PR #1029, opened 2022-03-25
feat: benchmark

### LINKED ISSUE #917: Built-in benchmarking support

Since we could write unit test in source files now with vitest I wish to also write some benchmarks like we do in rust/go. ~Deno cli will support bench command soon (https://github.com/denoland/deno/pull/13713)~ Deno added `deno bench` subcommand recently in [1.20.0 release](https://deno.com/blog/v1.20#new-subcommand-deno-bench).

In Node.js most of the project uses a library called [benchmark](https://www.npmjs.com/package/benchmark) whose last commit was 5 years ago. It works pretty well honestly but things changed a lot since then especially we are now suffering from the pain of esm/cjs interop. With vitest it could provide out-of-box ESM or even TypeScript support to bring a much more friendly dx.

### Suggested solution

Rust uses bench attribute and `--bench` flag should be passed in or just `cargo bench`  to run it:

```rust
// function to benchmark must be annotated with `#[bench]`
#[bench]
fn recursive_fibonacci(b: &mut Bencher) {
    // exact code to benchmark must be passed as a closure to the iter
    // method of Bencher
    b.iter(|| {
        (0..BENCH_SIZE).map(fibonacci).collect::<Vec<u32>>()
    })
}

#[bench]
fn iterative_fibonacci(b: &mut Bencher) {
    b.iter(|| {
        fibonacci_sequence().take(BENCH_SIZE).collect::<Vec<u32>>()
    })
}
```

Benchmarking is similar to unit testing in Golang, run it with `go test -bench=.`:

```go
// from fib_test.go
func BenchmarkFib10(b *testing.B) {
        // run the Fib function b.N times
        for n := 0; n < b.N; n++ {
                Fib(10)
        }
}
```

In deno, see https://github.com/denoland/deno/pull/13713:

```js
// Defaults to 1000 warmup iterations and 1000 bench iterations
Deno.bench("URL parsing", { n: 250, warmup: 100 }, () => {
  new URL("https://deno.land/");  
});


Deno.bench("URL resolving", { n: 1000 }, () => {
  new URL("./foo.js", import.meta.url);
});
```

```sh
$ deno bench --unstable example_bench.ts
running 2 benches from file:///Users/ib/dev/deno/example_bench.ts
bench URL parsing ... 250 iterations 23,063 ns/iter (208..356,041 ns/iter) ok (1s)
bench URL resolving ... 1000 iterations 48,045 ns/iter (250..128,625 ns/iter) ok (4s)

bench result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (5s)
```

My first thought was to provide a `benchmark` function, it works just like `describe` except it only runs with a bench flag:

```js
import { benchmark } from 'vitest'

benchmark('fib', (bench) => {
  bench('recursive', () => {
    recursiveFibonacci()
  })
  bench('iterative', () => {
    iterativeFibonacci()
  })
})
```

Also the output could be much prettier and modern comparing to `benchmark` package.

### Alternative

_No response_

### Additional context

_No response_

### Validations

poyoho:
fix #917

native support benchmark.

* Create a new `bench` command to start vitest and distinguish the current running mode of vitest (unit test or benchmark).
* Reuse the Vitest suite collection benchmark, add the `bench` API to distinguish between unit test and benchmark and collect the benchmark execution function.
* When running the test, unit-test normally enters the unit-test branch and executes the unit-test function once. And benchmark will use `Tinybench@2.1.3` as engine asynchronous multiple execution function.
* provide verbose and json reporter to show the benchmark result.


packages/vitest/package.json
```diff
@@ -85,6 +85,7 @@
   "dependencies": {
     "@types/chai": "^4.3.0",
     "@types/chai-subset": "^1.3.3",
+    "benchmark": "^2.1.4",
```
antfu, line 88: I haven't looked close to it but seems this package is no longer maintained for 5 years already. Do we have a better alternative?

poyoho: I saw the issues recommendation and didn't look for other 🙈

poyoho: @antfu There seems to be no better alternative https://www.npmjs.com/search?q=keywords:benchmark. I think it is possible to fork `benchmark` to maintain a new one?

antfu:
Let's discuss the design in #917 first.

About benchmark.js, if we really end up using it, I could try to contact the author and see if our team could grant the access and help on maintaining it.

packages/vitest/package.json
thelinuxlich: There is a fork by @Uzlopak actively updated in https://github.com/cthulhu-node/benchmark.js might be worthy joining forces

Uzlopak: I am now back from a three weeks vacation. I still want to implement the ability to automatically detect async functions and add the one PR to the original project which was fixing the statistics.
I did not publish a package yet. 
But we can surely join forces. :)

poyoho:
It seems need to make benchmark(`xxxx.bench.js`) whit diff files and add a new command to run benchmark 👏

sheremet-va:
I think we can now use [tinybench](https://github.com/tinylibs/tinybench) by @Uzlopak?

Uzlopak:
Published version 1.0.2 of tinybench.

poyoho:
image 
😁

Uzlopak:
Created a PR against github-action-benchmark. Does the color have some effect on the stdout and parsing by benchmark.js?

https://github.com/benchmark-action/github-action-benchmark/pull/126

test/benchmark/test/benchmark.test.ts
```diff
@@ -0,0 +1,21 @@
...
+
+describe('sort', () => {
+  benchmark('normal', () => {
+    const x = [1, 5, 4, 2, 3]
+    x.sort((a, b) => {
+      return a - b
+    })
+  })
+
+  benchmark('reverse', () => {
```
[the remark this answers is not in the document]
poyoho: oh, you make the `fn` to be promise?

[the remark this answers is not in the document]
Uzlopak: There is a wrapper for async functions in tinybench.

This code snippet just an example to see that tinybench is not going crazy and calling the async function a gazillion times. So you should see that async makes about 100 ops/s. 

Theoretically also your reverse example could `benchmark('reverse', async () => {` and should get also about 5.6 mio. ops/s.

packages/vitest/src/types/benchmark.ts
```diff
@@ -0,0 +1,60 @@
...
+  hz: number
+  rme: number
+  sampleSize: number
+}
+
+export type BenchmarkOptions = TinyBench.Options
+export type BenchFunction = (this: TinyBench.Suite, deferred: TinyBench.Deferred) => void
+export type BenchmarkAPI = ChainableFunction<
+'skip',
+[name: string, fn?: BenchFunction, options?: BenchmarkOptions],
```
Uzlopak, line 55:
```suggestion
[name: string, fn: BenchFunction, options?: BenchmarkOptions],
```

Why is fn optional? We need always a benchmark function

Uzlopak:
I realized that the async functionality is not working well.

packages/vitest/src/runtime/run.ts
```diff
@@ -237,10 +242,89 @@ export async function runSuite(suite: Suite) {
...
+    benchmarkLib.on('complete', () => {
+      suite.result!.duration = performance.now() - start
+      suite.result!.state = 'pass'
+      updateTask(suite)
+      defer.resolve(null)
+    })
+    benchmarkLib.on('error', (e) => {
+      defer.reject(e)
+    })
+    benchmarkLib.run({ async: true })
```
Uzlopak, line 318: also add `delay: 0`

poyoho: `benchmarkLib.run({ async: true, delay: 0 })` ?

Uzlopak: yeah

mangs:
Hmm, looking at the screenshot, average time per run is not mentioned, there is no breakdown by percentile, and which test was fastest is not mentioned; this can be important if you're comparing 4+ things. Can that be added?

I strongly suggest taking a look at [mitata](https://github.com/evanwashere/mitata) before this PR gets merged. This is how it looks integrated into [Deno](https://deno.land/) (it's what Deno uses for the `deno bench` command): 
image

packages/vitest/src/node/cli-api.ts
```diff
@@ -24,12 +24,17 @@ export async function startVitest(cliFilters: string[], options: CliOptions, vit
...
+  if (typeof options.coverage === 'boolean' && !options.benchmark)
     options.coverage = { enabled: options.coverage }
 
   const ctx = await createVitest(options, viteOverrides)
 
-  if (ctx.config.coverage.enabled) {
+  process.env.VITEST_MODE = ctx.config.watch ? 'WATCH' : 'RUN'
+
+  if (ctx.config.env)
+    Object.assign(process.env, ctx.config.env)
```
antfu, line 35: These lines have been moved to pool.ts

antfu:
I guess we should log an experiment warning when ppl use this command, addressing it "API might change in the future"

poyoho:
@mangs you mean deno use mitata?

mangs:
Hi @poyoho. The percentiles show outliers which can be important when tracking down performance issues which can get masked by averages. This is a very contrived example, but if you have 11 runs showing an average of 5 seconds, you can get that result with five items being zero, five 10, and one 5; percentiles would make that very clear. I think the function you found would do the trick and would be great! Thank you for considering my idea. As a big fan of what this project is doing, I don't mean to be a bother, I just want this feature to be as good as it can be. 😀

sheremet-va:
Sorry, I noticed with `rebase` came a lot of `@jest` packages. Is this intentional?

poyoho:
image

I think now log data is more, and the detailed data need to log in another way. so the pxx data I want to report in the file. 😁

Uzlopak:
Thank you for the link to the mitata repo. I analyzed the code and tbh I think their code is much more cleaner than tinybench/benchmark.js can ever be. Maybe we should really consider switching to mitata. Also I assume that as it is integrated into deno it will have much more traction in the development than tinybench will ever be.

Uzlopak:
@evanwashere

Maybe you want to add your thoughts to this feature?

poyoho:
@Uzlopak so Can I switch the tinybench to mitata ?

sheremet-va:
Do we have news on this PR? @Aslemammad you said, you wanted to take a look?

Aslemammad:
I'll take a look, I'm trying to implement a typesafe tinybench and modern API for it (based on the current)! and see if I can something out of it!

@poyoho could you please let me know what you like about other testing libs over tinybench, I'd like to add those features in the new huge refactor.

poyoho:
@Aslemammad sure! I think tinybench can export some runtime hook, so we can make the log to be  beautiful and show more runtime info. And mitata is seems just run the `fn` and show the benchmark info. 

tinybench can calculate many benchmark results based on his exposed hook. some like `mitata` I think we can even make it in vitest. It look like very easy 😁

Aslemammad:
@poyoho That's great, let's have a chat about it then. here's my discord: Aslemammad#6670

Aslemammad:
@poyoho The latest version of tinybench is `2.x.x` and the `1.x.x` is for the previous version which was based on benchmark.js.

poyoho:
I know. Rebase first 😁

poyoho:
image

@Aslemammad 🎉🎉🎉

Aslemammad:
@poyoho Happy to see this result! Thank you so much for the work. how about changing the `benchmark` function name to `bench` (in the benchmark.test.ts)? That'd be an easier naming in my opinion.

poyoho:
`benchmark` -> `bench`

packages/vitest/src/runtime/map.ts
```diff
@@ -26,9 +25,13 @@ export function isTest(task: Test | Benchmark): task is Test {
   return task.type === 'test'
 }
 
-export function getBenchmarkFactory(key: Suite): BenchFactory {
+export async function getBenchmarkFactory(key: Suite): Promise<BenchFactory> {
   let benchmark = benchmarkMap.get(key)
   if (!benchmark) {
+    if (!globalThis.EventTarget)
+      await import('event-target-polyfill' as any)
```
Aslemammad, line 32: Please consider putting the `any` after the import call!

poyoho: putting the any after the import call can to fix the types error 🤦

Uzlopak:
Just thought for improvements:

Consider running each benchmark in a separate process. So maybe it makes sense to recommend one benchmark per .bench.js/ts file. 

Also maybe consider adding a "benchmark normalizer". E.g. run first a benchmark calculating the fibonacci numbers and use the ops/s as 100%. So if you run the benchmarks on different machines, the benchmarks get more comparable.

Aslemammad:
What I also suggest is that we should promote warming up, it can help with having better and more accurate benchmarks, which is easily possible with tinybench.

Uzlopak:
Afaik warmup is done by default by benchmark.js. Maybe we should do it also automatically

Aslemammad:
@Uzlopak I guess having as a function in Tinybench (as it is right now) is better, so folks can choose to do it or not!

antfu:
I guess maybe we could have a `.bench.ts` convert to differentiated from testing

poyoho:
done!

antfu:
Pushed some changes,

- Introduce run mode. We should not rely on `config.benchmark` as it will present if users set the config regardless of running the command or not.
- Throw a hard error on calling `test()` in benchmark mode and vice versa. We could test that in `test/fails` later.

antfu:
image

@poyoho the ops/sec are identical, there might be something wrong

sheremet-va:
> image
> 
> @poyoho the ops/sec are identical, there might be something wrong

Even looking at this, it seems we show a lot less useful information than [Deno does](https://github.com/vitest-dev/vitest/pull/1029#issuecomment-1146744979)? (no percentile, for example) Didn't we want to take a look it this?

Uzlopak:
I guess we need to calculate ttables and utables for p75 and p995

Aslemammad:
@Uzlopak Tinybench already returns those percentiles https://github.com/tinylibs/tinybench#taskresult

antfu:
We probably need to think more about how we present the result for sure.

antfu:
We already support the reporter interface to access the full test result.

poyoho:
In fact, we had the results, but this style can't fill too much data 😁

poyoho:
> image
> 
> @poyoho the ops/sec are identical, there might be something wrong

two decimal places will be rounded off. The real result is:  0.006637481844709122, 0.005769233822295151...

poyoho:
image
updated

poyoho:
I think we can native support deno style benchmark result.

antfu:
For ops/sec <1, should we display the duration of one op instead?

poyoho:
maybe we can change ops/sec -> ops/ms

Uzlopak:
I dont think, that ops/ms is that relevant. I personally run benchmarks multiple times because it also depends on which core you run the benchmark and if there is currently a heavier process running on it. So basically benchmarks only indicate if you have some heavy performance degredation. So I would think it is totally OK to have the ops/s. Also you have the percentage on how they differ, so you can actually see which one is faster.

Is it possible to pass options to the benchmark like

```js
describe('sort', { minSample: 100 }, () => {
  bench('normal', () => {
    const x = [1, 5, 4, 2, 3]
    x.sort((a, b) => {
      return a - b
    })
  })
})
```

poyoho:
image

feat json format report to show all the result

sheremet-va:
What do we need to release this?

antfu:
I would like to improve the default output to a table, but not blocking as we could improve it later. Since this is experimental, I think we could ship it in the next minor.

packages/vitest/src/node/core.ts
```diff
@@ -101,7 +103,9 @@ export class Vitest {
       })
     }
 
-    this.reporters = await createReporters(resolved.reporters, this.runner)
+    this.reporters = resolved.benchmark
```
sheremet-va, line 106: shouldn't this be decided based on `mode`?

sheremet-va (changes requested):

packages/vitest/src/node/core.ts
antfu: Nice catch!

sheremet-va: I don't think `resolved.mode` is the same as VitestMode (`'test' | 'benchmark'`), it's custom user mode (`--mode`)

sheremet-va: We now have three modes 😄 

- `'WATCH' | 'RUN'`
- `'test' | 'benchmark'`
- custom, default `test` - btw, should we change it to `benchmark`, if Vitest is running as `vitest bench`?

packages/vitest/src/runtime/run.ts
```diff
@@ -230,23 +231,27 @@ export async function runSuite(suite: Suite) {
...
-        }
-        else {
-          const { sequence } = workerState.config
-          if (sequence.shuffle || suite.shuffle) {
-            // run describe block independently from tests
-            const suites = tasksGroup.filter(group => group.type === 'suite')
-            const tests = tasksGroup.filter(group => group.type === 'test')
-            const groups = shuffle([suites, tests], sequence.seed)
-            tasksGroup = groups.flatMap(group => shuffle(group, sequence.seed))
+      if (getRunMode()) {
```
sheremet-va, line 234: This looks weird 🤔 Shouldn't it be `isRunningInBenchmark`?

sheremet-va (changes requested):

packages/vitest/src/node/core.ts
sheremet-va: I see that now mode is always `'test' | 'benchmark'`, but it's still confusing, we will need to come up with better naming, I think :S

Still, I think we should pass down `benchmark` to `defineConfig` 🤔

packages/vitest/src/runtime/run.ts
sheremet-va: Yep, all tests were skipped, and always passed :)

PR #1031, opened 2022-03-26
feat: add parameter for test name of each

ouduidui:
I want to supplement a parameter about index of the test case for test name of `each`.

```ts
test.each([
  [1, 1, 2],
  [1, 2, 3],
  [2, 1, 3],
])('the index of the test case is %#', (a, b, expected) => {
  expect(a + b).toBe(expected)
})

// this will return
// √ the index of the test case is 0
// √ the index of the test case is 1
// √ the index of the test case is 2
```


patak-dev (approved):

packages/vitest/src/runtime/suite.ts
```diff
@@ -14,7 +14,10 @@ export const test = createTest(
   },
 )
 
-function formatTitle(template: string, items: any[]) {
+function formatTitle(template: string, items: any[], idx: number) {
+  // '%#' match index of the test case
+  template = template.replace(/\%\#/g, `${idx}`)
```
patak-dev, line 19: Thinking about it, we should check that the `%` isn't scaped here, no? To account for the case: `%%#`

Maybe it should be something like:
```
template.replace(/%%/g,`__vitest_escaped_%__`).replace(/\%\#/g, `${idx}`).replace(/__vitest_escaped_%__/g, '%%')
```

We could still merge this PR and review the escaping of `%` in general in a future PR, looking at the code below these edge cases may not be considered for other cases.

ouduidui: Oh, It's a detail what I ignored. 
Thanks for letting me learn something.

patak-dev: A nitpick, but we may also want to avoid doing this dance if not needed. So we could wrap it with `if (template.includes('%#')) {` so we only do it when there is a chance the replacement will be applied.

ouduidui: I thought about this before, At the time I thought `includes` and `replace` both need traverse the template that add a layer of judgment is redunant.

but now I think what you're saying makes sense.

patak-dev (approved):

Issue #1032, opened 2022-03-26
Show more details when no files found


togami2864:
### Clear and concise description of the problem

Currently vitest only displays 'No test files found,' when test files are not found. I sometimes forget that I set watchIgnore or include and get confused.
If vitest show more details, We will be able to avoid missing the config.

### Suggested solution

I implemented it in a local environment as a test.
- with passWithNoTests
<img width="574" alt="スクリーンショット 2022-03-26 19 41 53" src="https://user-images.githubusercontent.com/62130798/160237539-638e8322-e3d0-4215-b350-3867325d7d9f.png">

- without --passWithNoTests
<img width="521" alt="スクリーンショット 2022-03-26 19 41 26" src="https://user-images.githubusercontent.com/62130798/160237515-84ef7fdf-1574-413c-9e5c-90df623319c0.png">

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #1033, opened 2022-03-26
feat: show more details when no test files found

togami2864:
#1032 
Enable vitest to show 

- include
- watchIgnore

when no test files found.
screen shot

### DIFF
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -318,6 +318,27 @@ Isolate environment for each test file
 
 Coverage options
 
+### testNamePattern
+
+- **Type** `string | RegExp`
+
+Run tests with full names matching the pattern.
+If you add `do not run` to this property, tests containing the word `do not run` in the test name will be skipped.
+
+```js
+import { expect, it } from 'vitest'
+
+// run
+test('run', () => {
+  expect(true).toBe(true)
+})
+
+// skipped
+test('do not run', () => {
+  expect(true).toBe(true)
+})
+```
+
 ### open
 
 - **Type:** `boolean`
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -323,18 +323,18 @@ Coverage options
 - **Type** `string | RegExp`
 
 Run tests with full names matching the pattern.
-If you add `do not run` to this property, tests containing the word `do not run` in the test name will be skipped.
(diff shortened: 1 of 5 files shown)


packages/vitest/src/node/core.ts
```diff
@@ -114,12 +114,17 @@ export class Vitest {
     )
 
     if (!files.length) {
-      if (this.config.passWithNoTests)
-        this.log('No test files found\n')
+      const exitCode = this.config.passWithNoTests ? 0 : 1
+      if (this.config.passWithNoTests) { this.log(`No test files found, existing code with ${exitCode}\n`) }
```
dammy001, line 118:
```suggestion
      if (this.config.passWithNoTests) { this.log(`No test files found, exiting code with ${exitCode}\n`) }
```

docs/config/index.md
```diff
@@ -318,6 +318,27 @@ Isolate environment for each test file
 
 Coverage options
 
+### testNamePattern
+
+- **Type** `string | RegExp`
+
+Run tests with full names matching the pattern.
+If you add `do not run` to this property, tests containing the word `do not run` in the test name will be skipped.
```
sheremet-va, line 326: I feel like the opposite will happen? Only tests containing "do not run" will run, no?

packages/vitest/src/utils/base.ts
```diff
@@ -112,3 +112,11 @@ export function assertTypes(value: unknown, name: string, types: string[]): void
   const pass = types.includes(receivedType)
   if (!pass) throw new TypeError(`${name} value must be ${types.join(' or ')}, received "${receivedType}"`)
 }
+
+/**
+ * joinArrayItems
+ * convert Array<T> to a single string to display in stdout
+ */
+export function joinArrayItems(items: unknown[]): string {
```
sheremet-va, line 120: just out of interest, what is the difference between just calling `.join`? I see all test cases will return the same if called with `.join`

sheremet-va (changes requested):

packages/vitest/src/utils/base.ts
togami2864: Fixed to use `join` since `include` and `watchIgnore`  are absolutely the array. Thanks.

sheremet-va (approved):

PR #489, opened 2022-01-08
feat: json reporter

cexbrayat:
Fixes #108

Adds a JSON reporter, available with `--reporter json`.

It produces a JSON report of the test run in a similar format to the default Jest one (produced with `--json`, see https://jestjs.io/docs/cli#--json).

The Jest result format is defined [here](https://github.com/facebook/jest/blob/fe5e91c1b86323ab02b3de9d322b77e65a64e75a/packages/jest-test-result/src/types.ts#L58-L75).

Vitest JSON reporter uses a simplified version of this format:

```ts
interface TestResult {
  displayName?: string
  failureMessage?: string | null
  skipped: boolean
  status?: string
  testFilePath?: string
}
interface AggregatedResult {
  numFailedTests: number
  numFailedTestSuites: number
  numPassedTests: number
  numPassedTestSuites: number
  numPendingTests: number
  numTodoTests: number
  numPendingTestSuites: number
  numTotalTests: number
  numTotalTestSuites: number
  startTime: number
  success: boolean
  testResults: Array<TestResult>
}
```

PR #1034, opened 2022-03-27
fix: align json output with jest's

zxch3n:
In the current version(v0.7.11), the JSON reporter's output format deviates from jest's. See [Stackblitz example](https://stackblitz.com/edit/vitest-dev-vitest-gsjy9q?file=jest_output.json).

In PR https://github.com/vitest-dev/vitest/pull/489 json reporter is added, it implements the subset of jest's [AggregatedResult type](https://github.com/facebook/jest/blob/fe5e91c1b86323ab02b3de9d322b77e65a64e75a/packages/jest-test-result/src/types.ts#L58-L75). However, jest --json output is using [FormattedTestResults](https://github.com/facebook/jest/blob/fe5e91c1b86323ab02b3de9d322b77e65a64e75a/packages/jest-test-result/src/types.ts#L138-L154). 

And each object in `testResults` should represents a file rather than a test case.

### DIFF
--- a/packages/vitest/src/node/reporters/json.ts
+++ b/packages/vitest/src/node/reporters/json.ts
@@ -110,8 +110,11 @@ export class JsonReporter implements Reporter {
         } as FormattedAssertionResult
       })
 
-      if (tests.some(t => t.result?.state === 'run'))
-        this.ctx.console.warn('WARNING: Some tests are still running when generating the JSON report. This is likely an internal bug in Vitest.')
+      if (tests.some(t => t.result?.state === 'run')) {
+        this.ctx.console.warn('WARNING: Some tests are still running when generating the JSON report.'
+        + 'This is likely an internal bug in Vitest.'
+        + 'Please report it to https://github.com/vitest-dev/vitest/issues')
+      }
 
       testResults.push({
         assertionResults,
--- a/packages/vitest/src/node/reporters/json.ts
+++ b/packages/vitest/src/node/reporters/json.ts
@@ -1,38 +1,52 @@
 import { existsSync, promises as fs } from 'fs'
 import { dirname, resolve } from 'pathe'
 import type { Vitest } from '../../node'
-import type { File, Reporter } from '../../types'
+import type { File, Reporter, Suite, Test } from '../../types'
 import { getSuites, getTests } from '../../utils'
 
 // for compatibility reasons, the reporter produces a JSON similar to the one produced by the Jest JSON reporter
 // the following types are extracted from the Jest repository (and simplified)
+
+type Status = 'passed' | 'failed' | 'skipped' | 'pending' | 'todo' | 'disabled'
 type Milliseconds = number
-interface TestResult {
-  displayName?: string
-  failureMessage?: string | null
-  skipped: boolean
-  status?: string
-  testFilePath?: string
-  perfStats: {
-    end?: Milliseconds
-    runtime?: Milliseconds
-    start?: Milliseconds
-  }
+
+interface FormattedAssertionResult {
+  ancestorTitles: Array<string>
+  fullName: string
+  // location?: Callsite | null
+  status: Status
+  title: string
+  duration?: Milliseconds | null
 }
 
-interface AggregatedResult {
+interface FormattedTestResult {
+  message: string
+  name: string
+  summary: string
+  status: 'failed' | 'passed'
+  startTime: number
+  endTime: number
+  // coverage: unknown
+  assertionResults: Array<FormattedAssertionResult>
+}
+
+interface FormattedTestResults {
+  // coverageMap?: CoverageMap | null | undefined
   numFailedTests: number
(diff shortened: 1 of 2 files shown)


patak-dev:
Thanks for the PR @zxch3n! Things like status being `pass` instead of `passed` looks quite important to correct. Interesting that no one noticed this before. I wonder if tools consuming this format are able to accept both.

@cexbrayat would you help us review this PR and check that it won't be a breaking change for your use case?

packages/vitest/src/node/reporters/json.ts
```diff
@@ -60,20 +84,74 @@ export class JsonReporter implements Reporter {
...
+      const file = test.file
+      if (file) {
+        if (!fileToTestCases.has(file.filepath))
+          fileToTestCases.set(file.filepath, { file, tests: [] })
+
+        fileToTestCases.get(file.filepath)!.tests.push(test)
+      }
+    }
+
+    for (const [filepath, { tests, file }] of fileToTestCases) {
```
patak-dev, line 100: I think that we could avoid calling `getTests` and then reconstruct the file-to-tests map if we iterate by file here. See getTests that will give you all the tests flatten for a file
https://github.com/vitest-dev/vitest/blob/a1922a07c54bf3f3b2cd9e2f23376ba254f97bab/packages/vitest/src/utils/tasks.ts#L4

```js
for (const file of files) {
  const tests = getTests(file)
```

packages/vitest/src/node/reporters/json.ts
```diff
@@ -60,20 +84,74 @@ export class JsonReporter implements Reporter {
...
+      })
+
+      testResults.push({
+        assertionResults,
+        startTime,
+        endTime,
+        status: tests.every(t =>
+          t.result?.state === 'pass'
+           || t.result?.state === 'skip'
+           || t.result?.state === 'run'
```
patak-dev, line 132: Wouldn't be an internal error to still have a `run` state here?

packages/vitest/src/node/reporters/json.ts
zxch3n: Yeah. You're right

packages/vitest/src/node/reporters/json.ts
zxch3n: I'm not very familiar with the internal logic. Should I add a failure message in this case?

patak-dev: Maybe leave it as failed?

zxch3n: Should we throw an error here or append a failure message to the result? Or do anything to let the user know?

patak-dev: I think we could emit a warning and set it as failed in the output json.

packages/vitest/src/node/reporters/json.ts
```diff
@@ -47,33 +71,80 @@ export class JsonReporter implements Reporter {
...
-    const tests = getTests(files)
-    const numTotalTests = tests.length
+    const fileToTestCases = new Map<File, Test[]>()
 
+    for (const file of files) {
+      const relatedTests = getTests([file])
+      fileToTestCases.set(file, relatedTests)
+    }
+
+    const tests = [...fileToTestCases.values()].flat()
```
patak-dev, line 81: I think there is no problem in calling `getTests` twice. This line could remain as `const tests = getTests(files)` and the `fileToTestCases` map isn't needed then. You can directly use `for (const file of files) {` in line 94.

packages/vitest/src/node/reporters/json.ts
zxch3n: done

cexbrayat (approved):
LGTM, thanks @zxch3n 👍

krzysztofradomski:
Hey all. Just like to add wanted to weight in from a  hmm regular user's perspective ;)
In my current project we've recently switched from jest to vitest, very happy with the change, but the we could not consume the json output in Storybook... So I wrote a simple Storybook addon, which I published to npm just yesterday. This addon of mine depends on the current structure of the output file, and now today I've found this thread haha!

patak-dev:
> Hey all. Just like to add wanted to weight in from a hmm regular user's perspective ;) In my current project we've recently switched from jest to vitest, very happy with the change, but the we could not consume the json output in Storybook... So I wrote a simple Storybook addon, which I published to npm just yesterday. This addon of mine depends on the current structure of the output file, and now today I've found this thread haha!

😬 
Well, at least it should be only about removing the addon, no? Have you tried this PR? It would be good to get another confirmation that it is working as expected.

patak-dev (approved):

krzysztofradomski:
Hmm I might give this branch a try later today... Maybe I should just
update my addon - I was so proud of it, so not gonna ditch it so fast lol

pon., 28 mar 2022, 13:45 użytkownik patak ***@***.***>
napisał:

PR #1038, opened 2022-03-27
fix: update resolves docs

ouduidui:
I think there is something wrong that a sentence of resolves descrition .

```
Use it to unwrap value from pending promise and assert its value with usual assertions. 
```

Shouldn't we be able to wrap value when promise is fulfilled, instead of pending?

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -889,7 +889,7 @@ TODO
 
 - **Type:** `Promisify<Assertions>`
 
-  `resolves` is intended to remove boilerplate when asserting asynchronous code. Use it to unwrap value from pending promise and assert its value with usual assertions. If promise rejects, the assertion will fail.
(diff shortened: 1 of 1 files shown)


sheremet-va (changes requested):
Promise is `pending` if it was not fullfilled/rejected. Your sentence implies it should be called like this: 

```ts
expect(await promise).resolves.toBe('smth');
``` 

But if Promise will always be `pending` if it was passed just as is (because microtask wasn't fired yet), so current wording is more correct

ouduidui:
> Promise is `pending` if it was not fullfilled/rejected. Your sentence implies it should be called like this:
> 
> ```ts
> expect(await promise).resolves.toBe('smth');
> ```
> 
> But it Promise will always be `pending` if it was passed just as is (because microtask wasn't fired yet), so current wording is more correct

You're right, I'll have to think about it.

thanks.

PR #1039, opened 2022-03-27
fix(ui): open in IDE and error styles

userquin:
Moved the `useCodeError` logic to the `ViewEditor.vue` component from `error.ts` module, we would need to include `@unocss-include` on the `error.ts` module to style the elements created.

Anyway, on stackblitz I cannot make it working (missing links, it seems using oldest version), tested with `0.7.11` and `0.7.12` but still not working, it seems stackblitz doing strange things, for example: `taze` not showing `vitest 0.7.12`, I need to run `taze major`.

imagen

The repo I use with `0.7.12` with almost all core tests included: https://stackblitz.com/edit/vitejs-vite-nfjsbi

### DIFF
--- a/packages/ui/client/components/views/ViewEditor.vue
+++ b/packages/ui/client/components/views/ViewEditor.vue
@@ -28,13 +28,13 @@ const failed = computed(() => props.file?.tasks.filter(i => i.result?.state === 
 
 const widgets: CodeMirror.LineWidget[] = []
 const handles: CodeMirror.LineHandle[] = []
-const listeners: [el: HTMLSpanElement, l: Record<string, EventListener>, t: () => void][] = []
+const listeners: [el: HTMLSpanElement, l: EventListener, t: () => void][] = []
 
 const hasBeenEdited = ref(false)
 
 const clearListeners = () => {
   listeners.forEach(([el, l, t]) => {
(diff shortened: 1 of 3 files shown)


patak-dev:
When I tried locally on macOS, after making a test fail and pressing the `Open in IDE` I'm getting in the terminal (and no feedback in the app):

```
Could not open chainable.test.ts in the editor.
The editor process exited with an error: spawn code ENOENT.
```

Is there something else I need to setup so VS Code is picked up?

About the styles, could the icon color have more contrast so it is accessible?

image

We also have an `Open in Editor` button at the top-right, should both be called `Open in Editor` or `Open in IDE`?

userquin:
imagen

imagen

userquin:
@patak-dev the error comes from vite, vitest doesn't handle the `/__open-in-editor?file=` url, you can check the vite server middleware here: https://github.com/vitejs/vite/blob/main/packages/vite/src/node/server/index.ts#L528

Vite is using this library https://github.com/yyx990803/launch-editor from Evan

patak-dev (approved):

userquin:
We don't need tooltip listeners, the floating tooltip lib works also with focus and blur events

Issue #835, opened 2022-02-22
jsdom loading issue if threads false


bkuster:
### Describe the bug

# Issue
When starting vitest with `threads: false` tests fail with: `Cannot read properties of undefined (reading 'child')` The same test run while using `threads: true`.

# Possible Causes
The test uses the @vue/test-utils package for component testing. These _must_ be loaded after jsdom is loaded and the DOM is provided. (see [this issue](https://github.com/vuejs/vue-test-utils/issues/1192)).

# Failed workarounds
I tried loading jsdom via jsdom-global myself in a global setup hook, so far to no avail.

### Reproduction

# vite.config.js
```js
import { defineConfig } from 'vite';
import { createVuePlugin } from 'vite-plugin-vue2';

const configMain = defineConfig({
  plugins: [
    createVuePlugin(),
  ],
  test: {
    threads: false,
    environment: 'jsdom',
  },
});

export default configMain;
```

# Component.spec.js
```js
import { describe, expect, it } from 'vitest';
import { mount, createLocalVue } from '@vue/test-utils';
import Component from './Component.vue';

describe('Compontent', () => {
  const localVue = createLocalVue();

  it('mounts', () => {
    const wrapper = mount(Component, { localVue });
    expect(wrapper).toMatchSnapshot();
  });
});
```
# Component.vue
```vue
<template>
  <div>
    Foo Bar
  </div>
</template>

<script>
  export default {
    name: 'FooBar',
  };
</script>
```
### setup
```bash
npm i vitest vite vue@^2.6.0 @vue/test-utils@1 jsdom
```

### System Info

```shell
System:
    OS: Linux 5.4 Ubuntu 20.04.3 LTS (Focal Fossa)
    CPU: (6) x64 Intel(R) Core(TM) i7-8700 CPU @ 3.20GHz
    Memory: 643.27 MB / 31.36 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 16.13.0 - ~/.nvm/versions/node/v16.13.0/bin/node
    npm: 8.1.0 - ~/.nvm/versions/node/v16.13.0/bin/npm
  Browsers:
    Chromium: 98.0.4758.102
    Firefox: 97.0
  npmPackages:
    vite: ^2.8.4 => 2.8.4 
    vitest: ^0.5.3 => 0.5.3
```

### Used Package Manager

npm

### Validations

[...]

PR #1060, opened 2022-03-31
chore: repro --no-threads issues

cexbrayat:
Adds an additional step to the CI job to reproduce the --no-threads issues


antfu:
Can you share more context about this? Not all tests are capable of running in the same thread.

cexbrayat:
There are currently several issues when running tests with `--no-threads`:

- jsdom issues: it turns out that the teardown function of the jdsom env is never called in threaded mode. When threads are disabled, then the teardown function is called, and tests break
- some mocks and spies tests are failing as well, maybe for a similar reason (`pnpm test -- --no-threads`)

cexbrayat:
@antfu Just saw your comment. There is at least one issue with jsdom (and happy-dom probably) env: their teardown function is never called when threads are enabled if I'm not mistaken. That leads to behavior difference between threaded and non-threaded mode.

cexbrayat:
Related issue #835

Issue #1066, opened 2022-04-01
link to 404 in docs and readme


frenchcdsl44:
### Describe the bug

In the readme https://github.com/vitest-dev/vitest/tree/main/examples
In the docs https://vitest.dev/guide/#examples
the example link for testing-library-jest-dom points to a 404 error page
some other links might be broken

### Reproduction

https://github.com/vitest-dev/vitest/tree/main/examples

### System Info

```shell
chrome
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #1069, opened 2022-04-01
docs: update example links

dammy001:


dammy001:
closes #1066

userquin:
@dammy001 if the deployer has the old tests folders with the old examples and runs the `scripts/update-examples.ts` script, your changes will go straight to the trash

antfu:
Thanks, fixed in cadefe56

PR #1070, opened 2022-04-01
fix: inconsistency with globTestFiles and isTargetFile (fix #981, #982)

### LINKED ISSUE #981: file addition is not detected when using relative `include`
### Describe the bug

When a file is added the test reruns on watch mode.
But when `include` is relative like below, the rerun does not happen.
```ts
export default defineConfig({
  test: {
    include: ['test/**.test.ts'],
  },
});
```

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-b4spnq?file=vite.config.ts

1. Run `npm run test`.
2. Add `foo.test.ts` to `test` directory.
3. test does not rerun.

The rerun happens if you change `include` to `['**/test/**.test.ts']`.

### System Info

```shell
stackblitz

vitest 0.7.4
```

### Used Package Manager

npm

### Validations

### LINKED ISSUE #982: `dir` option is not respected with file addition detection
### Describe the bug

When there is a config like below, `test/foo.test.ts` will not run. (This is correct.)
But if you add `test/bar.test.ts` after running `vitest`, it gets included.
```ts
export default defineConfig({
  test: {
    dir: 'src',
    include: ['**/*.test.ts'],
  },
});
```

Maybe related to #981.

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-xrmrde?file=vite.config.ts&initialPath=__vitest__

1. Run `npm run test`.
2. See `src/bar.test.ts` is included and `test/bar.test.ts` **is not** included.
3. Add `test/foo.test.ts`.
4. Rerun will happen and `test/foo.test.ts` **is** included.

### System Info

```shell
stackblitz

vitest 0.7.4
```

### Used Package Manager

npm

### Validations

sapphi-red:
This PR fixes inconsistency with `globTestFiles` and `isTargetFile`.

fixes #981
fixes #982 

I actually tested with #981 and #982's reproduction and checked that it works.

### DIFF
--- a/packages/vitest/src/node/core.ts
+++ b/packages/vitest/src/node/core.ts
@@ -1,6 +1,6 @@
 import { existsSync, promises as fs } from 'fs'
 import type { ViteDevServer } from 'vite'
-import { toNamespacedPath } from 'pathe'
+import { toNamespacedPath, relative } from 'pathe'
 import fg from 'fast-glob'
 import mm from 'micromatch'
 import c from 'picocolors'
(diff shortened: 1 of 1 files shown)


packages/vitest/src/node/core.ts
```diff
@@ -429,12 +429,17 @@ export class Vitest {
     return testFiles
   }
 
+  private isMatchPattern(id: string, pattern: readonly string[]) {
+    const relativeId = relative(this.config.dir || this.config.root, id)
```
patak-dev, line 433: Is `this.config.dir` needed here? Isn't exclude/include/includeSource relative to the root?

```suggestion
    const relativeId = relative(this.config.root, id)
```

Also, maybe we want to directly make the path `relative` in isTargetFile, or before calling it (we are already calling slash), instead of creating a private function that is potentially called several times (and results in more complexity).

patak-dev (approved):

PR #1073, opened 2022-04-01
feat: extend `config.outputFile` to allow objects, close #1068

### LINKED ISSUE #1068: Multiple output-files / Allow extension of existing reporters

Before going to the problem I have to... Thank for this awesome lib, you're doing an insanely great job 👍🏿 

The problem:
I have the case where I need the json and junit reporter at once (one dataset goes into some documentary, the other to gitlab), when using both they merge their results into the specified `outputFile`. Since I cannot access the default reporter classes (https://github.com/vitest-dev/vitest/blob/main/packages/vitest/src/node/reporters/json.ts f.e.) to overwrite some stuff I'd need to create a custom one by merging junit & json reporters which seems kinda cubersome.

(sidenote: if the FR is not declined I'd try to contribute to this ofc, if I may :P)

### Suggested solution

If `outputFile` would accept a Record in a format like `{ [reporter]: outputFilePath }` (besides the string, the simple solution should ofc remain for the simple cases :D) the reporters could be run in parallel (would result in code like `this.ctx.config.outputFile.json || this.ctx.config.outputFile` or `typeof this.ctx.config.outputFile === 'string' ? this.ctx.config.outputFile  : this.ctx.config.outputFile.json` or sth).

### Alternative

By exposing the default reporters somehow they could be overwritten (would result in code like `class MyCustomReporter extends JsonReporter { ... }` where the ctx could be overwritten at a certain point).

### Additional context

_No response_

### Validations

steve-py96:
related #1068

### DIFF
--- a/test/reporters/tests/reporters.spec.ts
+++ b/test/reporters/tests/reporters.spec.ts
@@ -57,12 +57,60 @@ test('JUnit reporter', async() => {
   expect(context.output).toMatchSnapshot()
 })
 
+test('JUnit reporter (no outputFile entry)', async() => {
+  // Arrange
+  const reporter = new JUnitReporter()
+  const context = getContext()
+  context.vitest.config.outputFile = {}
+
+  vi.mock('os', () => ({
+    hostname: () => 'hostname',
+  }))
+
+  vi.setSystemTime(1642587001759)
+
+  // Act
+  await reporter.onInit(context.vitest)
+  await reporter.onFinished(files)
+
+  // Assert
+  expect(context.output).toMatchSnapshot()
+})
+
 test('JUnit reporter with outputFile', async() => {
   // Arrange
   const reporter = new JUnitReporter()
   const outputFile = resolve('report.xml')
   const context = getContext()
   context.vitest.config.outputFile = outputFile
+
+  vi.mock('os', () => ({
+    hostname: () => 'hostname',
+  }))
+
+  vi.setSystemTime(1642587001759)
+
+  // Act
+  await reporter.onInit(context.vitest)
+  await reporter.onFinished(files)
+
+  // Assert
+  expect(normalizeCwd(context.output)).toMatchSnapshot()
+  expect(existsSync(outputFile)).toBe(true)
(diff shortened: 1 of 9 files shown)


steve-py96:
note: I had (PR unrelated) errors when running `nr test:all` (on the `main` without changes)

errors:
```
test/fails test: ⎯⎯⎯⎯⎯⎯⎯ Failed Tests 6 ⎯⎯⎯⎯⎯⎯⎯
    test/fails test:  FAIL  test/runner.test.ts > should fails > empty.test.ts
    test/fails test:  FAIL  test/runner.test.ts > should fails > expect.test.ts
    test/fails test:  FAIL  test/runner.test.ts > should fails > hook-timeout.test.ts
    test/fails test:  FAIL  test/runner.test.ts > should fails > nested-suite.test.ts
    test/fails test:  FAIL  test/runner.test.ts > should fails > stall.test.ts
    test/fails test:  FAIL  test/runner.test.ts > should fails > timeout.test.ts
```

test/reporters/tests/reporters.spec.ts
```diff
@@ -83,6 +103,34 @@ test('JUnit reporter with outputFile', async() => {
   rmSync(outputFile)
 })
 
+test('JUnit reporter with outputFile object', async() => {
+  // Arrange
+  const reporter = new JUnitReporter()
+  const outputFile = resolve('report.xml')
```
AriPerkkio, line 109: Each of the `outputFile`'s should have unique names in tests. Also the `rootDirectory`'s should be unique. 
Earlier I saw that when tests were run on CI, some tests were locking the files when removing them, while other tests were trying to create them. Easy solution was to use unique names in each test so that they don't conflict.

steve-py96: makes sense tbh, gonna change that

steve-py96: fixed it, thx for the hint :D

sheremet-va (approved):

PR #1074, opened 2022-04-01
chore: don't publish sourcemaps

aleclarson:
Avoid publishing inline sourcemaps to npm, just like Vite

### DIFF
--- a/packages/vite-node/package.json
+++ b/packages/vite-node/package.json
@@ -58,7 +58,7 @@
   },
   "scripts": {
     "build": "rimraf dist && rollup -c",
-    "dev": "rollup -c --watch --watch.include=src/**",
+    "dev": "rollup -c --watch --watch.include=src/** -m inline",
     "prepublishOnly": "nr build",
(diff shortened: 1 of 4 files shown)


sheremet-va (changes requested):
What about other `packages`? They all have their own config

patak-dev (approved):

aleclarson:
ok I think that's all of them

sheremet-va (approved):

antfu:
What's the benefit of not publishing it? I actually did that on purpose.

aleclarson:
@antfu It increases the install size by a lot. Vite also used to publish sourcemaps, but stopped. Anyone debugging Vitest can fork the repo and build it locally.

patak-dev:
For reference, it saved 5MB (Vite is now at 18MB) when they were removed in Vite https://github.com/vitejs/vite/commit/1c1f82f93bc49ce3a94e23fe99eff377be691b60

PR #1078, opened 2022-04-02
feat: add commands to run tests depending on changed files

sheremet-va:
The code for getting files is based on jest implementation, but some of it was removed. Not sure how it will be working, if root and git is not the same? If someone can test in a monorepo, that would be cool. I am ashamed to admit, I am not familiar enough with git terminal commands 👀 

I was testing in in `tests/core` by changing code in `/src` and it worked fine. So i guess the solution should work for most people, using git.

I am getting the changed files and putting it into `related` array. Then the logic for related kicks in, and it tries to find tests based on related files. I have a guard `if (!related) { // find related }` in case `runTests` runs twice (bc of watch mode or else)

### DIFF
--- a/packages/vitest/src/node/cli.ts
+++ b/packages/vitest/src/node/cli.ts
@@ -33,6 +33,9 @@ cli
   .option('--environment <env>', 'runner environment (default: node)')
   .option('--passWithNoTests', 'pass when no tests found')
   .option('--allowOnly', 'Allow tests and suites that are marked as only (default: !process.env.CI)')
+  .option('--onlyChanged', 'Run tests that are affected by the changed files (default: false)')
+  .option('--lastCommit', 'Run tests that are affected by files changed in the last commit (default: false)')
+  .option('--changedSince <commit>', 'Run tests that are affected by files changed since commit hash or branch')
   .help()
 
 cli
--- a/packages/vitest/src/node/cli.ts
+++ b/packages/vitest/src/node/cli.ts
@@ -33,9 +33,7 @@ cli
   .option('--environment <env>', 'runner environment (default: node)')
   .option('--passWithNoTests', 'pass when no tests found')
   .option('--allowOnly', 'Allow tests and suites that are marked as only (default: !process.env.CI)')
-  .option('--onlyChanged', 'Run tests that are affected by the changed files (default: false)')
-  .option('--lastCommit', 'Run tests that are affected by files changed in the last commit (default: false)')
-  .option('--changedSince <commit>', 'Run tests that are affected by files changed since commit hash or branch')
+  .option('--changed [since]', 'Run tests that are affected by the changed files (default: false)')
   .help()
 
 cli
--- a/packages/vitest/src/node/git.ts
+++ b/packages/vitest/src/node/git.ts
@@ -1,0 +1,95 @@
+import { resolve } from 'pathe'
+import { execa } from 'execa'
+import type { ExecaReturnValue } from 'execa'
+
+export interface GitOptions {
+  lastCommit?: boolean
+  changedSince?: string
+}
+
+export class VitestGit {
+  private async resolveFilesWithGitCommand(
+    args: string[],
+    cwd: string,
+  ): Promise<string[]> {
+    let result: ExecaReturnValue
+
+    try {
+      result = await execa('git', args, { cwd })
+    }
+    catch (e: any) {
+      e.message = e.stderr
+
+      throw e
+    }
+
+    return result.stdout
+      .split('\n')
+      .filter(s => s !== '')
+      .map(changedPath => resolve(cwd, changedPath))
+  }
+
+  async findChangedFiles(cwd: string, options: GitOptions) {
+    const changedSince = options.changedSince
+
+    if (options && options.lastCommit) {
+      return this.resolveFilesWithGitCommand(
+        ['show', '--name-only', '--pretty=format:', 'HEAD', '--'],
+        cwd,
+      )
+    }
+    if (changedSince) {
+      const [committed, staged, unstaged] = await Promise.all([
+        this.resolveFilesWithGitCommand(
+          ['diff', '--name-only', `${changedSince}...HEAD`, '--'],
+          cwd,
+        ),
+        this.resolveFilesWithGitCommand(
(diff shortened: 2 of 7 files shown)


sheremet-va:
Don't know how to write tests for this :S

Maybe by mocking `exec`? But all it does is call it, so what would be the point 🤔

packages/vitest/src/node/cli.ts
```diff
@@ -33,6 +33,9 @@ cli
   .option('--environment <env>', 'runner environment (default: node)')
   .option('--passWithNoTests', 'pass when no tests found')
   .option('--allowOnly', 'Allow tests and suites that are marked as only (default: !process.env.CI)')
+  .option('--onlyChanged', 'Run tests that are affected by the changed files (default: false)')
+  .option('--lastCommit', 'Run tests that are affected by files changed in the last commit (default: false)')
+  .option('--changedSince <commit>', 'Run tests that are affected by files changed since commit hash or branch')
```
antfu, line 38: This feels a bit duplicated to me, do you think we could have a single command like `--changedSince [commit]` with a convention like `HEAD^3` or even simpler like `-3` for the previous 3 commits?

packages/vitest/src/node/git.ts
```diff
@@ -0,0 +1,95 @@
+import { resolve } from 'pathe'
+import { execa } from 'execa'
+import type { ExecaReturnValue } from 'execa'
```
antfu, line 3: We could use `simple-git` here

sheremet-va: I didn't want to bring other dependencies here, since what I need can be achieved easily with a simple command and `split('\n')`, so not sure it's justified

packages/vitest/src/node/cli.ts
sheremet-va: I would expect people to use actual hashes? I think testing by 'previous # commits' is not really an issue? You will run this only in CI, and CI usually have some global variables that contain hashes you can use, like `GITHUB_SHA`

What exactly do you find repetitive?

packages/vitest/src/node/git.ts
antfu: Sure. I am just not confident about maintaining it, since it might be a bit hard to test. No strong opinion on this.

sheremet-va: Are you afraid git will abandon support for this commands? I was looking at jest code for this, and they haven't changed it for 5 years 👀

packages/vitest/src/node/cli.ts
sheremet-va: I think `--changedSince` is the only one that really should be run in CI

`--onlyChanged` will run tests against files that are just that - changed, not commited or anything
`--lastCommit` is also to run it maybe on hook or if you forgot to test with `--onlyChanged`

packages/vitest/src/types/config.ts
```diff
@@ -329,6 +329,26 @@ export interface UserConfig extends InlineConfig {
...
+
+  /**
+   * Runs tests that are affected by the changes in the last commit
+   * Requires initialized git repository
+   * @default false
+   */
+  lastCommit?: boolean
+
+  /**
+   * Runs tests that are affected by the changes since the specified branch or commit hash
```
sheremet-va, line 348:
```suggestion
   * Runs tests that are affected by the changes between the specified branch or commit hash
```

packages/vitest/src/node/git.ts
antfu: No.. I mean, in different OS, or different git versions (or even global settings), it might emit different outputs. But I might be overthinking.

sheremet-va: I think this is fine until we don't have reports from users.

packages/vitest/src/node/cli.ts
antfu: I am still not sure if to have 3 options to do the same thing. It's also not clear to me if and how I should use them together.

I am thinking it like:

- `--changed` - from the latest commit, for unstaged changes. If the stage is clear, for the previous commit.
- `--changed ^3` - changes in the previous 3 commits (convention could be discussed)
- `--changed 5963a39` - since SHA
- `--changed main` - changes against main branch (or tags)

sheremet-va: I don't see a profit in cramming different usecases into a single option. Wouldn't it increase code heaviness? Also your suggestions solution makes a lot of assumptions, and I need a book to understand it (like running `--changed` without options)

I think having 3 different options is more approachable for users (I disagree they do the same thing). Especially coming from jest. Also I don't want to fix what is not broken (like previous 3 commits - who need this? what its usecase?)

For me you solution looks like instead of having 3 different functions with distinct names use one - I really don't see how this is better?

sheremet-va: > It's also not clear to me if and how I should use them together.

Well, you wouldn't. `--onlyChanged` can be used with both, it will be overriden.

But we also have `--watch` and `--run` options that you can't use at the same time.

antfu: What's behaviors would it be with?

- `vitest --lastCommit --changedSince fff`
- `vitest --lastCommit`
- `vitest --onlyChanged --lastCommit --changedSince fff`

3 options you got 8 different combinations (that's a multiplier if you have other options).

> I need a book to understand it (like running --changed without options)

It's common to have optional arguments as we had for `--api [api]`. I don't understand what is confusing.

sheremet-va: > It's common to have optional arguments as we had for `--api [api]`. I don't understand what is confusing.

It's not options that are confusing. Described behaviour is - why would it behave differently depending on staged changes?

> What's behaviors would it be with?

What's important (specific) will override what's less important (specific)

sheremet-va: Ok, I researched a little bit more, and what about:

- `--changed` - runs tests against unstaged changes
- `--changed hash/branch/HEAD~#` - runs against hash/branch (command is the same)

Documentation can describe how to do common tasks (like from last, use `HEAD`, `HEAD~1`)

antfu: Yeah, that's what I was proposing.

PR #1081, opened 2022-04-02
feat: reprint all tests, if one of the test files was deleted

### LINKED ISSUE #1005: Test status doesn't update if test file is deleted
### Describe the bug

If I run Vitest in watch mode, it reports the number of tests passing or failing as expected. But if I delete a test file, the said numbers don't update.

### Reproduction

Have two test files, say `a.test.js` and `b.test.js`. Each of them have one failing test, so we see 2 suites failing, 2 tests failing.

Now, we delete `b.test.js`. The status stays the same: 2 suites failing, 2 tests failing.

After quitting and restarting Vitest, it reports the correct count.

### System Info

```shell
System:
    OS: macOS 12.3
    CPU: (8) x64 Intel(R) Core(TM) i5-8257U CPU @ 1.40GHz
    Memory: 143.12 MB / 8.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.14.0 - ~/.volta/tools/image/node/16.14.0/bin/node
    npm: 8.3.1 - ~/.volta/tools/image/node/16.14.0/bin/npm
    Watchman: 2022.03.14.00 - /usr/local/bin/watchman
  Browsers:
    Firefox: 98.0.1
    Safari: 15.4
  npmPackages:
    @vitejs/plugin-react: ^1.0.7 => 1.2.0
    vite: ^2.8.0 => 2.8.6
    vitest: ^0.7.0 => 0.7.0
```

### Used Package Manager

npm

### Validations

sheremet-va:
Closes #1005

Debatable solution. I sometimes also was confused when after deleted a file, nothing happend. Maybe at least _something_ should happen?

### DIFF
--- a/packages/vitest/src/node/core.ts
+++ b/packages/vitest/src/node/core.ts
@@ -211,10 +211,6 @@ export class Vitest {
     return await this.runningPromise
   }
 
-  async reprintFiles(trigger?: string) {
-    await this.report('onReprint', trigger)
-  }
-
   async rerunFiles(files: string[] = this.state.getFilepaths(), trigger?: string) {
     await this.report('onWatcherRerun', files, trigger)
     await this.runFiles(files)
@@ -320,7 +316,7 @@ export class Vitest {
       if (this.state.filesMap.has(id)) {
         this.state.filesMap.delete(id)
         this.changedTests.delete(id)
-        this.reprintFiles(id)
+        this.report('onTestRemoved', id)
       }
     }
     const onAdd = async(id: string) => {
--- a/packages/vitest/src/node/core.ts
+++ b/packages/vitest/src/node/core.ts
@@ -320,7 +320,6 @@ export class Vitest {
       if (this.state.filesMap.has(id)) {
         this.state.filesMap.delete(id)
         this.changedTests.delete(id)
-        // rerun tests to update console output
         this.reprintFiles(id)
       }
     }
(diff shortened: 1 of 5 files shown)


antfu:
If you delete a source file, it should trigger the rerun for tests deps on it. but if you delete a test file, it should be ...deleted, isn't it? I think you can always rerun the tests by shortcuts. Consider in some projects they might have a lot of tests, delete to trigger the full rerun sounds a bit weird behavior to me.

Or maybe if you really need some feedback, we could simply print `xxxx.test.ts has been deleted`. WDYT?

sheremet-va:
> If you delete a source file, it should trigger the rerun for tests deps on it. but if you delete a test file, it should be ...deleted, isn't it? I think you can always rerun the tests by shortcuts. Consider in some projects they might have a lot of tests, delete to trigger the full rerun sounds a bit weird behavior to me.
> 
> Or maybe if you really need some feedback, we could simply print `xxxx.test.ts has been deleted`. WDYT?

I agree it looks weird. But if you consider testing as a continues process, it also feels weird that nothing happend, you know? Like, imagine you are writing tests and you see an output, that 3 tests passed, 1 test failed - then you delete a file with failed test, but nothing happens? Feels disjointed that it doesn't react.

But it would be an overkill, if I test one specific file, then delete another - then one of a sudden I run all tests? Also feels weird.

antfu:
Or instead of rerunning, we could remove the result of that file, and print the report again.

antfu (approved):

PR #1088, opened 2022-04-03
fix(logs): collect logs per test

userquin:
For context: 

The log spy currently uses a single buffer to collect all the logs from all the tests, when sending the logs to the reporters, there is no way of knowing which tests each of the logs correspond to, so as soon as are sent, all the logs that are collected in the corresponding tick are shown as corresponding to the current test:

imagen

This PR also separates the timers by tests, since otherwise we can lose logs:

imagen

PR #1086 by @sheremet-va fixed the identifier of the test to which the collected logs belong, but when they are sent we still have the problem that all the ones in the buffer are sent, so the logs of other tests will be included under the running test at tick.

This PR includes the current logic but separated by test, so that the logs are collected and included in their own buffer and thus, when sent to the reporters the logs are shown where they belong:

imagen

and for ui (the tab title on console should be 4, fixed also in this PR):

imagen

### DIFF
--- a/packages/ui/client/components/FileDetails.vue
+++ b/packages/ui/client/components/FileDetails.vue
@@ -29,6 +29,9 @@ const open = () => {
 const changeViewMode = (view: Params['view']) => {
   viewMode.value = view
 }
+const count = computed(() => {
+  return currentLogs.value?.map(l => l.size).reduce((s, a) => s + a, 0) ?? 0
+})
 </script>
 
 <template>
@@ -72,10 +75,10 @@ const changeViewMode = (view: Params['view']) => {
         </button>
         <button
           tab-button
-          :class="{ 'tab-button-active': viewMode === 'console', 'op20': viewMode !== 'console' && currentLogs?.length === 0 }"
+          :class="{ 'tab-button-active': viewMode === 'console', 'op20': viewMode !== 'console' && count === 0 }"
           @click="changeViewMode('console')"
         >
-          Console ({{ currentLogs?.length || 0 }})
+          Console ({{ count }})
         </button>
       </div>
     </div>
--- a/packages/ui/client/components/FileDetails.vue
+++ b/packages/ui/client/components/FileDetails.vue
@@ -30,7 +30,7 @@ const changeViewMode = (view: Params['view']) => {
   viewMode.value = view
 }
 const consoleCount = computed(() => {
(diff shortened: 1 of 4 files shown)


packages/vitest/src/runtime/setup.ts
```diff
@@ -38,16 +38,17 @@ function setupDefines(defines: Record<string, any>) {
...
 export function setupConsoleLogSpy() {
-  const stdoutBuffer: any[] = []
-  const stderrBuffer: any[] = []
-  let stdoutTime = 0
-  let stderrTime = 0
-  let timer: any
+  const stdoutBuffer = new Map<string, any[]>()
+  const stderrBuffer = new Map<string, any[]>()
+  const timers = new Map<string, { stdoutTime: number; stderrTime: number; timer: any }>()
+  const unknownTestId = '__vitest__unknown_test__'
```
userquin, line 44: should move it to a common place to detect it on the ui

userquin: uhmm, if sent to the client, logs shouldn't be displayed, and so there is no need to move it

packages/ui/client/components/FileDetails.vue
```diff
@@ -29,6 +29,9 @@ const open = () => {
 const changeViewMode = (view: Params['view']) => {
   viewMode.value = view
 }
+const count = computed(() => {
```
antfu, line 32:
```suggestion
const consoleCount = computed(() => {
```

packages/ui/client/components/FileDetails.vue
```diff
@@ -29,6 +29,9 @@ const open = () => {
 const changeViewMode = (view: Params['view']) => {
   viewMode.value = view
 }
+const consoleCount = computed(() => {
+  return currentLogs.value?.map(l => l.size).reduce((s, a) => s + a, 0) ?? 0
```
edimitchel, line 33: You use map and reduce, why don't do both in reduce?
```suggestion
  return currentLogs.value?.reduce((s, a) => s + a.size, 0) ?? 0
```

edimitchel (changes requested):

packages/ui/client/components/FileDetails.vue
userquin: fixed

edimitchel (approved):

sheremet-va (approved):

PR #1098, opened 2022-04-04
fix(mock-api): provide an explainatory error message when the mocks API hoisting doesn't work

### LINKED ISSUE #1084: ReferenceError: Cannot access '__vite_ssr_import_0__' before initialization when using mocks
### Describe the bug

I get the `ReferenceError: Cannot access '__vite_ssr_import_0__' before initialization` when i try to use the API `vi.mock`.

### Reproduction

This is a failing workflow: https://github.com/gdorsi/vite-mock-error-repro/runs/5804962645?check_suite_focus=true

This is the test file: https://github.com/gdorsi/vite-mock-error-repro/blob/redaxios/frontend/src/components/LoginForm/LoginForm.test.jsx

This is the repro: https://github.com/gdorsi/vite-mock-error-repro

### System Info

```shell
System:
    OS: macOS 11.6
    CPU: (8) arm64 Apple M1
    Memory: 186.25 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.13.1 - ~/.volta/tools/image/node/16.13.1/bin/node
    Yarn: 3.1.1 - ~/.volta/tools/image/yarn/1.22.17/bin/yarn
    npm: 8.1.2 - ~/.volta/tools/image/node/16.13.1/bin/npm
  Browsers:
    Chrome: 100.0.4896.60
    Firefox: 98.0.2
    Safari: 14.1.2
  npmPackages:
    @vitejs/plugin-react: ^1.3.0 => 1.3.0 
    vite: ^2.9.1 => 2.9.1 
    vitest: ^0.8.2 => 0.8.2
```

### Used Package Manager

yarn

### Validations

gdorsi:
closes #1084

### DIFF
--- a/packages/vitest/src/node/plugins/mock.ts
+++ b/packages/vitest/src/node/plugins/mock.ts
@@ -33,6 +33,13 @@ export function hoistMocks(code: string) {
   return m
 }
 
+const API_NOT_FOUND_ERROR = `There are some problems in resolving the mocks API.
+You may encounter this issue when importing the mocks API from another module other than 'vitest'.
+
+To fix this issue you can either:
+- import the mocks API directly from 'vitest'
+- enable the 'globals' options`
+
 export const MocksPlugin = (): Plugin => {
   return {
     name: 'vitest:mock-plugin',
@@ -43,12 +50,22 @@ export const MocksPlugin = (): Plugin => {
       if (m) {
         // hoist vitest imports in case it was used inside vi.mock factory #425
         const vitestImports = code.matchAll(vitestRegexp)
+        let found = false
+
         for (const match of vitestImports) {
           const indexStart = match.index!
           const indexEnd = match[0].length + indexStart
           m.remove(indexStart, indexEnd)
           m.prepend(`${match[0]}\n`)
+          found = true
         }
+
+        // if no vitest import found, check if the mock API is reachable after the hoisting
+        if (!found) {
+          m.prepend('try { vi } catch (_) { try { vitest } catch (__)'
+           + `{ throw new Error(${JSON.stringify(API_NOT_FOUND_ERROR)}) } }\n`)
+        }
+
         return {
           code: m.toString(),
           map: m.generateMap({ hires: true }),
--- a/test/fails/fixtures/mock-import-proxy-module.test.ts
+++ b/test/fails/fixtures/mock-import-proxy-module.test.ts
@@ -1,0 +1,8 @@
+import { expect, test, vi } from './proxy-module'
+
+// This can be used only when imported directly from vitest
+vi.mock('vite')
+
+test('hi', () => {
+  expect(1 + 1).toEqual(2)
+})
--- a/test/fails/fixtures/proxy-module.ts
+++ b/test/fails/fixtures/proxy-module.ts
@@ -1,0 +1,1 @@
+export * from 'vitest'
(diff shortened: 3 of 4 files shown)


sheremet-va:
I think it's better to move this to factory evaluation, and check for "ReferenceError". The code is in `mocker.ts`

sheremet-va:
I think it's better to move this to factory evaluation, and check for "ReferenceError". The code is in `mocker.ts`

gdorsi:
Thanks for the review!

Uhm, I'm not sure I fully understand what you mean, this error happens because after the hoising `vi.mock` is called before the `vi` variable declaration.

We never reach the factory evaluation (which happens somewhere around [here](https://github.com/vitest-dev/vitest/blob/b24fb2535991df1f1053973e48c5a7e46b543135/packages/vitest/src/runtime/mocker.ts#L221), right?), because the error is raised at the very start of the module execution.

sheremet-va:
> Thanks for the review!
> 
> Uhm, I'm not sure I fully understand what you mean, this error happens because after the hoising `vi.mock` is called before the `vi` variable declaration.
> 
> We never reach the factory evaluation (which happens somewhere around [here](https://github.com/vitest-dev/vitest/blob/b24fb2535991df1f1053973e48c5a7e46b543135/packages/vitest/src/runtime/mocker.ts#L221), right?), because the error is raised at the very start of the module execution.

Yes, that is why I left a review. I don't like raising the error here, I think better to format an error that actually happens, since parsing is unsafe.

gdorsi:
In this case `callFunctionMock` is never reached.

I've tried to throw an error at the start of the function:

```js
  private async callFunctionMock(dep: string, mock: () => any) {
    throw new Error('callFunctionMock')
```

I still get:
```js
ReferenceError: Cannot access '__vite_ssr_import_0__' before initialization
```

>  I think better to format an error that actually happen

WDYT if we wrap the hoisted statements in a try/catch?

I'm thinking about something like this:

```diff
-     m.prepend(`${m.slice(startIndex, endIndex)}\n`)
+    m.prepend(`try { ${m.slice(startIndex, endIndex)} } catch(err) { isReferenceError(err) ? throw MOCK_ERROR : throw err } \n`)
```

Anyway, if this becomes too complex we could simply close the PR.

sheremet-va:
If it will never be reached, then it shouldn't throw, that's the point.

gdorsi:
I'm still missing the point, sorry 😞 

I think that it throws because the `vi.mock(module)` is executed before the initialization of the `vi` variable.

I wanted to make the error look better than `Cannot access '__vite_ssr_import_0__' before initialization` because it doesn't even mention real the source of the error.

I'm very new to this codebase, so I'm sure I'm missing something big.

sheremet-va:
> I'm still missing the point, sorry 😞
> 
> I think that it throws because the `vi.mock(module)` is executed before the initialization of the `vi` variable.
> 
> I wanted to make the error look better than `Cannot access '__vite_ssr_import_0__' before initialization` because it doesn't even mention real the source of the error.
> 
> I'm very new to this codebase, so I'm sure I'm missing something big.

Oh, wait, you are write. I was thinking about usage inside `factory`, but it will throw on initializing `vi.mock`. Sorry, I got confused!

sheremet-va (approved):

PR #1104, opened 2022-04-05
fix: really disable deps optimization

### LINKED ISSUE #485: `--ui` attempts to load `.vue` files unrelated to test
### Describe the bug

Attempting to open the use the UI with `--ui` causes Vitest to load `.vue` files unrelated to the tests (from the docs in particular).

The following errors are printed:

```
error: No loader is configured for ".vue" files: demo/components/App.vue
```

```
[
    {
      detail: undefined,
      location: {
        column: 16,
        file: 'demo/main.ts',
        length: 25,
        line: 1,
        lineText: "import App from 'demo/components/App.vue'",
        namespace: '',
        suggestion: ''
      },
      notes: [],
      pluginName: '',
      text: 'No loader is configured for ".vue" files: demo/components/App.vue'
    }
],
```

`demo/main.ts` is not referenced in the source code or any test.

### Reproduction

Repository with reproduction: https://github.com/DerYeger/d3-graph-controller

Can be reproduced by running the test with `yarn test:watch --ui`.

### System Info

```shell
System:
    OS: Windows 10 10.0.22000
    CPU: (24) x64 AMD Ryzen 9 3900X 12-Core Processor            
    Memory: 16.04 GB / 31.92 GB
  Binaries:
    Node: 16.13.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.15 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 7.20.3 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22000.120.0), Chromium (96.0.1054.62)
  npmPackages:
    @vitejs/plugin-vue: 2.0.1 => 2.0.1 
    vite: 2.7.10 => 2.7.10 
    vitest: 0.0.139 => 0.0.139
```

### Used Package Manager

yarn

### Validations

sheremet-va:
Closes #485

### DIFF
--- a/packages/vitest/src/node/plugins/index.ts
+++ b/packages/vitest/src/node/plugins/index.ts
@@ -79,6 +79,9 @@ export async function VitestPlugin(options: UserConfig = {}, ctx = new Vitest())
           },
           // disable deps optimization
           cacheDir: undefined,
+          optimizeDeps: {
+            entries: [],
+          },
         }
       },
       async configResolved(viteConfig) {


packages/vitest/src/node/plugins/index.ts
```diff
@@ -79,6 +79,9 @@ export async function VitestPlugin(options: UserConfig = {}, ctx = new Vitest())
           },
           // disable deps optimization
           cacheDir: undefined,
+          optimizeDeps: {
+            entries: [],
+          },
```
patak-dev, line 84: I wonder if we should add an `optimizeDeps: false` option in Vite (this works totally fine, just thinking that we could be more explicit)

antfu: I'd like to see that (was a bit surprised this wasn't an option in Vite)

PR #1119, opened 2022-04-07
fix: make automocking aware of symbol indexed methods

### LINKED ISSUE #1082: automocking fails in some cases when implementation involves symbols
### Describe the bug

on vitest `v0.6.3` automocking with the [pino logging library ](https://github.com/pinojs/pino)worked as intended.

since vitest `v0.7.0` and up, automocking the `pino` logging library now introduces a behavior change such that `restoreAllMocks` does not properly restore some functions to its original implementation and any future test will error out if they depend on the implementation.

please note, this behavior doesn't happen if spies are directly used. i can only replicate this behavior with automocking.

### Reproduction

in `log.ts`

```typescript
import logger from 'pino'
export default logger() 
```

in `log.test.ts`

```typescript
import { vi, test, expect } from 'vitest'
import log from './log'

vi.mock('./log') 

test('always true', () => {
   log.warn('help')
   vi.restoreAllMocks()
   log.warn('me')

   expect(true).toBeTruthy()
 })
```

the following test will pass on vitest `v0.6.3` and fail on `v0.7.0` and up.

this test will fail with the following error

```
TypeError: this[writeSym] is not a function
```

`writeSym` is referring to the use of a [symbol to index into an object that maps log levels to a log function](https://github.com/pinojs/pino/blob/572ac0a2c0b270f39bfa6b95b01212ac33653b84/lib/tools.js#L58).

### System Info

```shell
Binaries:
    Node: 16.14.0 
    Yarn: 1.22.18 
    npm: 8.5.2
  npmPackages:
    vitest: 0.7.0 => 0.7.0
```

### Used Package Manager

npm

### Validations

eleith:
this PR intends to resolve #1082  by updating automocking to fully support symbol indexed methods (specifically, fixing the root issue which occurs when trying to restore and later use such methods)

## caveat

in a fresh project, i can get the below test snippet to fail for `vitest@0.9.0`. i can then get them to pass when linking vitest locally to this branch. however, when running the tests in `examples/mocks`, the below tests fail as the mock is not properly restored.

## local testing

in a fresh project, the following will fail for `v0.9.0` and pass on this PR

in `log.ts` 

```typescript
const sym = Symbol('x')

const logger = {
  warn() {
    return this[sym]()
  },
  [sym]() {
    return 'hello'
  }
}

export default logger
```

and in `log.test.ts` 

```typescript
import { vi, test, expect } from 'vitest'
import logger from './log'

vi.mock('./log')

test('auto mocked', () => {
  expect(logger.warn()).toBeUndefined()
})

test('restoring auto mock', () => {
  vi.restoreAllMocks()
  expect(logger.warn()).toBe('hello')
})
```

### DIFF
--- a/examples/mocks/test/automocking.spec.ts
+++ b/examples/mocks/test/automocking.spec.ts
@@ -48,14 +48,6 @@ test('automock properly restores mock', async() => {
 
   vi.restoreAllMocks()
 
-  // `spies` (on linux) is an empty Set()
-  // thus we need to restore manually
-  // however, on an external project or on windows,
-  // `spies` is a complete Set() of all (auto) mocks
-  // and the below two lines are not needed
-  vi.mocked(moduleWithSymbol[methodSymbol]).mockRestore()
-  vi.mocked(moduleWithSymbol.warn).mockRestore()
-
   expect(() => {
     log.warn()
   }).not.toThrow()
--- a/examples/mocks/test/automocking.spec.ts
+++ b/examples/mocks/test/automocking.spec.ts
@@ -1,8 +1,9 @@
 import type * as exampleModule from '../src/example'
-
 import log from '../src/log'
+import { methodSymbol, moduleWithSymbol } from '../src/moduleWithSymbol'
 
 vi.mock('../src/log')
+vi.mock('../src/moduleWithSymbol')
 
 test('all mocked are valid', async() => {
   const example = await vi.importMock<typeof exampleModule>('../src/example')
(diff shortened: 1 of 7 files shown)


eleith:
@antfu could you push a new updated build to `tinyspy`? 

i can confirm when installing locally that `v0.3.1` does not include your latest changes for symbol support.

sheremet-va:
> @antfu could you push a new updated build to `tinyspy`?
> 
> i can confirm when installing locally that `v0.3.1` does not include your latest changes for symbol support.

released in `0.3.2`

eleith:
on both windows and in a clean project directory (even linux) all tests in this PR pass with only a single call to `vi.restoreAllMocks()`

however when running tests in `examples/mocks` on linux (14.x and 16.x) the tests in this PR will fail unless the [auto mocks are manually restored ](https://github.com/vitest-dev/vitest/pull/1119/files#diff-8c3587b322423af978c24db382da7159c7f629c8950c5628a77dd6763dd69d7eR51) by calling `.mockRestore()`

when inspecting the value of `spies` in the [internals of vi.restoreAllMocks](https://github.com/vitest-dev/vitest/pull/1119/files#diff-8c3587b322423af978c24db382da7159c7f629c8950c5628a77dd6763dd69d7eR51), in a failed test run, it is incorrectly an empty `Set()`. 

in a successful test run, `spies` is a full `Set()` of all automocks.

examples/mocks/test/automocking.spec.ts
```diff
@@ -40,13 +41,25 @@ test('all mocked are valid', async() => {
...
-  // we are checking here that after unmocking
-  // these symbols are accessible
+test('automock properly restores mock', async() => {
+  expect(log.warn()).toBeUndefined()
+  expect(moduleWithSymbol.warn()).toBeUndefined()
+  expect(moduleWithSymbol[methodSymbol]()).toBeUndefined()
+
+  vi.restoreAllMocks()
+
+  // `spies` (on linux) is an empty Set()
```
sheremet-va, line 51: Hm, it seems `ensureSpy` is not working properly on Linux. Can you go deep into that?

This function should import the same module that Vitest uses, so `spies` will refer to the same Set. Maybe the path is incorrect on Linux, or it has double slashes or something?

I would start with logging the whole `moduleCache` to see if it has duplicate spy module.

sheremet-va (changes requested):

examples/mocks/test/automocking.spec.ts
sheremet-va: If not, I will try to find a time and look into it myself.

sheremet-va: So, I found some time, and this changed fixed it:

https://github.com/vitest-dev/vitest/blob/8b3d4128f01d6f046637679ab9b073add052a5fe/packages/vitest/src/runtime/mocker.ts#L246

```ts
    VitestMocker.spyModule = await this.request(
      `/@fs/${slash(resolve(distDir, 'spy.js'))}`,
    ) as typeof import('../integrations/spy')
    ```

eleith: thanks, that works

sheremet-va (approved):

PR #1123, opened 2022-04-08
chore: cleaning up test for ViewReport.cy.tsx

JessicaSachs:
This PR cleans up the tests for **`ViewReport.cy.tsx`** and begins to add faker in for more realistic test data.

There is one loss of coverage: we no longer check that a **b** tag is the element used in the DOM. For this, we should be using Percy for making sure that the styles are correct. Percy has a free, open-source plan.


userquin:
@JessicaSachs LGTM (aha faker in action, cool), but we need also the same for ansi colored tests, the original PR was to include the support via `ansi-to-html` dep, maybe we can add it later using changes in this PR.

antfu:
@userquin want to send another PR for your changes?

PR #1131, opened 2022-04-09
feat(ui): add the draft state * on code editor

### LINKED ISSUE #1130: ui: add the draft state * on code editor

userquin:
The logic on the `ViewEditor` done, we only need to add the mark/decoration on the tab.

closes #1130


userquin:
I need to change draft to use v-model

PR #1152, opened 2022-04-15
Update storybook in example

IanVS:
This is the result of running 

```shell
npx sb@next upgrade
```

in the `react-storybook-testing` example.  I checked that `pnpm run storybook` executes correctly.

If you get any warnings or errors, you may need to clear the vite cache, with `rm -rf node_modules/.vite-storybook`


patak-dev (approved):

patak-dev:
Thanks!

PR #1173, opened 2022-04-20
chore(ui): add missing `license` field

NMinhNguyen:


NMinhNguyen:
Haha that was quick! Not sure if I should’ve used `chore` or `fix` but thanks @antfu!

Issue #1165, opened 2022-04-18
`throw` from an `async` method,  exits `vitest watch`


dennismphil:
### Describe the bug

The same bug also exits `vitest --ui`

### Code

```ts
// myMethod.ts
export function myMethod(someRequiredString: string) {
  if (someRequiredString.trim().length ===  0) {
    throw new Error('String cannot be empty')
  }
}
```

### Test

```ts
// myMethod.spec.ts
describe('myMethod', () => {
  it('should throw for empty strings', () => {
    expect(() => myMethod('   ')).toThrow()

    // instead this works as expected
    // expect(getToken('  ')).rejects.toThrow();
  });
});
```

### Output

Exits from the watch mode

```
$ pnpm test:watch

> cloud-ui@0.0.1 test:watch /Users/dennis/code
> vitest watch --reporter=verbose

...
...
       press h to show help, press q to quit

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Unhandled Rejection ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
Error: String cannot be empty
 ❯ Module.myMethod src/services/accessApiService.ts:3:10
 ELIFECYCLE  Command failed with exit code 1. 

$ 
```

### Async function

![2022-04-18_12-28-27](https://user-images.githubusercontent.com/4141366/163871149-c1913a54-4c7c-4fc7-b168-6d09f0744a59.png)

### Sync function

[...]

Issue #1132, opened 2022-04-09
Clarify `Vitest was initialized with Native Node` error


sheremet-va:
### Clear and concise description of the problem

The error description is not clear to every user. 

### Suggested solution

Clarify that the error might be because:

- User imports from `vitest` directly not in a test file - suggest importing from `vitest/node` or use `import.meta.vitest`
- User imports from library that imports from `vitest` - suggest using `deps.inline` for that library
- User imports `vitest` inside `globalSetup`, which has different context - suggest using `setupFiles`
- Something bad happened, better create an issue in VItest repo

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

Issue #1201, opened 2022-04-27
Unhandled error when suite name is a function reference


underfisk:
### Describe the bug

Recently during migration from Jest to Vitest I've encountered that my tests were existing with process exit code 1 because one of my describe blocks was using the function reference as the name (Jest under the hood used `.name` therefore it was working as if we pass `MyFunction.name`) but with Vitest it just silently throws an error

### Reproduction

```typescript
function MyReactHook(){
  const myFunction = () => {}
  return {myFunction}
}

describe(MyReactHook, () => {
  it('should render', () => {
    // TODO: This is just a mimic, we could use renderHook
  })
})

```
The code above will lead to the problem shown on the screenshot below

![Screenshot 2022-04-27 at 11 48 16 PM](https://user-images.githubusercontent.com/15980884/165643772-de42be42-9a13-450f-84cb-8c90bdd4db50.png)

### System Info

```shell
System:
    OS: macOS 12.3.1
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 8.00 GB / 64.00 GB
    Shell: 3.2.2 - /usr/local/bin/fish
  Binaries:
    Node: 16.13.0 - ~/.nvm/versions/node/v16.13.0/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 8.7.0 - ~/.nvm/versions/node/v16.13.0/bin/npm
  Browsers:
    Chrome: 100.0.4896.127
    Safari: 15.4
```

### Used Package Manager

yarn

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

sheremet-va:
[...]

Issue #884, opened 2022-03-02
Huge debugging improvement (show enumerable error properties)


aalexgabi:
### Clear and concise description of the problem

Error enumerable properties are not shown for test case:
```
import { test } from 'vitest';

test('throw error with context', () => {
  const err = new Error('This is an error with context');
  err.userId = 32;
  err.companyId = 64;
  throw err;
})
```

![image](https://user-images.githubusercontent.com/600821/156337522-c37b328b-6b28-407a-a634-ef3bd4570edf.png)

### Suggested solution

Show enumerable properties on errors as node does by default:
![image](https://user-images.githubusercontent.com/600821/156337855-e1be743a-c3ce-4e95-a942-ba011a231372.png)

### Alternative

At the very least please add an option to opt in for this behavior.

### Additional context

When calling a function transformSomething(array) the function might have many iterations inside but one iteration will fail becaue the item may be invalid. It's very useful when throwing an error, especially for logging to specify which iteration failed. 

For more context here is the equivalent issue for chromium dev tools console https://bugs.chromium.org/p/chromium/issues/detail?id=1251322&q=enumerable&can=2

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

sheremet-va:
Resolved in https://github.com/vitest-dev/vitest/pull/1202

PR #1202, opened 2022-04-28
feat: error log improvements

sheremet-va:
Closes: #1132, #884, #1201
Fixes: #1165, #1203

- Errors show full stacktrace, since it's really hard to debug errors without an actual stacktrace (sometimes it has none, or look at #1201)
- Unhandled errors don't exit process, but fail the whole suite
- Unhandled errors are shown after the summary with a warning
- Clarified error message for "Vitest was initialized with Native Node"
- If errors have custom properties, they are shown under "Serialized Error" - previously I wanted to show them like Node does, but it looks bad, if error has only one stack trace with a code frame, so I decided to split it
- If error has "cause", print the case as the error without a code frame

### DIFF
--- a/packages/vitest/src/node/core.ts
+++ b/packages/vitest/src/node/core.ts
@@ -493,8 +493,8 @@ export class Vitest {
     return code.includes('import.meta.vitest')
   }
 
-  printError(err: unknown) {
-    return printError(err, this)
+  printError(err: unknown, fullStack = false) {
+    return printError(err, fullStack, this)
   }
 
   onServerRestarted(fn: () => void) {
--- a/packages/vitest/src/node/core.ts
+++ b/packages/vitest/src/node/core.ts
@@ -499,7 +499,7 @@ export class Vitest {
   }
 
   printError(err: unknown, fullStack = false, type?: string) {
-    return printError(err, type, fullStack, this)
+    return printError(err, type, fullStack, true, this)
   }
 
   onServerRestarted(fn: () => void) {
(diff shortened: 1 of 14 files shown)


packages/vitest/src/node/error.ts
```diff
@@ -36,10 +36,12 @@ export async function printError(error: unknown, ctx: Vitest) {
...
   )
 
+  const errorProperties = getErrorProperties(e)
+
   printErrorMessage(e, ctx.console)
-  await printStack(ctx, stacks, nearest, async (s, pos) => {
+  printStack(ctx, stacks, nearest, errorProperties, (s, pos) => {
     if (s === nearest && nearest) {
-      const sourceCode = await fs.readFile(fileFromParsedStack(nearest), 'utf-8')
+      const sourceCode = readFileSync(fileFromParsedStack(nearest), 'utf-8')
```
sheremet-va, line 44: I found a bug where `fs.readFile` stalls the terminal, and breaks `for of` loop - because of that some stack trace might be missing.

packages/vitest/src/node/core.ts
```diff
@@ -493,8 +493,8 @@ export class Vitest {
     return code.includes('import.meta.vitest')
   }
 
-  printError(err: unknown) {
-    return printError(err, this)
+  printError(err: unknown, fullStack = false) {
+    return printError(err, fullStack, this)
```
sheremet-va, line 497: I think we should show all stack trace for some errors (like unhandled ones), bacause it's really hard to debug otherwise.

patak-dev (approved):

patak-dev (approved):

aleclarson:
> * If errors have custom properties, they are shown under "Serialized Error" - previously I wanted to show them like Node does, but it looks bad, if error has only one stack trace with a code frame, so I decided to split it

Is this expected to still work? I'm not seeing `postgres`-specific error properties. When I try-catch the error and `console.error()` it, I see them as expected.

PR #1215, opened 2022-04-30
feat(vite-node): options via CLI (fixes #1208)

### LINKED ISSUE #1208: Recent versions of `vite-node` fail to execute a script when ESM is imported
### Describe the bug

**TLDR; `vite-node` higher than `0.8.5` fails to run a script where ESM is imported.**

I'm using `vite-node` to execute some code that relies on several dependencies. One of those dependencies (in my case `msw`) provides both ESM and UMD bundles. When executing the script, `vite-node` imports the ESM bundle, and immediately crashes.

The script works fine when running `vite-node@0.8.5`, but fails with any higher version. I believe that the turning point is the [c4a25ae](https://github.com/vitest-dev/vitest/commit/c4a25ae) commit, where some checks to inline dependencies where removed.

### Reproduction

https://github.com/antoinerey/vite-node-reproduction-esm-import

### System Info

```shell
System:
    OS: macOS 12.0.1
    CPU: (8) x64 Intel(R) Core(TM) i7-8557U CPU @ 1.70GHz
    Memory: 104.10 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 18.0.0 - /usr/local/bin/node
    npm: 8.6.0 - /usr/local/bin/npm
    Watchman: 2022.03.21.00 - /usr/local/bin/watchman
  Browsers:
    Brave Browser: 100.1.37.116
    Chrome: 101.0.4951.41
    Firefox: 98.0.1
    Safari: 15.1
  npmPackages:
    vite-node: ^0.10.0 => 0.10.0
```

### Used Package Manager

yarn

### Validations

antoinerey:
Fixes #1208

This PR adds a new flag to the `vite-node` CLI: `--server-options`. Using [dot syntax](https://github.com/cacjs/cac#dot-nested-options), it's now possible to define any option passed to the underlying ViteNodeServer instance.

Because CLIs do not support RegExp as input, we make sure to transform any string that looks like `/xxx/` to RegExps.


packages/vite-node/src/cli.ts
```diff
@@ -48,11 +49,12 @@ Usage:
...
-  -c, --config <file>    ${dim('[string]')} use specified config file
-  -w, --watch           ${dim('[boolean]')} restart on file changes, similar to "nodemon"
-  -s, --silent          ${dim('[boolean]')} do not emit errors and logs
-  --vue                 ${dim('[boolean]')} support for importing Vue component
+  -r, --root <path>           ${dim('[string]')} use specified root directory
+  -c, --config <file>         ${dim('[string]')} use specified config file
+  -w, --watch                ${dim('[boolean]')} restart on file changes, similar to "nodemon"
+  -s, --silent               ${dim('[boolean]')} do not emit errors and logs
+  --vue                      ${dim('[boolean]')} support for importing Vue component
+  -i, --inline-deps <deps>    ${dim('[string]')} inline specified dependencies
```
antoinerey, line 57: I'm not convinced that this is the best way to show that this flag may be set several times to inline several dependencies. Does anybody have a better idea?

packages/vite-node/src/cli.ts
```diff
@@ -62,18 +64,24 @@ export interface CliOptions {
...
     configFile: options.config,
     root: options.root,
   })
   await server.pluginContainer.buildStart({})
 
-  const node = new ViteNodeServer(server)
+  const node = new ViteNodeServer(server, {
+    deps: {
+      inline: Array.isArray(inlineDeps) ? inlineDeps : [inlineDeps],
+    },
```
antoinerey, line 83: Should we also port the rest of the relevant configuration as CLI flags? Or would you rather keep the bare minimum, and add stuff when needed?

packages/vite-node/src/cli.ts
```diff
@@ -48,11 +49,12 @@ Usage:
   $ vite-node [options] [files]
 
 Options:
-  -r, --root <path>      ${dim('[string]')} use specified root directory
-  -c, --config <file>    ${dim('[string]')} use specified config file
-  -w, --watch           ${dim('[boolean]')} restart on file changes, similar to "nodemon"
-  -s, --silent          ${dim('[boolean]')} do not emit errors and logs
-  --vue                 ${dim('[boolean]')} support for importing Vue component
```
antoinerey, line 55: The `--inline-deps` flag is quite long, so I had to align the rest of the help. But nothing changed other than that.

packages/vite-node/src/cli.ts
sheremet-va: I think we use `[...files]` for vitest.

packages/vite-node/src/cli.ts
sheremet-va: I think we should at least port `fallbackCJS` (with `no-` prefix for `false`) - I think `minimist` provides this functionality out of the box, but I am not sure.

Also, what about regexp?

sheremet-va (changes requested):

sheremet-va:
> I could not find any test associated to vite-node.

I don't think we explicitly tested vite-node 🤔 
Maybe we can add a `vite-node` folder to `test`.

> I've only taken care of adding --inline-deps, but should we also port all/most of the relevant configuration that people may define under the test property in their Vitest config file?

Not all are relevant there, but maybe we can provide the ones that are not used in `cli` file right now.

antfu:
Maybe we could introduce a field like `viteNode` inside `vite.config.ts` for `vite-node`?

For CLI options, I am imagining something more automated, like `--deps.inline=...` and `--x.y=` mapping the structure of the configurations so we don't need to introduce new options every time ppl need them.

antoinerey:
> For CLI options, I am imagining something more automated, like --deps.inline=... and --x.y= mapping the structure of the configurations so we don't need to introduce new options every time ppl need them.

Oh yes, good point. I'll try to look into it.

> Maybe we could introduce a field like viteNode inside vite.config.ts for vite-node?

I'm still trying to get a feeling of how it could work from a consumer perspective, especially how we could avoid duplicating the configuration between `vitest` (the `test` part), and `vite-node` (the `viteNode` part). Is this duplication really an issue, or is it a tradeoff that you're willing to accept?

---

Side notes:

- Let's not forget to add a similar type system as `vitest` through a triple slash command.
- Let's investigate wow we could automatically translate TS types into CLI documentation.

sheremet-va:
@antoinerey I think `cac` (cli tool used in Vitest) already supports object-like parameters. We can just pass them directly to options, I think (maybe with some processing).

ViteNode right now uses `minimist`, so it will require a small rewrite.

antoinerey:
I'll definitely take a look at that! Thanks for the tip.

packages/vite-node/src/cli.ts
```diff
@@ -24,6 +28,7 @@ export interface CliOptions {
   root?: string
   config?: string
   watch?: boolean
+  serverOptions?: ViteNodeServerOptions
```
antoinerey, line 31: Since the option is provided by the consumer, it might not match `ViteNodeServerOptions`, so I'm cheating a little bit here.

However, we could rely on a TS guard to assert that the given options match the TS interface, but it would mean hardcoding the checks, and thus duplicating the source of truth between `ViteNodeServerOptions` and the TS guard. Concretely, both could be out of sync, leading to issues. 

Do you see another way to achieve this without duplicating the source of truth? 🤔

packages/vite-node/src/cli.ts
```diff
@@ -12,6 +13,9 @@ cli
   .option('-r, --root <path>', 'Use specified root directory')
   .option('-c, --config <path>', 'Use specified config file')
   .option('-w, --watch', 'Restart on file changes, similar to "nodemon"')
+  // TODO: How could we document this? Ideally, we should link to the
+  // TODO: ViteNodeServerOptions type since it's the source of truth.
+  .option('--server-options <options>', 'Use specified Vite server options')
```
antoinerey, line 18: I could not find any proper example of such documentation. Should I add a few lines to the README?

packages/vite-node/src/cli.ts
```diff
@@ -81,3 +89,23 @@ async function run(files: string[], options: CliOptions = {}) {
...
+
+  if (serverOptions.deps && serverOptions.deps.external) {
+    serverOptions.deps.external = serverOptions.deps.external.map((dep) => {
+      return typeof dep === 'string' && dep.startsWith('/') && dep.endsWith('/')
+        ? new RegExp(dep)
+        : dep
+    })
+  }
+
+  // TODO: Handle serverOptions.transformMode.
```
antoinerey, line 110: I've noticed that `serverOptions.transformMode.ssr` (and `.web`) only accept Regexes, while `serverOptions.deps.inline` (and `.external`) accepts both strings and Regexes.

- Is that expected?
- If so, would you mind giving a bit of context as why? 🙏

antoinerey:
Converted the PR to draft since it's still WIP, and should probably not be merged.

---

@sheremet-va @antfu I've pushed a new version of this PR using `cac` as CLI, and it now handles arbitrary options through dot nesting. 🎉

```sh
vite-node --server-options.deps.inline="module-1" --server-options.deps.external="module-2" script.ts
```

Before going further (adding tests, polishing the code, etc.) I would like to get feedbacks, especially about ideas discussed in comments. Do you feel like this is going into the right direction? 🙏

packages/vite-node/src/cli.ts
antfu: Yeah I think that would do

sheremet-va:
> Do you feel like a TS guard could be a nice solution even though it duplicates the shape of the options? Or should we keep a single source of truth instead?

I think the whole idea why we are doing this is to just do it and forget kind of situations.

packages/vite-node/src/cli.ts
sheremet-va: doesn't have a reason 👀 
I guess just didn't notice. but to be honest I prefer using regexp always, because current solution for `deps.inline` as a string is not reliable on non-npm/pnpm package managers

antfu:
Renamed to `--options` for simplicity. Thanks

antfu (approved):

PR #1221, opened 2022-05-01
fix: self exported module call stack error

### LINKED ISSUE #1220: Maximum call stack when export the current file
### Describe the bug

Failing code:
```ts
import { describe, it, expect, vi } from "vitest";
import { dbContact } from "@judis/shared";

describe("contact spec", async () => {
  it("test", () => {
    try {
      const schema = dbContact(); // error here
    } catch (e) {
      console.error("sss", e);
    }
    expect(1).toBe(1);
  });
});
```
Error: 
```bash
stderr | src/composables/validation.spec.ts > contact spec > tt
sss RangeError: Maximum call stack size exceeded
    at Module.get [as dbContact] (file:///C:/dev/untitled-advogado-project/node_modules/.pnpm/vitest@0.10.0_happy-dom@3.1.0+sass@1.51.0/node_modules/vitest/dist/chunk-vite-node-utils.33447cc0.js:9073:13)
    at Module.get [as dbContact] (file:///C:/dev/untitled-advogado-project/node_modules/.pnpm/vitest@0.10.0_happy-dom@3.1.0+sass@1.51.0/node_modules/vitest/dist/chunk-vite-node-utils.33447cc0.js:9073:32)
    at Module.get [as dbContact] (file:///C:/dev/untitled-advogado-project/node_modules/.pnpm/vitest@0.10.0_happy-dom@3.1.0+sass@1.51.0/node_modules/vitest/dist/chunk-vite-node-utils.33447cc0.js:9073:32)
    at Module.get [as dbContact] (file:///C:/dev/untitled-advogado-project/node_modules/.pnpm/vitest@0.10.0_happy-dom@3.1.0+sass@1.51.0/node_modules/vitest/dist/chunk-vite-node-utils.33447cc0.js:9073:32
```

Code throwing the call stack: 
```ts
function exportAll(exports, sourceModule) {
  for (const key in sourceModule) {
    if (key !== "default") {
      try {
        Object.defineProperty(exports, key, {
          enumerable: true,
          configurable: true,
          get() {
            return sourceModule[key]; // here
          }
        });
      } catch (_err) {
      }
    }
  }
}
```
`@judis/shared` is a workspace dependency, it exports a nested dependency:
```ts
// @judis/shared/src/index.ts
export * from './db'

// @judis/shared/src/db.ts
export * from './firms'

// @judis/shared/firms/index.ts
export * from './contact'
export * from './index'

// @judis/shared/firms/contact.ts
export function dbContact(){
  console.log('it doesn't matter what the code is')
}
```

### Reproduction

This happens when you export the current `module` 
```ts
// mydep/index.ts
export * from './foo'
export * from './index' // error here
```

### System Info

```shell
System:
    OS: Windows 10 10.0.22000
    CPU: (24) x64 AMD Ryzen 9 3900X 12-Core Processor
    Memory: 25.24 GB / 63.94 GB
  Binaries:
    Node: 16.10.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.13 - C:\Program Files\nodejs\yarn.CMD
    npm: 7.24.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22000.120.0), Chromium (101.0.1210.32)
  npmPackages:
    vitest: ^0.10.0 => 0.10.0
```

### Used Package Manager

pnpm

### Validations

pikax:
fix #1220 

exporting itself is wrong, but it can be exported by mistake, `vite` handles the export just fine.

note: eslint was crashing :D so I ignored the files for this specific test

### DIFF
--- a/packages/vite-node/src/client.ts
+++ b/packages/vite-node/src/client.ts
@@ -253,6 +253,13 @@ function proxyMethod(name: 'get' | 'set' | 'has' | 'deleteProperty', tryDefault:
 }
 
 function exportAll(exports: any, sourceModule: any) {
+  // #1120 when a module exports itself it causes
+  // call stack error
+  if (exports === sourceModule) {
+    // eslint-disable-next-line no-console
+    console.warn('[vite-node] module is being self exported', sourceModule)
+    return
+  }
   // eslint-disable-next-line no-restricted-syntax
   for (const key in sourceModule) {
     if (key !== 'default') {
--- a/packages/vite-node/src/client.ts
+++ b/packages/vite-node/src/client.ts
@@ -255,11 +255,9 @@ function proxyMethod(name: 'get' | 'set' | 'has' | 'deleteProperty', tryDefault:
 function exportAll(exports: any, sourceModule: any) {
   // #1120 when a module exports itself it causes
   // call stack error
-  if (exports === sourceModule) {
-    // eslint-disable-next-line no-console
-    console.warn('[vite-node] module is being self exported', sourceModule)
+  if (exports === sourceModule)
     return
-  }
+
   // eslint-disable-next-line no-restricted-syntax
   for (const key in sourceModule) {
     if (key !== 'default') {
--- a/.eslintignore
+++ b/.eslintignore
@@ -5,3 +5,4 @@ node_modules
 *.d.ts
 coverage
 !.vitepress
+test/core/src/self
--- a/test/core/src/self/foo.ts
+++ b/test/core/src/self/foo.ts
@@ -1,0 +1,5 @@
+/* eslint-disable */
+
+export function foo(): true {
+  return true;
+}
--- a/test/core/src/self/index.ts
+++ b/test/core/src/self/index.ts
@@ -1,0 +1,4 @@
+/* eslint-disable */
+
+export * from './foo';
+export * from './index'; // <-- wrong
--- a/test/core/test/self.test.ts
+++ b/test/core/test/self.test.ts
@@ -1,0 +1,7 @@
+import { expect, it } from 'vitest'
+import { foo } from '../src/self'
+
+// #1220 self export module
+it('self export', () => {
+  expect(foo()).toBe(true)
+})


packages/vite-node/src/client.ts
```diff
@@ -253,6 +253,13 @@ function proxyMethod(name: 'get' | 'set' | 'has' | 'deleteProperty', tryDefault:
 }
 
 function exportAll(exports: any, sourceModule: any) {
+  // #1120 when a module exports itself it causes
+  // call stack error
+  if (exports === sourceModule) {
+    // eslint-disable-next-line no-console
+    console.warn('[vite-node] module is being self exported', sourceModule)
```
sheremet-va, line 260: do we need a warning, if this is expected behaviour? also, what will happen if module imports itself from the start? wouldn’t “exports” object be empty?

sheremet-va (changes requested):

packages/vite-node/src/client.ts
pikax: I wouldn't call "expected behaviour".
Even `eslint` was throwing exceptions
```bash
× eslint --fix:
Oops! Something went wrong! :(
ESLint: 8.14.0
Occurred while linting C:\dev\vitest\test\core\src\self\index.ts:3
Rule: "import/export"
    at C:\dev\vitest\node_modules\.pnpm\eslint-plugin-import@2.26.0_eslint@8.14.0\node_modules\eslint-plugin-import\lib\rules\export.js:190:15
```

I cannot think a use case where the module exporting itself makes sense 🤔

I don't think the warning is necessary, just a nice to have.

sheremet-va: what about order of importing? Im afraid it might lead to unexpected “exports” content. does it work correctly, if you export itself first? or what will happen if you import a module, that imports/exports the same module? 
is it covered by the same rules in exportAll, or is it a different issue?

i’m not sure about error because some people might use dependencies that do that and they don’t have the power to change it, and will see it everytime they test. maybe we can show it if debug flag is enabled, like we do with “request”?

pikax: > what about order of importing?


Not sure, what logically would differ.

>  afraid it might lead to unexpected “exports” content. does it work correctly, if you export itself first? 

Currently is throwing a maximum call stack error, which is very cryptic at best, not sure what you mean by "unexpected "exports" content".

> or what will happen if you import a module, that imports/exports the same module?
is it covered by the same rules in exportAll, or is it a different issue?

Do you mean [circular](https://github.com/vitest-dev/vitest/blob/main/test/core/test/circular.test.ts)? 



> i’m not sure about error because some people might use dependencies that do that and they don’t have the power to change it, and will see it everytime they test. maybe we can show it if debug flag is enabled, like we do with “request”?

I would say this is very unlikely because: 
- I couldn't find any issue in regards to this problem, that's why I created the issue https://github.com/vitest-dev/vitest/issues/1220
- Usually dependencies are build by rollup/webpack, so there's no actually imports/exports like this
- Eslint doesn't even allow you to commit or passing lint on this (I don't use eslint on the project this happened)
- I don't know a Valid use case for exporting itself, the only case I can think of is by mistake.


Warning the user is just a nice to have, don't have strong opinion about it, I'm not arguing about keeping it, I just explained it's a nice to have IMO.

sheremet-va: > I would say this is very unlikely because:

It might happen with libraries that do not transpile code. For example, React Native. Currently it is no officially supported, yes, just to give an example. Vitest is used to test code that might not even run in a browser, we support Node testing by default.

If this is such a bad behaviour, maybe we should throw a non-cryptic error instead?

sheremet-va: > Do you mean [circular](https://github.com/vitest-dev/vitest/blob/main/test/core/test/circular.test.ts)?

I mean:
```ts
// foo
export const foo = () => true
export * from './index'
```
```ts
// index
export * from './foo'
```

pikax: Sorry I don't understand what you are trying to say, are you saying the current behaviour (not allowing self import) is correct? or are you saying the warning is incorrect? 

Since this is easily fixed on my end without needing this patch, we can close this if you don't agree with this different behaviour.

> might happen with libraries that do not transpile code. For example, React Native. Currently it is no officially supported, yes, just to give an example. Vitest is used to test code that might not even run in a browser, we support Node testing by default.

You haven't provided an actual case, only a hypothetical scenario

sheremet-va: > Sorry I don't understand what you are trying to say, are you saying the current behaviour (not allowing self import) is correct? or are you saying the warning is incorrect?

I'm arguing what is the beast approach here. I think the current behaviour is incorrect, or at the very least not the best. We can either silently patch it, patch it with a warning, or throw an error. You propose the second, I merely ask why not throw an error or silently fix it with a warning on debug?

> You haven't provided an actual case, only a hypothetical scenario

What do you mean? React Native is an actual example.

pikax: > You propose the second, I merely ask why not throw an error or silently fix it with a warning on debug?

I've already mention why I think the current behaviour (main branch) is incorrect, why throwing is incorrect (it works on Vite) and why I think warning is nice (just nice to have but not strong opinion about it).



> What do you mean? React Native is an actual example.

You didn't provide an example, you just said: "It **might happen** with libraries that do not transpile code. For example, React Native. Currently it is no officially supported, yes, just to give an example."
Without providing a legit case, this is just "might happen" territory and I already argue why this is an unlikely issue. 


> I mean:
> 
> ```ts
> // foo
> export const foo = () => true
> export * from './index'
> ```
> 
> ```ts
> // index
> export * from './foo'
> ```

This example is clearly a code smell and most likely caused by a bad refactoring or an oversight on part of the developer, as I said many times, I don't see any legit usage of exporting self module (as per the test), ofc I might be forgetting something, but so far you haven't provided an actual code sample or scenario where it makes sense.

sheremet-va: You haven't explained why it's "nice" to have. You are just constantly repeating it.

If you say it's working in VIte (and I presume in Node - a lot of things that work in VIte don't work in Node, and we follow Node in that regard - module resolution is the example, because code is running in Node after all, and not in the browser), then why do I need a constant reminder in the console that my code works?

I totally agree it's nice to have a reminder, but I am not sure Vitest is the tool to give that reminder, that's all.

pikax: > You haven't explained why it's "nice" to have. You are just constantly repeating it.

To notify people that's probably a mistake, and as I said: "I don't have strong opinions about it."  - don't know why we are so many replies into this....


---

Apart from the warning do you have any other feedback on this?

sheremet-va: > Apart from the warning do you have any other feedback on this?

No, I really like what you did here 👍🏻

sheremet-va (approved):

PR #1224, opened 2022-05-01
feat: enable to remove shebang when executing the code via vite-node

### LINKED ISSUE #1175: vite-node fails to parse hashbang
### Describe the bug

When a file has a hashbang/shebang (most often `#!/usr/bin/env node`), vite-node fails with this error:

```
Error: Parse failure: Unexpected character '!' (1:1)
Contents of line 1: #!/usr/bin/env node
```

### Reproduction

StackBlitz link: [https://stackblitz.com/edit/node-fhzvli](https://stackblitz.com/edit/node-fhzvli).

### System Info

```shell
System:
  OS: Linux 5.10 Ubuntu 20.04.4 LTS (Focal Fossa)
  CPU: (8) x64 Intel(R) Core(TM) i5-1035G1 CPU @ 1.00GHz
  Memory: 1.02 GB / 3.72 GB
  Container: Yes
  Shell: 5.0.17 - /bin/bash
Binaries:
  Yarn: 1.22.15 - ~/.nvm/versions/node/v16.14.2/bin/yarn
npmPackages:
  vite: ^2.9.5 => 2.9.5 
  vitest: ^0.9.3 => 0.9.3
```

### Used Package Manager

pnpm

### Validations

togami2864:
fix #1175


packages/vite-node/src/cli.ts
```diff
@@ -70,6 +71,7 @@ async function run(options: CliOptions = {}) {
     logLevel: 'error',
     configFile: options.config,
     root: options.root,
+    plugins: [RemoveShebangPlugin()],
```
sheremet-va, line 74: We directly confirm what code is run in [vitest/packages/vite-node/src/client.ts:181](https://github.com/vitest-dev/vitest/blob/main/packages/vite-node/src/client.ts#L181), so we don't need any plugins here. Also, this will only work for `vite-node`, but not for `vitest`

sheremet-va (changes requested):

packages/vite-node/src/cli.ts
togami2864: > We directly edit what code is run in [vitest/packages/vite-node/src/client.ts:181](https://github.com/vitest-dev/vitest/blob/main/packages/vite-node/src/client.ts#L181)

The parse error occured when getting the variable `transformed` which would be embeded in `vm.runInThisContext`.
https://github.com/vitest-dev/vitest/blob/22f3c29610b415f78c8578365eb1a5aaeb0c8d59/packages/vite-node/src/client.ts#L132

 Inside `await this.options.fetchModule(id)`,  functions exported from vite are used, which can only be interfered with via plugin from `vite-node`.

> this will only work for vite-node, but not for vitest

Yes, I was thinking of the following use case with `vite-node` and had not considered about `vitest`.

```
// cli.ts
#!/usr/bin/env node
import("../your_module")
```

```shell
// users can run .ts code without thinking ts when developing cli tool.
$ npx vite-node cli.ts
```

sheremet-va: > Yes, I was thinking of the following use case with vite-node and had not considered about vitest.

Don't you want to also test that code 🤔 

> The parse error occured when getting the variable transformed which would be embeded in vm.runInThisContext.

Interesting, I though error happens when we call `vm.runInThisContext` 🤔 Maybe, this is a Vite bug, then? Does regular Vite files work?

togami2864: I found a solution ✨

`vite-node` uses the SSR mode of `vite` to transform. In doing so, the error occurred throw by the `acorn` parser.

https://github.com/vitejs/vite/blob/dddda1e05dd2184960afdcdf2960053acb30f5a9/packages/vite/src/node/ssr/ssrTransform.ts#L37

But, I found the [`allowHashBang`](https://github.com/acornjs/acorn/issues/629) option in `acron`. If it's enabled, 
we can treat shebang in `vite` and `vite-node` .

Also, `vm.runInthisContext` treates shebang as an Invalid token. You are right :)
```
SyntaxError: Invalid or unexpected token
    at new Script (node:vm:100:7)
    at createScript (node:vm:257:10)
    at Object.runInThisContext (node:vm:305:10)
    at ViteNodeRunner.directRequest (file:///Users/toueyoshiaki/Desktop/OSS/vitest/packages/vite-node/dist/client-900fb7db.js:144:19)
```

Therefore, need to remove shebang from the variable `transformed and both `vite-node` and `vitest` can run the code with shebang.

In conclusion, 
1: add allowHashBang in vite
2: change `transformed` => `transformed.replace(/^\#\!.*/, "")

should fix the bug and no plugins are required !!

togami2864:
The PR for vite is merged!! I'll rebase main branch after the next vite release and update this PR.

patak-dev:
Thanks for the PR in vite @togami2864! vite@2.9.8 has been released with it.

packages/vite-node/src/client.ts
```diff
@@ -178,7 +178,7 @@ export class ViteNodeRunner {
     })
 
     // add 'use strict' since ESM enables it by default
-    const fn = vm.runInThisContext(`'use strict';async (${Object.keys(context).join(',')})=>{{${transformed}\n}}`, {
+    const fn = vm.runInThisContext(`'use strict';async (${Object.keys(context).join(',')})=>{{${transformed.replace(/^\#\!.*/, '')}\n}}`, {
```
sheremet-va, line 181: I think we should also patch coverage in `src/integrations/coverage.ts#L100`, otherwise it will be off

sheremet-va (changes requested):

packages/vite-node/src/client.ts
sheremet-va: I'm not even sure how source maps work with this, maybe @Demivan can help us?

antfu, line 181:
```suggestion
    const fn = vm.runInThisContext(`'use strict';async (${Object.keys(context).join(',')})=>{{${transformed.replace(/^\#\!.*/, s => ' '.repeat(s.length))}\n}}`, {
```

To avoid index changes.

sheremet-va (approved):

PR #1225, opened 2022-05-01
feat: add config option logHeapUsage

### LINKED ISSUE #1187: highlighting long running tests

#1087 introduced a less verbose reporter mode, but I found that the test duration is always yellow, Initially I thought my tests were slow (some indeed, yes 🙈) 

For example in the screenshot attached to the PR, some tests runs in <10ms, and some in >500ms both are render with yellow

image
### 

### Suggested solution

I think, it could be use some code color convention for long running tests duration logging 

something like 

green: < 300ms
yellow: > 300ms and <500ms
red: >500ms 

actually, in the listRenderer, if the test duration was more than 300ms, we display with the yellow the duration

https://github.com/vitest-dev/vitest/blob/main/packages/vitest/src/node/reporters/renderers/listRenderer.ts#L53

base, in the base reporter, we don't have that validation 

https://github.com/vitest-dev/vitest/blob/main/packages/vitest/src/node/reporters/base.ts#L77
 

### Alternative

_No response_

### Additional context

_No response_

### Validations

sheremet-va:
Not sure if I placed `heap` calculations in the right place tho :P

Also fixes #1187

### DIFF
--- a/packages/vitest/src/runtime/run.ts
+++ b/packages/vitest/src/runtime/run.ts
@@ -51,9 +51,9 @@ export async function callSuiteHook<T extends keyof SuiteHooks>(
   return callbacks
 }
 
-const packs = new Map<string, TaskResult|undefined>()
+const packs = new Map<string, TaskResult | undefined>()
 let updateTimer: any
-let previousUpdate: Promise<void>|undefined
+let previousUpdate: Promise<void> | undefined
 
 function updateTask(task: Task) {
   packs.set(task.id, task.result)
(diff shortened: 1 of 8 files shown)


packages/vitest/src/runtime/run.ts
```diff
@@ -150,6 +150,7 @@ export async function runTest(test: Test) {
   getSnapshotClient().clearTest()
 
   test.result.duration = now() - start
+  test.result.heap = process.memoryUsage().heapUsed
```
antfu, line 153: Shall we only collect them when the flag is on?

yeion7:
This is awesome, thanks @sheremet-va

PR #1232, opened 2022-05-04
fix(inline-snapshots): detect linebreaks

### LINKED ISSUE #1134: Inline snapshots with line breaks are unstable
### Describe the bug

Test below keeps failing when run. It passes right after the snapshot is updated but fails immediately on the next run. Snapshot update doesn't actually change any content. I think `vitest` doesn't handle line changes properly when comparing inline snapshots.

```ts

/*
 * - npm run test
 * - Error -> Snapshot has changed. Press u
 * - Snapshot is updated, test passes. Press enter to re-run
 * - Error -> Snapshot has changed. Repeat previous steps...
 */

test('unstable snapshot', () => {
  expect(() => {
    throw new Error(['First line', 'Second line'].join('\n'));
  }).toThrowErrorMatchingInlineSnapshot(`
    "First line
    Second line"
  `);
});
```

https://user-images.githubusercontent.com/14806298/162612663-002cd153-ad18-42e3-b5df-cf14a3acd5b3.mp4

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-kxavky?file=test/basic.test.ts

### System Info

```shell
Default Stackblitz environment.
```

### Used Package Manager

npm

### Validations

AriPerkkio:
Fixes #1134

There is one unrelated typo fix commit and one test regression commit. I'm happy to move those into separate PR if needed.

### DIFF
--- a/packages/vitest/src/integrations/snapshot/port/utils.ts
+++ b/packages/vitest/src/integrations/snapshot/port/utils.ts
@@ -161,8 +161,18 @@ export async function saveSnapshotFile(
 
 export function prepareExpected(expected?: string) {
   function findStartIndent() {
-    const match = /^( +)}\s+$/m.exec(expected || '')
-    return match?.[1]?.length || 0
+    // Attemps to find indentation for objects.
+    // Matches the ending tag of the object.
+    const matchObject = /^( +)}\s+$/m.exec(expected || '')
+    const objectIndent = matchObject?.[1]?.length
+
+    if (objectIndent)
+      return objectIndent
+
+    // Attempts to find indentation for texts.
+    // Matches the quote of first line.
+    const matchText = /^\n( +)"/.exec(expected || '')
+    return matchText?.[1]?.length || 0
   }
 
   const startIndent = findStartIndent()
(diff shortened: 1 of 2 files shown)


packages/vitest/src/integrations/snapshot/port/utils.ts
```diff
@@ -161,17 +161,17 @@ export async function saveSnapshotFile(
 
 export function prepareExpected(expected?: string) {
   function findStartIndent() {
-    const match = /^( +)}\s+$/m.exec(expected || '')
+    const match = /^( +)[}|"]/m.exec(expected || '')
```
AriPerkkio, line 164: I'm quite unsure what this was supposed to do earlier:


`^( +)}\s+$`

- `^` From the beginning
- `( +)` Capture at least one space...
- `}` ... before } character
- `\s+$` ... only when whole expression ends with whitespaces

... right?

Should this have been without `^` instead?

AriPerkkio: Just realized there's the `m` modifier for multilines. Applied required changes for this.

PR #1249, opened 2022-05-06
refactor(vite-node): Replace minimist with cac

antoinerey:
As discussed in #1215, let's replace minimist with cac to allow further changes.

Note that `--silent` and `--vue` flags have been removed from the CLI. They were displayed in the help, but were not used in any way.


packages/vite-node/src/cli.ts
```diff
@@ -1,71 +1,46 @@
...
+cli.parse()
 
 export interface CliOptions {
-  files?: string[]
-  _?: string[]
   root?: string
   config?: string
   watch?: boolean
+  // silent?: boolean
+  // vue?: boolean
```
antoinerey, line 30: Are those two options still necessary? I could not find any reference to them. Were they forgotten in a previous refactor?

PR #1256, opened 2022-05-07
feat: self/window/global share state with globalThis

### LINKED ISSUE #1146: `globalThis` is not the same as `self` when using jsdom
## Observation

When testing react-components, you have to make sure that component-state changes are wrapped in an `act()` function call, otherwise the warning `Warning: An update to App inside a test was not wrapped in act(...).` will be displayed.

In the normal `create-react-app` setup, this warning is disabled when the code is currently running inside the `waitFor` function of `@testing-library/react`.

The same test, run with `vitest` shows the warning again. 

## Expectation

My expectation is that no warning is shown.

## Analysis

I figured out that `@testing-library/react` sets [self.IS_REACT_ACT_ENVIRONMENT](https://github.com/testing-library/react-testing-library/blob/c8c93f83228a68a270583c139972e79b1812b7d3/src/act-compat.js#L7) to `false`, while react [accesses `IS_REACT_ACT_ENVIRONMENT` directly](https://github.com/facebook/react/blob/d5b6b4b865ebf13a1eaf2342d623101056e5e197/packages/react-reconciler/src/ReactFiberAct.old.js#L44) to determine, whether to the warning or not.

In `jsdom`-environment of`vitest`, `self` and `globalThis` seem to be different object, so setting `self.IS_REACT_ACT_ENVIRONMENT` does not disable the warnings.

When I run the same test with `react-scripts test` (i.e. in jest), they are the same

## Expectation

`self` and `globalThis` should be the same object.

### Reproduction

I created the repository https://github.com/nknapp/vitest-react-18-testing-library-missing-act-workaround to reproduce the error. 

0. Setup dependencies

   ```
   yarn 
   ```

1. Run the test with `jest`, no warning will appear

   ```
   yarn test
   ```

2. Run the test with `vitest`, there will be a warning

   ```
   yarn vitest
   ```
    
3. Open [src/setupVitest.js](https://github.com/nknapp/vitest-react-18-testing-library-missing-act-workaround/blob/master/src/setupVitest.js) and remove the comments from the workaround-code.
The run `yarn vitest` again. No warning will appear

   The workaround defines `IS_REACT_ACT_ENVIRONMENT` as property that proxies its value to `self.IS_REACT_ACT_ENVIRONMENT`, so that the deactivation works correctly.

### System Info

```shell
System:
    OS: Linux 5.4 Ubuntu 20.04.4 LTS (Focal Fossa)
    CPU: (4) x64 Intel(R) Core(TM) i7-3687U CPU @ 2.10GHz
    Memory: 2.73 GB / 15.51 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 16.13.0 - ~/.nvm/versions/node/v16.13.0/bin/node
    Yarn: 1.22.15 - ~/.nvm/versions/node/v16.13.0/bin/yarn
    npm: 8.1.0 - ~/.nvm/versions/node/v16.13.0/bin/npm
  Browsers:
    Brave Browser: 99.1.36.122
    Chrome: 99.0.4844.84
    Firefox: 99.0
```

### Used Package Manager

yarn

### Validations

sheremet-va:
Fixes #1146


packages/vitest/src/integrations/env/jsdom.ts
```diff
@@ -40,26 +40,19 @@ export default <Environment>({
...
-        configurable: true,
-      })
-    }
+    const originals = new Map<string | symbol, any>(
+      allowRewrite.map(([key]) => [key, global[key]]),
+    )
 
     return {
       teardown(global) {
+        dom.window.document.head.innerHTML = ''
```
sheremet-va, line 51: This might break something with no-threads :)

sheremet-va:
Found it goes into infinite loop with `--no-threads`. Resolving.

Issue #256, opened 2021-12-21
Monorepo first class support


skimhugo:
### Clear and concise description of the problem

make monorepo first class support

### Suggested solution

make monorepo first class support

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

antfu:
Can you describe more about what you are expecting? Thanks

JakeGinnivan:
FWIW, I have taken a crack at switching the NX plugins repo for my old company across. https://github.com/sevenwestmedia-labs/nx-plugins/pull/61

My main issue so far are console output differences when running vitest via their command runners.

I will likely create a vitest plugin soon which will automate the conversion from jest -> vitest in a NX mono repo and provide the executors to run vitest tests for that project

ocavue:
Jest has a [`projects`](https://jestjs.io/docs/configuration#projects-arraystring--projectconfig) config option, which allows me to run test cases from multiple packages within one monorepo.

For example, my project could have this file structure:

```
my-monorepo/
    packages/
        pkg-a/
            e2e-tests/
            unit-tests/
            src/
            package.json
            jest.unit.config.mjs
            jest.e2e.config.mjs
        pkg-b/
            unit-tests/
            src/
            package.json
            jest.unit.config.mjs
    package.json
    jest.config.mjs
```


```ts
// my-monorepo/jest.config.mjs
export default {
    projects: ["./packages/*/jest*.config.mjs"],
}
```

```ts
// my-monorepo/packages/pkg-a/jest.unit.config.mjs
export default {
    displayName: { name: "pkg-a:unit", color: "blackBright" },
    setupFilesAfterEnv: ["jest-extended"],
    testEnvironment: "jsdom",  
}
```

```ts
// my-monorepo/packages/pkg-a/jest.e2e.config.mjs
export default {
    displayName: { name: "pkg-a:e2e", color: "blackBright" },
    setupFilesAfterEnv: ["jest-extended"],
    testEnvironment: "playwright",
}
```


```ts
// my-monorepo/packages/pkg-b/jest.config.mjs
export default {
[...]

PR #1278, opened 2022-05-09
feat!: always run tests inside worker

antfu:
## Breaking Changes

- `threads: false` now runs all tests in the same worker, instead of the main process
- the `isolate` option is now orthogonal to `threads`. By solely `threads: false`, the context will still be destroyed after each run. Combining `threads: false` and `isolate: false` to get the previous behavior of `threads: false` that context will leak across tests.
- `env` options will now only affect the tests but not the main process's `process.env`

This PR is for the preparation of supporting multiple projects #256


packages/vitest/src/node/pool.ts
```diff
@@ -74,32 +39,54 @@ export function createWorkerPool(ctx: Vitest): WorkerPool {
...
-        port.close()
-        workerPort.close()
-      }))
+      await pool.run(data, { transferList: [workerPort], name: 'run' })
+      port.close()
+      workerPort.close()
     }
+
+    if (!ctx.config.isolate && !ctx.config.threads)
+      await runFiles(files) // single thread with no isolation
```
sheremet-va, line 83: I think `--no-threads` should always run all files in a single worker, otherwise there is no point in `--no-threads`. It's _literally_ the same as running `--threads` - no alternative. And the name becomes misleading.

I think `--isolate` should just not affect `--no-threads`. I think in the future we might use `vm.runInContext` for isolation in `--no-threads`.

PR #1300, opened 2022-05-12
fix: `.resolves` and `.rejects` expectations

voraciousdev:
### Summary

The promise helpers, `.resolves` and `.rejects`, were throwing errors when the returned object could not be serialized. Additionally, the `.resolves` helper did not support a function argument even though a function was expected.

#### 1. Promise resolution

With a happy-path test for resolving a promise...

```ts
// example1.test.ts
await expect(() => Promise.resolve(1)).resolves.toEqual(1)
```

##### Before

The following error occurs.

```
TypeError: obj.then is not a function
```

##### After

The test passes.

```
1 passed (1)
```

#### 2. Dynamic imports

With an alternate-path test expecting a dynamic import (promise) to error...

```ts
// example2.ts
export default {}
```

```ts
// example2.test.ts
await expect(() => import('./example')).rejects.toThrowError()
```

##### Before

The following error occurs.

```
TypeError: Cannot convert object to primitive value
```

##### After

The test passes.

```
1 passed (1)
```

### DIFF
--- a/packages/vitest/src/integrations/chai/jest-expect.ts
+++ b/packages/vitest/src/integrations/chai/jest-expect.ts
@@ -61,6 +61,11 @@ export const JestChaiExpect: ChaiPlugin = (chai, utils) => {
             throw object
           })
         }
+        else if (promise === 'resolves') {
+          utils.flag(this, 'object', () => {
+            return object
+          })
+        }
         _super.apply(this, args)
       }
     })
--- a/packages/vitest/src/integrations/chai/jest-expect.ts
+++ b/packages/vitest/src/integrations/chai/jest-expect.ts
@@ -1,4 +1,5 @@
 import c from 'picocolors'
+import { AssertionError } from 'chai'
 import type { EnhancedSpy } from '../spy'
 import { isMockFunction } from '../spy'
 import { addSerializer } from '../snapshot/port/plugins'
@@ -56,15 +57,26 @@ export const JestChaiExpect: ChaiPlugin = (chai, utils) => {
       return function (this: Chai.Assertion & Chai.AssertionStatic, ...args: any[]) {
         const promise = utils.flag(this, 'promise')
         const object = utils.flag(this, 'object')
+        const isNot = utils.flag(this, 'negate') as boolean
         if (promise === 'rejects') {
           utils.flag(this, 'object', () => {
             throw object
           })
         }
-        else if (promise === 'resolves') {
-          utils.flag(this, 'object', () => {
-            return object
-          })
+        // if it got here, it's already resolved
+        // unless it tries to resolve to a function that should throw
+        // called as '.resolves[.not].toThrow()`
+        else if (promise === 'resolves' && typeof object !== 'function') {
+          if (!isNot) {
+            const message = utils.flag(this, 'message') || 'expected promise to throw an error, but it didn\'t'
+            const error = {
+              showDiff: false,
+            }
+            throw new AssertionError(message, error, utils.flag(this, 'ssfi'))
+          }
+          else {
+            return
+          }
         }
         _super.apply(this, args)
       }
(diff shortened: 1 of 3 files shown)


sheremet-va:
> Additionally, the .resolves helper did not support a function argument even though a function was expected.

What do you mean a function was expected? Promise is expected. Function _can_ be expected for `.rejects`, because it's a common pattern to wrap throwing in a function, and jest allows that, but I don't like putting it in `.resolves`.

voraciousdev:
@sheremet-va my apologies. I meant to say a function _should be allowed_ to maintain jest compatibility.

sheremet-va:
> @sheremet-va my apologies. I meant to say a function _should be allowed_ to maintain jest compatibility.

But your example will fail in jest. `resolves` expects a Promise, and should always expect a promise, not a function that returns a Promise. This is just extra steps.

voraciousdev:
@sheremet-va should it be removed from `.rejects` then? Either way, the two methods are currently inconsistent (I pulled that line from `.rejects`).

sheremet-va:
> @sheremet-va should it be removed from `.rejects` then? Either way, the two methods are currently inconsistent (I pulled that line from `.rejects`).

I know that you pulled that line from there. It is there for jest compatibility. As I said, it's a very common pattern that jest supports, and by extension, Vitest. Just remove the line from `.resolves` and PR can be merged.

> two methods are currently inconsistent

Being consistent is behaving the same as Jest does at this point of Vitest development. Maybe later, when we will want to diverge from that we will talk.

sheremet-va:
If you really want to help, I would advise adding a type guard inside `.resolves`/`.rejects` that makes sure that passed value is a Promise. To not get a weird error "obj.then is not a function".

voraciousdev:
@sheremet-va I see the [difference](https://github.com/facebook/jest/blob/3b0a975269a3b712d0235f67fbd072dd9cb867b2/packages/expect/src/index.ts#L220-L221) now. To be consistent with Jest, do you want me to just add that type guard to `.resolves` for now?

sheremet-va:
> @sheremet-va To be consistent with Jest, do you want me to just add that type guard to `.resolves` for now?

We can check inside both. In `.resolves` check the `obj` value, and in `.rejects` check that `wrapper` is a promise.

voraciousdev:
@sheremet-va I just pushed up another related commit. The reason I thought a function was required was due to the specific test I had written.

```ts
// make sure promise resolves
await expect(Promise.resolve(1)).resolves.not.toThrow()
```

This test caused the following error (hence why I thought it was required originally).

```
AssertionError: expected 1 to be a function
```

packages/vitest/src/integrations/chai/jest-expect.ts
```diff
@@ -579,6 +589,10 @@ export const JestChaiExpect: ChaiPlugin = (chai, utils) => {
     utils.flag(this, 'error', new Error('rejects'))
     const obj = utils.flag(this, 'object')
     const wrapper = typeof obj === 'function' ? obj() : obj // for jest compat
+
+    if (typeof wrapper?.then !== 'function')
+      throw new TypeError(`You must provide a Promise to expect() when using .rejects, not '${typeof wrapper}'.`)
```
antfu, line 594: Couldn't we do a `Promise.resolve` to normalize it?

sheremet-va: I think it’s better to have strict types, otherwise people test code that Vitest creates for them instead of the actual code.

packages/vitest/src/integrations/chai/jest-expect.ts
```diff
@@ -60,6 +61,11 @@ export const JestChaiExpect: ChaiPlugin = (chai, utils) => {
             throw object
           })
         }
+        else if (promise === 'resolves') {
```
sheremet-va, line 64: we don't need that :)

sheremet-va (changes requested):

sheremet-va:
> The reason I thought a function was required was due to the specific test I had written.

`.resolves` is just a hellper that passes awaited value down to the matcher, so you wouldn't have to do `expect(await promise)`. So, in your case it should be `expect(Promise.resolve(() => ({})))`. 

I think it was created before `async/await` syntax, so it was easier to do then doing `promise.then(v => expect(v))`. This exist mostly for compatibility with jest.

`.rejects` is the most useful, because you don't need to write `try/catch` block.

sheremet-va:
I wanted to do some changes to your PR, but cannot commit. Can you grant access? Specifically, you didn't cover this cases:

```js
    await expect((async () => new Error('msg'))()).resolves.not.toThrow() // calls chai assertion
    await expect((async () => new Error('msg'))()).resolves.not.toThrow(Error) // calls our assertion
    await expect((async () => () => {
      throw new Error('msg')
    })()).resolves.toThrow()
    await expect((async () => () => {
      return new Error('msg')
    })()).resolves.not.toThrow()
    await expect((async () => () => {
      return new Error('msg')
    })()).resolves.not.toThrow(Error)
  })

  it('resolves trows chai', async () => {
    const assertion = async () => {
      await expect((async () => new Error('msg'))()).resolves.toThrow()
    }

    await expect(assertion).rejects.toThrowError('expected promise to throw an error, but it didn\'t')
  })

  it('resolves trows jest', async () => {
    const assertion = async () => {
      await expect((async () => new Error('msg'))()).resolves.toThrow(Error)
    }

    await expect(assertion).rejects.toThrowError('expected promise to throw an error, but it didn\'t')
  })
```

voraciousdev:
@sheremet-va you should be able to commit now.

PR #891, opened 2022-03-03
feat: web support (browser as an environment)

Aslemammad:
It is surely ready for review!

PR #1302, opened 2022-05-13
feat: init @vitest/browser package

antfu:
Rework #891

## Checklist

- [x] Init package and integration
- [ ] Stub node packages for Vitest runtime (in progress, encounter some bug)
- [ ] Fulfill tests (Browser only APIs, Vue and React components)
- [ ] Better log in terminal
- [ ] Show progress and error on the browser page
- [ ] Headless mode in CI
- [ ] Docs


packages/vitest/src/node/plugins/index.ts
```diff
@@ -65,9 +70,12 @@ export async function VitestPlugin(options: UserConfig = {}, ctx = new Vitest())
...
         (options as ResolvedConfig).defines = defines
 
-        const open = preOptions.ui && preOptions.open
-          ? preOptions.uiBase ?? '/__vitest__/'
-          : undefined
+        let open: string | boolean | undefined
+
+        if (preOptions.ui && preOptions.open)
+          open = preOptions.uiBase ?? '/__vitest__/'
+        else if (preOptions.web)
```
userquin, line 77: should be `preOptions.browser`

I'm checking it rn, open my own branch or just work directly here?

antfu: Nice catch, you can directly commit to this branch

userquin:
@antfu the ui at least loading on the browser, and also the tests are loaded, but we have some problem with `node:util` package to format messages, I haven't found a way to bypass it, see screenshots below.

My commit also includes:

- added root script to copy ui to browser dist on root build script
- changes some `config.web` to `config.browser` (unresolved conversation)
- changed `vitest` to `vitest/browser` on browser node module on `/__vitest_index__` request, it was loading `vitest` instead the `browser` module
- added `isNode`, `isBrowser and `config.browser` logic to `vitest` modules to avoid run tests on server (fails)

imagen

imagen

mangs:
Big fan of this feature. 😀

Is the plan for this to eventually integrate with something like Playwright for cross-browser testing? Or are you targeting a single browser to minimize dependencies?

Some context for my question: Playwright provides the same API for Chromium, WebKit, and Gecko. Being able to run WebKit tests on macOS via GitHub Actions would be really useful for simulating iOS devices. We do that right now on my team and it saves us quite a bit of QA time via visual regression testing. Being able to use Vitest in the browser instead of jsdom's simulation would be a big upgrade. Playwright version 1.22 just added experimental component testing which is beginning to do what you're trying to achieve here, so maybe you can collaborate somehow? https://www.youtube.com/watch?v=y3YxX4sFJbM Might save you some time.

bjarkihall:
@mangs as I see it, this PR is about running tests in general in the browser (as an environment) instead of jsdom/happy-dom, so it's not **just** for component tests but the browser it's running in, in general:

```ts
test('browser', () => {
  document.body.innerHTML = `<div id="test" style="outline: 1px solid red; height: 10px; width: 10px;"></div>`;
  const testEl = document.getElementById('test');
  expect(testEl.clientHeight).toBe(testEl.offsetHeight);
  testEl.style.border = '2px solid blue';
  expect(testEl.clientHeight).not.toBe(testEl.offsetHeight);
})
```

As soon as vitest can run the tests within the browser, it should be possible to run them via playwright (or any other browser automation tool) instead of manually.

gautelo:
Is there any movement on this? I've found virtual DOM to be too limited/restrictive for testing custom web-components, and really would like to see this feature become a reality.

Manish3323:
Adding a comment to subscribe

Aslemammad:
A message for those who see this PR! suggested by @userquin.

to test the `browser` you will need:
- clone the repo and checkout the `feat/web-runner` branch
- run `pnpm install && pnpm run build` from root folder
- `cd test/browser && vitest --browser`: if the tests are not there just hard refresh

Aslemammad:
- [ ] use modern-node-polyfills
- [ ] when BROWSER=* is true, the page opens sooner rather than lets vitest finds files, that's why in this case, it doesn't find some tests (sometimes)

userquin:
Can we add a delay ib the browser ui to solve it? I will check it later.

pnpm-lock.yaml
```diff
@@ -140,7 +140,7 @@ importers:
       '@rollup/plugin-graphql': 1.1.0_nhmex7luzljoezdhc6xbg5ot4a
       graphql: 16.5.0
     devDependencies:
-      '@vitest/ui': link:../../packages/ui
+      '@vitest/ui': 0.18.1
```
sheremet-va, line 143: This doesn't seem right

sheremet-va (changes requested):

cristian-spiescu:
Hi there! We have library that enriches the react-testing-library dev/test workflow, with visual features. We are "polishing" this lib, and we'd like to release it as open source. We currently user mocha. We are interested to use Jest/Jest compatible also, because this would fit the programming habits of more people (compared to our current setup).

That's why I find vitest-browser really interesting. Could you please share your plans/roadmap regarding this feature?

Aslemammad:
@cristian-spiescu Hey mate, Lately I've been talking to @userquin @JessicaSachs! and here's my plan:
- Implement CI compatibility
- Bring Proper selectors
- More stable environment

PR #1310, opened 2022-05-15
fix: replace root when resolving dependency

### LINKED ISSUE #1294: Can't mock dynamically imported packages
### Describe the bug

Scoped packages, e.g. `@wdio/logger` can't be mocked at all. In Jest it was possible to have the following structure:

```ts
__mocks__/
  - @wdio/
    - logger.ts
```

Jest would automatically pick this up.

### Reproduction

See https://stackblitz.com/edit/vitest-dev-vitest-tx14cb?file=package.json,test%2Fautomocking.spec.ts,__mocks__%2F%40wdio%2Flogger.ts&initialPath=__vitest__

I mocked out `@wdio/logger` in a way I expect Jest to work. The mocked file looks like this:
```ts
export default 'foobar';
```
In my test file (automocking.spec.ts) I do:
```ts
vi.mock('@wdio/test');
test('can automock scoped NPM package', () => {
  expect(foo).toBe('foobar');
});
```

The test failed due to:
```
can automock scoped NPM package
AssertionError: expected [Function getLogger] to be 'foobar' // Object.is equality
 - /test/automocking.spec.ts:38:41
``` 

### System Info

```shell
n/a
```

### Used Package Manager

yarn

### Validations

nieyuyao:
fix #1294.

the reason for this issue is that when we resolve all mocks, store them in `mockMap`. it looks like below,

```javascript
{
    '/node_modules/@wdio/logger/build/node.js': '/__mocks__/@wdio/logger.ts'
}
```

but when we actually import, the incoming path is `@wdio/logger`, it becomes `/@fs/${rootPath}/node_modules/@wdio/logger/build/node.js` after `resolveId` function. so the mock fails.

### DIFF
--- a/packages/vite-node/src/utils.ts
+++ b/packages/vite-node/src/utils.ts
@@ -8,6 +8,10 @@ export function slash(str: string) {
   return str.replace(/\\/g, '/')
 }
 
+export function mergeSlashes(str: string) {
+  return str.replace(/\/\//g, '/')
+}
+
 export function normalizeRequestId(id: string, base?: string): string {
   if (base && id.startsWith(base))
     id = `/${id.slice(base.length)}`
--- a/packages/vite-node/src/client.ts
+++ b/packages/vite-node/src/client.ts
@@ -4,7 +4,7 @@ import vm from 'vm'
 import { dirname, extname, isAbsolute, resolve } from 'pathe'
 import { isNodeBuiltin } from 'mlly'
 import createDebug from 'debug'
-import { isPrimitive, normalizeModuleId, normalizeRequestId, slash, toFilePath } from './utils'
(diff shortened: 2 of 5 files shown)


packages/vite-node/src/utils.ts
```diff
@@ -8,6 +8,10 @@ export function slash(str: string) {
   return str.replace(/\\/g, '/')
 }
 
+export function mergeSlashes(str: string) {
```
sheremet-va, line 11: we already have this function in `utils/base.ts`

sheremet-va (changes requested):

packages/vite-node/src/utils.ts
sheremet-va: Oh, I didn't notice. I guess no, we shouldn't. Maybe sometime we can create a utils package, but not in this issue, I think

sheremet-va (approved):

PR #1312, opened 2022-05-15
feat: support npm setup files

tmkx:
For example:
```js
{
  test: {
    setupFiles: ['@testing-library/jest-dom/extend-expect.js'],
  },
}
```

### DIFF
--- a/packages/vitest/src/node/config.ts
+++ b/packages/vitest/src/node/config.ts
@@ -1,4 +1,5 @@
-import { resolve } from 'pathe'
+import { resolveModule } from 'local-pkg'
+import { normalize, resolve } from 'pathe'
 import c from 'picocolors'
 import type { ResolvedConfig as ResolvedViteConfig } from 'vite'
 
@@ -122,7 +123,7 @@ export function resolveConfig(
   if (process.env.VITEST_MIN_THREADS)
     resolved.minThreads = parseInt(process.env.VITEST_MIN_THREADS)
 
-  resolved.setupFiles = toArray(resolved.setupFiles || []).map(file => resolve(resolved.root, file))
+  resolved.setupFiles = toArray(resolved.setupFiles || []).map(file => normalize(resolveModule(file, { paths: [resolved.root] })!))
 
   // the server has been created, we don't need to override vite.server options
   resolved.api = resolveApiConfig(options)


packages/vitest/src/node/config.ts
```diff
@@ -122,13 +123,15 @@ export function resolveConfig(
   if (process.env.VITEST_MIN_THREADS)
     resolved.minThreads = parseInt(process.env.VITEST_MIN_THREADS)
 
-  resolved.setupFiles = toArray(resolved.setupFiles || []).map(file => resolve(resolved.root, file))
+  const require = createRequire(resolve(resolved.root, '_'))
```
antfu, line 126:
```suggestion
  const require = createRequire(resolve(resolved.root, '_'))
```

What's the `_` for?

tmkx: just a placeholder, `createRequire` needs a _file_, maybe using `index.js` more readable?

antfu: Can we use `resolveModule` from `local-pkg`?

tmkx: I have tried, `resolveModule` uses `const _require = createRequire(import.meta.url)`, it's hard code as `import.meta.url`, but we need relative to `root` here. otherwise causes some bad cases.

sheremet-va: This will incorrectly resolve esm files. We use Vite's resolver, when importing files.

tmkx: https://github.com/vitest-dev/vitest/blob/03a54e0651f51b739b7833817d14d2cbcf474ede/packages/vitest/src/node/config.ts#L125

The `setupFiles` will prepend `config.root` path directly,
```js
{
  test: {
    setupFiles: ['@testing-library/jest-dom/extend-expect.js'],
  },
}
```

will be transformed to:
```js
{
  test: {
    setupFiles: ['/path/to/@testing-library/jest-dom/extend-expect.js'],
  },
}
```

it should be:
```js
{
  test: {
    setupFiles: ['/path/to/node_modules/@testing-library/jest-dom/extend-expect.js'],
  },
}
```

sheremet-va (approved):

sheremet-va:
LGTM 👍🏻

dammy001 (approved):

PR #1327, opened 2022-05-17
fix: test context fixed on TypeScript

semoal:
Hi 👋
I've implemented some tests that need some asynchronous data before starting and passed the id's to the context, but the TestContext interface didn't work. 

<table>
<tr>
	<td>Before</td>
	<td>After</td>
<tr>
	<td>Screenshot 2022-05-17 at 3 13 37 PM
</td>
	<td>Screenshot 2022-05-17 at 3 13 43 PM

</td>
</table>

### DIFF
--- a/docs/guide/test-context.md
+++ b/docs/guide/test-context.md
@@ -48,7 +48,7 @@ To provide type for your custom context properties, you can aggregate the type `
 
 ```ts
 declare module 'vitest' {
-  export interface TestContext {
+  export interface Suite {
     foo?: string
   }
 }


semoal (commented):
Also I'm not sure about this L:41 , I had to change this line to:

`context.meta.suite.portfolioId`

If you're all agree I can change this too

sheremet-va:
I think Suite is another thing. Maybe we have an error in out type logics somewhere.

semoal:
That makes a lot of sense, sorry 4 the confusion. I'm used to Jest where "you could do anything" like global `let` and I've run into some troubles with Vitest but has much sense. 
 
Thanks, @antfu  and @sheremet-va!

PR #1340, opened 2022-05-20
fix: simplify globals access, fix global != global

### LINKED ISSUE #1323: Happy DOM - window.Blob is not equal to globalThis.Blob
### Describe the bug

Hi! :slightly_smiling_face: 

A bug was reported in [Happy DOM](https://github.com/capricorn86/happy-dom) related to "new Blob()" not being an instance of globalThis.Blob.

There seem to be two different implementations of Blob used in window.Blob vs globalThis.Blob. Node does not support Blob, so there must be something else adding the second implementation.

I investigated a bit further and I also noticed that window isn't the same as globalThis and global. Vitest should probably set `global.window` to be `global`.

Related bug in [Happy DOM](https://github.com/capricorn86/happy-dom):
https://github.com/capricorn86/happy-dom/issues/471

### Reproduction

```javascript
// Outputs "false"
console.log(window.globalThis === globalThis);

// Outputs "false"
console.log(window.Blob === globalThis.Blob);

// Outputs "false"
console.log(window.globalThis.Blob === globalThis.Blob);

// Outputs "false"
console.log(Blob === globalThis.Blob);

// Outputs "false"
console.log(Blob === globalThis.Blob);
```

### System Info

```shell
System:
    OS: Linux 5.17 Pop!_OS 22.04 LTS
    CPU: (4) x64 Intel(R) Core(TM) i5-7600K CPU @ 3.80GHz
    Memory: 6.43 GB / 15.56 GB
    Container: Yes
    Shell: 5.8.1 - /usr/bin/zsh
  Binaries:
    Node: 16.15.0 - /usr/bin/node
    Yarn: 1.22.10 - ~/.npm-global/bin/yarn
    npm: 8.5.5 - /usr/bin/npm
  Browsers:
    Chrome: 101.0.4951.41
    Firefox: 100.0
```

### Used Package Manager

npm

### Validations

### LINKED ISSUE #1335: Possible regression from 0.10.5 to 0.11.0 onward
### Describe the bug

Hi, I found when upgrading to a newer version to vitest, my tests in a react app (with react-easy-state) break which were fine before and the code itself is working as intended.

I get the following error:
```
 FAIL  test/component.spec.tsx [ test/component.spec.tsx ]
 ❯ timerInitializationSteps node_modules/jsdom/lib/jsdom/browser/Window.js:535:25
 ❯ Window.setTimeout node_modules/jsdom/lib/jsdom/browser/Window.js:465:12
 ❯ Object.apply node_modules/@risingstack/react-easy-state/dist/cjs.es6.js:280:24
```

### Reproduction

I created a minimal example to reproduce.
If the vitest version is set to 0.10.5 or ealier it works perfectly fine, later versions break.

https://stackblitz.com/edit/vitest-dev-vitest-cvqafs?file=test%2Fcomponent.tsx

### System Info

```shell
See Stackblitz
```

### Used Package Manager

npm

### Validations

sheremet-va:
Fixes #1323
Fixes #1335

### DIFF
--- a/test/core/test/dom.test.ts
+++ b/test/core/test/dom.test.ts
@@ -119,3 +119,10 @@ it('can call global functions without window works as expected', async () => {
   expect(() => self.requestAnimationFrame(noop)).not.toThrow()
   expect(() => globalThis.requestAnimationFrame(noop)).not.toThrow()
 })
+
+it('globals are the same', () => {
+  expect(window.globalThis).toBe(globalThis)
+  expect(window.Blob).toBe(globalThis.Blob)
+  expect(window.globalThis.Blob).toBe(globalThis.Blob)
+  expect(Blob).toBe(globalThis.Blob)
+})
--- a/packages/vitest/src/integrations/env/happy-dom.ts
+++ b/packages/vitest/src/integrations/env/happy-dom.ts
@@ -10,11 +10,7 @@ export default <Environment>({
     const { Window, GlobalWindow } = await importModule('happy-dom') as typeof import('happy-dom')
     const win = new (GlobalWindow || Window)()
 
-    const { keys, allowRewrite } = populateGlobal(global, win, { bindFunctions: true })
-
-    const originals = new Map<string | symbol, any>(
-      allowRewrite.map(([key]) => [key, global[key]]),
-    )
+    const { keys, originals } = populateGlobal(global, win, { bindFunctions: true })
 
     return {
       teardown(global) {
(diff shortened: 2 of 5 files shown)


test/core/test/dom.test.ts
```diff
@@ -119,3 +119,10 @@ it('can call global functions without window works as expected', async () => {
   expect(() => self.requestAnimationFrame(noop)).not.toThrow()
   expect(() => globalThis.requestAnimationFrame(noop)).not.toThrow()
 })
+
+it('globals are the same', () => {
```
sheremet-va, line 123: maybe we need `test/dom`, where we can change env in ci

Issue #1328, opened 2022-05-17
vue: dynamic component imports are not resolved


cexbrayat:
### Describe the bug

With Vue, you can define async components with a dynamic import like the following:

```vue
<template>
  <button @click="showHello = true">Show Hello</button>
  <div v-if="showHello">
    <div>Async component</div>
    <Hello :count="1"></Hello>
  </div>
</template>

<script setup lang="ts">
import { defineAsyncComponent, ref } from 'vue';

const Hello = defineAsyncComponent(() => import('./Hello.vue'));
const showHello = ref(false);
</script>
```

`Hello.vue` is then loaded only when the user clicks on the button `Show Hello`

But a unit test checking this behavior fails with Vitest:

```ts
import { mount, flushPromises } from '@vue/test-utils';
import AsyncComp from '../components/AsyncComp.vue';

test('mount component', async () => {
  const wrapper = mount(AsyncComp);

  expect(wrapper.text()).toContain('Show Hello');

  await wrapper.get('button').trigger('click');
  expect(wrapper.vm.showHello).toBe(true);
  await flushPromises();
  // I suppose this should work but does not
  expect(wrapper.text()).toContain('1 x 2');
});
```

Maybe I'm missing something obvious in the test setup, but it looks like the promise is never resolved.

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-bmfn8b

### System Info

```shell
System:
    OS: macOS 12.3.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 16.90 GB / 64.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.15.0 - ~/.volta/tools/image/node/16.15.0/bin/node
    Yarn: 1.22.17 - ~/.volta/tools/image/yarn/1.22.17/bin/yarn
    npm: 6.14.17 - ~/.volta/tools/image/npm/6.14.17/bin/npm
  Browsers:
    Chrome: 101.0.4951.64
    Firefox: 100.0
    Safari: 15.4
  npmPackages:
    @vitejs/plugin-vue: 2.3.3 => 2.3.3
    vite: 2.9.9 => 2.9.9
    vitest: 0.12.6 => 0.12.6
```

### Used Package Manager

yarn

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
[...]

PR #1359, opened 2022-05-23
feat: add dynamicImportSettled utility function

sheremet-va:
Context #1328

### DIFF
--- a/packages/vitest/src/integrations/vi.ts
+++ b/packages/vitest/src/integrations/vi.ts
@@ -1,7 +1,7 @@
 import type { FakeTimerInstallOpts } from '@sinonjs/fake-timers'
 import { parseStacktrace } from '../utils/source-map'
 import type { VitestMocker } from '../runtime/mocker'
-import { getWorkerState, resetModules } from '../utils'
+import { getWorkerState, resetModules, setTimeout } from '../utils'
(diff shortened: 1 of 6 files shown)


sheremet-va:
@cexbrayat what do you think about this?

packages/vitest/src/integrations/vi.ts
```diff
@@ -234,6 +234,24 @@ class VitestUtils {
...
+   */
+  public async dynamicImportSettled() {
+    const state = getWorkerState()
+    const promises: Promise<unknown>[] = []
+    for (const mod of state.moduleCache.values()) {
+      if (!mod.exports && mod.promise)
+        promises.push(mod.promise)
+    }
+    await Promise.all(promises)
+    // wait untill the end of the loop, so `.then` on modules called,
```
cexbrayat, line 251: nit: untill -> until

cexbrayat (approved):
That looks great 👍

packages/vitest/src/integrations/vi.ts
antfu, line 251:
```suggestion
    // wait until the end of the loop, so `.then` on modules called,
```

antfu (approved):

PR #1378, opened 2022-05-26
feat: Export snapshot state to `expect.extend`.

cpojer:
In order to use [`jest-image-snapshot`](https://github.com/americanexpress/jest-image-snapshot), the snapshot state must be exposed as `MatcherState` in `expect.extend()` calls. This makes the plugin work, and I added a quick example as well.


examples/image-snapshot/test/basic.test.ts
```diff
@@ -0,0 +1,17 @@
...
+    interface Matchers<R> {
+      toMatchImageSnapshot(): R
+    }
+  }
+}
+
+expect.extend({ toMatchImageSnapshot })
+
+test('image snapshot', () => {
+  expect(readFileSync('./test/stubs/input-image.png')).toMatchImageSnapshot()
```
antfu, line 16: Do we have a license/credit for this image?

cpojer: Updated it with the vitest logo, so we don't have to get into those concerns (Just figured I'll use a real example from my code ;) ).

antfu (approved):

PR #1379, opened 2022-05-26
fix: birpc using safe timers

### LINKED ISSUE #1154: Using `console.log` in test when `useFakeTimers` results in unexpected timer added from `birpc` package
### Describe the bug

If a test contains a call to `console`, a `setTimeout` is invoked by the `birpc` package, which when combined with `useFakeTimers` can lead to unexpected results.

This does appear to only occur on _first_ usage of `console`. Subsequent usages the behaviour is not exhibited. As such I've found a workaround by doing the following:

```ts
function timerBugWorkaround() {
    console.log();
    vi.clearAllTimers();
  }

beforeEach(() => {
    vi.useFakeTimers();
    timerBugWorkaround(); // Uncomment to invoke workaround
});
```

Unfortunately for me, I have spent many hours trying to figure out what was going on here when extra timers were being added from seemingly nowhere. Whilst this bug may not be a major issue, it does result in unexpected changes for something as common as using `console.log` (which is default go-to debugging tool before reaching for the inspector).

I believe this did also result in some issues from timers which are added later being automatically run without explicitly being advanced, however I can't seem to get a solid repro of that issue at this time—I haven't investigated further as I do believe it was ultimately caused by this issue.

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-qmkkke?file=test/basic.test.ts 

### System Info

```shell
Not relevant, repro in StackBlitz
```

### Used Package Manager

npm

### Validations

sheremet-va:
Fixes #1154

### DIFF
--- a/packages/vitest/src/runtime/rpc.ts
+++ b/packages/vitest/src/runtime/rpc.ts
@@ -1,5 +1,13 @@
-import { getWorkerState } from '../utils'
+import { getWorkerState, withSafeTimers } from '../utils'
 
 export const rpc = () => {
-  return getWorkerState().rpc
+  const { rpc } = getWorkerState()
+  return new Proxy(rpc, {
+    get(target, p, handler) {
+      const sendCall = Reflect.get(target, p, handler)
+      const safeSendCall = (...args: any[]) => withSafeTimers(() => sendCall(...args))
+      safeSendCall.asEvent = sendCall.asEvent
+      return safeSendCall
+    },
+  })
 }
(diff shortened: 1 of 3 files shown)


antfu:
Not very sure about the perf impact but I think we could have it for now

antfu (approved):

sxzz:
After this PR, the unit test of Element Plus failed... I don't know why.
https://github.com/element-plus/element-plus/blob/dev/packages/components/autocomplete/__tests__/autocomplete.test.tsx#L89

Issue #1290, opened 2022-05-11
deep equality check failing message is very different compare to Jest


ghiscoding:
### Clear and concise description of the problem

I recently switched from Jest to Vitest and very often in Jest I would use the `.toBe()` instead of `.toEqual()` since I never remember which one is which and in Jest when we do use the wrong one then we have a great useful message telling us that we probably meant to use another validity check. This is very helpful but in Vitest it just fails equality check without giving us any hint at all. 

For example, if we have this test that we run in both libs
```ts
expect([{ id: 123, name: 'John', age: 30 }])
 .toBe([{ id: 123, name: 'John', age: 30 }]);
```

in Jest we have this helpful message "serializes to the same string" and also other suggestions
```shell
expect(received).toBe(expected) // Object.is equality

    If it should pass with deep equality, replace "toBe" with "toStrictEqual"

    Expected: [{"age": 30, "id": 123, "name": "John"}]
    Received: serializes to the same string

      at Object.<anonymous> (/src/index.spec.ts:34:50)
```
but in Vitest, it just fails without providing any hint of our problem
```shell
 FAIL  src/index.spec.ts > My Test Suite > should have the same array
AssertionError: expected [ { id: 123, name: 'John', age: 30 } ] to be [ { id: 123, name: 'John', age: 30 } ] // Object.is equality
 ❯ src/stores/__tests__/viewStore.spec.ts:63:49
     61|     expect(store.selectedTab).toBe('compId');
     62| 
     63|     expect([{ id: 123, name: 'John', age: 30 }]).toBe([{ id: 123, name: 'John', age: 30 }]);
       |                                                 ^
     64|   });
     65| });
```

A new developer using Vitest will most certainly be confused and won't know why it failed, then do a search on Google or Stack Overflow. However in Jest, it gives us a great hint of how to modify our code and use the correct check. Half of the time I use the wrong one, so I will very often see this failing... to be or not to be 😄 

### Suggested solution

Use the same message as Jest which helps user finding the issue with the test. The message that always remind me that I used the wrong check is 
```shell
Received: serializes to the same string
```
then I know right away that I must use `.toEqual()` instead of `.toBe()` 

A very simple message that can so so helpful

### Additional context

I ran the same test with both libs at latest versions, Jest 28 and Vitest `0.12.4`
### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
[...]

PR #1383, opened 2022-05-27
feat: improve deep equality check failing message for toBe

TommyDew42:
## What does this PR do?

A PR for issue #1290.

We would like to have a message to remind users they should use deep equality checks, when object comparison fails with `toBe`. The error message would look like this now:
Screenshot 2022-05-27 at 12 12 49 PM

The implementation is similar to how [jest](https://github.com/facebook/jest/blob/main/packages/expect/src/matchers.ts#L85-L124) does it.

## How the change is tested?
Unit tests

### DIFF
--- a/packages/vitest/src/integrations/chai/jest-expect.ts
+++ b/packages/vitest/src/integrations/chai/jest-expect.ts
@@ -131,7 +131,7 @@ export const JestChaiExpect: ChaiPlugin = (chai, utils) => {
     const actual = this._obj
     const pass = Object.is(actual, expected)
 
-    let deepEqualityName = '';
+    let deepEqualityName = ''
 
     if (!pass) {
       const toStrictEqualPass = jestEquals(
--- a/packages/vitest/src/integrations/chai/jest-expect.ts
+++ b/packages/vitest/src/integrations/chai/jest-expect.ts
@@ -7,7 +7,7 @@ import type { Constructable, Test } from '../../types'
 import { assertTypes } from '../../utils'
 import { unifiedDiff } from '../../node/diff'
 import type { ChaiPlugin, MatcherState } from '../../types/chai'
-import { arrayBufferEquality, iterableEquality, equals as jestEquals, sparseArrayEquality, subsetEquality, typeEquality, generateToBeMessage } from './jest-utils'
+import { arrayBufferEquality, generateToBeMessage, iterableEquality, equals as jestEquals, sparseArrayEquality, subsetEquality, typeEquality } from './jest-utils'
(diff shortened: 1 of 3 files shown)


ghiscoding:
Thanks for working on the issue that I opened, however I still prefer how Jest shows the error, the text is more aligned especially around the expectations of expected/received. Also I see you added square brackets around the "serializes to the same string" but I find that a little bit confusing, I didn't know (until I looked at your code change) that the bracket was something that you added, I wasn't sure if the text inside it was dynamic or not (it's not) and that made it confusing, it's also not an array so why use brackets and not double quotes? Nonetheless, it's a great improvement compare to current code, so I'd be curious to see what the core team will reply. Thanks again for working on the issue

I like how Jest shows it, everything is aligned and more readable
```sh
expect(received).toBe(expected) // Object.is equality

    If it should pass with deep equality, replace "toBe" with "toStrictEqual"

    Expected: [{"age": 30, "id": 123, "name": "John"}]
    Received: serializes to the same string

      at Object.<anonymous> (/src/index.spec.ts:34:50)
```

packages/vitest/src/integrations/chai/jest-expect.ts
```diff
@@ -129,9 +129,44 @@ export const JestChaiExpect: ChaiPlugin = (chai, utils) => {
...
+        const toEqualPass = jestEquals(
+          actual,
+          expected,
+          [iterableEquality],
+        )
+
+        if (toEqualPass) {
+          reminderMessage = 'If it should pass with deep equality, replace "toBe" with "toEqual"\n\n'
+
+          toBeMessage = `${reminderMessage}expected #{this} to be [serializes to the same string] // Object.is equality`
```
sheremet-va, line 162: `expected to be` should be at the top. this is error message.

`expected/received` block is rendered independently

sheremet-va (changes requested):

TommyDew42:
@ghiscoding @sheremet-va 

Does the current error message look good to us?
Screenshot 2022-06-02 at 1 03 00 AM

ghiscoding:
> @ghiscoding @sheremet-va
> 
> Does the current error message look good to us? Screenshot 2022-06-02 at 1 03 00 AM

Yeah this is much closer to Jest message, love it

packages/vitest/src/integrations/chai/jest-expect.ts
TommyDew42: Solved in commit [9cc61](https://github.com/vitest-dev/vitest/pull/1383/commits/9cc61ea2bdba3ed7cab1a93bb9020ed98e8356c9).

sheremet-va (approved):

PR #1406, opened 2022-05-31
fix(vite-node): Prevent crash when passing single module as options

antoinerey:
Right now, when passing a single module with the new `--options` CLI flag, `cac` returns a string instead of an array. This causes a crash, trying to call `map` on it.

```
❯ ./packages/vite-node/vite-node.mjs --options.deps.inline="module" hello.js
file:///Users/antoine/Code/vitest/packages/vite-node/dist/cli.js:697
      inline: (_b = (_a = serverOptions.deps) == null ? void 0 : _a.inline) == null ? void 0 : _b.map((dep) => {
                                                                                                  ^

TypeError: _b.map is not a function
    at parseServerOptions (file:///Users/antoine/Code/vitest/packages/vite-node/dist/cli.js:697:99)
    at CAC.run (file:///Users/antoine/Code/vitest/packages/vite-node/dist/cli.js:660:49)
    at CAC.runMatchedCommand (file:///Users/antoine/Code/vitest/packages/vite-node/dist/cli.js:622:34)
    at CAC.parse (file:///Users/antoine/Code/vitest/packages/vite-node/dist/cli.js:549:12)
    at file:///Users/antoine/Code/vitest/packages/vite-node/dist/cli.js:652:5
    at ModuleJob.run (node:internal/modules/esm/module_job:185:25)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:281:24)
```

This PR updates the typings to reflect that, and makes sure that both `string` and `string[]` are properly handled.

### DIFF
--- a/packages/vite-node/src/cli.ts
+++ b/packages/vite-node/src/cli.ts
@@ -89,16 +89,32 @@ async function run(files: string[], options: CliOptions = {}) {
 }
 
 function parseServerOptions(serverOptions: ViteNodeServerOptionsCLI): ViteNodeServerOptions {
+  const serverOptionsDepsInline = typeof serverOptions.deps?.inline === 'string'
+    ? [serverOptions.deps?.inline]
+    : serverOptions.deps?.inline
+
+  const serverOptionsDepsExternal = typeof serverOptions.deps?.external === 'string'
+    ? [serverOptions.deps?.external]
+    : serverOptions.deps?.external
+
+  const serverOptionsTransformModeSsr = typeof serverOptions.transformMode?.ssr === 'string'
+    ? [serverOptions.transformMode?.ssr]
+    : serverOptions.transformMode?.ssr
+
+  const serverOptionsTransformModeWeb = typeof serverOptions.transformMode?.web === 'string'
+    ? [serverOptions.transformMode?.web]
+    : serverOptions.transformMode?.web
+
   return {
     ...serverOptions,
     deps: {
       ...serverOptions.deps,
-      inline: serverOptions.deps?.inline?.map((dep) => {
+      inline: serverOptionsDepsInline?.map((dep) => {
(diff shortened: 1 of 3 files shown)


packages/vite-node/src/cli.ts
```diff
@@ -108,18 +124,18 @@ function parseServerOptions(serverOptions: ViteNodeServerOptionsCLI): ViteNodeSe
     transformMode: {
       ...serverOptions.transformMode,
 
-      ssr: serverOptions.transformMode?.ssr?.map(dep => new RegExp(dep)),
-      web: serverOptions.transformMode?.ssr?.map(dep => new RegExp(dep)),
```
antoinerey, line 112: There was a mistake here. It should have been `web` instead. It's now fixed.

packages/vite-node/src/cli.ts
```diff
@@ -89,16 +89,32 @@ async function run(files: string[], options: CliOptions = {}) {
...
+  const serverOptionsTransformModeWeb = typeof serverOptions.transformMode?.web === 'string'
+    ? [serverOptions.transformMode?.web]
+    : serverOptions.transformMode?.web
+
   return {
     ...serverOptions,
     deps: {
       ...serverOptions.deps,
-      inline: serverOptions.deps?.inline?.map((dep) => {
+      inline: serverOptionsDepsInline?.map((dep) => {
```
antfu, line 112: You may just use [`toArray`](https://github.com/vitest-dev/vitest/blob/574d072c9250747bd66d09ca1e22df0aeb55681f/packages/vitest/src/utils/base.ts#L84) (you can move it from vitest to vite-node)

antoinerey:
CI failed, but I believe it has nothing to do with the changes. Should I do something?

PR #1416, opened 2022-06-02
docs: new theme

antfu:
blocked by https://github.com/vuejs/vitepress/pull/702


userquin:
@antfu wow man, looks awesome, congrats to Kia: you can also use the license entries just check the vue docs and its footer

@patak-dev the background seems to be missing... :cry:

PR #1424, opened 2022-06-04
feat: improve watcher performance, add forceRerunTriggers option

sheremet-va:
Few things to note here: 

- `watchIgnore` is renamed to `watchExclude`
- `watchExclude` now accepts only glob patterns

> Now, instead of ignoring a file for rerun, we don't even watch for it. This should improve errors in docker about too much watches, and generally improve performance, because no more excessive calls.

- added `forceRerunTriggers` option 
- `forceRerunTriggers` accepts glob pattern for files, that if changed, will cause a full suite rerun
- `forceRerunTriggers` respects `watchExclude`

> It is useful, when running CLI tests, because Vite cannot build a module graph for `exec` calls


antfu:
What's `dumb` stands for?

sheremet-va:
> What's `dumb` stands for?

"not smart" 👀 

I've recently worked on a project involving dumb devices, so I thought why not 😄  Also, the name was proposed as a joke in discord :P

jgoux:
> > What's `dumb` stands for?
> 
> "not smart" 👀
> 
> I've recently worked on a project involving dumb devices, so I thought why not 😄 Also, the name was proposed as a joke in discord :P

Sorry for this terrible joke. 😂

If the default watch mode is "smart", maybe instead we could have a `--watchMode=smart|dumb`, and if `watchMode` is set to `dumb`, then we could have `watchInclude` taken into account?

packages/vitest/src/node/core.ts
```diff
@@ -122,8 +122,8 @@ export class Vitest {
         this.console.error(c.dim('filter:  ') + c.yellow(filters.join(comma)))
       if (this.config.include)
         this.console.error(c.dim('include: ') + c.yellow(this.config.include.join(comma)))
-      if (this.config.watchIgnore)
-        this.console.error(c.dim('ignore:  ') + c.yellow(this.config.watchIgnore.join(comma)))
+      if (this.config.watchExclude)
+        this.console.error(c.dim('ignore:  ') + c.yellow(this.config.watchExclude.join(comma)))
```
antfu, line 126:
```suggestion
        this.console.error(c.dim('exclude:  ') + c.yellow(this.config.watchExclude.join(comma)))
```

packages/vitest/src/types/config.ts
```diff
@@ -198,11 +198,18 @@ export interface InlineConfig {
...
+  /**
+   * Glob patter of file paths that will trigger the whole suite rerun
    *
-   * @default [/\/node_modules\//, /\/dist\//]
+   * Useful if you are testing calling CLI commands
+   *
+   * @default []
    */
-  watchIgnore?: (string | RegExp)[]
+  dumbWatchInclude?: string[]
```
antfu, line 212: Emm, maybe:

- `forceRerunTargets`
- `forceRerunTriggers`
- `forceRerunEntries`

sheremet-va: I want to be consistent with "exclude/include"

packages/vitest/src/node/core.ts
sheremet-va: I think `exclude` is a bit misleading. Renamed to `watch exclude`

antfu (approved):

MichaelDeBoey:
> `watchIgnore` is renamed to `watchExclude`

This wasn't clear watching only the changelog

sheremet-va:
> > `watchIgnore` is renamed to `watchExclude`
> 
> This wasn't clear watching only the changelog

Sorry, thank you for pointing that out. Will do better next time.

PR #1431, opened 2022-06-05
fix: show full name for external mod at graph

### LINKED ISSUE #1374: Module graph reports multiple times same modules imports
### Describe the bug

I have been playing around with the module graph and I was curious if the reported modules are actually duplicated or if it's just an analysis of the file imports because it looks like lodash is represented more than once and some other packages (mostly @mui related but even testing library appears multiple times).

Screenshot 2022-05-25 at 7 19 38 PM

### Reproduction

https://github.com/underfisk/vitest-module-graph

### System Info

```shell
System:
    OS: macOS 12.3.1
    Memory: 2.49 GB / 64.00 GB
    Shell: 3.2.2 - /usr/local/bin/fish
  Binaries:
    Node: 16.15.0 - ~/.volta/tools/image/node/16.15.0/bin/node
    Yarn: 1.22.18 - ~/.volta/tools/image/yarn/1.22.18/bin/yarn
    npm: 8.5.5 - ~/.volta/tools/image/node/16.15.0/bin/npm
  Browsers:
    Chrome: 101.0.4951.64
    Safari: 15.4
```

### Used Package Manager

yarn

### Validations

nieyuyao:
fix #1374


packages/ui/client/composables/module-graph.ts
```diff
@@ -12,7 +12,7 @@ export type ModuleGraphConfig = GraphConfig<ModuleType, ModuleNode, ModuleLink>
 function defineExternalModuleNode(module: string): ModuleNode {
   let label = module
   if (label.includes('/node_modules/'))
-    label = label.split(/\/node_modules\//g).pop()!.split(/\//g).shift()!
+    label = label.split(/\/node_modules\//g).pop()!
```
sheremet-va, line 15: Will it show the full name? Like `@org/something/dist/esm/index.esm.js`? Maybe we should just detect `@org` and show `@org/something` in this case?

Although if we have something like `package/node` and `package/client`, i still will see only `package` 🤔 Hm, I wonder if there is a way to show only what was imported 🤔 I think we always get the full path 😞

nieyuyao: It would be better if it could be expressed in short form, such as `@org/somethingA` and `@org/somethingB`. I will modify the code.

nieyuyao: I read the code and found that it is unlikely to directly determine the appropriate short form based on module name.

sheremet-va: Yeah, let's do that!

sheremet-va (approved):

PR #1434, opened 2022-06-05
feat(vite-node): respect ssr field, allow inlining everything

sheremet-va:
Vite allows `true` for `noExternal`, so I added support to us

I am adding this logic to `ViteNode`, because it can be used outside of Vitest, and I think it can help Histoire in the future.

### DIFF
--- a/packages/vite-node/src/server.ts
+++ b/packages/vite-node/src/server.ts
@@ -37,7 +37,7 @@ export class ViteNodeServer {
       // }
 
       if (ssrOptions.noExternal === true) {
-        options.deps.inline = true
+        options.deps.inline ??= true
       }
       else if (options.deps.inline !== true) {
         options.deps.inline ??= []
--- a/packages/vite-node/src/server.ts
+++ b/packages/vite-node/src/server.ts
@@ -3,7 +3,7 @@ import type { TransformResult, ViteDevServer } from 'vite'
 import createDebug from 'debug'
 import type { FetchResult, RawSourceMap, ViteNodeResolveId, ViteNodeServerOptions } from './types'
 import { shouldExternalize } from './externalize'
-import { toFilePath, withInlineSourcemap } from './utils'
+import { toArray, toFilePath, withInlineSourcemap } from './utils'
 
 export * from './externalize'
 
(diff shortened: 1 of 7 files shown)


packages/vite-node/src/server.ts
```diff
@@ -24,7 +24,27 @@ export class ViteNodeServer {
...
+      options.deps ??= {}
+
+      // we don't externalize ssr, because it has different semantics in Vite
+      // if (ssrOptions.external) {
+      //   options.deps.external ??= []
+      //   options.deps.external.push(...ssrOptions.external)
+      // }
+
+      if (ssrOptions.noExternal === true) {
+        options.deps.inline = true
```
antfu, line 40:
```suggestion
        options.deps.inline ??= true
```

Inline config should have higher priority

sheremet-va: I am not sure that it should. What are your arguments? I think, if user decided he wants to inline everything for `ssr`, we should do the same no matter what's inside `deps.inline`, because it was a decision made by the user for his actual application.

sheremet-va: If we do `??=` it means `noExternal: true` will never call this `if`, because we always add our `inline` in Vitest, for example.

antfu: `test.deps.inline` is more specific to test and to me it should always have the highest priority in Vitest if it set. Since Vitest might shares the option with the app, `ssr` might be used for the app but not the tests.

sheremet-va: Again, this is `vite-node`. We always append something to `deps.inline` in Vitest, so if we remove this condition here, it won't ever run. There is no reason to enable inlining in the app, and make it partial in Vitest (which is `imposible` right now, because `inline` is **always** defined by us). If you don't want something to be inline, use `deps.external`.

And again, you are saying that if user defines `test.deps.inline`, it should have higher priority, but it is impossible here, because when code goes here, we already defined/appended `inlined` in Vitest.

sheremet-va: > which is `imposible` right now, because `inline` is **always** defined by us

To resolve this, we might change this to be more conditional:

https://github.com/vitest-dev/vitest/blob/c79f9f91c2c1871dcf1e04013d9307dd6f30cdba/packages/vitest/src/node/config.ts#L96

- If defined `true`, stay with true
- If defined `noExternal`, but not defined `inline`, `inline` becomes `true`
- If defined `noExternal`, and defined `inline`, `inline` stays the same, and extended by us
- If not defined, defined with our inlines

antfu: I see now. Given this is rather a niche case, I am fine either way. I'll leave the decision to you.

antfu (approved):

packages/vite-node/src/server.ts
sheremet-va: > I see now. Given this is rather a niche case, I am fine either way. I'll leave the decision to you.

I'll change it to `??=`, but will alter how Vitest appends it's inlines

sheremet-va:
Ready to be merged with other breaking changes :P

antfu:
The tests hangs 👀

sheremet-va:
> The tests hangs 👀

Fixed 👀

PR #1442, opened 2022-06-06
chore: move examples out of workspace

### LINKED ISSUE #1440: [proposal] move `examples` out of pnpm workspace

The projects under `examples` folder are currently in the pnpm workspace, which causes slow installation. And it would become slower as the number of examples grows, 

Since the examples are hosted at [StackBlitz](https://github.com/vitest-dev/vitest/blob/main/examples/README.md), and don't involve any development or CI process, I thought it might be better to move these out of pnpm workspace.

Just tested it locally: The line number of `pnpm-lock.yaml` went from 20k+ to less than 10k

### Suggested solution

Move `examples` out of pnpm workspace.

I'll submit a PR if this can be agreed.

### Alternative

_No response_

### Additional context

_No response_

### Validations

pd4d10:
Closes #1440

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -97,461 +97,21 @@ importers:
       vue: ^3.2.36
       workbox-window: ^6.5.3
     dependencies:
-      '@vueuse/core': 8.5.0_vue@3.2.33
-      vue: 3.2.33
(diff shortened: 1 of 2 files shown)


antfu:
Close as https://github.com/vitest-dev/vitest/issues/1440#issuecomment-1147110714

PR #1462, opened 2022-06-10
fix: null inside test.each is not turned into an empty array

### LINKED ISSUE #1461: `test.each`'s behaviour deviates from jest's
### Describe the bug

When calling `test.each()` with an array containing `null`, e.g. `test.each([null])`, the value will be converted to an empty array, as seen below:

https://github.com/vitest-dev/vitest/blob/3199dc7d689a58033336bdd99075c0833d39aa81/packages/vitest/src/runtime/suite.ts#L195-L202

https://github.com/vitest-dev/vitest/blob/3199dc7d689a58033336bdd99075c0833d39aa81/packages/vitest/src/utils/base.ts#L89-L91

This differs from Jest's behaviour, which converts `null` into `[null]`.

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-78askn?file=test/suite.test.ts

### System Info

```shell
System:
  OS: Windows 10 10.0.22000
  CPU: (16) x64 AMD Ryzen 7 5800H with Radeon Graphics
  Memory: 6.92 GB / 15.86 GB
Binaries:
  Node: 18.2.0 - ~\AppData\Local\Volta\tools\image\node\18.2.0\node.EXE
  Yarn: 3.2.1 - ~\AppData\Local\Volta\tools\image\yarn\1.22.18\bin\yarn.CMD
  npm: 8.9.0 - ~\AppData\Local\Volta\tools\image\node\18.2.0\npm.CMD
Browsers:
  Edge: Spartan (44.22000.120.0), Chromium (101.0.1210.53)
npmPackages:
  vitest: ^0.14.2 => 0.14.2
```

### Used Package Manager

yarn

### Validations

sheremet-va:
Fixes #1461

### DIFF
--- a/packages/vitest/src/runtime/suite.ts
+++ b/packages/vitest/src/runtime/suite.ts
@@ -1,6 +1,6 @@
 import { format } from 'util'
 import type { File, RunMode, Suite, SuiteAPI, SuiteCollector, SuiteFactory, SuiteHooks, Task, Test, TestAPI, TestFunction } from '../types'
-import { isObject, noop, toArray } from '../utils'
+import { isObject, noop } from '../utils'
 import { createChainable } from './chain'
 import { collectTask, collectorContext, createTestContext, runWithSuite, withTimeout } from './context'
 import { getHooks, setFn, setHooks } from './map'
@@ -167,7 +167,7 @@ function createSuite() {
   suite.each = <T>(cases: ReadonlyArray<T>) => {
     return (name: string, fn: (...args: T[]) => void) => {
       cases.forEach((i, idx) => {
-        const items = toArray(i) as any
+        const items = Array.isArray(i) ? i : [i]
         suite(formatTitle(name, items, idx), () => fn(...items))
       })
     }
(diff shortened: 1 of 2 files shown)


packages/vitest/src/runtime/suite.ts
```diff
@@ -167,7 +167,7 @@ function createSuite() {
   suite.each = <T>(cases: ReadonlyArray<T>) => {
     return (name: string, fn: (...args: T[]) => void) => {
       cases.forEach((i, idx) => {
-        const items = toArray(i) as any
+        const items = Array.isArray(i) ? i : [i]
```
patak-dev, line 170: Should `toArray()` be fixed to avoid transforming `null` into `[]` ?
Here: https://github.com/vitest-dev/vitest/blob/c23be109deb67f796bc31edb61d3acbb306fea49/packages/ui/client/composables/summary.ts#L50
Instead of `||` use `??`

sheremet-va: ?? will still transform it into an empty array. 
we have two implementations. the one you referenced has nothing to do with the PR. I think it is useful somewhere, but I am too afraid to remove it, but here it’s it definetly not needed.

patak-dev: Uh, sorry for the `??` noice. Still is strange that a `toArray` will convert `null` to `[]`, but you are right, it may be better to be explicit here.

kyranet: If a non-ternary alternative is desired, `[].concat(i)` may work.
image

I'm a bit late since the PR just merged, but can still be done in a follow-up PR 😅

PR #1467, opened 2022-06-12
feat: add `css` config option, don't process CSS by default

sheremet-va:


packages/vitest/src/types/config.ts
```diff
@@ -344,6 +344,15 @@ export interface InlineConfig {
...
   onConsoleLog?: (log: string, type: 'stdout' | 'stderr') => false | void
+
+  /**
+   * Should css be processed.
+   *
+   * If regexp is provided, it will transform matched files.
+   *
+   * @default [/\.module\./]
+   */
+  processCss?: boolean | RegExp[]
```
patak-dev, line 355: This looks like a great DX improvement. About naming, given that there is already a `css` option in Vite, maybe this new option could be aligned with it:
```
  css: {
    include: RegExp | RegExp[], // @default [/\.module\./]
    exclude: RegExp | RegExp[]
  }
```
Removing the `process` in the name could be helpful if there are other CSS options added later. Or maybe `css: boolean | { include, exclude }` if you still want users to use `css: true`

sheremet-va: I think if we want to add more features to css, they should be added to Vite’s options :)

About the object - I think yes, we can align it with other similar options.

packages/vitest/src/node/plugins/cssEnabler.ts
```diff
@@ -9,10 +10,14 @@ export function CSSEnablerPlugin(ctx: Vitest): VitePlugin {
...
-      return processCss
-    return processCss.some(re => re.test(id))
+    const { css } = ctx.config
+    if (typeof css === 'boolean')
+      return css
+    if (toArray(css.exclude).some(re => re.test(id)))
+      return false
+    if (toArray(css.include).some(re => re.test(id)))
+      return true
+    return true
```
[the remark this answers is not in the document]
sheremet-va: I know there is a utility like this, and I don't know why we don't use it. When doing this PR, I also haven't found it, so I didn't install it.

TBH I don't see any real value in using it here. If we want to install utils, we might as well change it's usages everywhere.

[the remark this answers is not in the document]
sheremet-va: Also, it's won't really work here. I need to return `true`, if the file is not inside `include` and `exclude`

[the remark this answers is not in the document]
patak-dev: Yes, probably better for another PR to do this for all include/exclude like options. The utils are already used in Vite core so it isn't adding anything new. I actually wonder if we shouldn't expose it from Vite as a util so more people in the ecosystem will use it. I think I'll propose that in a PR in vite.

patak-dev (approved):

packages/vitest/src/node/plugins/cssEnabler.ts
```diff
@@ -0,0 +1,31 @@
+import type { Plugin as VitePlugin } from 'vite'
+import { toArray } from '../../utils'
+import type { Vitest } from '../core'
+
+const isCss = (id: string) => {
+  return /\.(css|styl|less|sass|scss)$/.test(id)
```
antfu, line 6: https://github.com/vitejs/vite/blob/b3d965277741f03890d2dc6881a2c8605c6b5ac3/packages/vite/src/node/plugins/css.ts#L101-L102

packages/vitest/src/node/plugins/cssEnabler.ts
```diff
@@ -0,0 +1,31 @@
+import type { Plugin as VitePlugin } from 'vite'
+import { toArray } from '../../utils'
+import type { Vitest } from '../core'
+
+const isCss = (id: string) => {
```
antfu, line 5:
```suggestion
const isCSS = (id: string) => {
```

I personally prefer `css` and `CSS` whenever possible

packages/vitest/src/types/config.ts
```diff
@@ -344,6 +344,18 @@ export interface InlineConfig {
    * Return `false` to ignore the log.
    */
   onConsoleLog?: (log: string, type: 'stdout' | 'stderr') => false | void
+
+  /**
+   * Should css be processed. If not, `css` files will be empty.
```
antfu, line 349:
```suggestion
   * If CSS files should be processed.
   * When excluded, the CSS files will be replaced with empty strings to bypass the subsequent processing.
```

packages/vitest/src/types/config.ts
```diff
@@ -344,6 +344,18 @@ export interface InlineConfig {
...
+   *
+   * If regexp is provided, it will transform matched files.
```
antfu, line 351:
```suggestion
```

antfu (approved):

packages/vitest/src/node/plugins/cssEnabler.ts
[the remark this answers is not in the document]
patak-dev: For reference: 
- https://github.com/vitejs/vite/pull/8562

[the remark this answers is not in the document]
antfu: for deps' include / exclude is different from `createFilter` semantically (as `['vue']` stands for the `vue` package instead of glob). For `css` I think sure we could use it.

docs/config/index.md
```diff
@@ -533,3 +533,25 @@ Vitest will not fail, if no tests will be found.
 - **Default**: `false`
 
 Show heap usage after each test. Useful for debugging memory leaks.
+
+### css
+
+- **Type**: `boolen | { include?, exclude? }`
```
bgoscinski, line 539:
```suggestion
- **Type**: `boolean | { include?, exclude? }`
```

PR #1474, opened 2022-06-13
fix: add @types/node to dependencies

### LINKED ISSUE #1076: Build fails because of missing type definitions
### Describe the bug

Whenever I add a test to the project and try to run `npm run build`, I get type definition errors. Errors are related to @types/node and @types/ws. I know I could just install these within my own devDependencies, but I think this should not be the case?

image

image

### Reproduction

This is a fresh install of Vite+Vitest. The only change by me here is adding a test. Try running `npm run build` here:
https://stackblitz.com/edit/vitejs-vite-kmvhms

### System Info

```shell
-
```

### Used Package Manager

npm

### Validations

sheremet-va:
Fixes #1076

See repro: https://github.com/sheremet-va/vitest-types-bug

### DIFF
--- a/packages/vitest/package.json
+++ b/packages/vitest/package.json
@@ -84,7 +84,7 @@
   "dependencies": {
     "@types/chai": "^4.3.1",
     "@types/chai-subset": "^1.3.3",
-    "@types/node": "^17.0.35",
+    "@types/node": "*",
     "chai": "^4.3.6",
     "debug": "^4.3.4",
     "local-pkg": "^0.4.1",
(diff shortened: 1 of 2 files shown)


antfu:
But I am not very should about the version range here. Ideally, we should respect users' `@types/node` if ever specified. So probably `"@types/node": "*"` in `dependencies`

sheremet-va:
> But I am not very sure about the version range here. Ideally, we should respect users' `@types/node` if ever specified. So probably `"@types/node": "*"` in `dependencies`

Done :D

antfu (approved):

PR #1475, opened 2022-06-13
fix: spyOn default export of module

### LINKED ISSUE #855: Cannot spyOn default export of module
### Describe the bug

In Jest a common pattern for me was to `spyOn` the `default` export of a module. This no longer works in vitest. It throws the following error:
```
TypeError: Cannot redefine property: default
 ❯ eval test/suite.test.ts:5:16
      3| import b from '../src/b';
      4| 
      5| const aSpy = vi.spyOn(a, 'default');
       |                ^
      6| 
      7| describe('suite name', () => {
```

I get the feeling this is a limitation of ESM, a work around seems to be:
```
vi.mock('path/to/module', async () => {
  const actual = await requireActual('path/to/module');
  const mocked = {};
  for (const key of Object.keys(actual)) {
    if (typeof actual[key] === 'function') {
      mocked[key] = vi.fn(actual[key]);
    } else {
      mocked[key] = actual[key];
    }
  }
  return mocked;
})

import module from 'path/to/module';

const moduleSpy = vi.mocked(module);
```
However, this is a lot of code to include for every spy, so perhaps there could be an API like `vi.spyOnModule` that would automate this, so I could just do:

```
const moduleSpy = vi.spyOnModule('/path/to/module');
```

### Reproduction

I created a reproduction here: https://stackblitz.com/edit/vitest-dev-vitest-6igo9o?file=test/suite.test.ts

> Side note: the latest release (0.5.5) seems to have a regression where it does not show the error message, this seems to be because it checks for calls to `it`/`test` to detect if the file has tests. However, because `vi.spyOn` is failing before those tests are defined, the error message gets suppressed. Hence why I downgraded to `0.5.4` in the repro.

### System Info

```shell
System:
    OS: macOS 11.2.3
    Memory: 39.41 MB / 16.00 GB
    Shell: 5.8 - /usr/local/bin/zsh
  Binaries:
    Node: 16.13.2 - ~/.nvm/versions/node/v16.13.2/bin/node
    npm: 8.1.2 - ~/.nvm/versions/node/v16.13.2/bin/npm
  Browsers:
    Chrome: 98.0.4758.109
    Edge: 98.0.1108.62
    Firefox: 94.0.1
    Safari: 14.0.3
  npmPackages:
    @vitejs/plugin-react: ^1.2.0 => 1.2.0 
    vite: 2.8.3 => 2.8.3 
    vitest: 0.3.5 => 0.3.5
```

### Used Package Manager

yarn

### Validations

konpeki622:
Fixes #855 .

I changed the result of source-map transformation, which made the `default` property configurable so that it can create spy on default exports.

### DIFF
--- a/packages/vite-node/src/client.ts
+++ b/packages/vite-node/src/client.ts
@@ -205,6 +205,12 @@ export class ViteNodeRunner {
     if (transformed[0] === '#')
       transformed = transformed.replace(/^\#\!.*/, s => ' '.repeat(s.length))
 
+    // #855 make `default` property configurable
+    if (/Object\.defineProperty\(\_\_vite\_ssr\_exports\_\_\,\s\"default\"\,\s\{.*value/.test(transformed)) {
+      const exportsPropertyDefinition = 'Object.defineProperty(__vite_ssr_exports__, "default", { configurable: true, writable: true, enumerable: true, value'
+      transformed = transformed.replace(/Object\.defineProperty\(\_\_vite\_ssr\_exports\_\_\,\s\"default\"\,\s\{.*value/, exportsPropertyDefinition)
+    }
+
     // add 'use strict' since ESM enables it by default
     const fn = vm.runInThisContext(`'use strict';async (${Object.keys(context).join(',')})=>{{${transformed}\n}}`, {
       filename: fsPath,
--- a/test/core/src/default-import.ts
+++ b/test/core/src/default-import.ts
@@ -1,0 +1,5 @@
+import c from './module-esm'
+
+export function defaultImport() {
+  return c()
+}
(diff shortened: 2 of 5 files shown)


packages/vite-node/src/client.ts
```diff
@@ -205,6 +205,12 @@ export class ViteNodeRunner {
     if (transformed[0] === '#')
       transformed = transformed.replace(/^\#\!.*/, s => ' '.repeat(s.length))
 
+    // #855 make `default` property configurable
```
sheremet-va, line 208: This will break source maps. This transformations should be done in either plugin or by Vite itself.

sheremet-va (changes requested):

antfu:
This should be better done in Vite. Please send PR upstream. Thanks.

konpeki622:
Thanks, I will try it.

PR #1477, opened 2022-06-13
feat: add --shard command

### LINKED ISSUE #1464: Shard option to distribute tests across multiple machines

Jest v18 introduced as [`--shard`](https://jestjs.io/docs/cli#--shard) option which is pretty awesome.

You can split your tests execution by providing a `<shardIndex>/<shardCount>` ratio.

It goes hand in hand with Github Action matrix strategy: https://remarkablemark.org/blog/2022/05/13/jest-shard-test-github-actions/#workflow

This is very useful when dealing with long tests such as integration tests.

### Suggested solution

I think Jest API is great for this command, it should just be copied.

### Alternative

_No response_

### Additional context

_No response_

### Validations

sheremet-va:
Closes #1464

### DIFF
--- a/packages/vitest/src/node/config.ts
+++ b/packages/vitest/src/node/config.ts
@@ -56,16 +56,6 @@ export function resolveApiConfig<Options extends ApiConfig & UserConfig>(
   return api
 }
 
-const configError = (error: string): never => {
-  console.warn(
-    c.yellow(
-      `${c.inverse(c.red(' VITEST '))} ${error}\n`,
-    ),
-  )
-
-  process.exit(1)
-}
-
 export function resolveConfig(
   options: UserConfig,
   viteConfig: ResolvedViteConfig,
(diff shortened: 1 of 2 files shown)


packages/vitest/src/node/config.ts
```diff
@@ -56,6 +56,16 @@ export function resolveApiConfig<Options extends ApiConfig & UserConfig>(
...
 }
 
+const configError = (error: string): never => {
+  console.warn(
+    c.yellow(
+      `${c.inverse(c.red(' VITEST '))} ${error}\n`,
+    ),
+  )
+
+  process.exit(1)
```
antfu, line 66: Better to just throw the error, as `resolveConfig` can be used dynamically, and the caller might want to catch the error.

antfu: We can catch it and process.exit in the caller, instead of having side-effect for `resolveConfig`

sheremet-va: I've added throwing, but I still do `process.exit` when catching the error

PR #1478, opened 2022-06-13
docs: restructure docs api into three sections

sheremet-va:
I think splitting API into three separate sections makes it easier to access information.

But I can't come up with text for `/api/index`, so I just deleted it :S


patak-dev:
We are having the same issue in the Vite docs. The problem is that when you separate the API or Config section then the users can no longer see the full surface in a single page (to search with Cmd+f for example, even if there is algolia available). For Vite, we decided that the index would be the same as Vue: https://vuejs.org/api/

image

This is great because it gives users a complete list of the API + it gives you something to put in the index.

About the partition, should the Mock instance methods and properties also be split?

sheremet-va:
Not sure if this solution will work nicely with the count of expect's API methods 🤔

patak-dev:
Why? Check out the vue page, there are tons of methods there and look ok to me 🤔

sheremet-va:
> Why? Check out the vue page, there are tons of methods there and look ok to me 🤔

They are nicecly divided, but currently `expect` methods are just a list of 55 methods 👀

patak-dev:
Maybe the expect methods could be categorized a bit, at least for this API reference. For example:

not
toBe
toBeCloseTo
toBeDefined
toBeUndefined
toBeTruthy
toBeFalsy
toBeNull
toBeNaN
toBeTypeOf
toBeInstanceOf

toBeGreaterThan
toBeGreaterThanOrEqual
toBeLessThan
toBeLessThanOrEqual
toEqual
toStrictEqual

toContain
toContainEqual
toHaveLength
toHaveProperty

toMatch
toMatchObject
toMatchSnapshot
toMatchInlineSnapshot

toThrowError
toThrowErrorMatchingSnapshot
toThrowErrorMatchingInlineSnapshot

toHaveBeenCalled
toHaveBeenCalledTimes
toHaveBeenCalledWith
toHaveBeenLastCalledWith
toHaveBeenNthCalledWith
toHaveReturned
toHaveReturnedTimes
toHaveReturnedWith
toHaveLastReturnedWith
toHaveNthReturnedWith

toSatisfy
resolves
rejects
expect.assertions
expect.hasAssertions
expect.addSnapshotSerializer
expect.extend

antfu:
The `/api/index` should be a simple index of those three sections. Other than that I am good for it

antfu:
Close due to conflicts but I agree to do so. When we want to do this, we could directly push to main branch to avoid conflicts.

PR #1481, opened 2022-06-14
fix: add throat when running spec files

sheremet-va:
Fixes `VITETS_POOL_ID` logic.

I noticed no performance differences with this.

I _think_ this might also help with segfalts and Node errors? I have a feeling that running them all might cause it, but I don't have anything to back up that claim.

### DIFF
--- a/packages/vitest/LICENSE.md
+++ b/packages/vitest/LICENSE.md
@@ -1892,6 +1892,33 @@ Repository: sindresorhus/strip-final-newline
 
 ---------------------------------------
 
+## throat
+License: MIT
+By: ForbesLindesay
+Repository: https://github.com/ForbesLindesay/throat.git
+
+> Copyright (c) 2013 Forbes Lindesay
+> 
+> Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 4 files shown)


antfu:
https://www.npmtrends.com/p-limit-vs-p-throttle-vs-promise-throttle-vs-throat

Maybe p-limit?

sheremet-va:
> https://www.npmtrends.com/p-limit-vs-p-throttle-vs-promise-throttle-vs-throat
> 
> Maybe p-limit?

Sure. Replaced it. Also added comment to PR, if you didn't notice.

sheremet-va:
Closed in favor of #1482

antfu:
I guess tinypool already has the queueing mechanism, probably this might be redundant. About the Segfault, I guess it's about some deep bug of Node, where even if this makes it less likely to appear (I doubt tho), it doesn't really fix the root cause. 🤔

sheremet-va:
> It has, but we can't hook up to it.

Maybe we can add this functionality to `tinypool`?

> Doesn't hurt to try.

`react-enzym` failed once in the new PR, but fixed in the second one. I fear this change might introduce flaky tests now :P

Issue #1469, opened 2022-06-12
process.env.VITEST_WORKER_ID can be larger than maxThreads


flunderpero:
### Describe the bug

We are migrating our Jest test code over to vitest. We use `process.env.JEST_WORKER_ID` to allocate a separate database for each worker/thread. With Jest that `JEST_WORKER_ID` is always between 1 and the maximum number of workers.

In vitest `process.env.VITEST_WORKER_ID` seems to have another meaning and is ever increasing. The [documentation](https://vitest.dev/config/#setupfiles) states that `VITEST_WORKER_ID` can be used to distinguish between threads. In my opinion this is _not_ what it actually does.

### Reproduction

Run any test-sutie and see that `VITEST_WORKER_ID` increases. In watch-mode it increases with each run.

### System Info

```shell
System:
    OS: macOS 12.4
    CPU: (10) arm64 Apple M1 Max
    Memory: 1.36 GB / 64.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.13.1 - ~/.nvm/versions/node/v16.13.1/bin/node
    Yarn: 3.1.0 - ~/.nvm/versions/node/v16.13.1/bin/yarn
    npm: 8.10.0 - ~/.nvm/versions/node/v16.13.1/bin/npm
  Browsers:
    Chrome: 102.0.5005.115
    Firefox: 100.0.2
    Safari: 15.5
  npmPackages:
    vitest: 0.14.2 => 0.14.2
```

### Used Package Manager

yarn

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

flunderpero:
[...]

Issue #1212, opened 2022-04-29
max concurrency setting for test.concurrent


yamadashy:
### Clear and concise description of the problem

If you run a large number of tests with bottlenecks, such as database access, in `test.concurrent`, each test will affect each other and take time.

Also, in some cases, such as CI, you may want to tweak performance so that it does not become too heavy.

### Suggested solution

How about adding an option `maxConcurrency` to config to limit the number of concurrent tests to be run by `test.concurrent`?
```js
export default defineConfig({
  test: {
    maxConcurrency: 5,
  },
})
```

The implementation is to queue the tests for `test.concurrent` once, and run them without exceeding `maxConcurrency`.

Jest addresses these issues with an option `maxConcurrency` that allows you to tweak concurrent.
https://jestjs.io/ja/docs/configuration#maxconcurrency-number
> A number limiting the number of tests that are allowed to run at the same time when using test.concurrent. Any test above this limit will be queued and executed once a slot is released.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #1482, opened 2022-06-14
fix: add throat when running spec files

sheremet-va:
Fixes `VITETS_POOL_ID` logic.

I noticed no performance differences with this.

I _think_ this might also help with segfalts and Node errors? I have a feeling that running them all might cause it, but I don't have anything to back up that claim.

### DIFF
--- a/packages/vitest/package.json
+++ b/packages/vitest/package.json
@@ -119,6 +119,7 @@
     "micromatch": "^4.0.5",
     "mlly": "^0.5.2",
     "natural-compare": "^1.4.0",
+    "p-limit": "^4.0.0",
     "pathe": "^0.2.0",
     "picocolors": "^1.0.0",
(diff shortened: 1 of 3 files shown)


sheremet-va:
Also will reuse `p-limit` for #1212 in the PR after this

sheremet-va:
@Aslemammad maybe we can add a possibility to hook up into the internal tinypool queue? We need to generate `VITEST_POOL_ID` based on the running workers: https://github.com/vitest-dev/vitest/issues/1469#issuecomment-1154355502 (only not used IDs should be used)

Aslemammad:
@sheremet-va It's possible! I can work on it after my exams next week! Is it okay?

Aslemammad:
@sheremet-va Sure, thank you so much!

sheremet-va:
Continued in #1531

PR #1483, opened 2022-06-14
feat: add `maxConcurrency` option

### LINKED ISSUE #1212: max concurrency setting for test.concurrent

sheremet-va:
Closes #1212

### DIFF
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -533,3 +533,12 @@ Vitest will not fail, if no tests will be found.
 - **Default**: `false`
 
 Show heap usage after each test. Useful for debugging memory leaks.
+
+### maxConcurrency
+
+- **Type**: `number`
+- **Default**: `5`
+
+A number of tests that are allowed to run at the same time marked with `test.concurrent`.
+
+Test above this limit will be queued to run when available slot appears.
--- a/packages/vitest/package.json
+++ b/packages/vitest/package.json
@@ -119,6 +119,7 @@
     "micromatch": "^4.0.5",
     "mlly": "^0.5.2",
     "natural-compare": "^1.4.0",
+    "p-limit": "^4.0.0",
     "pathe": "^0.2.0",
     "picocolors": "^1.0.0",
     "pkg-types": "^0.3.2",
--- a/packages/vitest/src/defaults.ts
+++ b/packages/vitest/src/defaults.ts
@@ -72,6 +72,7 @@ const config = {
   open: true,
   coverage: coverageConfigDefaults,
   fakeTimers: fakeTimersDefaults,
+  maxConcurrency: 5,
 }
 
 export const configDefaults: Required<Pick<UserConfig, keyof typeof config>> = Object.freeze(config)
--- a/packages/vitest/src/runtime/run.ts
+++ b/packages/vitest/src/runtime/run.ts
@@ -1,3 +1,4 @@
+import limit from 'p-limit'
 import type { File, HookCleanupCallback, HookListener, ResolvedConfig, Suite, SuiteHooks, Task, TaskResult, TaskState, Test } from '../types'
 import { vi } from '../integrations/vi'
 import { getSnapshotClient } from '../integrations/snapshot/chai'
@@ -195,6 +196,8 @@ export async function runSuite(suite: Suite) {
 
   updateTask(suite)
 
+  const workerState = getWorkerState()
+
   if (suite.mode === 'skip') {
     suite.result.state = 'skip'
   }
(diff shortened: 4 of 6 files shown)


antfu:
It's better to release this with the next minor bump.

sheremet-va:
> It's better to release this with the next minor bump.

I am waiting for you to review `--shard` before releasing a new version 😄

gajus:
Is there a way to set this globally?

```
Usage:
  $ vitest run [...filters]

Options:
  -r, --root <path>                      Root path
  -c, --config <path>                    Path to config file
  -u, --update                           Update snapshot
  -w, --watch                            Enable watch mode
  -t, --testNamePattern <pattern>        Run tests with full names matching the specified regexp pattern
  --dir <path>                           Base directory to scan for the test files
  --ui                                   Enable UI
  --open                                 Open UI automatically (default: !process.env.CI))
  --api [api]                            Serve API, available options: --api.port <port>, --api.host [host] and --api.strictPort
  --threads                              Enabled threads (default: true)
  --single-thread                        Run tests inside a single thread, requires --threads (default: false)
  --experimental-vm-threads              Run tests in a worker pool using VM isolation (default: false)
  --experimental-vm-worker-memory-limit  Set the maximum allowed memory for a worker. When reached, a new worker will be created instead
  --silent                               Silent console output from tests
  --hideSkippedTests                     Hide logs for skipped tests
  --isolate                              Isolate environment for each test file (default: true)
  --reporter <name>                      Specify reporters
  --outputFile <filename/-s>             Write test results to a file when supporter reporter is also specified, use cac's dot notation for individual outputs of multiple reporters
  --coverage                             Enable coverage report
  --run                                  Disable watch mode
  --mode <name>                          Override Vite mode (default: test)
  --globals                              Inject apis globally
  --dom                                  Mock browser api with happy-dom
  --browser [options]                    Run tests in the browser (default: false)
  --environment <env>                    Specify runner environment, if not running in the browser (default: node)
  --passWithNoTests                      Pass when no tests found
  --logHeapUsage                         Show the size of heap for each test
  --allowOnly                            Allow tests and suites that are marked as only (default: !process.env.CI)
  --dangerouslyIgnoreUnhandledErrors     Ignore any unhandled errors that occur
  --shard <shard>                        Test suite shard to execute in a format of <index>/<count>
  --changed [since]                      Run tests that are affected by the changed files (default: false)
  --sequence <options>                   Define in what order to run tests (use --sequence.shuffle to run tests in random order, use --sequence.concurrent to run tests in parallel)
  --segfaultRetry <times>                Return tests on segment fault (default: 0) (default: 0)
  --no-color                             Removes colors from the console output (default: true)
  --inspect                              Enable Node.js inspector
  --inspect-brk                          Enable Node.js inspector with break
  --test-timeout <time>                  Default timeout of a test in milliseconds (default: 5000)
  --bail <number>                        Stop test execution when given number of tests have failed (default: 0)
  --retry <times>                        Retry the test specific number of times if it fails (default: 0)
  --diff <path>                          Path to a diff config that will be used to generate diff interface
```

We have a problem where vitest hogs too much CPU and ends up suffocating database operations (which is hosted on the same machine). I am not clear on how to tell Vitest to at most parallelize 80% CPU worth of threads.

gajus:
Found it:

```diff
+ import os from 'node:os';
import path from 'node:path';
import { defineConfig } from 'vitest/config';

export default defineConfig({
  resolve: {
    alias: {
      '@': path.resolve(__dirname, './src'),
    },
  },
  test: {
    deps: {
      interopDefault: true,
    },
+    maxThreads: Math.ceil(os.cpus().length * 0.5),
    mockReset: true,
    outputFile: './src/__generated__/vitest/results.xml',
    reporters: ['verbose', 'junit'],
    testTimeout: 60_000,
  },
});
```

gajus:
On a related note, I am not sure if defaulting to 100% of the CPUs is a great idea. It diminishes test stability.

PR #1510, opened 2022-06-20
fix: correctly generate inline snapshots inside js files

### LINKED ISSUE #1376: Inline snapshots not being created
### Describe the bug

Under some specific settings unkown to me yet (dependencies, maybe?), inline snapshots do not work. To be more specific:
- They are not written when there is only one `toMatchInlineSnapshot` assertion
- They are written to the wrong assertion when there are more than one

This happened to my first project with vite/vitest and for a while assumed that they just weren't working but trying out a new project from scratch didn't have this issue. I believe this is somewhat of an edge case, and even though I narrowed my dependencies to the bare minimum in a StackBlitz project, I cannot seem to find the root cause.

Thanks!

### Reproduction

1. Go to https://stackblitz.com/edit/vitejs-vite-vxrdhg?file=tests/unit/utils/dates.spec.js&view=editor
2. Run `npm run test`
3. See how the second snapshot gets created with the value of the first snapshot.

More variants:
- Remove the second test block, see how vitest says the snapshot was written but it is not

### System Info

```shell
I don't think this is relevant given there is a StackBlitz repro, but still

  System:
    OS: macOS 11.4
    CPU: (16) x64 Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
    Memory: 963.97 MB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
  Browsers:
    Brave Browser: 97.1.34.81
    Chrome: 101.0.4951.64
    Firefox: 100.0.2
    Safari: 14.1.1
  npmPackages:
    vite: ^2.9.9 => 2.9.9
    vitest: ^0.12.9 => 0.12.9
```

### Used Package Manager

npm

### Validations

sheremet-va:
Fixes #1376


packages/vitest/src/integrations/snapshot/port/state.ts
```diff
@@ -109,12 +109,16 @@ export default class SnapshotState {
...
-      const stack = this._getInlineSnapshotStack(stacks)
+      const stack = this._inferInlineSnapshotStack(stacks)
       if (!stack) {
         throw new Error(
           `Vitest: Couldn't infer stack frame for inline snapshot.\n${JSON.stringify(stacks)}`,
         )
       }
+      // removing 1 column, because source map points to the wrong
+      // location for js files, but `column-1` points to the same in both js/ts
+      // #vite/8657
```
antfu, line 120: Better with a full link

antfu (approved):

PR #1516, opened 2022-06-20
feat(vite-node): vite hmr support

poyoho:
fix: #1028
fork vite client hmr handle in vite-node


sheremet-va (commented):
Should we export a public interface for it, so people can reuse it? I mean, define an entry point in `rollup.config` and in `exports` in package.json

packages/vite-node/src/client.ts
```diff
@@ -187,7 +187,10 @@ export class ViteNodeRunner {
       __vite_ssr_dynamic_import__: request,
       __vite_ssr_exports__: exports,
       __vite_ssr_exportAll__: (obj: any) => exportAll(exports, obj),
-      __vite_ssr_import_meta__: { url },
+      __vite_ssr_import_meta__: {
+        url,
+        hot: this.options.createHotContext?.(this, `/@fs/${fsPath}`),
```
antfu, line 192: Can this be a getter? I am a bit worried about the performance impact on it.

poyoho: createHotContext when transformd code had `__vite_ssr_import_meta__.hot`?

antfu (approved):

sheremet-va (approved):

PR #1525, opened 2022-06-22
feat!: rename mock types to match jest's

sheremet-va:
Current types were names without much consideration about compatibility with Jest, or any actual thought behind it 👀 

I think we might as well align them with Jest. This is a breaking change

### DIFF
--- a/packages/vitest/src/integrations/spy.ts
+++ b/packages/vitest/src/integrations/spy.ts
@@ -16,7 +16,7 @@ interface MockResultThrow {
 
 type MockResult<T> = MockResultReturn<T> | MockResultThrow | MockResultIncomplete
 
-export interface SpyContext<TArgs, TReturns> {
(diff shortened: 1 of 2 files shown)


antfu:
Maybe we could expose alias like:

```ts
type MockContext = SpyContext
```

antfu:
We have `SpyInstanceFn` renamed to `Mock`, but kept `SpyInstance`, which feels a bit inconsistent to me. Should we also rename it, or we just have two set of naming logic.

sheremet-va:
> We have `SpyInstanceFn` renamed to `Mock`, but kept `SpyInstance`, which feels a bit inconsistent to me. Should we also rename it, or we just have two set of naming logic.

jest has SpyInstance 😄 I've added an alias for MockInstance, since jest also has that one.

antfu (approved):

PR #1541, opened 2022-06-24
feat: add caching to run failed and longer tests first

### LINKED ISSUE #1532: Run slowest tests first or provide a way to provide the order of tests

Jest runs the slowest tests first. That decreases overall runtime of our test-suite from ~90s to ~70s. In vitest the same suite always takes ~110s mainly due to the increased isolation (which is _the_ greatest feature of vitest imo) but also because it so happens that the slow test files are started near the end of the run.

### Suggested solution

Either implement Jest's approach and run the slowest test suites first or allow the user to specify the order in which tests are started. I would prefer Jest's approach.

### Alternative

_No response_

### Additional context

_No response_

### Validations

sheremet-va:
Closes #1532

TODO
- [x] Tests
- [x] Clear cache command
- [x] Sort by file size, if info not provided


packages/vitest/src/types/config.ts
```diff
@@ -427,4 +435,8 @@ export interface ResolvedConfig extends Omit<Required<UserConfig>, 'config' | 'f
     index: number
     count: number
   }
+
+  cache: {
+    path: string
```
antfu, line 440:
```suggestion
    dir: string
```

antfu:
Looks good for the overall direction.

sheremet-va:
@antfu I did some changes since your last review:

- add sequelizer in preparation for #1570
- cache file stats
- new command "clearCache"

docs/guide/cli.md
```diff
@@ -36,6 +36,10 @@ Useful to run with [`lint-staged`](https://github.com/okonet/lint-staged) or wit
 vitest related /src/index.ts /src/hello-world.js
 ```
 
+### `vitest clearCache`
```
antfu, line 39:
```suggestion
### `vitest purge`
```

I am not a fan of pascal case in CLI. Maybe `purge`, WDYT?

sheremet-va: Why `purge` tho? Don't see any connection to cache. We can do `clear-cache`, I guess.

antfu: Just want to find a single word to represent it. As for reference, here is what package managers are doing:

```ts
npm cache clean --force
yarn cache clean
pnpm store prune
```

And for Vite: https://vitejs.dev/guide/dep-pre-bundling.html#file-system-cache

/cc @patak-dev in case we are also going to add a clear cache command for Vite, I think we better align the commands

sheremet-va: We can do `vitect clean`, and pass `cache` as an argument. In future we might add option to clear only some specific caches.

sheremet-va: I've updated to proposed solution. What do you think, @antfu, @patak-dev?

sheremet-va: I don't like solution with `vitest cache <action>`, because I think we shouldn't all anything done to cache except clearing.

antfu: `clean` looks good to me

patak-dev: This is something we should also clean in Vite, but whenever we discuss in a meeting, 10 options appear and we get stuck deciding 👀
For vite, we now ended up moving force to optimizeDeps.force, hinting that there won't be a global cache object but each feature may have its own? 

I think that the global cache config is good for Vitest though (and probably also we should move cacheDir to cache.dir like you have in the PR)

I thought while reading the PR that you could piggyback on vite here and use `${cacheDir}/vitest`... but probably a good idea to keep them separate as there is a log of people used to do rm -rf ${cacheDir}

About the command, vitest clean sounds good, I agree with you that we may now want to follow the yarn cache clean here
I would prefer a flag instead of a command though, as it is very useful to run something while clearing the cache (like --force):
```
vitest run --coverage --clean
```
And maybe by default if no params are passed it means to clean everything?

sheremet-va: > I would prefer a flag instead of a command though, as it is very useful to run something while clearing the cache (like --force):

You have a point! Maybe we can do something like

```
vitest run --no-cache
```

I don't like `--force`, because Vite already uses it and it may be confusing. And `--clean` or `--clean` is not expressive enough for me 👀

sheremet-va: Although the fact that `--no-cache` removes cache is also not apparent 🤔

sheremet-va: I guess we can just do `vitest run --clear-cache`

patak-dev: I actually wanted `--no-cache` for Vite too but it isn't clear if you start partitioning the cache. I think that `vitest run --clean` is clear enough, it is something you use a lot so it is better if it is a single word

sheremet-va: > it is something you use a lot so it is better if it is a single word

Hardly disagree with using a lot - I don't know why you would even use it in Vitest, only to clear malformed cache maybe.

Also single words only _look_ good, but confuse a lot. What `--clean` means? What happens when you call it?

patak-dev: I use `--force` a lot because ya, I'm debugging Vite cache itself 👀 
So I agree with you that normally isn't something that final users will be reaching a lot. I'm fine with `--clearCache` (maybe good to check how it will look later for `--clearCache="parts of it"`)

sheremet-va: > maybe good to check how it will look later

It will look something like this:
```
vitest run --clearCache results
```
We use cac's notation

sheremet-va: @antfu what do you think?

sheremet-va: I will remove command for now, since I want this feature to be released today. Command will be in another PR.

PR #1545, opened 2022-06-25
fix: cache path of modules imported in setup file

### LINKED ISSUE #1450: `vi.mock` ignored if module is imported in setup file
### Describe the bug

Registration of `vi.mock` on a module will not work if that module is imported in a file added to the `setupFiles` array in `vite.config.ts`.

### Use case

We use Vue 3 with global component registration (this is needed because of the way our CMS works). To accomplish this in our tests, we register these components in `vitest.setup.ts` to the `@vue/test-utils` config as the example below.

```js
import { config } from '@vue/test-utils';

config.global.components[name] = component;
```

This causes the mocks set on the imports of these components to be ignored. Although this is our use case, this issue is unrelated to Vue. See the minimal reproduction for more details.

### Reproduction

Here is a minimal repo: https://stackblitz.com/edit/vitest-dev-vitest-etmse4?file=vitest.setup.ts

In `test/foo.test.ts` a mock is used. This fails, unless I remove the only line in `vitest.setup.ts`.

### System Info

```shell
System:
    OS: Windows 10 10.0.17763
    CPU: (4) x64 Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz
    Memory: 4.85 GB / 16.00 GB
  Binaries:
    Node: 14.17.6 - C:\Program Files\nodejs\node.EXE
    npm: 6.14.15 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 102.0.5005.63
    Internet Explorer: 11.0.17763.2145
  npmPackages:
    vite: ^2.9.9 => 2.9.9
    vitest: 0.14.1 => 0.14.1
```

### Used Package Manager

npm

### Validations

konpeki622:
closes #1450


sheremet-va:
This doesn't seem right. `vi.mock` doesn't work because we already imported a file and it exists inside a cache. We need to remove it, if we are mocking it, that's all.

konpeki622:
> This doesn't seem right. `vi.mock` doesn't work because we already imported a file and it exists inside a cache. We need to remove it, if we are mocking it, that's all.

I mean that once a module was imported before mocking, its path should be cached. But now it cached two different paths from setup files and test files, though these two paths refered to one module, that means it will mock 'another' module. I made them to be the same absolute path.

sheremet-va:
> I mean that once a module was imported before mocking, its path should be cached. But now it cached two different paths from setup files and test files, though these two paths refered to one module, that means it will mock 'another' module. I made them be the same absolute path.

You are incorrect.

Their imported paths are the same. `vi.mock` mocks the correct module. It is not mocked in test, because in `request` it imports the cached one (`vite-node` doesn't know about mocks). This is why we should removed actual module from cache - it is no longer needed.

test/core/test/imports.test.ts
```diff
@@ -23,7 +23,7 @@ test('dynamic aliased import works', async () => {
   const timeoutPath = '@/timeout'
   const variableTimeoutMod = await import(timeoutPath)
 
-  expect(stringTimeoutMod).toBe(variableTimeoutMod)
+  expect(stringTimeoutMod).toStrictEqual(variableTimeoutMod)
```
sheremet-va, line 26: No, it should `BE` the same module - have the same reference

packages/vitest/src/runtime/setup.ts
```diff
@@ -167,7 +168,8 @@ export async function runSetupFiles(config: ResolvedConfig) {
   await Promise.all(
     files.map(async (file) => {
       getWorkerState().moduleCache.delete(file)
-      await import(file)
+      if (existsSync(file))
+        await import(file)
```
sheremet-va, line 172: We shouldn't silently skip this

sheremet-va (changes requested):

konpeki622:
Well, I will leave this question open and find why that makes test fail. 🤔

packages/vite-node/src/utils.ts
```diff
@@ -45,14 +45,16 @@ export function isPrimitive(v: any) {
   return v !== Object(v)
 }
 
-export function toFilePath(id: string, root: string): string {
+export function toFilePath(id: string, root: string, raw = false): string {
```
sheremet-va, line 48: This became very confusing. Why as a developer I should use `raw` here? This function should just work, I already gave it path

sheremet-va (changes requested):
As I said, this is wrong solution. To fix the problem you are linking to, we need to load mocked version first. They have the same path. It has nothing to do with this function.

The mocked version is not loaded, because module is already cached:

https://github.com/vitest-dev/vitest/blob/b054cd5e3664055aed47a5818c7f2b79fb90ed97/packages/vite-node/src/client.ts#L85

All caches a have a key similar to `${path}__mock`.

Issue #1470, opened 2022-06-13
Custom test runner


black7375:
### Clear and concise description of the problem

Jest ecosystem has [jest-runner](https://github.com/facebook/jest/tree/main/packages/jest-runner) or [create-jest-runner](https://github.com/jest-community/create-jest-runner).

It can be extended to test in an environment such as an [electron](https://github.com/d4rkr00t/jest-electron-runner).
Also, It would be nice to be able to perform [lint](https://github.com/jest-community/jest-runner-eslint) or [typecheck](https://github.com/jest-community/jest-runner-tsd) with the efficient worker of tinypool.

### Suggested solution

I'd like to provide an API for setting up a custom runner.

### Alternative

It can also refer to Jest's [`projects` config](https://jestjs.io/docs/configuration#projects-arraystring--projectconfig).

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

acutmore:
Some more info here on Jest's custom test runner: 
https://www.youtube.com/watch?v=NtjyeojAOBs "ReactiveConf 2017 - Rogelio Guzman: Jest as a Platform"

PR #1574, opened 2022-06-30
feat: support edge runtime

promer94:
add environment support for https://github.com/vercel/edge-runtime

## limitations
`eval` and `new Function()` still works in vitest edge environment. 
#1470 might help to solve this problem
```js
// ❌ Fail
// Error name:    "EvalError"
// Error message: "Code generation from strings disallowed for this context"
it('should return the correct value', () => {
  let val = eval('2 + 2');
  expect(val).toBe(4);
});

```

### DIFF
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -153,13 +153,14 @@ export default defineConfig({
 
 ### environment
 
-- **Type:** `'node' | 'jsdom' | 'happy-dom'`
+- **Type:** `'node' | 'jsdom' | 'happy-dom' | 'edge-runtime'`
 - **Default:** `'node'`
 
 The environment that will be used for testing. The default environment in Vitest
 is a Node.js environment. If you are building a web application, you can use
 browser-like environment through either [`jsdom`](https://github.com/jsdom/jsdom)
 or [`happy-dom`](https://github.com/capricorn86/happy-dom) instead.
+If you are building edge functions, you can use [`edge-runtime`](https://edge-runtime.vercel.app/packages/vm) environment
 
 By adding a `@vitest-environment` docblock or comment at the top of the file,
 you can specify another environment to be used for all tests in that file:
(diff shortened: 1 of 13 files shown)


docs/config/index.md
```diff
@@ -153,13 +153,14 @@ export default defineConfig({
 
 ### environment
 
-- **Type:** `'node' | 'jsdom' | 'happy-dom'`
+- **Type:** `'node' | 'jsdom' | 'happy-dom' | 'edge'`
```
antfu, line 156:
```suggestion
- **Type:** `'node' | 'jsdom' | 'happy-dom' | 'edge-runtime'`
```

`edge` feels too general without knowing the context, maybe better to be `edge-runtime`?

antfu (approved):
I like this in general. Just a few details need to discuss

packages/vitest/src/integrations/env/index.ts
```diff
@@ -1,9 +1,19 @@
...
 export const environments = {
   node,
   jsdom,
   'happy-dom': happy,
+  'edge-runtime': edge,
+}
+
+export const envs = Object.keys(environments)
+
+export const envNpm = {
```
antfu, line 15:
```suggestion
export const envPackageNames: Record<keyof typeof environments, string> = {
```

packages/vitest/src/node/cli-api.ts
```diff
@@ -37,7 +38,8 @@ export async function startVitest(cliFilters: string[], options: CliOptions, vit
   }
 
   if (ctx.config.environment && ctx.config.environment !== 'node') {
-    if (!await ensurePackageInstalled(ctx.config.environment)) {
+    const packName = envNpm[ctx.config.environment]
```
antfu, line 41:
```suggestion
    const packageName = envNpm[ctx.config.environment]
```

antfu (approved):

sheremet-va (changes requested):
Should we also add `@edge-runtime/vm` to peerDependencies? We have all other envs there

sheremet-va (approved):

PR #1582, opened 2022-07-03
feat: make sequencer public, add option to run tests in random order

### LINKED ISSUE #1570: Option to run tests in random order

I may have missed the option in the docs, but it would be great to be able to execute tests in a random order, as Jest or Jasmine allow to.

This is particularly useful to track tests that accidentally depend on another run previously, and that fail when run in isolation.

### Suggested solution

Ideally, having a `random` option that can be set to true to run the tests in random order would be great.
The output could then log the seed used.

Jasmine also allows setting a seed, allowing to run the test in a specific order (for example when a seed failed locally or on CI). So another `seed` option that could be set to a specific number would be great

### Alternative

_No response_

### Additional context

See Jasmine options: https://jasmine.github.io/setup/nodejs.html

### Validations

sheremet-va:
Closes #1570

TODO
- [x] Docs
- [x] Tests

This PR adds a few new options to config:
1. `--shuffle` to CLI flags
2. `sequence` to options
  - `sequencer` allows users to overwrite builtin logic for sharding and sorting 
  - `shuffle` allows tests to be run in random order
  - `seed` makes random order more deterministic

Also added `shuffle` flag to `describe` block to make only part of suite random. I've noticed we have some tests rely on running in the same order, so I though this option might be helpful.

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -316,7 +316,7 @@ When you use `test` in the top level of file, they are collected as part of the 
 
 - **Type:** `(name: string, fn: TestFunction, timeout?: number) => void`
 
-  Vitest provides a way to run all tests in random order via CLI flag [`--random`](/guide/cli) or config option [`sequence.shuffle`](/config/#sequence-shuffle), but if you want to have only part of your test suite to run tests in random order, you can mark it with this flag.
+  Vitest provides a way to run all tests in random order via CLI flag [`--sequence.shuffle`](/guide/cli) or config option [`sequence.shuffle`](/config/#sequence-shuffle), but if you want to have only part of your test suite to run tests in random order, you can mark it with this flag.
 
   ```ts
   describe.shuffle('suite', () => {
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -320,9 +320,9 @@ When you use `test` in the top level of file, they are collected as part of the 
 
   ```ts
   describe.random('suite', () => {
-    test('concurrent test 1', async () => { /* ... */ })
-    test('concurrent test 2', async () => { /* ... */ })
-    test('concurrent test 3', async () => { /* ... */ })
+    test('random test 1', async () => { /* ... */ })
+    test('random test 2', async () => { /* ... */ })
+    test('random test 3', async () => { /* ... */ })
   })
   // order depends on sequence.seed option in config (Date.now() by default)
   ```
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -312,6 +312,23 @@ When you use `test` in the top level of file, they are collected as part of the 
   describe.todo.concurrent(/* ... */) // or describe.concurrent.todo(/* ... */)
   ```
 
+### describe.random
+
+- **Type:** `(name: string, fn: TestFunction, timeout?: number) => void`
+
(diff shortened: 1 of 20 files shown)


docs/api/index.md
```diff
@@ -312,6 +312,23 @@ When you use `test` in the top level of file, they are collected as part of the
...
   ```
 
+### describe.random
+
+- **Type:** `(name: string, fn: TestFunction, timeout?: number) => void`
+
+  Vitest provides a way to run all tests in random order via CLI flag [`--random`](/guide/cli) or config option [`sequence.random`](/config/#sequence-random), but if you want to have only part of your test suite to run tests in random order, you can mark it with this flag.
+
+  ```ts
+  describe.random('suite', () => {
```
antfu, line 322: If so, should we also support `repeat` for `describe`? For example:

describe.repeat(10).random('repect 10 times randomly', () => {
  // ...
})

sheremet-va: How is this has anything to do with random? 👀

antfu: No necessary in this PR. Using random feels more unstable to identify the ordering bug (like sometimes the change passes the CI because the order does not hit the bug, but will later). Running random tests multiple times could reduce such scenarios and make the tests more stable.

sheremet-va: > No necessary in this PR, just an idea. Using random feels more unstable to identify the ordering bug (like sometimes the change passes the CI because the order does not hit the bug, but will later). Running random tests multiple times could reduce such scenarios and make the tests more stable.

Sure, but this also should be done on the global level, yes? This PR adds `vitest --random`, so there should also be `vitest --random --repeat 10`.

It's also not clear what should be done on error. Do we fail the suite when it hits the first error? Do we show summary of all 10 runs? I think if we want to add this, it's better be done in other PR, and if/when people will ask for it.

I really want to switch to `@vitest/browser`, but all this enhancements PRs stop me 😄

antfu: Sure we could do that later 👍

antfu (approved):
The feature looks good to me. But do you think `shuffle` would be better than `random`?

sheremet-va:
> I mean the term `random` -> `shuffle` in general. From my understanding:
> 
> * `random` is a more general term, it could combine with many things, like random in order, or decide if a test should run randomly, or pick a random test, or both.
> * `shuffle` specifically means random the order of an array, in which the items of that array are still the same.
> 
> Refs:
> 
> * https://wikidiff.com/random/shuffle
> * https://www.italki.com/post/question-297430

Sure, let's name it shuffle instead.

docs/config/index.md
```diff
@@ -582,3 +582,34 @@ Options to configure Vitest cache policy. At the moment Vitest stores cache for
...
+A custom class that defines methods for sharding and sorting. You can extend `BaseSequencer` from `vitest/node`, if you only need to redefine one of the `sort` and `shard` methods, but both should exist.
+
+Sharding is happening before sorting, and only if `--shard` option is provided.
+
+#### sequence.shuffle
+
+- **Type**: `boolean`
+- **Default**: `false`
+
+If you want tests to run randomly, you can enable it with this option, or CLI argument [`--shuffle`](/guide/cli).
```
antfu, line 606:
```suggestion
If you want tests to run randomly, you can enable it with this option, or CLI argument [`--sequence.shuffle`](/guide/cli).
```

PR #1600, opened 2022-07-05
refactor: update vue 2 example

sheremet-va:
Updated Vue 2 example. We still don't know why script setup doesn't update HTML tho 🤔


sheremet-va:
@antfu did you have a change to see why it doesn't re-render HTML? I checked and values do in fact change, and click handler is called.

sheremet-va:
Continued in #1752

Issue #1603, opened 2022-07-06
Cache implementation differs from Vite


seivan:
I see the implementation is fairly recent form  [https://github.com/vitest-dev/vitest/pull/1541](this), but could they have gotten the keys wrong? 

The cache config is `{dir: string}` while on Vite it's just a straight up `string`
Vite also calls the key `cacheDir` and not `cache`.

Because of that, this will always fail:
https://github.com/vitest-dev/vitest/blob/9c607572800a90e1b27d433239bfd05a7b4354c3/packages/vitest/src/node/config.ts#L185-L189

And the cache will default to `node_modules/.vitest` as opposed to what the user has defined. 

The workaround for this is of course, to repeat the key a second time for vitest, but that's a bit unnecessary. 

In fact, `cacheDir ` isn't even successfully omitted from the config type as the name is wrong. 
https://github.com/vitest-dev/vitest/blob/9c607572800a90e1b27d433239bfd05a7b4354c3/packages/vitest/src/types/config.ts#L418

I can do the PR if we're all on the same page. 

### Reproduction

Set `cacheDir` config on `vite.config.ts` to be `somethingElse`.
Run tests with said config. 
Notice `somethingElse` doesn't exist but `node_modules/.vitest` does. 

### System Info

```shell
-
```

### Used Package Manager

yarn

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

sheremet-va:
Caches should be in different folders to not cause any collision. They serve different purposes.

sheremet-va:
Well, `cacheDir` is a final destination for Vite cache right now. We can't put our cache inside, because it might collide, or you might delete it with `--force` when running application.

seivan:
[...]

PR #1604, opened 2022-07-06
Fix: Using Vite cacheDir key unless user define.

seivan:
Quick try to see if any tests fail.

Related: https://github.com/vitest-dev/vitest/issues/1603

### DIFF
--- a/packages/vitest/src/node/config.ts
+++ b/packages/vitest/src/node/config.ts
@@ -182,7 +182,7 @@ export function resolveConfig(
   if (typeof resolved.css === 'object')
     resolved.css.include ??= [/\.module\./]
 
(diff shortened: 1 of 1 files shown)


sheremet-va:
As I said in discussion, cache dirs should be separate. If you want them to be the same in your project, copy it to `cache.dir`, but beware that they might collide (vite can unexpectedly clear vitest cache and vice versa).

seivan:
Yeah I don't see it. 
If you're picking the parent folder, then vite clearing its cache dir, won't affect the parent directory at all.
And if they pick root  straight up as their cache directory, then they have much bigger problem. 
I am just trying to unify the settings so the config keys and choices don't differ, but you do you.

PR #1620, opened 2022-07-09
Add options to resolveSnapshotPath

hermanbanken:
My team wants to save our snapshots in different files, depending on some variables that are defined by a table (`describe.each`). Specifically we want to run the same test for various versions of clients. Storing all of these snapshots in the same file has some very annoying downsides, because the diff tools will confuse the addition and deletion of versions and display generally unhelpful outputs for human review.

Today the snapshot path depends only on the test file and the extension (`.snap`) and given that it runs in a different thread (using some RPC protocol) we can't pass down variables using AsyncLocalStorage or other mechanisms with global state.

This PR adds the easily customizable Test.Context object to the snapshot resolver, allowing users to conveniently specify a different path per context.

I'm not sure this works due to the nature of the RPC: can it send other parameters than strings? Does it fall apart if the object isn't JSON serialisable? I'm not sure. But I wanted to create this PR already, to make the proposal and gather some opinions.

### DIFF
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -503,7 +503,7 @@ Format options for snapshot testing. These options are passed down to [`pretty-f
 
 ### resolveSnapshotPath
 
-- **Type**: `(testPath: string, snapExtension: string) => string`
+- **Type**: `(testPath: string, snapExtension: string, options: { context: any }) => string`
 - **Default**: stores snapshot files in `__snapshots__` directory
 
 Overrides default snapshot path. For example, to store snapshots next to test files:
--- a/packages/vitest/src/integrations/snapshot/client.ts
+++ b/packages/vitest/src/integrations/snapshot/client.ts
@@ -40,7 +40,7 @@ export class SnapshotClient {
           filePath,
           new SnapshotState(
             filePath,
-            await rpc().resolveSnapshotPath(filePath),
+            await rpc().resolveSnapshotPath(filePath, undefined, { context: this.test.context }),
             getWorkerState().config.snapshotOptions,
           ),
         )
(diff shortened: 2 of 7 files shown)


sheremet-va:
> I'm not sure this works due to the nature of the RPC: can it send other parameters than strings? Does it fall apart if the object isn't JSON serialisable? I'm not sure. But I wanted to create this PR already, to make the proposal and gather some opinions.

These are the rules for serialisation: https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm

If it fails to clone, it will throw an error.

sheremet-va:
I am not against the idea, but we will need to change how we resolve snapshot paths. If we change `resolveSnapshotPath` from function to a string (path to a module), then we can import it inside test thread, and pass whatever we want. This will be a breaking change:

```diff
- resolveSnapshotPath?: (path: string, extension: string) => string
+ resolveSnapshotPath: string
```

The module should return `default` as a function that we used previously.

thomasmattheussen:
> My team wants to save our snapshots in different files

We have the exact same problem. We'd like to use vitest but are now stuck using jest with https://github.com/igor-dv/jest-specific-snapshot

unional:
I'm using workspace and browser mode and here is my 2 cents:

- to work with workspace, the approach also need `projectRoot`. For string approach, that means some templating (`{projectRoot}/...` or `<projectRoot>/...`)
- `page.screenshot()` creates images with counter (`a.spec.tsx/test-something-{1..n}.png`). `resolveSnapshotPath()` does not have a way to do the same. Need that counter or additional templating.

PR #1633, opened 2022-07-12
feat: make fallbackCJS mode try resolve from `/es/...` path to `/lib/...` path

zthxxx:


packages/vite-node/src/externalize.ts
```diff
@@ -30,9 +30,9 @@ export function guessCJSversion(id: string): string | undefined {
   }
   if (id.match(ESM_FOLDER_RE)) {
     for (const i of [
-      id.replace(ESM_FOLDER_RE, '/lib/$1'),
       id.replace(ESM_FOLDER_RE, '/umd/$1'),
       id.replace(ESM_FOLDER_RE, '/cjs/$1'),
+      id.replace(ESM_FOLDER_RE, '/lib/$1'),
```
zthxxx, line 35: I think the order of try to resolve maybe from non-standard to more standard.

antfu: How do you define "more standard"?

zthxxx: like `"umd"` is `"Universal Module Definition"`, `"cjs"` is `"CommonJS"`;

However `"lib"` usually means put `"CommonJS"` format product into this directory, but it isn't any abbreviation of standard, it maybe trace back to authors without compile system (babel/typescript) that they put **source code** in `"lib/"` directory without other dirs ("src/", "es/", "build/").

Issue #1628, opened 2022-07-12
Support jest-like `moduleNameMapper` in vitest


zthxxx:
### Clear and concise description of the problem

Sometime we have to deal with non-standard third-party packages, such as replace `xxx/es/` to `xxx/lib` which are no need in "vite dev / build" but ONLY need in testing phase, and it's also NOT satisfy vitest [`deps.fallbackCJS`](https://vitest.dev/config/#deps-fallbackcjs) mode.

### Suggested solution

Maybe support some configuration like [jest `moduleNameMapper`](https://jestjs.io/docs/configuration#modulenamemapper-objectstring-string--arraystring).

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

sheremet-va:
Use config option [`resolve.alias`](https://vitejs.dev/config/#resolve-alias).

zthxxx:
@sheremet-va  such as replace xxx/es/ to xxx/lib which are **NO need in "vite dev / build"** but ONLY need in testing phase

sheremet-va:
> @sheremet-va such as replace xxx/es/ to xxx/lib which are **NO need in "vite dev / build"** but ONLY need in testing phase

Use `process.env.VITEST` or `mode === 'test'` to distinguish between test and dev env.

zthxxx:
Ugly, and that's why vitest has a [`fallbackCJS` mode to do those alias](https://github.com/vitest-dev/vitest/blob/v0.18.0/packages/vite-node/src/externalize.ts#L19).

sheremet-va:
> Expose the capability to the user is better than [hide everything behind guessCJSversion](https://github.com/vitest-dev/vitest/blob/v0.18.0/packages/vite-node/src/externalize.ts#L6-L41).

We expose this capability with `resolve.alias`. 

[...]

PR #1634, opened 2022-07-12
feat: add alias option

sheremet-va:
Linked #1628


antfu:
I am not so sure. I understand it could be easier for users to have env-specific alias, yet it might still be limited. I am concerned this would introduce some slippery slope effect that we might need to re-expose every vite option under `test`. Since we have `vitest.config` and `process.env.VITEST` on top of the fact that the config is js, I would prefer to reuse Vite's options as much as possible.

sheremet-va:
> I am not so sure. I understand it could be easier for users to have env-specific alias, yet it might still be limited. I am concerned this would introduce some slippery slope effect that we might need to re-expose every vite option under `test`. Since we have `vitest.config` and `process.env.VITEST` on top of the fact that the config is js, I would prefer to reuse Vite's options as much as possible.

I think we should add `plugins` and `alias` support to `test` property, since they are highly requested, and it's easier to configure with a simple `test.alias/test.plugins` than go to the issues/discord and ask. Plugins are harder to implement right now, so I am waiting on multi repo support.

Anything else should be on Vite config. What do you think, @patak-dev?

antfu:
OK, I am ok to have `alias` and `plugins` for the DX.

docs/config/index.md
```diff
@@ -109,6 +109,12 @@ This might potentially cause some misalignment if a package has different logic
 
 Interpret CJS module's default as named exports.
 
+### alias
+
+- **Type:** `Record<string, string> | Array<{ find: string | RegExp, replacement: string, customResolver?: ResolverFunction | ResolverObject }>`
+
+Define custom aliases when running inside tests. They will be merged with aliases from `resolve.alias`.
```
belgattitude, line 116: Could it be named something else than `alias` ? 

Genuinely asking cause it's not immediately clear to me how they compare/plug into the vite ones (`alias` in vite 2 and `resolve.alias` in vite 3). Easy to be confused when quickly checking config or looking for an existing issue.

Could it be renamed ? (ie: extraAlias, moduleMapper, testAlias...)

sheremet-va: It’s on test.alias, what is confusing here?

belgattitude: You're absolutely right, no confusion. My bad, I was confused :)

test/core/vitest.config.ts
```diff
@@ -56,5 +56,13 @@ export default defineConfig({
     sequence: {
       seed: 101,
     },
+    alias: [
+      {
+        find: 'test-alias',
+        replacement: '',
+        // vitest doesn't crach because function is defined
```
sheremet-va, line 63:
```suggestion
        // vitest doesn't crash because function is defined
```

PR #1654, opened 2022-07-16
chore: disable netlify build if there are no changes to docs

Demivan:


Demivan:
@antfu would be nice to disable comments when Netlify build starts and when it is cancelled. It is a little bit annoying to get notifications about this.

Demivan:
Closing this. It breaks updating Vitest version in the header when releasing

antfu (approved):

antfu:
Thanks! I have removed the comments - I don't like it either, donno why it's been set in the first place.

PR #1666, opened 2022-07-18
docs(fix): lang and copy in code blocks in light mode

userquin:
https://streamable.com/y16j3q


sheremet-va (approved):

antfu:
I think it better be fixed in VitePress

PR #1673, opened 2022-07-19
fix: apply Vite resolving algorithm to node_modules libraries

### LINKED ISSUE #1638: unable to alias modules using Vitest/Vite
### Describe the bug

I am replacing my old Jest tests with Vitest (awesome work btw) but I am unable to alias a module in similar manner. Failing to do so the different versions conflict and crash, failing the test. This might be more related to Vite but I succeeded aliasing the same module using just Vite so I'm not sure what's the cause of the problem.

### Reproduction

https://github.com/TeemuKoivisto/vite-unable-to-alias-module

### System Info

```shell
System:
    OS: macOS 11.5.2
    CPU: (8) x64 Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
    Memory: 232.34 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.14.0 - ~/.nvm/versions/node/v16.14.0/bin/node
    Yarn: 1.22.17 - ~/.nvm/versions/node/v16.14.0/bin/yarn
    npm: 8.3.1 - ~/.nvm/versions/node/v16.14.0/bin/npm
  Browsers:
    Chrome: 103.0.5060.114
    Firefox: 102.0
    Safari: 14.1.2
  npmPackages:
    vite: ^3.0.0 => 3.0.0 
    vitest: ^0.18.0 => 0.18.0
```

### Used Package Manager

pnpm

### Validations

sheremet-va:
This PR adds experimental loader and conditions, so native Node resolving algorithm knows about Vite's resolving algorithm.

This is a giant breaking change, but still should not affect a lot of people. This fixes issues when people applied custom conditions and they were working in Vite, but didn't work in Vitest, because it was imported from another library. This PR also allows applying `alias` to external libraries - this is useful for `preact`, for example.

We still don't allow mocking external libraries within another external library!

Closes #1638

### DIFF
--- a/packages/vitest/src/node/pool.ts
+++ b/packages/vitest/src/node/pool.ts
@@ -7,7 +7,7 @@ import { Tinypool } from 'tinypool'
 import { createBirpc } from 'birpc'
 import type { RawSourceMap } from 'vite-node'
 import type { ResolvedConfig, WorkerContext, WorkerRPC } from '../types'
-import { distDir } from '../constants'
+import { distDir, rootDir } from '../constants'
 import { AggregateError } from '../utils'
 import type { Vitest } from './core'
 
@@ -20,6 +20,8 @@ export interface WorkerPool {
 
 const workerPath = _url.pathToFileURL(resolve(distDir, './worker.mjs')).href
 const loaderPath = _url.pathToFileURL(resolve(distDir, './loader.mjs')).href
+
+const suppressLoaderWarningsPath = resolve(rootDir, './suppress-warnings.cjs')
 
 export function createPool(ctx: Vitest): WorkerPool {
   const threadsCount = ctx.config.watch
@@ -41,10 +43,11 @@ export function createPool(ctx: Vitest): WorkerPool {
     minThreads,
 
     execArgv: [
+      '--require',
+      suppressLoaderWarningsPath,
       '--experimental-loader',
       loaderPath,
       ...conditions || [],
-      '--no-warnings',
     ],
   }
 
--- a/packages/vitest/src/node/pool.ts
+++ b/packages/vitest/src/node/pool.ts
@@ -19,6 +19,7 @@ export interface WorkerPool {
 }
 
 const workerPath = _url.pathToFileURL(resolve(distDir, './worker.mjs')).href
+const loaderPath = _url.pathToFileURL(resolve(distDir, './loader.mjs')).href
 
 export function createPool(ctx: Vitest): WorkerPool {
   const threadsCount = ctx.config.watch
@@ -36,6 +37,13 @@ export function createPool(ctx: Vitest): WorkerPool {
 
     maxThreads,
     minThreads,
+
+    execArgv: [
+      '--experimental-loader',
+      loaderPath,
+      ...ctx.server.config.resolve.conditions?.flatMap(c => ['-C', c]) ?? [],
+      '--no-warnings',
+    ],
   }
 
   if (ctx.config.isolate) {
--- a/packages/vitest/src/node/pool.ts
+++ b/packages/vitest/src/node/pool.ts
@@ -29,6 +29,8 @@ export function createPool(ctx: Vitest): WorkerPool {
   const maxThreads = ctx.config.maxThreads ?? threadsCount
   const minThreads = ctx.config.minThreads ?? threadsCount
 
+  const conditions = ctx.server.config.resolve.conditions?.flatMap(c => ['-C', c])
+
   const options: TinypoolOptions = {
     filename: workerPath,
     // TODO: investigate further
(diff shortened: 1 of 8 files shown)


packages/vitest/src/node/pool.ts
```diff
@@ -36,6 +37,13 @@ export function createPool(ctx: Vitest): WorkerPool {
 
     maxThreads,
     minThreads,
+
+    execArgv: [
+      '--experimental-loader',
+      loaderPath,
+      ...ctx.server.config.resolve.conditions?.flatMap(c => ['-C', c]) ?? [],
+      '--no-warnings',
```
antfu, line 45: Is no warnings bypass all warnings from Node? Or can it be selective?

packages/vitest/src/runtime/loader.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+export const resolve: Resolver = async (url, context, next) => {
+  const { parentURL } = context
+  if (!parentURL || !parentURL.includes('node_modules'))
+    return next(url, context, next)
+
+  const id = normalizeModuleId(url)
+  const importer = normalizeModuleId(parentURL)
+  const state = getWorkerState()
```
sheremet-va, line 49: This won't be possible in future versions of loader, because it will be running in a separate thread. Is there a way to communicate between the main thread? @antfu I didn't find a way to pass down `workerPort` to the loader.

packages/vitest/src/node/pool.ts
sheremet-va: I don't think you can select what warning to suppress. Here I am suppressing `experimental` warning.

sheremet-va: We can do something like this:

```ts
process.removeAllListeners('warning')
process.addEventListener('warning', e => !e.message.includes('experimental') && process.stdout.write(e.message))
```

packages/vitest/src/runtime/loader.ts
sheremet-va: We can pass down `argv` to the loader, so it would know what to listen for example, but we need to setup something to listen to.

packages/vitest/src/node/pool.ts
antfu: Ref: how `tsx` handles it
https://github.com/esbuild-kit/tsx/blob/develop/src/suppress-warnings.cts

packages/vitest/src/runtime/loader.ts
antfu: I guess we need to count on Node to provide a way for communication? (and I think they should if they move it to a thread)

antfu:
Can we have a flag to toggle this behavior?

sheremet-va:
> Can we have a flag to toggle this behavior?

How would you call it? I personally don't think it should be toggable. From my experience with Vitest issues for the past several months, is that people expect aliasing to work as it works in Jest (also inside `node_modules`)

antfu (approved):

antfu:
Ok, we could try and see how it works

thebuilder:
After upgrading to `v0.20.1`, most of my tests are now failing with:

```
SyntaxError: Named export '__assign' not found. The requested module 'tslib' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'tslib';
const { __assign } = pkg;
```

It seems to be caused by `framer-motion` using [tslib](https://github.com/microsoft/tslib). The package does include an [exports field](https://github.com/microsoft/tslib/blob/main/package.json#L29-L36) and module support, so a bit weird why this is popping up. Thinking it's related to how node modules are resolved now?

sheremet-va:
> After upgrading to `v0.20.1`, most of my tests are now failing with:
> 
> ```
> SyntaxError: Named export '__assign' not found. The requested module 'tslib' is a CommonJS module, which may not support all module.exports as named exports.
> CommonJS modules can always be imported via the default export, for example using:
> 
> import pkg from 'tslib';
> const { __assign } = pkg;
> ```
> 
> It seems to be caused by `framer-motion` using [tslib](https://github.com/microsoft/tslib). The package does include an [exports field](https://github.com/microsoft/tslib/blob/main/package.json#L29-L36) and module support, so a bit weird why this is popping up. Thinking it's related to how node modules are resolved now?

Please, open a separate issue with minimal reproduction. Thank you! 😄

cameronbraid:
This change breaks projects using yarn pnp

0.19.1 works, when upgrading to 0.20 I get 

```
Serialized Error: {
  "errors": [
    [Error: Cannot find module '/workspace/octane/.yarn/__virtual__/vitest-virtual-b7339852d0/0/cache/vitest-npm-0.20.2-9730ec2bf1-eea156689f.zip/node_modules/vitest/dist/loader.mjs' imported from /workspace/octane/
Did you mean to import /workspace/octane/.yarn/__virtual__/vitest-virtual-b7339852d0/0/cache/vitest-npm-0.20.2-9730ec2bf1-eea156689f.zip/node_modules/vitest/dist/loader.mjs?],
  ],
}
```

cameronbraid:
Looks like its already reported https://github.com/vitest-dev/vitest/issues/1758

PR #1676, opened 2022-07-19
feat: support `istanbul` coverage provider

### LINKED ISSUE #1252: Coverage: give the choice between c8 and nyc

I observed huge coverage gaps when transitioning from Jest to Vitest, the reason coming from C8 not ignoring a few things:
- `types.ts` were ignored by default in Jest and not in Vitest
- `*.entity.ts` were ignored by default in Jest and not in Vitest
- `*.test.ts` were ignored by default in Jest and not in Vitest
- For includes/excludes, Vitest takes absolute paths, Jest accepts relative paths (that may lead to include or exclude paths not working and giving erroneous results in your coverage)
- Comments are ignored by default in Jest and not in Vitest
- Import lines are ignored by default in Jest and not in Vitest
- Objects spread in multiple lines are considered as one line in Jest, and as many lines as there is in Vitest

The "deep reason" seems that Jest uses nyc, and Vitest uses c8, and they don't have the same way of running through the code to make the coverage report.
Besides, the way c8 does it (not ignoring comments) seems objectively wrong, because adding comments in any covered branch in any file will actually increase the code coverage when it really shouldn't. Same for multiple line objects.

### Suggested solution

Give the choice to use nyc instead of c8.

### Alternative

Fix c8? There is actually an issue dating from 2019 talking about the comments counted as covered here: https://github.com/bcoe/c8/issues/182

### Additional context

_No response_

### Validations

AriPerkkio:
Closes #1252

Documentation of the istanbul packages is not perfect. I had to do some analyzing of `jest` and `nyc` to get things working.

- https://github.com/facebook/jest/blob/main/packages/jest-reporters/src/CoverageReporter.ts
- https://github.com/istanbuljs/nyc/tree/master/lib

Creates common interface for coverage providers to implement and adds new coverage provider:

- `onFileTransform`:
  - c8: No file transforms needed
  - istanbul: Instrument files under testing with `istanbul-lib-instrument`. Coverage objects are added to `globalThis.__VITEST_COVERAGE__`

- `onBeforeFilesRun`:
  - c8: Instruct v8 to enable coverage collection by setting up `process.env.NODE_V8_COVERAGE`
  - istanbul: No actions needed

- `onAfterSuiteRun`:
  - c8: Write v8 coverage reports to file system using `v8.takeCoverage()`
  - istanbul: Collect the `globalThis.__VITEST_COVERAGE__` from workers through birpc into main thread

- `onAfterAllFilesRun`:
  - c8: Create report using `c8/lib/report` 
  - istanbul: Process and merge coverage objects with `istanbul-lib-coverage` and create report with `istanbul-reports`

### Example

`vitest/test/coverage-test/src/utils.ts`: Istanbul on the left with 57% coverage, v8 on the right with 70% coverage: 

image


AriPerkkio:
This is still a work-in-progress but it might be worth to create a draft PR at this point to prevent duplicate work.

I think the `refactor: create interface for coverage logic` part is complete, but istanbul integration still requires work. I can continue this later on, but if anyone is eager to finish this PR let me know.

I think it's also good time for Vitest team to take a look at this overall and call for a stop if this doesn't look good.

packages/vitest/src/node/core.ts
```diff
@@ -84,12 +87,17 @@ export class Vitest {
...
+    this.config.coverage = this.coverageReporter.resolveOptions()
+
     this.runningPromise = undefined
 
     this._onRestartListeners.forEach(fn => fn())
 
-    if (resolved.coverage.enabled)
-      await cleanCoverage(resolved.coverage, resolved.coverage.clean)
+    if (this.config.coverage.enabled)
+      await this.coverageReporter.clean(this.config.coverage.clean)
```
Demivan, line 100: Suggestion: create `NullCoverageProvider` (name is WIP :slightly_smiling_face: ) that does nothing and all `this.config.coverage.enabled` can be removed.

packages/vitest/src/node/plugins/index.ts
```diff
@@ -146,6 +147,7 @@ export async function VitestPlugin(options: UserConfig = {}, ctx = new Vitest())
     MocksPlugin(),
     GlobalSetupPlugin(ctx),
     CSSEnablerPlugin(ctx),
+    InstrumenterPlugin(ctx),
```
Demivan, line 150: Can this plugin be excluded like `UIPlugin` if coverage is disabled?

packages/vitest/src/runtime/run.ts
```diff
@@ -288,7 +287,17 @@ export async function startTests(paths: string[], config: ResolvedConfig) {
...
+  // so that it would only be declared inside coverage/istanbul.ts?
+  // Passing globalThis to onFilesRun does not work
+
+  // @ts-expect-error -- untyped global
+  const coverage = globalThis.__VITEST_COVERAGE__
+
+  // TODO: Not sure if this will work for c8. Previously v8.takeCoverage() was called
+  // here inside the worker. Now, this will simply inform the main process to call it.
+
+  rpc().onFilesRun(coverage)
```
Demivan, line 300: I think we can keep `takeCoverage` c8 specific logic here. If v8 coverage is not enabled `v8.takeCoverage` should be a noop

Demivan (changes requested):

packages/vitest/src/integrations/coverage/c8.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    await report.run()
+
+    if (this.ctx.config.coverage.enabled && this.ctx.config.coverage.provider === 'c8') {
+      if (this.ctx.config.coverage['100']) {
+        this.ctx.config.coverage.lines = 100
+        this.ctx.config.coverage.functions = 100
+        this.ctx.config.coverage.branches = 100
+        this.ctx.config.coverage.statements = 100
+      }
```
Demivan, line 111: Move to `resolveC8Options`

packages/vitest/src/integrations/coverage/istanbul.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  resolveOptions(): ResolvedCoverageOptions {
+    return this.options
+  }
+
+  onFileTransform(sourceCode: string, id: string, pluginCtx: TransformPluginContext) {
+    if (!this.testExclude.shouldInstrument(id))
+      return
+
+    // TODO: Sourcemaps are not accurate.
```
AriPerkkio, line 71: Reminder for later: The key is to use `istanbul-lib-source-maps` here. Sourcemaps even for TS seem to be correct with that one. 🎉

packages/vitest/src/node/plugins/index.ts
AriPerkkio: Sure, we can add check for the `options.coverage?.enabled` here.

Note that there is a TODO for this in the plugin itself. It would be returning `null` which gets filtered out here. 

```ts
export function InstrumenterPlugin(ctx: Vitest): VitePlugin | null {
  // // Skip coverage reporters which do not need code transforms, e.g. native v8
  // TODO: This would be great but ctx has not yet been initialized
  // if (typeof ctx.coverageReporter.onFileTransform !== 'function')
  //  return null
```

I'm not 100% sure that one can be achieved though since `ctx` has not yet been initialized at this point. That solution would be ideal but since we don't need the `InstrumenterPlugin` at all with `c8`.

packages/vitest/src/integrations/coverage/base.ts
```diff
@@ -0,0 +1,22 @@
+import type { ExistingRawSourceMap, TransformPluginContext } from 'rollup'
```
AriPerkkio, line 1: Importing these rollup types seems to cause out-of-memory locally, CI is OK. These errors started showing up after merging the `browser` feature. 

```
v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool)
```

I've run into similar issues in other projects before. I think rollup is not that memory efficient when it comes to bundling multiple entries.

AriPerkkio: Build seems to work with `export NODE_OPTIONS="--max-old-space-size=8192"`.

packages/vitest/src/node/core.ts
AriPerkkio: This is now implemented in https://github.com/vitest-dev/vitest/pull/1676/commits/b699c0ce17e05b6a8c531901af635e977cd8f09f.

packages/vitest/src/node/core.ts
```diff
@@ -83,12 +87,22 @@ export class Vitest {
 
     this.reporters = await createReporters(resolved.reporters, this.runner)
 
+    if (options.coverage?.enabled && options.coverage?.provider === 'c8')
+      this.coverageReporter = new C8Reporter()
+    else if (options.coverage?.enabled && options.coverage.provider === 'istanbul')
+      this.coverageReporter = new IstanbulReporter()
+    else
+      this.coverageReporter = new NullCoverageProvider()
```
Demivan, line 95: Move this to some `getCoverageProvider(options.coverage)` method in `../integrations/coverage`

Off topic: Should this feature be named coverage reporters or coverage providers?

packages/vitest/src/types/coverage.ts
```diff
@@ -15,41 +15,101 @@ export type CoverageReporter =
...
+
+}
+
+export interface NullCoverageOptions extends BaseCoverageOptions {
+  enabled: false
+}
+
+export interface IstanbulOptions extends BaseCoverageOptions {
+  /* Name of global coverage variable. (optional, default __coverage__) */
+  coverageVariable?: string
```
Demivan, line 67: This should not be public as users should not change it. Same for `coverageGlobal*` options.

packages/vitest/src/integrations/coverage/istanbul.ts
```diff
@@ -0,0 +1,145 @@
...
+  const reportsDirectory = resolve(root, options.reportsDirectory || configDefaults.coverage.reportsDirectory!)
+
+  const resolved = {
+    ...configDefaults.coverage,
+
+    // Custom
+    provider: 'istanbul',
+    coverageVariable,
+    coverageGlobalScope: 'globalThis',
+    coverageGlobalScopeFunc: false,
```
Demivan, line 129: This should not be overridable by `options`

packages/vitest/src/node/core.ts
AriPerkkio: I think we should keep naming these as coverage providers - at least internally in the codebase. I'll rename all new coverage reporters to providers.

Coverage reporters are then the ones `istanbul-repoters` (also used by `c8` internally) support, e.g. `lcov`, `clover`.

What Jest has is [`coverageProviders` configuration option](https://jestjs.io/docs/configuration#coverageprovider-string) with values of `babel` and `v8`. They also expose only a few coverage related options, instead of the whole `c8` API. Both coverage providers use the same public API. If we aim to do the same, that is a breaking change and requires some more work.

Demivan: I don't think we need to make c8 and Istanbul options the same. Proper types for options as it is now is enough in my opinion.

AriPerkkio: Changes done in
- https://github.com/vitest-dev/vitest/compare/ebb21c001a5ea389ad610dee931f4ee16c6be422..fea9cce8ec88641a632de524d46a833dff336769
- https://github.com/vitest-dev/vitest/compare/fea9cce8ec88641a632de524d46a833dff336769..fc1375028b0f44ec8ce16b71f2c602c8a5af3375

packages/vitest/src/integrations/coverage/istanbul.ts
```diff
@@ -0,0 +1,146 @@
...
+  options!: ResolvedCoverageOptions & { provider: 'istanbul' }
+  instrumenter!: Instrumenter
+  testExclude!: InstanceType<TestExclude>
+  coverages: any[] = []
+
+  initialize(ctx: Vitest) {
+    this.ctx = ctx
+    this.options = resolveIstanbulOptions(ctx.config.coverage, ctx.config.root)
+
+    const { createInstrumenter } = require('istanbul-lib-instrument')
```
wight554, line 50: Would be cool to see this configurable to use custom instrumenters like https://github.com/kwonoj/swc-coverage-instrument/

AriPerkkio: This seems like a cool idea, thanks for improvement idea @wight554. However I think it might be best if we leave this customization out of this PR and introduce it in a separate feature afterwards. 
Do you think there is anything important we need to take into consideration at this point? Leave some abstractions in place, etc?
My Rust knowledge is a bit _rusty_ and to me it's not that clear what exactly `import('swc-plugin-coverage-instrument')` would return here. Does it match the API of `istanbul-lib-instrument`?

wight554: > This seems like a cool idea, thanks for improvement idea @wight554. However I think it might be best if we leave this customization out of this PR and introduce it in a separate feature afterwards. Do you think there is anything important we need to take into consideration at this point? Leave some abstractions in place, etc? My Rust knowledge is a bit _rusty_ and to me it's not that clear what exactly `import('swc-plugin-coverage-instrument')` would return here. Does it match the API of `istanbul-lib-instrument`?

I've checked how jest handles this, see https://github.com/facebook/jest/blob/24e0472ac41ea03cdd89ffaea8c5f410ecf6054f/packages/jest-transform/src/ScriptTransformer.ts#L410
In this case all that needs to be added is config value to skip instrumenter part and return unmodified map and code in `onFileTransform`

wight554:
looks good overall, noticed few issues:
1. exclude option seems to be ignored when using instanbul
2. when setting provider to c8 explicitly `vitest run --coverage` skips coverage
3. vitest doesn't suggest installing missing istanbul packages ans skips coverage instead
4. c8 randomly ignores coverage for 1 file in 100% covered code when using c8 with these changes

codebase I used for testing: https://github.com/wight554/blog-template
You might notice schemas are added to coverage report when using istanbul [1] if you try

antfu:
The overall direction looks great to me. Please ping me when you think the work on your side is done, and I can help to do the refactoring and make the packages more general as `@vitest/coverage-istanbul` and even `@vitest/coverage-c8` later.

packages/vitest/src/node/cli-api.ts
```diff
@@ -35,9 +35,21 @@ export async function startVitest(cliFilters: string[], options: CliOptions, vit
...
-      process.exitCode = 1
-      return false
+    // TODO: Requring all these packages to be installed by users may be too much
+    const requiredPackages = ctx.config.coverage.provider === 'istanbul'
+      ? [
+          'istanbul-lib-coverage',
+          'istanbul-lib-instrument',
+          'istanbul-lib-report',
+          'istanbul-reports',
+        ]
```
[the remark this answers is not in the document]
antfu: Yes, it sounds great to me
https://github.com/vitest-dev/vitest/pull/1676#issuecomment-1195844043

AriPerkkio:
From my side all that is left is to do unit+manual testing and add documentation. I think everything else (besides refactoring into separate packages) is done. I'm happy to leave these tasks to other people as well. 😄 

I'm currently travelling and won't be doing anything for the next ~10 days. @antfu if you'd like to start working on this now, feel free to take over the PR. 👍

antfu:
Great, thanks! Have a good time traveling!

packages/vitest/src/node/plugins/index.ts
AriPerkkio: Looks like `options` can only be used for resolving CLI flags here. The configuration file has not yet been processed - at this point we cannot know whether coverage is enabled by the configuration file only.

We'll need to revert this change and always enable the plugin.

packages/vitest/src/runtime/run.ts
AriPerkkio: Alright it seems we need to add the explicit `takeCoverage()` here. Otherwise it will cause random flakiness where the coverage report is partially missed, as reported by @wight554 in https://github.com/vitest-dev/vitest/pull/1676#issuecomment-1195793478.

> 4. c8 randomly ignores coverage for 1 file in 100% covered code when using c8 with these changes

packages/vitest/src/types/coverage.ts
```diff
@@ -15,41 +15,89 @@ export type CoverageReporter =
...
    * @default 'text'
    */
   reporter?: Arrayable<CoverageReporter>
+}
+
+export interface NullCoverageOptions extends BaseCoverageOptions {
+  enabled: false
+}
+
+export interface IstanbulOptions extends BaseCoverageOptions {
```
AriPerkkio, line 62: These `IstanbulOptions` require some API design. Currently these are options of `istanbul-lib-instrument` which already includes some options which are likely not to be configured by the user.

Instead we should be providing a subset of configration options of all used [istanbul packages](https://github.com/istanbuljs/istanbuljs/tree/master/packages). Currently we are using `istanbul-lib-coverage`, `istanbul-lib-instrument`, `istanbul-lib-report`, `istanbul-lib-source-maps` and `istanbul-reports`. For example the `istanbul-lib-report` package contains [`skipFull` and `skipEmpty` flags](https://github.com/istanbuljs/istanbuljs/tree/master/packages/istanbul-lib-report#istanbul-lib-report) which are something users might be interested in.

Before deciding the API I don't think there is much point of writing any documentation. One option is to start with a really minimal API and wait for feature requests. 😁

AriPerkkio: I went through the undocumented configuration options of used `istanbul`-packages and checked options which should be exposed to users. There's now `skipFull`, `ignoreClassMethods` and `watermarks`. Also added a custom `exclude` option. 

`nyc` has a `check-coverage` option which can be used to compare coverage against thresholds: https://github.com/istanbuljs/nyc#coverage-thresholds. However this is not something `istanbul`-packages provide out-of-the-box and requires custom implementation. 

References from nyc and c8:
- https://github.com/istanbuljs/nyc/blob/master/index.js#L480
- https://github.com/bcoe/c8/blob/main/lib/commands/check-coverage.js#L38

I think support for thresholds could be left for later - otherwise this PR will just keep growing.

AriPerkkio:
There's now some initial documentation and unit tests in place. I think the next logical step is to refactor these coverage providers into their own packages - https://github.com/vitest-dev/vitest/pull/1676#issuecomment-1195844043.

sheremet-va:
> There's now some initial documentation and unit tests in place. I think the next logical step is to refactor these coverage providers into their own packages - [#1676 (comment)](https://github.com/vitest-dev/vitest/pull/1676#issuecomment-1195844043).

I think we should have these packages:

- @vitest/coverage-v8
- @vitest/coverage-istanbul

Then Vitest checks for `coverage` field, if package is installed. We might also expose this as `CoverageProvider` for extending with custom provider.

packages/vitest/src/types/coverage.ts
AriPerkkio: Thresholds are now implemented - it wasn't that big feature after all. 

I think all needed configuration options are now supported.

antfu:
Pushed some refactorings:

- [x] Move to separate packages
- [ ] Make it work
- [ ] Allow `--coverage=c8`
- [ ] Prompt of choice when passing `--coverage`
- [ ] Prompt for auto-installing
- [ ] Support custom coverage provider

antfu (approved):

Demivan:
@antfu `coverage-c8` and `coverage-istanbul` build is failing locally.

```
src/provider.ts(6,10): error TS2305: Module '"vitest/config"' has no exported member 'configDefaults'.
```

Demivan:
https://github.com/vitest-dev/vitest/runs/7835819196?check_suite_focus=true#step:6:107

Need to add `node` to tsconfig types

JounQin:
It should be noticed `provider: 'istanbul'` is required if the user wants to use `istanbul`, ideally, if `@vitest/coverage-istanbul` is detected, `istanbul` should be used by default IMO.

antfu:
I thought about it as once, but it could be complex to handle, like if you have both c8 and Istanbul packages installed, or if we have more providers in the future. Also, it might be dangerous to have dependencies change the behavior implicitly.

PR #1678, opened 2022-07-19
feat: add option for ignoring unhandled errors

DerYeger:
I recently migrated a large project from Jest to Vitest.
During the migration, I encountered issues with unhandled errors due to race conditions (caused by poor tests that don't clean up their async React hooks).

Since there's a huge number of tests, without any way of knowing which tests cause these issues, I had to use `patch-package` to prevent Vitest from failing due to unhandled errors.

I am aware that this isn't the proper way, but there was no other way of getting around this issue.

With this PR, I propose a new config option for ignoring unhandled errors.
It's fittingly labelled `dangerouslyIgnoreUnhandledErrors`.

### DIFF
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -538,6 +538,13 @@ export default defineConfig({
 
 Allow tests and suites that are marked as only.
 
+### dangerouslyIgnoreUnhandledErrors
+
+- **Type**: `boolean`
+- **Default**: `false`
+
+Ignore any unhandled errors that occur.
+
 ### passWithNoTests
 
 - **Type**: `boolean`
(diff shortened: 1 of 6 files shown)


sheremet-va (changes requested):
I would much prefer if we could provide a better way of displaying errors to help users find the issues, instead of ignoring it.

DerYeger:
> I would much prefer if we could provide a better way of displaying errors to help users find the issues, instead of ignoring it.

While I agree, I'm not sure if much can be done about improving the reports of unhandled errors.
I'd be happy to work on a solution if someone has an idea.

With the proposed option enabled, Vitest's behaviour would match the behaviour of older Jest versions, making iterative migration much simpler.

antfu:
Or maybe we return a different exitCode so you could ignore it with a UNIX pipe or some ci config?

DerYeger:
> Or maybe we could return a different exitCode so you could ignore it with a UNIX pipe or some ci config?

I guess that would work as well (if actually failing tests in presence of unhandled errors still result in exit code 1).

Though it would also have downsides compared to the config option (no central config, no platform independence, can't call package script differently like `nr test --no-ui`).

antfu:
I am fine with adding the option, since we have the `--dangerously` prefix

antfu (approved):

sheremet-va (approved):

Issue #1675, opened 2022-07-19
expect.any and toHaveProperty are not working as expected


aneem:
### Describe the bug

This works with [Jest](https://stackblitz.com/edit/node-y6rpdc?file=index.test.js):
```
expect({ str: 'abcd'}).toHaveProperty('str', expect.any(String));
```
Doesn't with Vitest:

```
❯ test/basic.test.ts:6:26
      4| 
      5| test('expect.any', () => {
      6|   expect({ str: 'abcd' }).toHaveProperty('str', expect…
       |                          ^
      7| });
      8| 

  - Expected   "Any<String>"
  + Received   "abcd"
```

<img width="1218" alt="CleanShot 2022-07-19 at 14 06 03@2x" src="https://user-images.githubusercontent.com/5670204/179702840-c8dc85b7-39a1-4b5f-b4fa-b90b65af1ffc.png">

### Reproduction

https://stackblitz.com/edit/node-fhkkmq?file=test/basic.test.ts

### System Info

```shell
System:
    OS: macOS 12.4
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 893.54 MB / 16.00 GB
    Shell: 3.4.1 - /usr/local/bin/fish
  Binaries:
    Node: 16.14.2 - ~/.asdf/installs/nodejs/16.14.2/bin/node
    Yarn: 1.22.18 - /usr/local/bin/yarn
    npm: 8.5.0 - ~/.asdf/plugins/nodejs/shims/npm
    Watchman: 2022.05.30.00 - /usr/local/bin/watchman
  Browsers:
    Brave Browser: 101.1.38.119
    Chrome: 103.0.5060.114
    Firefox: 102.0
    Safari: 15.5
  npmPackages:
    vite: ^3.0.2 => 3.0.2
    vitest: ^0.18.1 => 0.18.1
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
[...]

PR #1682, opened 2022-07-19
fix: toHaveProperty with expect.any (fix: #1675)

skarab42:
I'm not sure I implemented the fix in the right place, but it seems to fix the problem in #1675.

### DIFF
--- a/packages/vitest/src/integrations/chai/jest-expect.ts
+++ b/packages/vitest/src/integrations/chai/jest-expect.ts
@@ -1,5 +1,5 @@
 import c from 'picocolors'
-import { AssertionError, util as chaiUtil } from 'chai'
+import { AssertionError } from 'chai'
 import type { EnhancedSpy } from '../spy'
 import { isMockFunction } from '../spy'
 import { addSerializer } from '../snapshot/port/plugins'
@@ -303,7 +303,7 @@ export const JestChaiExpect: ChaiPlugin = (chai, utils) => {
 
     const actual = this._obj
     const [propertyName, expected] = args
-    const { value, exists } = chaiUtil.getPathInfo(actual, propertyName)
+    const { value, exists } = utils.getPathInfo(actual, propertyName)
     const pass = exists && (args.length === 1 || jestEquals(expected, value))
 
     return this.assert(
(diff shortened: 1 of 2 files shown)


sheremet-va:
> @sheremet-va Does `jestEquals` support dot notation?
> 
> https://github.com/vitest-dev/vitest/blob/7c2138f2b0439dfba084f8b104af4dd628adca9b/test/core/test/jest-expect.test.ts#L199-L205

`jestEquals` compares two values. How you get them is only for you to decide.

skarab42:
> `jestEquals` compares two values. How you get them is only for you to decide.

Is there any method already used in vitest? like [lodash.get](https://lodash.com/docs/4.17.15#get)? If not, can I use lodash? Or do I have to do my implementation? Sorry for all the questions 😅

sheremet-va:
> > `jestEquals` compares two values. How you get them is only for you to decide.
> 
> Is there any method already used in vitest? like [lodash.get](https://lodash.com/docs/4.17.15#get)? If not, can I use lodash? Or do I have to do my implementation? Sorry for all the questions 😅

I don't think we use it anywhere. You can check source code for this matcher in chai repo, maybe they have a public API for this things.

skarab42:
@sheremet-va I refactored using `jestEquals` and `chai.util.getPathInfo`. Thanks for pointing me to the solution ;)

packages/vitest/src/integrations/chai/jest-expect.ts
```diff
@@ -1,5 +1,5 @@
 import c from 'picocolors'
-import { AssertionError } from 'chai'
+import { AssertionError, util as chaiUtil } from 'chai'
```
sheremet-va, line 2: We already have utils on 

https://github.com/vitest-dev/vitest/blob/a43df9e49f2fe98bb2791d12d1c9ac3350f1ac01/packages/vitest/src/integrations/chai/jest-expect.ts#L35

sheremet-va (changes requested):

packages/vitest/src/integrations/chai/jest-expect.ts
skarab42: Sorry too enthusiastic, I don't look enough at what's already there... I made the change.

sheremet-va (approved):

PR #1693, opened 2022-07-21
docs: add example of using Vitest with Gatsby

apatel369:
Gatsby is a popular react.js framework.
This example will be helpful for integrating vitest with it.


LekoArts:
For a more complete example one could also show how to mock out data pieces from Gatsby, e.g. `useStaticQuery` as shown here:

https://github.com/LekoArts/portfolio-v2/blob/cfee562ca51f2927b85aeb6096f614d0d0491bd2/www/src/components/__tests__/seo.tsx

apatel369:
I had build issues with this branch and never had the chance to fix them. I am abandoning this PR.

PR #1697, opened 2022-07-21
feat: make close timeout configurable

### LINKED ISSUE #1659: Configurable `CLOSE_TIMEOUT`

When running tests with the `globalSetup` option, the tests often timeout before the full teardown is completed. 

For example, I spin up a database and then shut it down in the teardown, The current 1000ms is not sufficient for the task.

### Suggested solution

Allow this timeout to be configured. Perhaps `teardownTimeout`.

### Alternative

_No response_

### Additional context

_No response_

### Validations

krishnan-chandra:
(Closes #1659)

This PR replaces the hardcoded `CLOSE_TIMEOUT` value for timeout on test teardown with a `teardownTimeout` setting. It retains the current default value of 1000 ms (1 second).

### DIFF
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -313,6 +313,13 @@ Default timeout of a test in milliseconds
 
 Default timeout of a hook in milliseconds
 
+### teardownTimeout
+
+- **Type:** `number`
+- **Default:** `1000`
+
+Default timeout to wait for close when Vitest shuts down, in milliseconds
+
 ### silent
 
 - **Type:** `boolean`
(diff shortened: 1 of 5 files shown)


sheremet-va (changes requested):
I don't think changing pnpm-lock is needed

krishnan-chandra:
You're correct! My apologies, pnpm-lock change has been removed now.

Demivan (approved):

sheremet-va (approved):

PR #1723, opened 2022-07-26
feat(vite-node): allow dumpping the transformed code to debug

antfu:


sheremet-va (approved):
I just used `@vitest/ui` to see transformed modules before :P

Issue #1714, opened 2022-07-25
Unable to mock 'vue-meta'


jagoncalves14:
### Describe the bug

I was trying to upgrade Vitest from version `0.12.9` to `0.19.0` and now tests that are asserting a specific function from a module to have been called started failing.

That module is an npm package called `vue-meta`. ([Link](https://www.npmjs.com/package/vue-meta/v/3.0.0-alpha.10))
I have components that `import { useMeta } from 'vue-meta'`.

That `useMeta` function is called when some components are mounted or created.
All I want to do is to spy `useMeta` and assert it, expecting that it gets called. Should be easy, right? 

But for some reason this stopped working after the upgrade and I'm clueless in understanding why.

I was able to create StackBlitz reproducing this. Just open the link and the test will run automatically.

### Reproduction

https://stackblitz.com/edit/vue-vitest-demo

### System Info

```shell
System:
    OS: macOS 12.4
    CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
    Memory: 819.89 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 14.20.0 - ~/.nvm/versions/node/v14.20.0/bin/node
    npm: 8.3.2 - ~/Projects/clearancejobs/node_modules/.bin/npm
  Browsers:
    Chrome: 103.0.5060.134
    Firefox: 100.0
    Safari: 15.5
  npmPackages:
    @vitejs/plugin-vue: 2.1.0 => 2.1.0
    vite: 2.9.9 => 2.9.9
    vitest: 0.19.0 => 0.19.0
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
[...]

PR #1740, opened 2022-07-29
fix: normalize `moduleCache` path

ChpShy:
fix: #1714

This problem is caused by the `moduleCache` key is not same in setup file and the vue component.
In setup file, the key is `/@fs/user/../node_modules/vue-meta/dist/vue-meta.cjs.js`, but in the component, is `/node_modules/vue-meta/dist/vue-meta.cjs.js`.

I'm not sure if this solution is the best, if not, modifications or rewrites are welcome~

### DIFF
--- a/packages/vitest/src/runtime/mocker.ts
+++ b/packages/vitest/src/runtime/mocker.ts
@@ -103,6 +103,7 @@ export class VitestMocker {
   }
 
   private async callFunctionMock(dep: string, mock: () => any) {
+    dep = toFilePath(dep, this.root)
     const cached = this.moduleCache.get(dep)?.exports
     if (cached)
       return cached


sheremet-va (changes requested):
Can you add a test for this?

ChpShy:
> Can you add a test for this?

ok. I will do it later.

jagoncalves14:
Thanks for this @ChpShy.
I applied your change directly in my node_modules/vitest folder to see if it fixes it — and it did! 👏

ChpShy:
I'm confused that I can't reproduce this in `vitest` test. Maybe someone more familiar with `vite` and `vitest` could help with add test? Or I need more time to reproduce it.

ChpShy:
It looks like it is caused by `pnpm workspace`,  In workspace package, test success, because `vue component` also compiled to be `/@fs/Users/xxx/node_modules/.pnpm/vue-meta@3.0.0-alpha.10_vue@3.2.37/node_modules/vue-meta/dist/vue-meta.cjs.js`.

But in demo project, test failed, because `vue component` compiled to be `/node_modules/.pnpm/vue-meta@3.0.0-alpha.10_vue@3.2.37/node_modules/vue-meta/dist/vue-meta.cjs.js`.

And `callFunctionMock` is private, So, how can I add the test? @sheremet-va

ChpShy:
Also, this maybe caused by vite. I'm not sure. Maybe someone can help to provide a reason? Hope for feedback~

ChpShy:
> because

I also test the different version of `vite`. It get the same result.

sheremet-va:
> So, how can I add the test?

You can add a unit test:

```ts
const mockMap = {
  ...your mocks
}
const mocker = new Mocker({ ...options })
const result = mocker.request(pathThatFailedBefore)
```

ChpShy:
I think this had been resolved by pr  #1726

line: https://github.com/vitest-dev/vitest/blob/main/packages/vitest/src/runtime/mocker.ts#L298

Waiting for release~

PR #1742, opened 2022-07-29
feat(types): add partial flag to mocked TypeScript helper (#1739)

### LINKED ISSUE #1739: Partial types for mocked return values

As a developer using Vitest I want to be able to mock only certain properties or functions of a module so that I can keep the type checking from TypeScript but only mock what I need for the test.

Currently when using `vi.mocked` and `mockReturnValue` with TypeScript it will expect a fully compatible type, but most of the time since you're mocking you only need some of the properties to be set.

This is an issue I face a lot when mocking but I haven't found a proper solution, the most common solution to this seems to be casting to `unknown` and but then you lose the type checking.

### Suggested solution

There's a solution suggested [here](https://github.com/kulshekhar/ts-jest/issues/1065) but I think just changing the return type to be a `Partial<T>` would work?

(Copied suggestion just in case:)

```ts
type DeepPartial<T> = {
    [P in keyof T]?: T[P] extends Array<infer U>
        ? Array<DeepPartial<U>>
        : T[P] extends ReadonlyArray<infer U>
            ? ReadonlyArray<DeepPartial<U>>
            : DeepPartial<T[P]>
};

type PartialReturnType<T extends MockableFunction> = 
    ReturnType<T> extends Promise<infer U>
        ? Promise<DeepPartial<U>>
        : ReturnType<T> extends Array<infer P>
            ? Array<DeepPartial<P>>
            : DeepPartial<ReturnType<T>>;

interface MockWithArgs<T extends MockableFunction> extends jest.MockInstance<PartialReturnType<T>, ArgumentsOf<T>> {
    new (...args: ConstructorArgumentsOf<T>): T;
    (...args: ArgumentsOf<T>): PartialReturnType<T>;
}
```

### Alternative

I use `@ts-ignore` so I can keep the types but disable the error.

There's also [this](https://github.com/vitest-dev/vitest/issues/351) related issue but it solves a different problem I think.

### Additional context

My current use case for this is mocking Pinia stores:

```ts
vi.mocked(useAuthStore).mockReturnValue({
  loggedIn: true
});
```
This will produce an error because the return value is missing a lot of Store properties and functions.

### Validations

jfrs:
I'm not sure about the implementation, happy to get any feedback on this. Closes #1739.

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -1830,9 +1830,11 @@ Vitest provides utility functions to help you out through it's **vi** helper. Yo
 ### vi.mocked
 
 - **Type**: `<T>(obj: T, deep?: boolean) => MaybeMockedDeep<T>`
+- **Type**: `<T>(obj: T, options?: { partial?: boolean; deep?: boolean }) => MaybePartiallyMockedDeep<T>`
 
   Type helper for TypeScript. In reality just returns the object that was passed.
 
+  When `partial` is `true` it will expect a `Partial<T>` as a return value.
   ```ts
   import example from './example'
   vi.mock('./example')
--- a/packages/vitest/src/integrations/spy.ts
+++ b/packages/vitest/src/integrations/spy.ts
@@ -61,6 +61,10 @@ export interface Mock<TArgs extends any[] = any, TReturns = any> extends SpyInst
   new (...args: TArgs): TReturns
   (...args: TArgs): TReturns
 }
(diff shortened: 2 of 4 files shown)


sheremet-va:
Maybe it's better to refactor `vi.mocked` to this? 🤔 

```ts
vi.mocked(obj, { partial?, deep? })
```

sheremet-va:
> Something like this?

Yes, something like this 😄

sheremet-va (approved):

antfu (approved):

PR #1755, opened 2022-07-31
refactor: update Enzym example, upgrade Vue plugin

sheremet-va:


sheremet-va:
@antfu TOO SOON 😆 

Now docs are not building

PR #1757, opened 2022-07-31
fix: filter out Symbol.hasInstance

### LINKED ISSUE #1671: [0.18.1] TypeError: Symbol.hasInstance is read-only
### Describe the bug

```
      let [s, l] = e.next;
  ...<omitted>... }'
```

### Reproduction

### System Info

```shell
System:
    OS: Windows 10 10.0.19044
    Memory: 17.74 GB / 31.92 GB
  Binaries:
  Browsers:
    Chrome: 103.0.5060.114
  npmPackages:
    vitest: 0.18.1 => 0.18.1
```

### Used Package Manager

npm

### Validations

nieyuyao:
fix #1671

### DIFF
--- a/packages/vitest/src/runtime/mocker.ts
+++ b/packages/vitest/src/runtime/mocker.ts
@@ -32,8 +32,7 @@ type Key = string | symbol
 
 function isSpecialProp(prop: Key, parentType: string) {
   return parentType.includes('Function')
-      && typeof prop === 'string'
-      && ['arguments', 'callee', 'caller', 'length', 'name'].includes(prop)
+      && ['arguments', 'callee', 'caller', 'length', 'name', Symbol.hasInstance].includes(prop)
 }
 
 interface ViteRunnerRequest {


QGerrit (approved):

QGerrit (commented):
This unfortunately doesn't seem to quite be enough to fix #1671.... at least fully. After manually applying this patch locally, my tests started failing with `TypeError: Cannot read properties of undefined (reading 'dataEmitted')`. `jsdom` test environment in a very basic vue3 app.... oddly, `dataEmitted` doesn't seem to appear anywhere in my node_modules tree.

nieyuyao:
> This unfortunately doesn't seem to quite be enough to fix #1671.... at least fully. After manually applying this patch locally, my tests started failing with `TypeError: Cannot read properties of undefined (reading 'dataEmitted')`. `jsdom` test environment in a very basic vue3 app.... oddly, `dataEmitted` doesn't seem to appear anywhere in my node_modules tree.

emmm.... I tried to find the reason and converted the PR into draft

nieyuyao:
This `TypeError: Cannot read properties of undefined (reading 'dataEmitted')` comes from [here](https://github.com/nodejs/node/blob/948712bc4c4b24edc238222ba6a7e585f51d01ef/lib/internal/streams/readable.js#L1172). Try to read the readableDidRead property and assign to the container when mocking. However, the Readable has not been initialized yet, so an error is thrown.

QGerrit:
Wow, good find! I forgot to mention I was doing `vi.mock('axios')`, that break location makes more sense now.

packages/vitest/src/runtime/mocker.ts
```diff
@@ -32,8 +32,7 @@ type Key = string | symbol
 
 function isSpecialProp(prop: Key, parentType: string) {
   return parentType.includes('Function')
-      && typeof prop === 'string'
-      && ['arguments', 'callee', 'caller', 'length', 'name'].includes(prop)
+      && ['arguments', 'callee', 'caller', 'length', 'name', Symbol.hasInstance].includes(prop)
```
simon-abbott, line 35: I don't like this solution. There are valid cases where we _do_ want to mock `Symbol.hasInstance`, and, as you've noticed, this doesn't actually fix the root cause.

simon-abbott (changes requested):
I don't like this solution. There are valid cases where we _do_ want to mock `Symbol.hasInstance`, and, as you've noticed, this doesn't actually fix the root cause.

I'm working on a fix for the underlying issue which should be up shortly.

simon-abbott:
I've solved #1671 in a more general way over in #1786.

PR #1778, opened 2022-08-03
feat: treat pseudo ESM as ESM with custom loader, disable custom Node Loader by default

### LINKED ISSUE #1753: [0.20.2]: Tets fails if importing `tslib`
### Describe the bug

After upgrading to v0.20.x, most of my tests are now failing with:

```
SyntaxError: Named export '__assign' not found. The requested module 'tslib' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'tslib';
const { __assign } = pkg;
```

It's caused by framer-motion using [tslib](https://github.com/microsoft/tslib). The package does include an [exports field](https://github.com/microsoft/tslib/blob/main/package.json#L29-L36) and module support, so a bit weird why this is popping up. Thinking it's related to how node modules are resolved now after #1673.

### Reproduction

The issue can be replicated by importing `<framer-motion>` in a test file. 
https://stackblitz.com/edit/vitest-dev-vitest-ue9bu2?file=test/App.test.tsx

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    CPU: (4) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Yarn: 1.22.10 - /bin/yarn
    npm: 7.17.0 - /bin/npm
  npmPackages:
    @vitejs/plugin-react: ^1.3.2 => 1.3.2 
    vite: ^2.9.9 => 2.9.14 
    vitest: 0.20.2 => 0.20.2
```

### Used Package Manager

npm

### Validations

### LINKED ISSUE #1776: Mistaken ESM as CJS when using svelte
### Describe the bug

Although `npm run dev/build` works perfectly, `npm run test` throws out an error:

> SyntaxError: Named export 'parse' not found. The requested module 'css-what' is a CommonJS module, which may not support all module.exports as named exports.

After some investigation, the offending code (deep in my dependency chain) turns out to be:

```js
import { parse } from "css-what"
```

It seems that somehow vitest incorrectly considers a perfect es module (`css-what`) as cjs, thus refusing to import it.

---

However, commenting out `svelte` in vite.config can make the error go away in vitest (and obviously breaks my code):

```js
export default defineConfig({
  // plugins: [svelte()],
  test: { globals: true, environment: 'happy-dom' },
})
```

---

Dependency chain:

My code:
```js
import { generateJSON } from '@tiptap/html'
```

`@tiptap/html`:
```js
import { createHTMLDocument, parseHTML } from 'zeed-dom';
```

`zeed-dom`:
```js
import { parse } from "css-what"
```

### Reproduction

https://stackblitz.com/github/andy1li/vitest-bug-report

`npm run dev` to see that vite and svelte work fine:
> The root page should display some json, indicating that the miminal code works.

`npm run test` to reproduce the error msg:

> SyntaxError: Named export 'parse' not found. The requested module 'css-what' is a CommonJS module, which may not support all module.exports as named exports.
> CommonJS modules can always be imported via the default export, for example using:
> 
> import pkg from 'css-what';
> const { parse } = pkg;

### System Info

```shell
Stackblitz:
  System:
    OS: Linux 5.0 undefined
    CPU: (4) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Yarn: 1.22.10 - /bin/yarn
    npm: 7.17.0 - /bin/npm
  npmPackages:
    vite: latest => 3.0.4 
    vitest: latest => 0.20.3 

Mac:
  System:
    OS: macOS 11.6.7
    CPU: (4) x64 Intel(R) Core(TM) i5-6360U CPU @ 2.00GHz
    Memory: 18.48 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.16.0 - ~/.nvm/versions/node/v16.16.0/bin/node
  Browsers:
    Chrome: 103.0.5060.134
    Firefox: 102.0.1
    Safari: 15.5
  npmPackages:
    vite: latest => 3.0.4
    vitest: latest => 0.20.3
```

### Used Package Manager

npm

### Validations

sheremet-va:
With this we can technically externalise every `node_modules` file:

https://github.com/vitest-dev/vitest/blob/5e6a8dab31baa83f142d73c1b60e196d60818a8f/packages/vite-node/src/externalize.ts#L15

This actually **speeds up** tests.

Fixes #1776
Fixes #1753


antfu (approved):

packages/vitest/src/runtime/loader.ts
```diff
@@ -18,15 +27,31 @@ export const resolve: Resolver = async (url, context, next) => {
...
     const resolved = await resolver(id, importer)
     if (resolved) {
-      return {
-        url: pathToFileURL(resolved.id).toString(),
+      const filepath = pathToFileURL(resolved.id).toString()
+      const result: ResolveResult = {
+        url: filepath,
         shortCircuit: true,
       }
+      const source = cache.get(resolved.id)?.source ?? await readFile(resolved.id, 'utf8')
```
antfu, line 35: Maybe we could cache in disk based on package hash (we might be able read it from Vite)

sheremet-va: But why, they are already on disk 👀 Are you afraid it will not update, if I change something inside? If so, it won't update either way, because HMR doesn't work for node_modules 👀

antfu: Oh I didn't notice we need the content 😅

antfu (approved):

PR #1782, opened 2022-08-04
fix: auto-mock based on descriptor list, not property keys (fix #1671)

### LINKED ISSUE #1671: [0.18.1] TypeError: Symbol.hasInstance is read-only

simon-abbott:
These two methods seem like they should be the same, but they're not. Some objects, especially native Node exports, behave strangely; they have properties that are enumerated by `Object.getOwnPropertyNames()` and/or `Object.getOwnPropertySymbols()`, but when you try to get the descriptor for that property using `Object.getOwnPropertyDescriptor()` it returns `undefined`. To combat this, instead of getting the property names / symbols manually, we instead offload that work to the JS engine itself via `Object.getOwnPropertyDescriptors`, and just iterate through the result.

Fixes #1671

### DIFF
--- a/packages/vitest/src/runtime/mocker.ts
+++ b/packages/vitest/src/runtime/mocker.ts
@@ -3,7 +3,7 @@ import { isNodeBuiltin } from 'mlly'
 import { basename, dirname, join, resolve } from 'pathe'
 import { normalizeRequestId, toFilePath } from 'vite-node/utils'
 import type { ModuleCacheMap } from 'vite-node/client'
-import { getAllProperties, getType, getWorkerState, isWindows, mergeSlashes, slash } from '../utils'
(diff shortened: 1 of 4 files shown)


simon-abbott:
Hmm... That didn't seem to work quite as smoothly as I'd hoped... Back to the drawing board I guess!

simon-abbott:
Closed in favor of #1786.

PR #1784, opened 2022-08-04
fix: plugins don't think it's watch mode

### LINKED ISSUE #1003: vitest doesn't work well with custom compiler
### Describe the bug

When using @rollup/plugin-typescript as compiler for tests I'm getting `close timed out after 1000ms` warning after successful run

### Reproduction

https://github.com/wight554/blog-template/pull/30
CI checks can be used for reference

### System Info

```shell
System:
    OS: macOS 12.3
    CPU: (16) x64 Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
    Memory: 2.62 GB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.14.0 - ~/.nvm/versions/node/v16.14.0/bin/node
    Yarn: 3.1.1 - ~/.nvm/versions/node/v16.14.0/bin/yarn
    npm: 8.3.1 - ~/.nvm/versions/node/v16.14.0/bin/npm
    Watchman: 2021.10.04.00 - /usr/local/bin/watchman
  Browsers:
    Chrome: 99.0.4844.83
    Edge: 99.0.1150.46
    Firefox: 98.0.1
    Safari: 15.4
  npmPackages:
    vite: ^2.8.6 => 2.8.6 
    vitest: ^0.7.4 => 0.7.4
```

### Used Package Manager

yarn

### Validations

sheremet-va:
Fixes #1003

### DIFF
--- a/packages/vitest/src/node/plugins/index.ts
+++ b/packages/vitest/src/node/plugins/index.ts
@@ -26,6 +26,9 @@ export async function VitestPlugin(options: UserConfig = {}, ctx = new Vitest())
     <VitePlugin>{
       name: 'vitest',
       enforce: 'pre',
+      options() {
+        this.meta.watchMode = false
+      },
       config(viteConfig: any) {
         // preliminary merge of options to be able to create server options for vite
         // however to allow vitest plugins to modify vitest config values


wight554:
Is it possible to add some custom meta flag e.g. vitestWatch that'll rely on options.watch so plugins can rely on for adding watch specific logic?
Would it work for non-pre plugins when added here https://github.com/vitest-dev/vitest/blob/f027b05525424e943ceb951c4eafb8f49db98832/packages/vitest/src/node/plugins/index.ts#L161 ?

sheremet-va:
> Is it possible to add some custom meta flag e.g. vitestWatch that'll rely on options.watch so plugins can rely on it for adding watch specific logic?
> 
> Would it work for non-pre plugins when added here? https://github.com/vitest-dev/vitest/blob/f027b05525424e943ceb951c4eafb8f49db98832/packages/vitest/src/node/plugins/index.ts#L161

Options are resolved after configResolve, so no it's not possible, and I am tired to answer that question. 🙁

wight554:
> > Is it possible to add some custom meta flag e.g. vitestWatch that'll rely on options.watch so plugins can rely on it for adding watch specific logic?
> > Would it work for non-pre plugins when added here? https://github.com/vitest-dev/vitest/blob/f027b05525424e943ceb951c4eafb8f49db98832/packages/vitest/src/node/plugins/index.ts#L161
> 
> Options are resolved after configResolve, so no it's not possible, and I am tired of answering the same question to you. 🙁
> 
> If for some reason plugin wants to do something vitest-specific, it should use VITEST env variable.

Not talking bout vitest specific, but vitest watch specific, but if api design doesn't allow to expose it in any way, then it might be better to kill watch mode at all
Tbh I'd check for possible performance regressions if esbuild vite logic relies on watch mode before merging it
In my own testing I haven't noticed issues when comparing w/ and w/o watchMode with rollup plugin typescript as transformer

wight554:
Nvm just realized vitest has own watcher instead of relying on vite/transformer, then I see no reason not to merge this one

Issue #1619, opened 2022-07-09
`config.sequence.sequencer`: Type declaration conflicts


AriPerkkio:
### Describe the bug

When creating custom `sequencer` by extending the `BaseSequencer` there are some type errors. I think the root cause is similar as https://github.com/vitest-dev/vitest/issues/991. 

```ts
import { defineConfig } from 'vitest/config';
import { BaseSequencer, Vitest } from 'vitest/node';

export default defineConfig({
    test: {
        sequence: {
            sequencer: class Seqencer extends BaseSequencer {
                protected ctx: Vitest;

                constructor(ctx: Vitest) {
                    super(ctx);
                    this.ctx = ctx;
                }

                async shard(files: string[]) {
                    return files;
                }

                async sort(files: string[]) {
                    return files;
                }
            },
        },
    },
});

// Type 'typeof Seqencer' is not assignable to type 'TestSequencerContructor'.
//  Types of parameters 'ctx' and 'ctx' are incompatible.
//   Type 'import("/workspaces/vitest-sonar-reporter/node_modules/vitest/dist/index").Vitest' is not assignable to type 'import("/workspaces/vitest-sonar-reporter/node_modules/vitest/dist/node").Vitest'.
```

https://vitest.dev/config/#sequence-sequencer 

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-wr6xms?file=vite.config.ts&initialPath=__vitest__

### System Info

```shell
System:
    OS: Linux 5.4 Ubuntu 20.04.4 LTS (Focal Fossa)
    CPU: (4) x64 Intel(R) Xeon(R) Platinum 8272CL CPU @ 2.60GHz
    Memory: 5.27 GB / 7.77 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 16.14.2 - /usr/bin/node
    Yarn: 1.22.15 - /opt/yarn/stable/bin/yarn
    npm: 8.5.0 - /usr/bin/npm
  npmPackages:
    vite: ^2.9.14 => 2.9.14 
    vitest: ^0.17.1 => 0.17.1
```

### Used Package Manager

yarn

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
[...]

PR #1785, opened 2022-08-04
fix: better bundling of types

sheremet-va:
Reduces the bundle size (and time for `build`), since we no longer bundle the same `Vitest` type (and its dependencies).

Before:
```
➜ du -sh dist/                                              
1.5Mb    dist/
```
After:
```
➜ du -sh dist/
1.4Mb    dist/
```

Possible fix for #1619?

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "test:ci": "cross-env CI=true pnpm -r --stream --filter !test-fails --filter !test-browser run test --allowOnly",
     "test:ci:single-thread": "cross-env CI=true pnpm -r --stream --filter !test-fails run test --allowOnly --no-threads",
     "typecheck": "tsc --noEmit",
+    "typecheck:why": "tsc --noEmit --explainFiles > explainTypes.txt",
     "ui:build": "vite build packages/ui",
     "ui:dev": "vite packages/ui",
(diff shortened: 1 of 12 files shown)


sheremet-va:
Looking for a way to fully ignore `dist` folder 🤔

sheremet-va:
> Looking for a way to fully ignore `dist` folder by TypeScript 🤔

I found a wonderful flag `--explainFiles` 🔥

package.json
```diff
@@ -25,6 +25,7 @@
     "test:ci": "cross-env CI=true pnpm -r --stream --filter !test-fails --filter !test-browser run test --allowOnly",
     "test:ci:single-thread": "cross-env CI=true pnpm -r --stream --filter !test-fails run test --allowOnly --no-threads",
     "typecheck": "tsc --noEmit",
+    "typecheck:why": "tsc --noEmit --explainFiles > explainTypes.txt",
```
antfu, line 28: 👀? Cool magic good to know!

sheremet-va: Wanted to keep it in case someone will mess up the types again 👀

antfu (approved):
I don't completely understand what is happening but I trust your judge :P

PR #1790, opened 2022-08-04
feat: use child_process for --no-threads

sheremet-va:
Possible solution for running tests in child_process, if your tests are failing with v8 segfault.

### DIFF
--- a/packages/vitest/rollup.config.js
+++ b/packages/vitest/rollup.config.js
@@ -21,6 +21,7 @@ const entries = [
   'src/node.ts',
   'src/runtime/worker.ts',
   'src/runtime/loader.ts',
+  'src/runtime/child.ts',
   'src/runtime/entry.ts',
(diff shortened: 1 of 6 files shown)


sheremet-va:
@antfu this is just an idea for now, tell me what you think 👀 

Most of the things are not optimized. For example, flow (https://github.com/avajs/ava/blob/main/lib/ipc-flow-control.cjs) and serializaton/deserelization with v8 can be improved.

Also, this processes are firing up really slowly, haven't check why yet 👀

sheremet-va:
The fix for workers already exist for Node 18, the fix for Node 16 will land on August 16th (https://github.com/nodejs/node/pull/44098).

The only problem is Node 14. Its LTS period is ending on April 2023 😞 Although someone can open a PR for staging, explained in https://github.com/nodejs/node/pull/42874#issuecomment-1176581788

PR #1799, opened 2022-08-05
feat(types): better local test context support

### LINKED ISSUE #1798: Types: Better local test context support

I use [test context](https://vitest.dev/guide/test-context.html#extend-test-context) quite often for setting up same fixtures for test. Let's say we have the following test spec code: 

```ts
describe('Test A', () => {
  beforeEach((context) => {
    context.foo = 'foo';
  });
  it('Test A - SubTest A', (context) => {
    expect(context.foo.toUpperCase()).toBe('FOO');
  });
});

describe('Test B', () => {
  beforeEach((context) => {
    context.bar = 'bar';
  });
  it('Test B - SubTest A', (context) => {
    expect(context.bar.toUpperCase()).toBe('BAR');
  });
});

```

According to the official docs, We need to aggregate the `TestContext` type:

```ts
declare module 'vitest' {
  export interface TestContext {
    foo?: string;
    bar?: string;
  }
}
```

This introduces some problems, which are caused by the fact that different tests have different context but we cannot do a "local aggregation", so the `TestContext` must be the supertype of all contexts. Those problems are:

* Introducing unnecessary optional properties. I have to check those optional properties in each test, which brings unnecessary code for assertions (`!`), nullish operators (`?.`), or type guards. If I made them required, they become required for all tests which may cause typos (e.g. it is legal to write `context.foo` in "Test B - SubTest A" but we don't set it up yet)
* Type interference.  What if we have two `context.foo` in different tests but with different types? We could write `context.testA.foo` but it seems ugly.

So my question here is that is it possible to declare "local test context" rather than a global context?

As I search the typings, there's a `ExtraContext` passing around. But the `test` has a fixed `{}` as `ExtraContext` so it won't works.

### Suggested solution

There could be a solution for add some TypeScript helpers, without changing the underlying js code while keep it compatible for global context as following. [(playground)](https://www.typescriptlang.org/play?#code/C4TwDgpgBAwgFgQwJYDsECMA2EBiBXFAY2CQHsUAeAFSggA9gIUATAZyleACdUBzAGig4AfFAC8QqADIoAbwBQUKAG0A1hBBRUUKgF0AXLESoM2fERLlqgkfIC+8+akZcAZgkLQqETjHKMGOShXUlJDFDwAW3QILigHUEgoAEEAd2RgUwhqUQkaAB8oAAUuUkikVggAGSR1HIBuR0SvH2BzYjJKAFEGLgQ-FADgcTk7XKgACkJ-emBDb18ZwJke7n6l4ABKcVE0jKyKBBQQUUKAN1IkZkbmo2Q0LAgF4GSigEkKVb6BoZHZMZG8HuWXalkoAHJpkQ8FwuExgOCoIVweRMCBEcjWKokGAMVBwcBSMxSHjwe4kJhWODBIooNRaAwmGwoF91oNZiNWT9ZsIJmhIhB9JweCgBMEUAB+eatUGdHKCEgC0h4YBSqARaKxbZiUQXK7yYSOZyxdyeKBVUiEBCYZ7cwKyKDoBBcQzCvjxRwAek9OlarzeUFSfTA7CBJkez39gjKSGGcFiEHkzAghEwzugUM4UEYnEMYYe2Ej7wo-0N8hzwAmBNaUChQ2pkzrs21ogUSluduAVHA0AkzVIrlrG0aSm9vsW7IY9k2TVaFAtVptrU7vOrWabDAbUw2LbktI7G27ST7PYHQ8nwBHUDHto20nNlutt4v08c5daADpCcSP1icfPHyXCchlXCtz3rQRtwvXc22zHtYEPeCT0gM8N0vWkb2XO8ZAXJ8sJfOwZyAA)

```ts
type ChainableFunction<T extends string, F> = F & {
  [key in T]: ChainableFunction<T, F>
}

interface TestContext { foo: number }
type Awaitable<T> = T | PromiseLike<T>;

type TestFunction<ExtraContext = {}> = (context: TestContext & ExtraContext) => Awaitable<any> | void;
type ChainableTestAPI<ExtraContext = {}> = ChainableFunction<'concurrent' | 'only' | 'skip' | 'todo' | 'fails', 
  <T extends ExtraContext = ExtraContext>(name:string, fn?: TestFunction<T>, timeout?: number) => void
>

interface LocalTestContext { bar: string }

// TestAPI wraps ChainableTestAPI, omit here
declare const test: ChainableTestAPI<{}>

test('test context', (context) => {
  type ContextType = typeof context;
  // TestContext
})

test<LocalTestContext>('test context', (context) => {
  type ContextType = typeof context;
  // TestContext & LocalTestContext
})


test.todo.skip<LocalTestContext>('test context', (context) => {
  type ContextType = typeof context;
  // TestContext & LocalTestContext
})

```

The story is the same for `beforeEach` and `afterEach`.

### Alternative

Alternatively, I have to use type casts to shim the `LocalTestContext` to `{}`.

```ts
// extract BeforeEach hooks
type BeforeEachFunction<T = unknown> = Parameters<typeof beforeEach>[0] extends (
  context: infer C,
  ...args: infer P
) => infer R
  ? (context: C & T, ...args: P) => R
  : never;

describe('Test A', () => {
  interface LocalTestContext {
    foo: string;
  }
  beforeEach(((context) => {
    context.foo = 'foo';
  }) as BeforeEachFunction<LocalTestContext> as BeforeEachFunction);
  it('Test A - SubTest A', ((context) => {
    expect(context.foo.toUpperCase()).toBe('FOO');
  }) as TestFunction<LocalTestContext> as TestFunction);
});

describe('Test B', () => {
  interface LocalTestContext {
    bar: string;
  }
  beforeEach(((context) => {
    context.bar = 'bar';
  }) as BeforeEachFunction<LocalTestContext> as BeforeEachFunction);
  it('Test B - SubTest A', ((context) => {
    expect(context.bar.toUpperCase()).toBe('BAR');
  }) as TestFunction<LocalTestContext> as TestFunction);
});
```

Although the type of `context` is correct (`TestContext & LocalTestContext`) here, it seems less elegant than the proposed apporach.

### Additional context

_No response_

### Validations

Tanimodori:
fix #1798.

This pr only changes types, no underlying JavaScript codes are altered.

This pr passed build and all tests on my local machine.

### DIFF
--- a/packages/vitest/src/runtime/chain.ts
+++ b/packages/vitest/src/runtime/chain.ts
@@ -1,15 +1,13 @@
-export type ChainableFunction<T extends string, Args extends any[], R = any> = {
-  (...args: Args): R
-} & {
-  [x in T]: ChainableFunction<T, Args, R>
+export type ChainableFunction<T extends string, F extends (...args: any[]) => any> = F & {
+  [x in T]: ChainableFunction<T, F>
 }
 
-export function createChainable<T extends string, Args extends any[], R = any>(
+export function createChainable<T extends string, F extends (this: Record<T, boolean | undefined>, ...args: any[]) => any>(
   keys: T[],
-  fn: (this: Record<T, boolean | undefined>, ...args: Args) => R,
-): ChainableFunction<T, Args, R> {
+  fn: F,
+): ChainableFunction<T, ((...args: Parameters<F>) => ReturnType<F>)> {
   function create(obj: Record<T, boolean | undefined>) {
-    const chain = function (this: any, ...args: Args) {
+    const chain = function (this: any, ...args: Parameters<F>) {
       return fn.apply(obj, args)
     }
     for (const key of keys) {
--- a/packages/vitest/src/runtime/hooks.ts
+++ b/packages/vitest/src/runtime/hooks.ts
@@ -5,5 +5,5 @@ import { getCurrentSuite } from './suite'
 // suite hooks
 export const beforeAll = (fn: SuiteHooks['beforeAll'][0], timeout?: number) => getCurrentSuite().on('beforeAll', withTimeout(fn, timeout ?? getDefaultHookTimeout(), true))
 export const afterAll = (fn: SuiteHooks['afterAll'][0], timeout?: number) => getCurrentSuite().on('afterAll', withTimeout(fn, timeout ?? getDefaultHookTimeout(), true))
-export const beforeEach = (fn: SuiteHooks['beforeEach'][0], timeout?: number) => getCurrentSuite().on('beforeEach', withTimeout(fn, timeout ?? getDefaultHookTimeout(), true))
-export const afterEach = (fn: SuiteHooks['afterEach'][0], timeout?: number) => getCurrentSuite().on('afterEach', withTimeout(fn, timeout ?? getDefaultHookTimeout(), true))
+export const beforeEach = <ExtraContext = {}>(fn: SuiteHooks<ExtraContext>['beforeEach'][0], timeout?: number) => getCurrentSuite<ExtraContext>().on('beforeEach', withTimeout(fn, timeout ?? getDefaultHookTimeout(), true))
+export const afterEach = <ExtraContext = {}>(fn: SuiteHooks<ExtraContext>['afterEach'][0], timeout?: number) => getCurrentSuite<ExtraContext>().on('afterEach', withTimeout(fn, timeout ?? getDefaultHookTimeout(), true))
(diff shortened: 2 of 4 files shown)


sheremet-va (changes requested):
Can you add more tests with examples please? Other then that LGTM 👍🏻

Tanimodori:
Thanks, I've found a better approach without changing the `ChainableFunction` signature (since it is exported and may be troublesome for dependencies). I also add a unit test. I'll open another pr soon.

PR #1803, opened 2022-08-05
feat!: if not processed, CSS Modules return a proxy, scope class names by filename

### LINKED ISSUE #1512: Automatically mock css module imports

When using snapshots, any change to a ``module.(css|scss)`` file will change the generated class.
image
This creates a massive amount of noise in commits and pr. This random class name behavior is not desirable in a testing context, where you want predictability.

I would be open to submitting a PR for this issue if that was desired.

### Suggested solution

The object imported from a ``module.(css|scss)`` file could be mocked to:
```js
new Proxy(new Object(), {
  get(_, style) {
    return style;
  },
});
```
This means that when a class is accessed, it returns the raw string, reproducably:
```js
import styles from "./Renderer.module.scss";
console.log(styles.underline) // "underline"
```
Instead of the occasionally randomized variant:
```js
import styles from "./Renderer.module.scss";
console.log(styles.underline) // "_underline_i69w2_32" (until someone modifies Renderer.module.scss)
```

### Alternative

This can be solved in each test file, by just mocking the module.
```js
vi.mock("./Renderer.module.scss", () => ({
  default: new Proxy(new Object(), {
    get(_, style) {
      return style;
    },
  }),
}));
```
This is perfectly workable, but there is no reason a developer should need to do this - returning the same string every time is expected behavior in my opinion.

### Additional context

Diffs where every class had its name changed (but nothing was actually changed) are massive, and can obscure actual changes - they suck. For example, if a pr includes a change to a css file *and* a modification to the classes of an object, the second change will be hidden behind the noise of the first, reducing the utility of the snapshot as a change detector.

### Validations

sheremet-va:
Closes #1512

This PR implements the following logic:

- By default, CSS processing is disabled for every css file
  - `module.*` exports a proxy in that case to not affect runtime (meaning, `styles.{value}` will always return a key with filename hash)
- If processing is enabled (`css.include`), css is processed and injected into DOM
  - By default, `module.*` returns pseudo-scoped class names (`styles.module` always returns `_module_${filenameHash}`, `styles.someRandomValue` will return `undefined`)
  - If you want to enable original scoped names (so production and tests generate the same names), use `css.modules.classNameStrategy: 'scoped'`
- Added `css.modules.classNameStrategy` option:
  - `scoped` - use `css.modules.generateScopeName` from user config or fallback to default behaviour, this is how it works in dev and prod
  - `non-scoped` - returns just the name
  - `stable` - returns the nama and hash, based on filepath, relative to root

### DIFF
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -706,14 +706,12 @@ Show heap usage after each test. Useful for debugging memory leaks.
 
 - **Type**: `boolean | { include?, exclude? }`
 
-Configure if CSS should be processed. When excluded, CSS files will be replaced with empty strings to bypass the subsequent processing.
-
-By default, processes only CSS Modules, because it affects runtime. JSDOM and Happy DOM don't fully support injecting CSS, so disabling this setting might help with performance.
+Configure if CSS should be processed. When excluded, CSS files will be replaced with empty strings to bypass the subsequent processing. CSS Modules will return a proxy to not affect runtime.
 
 #### css.include
 
 - **Type**: `RegExp | RegExp[]`
-- **Default**: `[/\.module\./]`
+- **Default**: `[]`
 
 RegExp pattern for files that should return actual CSS and will be processed by Vite pipeline.
 
@@ -723,6 +721,33 @@ RegExp pattern for files that should return actual CSS and will be processed by 
 - **Default**: `[]`
 
 RegExp pattern for files that will return an empty CSS file.
+
+#### css.modules
+
+- **Type**: `{ scopeClassNames? }`
+- **Default**: `{}`
+
+#### css.modules.scopeClassNames
+
+- **Type**: `boolean`
+- **Default**: false
+
+If you decide to process CSS files, you can configure if class names inside CSS modules should be scoped. By default, Vitest exports a proxy, bypassing CSS Modules processing.
+
+You might want to enable this, if your CSS classes are conflicting with each other, when CSS is inlined. For example, when you are accessing computed styles:
+
+```tsx
+// global.module.css
+// .error { width: 600px }
+
+// element.module.css
+// .error { width: 100px }
+
+// test
+const styles = window.getComputedStyles(<div className={error}></div>)
+// it's possible to have two different classes with conflicting styles
+expect(styles).toMatchObject({ with: '100px' })
+```
 
 ### maxConcurrency
 
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -745,7 +745,7 @@ You might want to enable this, if your CSS classes are conflicting with each oth
 
 // test
 const styles = window.getComputedStyles(<div className={error}></div>)
-// it's possible to have two different classes with conflicting styles
+// this will fail, if global css is loaded after element
 expect(styles).toMatchObject({ with: '100px' })
 ```
 
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -724,30 +724,19 @@ RegExp pattern for files that will return an empty CSS file.
 
 #### css.modules
 
-- **Type**: `{ scopeClassNames? }`
+- **Type**: `{ classNamesStrategy? }`
 - **Default**: `{}`
 
-#### css.modules.scopeClassNames
+#### css.modules.classNamesStrategy
 
-- **Type**: `boolean`
-- **Default**: false
+- **Type**: `'stable' | 'scoped' | 'non-scoped'`
+- **Default**: `'stable'`
 
-If you decide to process CSS files, you can configure if class names inside CSS modules should be scoped. By default, Vitest exports a proxy, bypassing CSS Modules processing.
+If you decide to process CSS files, you can configure if class names inside CSS modules should be scoped. By default, Vitest exports a proxy, bypassing CSS Modules processing. You can choose one of the options:
 
-You might want to enable this, if your CSS classes are conflicting with each other, when CSS is inlined. For example, when you are accessing computed styles:
-
-```tsx
-// global.module.css
-// .error { width: 600px }
-
-// element.module.css
-// .error { width: 100px }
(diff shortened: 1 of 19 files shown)


docs/config/index.md
```diff
@@ -697,6 +695,33 @@ RegExp pattern for files that should return actual CSS and will be processed by
...
+// global.module.css
+// .error { width: 600px }
+
+// element.module.css
+// .error { width: 100px }
+
+// test
+const styles = window.getComputedStyles(<div className={error}></div>)
+// this will fail, if global css is loaded after element
+expect(styles).toMatchObject({ with: '100px' })
```
Demivan, line 722:
```suggestion
expect(styles).toMatchObject({ width: '100px' })
```

Demivan (changes requested):

Demivan:
How about making `generateScopedName` depend on classname and filename? This should should prevent class name conflict issues and make it so generated class names don't change when css content changes.

This is what `postcss-modules` uses in their tests.
https://github.com/madyankin/postcss-modules/blob/a400e145e0ae9e996e51eed81a2e8e2a1f604527/test/test.js#L31-L34

sheremet-va:
> How about making `generateScopedName` depend on classname and filename? This should should prevent class name conflict issues and make it so generated class names don't change when css content changes.
> 
> This is what `postcss-modules` uses in their tests. [madyankin/postcss-modules@`a400e14`/test/test.js#L31-L34](https://github.com/madyankin/postcss-modules/blob/a400e145e0ae9e996e51eed81a2e8e2a1f604527/test/test.js#L31-L34)

They are using `basename`, so it can still be incompatible, if you have `/foo/index.css` and `/baz/index.css` for example. We can encode the whole path to fix this, of course. @isaec, what do you think?

Demivan:
> This also changes semantics and requires different naming.

Maybe enum `css.modules.classNames`:

- `stable` - (default) class name + hashed path
- `scoped` - original implementation
- `raw`?, `non-scoped`?, `source`? - source class names

isaec:
> > How about making `generateScopedName` depend on classname and filename? This should should prevent class name conflict issues and make it so generated class names don't change when css content changes.
> > This is what `postcss-modules` uses in their tests. [madyankin/postcss-modules@`a400e14`/test/test.js#L31-L34](https://github.com/madyankin/postcss-modules/blob/a400e145e0ae9e996e51eed81a2e8e2a1f604527/test/test.js#L31-L34)
> 
> They are using `basename`, so it can still cause conflicts, if you have `/foo/index.css` and `/baz/index.css` for example. We can encode the whole path to fix this, of course, generating `_${className}_${encodedFilepath}` (`_main_32rsfdr3`, it won't change when content changes, but will change, if filename is changed, or file is moved to other directory) @isaec, what do you think?
> 
> This also changes semantics and requires different naming.

For my uses, the stability was the desired characteristic. If importing the stylesheet has the correct property string mapping, then ``stable`` or ``source`` both work for me. Stable seems like a good default, still discouraging selector matching for tests (unless you import the style object), but not providing unneeded noise in snapshot testing.

sheremet-va:
Option `scopeClassNames` is removed, option `classNameStrategy` is added with options:

- `scoped` - use `css.modules.generateScopeName` from user config or fallback to default behaviour, this is how it works in dev and prod
- `non-scoped` - returns just the name
- `stable` - returns the nama and hash, based on filepath, relative to root

packages/vitest/src/integrations/css/css-modules.ts
```diff
@@ -0,0 +1,19 @@
+import type { CSSModuleScopeStrategy } from '../../types'
+
+export function generateCssFilenameHash(filename: string) {
+  return Buffer.from(filename).toString('base64').substring(0, 6)
```
Demivan, line 4: This is probably a bad hash function as first characters will be the same for different files in the same folder.

Demivan, line 4: This is not a good hash function as different files in the same folder will get the same hash

Demivan (changes requested):

packages/vitest/src/integrations/css/css-modules.ts
sheremet-va: What do you propose instead?

antfu:
I am not familiar with CSS modules, so I'd trust your judgment here. Feel free to merge once you meet the agreement :)

packages/vitest/src/integrations/css/css-modules.ts
Demivan: Something like md5 that uses entire length of the string

sheremet-va: We still need to truncate it though to 5-6 chars, otherwise the length might be too big

Demivan: Problem is that base64 is not a hash:

`Buffer.from(filename).toString('base64').substring(0, 6)`:

`folder/file1.module.css` -> `Zm9sZG`  
`folder/file2.module.css` -> `Zm9sZG`

`md5(filename).substring(0, 6)`

`folder/file1.module.css` -> `9d8841`   
`folder/file2.module.css` -> `34c715`

postcss-modules uses [string-hash](https://github.com/darkskyapp/string-hash/blob/master/index.js)

Demivan, line 4: We need to normalize path separators so snapshots created on linux will still work on windows

sheremet-va: We are using `relative` from pathe which normilizes it already (you can see out Windows tests are passing)

sheremet-va: updated to use md5

Demivan (approved):

PR #1812, opened 2022-08-06
 fix: .only annotation is ignored (fix #1811)

### LINKED ISSUE #1811: `.only` annotation is ignored
### Describe the bug

`.only` annotations are ignored and all tests are executed all the time. 

Also the API reference does not mention that `allowOnly` requires `threads` to be `false`.

### Reproduction

Sorry, I have no reproduction repository at hand but I use the following vitest.config.ts:

```
// <reference types="vitest" />
import { resolve } from 'path';

export default defineConfig({
  test: {
    environment: 'happy-dom',
    alias: {
      // as soon as we use an alias for '@', allowOnly stops working
      '@': `${resolve(__dirname, 'src')}`,
      vue: 'vue/dist/vue.esm-bundler.js',
    },
    globals: true,
    allowOnly: true,
    threads: false,
    // silent: true,
  },
});
```

### System Info

```shell
System:
    OS: macOS 12.2
  Binaries:
    Node: 18.1.0 - ~/.nvm/versions/node/v18.1.0/bin/node
    npm: 8.8.0 - ~/.nvm/versions/node/v18.1.0/bin/npm
  npmPackages:
    vite: ^2.9.9 => 2.9.9
    vitest: ^0.21.0 => 0.21.0
```

### Used Package Manager

npm

### Validations

softwarerero:
Also add a hint to the API reference that `threat` needs to be `false` when `allowOnly` is used.

### DIFF
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -545,7 +545,7 @@ export default defineConfig({
 - **Type**: `boolean`
 - **Default**: `false`
 
-Allow tests and suites that are marked as only.
+Allow tests and suites that are marked as only. This also requires to set `threads` to `false`.
 
 ### dangerouslyIgnoreUnhandledErrors
 
--- a/packages/vitest/src/runtime/collect.ts
+++ b/packages/vitest/src/runtime/collect.ts
@@ -22,6 +22,7 @@ function hash(str: string): string {
 
 export async function collectTests(paths: string[], config: ResolvedConfig) {
   const files: File[] = []
+  const filesWithOnlyTasks = []
 
   const browserHashMap = getWorkerState().browserHashMap!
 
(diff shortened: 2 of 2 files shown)


docs/config/index.md
```diff
@@ -545,7 +545,7 @@ export default defineConfig({
 - **Type**: `boolean`
 - **Default**: `false`
 
-Allow tests and suites that are marked as only.
+Allow tests and suites that are marked as only. This also requires to set `threads` to `false`.
```
sheremet-va, line 548: First of all, I think you misinterpret what this flag does. 

Usually, if you have `.only` test and run it in CI, you tests will fail. This flag just allows your tests to not fail in this situation.

packages/vitest/src/runtime/collect.ts
```diff
@@ -91,9 +92,11 @@ export async function collectTests(paths: string[], config: ResolvedConfig) {
     interpretTaskModes(file, config.testNamePattern, hasOnlyTasks, false, config.allowOnly)
 
     files.push(file)
+    if (hasOnlyTasks)
+      filesWithOnlyTasks.push(file)
   }
 
-  return files
+  return config.allowOnly ? filesWithOnlyTasks : files
```
sheremet-va, line 99: This won't work, because this is not what `allowOnly` flag does.

sheremet-va (changes requested):

PR #1818, opened 2022-08-07
fix: Add error message when mock is missing export (fix #1763)

### LINKED ISSUE #1763: Add the required "default" export mocking caveat to Jest migration guide.

It took me a while to realise that I needed to explicitly mock default module exports in vitest. Since this is a rather big change from the way we mock modules in jest, I think it's worth to mention this in the Jest migration guide.

This is well documented in https://vitest.dev/api/#vi-mock, but is easy to miss as it's not a part of the general mocking guide.

### Suggested solution

Add a quick note about default export mocking in https://vitest.dev/guide/migration.html#migrating-from-jest. I think this will help people alot.

### Alternative

Provide better error messages when people accidentally mock named exports with `vi.mock`. For example `cannot read property 'get' of undefined, did you intend to mock the "default" export? See https://vitest.dev/api/#vi-mock for details.` 

### Additional context

_No response_

### Validations

jereklas:


jereklas:
missed testing. will re-open when fixed.

PR #1819, opened 2022-08-08
fix: add error message when mock is missing export

jereklas:


packages/vitest/src/runtime/mocker.ts
```diff
@@ -335,7 +335,38 @@ export class VitestMocker {
...
+      apply(target: () => Object) {
+        return new Proxy(target(), exportHandler)
+      },
+    }
+
+    return new Proxy(factory, factoryHandler) as () => unknown
+  }
+
   public queueMock(id: string, importer: string, factory?: () => unknown) {
+    factory = this.wrapFactoryInProxy(id, importer, factory)
```
sheremet-va, line 369: Can't we just wrap exports here:

https://github.com/jereklas/vitest/blob/9d5234ef74e092a902314d2baac7cfbdd6cb2729/packages/vitest/src/runtime/mocker.ts#L109

?

sheremet-va (changes requested):

packages/vitest/src/runtime/mocker.ts
sheremet-va: And while we are here, let's throw an error, if factory returns non-object. (With a helper that it should return `{ default }`, if user wants to mock default export)

jereklas: I split the proxy objects. The export object is now happening where you suggested, and the factory proxy object is now returning an error if the return value is not an object.

packages/vitest/src/runtime/mocker.ts
```diff
@@ -336,6 +354,16 @@ export class VitestMocker {
...
 
   public queueMock(id: string, importer: string, factory?: () => unknown) {
+    const factoryHandler = {
+      apply(target: () => unknown) {
+        const value = target()
+        if (typeof value !== 'object')
+          throw new Error(`[vitest] vi.mock("${id}", factory?: () => unknown) is not returning an object. Did you mean to return an object with a "default" key?`)
+        return value
+      },
+    }
```
sheremet-va, line 364: Why do we need a proxy, if we can just check it here:
https://github.com/jereklas/vitest/blob/bdd74e50836a70d9aae410781f5c6fdbe5093f3a/packages/vitest/src/runtime/mocker.ts#L109
?

sheremet-va (changes requested):

packages/vitest/src/runtime/mocker.ts
```diff
@@ -336,6 +354,16 @@ export class VitestMocker {
   }
 
   public queueMock(id: string, importer: string, factory?: () => unknown) {
+    const factoryHandler = {
+      apply(target: () => unknown) {
+        const value = target()
+        if (typeof value !== 'object')
```
sheremet-va, line 360: Should also check `null`, i think?

packages/vitest/src/runtime/mocker.ts
jereklas: You need the proxy on the exports object because you don't know which export is being accessed until the calling code imports something?

I need to head to bed, but will try taking a look at your factory related concern tomorrow.

sheremet-va: We do need a roxy for exports object, but I don't see why we need proxy for calling a factory - we are calling it ourselves on https://github.com/jereklas/vitest/blob/bdd74e50836a70d9aae410781f5c6fdbe5093f3a/packages/vitest/src/runtime/mocker.ts#L109

jereklas: Ah gotcha. I misunderstood your statement.

sheremet-va (approved):

jereklas:
Looks like a test unrelated to my changes failed in CI? I ran them locally without any issue. What's the process for getting tests re-ran?

alfiomy:
Good

PR #1826, opened 2022-08-09
docs: improve (describe|test).only description

Dragomir-Ivanov:
Hi,
Here is initial version of the Docs improvement. Sorry for the delay.

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -102,7 +102,7 @@ In Jest, `TestFunction` can also be of type `(done: DoneCallback) => void`. If t
   })
   ```
 
-  Sometimes it is very usefull to run `only` tests, in a certain file ignoring all other tests from the whole test suite, which polute the output.
+  Sometimes it is very useful to run `only` tests in a certain file, ignoring all other tests from the whole test suite, which pollute the output.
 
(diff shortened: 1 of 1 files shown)


docs/api/index.md
```diff
@@ -102,6 +102,13 @@ In Jest, `TestFunction` can also be of type `(done: DoneCallback) => void`. If t
   })
   ```
 
+  Sometimes it is very usefull to run `only` tests, in a certain file ignoring all other tests from the whole test suite, which polute the output.
```
sheremet-va, line 105:
```suggestion
  Sometimes it is very useful to run `only` tests in a certain file, ignoring all other tests from the whole test suite, which pollute the output.
```

docs/api/index.md
```diff
@@ -290,6 +297,13 @@ When you use `test` in the top level of file, they are collected as part of the
...
```
sheremet-va, line 300:
```suggestion
  Sometimes it is very useful to run `only` tests in a certain file, ignoring all other tests from the whole test suite, which pollute the output.
```

sheremet-va (changes requested):

PR #1805, opened 2022-08-05
feat(types): better local test context support

Tanimodori:
The second try for fixing #1798
Previous pr #1799
Related issue: #1660

PR #1829, opened 2022-08-10
docs: clarify how to use local test context

Tanimodori:
Add docs for how local test context works introduced in PR #1805, which is available since Vitest v0.21.1

### DIFF
--- a/docs/guide/test-context.md
+++ b/docs/guide/test-context.md
@@ -58,7 +58,7 @@ declare module 'vitest' {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


docs/guide/test-context.md
```diff
@@ -58,3 +58,20 @@ declare module 'vitest' {
 }
 ```
 
+Since Vitest v0.21.1, if you want to provide property types only for specific `beforeEach`, `afterEach`, `it` and `test` hooks, you can provide an extra context type as type parameter.
```
antfu, line 61:
```suggestion
If you want to provide property types only for specific `beforeEach`, `afterEach`, `it` and `test` hooks, you can pass the type as a generic.
```

PR #1830, opened 2022-08-10
fix: handle undefined returns of module mocks, and update migration docs (#1763)

jereklas:
This should wrap up the loose ends on https://github.com/vitest-dev/vitest/issues/1763, so that it can be closed out.

### DIFF
--- a/docs/guide/migration.md
+++ b/docs/guide/migration.md
@@ -14,6 +14,19 @@ Jest has their [globals API](https://jestjs.io/docs/api) enabled by default. Vit
 
 If you decide to keep globals disabled, be aware that common libraries like [`testing-library`](https://testing-library.com/) will not run auto DOM [cleanup](https://testing-library.com/docs/svelte-testing-library/api/#cleanup).
 
+**Module mocks**
+
+When mocking a module in Jest, if the factory argument returned a primitive, the mock would implicitly mock the default export. In Vitest, the factory argument has to return an object where each export is explicitly defined. For example, the following `jest.mock` would have to be updated as follows:
+
+```diff
+- jest.mock('./some-path', () => 'hello')
++ vi.mock('./some-path', () => ({
++   default: 'hello';
++ })
+```
+
+For more details please refer to the [vi.mock api](/api/#vi-mock)
+
 **Auto-Mocking Behaviour**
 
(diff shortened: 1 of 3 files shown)


jereklas:
Looks like a flaky test unrelated to my change caused one of the builds to fail.

docs/guide/migration.md
```diff
@@ -14,6 +14,19 @@ Jest has their [globals API](https://jestjs.io/docs/api) enabled by default. Vit
 
 If you decide to keep globals disabled, be aware that common libraries like [`testing-library`](https://testing-library.com/) will not run auto DOM [cleanup](https://testing-library.com/docs/svelte-testing-library/api/#cleanup).
 
+**Module mocks**
+
+When mocking a module in Jest, if the factory argument returned a primitive, the mock would implicitly mock the default export. In Vitest, the factory argument has to return an object where each export is explicitly defined. For example, the following `jest.mock` would have to be updated as follows:
```
sheremet-va, line 19: I don't think it has to be a primitive. Everything returned from a mock was available on `default`, since it is default CJS behaviour, no?

sheremet-va (changes requested):

docs/guide/migration.md
jereklas: I'm not actually sure, I assumed looking at the jest docs that a factory returning an object would result in named exports, and or require default explicitly be defined (if you were wanting default?) I can play with jest to better understand.

jereklas: You are indeed correct, will post an update shortly

sheremet-va (approved):

docs/guide/migration.md
```diff
@@ -14,6 +14,19 @@ Jest has their [globals API](https://jestjs.io/docs/api) enabled by default. Vit
 
 If you decide to keep globals disabled, be aware that common libraries like [`testing-library`](https://testing-library.com/) will not run auto DOM [cleanup](https://testing-library.com/docs/svelte-testing-library/api/#cleanup).
 
+**Module mocks**
+
+When mocking a module in Jest, the factory argument's return value is the default export. In Vitest, the factory argument has to return an object with each export explicitly defined. For example, the following `jest.mock` would have to be updated as follows:
```
vilbergs, line 19: Nicely articulated! 👍

PR #1842, opened 2022-08-12
fix(reporter-junit): remove trailing zeros in duration

### LINKED ISSUE #1825: Keep time attribute in JUnit XML file
### Describe the bug

Currently in junit reporter, when `task.result.duration`'s value is 0, the duration got undefined as fallback value by JavaScript's implicit conversion, which results in that generated report file will not got `time` attribute in `<testcase>` tag.

```typescript
function getDuration(task: Task): string | undefined {
  return task.result?.duration ? (task.result.duration / 1000).toFixed(10) : undefined
}
```

The produced xml file:

```xml
<testsuites id="" name="" tests="1" failures="0" skipped="0" errors="0" time="0.014961999997496605">
  <testsuite name="x" timestamp="2022-08-09T09:17:24.527Z" hostname="" tests="5" failures="0" errors="0" skipped="0" time="0.0070000000">
    <!-- no time attribute was present -->
    <testcase classname="x" name="should xxx">
    </testcase>
  </testsuite>
</testsuites>
```

In junit standard, time attribute should be kept even using value '0'. Also, this will cause some tools work incorrectly.

Use 0 as fallback value should fix this:

```typescript
function getDuration(task: Task): string {
  return typeof task.result?.duration === 'number' ? task.result.duration === 0 ? '0' : (task.result.duration / 1000).toFixed(10) : '0'
}
```

### Reproduction

I'm not sure why the `task.result.duration` reveived value 0, but in some simple test cases(e.g. only one expect statement) it always appear, so I suspect that this is also a precision loss problem caused by JavaScript.

### System Info

```shell
System:
    OS: macOS 12.3.1
    CPU: (8) arm64 Apple M1
    Memory: 90.95 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.17.4 - ~/.nvm/versions/node/v14.17.4/bin/node
    Yarn: 1.22.18 - ~/.yarn/bin/yarn
    npm: 8.10.0 - ~/.nvm/versions/node/v14.17.4/bin/npm
  Browsers:
    Chrome: 104.0.5112.79
    Edge: 104.0.1293.47
    Firefox: 89.0.2
    Safari: 15.4
```

### Used Package Manager

npm

### Validations

nieyuyao:
fix #1825

### DIFF
--- a/packages/vitest/src/node/reporters/junit.ts
+++ b/packages/vitest/src/node/reporters/junit.ts
@@ -60,7 +60,7 @@ function escapeXML(value: any): string {
 
 export function getDuration(task: Task): string | undefined {
   const duration = task.result?.duration ?? 0
-  return (duration / 1000).toFixed(10).replace(/^(\d+)((?:\.0+)|(\.\d+?)0+)$/, '$1$3')
+  return (duration / 1000).toLocaleString(undefined, { useGrouping: false, maximumFractionDigits: 10 })
 }
 
 export class JUnitReporter implements Reporter {
--- a/packages/vitest/src/node/reporters/junit.ts
+++ b/packages/vitest/src/node/reporters/junit.ts
@@ -58,8 +58,9 @@ function escapeXML(value: any): string {
     true)
 }
 
-function getDuration(task: Task): string | undefined {
-  return task.result?.duration ? (task.result.duration / 1000).toFixed(10) : undefined
+export function getDuration(task: Task): string | undefined {
+  const duration = task.result?.duration ?? 0
+  return (duration / 1000).toFixed(10).replace(/^(\d+)((?:\.0+)|(\.\d+?)0+)$/, '$1$3')
 }
 
 export class JUnitReporter implements Reporter {
--- a/test/reporters/tests/__snapshots__/reporters.spec.ts.snap
+++ b/test/reporters/tests/__snapshots__/reporters.spec.ts.snap
@@ -3,32 +3,32 @@
 exports[`JUnit reporter (no outputFile entry) 1`] = `
 "<?xml version=\\"1.0\\" encoding=\\"UTF-8\\" ?>
 <testsuites>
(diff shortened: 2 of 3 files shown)


packages/vitest/src/node/reporters/junit.ts
```diff
@@ -58,8 +58,9 @@ function escapeXML(value: any): string {
     true)
 }
 
-function getDuration(task: Task): string | undefined {
-  return task.result?.duration ? (task.result.duration / 1000).toFixed(10) : undefined
+export function getDuration(task: Task): string | undefined {
+  const duration = task.result?.duration ?? 0
+  return (duration / 1000).toFixed(10).replace(/^(\d+)((?:\.0+)|(\.\d+?)0+)$/, '$1$3')
```
Demivan, line 63: You can `parseFloat()` or cast to number to remove trailing zeroes. Using regex seems like an overkill for this

nieyuyao: Using `parseFloat` will result in exponential values.

antfu: I guess we could use `Math.round(duration * 10e10) / 10e10` instead.

Demivan: Then lets keep it as a regex

Demivan (approved):

packages/vitest/src/node/reporters/junit.ts
Demivan: `String(Math.round(0.0000000000100 * 10e10) / 10e10)` this returns `"1e-11"`

antfu: This might help? https://stackoverflow.com/a/50978675

Demivan: `0.0000000100.toLocaleString(undefined, {useGrouping:false, maximumFractionDigits:10 })` - `"0.00000001"` looks good

antfu (approved):

PR #1843, opened 2022-08-12
fix: trim URL parameters from file paths in coverage

OrestHk:


sheremet-va:
I think PR #1841 tries to do the same, but in a different way? @ChpShy

OrestHk:
Hello, I've tried https://github.com/vitest-dev/vitest/pull/1841, the problem is that it will indead remove the error, but also remove the .ts file from the coverage result list, keeping only the .vue file in the result
This is only a draft ATM, but the following seems to allow to keep both ts and vue file in the coverage result, with the creation of the HTML file ine the html reporter

ChpShy:
Yes, you are right. This is not `v0.21.1`'s bug, to be precise, this is `v0.16.x`'s bug.In `v0.16.0`, `testing.ts` is skipped. And I  misunderstood it.

sheremet-va:
Coverage is now implemented in different packages. @OrestHk do you think the fix is needed, or we can close this PR?

OrestHk:
Tried with vitest 0.24.1 and @vitest/coverage-c8 the problem is still here
I'll close this PR and look into @vitest/coverage-c8 to see what can be done !

OrestHk:
@sheremet-va I openned a new PR https://github.com/vitest-dev/vitest/pull/2181, as this one is a bit stale and I add a bit of trouble with the lock file

Issue #1850, opened 2022-08-14
Module '"vitest"' declares 'spyOn' locally, but it is not exported.


artokun:
### Describe the bug

By inspecting the package I can see that the type definitions are emitted, but the actual function is not being exported
![Screen Shot 2022-08-13 at 11 23 50 PM](https://user-images.githubusercontent.com/7666455/184525233-c286c554-47c0-419e-947e-ade50d73b7cd.png)

"vitest": "^0.21.1"

### Reproduction

Try to export spyOn from the vitest package

### System Info

```shell
System:
    OS: macOS 12.5
    CPU: (10) arm64 Apple M1 Max
    Memory: 15.98 GB / 64.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.14.0 - ~/.volta/tools/image/node/16.14.0/bin/node
    Yarn: 1.22.17 - ~/.volta/tools/image/yarn/1.22.17/bin/yarn
    npm: 8.5.2 - ~/.volta/tools/image/npm/8.5.2/bin/npm
  Browsers:
    Brave Browser: 103.1.40.105
    Chrome: 104.0.5112.79
    Firefox: 101.0.1
    Safari: 15.6
  npmPackages:
    vitest: ^0.21.1 => 0.21.1
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

artokun:
This is a non issue! Please see my PR comment for anyone who has the same issue
https://github.com/vitest-dev/vitest/pull/1851#issuecomment-1214301377

PR #1851, opened 2022-08-14
fix: include spy directory export in runtime barrel

artokun:
This PR simply exports the spy module from the integrations directory so that its exports are available as expected

Fixes [#1850]


sheremet-va:
They were removed from exports intentionally. Users should use `vi.fn` and `vi.spyOn`.

artokun:
Ah I see, well I hope whoever searches for this in the future:

```ts
import { vi } from "vitest";
vi.spyOn(console, "log")
```
indeed works. Will close, thanks for the prompt review @sheremet-va

PR #1854, opened 2022-08-15
feat: support `--segfault-retry=3`

antfu:
Context: https://github.com/vitejs/vite/pull/9508

Changes:
- Wrapper the main CLI in a child_process
- Supported `--segfault-retry` CLI flag and `VITEST_SEGFAULT_RETRY` env
- When retries set but still detecting the segfault, a tip will be displayed

image

Co-authored-by: tony19 <tony19@users.noreply.github.com>


packages/vitest/src/node/cli-wrapper.ts
```diff
@@ -0,0 +1,96 @@
...
+have Vitest auto retries on flaky segfaults.\n`))
+    }
+  }
+  process.exit(exitCode)
+}
+
+main()
+
+async function start(args: string[]) {
+  const child = execa('node', [fileURLToPath(ENTRY), ...args], {
```
sheremet-va, line 72: Doesn't it hurt performance for "--segfault-retry=1" that we always starts a child process? Shouldn't we just import `./cli.mjs` in this case?

sheremet-va (changes requested):

packages/vitest/src/node/cli-wrapper.ts
sheremet-va: Hm, but we can't show useful error message that way 🤔 

I guess it's ok for now.

sheremet-va (approved):

packages/vitest/src/node/cli-wrapper.ts
antfu: I didn't count how many times we invoke `vitest` in our test, but at the high level, it doesn't make CI time increase. Using this would allow us to provide information about the segfault reasoning and issue link.

PR #1869, opened 2022-08-17
chore(deps): update tinyspy to 1.0.2

simon-abbott:
Tinyspy has had a couple updates, we should make sure they are included in the next vitest release. :)


simon-abbott:
Whoops, wrong base. One sec.

simon-abbott:
Same weird error as in #1868. Something's broken that isn't this change.

Issue #1877, opened 2022-08-18
console.Console is undefined


ahaoboy:
### Describe the bug

In vitest console.Console is undefined

### Reproduction

```
console.log(typeof console.Console)
//  vitest: undefined
//  node: function
```

### System Info

```shell
System:
    OS: macOS 12.5
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 215.25 MB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 14.19.0 - ~/Library/Caches/fnm_multishells/35067_1659959585448/bin/node
    Yarn: 1.22.17 - ~/Library/pnpm/yarn
    npm: 6.14.16 - ~/Library/Caches/fnm_multishells/35067_1659959585448/bin/npm
  Browsers:
    Chrome: 104.0.5112.79
    Edge: 104.0.1293.54
    Safari: 15.6
```

### Used Package Manager

pnpm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #1878, opened 2022-08-18
fix: console.Console is undefined (close: #1877)

ahaoboy:


antfu:
It sould be done in `packages/vitest/src/integrations/env/node.ts` as this is a Node specific API

sheremet-va (approved):

PR #1880, opened 2022-08-18
feat(coverage-istanbul): add "all" option

AriPerkkio:
Adds similar `all` option as [`c8`](https://github.com/bcoe/c8#checking-for-full-source-coverage-using---all) and [`nyc`](https://github.com/istanbuljs/nyc#common-configuration-options) have.

References from other projects:
- `c8`: https://github.com/bcoe/c8/blob/v7.12.0/lib/report.js#L187-L222
- `nyc`: https://github.com/istanbuljs/nyc/blob/v15.1.0/index.js#L186-L217
- `jest`https://github.com/facebook/jest/blob/v28.1.3/packages/jest-reporters/src/CoverageReporter.ts#L107-L222

### DIFF
--- a/packages/coverage-istanbul/src/provider.ts
+++ b/packages/coverage-istanbul/src/provider.ts
@@ -23,7 +23,10 @@ interface TestExclude {
     exclude?: string | string[]
     extension?: string | string[]
     excludeNodeModules?: boolean
-  }): { shouldInstrument(filePath: string): boolean }
+  }): {
+    shouldInstrument(filePath: string): boolean
+    glob(cwd: string): Promise<string[]>
+  }
 }
 
 export class IstanbulCoverageProvider implements CoverageProvider {
@@ -95,11 +98,14 @@ export class IstanbulCoverageProvider implements CoverageProvider {
   }
 
   async reportCoverage() {
-    const mergedCoverage = this.coverages.reduce((coverage, previousCoverageMap) => {
+    const mergedCoverage: CoverageMap = this.coverages.reduce((coverage, previousCoverageMap) => {
       const map = libCoverage.createCoverageMap(coverage)
       map.merge(previousCoverageMap)
       return map
     }, {})
+
+    if (this.options.all)
+      await this.includeUntestedFiles(mergedCoverage)
 
     const sourceMapStore = libSourceMaps.createSourceMapStore()
     const coverageMap: CoverageMap = await sourceMapStore.transformCoverage(mergedCoverage)
@@ -172,6 +178,39 @@ export class IstanbulCoverageProvider implements CoverageProvider {
             console.error(errorMessage)
           }
         }
+      }
+    }
+  }
+
+  async includeUntestedFiles(coverageMap: CoverageMap) {
+    // Load, instrument and collect empty coverages from all files which
+    // are not already in the coverage map
+    const includedFiles = await this.testExclude.glob(this.ctx.config.root)
+    const uncoveredFiles = includedFiles
+      .map(file => resolve(this.ctx.config.root, file))
+      .filter(file => !coverageMap.data[file])
+
+    const transformResults = await Promise.all(uncoveredFiles.map(async (filename) => {
+      const transformResult = await this.ctx.vitenode.transformRequest(filename)
+      return { transformResult, filename }
+    }))
+
+    for (const { transformResult, filename } of transformResults) {
+      const sourceMap = transformResult?.map
+
+      if (sourceMap) {
+        this.instrumenter.instrumentSync(
+          transformResult.code,
+          filename,
+          {
+            ...sourceMap,
+            version: sourceMap.version.toString(),
+          },
+        )
+
+        const lastCoverage = this.instrumenter.lastFileCoverage()
+        if (lastCoverage)
+          coverageMap.data[lastCoverage.path] = lastCoverage
       }
     }
   }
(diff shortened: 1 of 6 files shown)


packages/coverage-istanbul/src/provider.ts
```diff
@@ -173,6 +180,35 @@ export class IstanbulCoverageProvider implements CoverageProvider {
...
+
+  async includeUntestedFiles(coverageMap: CoverageMap) {
+    const files = await this.testExclude.glob(this.ctx.config.root)
+
+    for (const file of files) {
+      const filename = resolve(this.ctx.config.root, file)
+
+      if (!coverageMap.data[filename]) {
+        const transformResult = await this.ctx.vitenode.transformRequest(filename) as TransformResult
+        const sourceMap = transformResult?.map as NonNullable<TransformResult['map']>
```
AriPerkkio, line 192: I'm not sure about the `this.ctx.vitenode.transformRequest` part. Just saw that this was used elsewhere to get source maps. 

Does this look alright? The goal here is to load transformed code and source maps for it. After quick testing this seemed to work OK.

sheremet-va (approved):

packages/coverage-istanbul/src/provider.ts
```diff
@@ -175,6 +181,36 @@ export class IstanbulCoverageProvider implements CoverageProvider {
       }
     }
   }
+
+  async includeUntestedFiles(coverageMap: CoverageMap) {
+    const files = await this.testExclude.glob(this.ctx.config.root)
+
+    for (const file of files) {
```
sheremet-va, line 188: Maybe we should do `transformRequest` inside `Promise.all`?

sheremet-va (changes requested):

packages/coverage-istanbul/src/provider.ts
AriPerkkio: Sure. Should we still set some maximum limit of parallel `transformRequest` calls - so that we don't end up transforming 1000 files inside a single `Promise.all`. Maybe loop the `files` in sets of 20-25?

sheremet-va: Now that you say this, I guess we should 🤔 We already use `p-limit` library for `.concurrent`, so you can reuse it here. But I think we can be liberal with the maximum number, tbh - maybe base it on cores/available memory?

We already do `Promise.all` on codebase here with our limit tho:

https://github.com/vitest-dev/vitest/blob/5c064944126c36f3e1df0ecd5289da3b67b4353b/packages/vitest/src/node/core.ts#L184

AriPerkkio: Now using `p-limit` and limiting parallel transforms with the count of threads (minus one so that main thread is not blocked). I tested this setup with +100 generated uncovered files, with artificial timeout. Batching seems to work nicely. 

One thing to pay attention here @sheremet-va. I added `p-limit` as devDependency - this seems wrong. In `packages/vitest` it is declared as devDependency. Shouldn't it be regular dependency since it is used in runtime by consumers? 

https://github.com/vitest-dev/vitest/blob/03f7450957de6d18826040c7c5a1ea98712b251c/packages/vitest/package.json#L133

sheremet-va: > Shouldn't it be regular dependency since it is used in runtime by consumers?

dev dependencies are inlined, when Vitest is bundled, so no, it shouldn't.

packages/coverage-istanbul/src/provider.ts
```diff
@@ -175,6 +183,44 @@ export class IstanbulCoverageProvider implements CoverageProvider {
...
+
+  async includeUntestedFiles(coverageMap: CoverageMap) {
+    // Load, instrument and collect empty coverages from all files which
+    // are not already in the coverage map
+    const includedFiles = await this.testExclude.glob(this.ctx.config.root)
+    const uncoveredFiles = includedFiles
+      .map(file => resolve(this.ctx.config.root, file))
+      .filter(file => !coverageMap.data[file])
+
+    const maxThreads = this.ctx.config.maxThreads ?? Math.max(cpus().length - 1, 1)
```
sheremet-va, line 195: I feel like this is bad idea.

1. `maxThreads` is not meant to be used like this
2. `cpus()`just returns virtual cores, which are usually not a lot

As I said before, we can be very liberal with this values - you proposed 25-30, I was thinking much more (transforming is not a hard operation), but in this PR we might even get 1 for `--no-threads`.

This value can be much higher, we don't even have limits in other places.

sheremet-va (changes requested):

packages/coverage-istanbul/src/provider.ts
sheremet-va: @antfu can you back me up on this? is this premature optimisation?

AriPerkkio: Alright, I must have misunderstood this part.

> maybe base it on **cores**/available memory?

I've now set the limit to a randomly picked value of 50. If this indeed seems unnecessary optimization, I'm open to remove it completely and run all transforms in a single `Promise.all`.

antfu: I guess we could use `Promise.all` here, it shouldn't have much difference at this operation.

AriPerkkio: Now using `Promise.all` for all transforms, without `p-limit`.

sheremet-va (approved):

antfu (approved):

PR #1885, opened 2022-08-19
Show example chai.use example

emahuni:
Show how to extend vitest.chai using `chai.use` for existing plugins or ones that the developer creates. It's not clear in the documentation how to do this and I believe this is the place to just briefly mention it.

### DIFF
--- a/docs/guide/extending-matchers.md
+++ b/docs/guide/extending-matchers.md
@@ -6,7 +6,18 @@ title: Extending Matchers | Guide
 
 Since Vitest is compatible with both Chai and Jest, you can use either the `chai.use` API or `expect.extend`, whichever you prefer.
 
-This guide will explore extending matchers with `expect.extend`. If you are interested in Chai's API, check [their guide](https://www.chaijs.com/guide/plugins/).
+This guide will explore extending matchers with `expect.extend`, however, here is a little guide on how to extend Chai assertions in Vitest using an existing plugin. If you are interested in Chai's API on how to create Chai plugins check [their guide](https://www.chaijs.com/guide/plugins/). 
+
+To extend Vitest's chai assertions with an existing plugin or one you create, here is a quick example on how to `chai.use` [sinon-chai](https://www.chaijs.com/plugins/sinon-chai/) plugin with Vitest:
+
+```ts
+import { chai } from 'vitest'
+import SinonChai from 'sinon-chai'
+
+chai.use(SinonChai)
+
+// then use it in your tests as indicated in the plugin's documentation.
+```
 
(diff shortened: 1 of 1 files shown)


docs/guide/extending-matchers.md
```diff
@@ -6,7 +6,18 @@ title: Extending Matchers | Guide
 
 Since Vitest is compatible with both Chai and Jest, you can use either the `chai.use` API or `expect.extend`, whichever you prefer.
 
-This guide will explore extending matchers with `expect.extend`. If you are interested in Chai's API, check [their guide](https://www.chaijs.com/guide/plugins/).
+This guide will explore extending matchers with `expect.extend`, however, here is a little guide on how to extend Chai assertions in Vitest using an existing plugin. If you are interested in Chai's API on how to create Chai plugins check [their guide](https://www.chaijs.com/guide/plugins/). 
+
+To extend Vitest's chai assertions with an existing plugin or one you create, here is a quick example on how to `chai.use` [sinon-chai](https://www.chaijs.com/plugins/sinon-chai/) plugin with Vitest:
```
dammy001, line 11:
```suggestion
To extend Vitest's chai assertions with an existing plugin or one you create, here is a quick example of how to `chai.use` [sinon-chai](https://www.chaijs.com/plugins/sinon-chai/) plugin with Vitest:
```

antfu:
I am not sure if we should guide this. While we already provide built-in mocking, I'd consider using Sinon would be a bit niche and advanced usage. Guiding here might make it a bit misleading.

unicornware:
this would be a great addition to the documentation.  since migrating to vitest, i've been using `import chai from 'chai'` rather than `import { chai } from 'vitest'` in my setup file. i didn't realize i should've been using the latter until upgrading to `0.28.4` and i started getting `Invalid Chai property` errors

unicornware:
also, as far as examples go, would showing how to modify `chai.config` work?

maybe something like this:

```ts
import { chai } from 'vitest'

chai.config.includeStack = true
chai.config.truncateThreshold = 0
```

sheremet-va:
> also, as far as examples go, would showing how to modify `chai.config` work?

We support `chaiConfig` directly in the config now, so I don't think this is necessary.

But it would be nice to have a mention about `chai.use` in "Extending Matchers", but I think it should be explained in detail - not just reference other libraries, but shows how to add your own matcher.

@emahuni do you think you can add this? Or should this PR be closed?

sheremet-va:
Seems like the original author cannot work on this PR anymore (I asked if this should be closed a year ago). Feel free to open a new PR.

PR #1890, opened 2022-08-21
fix(reporter-junit): escape XML when in error message (fix: #1823)

### LINKED ISSUE #1823: Junit reporter does not escape error messages and stacktraces
### Describe the bug

The Junit reporter produces illegal XML if something resembling an XML tag shows up in the error message or stack trace.

As an example, we get this Junit-XML which is clearly malformed:
```
<testcase classname="packages/server.faas.scrape/__tests__/scrape_basics.spec.ts" name="Extract title from title tag" time="3.3660000000">
            <failure message="expected &apos;&apos; to be &apos;English Title&apos; // Object.is equality" type="AssertionError">
AssertionError: expected '' to be 'English Title' // Object.is equality
 ❯ test_function packages/server.faas.scrape/__tests__/scrape_basics.spec.ts:26:30
 ❯ Server.<anonymous> packages/lib.shared.test/fixtures.ts:16:16
            </failure>
        </testcase>

```

The string `<anonymous>` should be escaped.

### Reproduction

See the following Stackblitz for an example of a test that produces malformed XML:

https://stackblitz.com/edit/vitest-dev-vitest-b4w9ta?file=test/basic.test.ts

### System Info

```shell
System:
    OS: Linux 5.10 Ubuntu 20.04.3 LTS (Focal Fossa)
    CPU: (4) x64 Intel(R) Xeon(R) CPU @ 2.20GHz
    Memory: 2.61 GB / 3.84 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 18.7.0 - /usr/bin/node
    Yarn: 3.1.0 - /usr/bin/yarn
    npm: 8.15.0 - /usr/bin/npm
  npmPackages:
    vitest: 0.21.0 => 0.21.0
```

### Used Package Manager

yarn

### Validations

ChrisTowles:
When a error message has XML it made the junit reporter wasn't escaping when writing it to the log. 

This should fix #1823 

I don't love how i created `createSuiteHavingFailedTestWithXmlInError` to have an test case and welcome feedback.

### DIFF
--- a/test/reporters/src/data.ts
+++ b/test/reporters/src/data.ts
@@ -176,58 +176,5 @@ suite.tasks = tasks
 
 const files = [file]
 
-function createSuiteHavingFailedTestWithXmlInError(): File[] {
-  const file: File = {
-    id: '1223128da3',
-    name: 'test/core/test/basic.test.ts',
-    type: 'suite',
-    mode: 'run',
-    filepath: '/vitest/test/core/test/basic.test.ts',
-    result: { state: 'fail', duration: 145.99284195899963 },
-    tasks: [],
-  }
-
-  const suite: Suite = {
-    id: '',
-    type: 'suite',
-    name: 'suite',
-    mode: 'run',
-    file,
-    result: { state: 'pass', duration: 1.90183687210083 },
-    tasks: [],
-  }
-
-  const errorWithXml = new AssertionError({
-    message: 'error message that has XML in it <tag>',
-  })
-
-  errorWithXml.stack = 'Error: error message that has XML in it <tag>\n'
-    + '    at /vitest/test/core/test/basic.test.ts:8:32\n'
-    + '    at etc....'
-
-  const tasks: Task[] = [
-    {
-      id: '123_0',
-      type: 'test',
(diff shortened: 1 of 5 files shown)


ChrisTowles:
I believe the reason the tests are failing is the new test that requires a snapshot..

ChrisTowles:
Windows Tests was flakey, It broke 3 times with different errors seemingly not related to this change.

I force pushed a no-op change and it worked last time....

test/reporters/src/data.ts
```diff
@@ -174,4 +174,60 @@ const tasks: Task[] = [
 file.tasks = [suite]
 suite.tasks = tasks
 
-export const files = [file]
+const files = [file]
+
+function createSuiteHavingFailedTestWithXmlInError(): File[] {
```
Demivan, line 179: To make test cleaner you can move this to a separate file.

ChrisTowles: Moved to `data-for-junit.ts`

Demivan (approved):

Issue #1882, opened 2022-08-19
Subordinate confusing "Failed to fully serialize error" to the original error information


kaorukobo:
### Clear and concise description of the problem

A minimum reproduction of the problem is as follows:

```sh
# Use latest vite, vitest, and axios
npm i -D vitest@latest axios@latest

# Create a test which throws AxiosError object
mkdir test
cat > test/issue.test.js <<'EOS'
import axios from 'axios'
import { test } from 'vitest'

test("it", async () => {
  await axios.get('https://github.com/vitest-dev/vitest/issues/99999999')
})
EOS

# Run
./node_modules/.bin/vitest run
```

Then I get the following output:

```
Error: Failed to fully serialize error: Illegal invocation
Inner error message: Request failed with status code 404
```

While what developers would want to get are...

- The original error message `Request failed with status code 404`
- Codeframe/stacktrace of the original error.

But the above output looks that something weird happens, such as a bug from the vitest.
Also it does not provide enough information to solve the original error.

### Suggested solution

I suspect that there are many cases that an error object cannot be serialized. (While I don't know what this error serialization is for.)

So in such case, it would be better if the output could be as follows:

```
FAIL  test/issue.test.ts > it
AxiosError: Request failed with status code 404  (Internal Error: Failed to fully serialize error: Illegal invocation) 
 ❯ test/issue.test.js:5:X
      3| 
      4| test("it", async () => {
      5|   await axios.get('https://github.com/vitest-dev/vitest/issues/99999999')
       |        ^
      6| })
```

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

sheremet-va:
You cannot see stack trace **because** Vitest failed to serialize error. Serialisation happens here:

https://github.com/vitest-dev/vitest/blob/285ac7c24ef74ee244ce927b5e3bc757feaa918a/packages/vitest/src/runtime/error.ts#L56

PR welcome.

kaorukobo:
@sheremet-va
Thanks. I got why serialization is needed. I might try a PR.

[...]

Issue #1822, opened 2022-08-08
Importing @node-rs/crc32 results in segfault on Linux


flunderpero:
### Describe the bug

Merely importing @node-rs/crc32 leads to a segfault after tests are finished. 

```
vitest-dev-vitest-hkkxno llnode `which node` -c core.node.5641                                                                                                    [main ✗] [97c6c0a7b131]
(lldb) target create "/usr/bin/node" --core "core.node.5641"
Core file '/tmp/vitest-dev-vitest-hkkxno/core.node.5641' (x86_64) was loaded.
(lldb) plugin load '/root/llnode/llnode.so'
(lldb) settings set prompt '(llnode) '
(llnode) bt all
* thread #1, name = 'node', stop reason = signal SIGSEGV
  * frame #0: 0x00007f905d38bd40
    frame #1: 0x00007f90663ba5a1 libpthread.so.0`__nptl_deallocate_tsd.part.0 + 145
    frame #2: 0x00007f90663bb62a libpthread.so.0`start_thread + 250
    frame #3: 0x00007f90662e0133 libc.so.6`clone + 67
  thread #2, stop reason = signal 0
    frame #0: 0x00007f90663c2376 libpthread.so.0`__pthread_cond_wait + 534
    frame #1: 0x000000000164ca09 node`uv_cond_wait at thread.c:780
    frame #2: 0x0000000001638d04 node`worker(arg=<unavailable>) at threadpool.c:76
    frame #3: 0x00007f90663bb609 libpthread.so.0`start_thread + 217
    frame #4: 0x00007f90662e0133 libc.so.6`clone + 67
  thread #3, stop reason = signal 0
    frame #0: 0x00007f90663c5454 libpthread.so.0`do_futex_wait.constprop.0 + 52
    frame #1: 0x00007f90663c5548 libpthread.so.0`__new_sem_wait_slow.constprop.0 + 152
    frame #2: 0x000000000164c842 node`uv_sem_wait at thread.c:626
    frame #3: 0x000000000164c830 node`uv_sem_wait(sem=0x0000000004e15180) at thread.c:682
    frame #4: 0x0000000000c5e515 node`node::inspector::(anonymous namespace)::StartIoThreadMain(void*) + 53
    frame #5: 0x00007f90663bb609 libpthread.so.0`start_thread + 217
    frame #6: 0x00007f90662e0133 libc.so.6`clone + 67
  thread #4, stop reason = signal 0
[...]

PR #1892, opened 2022-08-21
feat: try to serialize error as possible

kaorukobo:
This PR is based on issue #1882 and improves the error serialization.

As per an [test example](https://github.com/kaorukobo/vitest/commit/2a8abd9bb1c35ced983401e7d6dd397c3a169697#diff-bfdbab86fc478fa36ab4a47c11aeae9a5a1aaa0ce4b1ac8bf89495711654e867), it serializes an error object as possible even if the object contains unserializable values.

Therefore, vitest will not show `Failed to fully serialize` error anymore and the stacktrace/codeframe comes up as usual.

## Substitution of unserializable value

Values that failed to serialize are substituted with an special object as follows: 
`{ UNSERIALIZABLE_VALUE: { error: 'ErrorMessage' } }`.

## Note on axios

As for axios mentioned in #1822, the AxiosError object does not contain `stack` property in the first place. So the stacktrace does not appear even though this patch is applied.

### DIFF
--- a/packages/vitest/src/runtime/error.ts
+++ b/packages/vitest/src/runtime/error.ts
@@ -7,45 +7,50 @@ const OBJECT_PROTO = Object.getPrototypeOf({})
 
 // https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm
 export function serializeError(val: any, seen = new WeakMap()): any {
-  if (!val || typeof val === 'string')
-    return val
-  if (typeof val === 'function')
-    return `Function<${val.name}>`
-  if (typeof val !== 'object')
-    return val
-  if (val instanceof Promise || 'then' in val || (val.constructor && val.constructor.prototype === 'AsyncFunction'))
-    return 'Promise'
-  if (typeof Element !== 'undefined' && val instanceof Element)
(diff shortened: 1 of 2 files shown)


sheremet-va:
Tbh instead I would return an error with `message` and `stack` copied, and just a list of other available props:

```
SomeError: message
  stacktrace {
  "someValue",
  "unserializable"
}
```

kaorukobo:
@sheremet-va Looks cleaner. Are you planning to implement the idea by yourself?

kaorukobo:
Or, the entire `serializeError()` could be far simplified, by making it return an object which contains only the basic Error's properties (`name, message, stack, cause`).

...if there are no reasons for keeping all of original properties.

sheremet-va:
> @sheremet-va Looks cleaner. Are you planning to implement the idea by yourself?

No, I don't have time.

> by making it return an object which contains only the basic Error's properties

No, we should try and return as much information as possible. In my proposed solution we can just skip failed property, and mark it as `unserialisable`, instead of returning only names.

kaorukobo:
@sheremet-va Thank you for implementing it. I've confirmed that issue #1882 cannot be reproduced anymore.

PR #1896, opened 2022-08-22
fix: invalidate new worker and its dependencies (fix #1873)

### LINKED ISSUE #1873: Can't run two web worker test cases in one file
### Describe the bug

Firstly, I used `@vitest/web-worker` to simulate web worker: 
```
export default defineConfig({
  test: {
    globals: true,
    environment: 'jsdom',
    setupFiles: ['@vitest/web-worker'],
  },
})
```
Next, I wrote two test in one file(If I only add one test, It will be OK).

```typescript
const createWorker = () => new Worker(new URL('./fibonacci', import.meta.url), {
  type: 'module'
})

test('run MyWorker first time', async () => {
  const worker = createWorker()
  const ret = await promiseWorker(worker, 10)
  expect(ret).toBe(89)
})

// The second test will print such errors:
// Error: Test timed out in 5000ms.
// If this is a long-running test, pass a timeout value as the last argument or configure it globally with "testTimeout".
test('run MyWorker second time', async () => {
  const worker = createWorker()
  const ret = await promiseWorker(worker, 10)
  expect(ret).toBe(89)
})

function promiseWorker(worker: Worker, n: number) {
    worker.onmessage = function({data}) {
      resolve(data)
    }
    worker.onmessageerror = function({data}) {
      reject(data)
    }

    worker.onerror = function(e) {
      reject(e)
    }

    worker.postMessage(n)
  }) 
}
```

This is my `fibonacci.ts` file, If I not import `exportWorker` but inline it, the tests will all passed : 

```typescript
import exportWorker from "./exportWorker"

// If I use the `exportWorker` from ./exportWorker.ts , It will be broken.
exportWorker(fibonacci)

// If I use inline exportWorker, It will passed
// function exportWorker(fn: CallableFunction) {
//   self.onmessage = function({data}) {
//     const result = fn(data)

//     self.postMessage(result)
//   }
// }
```

### Reproduction

[vitest-web-worker-reproduction](https://github.com/mysteryven/vitest-web-worker-reproduction)

### System Info

```shell
System:
    OS: macOS 12.4
    CPU: (4) x64 Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz
    Memory: 16.84 MB / 8.00 GB
    Shell: 3.2.57 - /bin/sh
  Binaries:
    Yarn: 1.22.18 - ~/.nvm/versions/node/v16.14.2/bin/yarn
  Browsers:
    Firefox: 80.0.1
    Safari: 15.5
  npmPackages:
    vite: ^3.0.0 => 3.0.8 
    vitest: latest => 0.22.1
```

### Used Package Manager

pnpm

### Validations

sun0day:
Worker and its dependencies should be invalidated every time it is created so that Worker module and its dependencies will be injected into a new `context` on the next creation.

fix #1873

Before PR:

image

After PR:

image

### DIFF
--- a/packages/web-worker/src/pure.ts
+++ b/packages/web-worker/src/pure.ts
@@ -133,11 +133,11 @@ export function defineWebWorker() {
 
       runner.executeFile(fsPath)
         .then(() => {
-          invalidates.forEach((fsPath) => {
-            // worker should be new every time
-            moduleCache.delete(fsPath)
-            moduleCache.delete(`mock:${fsPath}`)
-          })
+          // worker should be new every time, invalidate its sub dependency
+          moduleCache.invalidateSubDepTree(
+            invalidates.reduce(
+              (acc, fsPath) => acc.concat([fsPath, `mock:${fsPath}`]), [] as string[]
+          ))
           const q = this.messageQueue
           this.messageQueue = null
           if (q)
--- a/packages/web-worker/src/pure.ts
+++ b/packages/web-worker/src/pure.ts
@@ -44,7 +44,6 @@ interface InlineWorkerContext {
   postMessage: (data: any) => void
   self: InlineWorkerContext
   global: InlineWorkerContext
-  invalidates: string[]
   importScripts?: any
 }
 
@@ -95,7 +94,6 @@ export function defineWebWorker() {
     public onerror: null | Procedure = null
 
     constructor(url: URL | string) {
-      const invalidates: string[] = []
       const context: InlineWorkerContext = {
         onmessage: null,
         dispatchEvent: (event: Event) => {
@@ -113,7 +111,6 @@ export function defineWebWorker() {
         get global() {
           return context
         },
-        invalidates,
       }
 
       this.inside.on('message', (e) => {
@@ -129,15 +126,11 @@ export function defineWebWorker() {
       const id = (url instanceof URL ? url.toString() : url).replace(/^file:\/+/, '/')
 
       const fsPath = toFilePath(id, config.root)
-      invalidates.push(fsPath)
 
       runner.executeFile(fsPath)
         .then(() => {
           // worker should be new every time, invalidate its sub dependency
-          moduleCache.invalidateSubDepTree(
-            invalidates.reduce(
-              (acc, fsPath) => acc.concat([fsPath, `mock:${fsPath}`]), [] as string[],
-            ))
+          moduleCache.invalidateSubDepTree([fsPath, `mock:${fsPath}`])
           const q = this.messageQueue
           this.messageQueue = null
           if (q)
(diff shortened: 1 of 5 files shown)


packages/web-worker/src/pure.ts
```diff
@@ -133,11 +133,11 @@ export function defineWebWorker() {
 
       runner.executeFile(fsPath)
         .then(() => {
-          invalidates.forEach((fsPath) => {
-            // worker should be new every time
-            moduleCache.delete(fsPath)
-            moduleCache.delete(`mock:${fsPath}`)
-          })
+          // worker should be new every time, invalidate its sub dependency
```
sheremet-va, line 136: Currently `invalidates` is not populated, and always contains only `fsPath`, so we can just remove it's usage entirely.

sheremet-va (changes requested):
Please, add a test for his use case.

packages/web-worker/src/pure.ts
sun0day: In fact, every time vitest creates a new worker via the same URL, the `context` (or `self`) injected into the worker and its dependencies is not always up-to-date because of the `mod` cache.

So we need to clear both the worker module and its dependencies `mod` cache.

sun0day:
> Please, add a test for this use case.

Yeah, I will add some test cases later.

packages/web-worker/src/pure.ts
sun0day: For example:

```ts
// entry.test.ts
test('case1', async () => {
  new Worker(new URL('./worker', import.meta.url), {
    type: 'module'
  })
})

test('case2', async () => {
  new Worker(new URL('./worker', import.meta.url), {
    type: 'module'
  })
})
```

```ts
// worker.ts
import getDepSelf from './worker-dep'

console.log(getDepSelf() === self) // in case1, output is true while in case2, output is false
```

```ts
// worker-dep.ts
export default () => {
  return self
}
```

sheremet-va: Well, yes, I was hoping this is what this PR does, no?

sun0day: Yes, this PR does solve this problem

sun0day:
> Please, add a test for this use case.

Done. @sheremet-va

packages/web-worker/src/pure.ts
sheremet-va: Still, I don't think we need `invalidates` at all here, it is not used anywhere except to add `fsPath`. 

It's original idea was to keep track of dependencies and remove them after (what this PR does in a different way).

I was asking you to remove it

sheremet-va (changes requested):

packages/web-worker/src/pure.ts
sun0day: So you mean it's OK to use `invalidateSubDepTree` to delete the dependency mod cache, besides that I should also remove the `invalidates` usage and not use it anymore？

sheremet-va: Yes

sun0day: Done. Sorry, I didn't catch what you mean before.

sheremet-va (approved):

PR #1648, opened 2022-07-15
fix: mock properties of classes/functions (fix #1523)

simon-abbott:
Previously classes weren't mocked properly because their prototypes were not being copied, so they lost all their methods. This PR rewrites the automocker to properly handle:

- Functions with properties (namely classes)
- Circular objects (also to handle classes, thanks JavaScript)

With these changes classes can now be properly automocked!

This fixes #1523.

PR #1903, opened 2022-08-23
fix: automock getters and setters

### LINKED ISSUE #1899: cannot spy on a Getter from a Service that is mocked - regression since version 0.21.x
### Describe the bug

I have a Vue Pinia Store implementation that I want to test and I want to mock an Service instead of using the actual service, it works fine for nearly everything **except** when trying to spy on a Getter where it throws an error saying that `"x does not exist"`, if however I rename my Getter to a function from `get x()` to `getX()` then it works as intended

This seems to be a regression since it was working properly in older Vitest version 0.17.x. I'm upgrading my Vue 3 project to Vite 3 and I started updating to latest Vitest version but the highest I could go before I get this new error was version 0.20.x and anything higher than that would throw the error. I also started upgrading to Vite 3 and now even older version of Vitest throws the error (probably because Vitest 0.17.x wasn't working properly with Vite 3). 

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-od71ph?file=test%2FsolutionDataService.test.ts,services%2FsolutionDataService.ts&initialPath=__vitest__

Run the Service test `solutionDataService.test.ts`

Simple code
```ts
vi.mock('../services/solutionDataService');
import SolutionDataService from '../services/solutionDataService';

test('service with getter', async () => {
  vi.spyOn(SolutionDataService, 'errorsAndWarnings', 'get').mockReturnValue({ msg: 'something' });
});
```
Error thrown

```sh
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯

FAIL  test/solutionDataService.test.ts > service with getter
Error: errorsAndWarnings does not exist
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
```

If we modify the `package.json` of that same stackblitz with older version (as shown below) then everything passes
```json
"devDependencies": {
  "@vitejs/plugin-vue": "^2.3.3",
  "@vue/test-utils": "latest",
  "jsdom": "latest",
  "vite": "^2.9.15",
  "vitest": "^0.20.3"
}
```
if we try to bump only Vitest to `0.21.0` (fixed version), then it starts failing again with the error, so it looks like the regression was introduced in version `0.21.0` and just guessing that it might be caused by PR #1786 (very wild guess just by looking at the code)

### System Info

```shell
System:
    OS: Windows 10 10.0.19043
    CPU: (12) x64 Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
    Memory: 13.52 GB / 31.75 GB
  Binaries:
    Yarn: 1.22.19 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 8.15.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 104.0.5112.82
    Edge: Spartan (44.19041.1266.0), Chromium (104.0.1293.54)
  npmPackages:
    @vitejs/plugin-vue: ^3.0.3 => 3.0.3
    vite: ^3.0.9 => 3.0.9
    vitest: ^0.22.1 => 0.22.1
```

### Used Package Manager

yarn

### Validations

simon-abbott:
This fixes a regression in 0.21.0.

Fixes #1899

### DIFF
--- a/packages/vitest/src/runtime/mocker.ts
+++ b/packages/vitest/src/runtime/mocker.ts
@@ -209,10 +209,21 @@ export class VitestMocker {
       const isModule = containerType === 'Module' || !!container.__esModule
       for (const { key: property, descriptor } of getAllMockableProperties(container)) {
         // Modules define their exports as getters. We want to process those.
-        if (!isModule) {
-          // TODO: Mock getters/setters somehow?
-          if (descriptor.get || descriptor.set)
-            continue
+        if (!isModule && descriptor.get) {
+          try {
+            const newDescriptor: PropertyDescriptor = {
+              enumerable: true,
+              configurable: true,
+              get: spyModule.fn(),
+            }
+            if (descriptor.set)
+              newDescriptor.set = spyModule.fn()
+            Object.defineProperty(newContainer, property, newDescriptor)
+          }
+          catch (error) {
+            // Ignore errors, just move on to the next prop.
+          }
+          continue
         }
 
         // Skip special read-only props, we don't want to mess with those.
--- a/packages/vitest/src/runtime/mocker.ts
+++ b/packages/vitest/src/runtime/mocker.ts
@@ -211,14 +211,7 @@ export class VitestMocker {
         // Modules define their exports as getters. We want to process those.
         if (!isModule && descriptor.get) {
           try {
-            const newDescriptor: PropertyDescriptor = {
-              enumerable: true,
-              configurable: true,
-              get: spyModule.fn(),
-            }
-            if (descriptor.set)
-              newDescriptor.set = spyModule.fn()
-            Object.defineProperty(newContainer, property, newDescriptor)
+            Object.defineProperty(newContainer, property, descriptor)
           }
           catch (error) {
             // Ignore errors, just move on to the next prop.
--- a/examples/mocks/test/automocking.spec.ts
+++ b/examples/mocks/test/automocking.spec.ts
@@ -1,8 +1,10 @@
 import type * as exampleModule from '../src/example'
 import log from '../src/log'
+import { A } from '../src/moduleA'
 import { methodSymbol, moduleWithSymbol } from '../src/moduleWithSymbol'
 
 vi.mock('../src/log')
+vi.mock('../src/moduleA')
 vi.mock('../src/moduleWithSymbol')
 
 test('all mocked are valid', async () => {
@@ -54,4 +56,8 @@ test('automock properly restores mock', async () => {
 
   expect(moduleWithSymbol[methodSymbol]()).toBe('hello')
   expect(moduleWithSymbol.warn()).toBe('hello')
+})
+
+test('automock has a getter', () => {
+  expect(A).toBe('A')
 })
--- a/test/core/src/mockedC.ts
+++ b/test/core/src/mockedC.ts
@@ -11,6 +11,10 @@ export class MockedC {
     return mockedA()
   }
 
+  get getOnlyProp(): number {
+    return 42
+  }
+
   get getSetProp(): number {
     return 123
   }
--- a/test/core/test/mocked.test.js
+++ b/test/core/test/mocked.test.js
@@ -1,4 +1,4 @@
-// this file should not be converted to js
+// this file should not be converted to ts
 // so it won't be transformed by esbuild
 
 import { assert, test, vi } from 'vitest'
--- a/test/core/test/mocked.test.ts
+++ b/test/core/test/mocked.test.ts
@@ -13,6 +13,25 @@ vitest.mock('virtual-module', () => ({ value: 'mock' }))
 vitest.mock('../src/mockedC')
 vitest.mock('../src/mockedD')
 
+/**
+ * Get a property descriptor from an object.
+ *
+ * This is different from `Object.getOwnPropertyDescriptor` because it recurses
+ * into the prototype chain until it either finds a match or reaches the end.
+ *
+ * @param object The object that contains the property.
+ * @param property The property.
+ * @returns The property's descriptor, or undefined if no matching property was found.
+ */
+function getPropertyDescriptor(object: any, property: PropertyKey) {
+  for (let o = object; o; o = Object.getPrototypeOf(o)) {
+    const descriptor = Object.getOwnPropertyDescriptor(o, property)
+    if (descriptor)
+      return descriptor
+  }
+  return undefined
+}
+
 test('submodule is mocked to return "two" as 3', () => {
   assert.equal(3, two)
 })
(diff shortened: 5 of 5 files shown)


packages/vitest/src/runtime/mocker.ts
```diff
@@ -209,10 +209,21 @@ export class VitestMocker {
...
+              get: spyModule.fn(),
+            }
+            if (descriptor.set)
+              newDescriptor.set = spyModule.fn()
+            Object.defineProperty(newContainer, property, newDescriptor)
+          }
+          catch (error) {
+            // Ignore errors, just move on to the next prop.
+          }
+          continue
```
ghiscoding, line 226: isn't this `continue` redundant? perhaps you could move it into the `catch`

simon-abbott: No, if we encounter a getter we don't want to do any of the other mocking logic afterwards. The `continue` here says "we're done with this prop, `continue` on to the next one". The `catch` is just to allow graceful handling of trying to overwrite non-configurable props.

packages/vitest/src/runtime/mocker.ts
```diff
@@ -209,10 +209,21 @@ export class VitestMocker {
...
-        if (!isModule) {
-          // TODO: Mock getters/setters somehow?
-          if (descriptor.get || descriptor.set)
-            continue
+        if (!isModule && descriptor.get) {
+          try {
+            const newDescriptor: PropertyDescriptor = {
+              enumerable: true,
+              configurable: true,
+              get: spyModule.fn(),
```
sheremet-va, line 217: Do we want to _mock_ getters/setters? I assumed we just place them on a new object as is, no? If not, that's different from jest, as I understand it, and needs to be documented in migration guide and mocking guide.

sheremet-va (changes requested):

packages/vitest/src/runtime/mocker.ts
simon-abbott: > Do we want to _mock_ getters/setters? I assumed we just place them on a new object as is, no?

I think we want to mock them. Since they are functions it makes sense to treat them the same as any other function during the automocking process, which is to completely stub it out to ensure it doesn't cause any side effects. If you want to have partial implementation you can always do a custom `__mocks__` definition, use `importActual`, or just not mock it in the first place.

I can't think of a good reason to mock everything _except_ getters/setters. I see no reason why they should be special here, but if you have a compelling argument I'm happy to consider it.

> If not, that's different from jest, as I understand it, and needs to be documented in migration guide and mocking guide.

This is already a difference from Jest. Jest doesn't mock getters or setters _at all_, which is why I had implemented it the way I did in #1648.

sheremet-va: What do you mean by `Since they are functions`? If they are on first level, then the getter was added by Vite: 

```ts
const someValue = 3
Object.defineProperty(__vite_exports__, 'someValue', { get: () => someProperty })
```

> I can't think of a good reason to mock everything except getters/setters. I see no reason why they should be special here, but if you have a compelling argument I'm happy to consider it.

In the same message you are saying that Jest doesn't mock them. I see no reason to automock getters/setters at all, to be honest.

> Jest doesn't mock getters or setters at all, which is why I had implemented it the way I did in https://github.com/vitest-dev/vitest/pull/1648.

This doesn't explain why you just didn't put the same descriptor on new object. Why does a new object has a hole in it, compared to the original?

simon-abbott: > What do you mean by `Since they are functions`? If they are on first level, then the getter was added by Vite:
> 
> ```ts
> const someValue = 3
> Object.defineProperty(__vite_exports__, 'someValue', { get: () => someProperty })
> ```

Getters are functions with 0 arguments that are accessed without parenthesis. Even in your example right there the getter is defined as an arrow function that returns `someProperty`.

---

> In the same message you are saying that Jest doesn't mock them. I see no reason to automock getters/setters at all, to be honest.

This PR exists because of #1899. If we want to just stick with Jest compatibility then close this PR and mark #1899 as wontfix. 

---

> Also, your implementation doesn't allow restoring mocks to original setters/getters.

This is true, but imo that doesn't matter. Why would you want to restore the mocks on an automocked import anyway? Plus, if we're going with the "the way Jest does it" argument, Jest doesn't allow restoring _any_ automocked properties. If you want to mock a getter and restore it later just use `spyOn` yourself.

---

> This doesn't explain why you just didn't put the same descriptor on new object. Why does a new object has a hole in it, compared to the original?

What do you mean "has a hole in it"? This PR _fills_ a hole in the mocked objects.

sheremet-va: > Getters are functions with 0 arguments that are accessed without parenthesis

I know what a function is. I don't know however how this is important to this discussion? Why do we need to treat all "technically" functions the same?

> This PR exists because of https://github.com/vitest-dev/vitest/issues/1899. If we want to just stick with Jest compatibility then close this PR and mark https://github.com/vitest-dev/vitest/issues/1899 as wontfix, and I'll put up another PR removing this FIXME:

This PR fixes it in a way that is not conventional. I would prefer to just save the descriptors as they are, no need to "auto" mock it.

> Why would you want to restore the mocks on an automocked import anyway?

"Because I want to" is enough of a reason, when implementation is a one liner and doesn't contradict anything else here. By your logic of "it's a function, it should follow what other functions do", it should be restorable, since it's possible to restore automocked functions. But it doesn't matter, because we don't want to automock setters/getters, just move them on a new object.

> What do you mean "has a hole in it"? This PR fills a hole in the mocked objects.

I was refering to your previous PR that had a hole, because you were saying that you implemented in this way intentionally ("which is why I had implemented it the way I did").

simon-abbott: > Why do we need to treat all "technically" functions the same?

> I would prefer to just save the descriptors as they are, no need to "auto" mock it.

> But it doesn't matter, because we don't want to automock setters/getters, just move them on a new object.

Say you have the following class:

```js
import { readFile } from 'node:fs/promises';

class SettingsWrapper {
  constructor(settingsFilePath) {
    this.settingsFilePath = settingsFilePath;
  }

  get settings() {
    return readFile(this.settingsFilePath, { encoding: 'utf-8' }).then(
      (json) => {
        return JSON.parse(json);
      },
    );
  }
}
```

If we just blindly copy the descriptor over to the new object then accessing the `settings` getter on a mocked instance would perform an _actual disk read_. Sure you can argue that this is bad design, but it's still possible and thus needs to be accounted for in how we handle automocking. Since getters and setters are functions they can do _anything_ a normal function can do, which includes some _serious_ side effects that people want to mock out.

---

> I was referring to your previous PR that had a hole, because you were saying that you implemented in this way intentionally ("which is why I had implemented it the way I did").

I implemented it that way before because that's how Jest has been doing it for over 5 years<sup>1</sup>, and they seemed to be doing just fine.

--- 

This whole conversation has made me think we really have two options here: either do exactly what Jest does and don't mock getters/setters at all, and if you want to stub them out during tests just set them yourself<sup>2</sup>, or one-up Jest and automock the getters/setters as in this PR. I feel _very_ strongly that just blindly copying them over to the final object is the _wrong_ solution and _will_ cause problems in the future if we try to go that route.

---

<sup>1</sup> https://bambielli.com/til/2017-09-17-jest-doesnt-mock-getters/
<sup>2</sup> https://bambielli.com/til/2017-09-17-jest-doesnt-mock-getters/#option-2-just-assign-a-value

simon-abbott: As for the restorability of Mocks, that's *another* difference from Jest. Jest doesn't allow automatic restoring of _any_ automocked attributes. If you want the original again in Jest you _have_ to use `requireActual`. The only reason I _didn't_ mirror that behavior in my original PR is because some of the existing tests already relied on that behavior, and I was trying to make my changes have as few breakages as possible.

sheremet-va: > Say you have the following class

This is just a hypothetical situation. By your own logic, we can ask user to take care of it.

>  I feel very strongly that just blindly copying them over to the final object is the wrong solution and will cause problems in the future if we try to go that route.

This is how it was working before in Vitest. Just restore this behaviour. 

As I said before, getters/setters don't only exist on classes and objects. Our module system is implemented with getters/setters. Exported constants *SHOULD* stay the same. This is the end of the conversation.

PR #1908, opened 2022-08-24
fix: test files path in windows

YaDev:
This is a fix for "Error: No test suite found" in Windows.

This should force it to use uppercase letters for the drivers.

For more info :
https://github.com/nodejs/node/issues/6624


sheremet-va (changes requested):
Maybe we should make a util function out of this? Or open PR to `pathe` for this? We don't use Node's `path`, and I thought the goal of `pathe` is to make pathing consistent.

YaDev:
Unfortunately, pathe uses path and does not solve this problem.
https://github.com/unjs/pathe/blob/main/src/path.ts

I think the filtering is not going to work well in windows 
"testFiles = testFiles.filter(i => filters.some(f => i.includes(f)))"

cause 
c:\project\src\tests\file.ts
is equal to 
C:\project\src\tests\file.ts
in Windows.

so, one of these commands 
"npx vitest C:\......." or "npx vitest c:\......." 

will fail 100%.

cmd, powershell, vscode & other IDEs can run into this issue. (had it on powershell and vscode)

fs.realpath.native might fix it but I think this solution should the best for now until a refactoring is done to the globTestFiles function.


So, move it to util ?

sheremet-va:
> Unfortunately, pathe uses path and does not solve this problem.
> 
> https://github.com/unjs/pathe/blob/main/src/path.ts
> 
> 
> 
> I think the filtering is not going to work well in windows 
> 
> "testFiles = testFiles.filter(i => filters.some(f => i.includes(f)))"
> 
> 
> 
> cause 
> 
> c:\project\src\tests\file.ts
> 
> is equal to 
> 
> C:\project\src\tests\file.ts
> 
> in Windows.
> 
> 
> 
> so, one of these commands 
> 
> "npx vitest C:\\......." or "npx vitest c:\\......." 
> 
> 
> 
> will fail 100%.
> 
> 
> 
> cmd, powershell, vscode & other IDEs can run into this issue. (had it on powershell and vscode)
> 
> 
> 
> fs.realpathSync.native might fix it but I think this solution should be the best for now until a refactoring is done to the globTestFiles function.
> 
> 
> 
> 
> 
> So, move it to util ? (I totally agree with that)

Of course it's using path, but I don't see the problem? They are wrapping it to make it consistent across platforms, which you are trying to do here. We can open PR to fix this issue on their side, I think.

YaDev:
It is pointless to argue.
You should read pathe's documentations carefully. If pathe's team should fix this problem, then go ahead and create a PR. Otherwise, submit a PR for nodejs like :
https://github.com/nodejs/node/issues/38459

or fix it on your end.

PR #1915, opened 2022-08-24
fix: add possibility to filter test files by fileName

### LINKED ISSUE #1797: Filter tests by filename in watch mode

As a developer using Vitest I want to have the option to filter test by filename in watch mode so that I can simply run only the test file I am currently working on and do not have to execute all test suites. 

I am coming from jest where I really often used this filter option (jest-watch-typeahead/filename) and it is very helpful for me.

### Suggested solution

Add a filter option like [jest-watch-typeahead/filename](https://github.com/jest-community/jest-watch-typeahead)

### Alternative

_No response_

### Additional context

vitest
image

jest
image

### Validations

bartoszgolebiowski:
Hello,

fix #1797

I have created PR with a new option for "node-vite".
I added to vitest config new property "fileNamePattern", it is similar to "testNamePattern"
It is responsible for skipping test files when the pattern does not match.
I added documentation for "fileNamePattern" next to "testNamePattern" property.
I added additional information for the CLI help option about the new property.
I added a test case to check if the new functionality works as expected. 

To test it you can:
1. test/core/vitest.config.ts -> remove "fileNamePattern" property. -> it will fail because I added failing file-name-pattern.test.ts
2. pnpm run test -p .test -> it will fail because I added failing file-name-pattern.test.ts


bartoszgolebiowski:
**examples/playwright** on CI / test (windows-latest, 16).
I think this is not related to my changes.

sheremet-va (changes requested):
I don't like basing solution on `testNamePattern`. We can already define a pattern for tests with `include`. 

If you just want to add `-p` option to CLI, you can call `this.rerunFiles` with a new set of files, filtered by the new pattern.

bartoszgolebiowski:
Hello,

I added a new folder `test-skip`, which is used to create the test scenario to check if filtering works correctly.
I set ` fileNamePattern: '/core/test/',`
Due to that it will match all tests inside the **_test_** folder, and will not match tests inside the **_test-skip_** folder.
I could not find any other method to create a test for this scenario, but I am open to discuss. 

According to your review, I have updated the `changeFilenamePattern `method to filter files array and pass results to `this.rerunFiles`

packages/vitest/src/node/stdin.ts
```diff
@@ -8,6 +8,7 @@ const keys = [
   ['a', 'rerun all tests'],
   ['f', 'rerun only failed tests'],
   ['u', 'update snapshot'],
+  ['p', 'filter by a filename regex pattern'],
```
bartoszgolebiowski, line 11: Typo

sheremet-va:
> According to your review, I have updated the `changeFilenamePattern `method to filter files array and pass results to `this.rerunFiles`

No, what I meant is that we don't need this property at all, because we already have `include` and `exclude`. When pressing `p`, we should just pass down correct files to `rerunFiles`.

I don't mind using regex for this.

packages/vitest/src/runtime/collect.ts
```diff
@@ -34,7 +33,7 @@ export async function collectTests(paths: string[], config: ResolvedConfig) {
       return await import(`${filepath}?v=${hash}`)
   }
 
-  for (const filepath of paths) {
+  for (const filepath of matchPattern(paths, config.fileNamePattern)) {
```
sheremet-va, line 36: We don't need this. Files are already filtered by Vitest by this point.

sheremet-va (changes requested):

packages/vitest/src/runtime/collect.ts
bartoszgolebiowski: OK, without that I cannot filter the initial test run, to check functionality in `pnpm test:ci`
I will think about how to deal with it later. 

Thank you

sheremet-va:
@bartoszgolebiowski I pushed some changes to your PR. We don't have tests for stdin right now, unfortunately tho, and creating infrastructure for it is out of scope of this PR, I think. But would be great to have this stuff tested.

I also noticed that after rerunning it doesn't update test count, can you look into that?

bartoszgolebiowski:
Hello @sheremet-va 

I added one **failing** test in the `"skip-test"` directory and new property `"fileNamePattern"`
 in `vitest.config.ts` which was responsible for setting an initial value for filter https://github.com/vitest-dev/vitest/pull/1915/commits/93056db8120d83cdc5b8b5c11b32516c3566a273#diff-644652b4f2da0d83159d02c25d7368a9480fe293826cc3314d4f25cd26672ea6L82

This filter runs all tests inside `"test"` directory and does not run test inside `"skip-test"` directory. 
I can see that the new test file and logic for injecting value are gone. 

To resolve it, we can still use https://github.com/vitest-dev/vitest/pull/1915/commits/0fcdd52b6b0ca97f54e819cab88438a69b161bdf#diff-ff14f3092ee007d54e9c9f53fa4044149899b44fa12979cc92f1d182d1a2b057R45 and assigning this value to 
https://github.com/vitest-dev/vitest/pull/1915/commits/93056db8120d83cdc5b8b5c11b32516c3566a273#diff-644652b4f2da0d83159d02c25d7368a9480fe293826cc3314d4f25cd26672ea6R26 and add failing test in the `"skip-test"` folder again.

```js
let latestFilename = String(ctx.config.fileNamePattern || '')
```

BR
Bartosz Golebiowski

sheremet-va:
> To resolve it, we can still use https://github.com/vitest-dev/vitest/commit/0fcdd52b6b0ca97f54e819cab88438a69b161bdf#diff-ff14f3092ee007d54e9c9f53fa4044149899b44fa12979cc92f1d182d1a2b057R45 and assigning this value to
https://github.com/vitest-dev/vitest/commit/93056db8120d83cdc5b8b5c11b32516c3566a273#diff-644652b4f2da0d83159d02c25d7368a9480fe293826cc3314d4f25cd26672ea6R26 and add failing test in the "skip-test" folder again.

Yes, but this doesn't actually test the interaction that user has - pressing "p" and inserting filter.

unional:
The file name filter doesn't work at all. Using `vitest` 0.25.5.

PR #1925, opened 2022-08-27
fix: clean coverage-c8 tmp before reporting (fix #1917)

### LINKED ISSUE #1917: v0.22 with c8 coverage does not clear tmp
### Describe the bug

After upgrading to v0.22 we noticed each run of vitest was getting progressively slower. After awhile it stopped working entirely and would fail with an out of memory error. After some digging - it seems that the files in `/coverage/tmp` are not being cleared out between runs. Prior to v0.22 this was happening automatically and seems like it should be the default behavior per c8's documentation.

### Reproduction

We are running `vitest run --environment jsdom --coverage`. After running - look in `/coverage/tmp` and note the number of `.json` files present. Each time you run, this folder gets larger.

### System Info

```shell
System:
    OS: macOS 12.5.1
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 8.64 GB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.13.2 - ~/.nvm/versions/node/v16.13.2/bin/node
    npm: 8.1.2 - ~/.nvm/versions/node/v16.13.2/bin/npm
  Browsers:
    Chrome: 101.0.4951.54
    Firefox: 100.0.2
    Safari: 15.6.1
  npmPackages:
    @vitejs/plugin-vue: ^3.0.0 => 3.0.3 
    vite: ^3.0.0 => 3.0.9 
    vitest: ^0.22.1 => 0.22.1
```

### Used Package Manager

npm

### Validations

sun0day:
fix #1917 

Clean tmp dir before coverage-c8 report each time.

Same as:

https://github.com/bcoe/c8/blob/7f1069dcce6821a6794557b39d6a13f620c64bad/bin/c8.js#L31-L35

### DIFF
--- a/packages/coverage-c8/src/provider.ts
+++ b/packages/coverage-c8/src/provider.ts
@@ -1,4 +1,4 @@
-import { existsSync, promises as fs } from 'fs'
+import { existsSync, promises as fs, mkdirSync, rmSync } from 'fs'
 import _url from 'url'
 import type { Profiler } from 'inspector'
 import { takeCoverage } from 'v8'
@@ -29,6 +29,18 @@ export class C8CoverageProvider implements CoverageProvider {
   }
 
   onBeforeFilesRun() {
+    let tmpExist = existsSync(this.options.tempDirectory)
+
+    // clean tmp dir before run
+    if (tmpExist) {
+      rmSync(this.options.tempDirectory, { recursive: true, force: true })
+      tmpExist = false
+    }
+
+    // make sure tmp dir exist before run
+    if (!tmpExist)
+      mkdirSync(this.options.tempDirectory, { recursive: true })
+
     process.env.NODE_V8_COVERAGE ||= this.options.tempDirectory
   }
 
--- a/packages/coverage-c8/src/provider.ts
+++ b/packages/coverage-c8/src/provider.ts
@@ -1,4 +1,4 @@
-import { existsSync, promises as fs, mkdirSync, rmSync } from 'fs'
+import { existsSync, promises as fs } from 'fs'
 import _url from 'url'
 import type { Profiler } from 'inspector'
 import { takeCoverage } from 'v8'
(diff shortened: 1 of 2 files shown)


packages/coverage-c8/src/provider.ts
```diff
@@ -29,6 +29,18 @@ export class C8CoverageProvider implements CoverageProvider {
   }
 
   onBeforeFilesRun() {
+    let tmpExist = existsSync(this.options.tempDirectory)
+
+    // clean tmp dir before run
+    if (tmpExist) {
+      rmSync(this.options.tempDirectory, { recursive: true, force: true })
+      tmpExist = false
+    }
```
AriPerkkio, line 38: There should be no need to re-implement this logic. It is already done in `clean` function.

https://github.com/vitest-dev/vitest/blob/5b190e8490092f835240b60112fbef449c1af801/packages/coverage-c8/src/provider.ts#L35-L41

This function should be called from `packages/vitest` as it was before `0.22` release. I think it broke somewhere in https://github.com/vitest-dev/vitest/pull/1676.

sun0day: I'm not sure whether `onBeforeFilesRun` supports calling async `clean`. If support, I can change `onBeforeFilesRun` to call async `clean`. Besides, I think it's better to handle `clean` logic inside `C8CoverageProvider`.

sun0day: Or we can call `C8CoverageProvider.clean` in `initCoverageProvider`

https://github.com/vitest-dev/vitest/blob/5b190e8490092f835240b60112fbef449c1af801/packages/vitest/src/node/core.ts#L119-L127

AriPerkkio: Coverage provider is instructed to clean here.

https://github.com/vitest-dev/vitest/blob/5b190e8490092f835240b60112fbef449c1af801/packages/vitest/src/node/core.ts#L108

But it seems that at this point `this.coverageProvider` has not yet been initialized at this point. This is clearly a bug. 

The `this.coverageProvider.clean()` should be moved so that it is called after `this.initCoverageProvider()`. Maybe here into `start`?

https://github.com/vitest-dev/vitest/blob/5b190e8490092f835240b60112fbef449c1af801/packages/vitest/src/node/core.ts#L148-L151

sun0day: Alright. I will modify my commit code

sun0day: done @AriPerkkio

AriPerkkio (approved):

antfu (approved):

PR #1928, opened 2022-08-29
fix!: drop support for Vite 2

antfu:
Given the fact that Vite 3 is out for about two months and most of the ecosystem is finished upgrading, we could drop the compact range for Vite to allow non-vite usage always to install Vite 3 as the dependency.

### DIFF
--- a/.tazerc.json
+++ b/.tazerc.json
@@ -1,6 +1,6 @@
 {
   "exclude": [
-    "vite",
+    "vue",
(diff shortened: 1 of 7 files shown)


dammy001:
@antfu we might need to update the version on docs as well

alexjidras:
@antfu will we still be able to use vitest with vite 2 ?

PR #1929, opened 2022-08-29
feat: introduce `retry` option for tests

### LINKED ISSUE #1057: Flaky test marking and retrying

Often some tests tend to be flaky which is a nightmare in large CI/CD pipelines as you have to retrigger and rerun.

There is not always the time to fix those flaky tests (what would be the ideal case - no doubt)
But in reality it happens.

It would be nice to have a retry mechanism that does not block CI runs.

### Suggested solution

I propose a feature that would mark tests as flaky (and then it would retry them up to a max that is specified when it fails)

like:

```typescript
test.flaky('This test is marked as flaky', () => { /* some flaky logic */ })
```

in the `vitest.config.ts` it could be then configured how often a flaky test should be retried on failure (default to 3 times)

```typescript

export default defineConfig({
  test: {
    flakyTestRetries: 4, // default would be 3
  },
})
```

the test would then be rerun if it fails up to the maximum specified.

It would only do that for tests that are marked as `test.flaky` (Therefore it needs an incentive of the user to say this test is really flaky I know but I don't have time to fix it now)

In the summary of the end it could then be written 3 flaky tests detected in addition to how many are passed or skipped.

### Alternative

Not implement this feature

### Additional context

_No response_

### Validations

antfu:
close #1057

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -9,6 +9,11 @@ The following types are used in the type signatures below
 ```ts
 type Awaitable<T> = T | PromiseLike<T>
 type TestFunction = () => Awaitable<void>
+
+interface TestOptions {
+  timeout?: number
+  retry?: number
+}
 ```
 
 When a test function returns a promise, the runner will wait until it is resolved to collect async expectations. If the promise is rejected, the test will fail.
@@ -19,7 +24,7 @@ In Jest, `TestFunction` can also be of type `(done: DoneCallback) => void`. If t
 
 ## test
 
-- **Type:** `(name: string, fn: TestFunction, timeout?: number) => void`
+- **Type:** `(name: string, fn: TestFunction, timeout?: number | TestOptions) => void`
 - **Alias:** `it`
 
   `test` defines a set of related expectations. It receives the test name and a function that holds the expectations to test.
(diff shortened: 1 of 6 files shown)


sheremet-va:
Can we show anything about retries in terminal for debugging?

antfu:
What would the format be like in your mind? Like `✅ test name (x3)` or `✅ test name (retry x3)`?

sheremet-va:
> What would the format be like in your mind? Like `✅ test name (x3)` or `✅ test name (retry x3)`?

I think we should give brief description about errors, not just the counter.

antfu:
Can you share a more detailed layout/text, or direct push/pr to this branch?

sheremet-va:
> Can you share a more detailed layout/text design, or directly push/pr to this branch?

tbh this question is why I didn't start this PR by myself 😄 

I feel like we need some kind of indicator, but I have poor imagination 😃

antfu:
I will start with indicators and leave it for future improvements then

sheremet-va:
Can we add a global option for this?

antfu:
I am not yet sure if we want the global option for this tbh. We already support describe level and it should be enough for now I think.

sheremet-va (approved):

lnhrdt:
> I am not yet sure if we want the global option for this tbh. We already support describe level and it should be enough for now I think.

Today I looked for a way to configure this globally and came across this PR. I was hoping to configure test retries globally in CI (where resources are limited and it makes my tests flakier). I was hoping to configure it against something like `process.env.CI` but I can't since there's no global configuration for test retries. Would you reconsider providing a global configuration option @antfu?

xsjcTony:
Yeah, that would be awesome if we can have this option globally since CI is just unstable sometimes. The approach that `Playwright` takes is just ✨

giovannivocale-molar:
Is there an option to specify `retry` in the global settings for `vitest`, so that automatically retries X number of times if a test is failing?

sheremet-va:
> Is there an option to specify `retry` in the global settings for `vitest`, so that automatically retries X number of times if a test is failing?

I would recommend opening an enhancement issue if you need something,

PR #1935, opened 2022-08-30
fix: don't throw on null error, hide JSON results warning

sheremet-va:
We might corrupt JSON somehow, if user ends the process, so I think we should just skip it.

### DIFF
--- a/packages/vitest/src/utils/base.ts
+++ b/packages/vitest/src/utils/base.ts
@@ -177,3 +177,21 @@ export function shuffle<T>(array: T[], seed = RealDate.now()): T[] {
 
   return array
 }
+
+// removes stack trace from error, when we don't really need it
+// by default stack trace limit is 100 in vitest, but we might
+// create errors just to get the last line of stack trace behind the scenes
+export function createSimpleError(message?: string, limit = 0) {
+  if (!('stackTraceLimit' in Error)) {
+    const err = new Error(message)
+    if (limit === 0)
+      err.stack = ''
+    return err
+  }
+
+  const oldLimit = Error.stackTraceLimit
+  Error.stackTraceLimit = limit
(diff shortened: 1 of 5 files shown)


antfu:
I am not sure if I understand it, why do we need to limit the stacktrace?

sheremet-va:
> I am not sure if I understand it, why do we need to limit the stacktrace?

I left a comment explaining why. Vitest sets a limit to 100 by default, which is a lot for something we don't use internally. Building stack trace is a heavy operation in Node, as we saw in [Vite](https://github.com/vitejs/vite/pull/6868) for example.

antfu:
> Vitest sets a limit to 100 by default

Should we low it down a bit and make it configurable?

sheremet-va:
> > Vitest sets a limit to 100 by default
> 
> Should we low it down a bit and make it configurable?

User errors are few and always visible, so I don't think it's necessary. The problem is with internal usage of `new Error` for parsing stack trace.

sheremet-va:
> User errors are always visible, so I don't think it's necessary. The problem is with internal usage of `new Error` for parsing stack trace.

But I am not against making it configurable.

packages/vitest/src/utils/base.ts
```diff
@@ -177,3 +177,21 @@ export function shuffle<T>(array: T[], seed = RealDate.now()): T[] {
 
   return array
 }
+
+// removes stack trace from error, when we don't really need it
+// by default stack trace limit is 100 in vitest, but we might
+// create errors just to get the last line of stack trace behind the scenes
+export function createSimpleError(message?: string, limit = 0) {
```
antfu, line 184:
```suggestion
export function createErrorWithTraceLimit(message?: string, limit = 0) {
```

PR #1938, opened 2022-08-30
fix(vite-node): tolerate error in watch mode, fix #1715

### LINKED ISSUE #1715: [vite-node] Do not crash on error in watch mode

If you run vite-node in watch mode (e.g. `vite-node -w app.ts`), any time your code crashes (e.g. saved with invalid syntax) you have to restart `vite-node` manually.

### Suggested solution

If an error (invalid syntax, code exception etc.) is encountered, **do not stop** the `vite-node` process and wait for the next file change to rerun (potentially fixed) code.

### Alternative

Running `vite-node -w app.ts` manually every time you accidentally saved a file in an invalid state.

### Additional context

_No response_

### Validations

andrew-w-ross:
This pr is a partial fix for watch mode in `vite-node`, It allows the user to recover on a bad transform. It however won't attempt to keep a process running if it fails think forever or pm2.


packages/vite-node/src/client.ts
```diff
@@ -129,7 +129,13 @@ export class ViteNodeRunner {
...
   async executeFile(file: string) {
-    return await this.cachedRequest(`/@fs/${slash(resolve(file))}`, [])
+    try {
+      return await this.cachedRequest(`/@fs/${slash(resolve(file))}`, [])
+    }
+    catch (err) {
+      if (this.options.exitOnFailedTransform)
+        throw err
+    }
   }
```
antfu, line 139: I think we should do the error handling on the `cli.ts` caller side instead

andrew-w-ross: @antfu Only getting back to this pr now, let me know if this works for you.

packages/vite-node/src/cli.ts
```diff
@@ -75,8 +75,15 @@ async function run(files: string[], options: CliOptions = {}) {
...
-  for (const file of files)
-    await runner.executeFile(file)
+  for (const file of files) {
+    try {
+      await runner.executeFile(file)
+    }
+    catch (e) {
+      if (!options.watch)
+        throw e
+    }
```
antfu, line 85: I think we should still print out the error and set exitcode

antfu:
And I think we should have an option to make this opt-in.

packages/vite-node/src/cli.ts
andrew-w-ross: Sure any need to set the exit code back to 0 on hmr?

packages/vite-node/src/cli.ts
```diff
@@ -75,8 +75,22 @@ async function run(files: string[], options: CliOptions = {}) {
...
+  for (const file of files) {
+    try {
+      await runner.executeFile(file)
+    }
+    catch (e) {
+      if (options.watch) {
+        process.exitCode = 1
+        // Circular reference throws a TypeError
+        if (e instanceof Error && e.name === 'TypeError')
+          continue
```
antfu, line 87: I am not sure about this. Can you give more specific examples of why this is necessary?

PR #1942, opened 2022-08-31
chore: workaround test-ui ci failure

antfu:
Try to solve test ui hanging. Merge it if all went green.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,7 +107,7 @@ jobs:
       - name: Set node
         uses: actions/setup-node@v3
         with:
-          node-version: 16.x
+          node-version: 16.16
 
       - name: Install
         run: pnpm i
--- a/examples/lit/src/my-button.ts
+++ b/examples/lit/src/my-button.ts
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/no-unused-vars */
 import { LitElement, css, html } from 'lit'
 import { customElement, property } from 'lit/decorators.js'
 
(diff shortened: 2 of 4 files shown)


antfu:
For reference: https://github.com/cypress-io/cypress/issues/22795

sheremet-va:
> For reference: [cypress-io/cypress#22795](https://github.com/cypress-io/cypress/issues/22795)

16.16 doesn't have a fix for node segfaults :S

lmiller1990:
Do you intend to try to support Node.js 16.17 or just recommend against using it? It seems 16.17, specifically noted here: https://github.com/nodejs/node/pull/42623#issuecomment-1223992512 introduced a breaking change.

I am looking at support for Node.js 18.6 for Cypress which has the same issue.

sheremet-va:
> Do you intend to try to support Node.js 16.17 or just recommend against using it? It seems 16.17, specifically noted here: [nodejs/node#42623 (comment)](https://github.com/nodejs/node/pull/42623#issuecomment-1223992512) introduced a breaking change.
> 
> I am looking at support for Node.js 18.6 for Cypress which has the same issue. I will let you know if I find a fix.

Vitest supports Node 16.17 (we have `registerNodeLoader` option, which has `shortCircuit` property), this is just for running Cypress tests 😄

lmiller1990:
Hmm interesting, I think you won't need that anymore once https://github.com/cypress-io/cypress/pull/23637 lands -- we just needed to update our `ts-node` internally. 

I could be mistaken, though. Also -- what do you mean "for running Cypress tests"? Ohhhh - I see you are using Cypress to test the Vitest UI? 

If this is correct that is very cool, I can contribute/help out in some way?

PR #1944, opened 2022-08-31
feat!: improve interchangeability with cjs

### LINKED ISSUE #1939: Dynamic import of TS files from inside node_modules
### Describe the bug

If I do `await import('./user.ts')` from inside a test, it works just fine. If I move this dynamic import into a dependency, it fails with following:

```
TypeError: Unknown file extension ".ts" for ./user.ts
      1| export async function importTest() {
      2|     return await import('./user.ts');
       |            ^
      3| }
      4| 
```

In actual test i am using absolute paths to be sure, simplifying here for readability. The contents of the file as simplified for testing to just `export const User: any = {}`.

I even tried to create a simple `pkg` folder with package.json and a single file exporting that function. Works fine if its outside of node_modules, once I move it there, it fails this way. Doesnt matter if I use `import from 'pkg'` or `import from '../node_modules/pkg/index.mjs'`.

Tried to enable `registerNodeLoader`, that changes the error to following:

```
SyntaxError: Missing initializer in const declaration
```

This is because it is TS file, and it fails on the first TS part. If I remove the type hint (which makes it a valid JS file), it works fine.

Then I finally tried `inline: ['pkg']` and it started working. But this workaround does not work for me in the actual use case, where it breaks the dependency (`@mikro-orm/core`).

### Reproduction

1. Create `user.ts` file with just `export const AuthorListing: any = {}`.
2. Create `node_modules/pkg` folder with `index.mjs` file:

```ts
export async function importTest(file) {
    return await import(file);
}
```

3. Run `importTest('./user.ts')` from a test file.

### System Info

```shell
System:
    OS: macOS 12.5.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 97.61 MB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.8.0 - ~/Library/Caches/fnm_multishells/98665_1661811231718/bin/node
    npm: 8.18.0 - ~/Library/Caches/fnm_multishells/98665_1661811231718/bin/npm
  Browsers:
    Chrome: 104.0.5112.101
    Firefox: 103.0.1
    Safari: 15.6.1
```

### Used Package Manager

yarn

### Validations

sheremet-va:
Fixes #1939


antfu (approved):

sheremet-va:
@antfu made some changes - noticed that new behaviour also triggers for ESM, which is incorrect

Also added usecase, where default is not an object - Node finds named exports, but returnes them as undefined (because tries to access it on `default`): https://stackblitz.com/edit/node-kyzwfi?file=cjs.js,index.js

PR #1945, opened 2022-08-31
feat: improve performance for node loader

sheremet-va:


sheremet-va:
Loader will be deprecated in the future versions.

PR #1952, opened 2022-09-01
fix(dev) fix issue with decoding URI for import expression

### LINKED ISSUE #1937: Import failure on file path with spaces
### Describe the bug

Related issue: https://github.com/unjs/nitro/issues/448

When importing a module path with spaces, vitest will throw `Error: [vite-node] Failed to load`, but it works fine in node env.

### Reproduction

Let's define a simple `test dir/util.mjs`

```js
export default 123
```

This file can be improved via Node.js:

```js
const util = await import("file:///test/test%20dir/util.mjs")
console.log(util) // 123
```

But when we import this file in vitest, we'll see error:

```js
import { describe } from 'vitest'

describe('test:suite', async () => {
  const util = await import("file:///test/test%20dir/util.mjs") // Error: [vite-node] Failed to load /test/test%20dir/util.mjs
  console.log(util)
})
```

### System Info

```shell
System:
    OS: macOS 12.5.1
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.7.0
    npm: 8.15.0
  Browsers:
    Chrome: 104.0.5112.101
  npmPackages:
    vitest: ^0.22.1 => 0.22.1
```

### Used Package Manager

pnpm

### Validations

bartoszgolebiowski:
close #1937

I just add `decodeURI` on the argument for the method responsible for resolving import expression.

I create a directory `test utils`+ one ts file with default export.
My unit test is importing this ts file via import expression and assert default export.
The import expression argument is encoded 
`const util = await import('file:///test/test%20util/helper')`

### DIFF
--- a/packages/vite-node/src/client.ts
+++ b/packages/vite-node/src/client.ts
@@ -15,12 +15,12 @@ export const DEFAULT_REQUEST_STUBS = {
     injectQuery: (id: string) => id,
     createHotContext() {
       return {
-        accept: () => {},
-        prune: () => {},
-        dispose: () => {},
-        decline: () => {},
-        invalidate: () => {},
-        on: () => {},
+        accept: () => { },
+        prune: () => { },
+        dispose: () => { },
+        decline: () => { },
+        invalidate: () => { },
+        on: () => { },
       }
     },
     updateStyle(id: string, css: string) {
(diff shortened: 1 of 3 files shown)


sheremet-va:
I think this should be fixed on Vite side. If you look at the transpiled code, you can see that Vite doesn't resolve this import correctly. (all imports are transformed by Vite first with "analyze" plugin, so Vite should understand the import it transformed, I think)

PR #1963, opened 2022-09-03
feat: allow custom environments

### LINKED ISSUE #1961: Custom environment support

Jest supports running tests in custom environments using the [`testEnvironment` option](https://jestjs.io/docs/configuration#testenvironment-string). I'd like to have a similar feature in Vitest, both to make it easier to migrate from Jest to Vitest, and so users of [my custom environment](https://github.com/nickmccurdy/jest-environment-url) can use Vitest. This would also make it easier to support Vitest in the [Testing Library Recorder Extension](https://github.com/testing-library/testing-library-recorder-extension), since it uses my custom environment.

### Suggested solution

1. Support using custom strings with the `environment` option
   - Optionally, we could also support passing an environment class directly to the option
2. Support using custom strings with `@vitest-environment` (also `@jest-environment` for compatibility)
3. Resolve strings like `jsdom` to `vite-environment-jsdom` (or `jest-environment-jsdom` if we can figure out how to support Jest environments directly)

### Alternative

Use the [adapter pattern](https://sourcemaking.com/design_patterns/adapter) to load Jest environments instead of supporting them directly

### Additional context

_No response_

### Validations

sheremet-va:
Closes #1961


docs/config/index.md
```diff
@@ -235,7 +235,34 @@ test('use jsdom in this test file', () => {
...
+
+Starting from 0.23.0, you can also define custom environment. When non-builtin environment is used, Vitest will try to load package `vitest-environment-${name}`. That package should export an object with the shape of `Environment`:
+
+```ts
+import type { Environment } from 'vitest'
+
+export default <Environment>{
+  name: 'custom',
+  setup() {
+    // cusotm setup
```
Demivan, line 248:
```suggestion
    // custom setup
```

docs/guide/environment.md
```diff
@@ -0,0 +1,55 @@
+# Test Environment
+
+Vitest provides [`environment`](/config/#environment) option to run code inside a specific environment. You can modify how environment behaves with [`environmentOptions`](/config/#environmentoptions) option.
+
+By default, you can use this environments:
```
Demivan, line 5:
```suggestion
By default, you can use these environments:
```

docs/guide/environment.md
```diff
@@ -0,0 +1,55 @@
...
+  }
+}
+```
+
+You also have access to default Vitest environments through `vitest/environments` entry:
+
+```ts
+import { environments, populateGlobal } from 'vitest/environments'
+
+console.log(environments) // { jsdom, happy-dom, node, edge-runtime }
```
antfu, line 35: What is this for? To reuse the built-in env in the custom one? If so, maybe explicitly name it `builtinEnvironments`?

antfu:
Overall direction looks good to me

docs/guide/environment.md
sheremet-va: Yes, to call `setup` directly. Sure, we can rename it, I just didn't touch the original name

antfu: We could rename it on re-exporting.

antfu (approved):

docs/guide/environment.md
```diff
@@ -0,0 +1,55 @@
...
+
+Starting from 0.23.0, you can create your own package to extend Vitest environment. To do so, create package with the name `vitest-environment-${name}`. That package should export an object with the shape of `Environment`:
+
+```ts
+import type { Environment } from 'vitest'
+
+export default <Environment>{
+  name: 'custom',
+  setup() {
+    // cusotm setup
```
Demivan, line 20:
```suggestion
    // custom setup
```

Demivan (approved):

docs/guide/environment.md
sheremet-va: It's fixed already:

https://github.com/sheremet-va/vitest/blob/27f947a4ba40eb277fce3aa01d0225d5105f46fe/docs/config/index.md#L248

What commit are you watching?

Demivan: That was in docs/config/index.md

sheremet-va: Oh, I didn't notice

nickmccurdy (commented):
Nice work! I got busy but thanks for working on this.

nickmccurdy:
Also, please remember to publish this as a breaking change incrementing the minor version.

PR #1984, opened 2022-09-06
fix: parse env options from comments (fix #1962)

### LINKED ISSUE #1962: Environment options are not parsed from comments
### Describe the bug

Vitest supports parsing `@{vitest,jest}-environment` comments to set an environment, but it should also support `@{vitest,jest}-environment-options` comments for [compatibility with Jest](https://jestjs.io/docs/configuration#testenvironmentoptions-object).

#### `index.test.js`
```js
/**
 * @jest-environment jsdom
 * @jest-environment-options {"url": "https://example.com/"}
 */
test("use jsdom and set the URL in this test file", () => {
  expect(location.href).toBe("https://example.com/")
})
```
#### `vite.config.js`
```js
export default {
  test: {
    globals: true,
  },
}
```
#### Expected result from Jest
```
 PASS  ./index.test.js
  ✓ use jsdom and set the URL in this test file (2 ms)
```

#### Actual result from Vitest
```
 FAIL  index.test.js > use jsdom and set the URL in this test file
AssertionError: expected 'http://localhost:3000/' to be 'https://example.com/' // Object.is equality
 ❯ index.test.js:6:25
      4|  */
      5| test("use jsdom and set the URL in this test file", () => {
      6|   expect(location.href).toBe("https://example.com/")
       |                         ^
      7| })
      8| 

  - Expected   "https://example.com/"
  + Received   "http://localhost:3000/"
```

### System Info

```shell
System:
    OS: macOS 12.5.1
    CPU: (4) x64 Intel(R) Core(TM) i5-5257U CPU @ 2.70GHz
    Memory: 43.91 MB / 8.00 GB
    Shell: 3.5.1 - /usr/local/bin/fish
  Binaries:
    Node: 18.8.0 - /usr/local/bin/node
    npm: 8.19.1 - /usr/local/bin/npm
    Watchman: 2022.08.29.00 - /usr/local/bin/watchman
  npmPackages:
    vitest: ^0.18.0 => 0.18.1
```

### Used Package Manager

npm

### Validations

nickmccurdy:


packages/vitest/src/runtime/entry.ts
```diff
@@ -22,45 +21,29 @@ export async function run(files: string[], config: ResolvedConfig): Promise<void
...
     const env = code.match(/@(?:vitest|jest)-environment\s+?([\w-]+)\b/)?.[1] || config.environment || 'node'
+    const envOptions = JSON.parse(code.match(/@(?:vitest|jest)-environment-options\s+?(.+)/)?.[1] || 'null')
     return {
       file,
       env: env as VitestEnvironment,
+      envOptions: envOptions as EnvironmentOptions | undefined,
     }
   }))
 
-  const filesByEnv = filesWithEnv.reduce((acc, { file, env }) => {
```
sheremet-va, line 31: Just a hint. It is required to group files by env for `--no-threads`. Now, you need to group them by env AND options: https://vitest.dev/config/#environment

> If you are running Vitest with [--no-threads](https://vitest.dev/config/#threads) flag, your tests will be run in this order: node, jsdom, happy-dom, edge-runtime, custom environments. Meaning, that every test with the same environment is grouped together, but is still running sequentially.

nickmccurdy: Out of curiosity, is there a specific reason we need to sort by configs, and not just envs?

sheremet-va: Well, if environment has different config, it's a different environment, isn't it? If you only group by environments, then only one config will be applied.

nickmccurdy:
I seem to have fixed the bug, but I'm not sure if I have the environment option merging right, and things could probably be cleaner in the long run (see #1987).

packages/vitest/src/runtime/entry.ts
```diff
@@ -45,22 +52,31 @@ export async function run(files: string[], config: ResolvedConfig): Promise<void
...
-      for (const file of files) {
+    const filesByOptions = groupBy(files, ({ envOptions }) => JSON.stringify(envOptions))
+
+    for (const options of Object.keys(filesByOptions)) {
+      const files = filesByOptions[options]
+
+      if (!files || !files.length)
+        continue
+
+      await withEnv(environment, JSON.parse(options) || config.environmentOptions || {}, async () => {
```
sheremet-va, line 63: I think we can get parsed options from `files[0]`? To remove overhead of parsing.

packages/vitest/src/runtime/entry.ts
```diff
@@ -45,22 +52,31 @@ export async function run(files: string[], config: ResolvedConfig): Promise<void
     if (!files || !files.length)
       continue
 
-    await withEnv(environment, config.environmentOptions || {}, async () => {
-      for (const file of files) {
+    const filesByOptions = groupBy(files, ({ envOptions }) => JSON.stringify(envOptions))
```
sheremet-va, line 55: What about this case?

```ts
// file1.ts
// @vitest-environment custom
// @vitest-environment-options {"option1":1,"option2":2}

// file2.ts
// @vitest-environment custom
// @vitest-environment-options {"option2":2,"option1":1}
```

These files should run in the same environment, because they have the same values.

sheremet-va (changes requested):

packages/vitest/src/runtime/entry.ts
nickmccurdy: Personally I don't think that's beneficial enough to be worth the downsides. We would have to make the equality logic much more complex. Also from what I understand, the JSON and ECMAScript spec do not specify whether parsers need to be ordered, so we would need to herd code different logic for every JavaScript runtime and environment.

If you wanted to ensure better performance, you could simple sort your kegs with a linter.

sheremet-va: > Also from what I understand, the JSON and ECMAScript spec do not specify whether parsers need to be ordered, so we would need to hard code different logic for JavaScript runtimes and environments.

I think you overcomplicate things here, I'm sure we can create a string key that doesn't depend on key order. Sorting keys seems like an obvious solution. 

> If you wanted to ensure better performance as a user, you could simply sort your keys with a linter. 

I don't think you can store them with linter in our docblock?

> Would something break if we made the wrong assumption about option equality for grouping envs?

Most things that expect single JSOM instance will break, because we don't clear ESM cache in `--no-threads`: Vue, for example, stores "document" globally on import, so next tests with different "options" will fail, because global "document" will be different. This is actually why we have grouping.

I am sure this usecase will emerge, if people will use `@vitet-environment-options`.

nickmccurdy: > > Also from what I understand, the JSON and ECMAScript spec do not specify whether parsers need to be ordered, so we would need to hard code different logic for JavaScript runtimes and environments.
> 
> I think you overcomplicate things here, I'm sure we can create a string key that doesn't depend on key order. Sorting keys seems like an obvious solution.

That sounds like a good place to start, but I still feel like that's a lot of time complexity.

> > Would something break if we made the wrong assumption about option equality for grouping envs?
> 
> Most things that expect single JSOM instance will break, because we don't clear ESM cache in `--no-threads`: Vue, for example, stores "document" globally on import, so next tests with different "options" will fail, because global "document" will be different. This is actually why we have grouping.

So does that mean you can't have multiple tests with different options for the same env when using Vue with `--no-threads`?

sheremet-va:
Let's merge it now and see, if the order is really a problem later.

PR #2014, opened 2022-09-09
docs: fix `vi.fn` signature

kidonng:
Make the argument optional as described:

> Creates a spy on a function, though can be initiated without one.


kidonng:
> [test/basic.test.ts > basic > should have the correct title](https://github.com/vitest-dev/vitest/runs/8263594146?check_suite_focus=true#step:7:177)

Looks like some flaky test needs to resolved.

PR #2015, opened 2022-09-09
fix typo jsdon -> jsdom

haikyuu:


haikyuu:
@sheremet-va not sure why this is closed. It's just a typo

patak-dev:
@haikyuu it was a duplicate of https://github.com/vitest-dev/vitest/pull/2011 (that was the first one and got merged). There were actually 3 PRs to fix the same typo

haikyuu:
Glad to hear that 😁 thank you for the explanation

PR #2027, opened 2022-09-12
fix: use correct source maps in stacktrace

### LINKED ISSUE #2024: Error shown with line:column number that are off (source map not used)
### Describe the bug

Logged failing tests have the wrong line:column number.

### Reproduction

https://github.com/imba/imba-vite-template/tree/vitest-repro

Install dependencies and run npm test to see failing tests.

### System Info

```shell
System:
    OS: macOS 12.4
    CPU: (10) arm64 Apple M1 Max
    Memory: 197.63 MB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    npm: 8.5.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 105.0.5195.102
    Chrome Canary: 107.0.5293.0
    Safari: 15.5
  npmPackages:
    vite: ^3.1.0 => 3.1.0 
    vitest: ^0.23.1 => 0.23.1
```

### Used Package Manager

npm

### Validations

haikyuu:
use vitenode server to get a fetch result instead of a transform result.
And then get the sourcemap from it.

This fixes #2024


packages/vitest/src/utils/source-map.ts
```diff
@@ -25,10 +25,14 @@ export async function interpretSourcePos(stackFrames: ParsedStack[], ctx: Vitest
   for (const frame of stackFrames) {
     if ('sourcePos' in frame)
       continue
-    const transformResult = ctx.server.moduleGraph.getModuleById(frame.file)?.ssrTransformResult
-    if (!transformResult)
+    const ssrTransformResult = ctx.server.moduleGraph.getModuleById(frame.file)?.ssrTransformResult
+    const transformResult = ctx.server.moduleGraph.getModuleById(frame.file)?.transformResult
+    const result = ssrTransformResult || transformResult
```
sheremet-va, line 30: Why do we need this, if it's not used later? Maybe you wanted to put it on L#34?

sheremet-va (changes requested):
Can you add a test for this, please?

packages/vitest/src/utils/source-map.ts
haikyuu: It's only used in the check `if (!result) continue`
The rationale behind it is that there is no need to calculate the position if the file hasn't been transformed.

And sometimes, the file has been transformed but `ssrTransformResult` is unedfined while `transformResult` isn't.

haikyuu: The reason I didn't put it in line `#34` is that it has the wrong sourcemap. 

I think it's the sourcemap of transforming files to run in vite-node context where imports are mapped differently, and that's not the only one, so it has different sourcemaps.

haikyuu:
> Can you add a test for this, please?

Sure, is there a test example for errors/stacktraces I can check? @sheremet-va

packages/vitest/src/utils/source-map.ts
sheremet-va: But it means `map` on L35 can be `undefined`? I don't really see a point in this check, tbh. It makes more sense to `continue` on L34, if map is not defined.

sheremet-va:
> > Can you add a test for this, please?
> 
> Sure, is there a test example for errors/stacktraces I can check? @sheremet-va

To be honest, I don't know.

haikyuu:
The test fails because of this error https://github.com/vitest-dev/vitest/runs/8301687675?check_suite_focus=true#step:7:668

```
Error: Failed to parse source for import analysis because the content contains invalid JS syntax. 
You may need to install appropriate plugins to handle the .imba file format, 
or if it's an asset, add \\"**/*.imba\\" to \`assetsInclude\` in your configuration.
```

I used my complete plugin to generate the snapshot, but I created a small version of the plugin just for the test. But I don't understand why Vite is complaining about this error. I'd use some help here 🆘

sheremet-va:
> I used my complete plugin to generate the snapshot, but I created a small version of the plugin just for the test. But I don't understand why Vite is complaining about this error. I'd use some help here 🆘

Try adding `enforce: 'pre'`

haikyuu:
Thanks @sheremet-va I was including extra quotes because I copied the value from the debugger. Fixed now :)

haikyuu:
@sheremet-va The tests are failing because of another error  in https://github.com/vitest-dev/vitest/blob/main/test/fails/fixtures/timeout.test.ts#L4 that's not related to my PR

```
 Snapshots  1 failed
  - Expected   ""Error: Test timed out in 10ms.""
  + Received   ""Error: Hook timed out in 10ms.""
```

packages/vitest/src/utils/source-map.ts
haikyuu: Makes sense @sheremet-va Done 👍

test/stacktraces/test/runner.test.ts
```diff
@@ -0,0 +1,41 @@
...
+      })
+        .catch((e) => {
+          error = e
+        })
+
+      expect(error).toBeTruthy()
+      const msg = String(error)
+        .split(/\n/g)
+        .reduce((acc, line) => {
+          if (line.includes('Start at') || line.includes('Duration') || line.includes('(1 test | 1 failed)') || line.includes(root))
```
sheremet-va, line 33: Maybe we should only include this part in snapshot?

Screenshot 2022-09-13 at 10 36 57

sheremet-va (changes requested):

test/stacktraces/test/runner.test.ts
haikyuu: Agree, will do

haikyuu: Done, they look nicer now

haikyuu:
@sheremet-va now one test which is not related to my changes seems flaky: https://github.com/vitest-dev/vitest/blob/main/test/fails/test/runner.test.ts

```
  - Expected   ""Error: Test timed out in 10ms.""
  + Received   ""Error: Hook timed out in 10ms.""
```

haikyuu:
@sheremet-va can you take a look again? I fixed the previous failing test. But I'm not sure why single threaded tests are failing in Ubuntu?

sheremet-va (approved):

PR #2039, opened 2022-09-14
feat(benchmark): only mode

Aslemammad:
I'll address bench todo mode and reporters' issues in the next PRs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "packageManager": "pnpm@7.11.0",
   "description": "A blazing fast unit test framework powered by Vite",
   "scripts": {
-    "ci": "ni && nr typecheck && nr lint && nr build && nr test:all && nr bench",
+    "ci": "ni && nr typecheck && nr lint && nr build && nr test:all && nr bench:all",
     "build": "pnpm -r --filter=./packages/** run build",
     "dev": "cross-env NODE_OPTIONS=\"--max-old-space-size=8192\" pnpm -r --parallel --filter=./packages/** run dev",
     "docs": "pnpm -C docs run dev",
@@ -19,7 +19,7 @@
     "lint": "eslint .",
     "lint:fix": "nr lint --fix",
     "release": "bumpp package.json packages/*/package.json --commit --push --tag && pnpm -r publish --access public",
-    "bench": "cross-env CI=true pnpm -r --stream run bench --allowOnly",
+    "bench:all": "cross-env CI=true pnpm -r --stream run bench --allowOnly",
     "test": "vitest --api -r test/core",
     "test:run": "vitest run -r test/core",
     "test:all": "cross-env CI=true pnpm -r --stream run test --allowOnly",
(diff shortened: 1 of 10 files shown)


test/benchmark/test/only.bench.ts
```diff
@@ -0,0 +1,68 @@
+import assert from 'node:assert'
```
sheremet-va, line 1: this file is not running by `test` script

and even then, it will never fail, because it just checks json

sheremet-va (changes requested):

antfu (approved):

test/benchmark/test/only.bench.ts
```diff
@@ -0,0 +1,59 @@
+import { bench, describe, expect } from 'vitest'
```
sheremet-va, line 1: This file is still not running, I just put the previous comment on the first line. Please, ensure that your tests are passing before requiring review.

sheremet-va (changes requested):

test/benchmark/test/only.bench.ts
Aslemammad: I see, I didn't know these benchmarks are not running, so misunderstood you! Let me see.

Aslemammad: I see, I didn't know these benchmarks are not running, so misunderstood you! Let me see.

test/benchmark/package.json
```diff
@@ -2,8 +2,6 @@
   "name": "@vitest/benchmark",
   "private": true,
   "scripts": {
-    "test": "node test.mjs",
-    "testu": "vitest -u",
-    "coverage": "vitest run --coverage"
+    "bench": "vitest bench"
```
sheremet-va, line 5: Benchmarks never fail. I specifically wrote `test.mjs` to correctly test benchmarks output, and now two other files are not tested.

package.json
```diff
@@ -19,6 +19,7 @@
     "lint": "eslint .",
     "lint:fix": "nr lint --fix",
     "release": "bumpp package.json packages/*/package.json --commit --push --tag && pnpm -r publish --access public",
+    "bench:all": "cross-env CI=true pnpm -r --stream run bench --allowOnly",
```
sheremet-va, line 22: We don't need this, we should only run tests in CI, it doesn't make sense to "just" run benchmarks in CI, we should test benchmark functionality, which is not testable with `vitest bench`.

sheremet-va (changes requested):

test/benchmark/package.json
sheremet-va: Try to run a failing script first. Only after you ensured that tests __can actually__ fail CI, run an actual test.

Right now you are making _assumptions_ that something works, but reality disagrees.

Aslemammad: Ok, Let me analyze test.js.

sheremet-va (approved):

sheremet-va (changes requested):
I just noticed, there is no documentation for this feature. Please, add a short documentation alongside `bench.skip` (already in the docs)

sheremet-va (approved):

PR #2064, opened 2022-09-20
fix: align `.each` behavior with jest

### LINKED ISSUE #1858: test.each's behaviour deviates from jest's
### Describe the bug

As noted previously in https://github.com/vitest-dev/vitest/issues/1461, there are significant differences in `test.each()` between `vitest` and `jest`.

### Reproduction

1. When invoking `test.each()` with an array containing null, e.g. `test.each([null])`, the value will be processed as an array containing null (`[null]`) on Jest and converted to a `null` on Vitest.

Vitest used to return an empty array in that scenario, but this was changed in https://github.com/vitest-dev/vitest/pull/1462) and is enforced in this test:

```js
test.each([
  null,
  [null],
])('null is null', (value) => {
  expect(value).toBe(null)
})
```
https://github.com/vitest-dev/vitest/blob/main/test/core/test/each.test.ts#L11

However, the same test will fail in Jest:
```
 FAIL  core/test/each.test.ts
  ●  null is null
    expect(received).toBe(expected) // Object.is equality

    Expected: null
    Received: [null]

      43 |       [null],
      44 |     ])('null is null', (value) => {
    > 45 |       expect(value).toBe(null)
         |                     ^
      46 |     })
      47 |
      48 |     it.each([true, false])(

      at core/test/each.test.ts:45:21

Test Suites: 1 failed, 0 passed, 1 total
Tests:       1 failed, 0 skipped, 0 passed, 1 total
```

2. In addition, empty arrays are treated as empty arrays in Jest and `undefined` in Vitest:

```js
test.each([
  true, false,
  0, 1234,
  new Date(),
  [], {}, () => {},
])('primitives', (value) => {
  console.log(value, typeof value);
});
```

On Jest, the output is:
```
      true boolean
      false boolean
      0 number
      1234 number
      2022-08-15T21:52:20.000Z object
      [] object // <-- Difference
      {} object
      [Function (anonymous)] function
```

On Vitrest, the output is:
```
      true boolean
      false boolean
      0 number
      1234 number
      2022-08-15T21:52:20.000Z object
      undefined undefined // <-- Difference
      {} object
      [Function (anonymous)] function
```

### System Info

```shell
Binaries:
  Browsers:
    Chrome: 104.0.5112.79
    Safari: 15.6
  npmPackages:
    vite: 3.0.x => 3.0.7 
    vitest: 0.21.x => 0.21.1
```

### Used Package Manager

npm

### Validations

bartoszgolebiowski:
close #1858

### DIFF
--- a/test/core/test/each.test.ts
+++ b/test/core/test/each.test.ts
@@ -9,13 +9,6 @@ test.each([
 })
 
 test.each([
-  null,
-  [null],
-])('null is null', (value) => {
-  expect(value).toBe(null)
-})
-
-test.each([
   ['string', true],
   ['string', false],
 ])('can be parsed', (a, b) => {
@@ -135,4 +128,37 @@ describe('context with each - concurrent', () => {
       expect(number1 + number2).toBe(number3)
     })
   })
+})
+
+describe('not all arguments are array describe.each', () => {
+  const results = [null, [null]]
+  let i = 0
+
+  describe.each([null, [null]])('null is null', (value) => {
+    test('null is null', () => {
+      expect(value).toEqual(results[i++])
+    })
+  })
+})
+
+describe('not all arguments are array test.each', () => {
+  const results = [null, [null]]
+  let i = 0
+
+  test.each([null, [null]])('null is null', (value) => {
+    expect(value).toEqual(results[i++])
+  })
+})
+
+test.each([
+  null,
+])('value is null', (value) => {
+  expect(value).toBeNull()
+})
+
+test.each([
+  [null, null],
+  [null, null],
+])('if all cases are arrays of equal length, treats array elements as arguments', (value) => {
+  expect(value).toBeNull()
 })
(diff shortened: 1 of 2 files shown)


test/core/test/each.test.ts
```diff
@@ -136,3 +129,36 @@ describe('context with each - concurrent', () => {
...
+test.each([
+  null,
+])('value is null', (value) => {
+  expect(value).toBeNull()
+})
+
+test.each([
+  [null, null],
+  [null, null],
+])('if all cases are arrays of equal length, treats array elements as arguments', (value) => {
```
sheremet-va, line 162: Shouldn't we check both `(value1, value2)`?

test/core/test/each.test.ts
```diff
@@ -136,3 +129,36 @@ describe('context with each - concurrent', () => {
     })
   })
 })
+
+describe('not all arguments are array describe.each', () => {
+  const results = [null, [null]]
+  let i = 0
+
+  describe.each([null, [null]])('null is null', (value) => {
```
sheremet-va, line 137: maybe there are better names for these tests? there are 3 tests named `null is null`, but they check `[null] === [null]` 🤔

sheremet-va (changes requested):

antfu (approved):

PR #2079, opened 2022-09-24
trigger beforeFullReload hook before full-reload

ycmjason:


sheremet-va:
@ycmjason do you think you can add other missing events?

ycmjason:
@sheremet-va done. hope that's fine.

sheremet-va (approved):

PR #2101, opened 2022-09-30
docs: more resources

rwieruch:
Not sure whether this is wanted in the documentation, however, when I set up Vitest with React Testing Library, I didn't find any resources about it.

Context: [The Road to React](https://www.amazon.com/dp/B077HJFCQX) is using Vite/Vitest instead of CRA/Jest now 🎆 Hence I needed to come up with a short guide which helps readers to install React Testing Library in Vitest.


antfu:
Thank you. But I think the docs are not the right place for introducing community resources. Maybe we could have an awesome-vitest at some point in a separate repo or so.

PR #2107, opened 2022-10-01
feat: add typechecking functionality

### LINKED ISSUE #1954: [RFC] Testing types with expectTypeOf and assertType

Currently users have to come up with their own solutions to test types. Fo example, Vue defines a set of [small utils](https://github.com/vuejs/core/blob/7c8f4578e9e7178e326cf8e343f7a8b4143ba63b/test-dts/index.d.ts):

```ts
export function expectType<T>(value: T): void
export function expectError<T>(value: T): void
export function expectAssignable<T, T2 extends T = T>(value: T2): void
```

This is not helpful for people who are coming into the world of types and testing their types. We want to provide a better developer experience.

### Suggested solution

Vitest could provide an API for testing TypeScript types, since Vitest can already _run_ typescript files.

Vitest can provide API, similar to `expect`, inspired by [`expect-type`](https://www.npmjs.com/package/expect-type) package:

```ts
expectTypeOf({a: 1}).toEqual<{a: number}>()
expectTypeOf({a: 1}).not.toMatch({b: 1})
```

Vitest can also provide API, similar to what Vue is using:

```ts
assertType<Type>(obj)
// @ts-expect-error
assertType<Type>(obj)
```

That way Vitest can provide similar set of tools that it already provides for runtime tests (`expect` and `assert` APIs).

Vitest will not fail, if tests are running with `vitest` and there is a type error, because there is no type checking involved, when running regular tests.

Vitest will fail, if tests are running with `vitest type` and there is a type error. When running `type` command, Vitest will run `tsc` under the hood. There will be an option in config to change command:

```jsonc
{
  test: {
    type: {
      command?: string // tsc --noEmit by default
      // potentially other options?
    }
  }
}
```

It will be recommended to have two pipelines for testing your application (currently this practice already exists, but tests are running with `tsc` or similar tool directly).

sheremet-va:
Currently implemented running typechecking and showing errors. Exports `expectTypeOf`.

- [x] Determines the group `expect` is in
- [x] Don't show "skip"/"todo" tests
- [x] Show only "only" tests, if suite has "only" flag
- [x] Add options to ignore source type errors
- [ ] ~Support shortcuts~
- [x] Show list of potentially tested files (`include`)
- [x] Allow running `vitest typecheck` even, if there are no potential tests (vitest still shows source errors)
- [x] Allow checking JS files with `@ts-check`
- [x] Documentation
- [x] Tests

Closes #1954


docs/config/index.md
```diff
@@ -849,3 +849,58 @@ Vitest usually uses cache to sort tests, so long running tests start earlier - t
 - **Default**: `Date.now()`
 
 Sets the randomization seed, if tests are running in random order.
+
+### typechecker
```
antfu, line 853:
```suggestion
### typecheck
```

Seems a typo?

Demivan: And on `docs/guide/testing-types.md` page

docs/api/index.md
```diff
@@ -472,6 +509,10 @@ When you use `test` or `bench` in the top level of file, they are collected as p
   describe.todo.concurrent(/* ... */) // or describe.concurrent.todo(/* ... */)
   ```
 
+::: warning
+You cannot use this syntax, when using Vitest as [type checker](/guide/testing-types).
+:::
```
antfu, line 514: This feels a bit redundent to repeat, maybe we just mention you can't use them in the `test-types.md` instead?

sheremet-va: I think we try to write self contained docs, no? It feels kind of weird to me to move a warning that you can't use certain API from its API page 🤔

sheremet-va:
@mmkal hello 👋🏻 I don't know if you saw our proposal about using `expect-type` in Vitest, and I was wondering if you have any thoughts on this proposal. 

We are thinking about inlining you package entirely to have more control over error messages and such. Do you think, you can merge https://github.com/mmkal/expect-type/pull/13 for at least partial support?

And maybe you can also review documentation, I mentioned error messages there. Thank you 😄

docs/guide/testing-types.md
```diff
@@ -0,0 +1,81 @@
...
+Unfortunately, TypeScript doesn't provide type metadata without patching, so we cannot provide useful error messages at this point, but there are <a href="https://github.com/microsoft/TypeScript/pull/40468" tatger="_blank">works in TypeScript project</a> to fix this. If you want better messages, please, ask TypeScript team to have a look at mentioned PR.
+
+If you find it hard working with `expectTypeOf` API and figuring out errors, you can always use more simple `asserType` API:
+
+```ts
+const answer = 42
+
+assertType<number>(answer)
+// @ts-expect-error answer is not a string
+assertType<string>(answer)
```
mrazauskas, line 53: Just to add my two cents: in type testing `// @ts-expect-error` comment should be considered an anti pattern, because it hides any kind of error. For instance, a test is a type is passing:

```ts
// @ts-expect-error answer is not a string
 assertType<string>(answr)
```

Or, for example, `assertType` can be renamed or removed, but the test will be perfectly passing. Ouch... Perhaps executing the file could be a solution? `expect-type` has no runtime, but Vitest has one.

sheremet-va: Yeah, but if you don't use this pattern (which is not a recommended way to do it, we encourage `expect-type`), then we just spent a lot of time by running code that could've been just statically analysed.

If you want, you can always add both runtime and type tests to both `test.include` and `test.typecheck.include`

mrazauskas: Ah right! Could be very useful note to add. (Ignore me, if it is already here and I missed it)

sheremet-va: Added it to the docs, thanks for your input 👍🏻

mmkal:
> @mmkal hello 👋🏻 I don't know if you saw our proposal about using `expect-type` in Vitest, and I was wondering if you have any thoughts on this proposal. 
> 
> 
> 
> We are thinking about inlining you package entirely to have more control over error messages and such. Do you think, you can merge https://github.com/mmkal/expect-type/pull/13 for at least partial support?
> 
> 
> 
> And maybe you can also review documentation, I mentioned error messages there. Thank you 😄 

I'll try to take another look this weekend, there is another open PR which is also an attempt to improve error messages. Is there a particular docs file you'd like me to look at here in the meantime? There are 50+ changed 😬

sheremet-va:
> Is there a particular docs file you'd like me to look at here in the meantime?

This file: [docs/guide/testing-types.md](https://github.com/vitest-dev/vitest/pull/2107/files#diff-58c52997981aa78c64ba261e445d5f5866f73205518e556a06beaa2d11fbcbff)

Particularly "Reading Errors" section. 😄

antfu (approved):

PR #2112, opened 2022-10-03
docs: add a NuxtJS example

r9mp:
Hi,

This PR adds a NuxtJS example in the examples/ directory.

Don't hesitate if you need me to change / improve stuff.


antfu (changes requested):
Please

- Apply the linter
- Use pnpm instead of yarn, and omit the lock

r9mp:
@antfu I solved the linter problem, and used pnpm ; but the ci tests are still down. Il seems that there is en import problem of my vue components. Maybe the auto-import of vue component by vue is ko...
I don't know how to fix it. If you have ideas!!

sheremet-va:
> @antfu I solved the linter problem, and used pnpm ; but the ci tests are still down. Il seems that there is en import problem of my vue components. Maybe the auto-import of vue component by vue is ko... I don't know how to fix it. If you have ideas!!

Your `package.json` doesn't have all needed dependencies. Don't rely on pnpm workspaces support for examples, they can be initialized without the main `package.json` via stackblitz.

Issue #2134, opened 2022-10-11
Errors in benchmarks are displayed as `Unknown Error: undefined`


nickmccurdy:
### Describe the bug

It seems like if a benchmark throws an error, Vitest doesn't properly print its name and stack trace, as I'd expect from a Vitest test.

### Reproduction

```js
import { bench } from "vitest";

bench("Throwing error", () => {
  throw new Error("Example");
});
```
```
 FAIL  test/basic.bench.ts [ test/basic.bench.ts ]
Error: Transform failed with 1 error:
/project/sandbox/test/basic.bench.ts:3:30: ERROR: Unexpected "throw"

 RERUN  test/basic.bench.ts x3

 ✓ test/basic.bench.ts (1) 775ms
   · Throwing error  218.04 ops/sec ±0.71% (109023 samples) fastest

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Failed Suites 1 ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯

 FAIL  test/basic.bench.ts [ test/basic.bench.ts ]
Unknown Error: undefined


⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯
```
https://codesandbox.io/p/sandbox/vitest-benchmark-throwing-error-s1uxfk?file=%2Ftest%2Fbasic.bench.ts

### System Info

```shell
System:
    OS: Linux 5.4 Debian GNU/Linux 11 (bullseye) 11 (bullseye)
    CPU: (4) x64 AMD EPYC
    Memory: 380.22 MB / 1.95 GB
    Container: Yes
[...]

PR #2137, opened 2022-10-11
fix(bench): throw root cause

azaleta:
linked issue: #2134

### DIFF
--- a/packages/vitest/src/runtime/run.ts
+++ b/packages/vitest/src/runtime/run.ts
@@ -369,13 +369,9 @@ async function runBenchmarkSuite(suite: Suite) {
         }
       })
       benchmark.task!.addEventListener('error', (e) => {
-        let causeError
         const task = e.task
         const _benchmark = benchmarkMap[task.name || '']
-        if (_benchmark)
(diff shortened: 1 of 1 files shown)


packages/vitest/src/runtime/run.ts
```diff
@@ -369,7 +369,13 @@ async function runBenchmarkSuite(suite: Suite) {
...
       })
       benchmark.task!.addEventListener('error', (e) => {
-        defer.reject(e)
+        let causeError
+        const task = e.task
+        const _benchmark = benchmarkMap[task.name || '']
+        if (_benchmark)
+          causeError = task.result!.error
+
+        defer.reject(causeError || e)
```
nickmccurdy, line 378: I think this is easier to read with a slightly more declarative style:
```suggestion
        const task = e.task
        const _benchmark = benchmarkMap[task.name || '']
        defer.reject(_benchmark ? task.result!.error : e)
```

nickmccurdy (changes requested):
Thanks, the output looks good! I just have a minor comment on code style.

sheremet-va (approved):

Issue #2154, opened 2022-10-14
Unexpected test failure output "[Function spy]" with vi.fn()


danawoodman:
### Describe the bug

When using a mocked function via `vi.fn(...)` the failure output is unintuitive and makes it near impossible to understand what the failure condition was when using `expect(mock).toHaveReturnedWith(...)`.

Rather than the actual return value, the error output is `"[Funciton spy]"` which makes it impossible to know what the spy actually returned with.

### Reproduction

```ts
const mock = vi.fn(() => 1)
mock()
expect(mock).toHaveReturnedWith(2)

// same result with:
// const mock = vi.fn().mockReturnValue(1)
```

Results in:

```
  - Expected   "2"
  + Received   "[Function spy]"
```

When the expected outcome is:

```
  - Expected   "2"
  + Received   "1"
```

Current workaround:

```ts
expect(handler.calls[0][0]).toBe(1)
```

### System Info

```shell
System:
    OS: macOS 12.6
    CPU: (8) x64 Apple M2
    Memory: 40.22 MB / 24.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.9.0 - ~/.asdf/installs/nodejs/18.9.0/bin/node
    Yarn: 1.22.19 - ~/.asdf/installs/nodejs/18.9.0/.npm/bin/yarn
    npm: 8.19.1 - ~/.asdf/plugins/nodejs/shims/npm
  Browsers:
    Brave Browser: 105.1.43.89
    Chrome: 106.0.5249.119
    Firefox: 103.0
    Safari: 16.0
  npmPackages:
    vitest: 0.24.3 => 0.24.3
```

### Used Package Manager

yarn

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
[...]

PR #2163, opened 2022-10-17
fix(vitest): return detailed error when using toHaveReturnedWith

azaleta:
Linked issue: #2154

1. repalce Expected/Received with more detailed info
2. I think the return  error message is not correct
    `to be successfully called with #{exp}` refers to args
3. For long return array, truncate them for a clean output format


packages/vitest/src/integrations/chai/jest-expect.ts
```diff
@@ -589,11 +589,13 @@ export const JestChaiExpect: ChaiPlugin = (chai, utils) => {
...
     const pass = spy.mock.results.some(({ type, value: result }) => type === 'return' && jestEquals(value, result))
+    const returns = spy.mock.results.filter(({ type }) => type === 'return').map(({ value: result }) => result).toString()
     this.assert(
       pass,
-      `expected "${spyName}" to be successfully called with #{exp}`,
-      `expected "${spyName}" to not be successfully called with #{exp}`,
-      value,
+      `expected "${spyName}" to be successfully called having return #{exp} at least once`,
+      `expected "${spyName}" to not be successfully called having return #{exp}`,
+      `executions have a successful return value: ${value}`,
```
sheremet-va, line 597: `{}` will return `[Object Object]`

Please, don't reinvent the wheel.

sheremet-va (changes requested):
Vitest already provides a list of calls, when using `toHaveBeenCalled`, just reuse it with modifications in other assertions.

packages/vitest/src/integrations/chai/jest-expect.ts
azaleta: try to use chai untils to format message.

sheremet-va: Please, look at how `toHaveBeenCalledWith` is processing errors. Try to use the same approach, it gives more useful information.

azaleta: May I ask a question?
According the `toHaveBeenCalledWith` which use a util `formatCalls` to specify call stack.
I think a new until such as `formatReturns` should be added to format returns for all calls.
Is this what you want?

sheremet-va: Yes. It would look basically like `formatCalls`, but uses `returns` instead of `calls`

test/core/test/mocked.test.ts
```diff
@@ -134,6 +135,130 @@ test('async functions should be mocked', () => {
   expect(asyncFunc()).resolves.toBe('foo')
 })
 
+describe('mocked function which fails on toReturnWith', () => {
+  test('zero call', () => {
+    const mock = vi.fn(() => 1)
+    try {
```
sheremet-va, line 141: Lets' rewrite these tests with `toThrow`:

```ts
expect(() => expect(mock).toReturnWith(2)).toThrowErrorMatchingInlineSnapshot()
```

Two reasons:

- 1) less code
- 2) more reliable code (if first assertion won't fail, we will not make assertions inside a "catch", but the test will pass)

sheremet-va (changes requested):

Issue #2178, opened 2022-10-20
Support test file path with leading "./" or ".\" segments


kleinfreund:
### Clear and concise description of the problem

In some environments (e.g. PowerShell on Windows), auto completion for relative paths **will always** add a dot segment (e.g. "./" or ".\") at the start of the path. For example, in PowerShell, writing `npx vitest run t` and pressing <kbd>Tab</kbd> while inside a directory that contains a file or directory starting with `t` will provide auto completion entries for that object **and** add `.\` to the start of the path.

![Screenshot showing a PowerShell terminal window with the buffer containing `npx vitest run t` and the respective auto completion results](https://user-images.githubusercontent.com/5774638/197006894-42721d82-fb8a-432b-8cb4-599c714d7380.png)

### Suggested solution

Vitest should process and allow path segments starting with dot segments.

It currently doesn't and reports "No test files found, exiting with code 1" instead.

### Additional context

I've tried looking through the Vitest code base to possibly work on an implementation of this myself, but while I found related code, I didn't find any tests for CLI interactions or the CLI API starting Vitest with test files and am not sure how I would possibly test such a contribution.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #2180, opened 2022-10-20
fix(cli): normalize argument path

poyoho:
fix: #2178


sheremet-va (changes requested):
Can you add a test case for this?

Issue #1839, opened 2022-08-11
Coverage with HTML reporter (c8) for .vue + .ts files, no SFC


OrestHk:
### Describe the bug

Hi, updating vitest from `0.16` to `0.21.1` I found out the following

`yarn vitest --coverage` with an `html` reporter will throw an error when the component that is tested is not a SFC (.vue + .ts), as the reporter will list the ts file as `file.ts?vue&type=script&src=true&lang.ts`

The same error occurs for a project using vue@2.7, @vue/test-utils@1 with the last version of vitest

Rollbacking to `0.16` it seems that the ts file does not appear at all in the report, only the .vue file is listed

Sorry if I missed something in the documentation about that case !

### Reproduction

https://github.com/OrestHk/vitest-coverage-no-sfc

- yarn install
- yarn vitest --coverage

### System Info

```shell
System:
    OS: Windows 10 10.0.19044
    CPU: (4) x64 Intel(R) Core(TM) i5-4690K CPU @ 3.50GHz
    Memory: 6.30 GB / 15.85 GB
  Binaries:
    Node: 18.7.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.15 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 8.15.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 104.0.5112.81
    Edge: Spartan (44.19041.1266.0), Chromium (104.0.1293.47)
    Internet Explorer: 11.0.19041.1566
  npmPackages:
    @vitejs/plugin-vue: ^3.0.1 => 3.0.1
    vite: ^3.0.5 => 3.0.5
    vitest: ^0.21.1 => 0.21.1
```

### Used Package Manager

yarn

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
[...]

Issue #2155, opened 2022-10-15
Coverage report not clear and inconsistent with vue


Gnuk:
### Describe the bug

On a vue with TypeScript, vitest and istanbul plugin:

Some coverage information are not clear and inconsistent especially in the `.vue` files but it's not the case using jest.

_In the reproduction case, the coverage is `100%` for each measures using jest but not with vitest_

Actually, the problem is blocking to merge the PR https://github.com/jhipster/jhipster-lite/pull/3231 on JHipster Lite generator, that's why I put a link here.

_Sorry by advance if the reproduction case is not so minimal, it's because it's linked to the PR and this may evolve with the time. If really needed, please answer this issue but I think it's interesting to have the history from the JHipster Lite PR._

### Reproduction

Prerequisites: latest LTS node version

Reproduction:

* Clone my [fork](https://github.com/Gnuk/jhipster-lite) and go to the [move-to-vitest](https://github.com/Gnuk/jhipster-lite/tree/move-to-vitest) branch.
* Launch `npm ci` (or `install`)
* Launch `npm test`
* Look at the coverage (you can find the complete report under `target/test-results`)

You may get this in the summary:

```
=============================== Coverage summary ===============================
Statements   : 99.6% ( 996/1000 )
Branches     : 100% ( 175/175 )
Functions    : 99.17% ( 483/487 )
Lines        : 99.56% ( 920/924 )
================================================================================
```

The expected coverage can be reproduced using `jest`, it's just the same way to reproduce:

* Clone the [original](https://github.com/jhipster/jhipster-lite) repository.
* Launch `npm ci` (or `install`)
* Launch `npm test`
* Look at the coverage (you can find the complete report under `target/test-results`)

You may get this in the summary:

```
=============================== Coverage summary ===============================
Statements   : 100% ( 1033/1033 )
Branches     : 100% ( 110/110 )
Functions    : 100% ( 451/451 )
Lines        : 100% ( 948/948 )
================================================================================
```

_Note: if you want the exact behaviour between jest and vitest, you can also switch to the previous commit from my fork because the goal of the fork is to migrate from jest to vitest._

### System Info

```shell
System:
    OS: Linux 5.19 Arch Linux
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 5.92 GB / 15.27 GB
    Container: Yes
    Shell: 5.9 - /usr/bin/zsh
  Binaries:
[...]

PR #2181, opened 2022-10-20
fix(coverage): trim URL parameters from file paths in c8 coverage

OrestHk:
This should fix https://github.com/vitest-dev/vitest/issues/1839 but only for C8, the problem still exists for istanbul provider
This will trim URL parameters from files (`?vue&type=script&src=true&lang.ts`) allowing them to be properly reported in text report and to be browsable in html report

Not sure what's the policy on tests for the c8 package

### DIFF
--- a/packages/coverage-c8/src/provider.ts
+++ b/packages/coverage-c8/src/provider.ts
@@ -58,7 +58,7 @@ export class C8CoverageProvider implements CoverageProvider {
         if (!map)
           return
 
-        const url = _url.pathToFileURL(file.replace(/\?(.*)/, '')).href
+        const url = _url.pathToFileURL(file.split('?')[0]).href
 
         let code: string | undefined
         try {
--- a/packages/coverage-c8/src/provider.ts
+++ b/packages/coverage-c8/src/provider.ts
@@ -58,7 +58,7 @@ export class C8CoverageProvider implements CoverageProvider {
         if (!map)
           return
 
-        const url = _url.pathToFileURL(file).href
+        const url = _url.pathToFileURL(file).href.replace(/\?(.*)/, '')
 
         let code: string | undefined
         try {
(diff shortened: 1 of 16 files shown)


AriPerkkio:
The `istanbul` coverage provider has same issue. https://github.com/vitest-dev/vitest/issues/2155#issuecomment-1280359673

@OrestHk would you mind fixing that one as well here?

Also it would be great to see test cases for this issue.

sheremet-va:
What happens, when you have the same file, but different queries? Like vue does for transforming html/script/css?

OrestHk:
@AriPerkkio I also looked a bit at the istanbul part.
For the time being I only found a monkey patch for istanbul, but I don't think it's viable as the modification implied playing with `coverage` provided by `onAfterSuiteRun` to modify it as needed before the push
```
  onAfterSuiteRun({ coverage }: AfterSuiteRunMeta) {
    this.coverages.push(coverage)
  }
```
The type of coverage being unknown, I don't have enough knowledge on istanbul to tell that the solution was viable. That's why I focused on C8 ATM, I'll look a bit more into istanbul after this PR

@sheremet-va when a vue SFC is provided it will be listed as `component.vue` and the coverage will be provided for the component file passing through `<template>` and `<script>` if I remember correctly.

I'll try to give some time to this around this week and reimplement the tests I made for a previous PR about this issue for the SFC case and a combination of VUE + TS file.

packages/coverage-c8/src/provider.ts
```diff
@@ -58,7 +58,7 @@ export class C8CoverageProvider implements CoverageProvider {
         if (!map)
           return
 
-        const url = _url.pathToFileURL(file).href
+        const url = _url.pathToFileURL(file).href.replace(/\?(.*)/, '')
```
AriPerkkio, line 61: Isn't there already a `pathname` which does not contain the query string?

- https://nodejs.org/api/url.html#urlpathname

```suggestion
        const url = _url.pathToFileURL(file).pathname
```

OrestHk: Using `pathname` will return only the path of the URL, without the [origin](https://nodejs.org/api/url.html#urlorigin) part, which is needed for the `html` report (`{ reporter: 'html' }`) to redirect to the proper .html coverage file
I could do a concatenation of `.orign` + `.pathname` ? Looks less coming out of nowhere than the replace

Issue #2176, opened 2022-10-20
Cannot import/require from custom reporter if it's outside project root


segrey:
### Describe the bug

When reporter is located outside project root, it cannot be loaded:
```
vitest --root ./my-project-root --reporter $(pwd)/custom-reporter/my-reporter.js
Started...
Cannot find module './foo'
Require stack:
- /home/segrey/temp/vitest-reporter-with-imports/my-project-root/home/segrey/temp/vitest-reporter-with-imports/custom-reporter/my-reporter.js
Error: Cannot find module './foo'
Require stack:
- /home/segrey/temp/vitest-reporter-with-imports/my-project-root/home/segrey/temp/vitest-reporter-with-imports/custom-reporter/my-reporter.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:956:15)
    at Module._load (node:internal/modules/cjs/loader:804:27)
    at Module.require (node:internal/modules/cjs/loader:1022:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at /home/segrey/temp/vitest-reporter-with-imports/my-project-root/home/segrey/temp/vitest-reporter-with-imports/custom-reporter/my-reporter.js:3:5
    at ViteNodeRunner.directRequest (file:///home/segrey/temp/vitest-reporter-with-imports/node_modules/vitest/dist/chunk-vite-node-client.e22799cd.js:250:11)
    at async ViteNodeRunner.cachedRequest (file:///home/segrey/temp/vitest-reporter-with-imports/node_modules/vitest/dist/chunk-vite-node-client.e22799cd.js:127:12)
    at async ViteNodeRunner.executeId (file:///home/segrey/temp/vitest-reporter-with-imports/node_modules/vitest/dist/chunk-vite-node-client.e22799cd.js:110:12)
    at async loadCustomReporterModule (file:///home/segrey/temp/vitest-reporter-with-imports/node_modules/vitest/dist/chunk-vite-node-externalize.efa824b9.js:8524:28)
    at async file:///home/segrey/temp/vitest-reporter-with-imports/node_modules/vitest/dist/chunk-vite-node-externalize.efa824b9.js:8539:32
```

Looks like the problem is in `node_modules/vitest/dist/chunk-vite-node-utils.835c8b2c.js`
```
function toFilePath(id, root) {
[...]

PR #2184, opened 2022-10-21
fix(reporter): load custom reporter using `executeFile`

poyoho:
fix: #2176
load the reporter file base on file system

### DIFF
--- a/packages/vitest/src/node/reporters/utils.ts
+++ b/packages/vitest/src/node/reporters/utils.ts
@@ -1,4 +1,5 @@
 import type { ViteNodeRunner } from 'vite-node/client'
+import { normalize } from 'pathe'
 import type { Reporter } from '../../types'
 import { BenchmarkReportsMap, ReportersMap } from './index'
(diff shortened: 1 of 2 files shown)


packages/vitest/src/node/reporters/utils.ts
```diff
@@ -1,12 +1,13 @@
...
+import { normalize } from 'pathe'
 import type { Reporter } from '../../types'
 import { BenchmarkReportsMap, ReportersMap } from './index'
 import type { BenchmarkBuiltinReporters, BuiltinReporters } from './index'
 
 async function loadCustomReporterModule<C extends Reporter>(path: string, runner: ViteNodeRunner): Promise<new () => C> {
   let customReporterModule: { default: new () => C }
   try {
-    customReporterModule = await runner.executeId(path)
+    customReporterModule = await runner.executeId(normalize(`/@fs/${path}`))
```
antfu, line 10:
```suggestion
    customReporterModule = await runner.executeFile(path)
```

I think this would do.

antfu (approved):

Issue #2172, opened 2022-10-19
No test suite found in vite.config.js when it's not included in include glob


haikyuu:
### Describe the bug

No test suite found in vite.config.js when it's not included in include glob. This is caused by `includeSource` since the error is gone when that's commented.

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-jqgret?file=vite.config.ts

### System Info

```shell
System:
    OS: macOS 12.4
    CPU: (10) arm64 Apple M1 Max
    Memory: 79.06 MB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.17.0 - /usr/local/bin/node
    Yarn: 1.22.17 - /usr/local/bin/yarn
    npm: 8.15.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 106.0.5249.119
    Chrome Canary: 109.0.5366.0
    Safari: 15.5
    Safari Technology Preview: 16.4
  npmPackages:
    vite: ^3.1.8 => 3.1.8 
    vitest: ^0.24.3 => 0.24.3
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

sheremet-va:
Not sure what to do here? You are including a lot of source tests, and `vite.config` is part of glob, and had `"import.meta.vitest"`, so it is treated as a test file, and test files throw error if there are no tests inside.

haikyuu:
Ah, got you.

I think it's safe to exclude `vite.config.{j,t)s` from the glob in that case. Thoughts?

sheremet-va:
> Ah, got you.
> 
[...]

PR #2188, opened 2022-10-21
fix: improve config's default exclude

haikyuu:
This PR fixes https://github.com/vitest-dev/vitest/issues/2172.

- [x] Added to the docs


packages/vitest/src/defaults.ts
```diff
@@ -1,8 +1,7 @@
 import type { BenchmarkUserOptions, ResolvedCoverageOptions, UserConfig } from './types'
 
 export const defaultInclude = ['**/*.{test,spec}.{js,mjs,cjs,ts,mts,cts,jsx,tsx}']
-export const defaultExclude = ['**/node_modules/**', '**/dist/**', '**/cypress/**', '**/.{idea,git,cache,output,temp}/**']
-
+export const defaultExclude = ['**/node_modules/**', '**/dist/**', '**/cypress/**', '**/.{idea,git,cache,output,temp}/**', 'vite.config.*', 'vitest.config.*']
```
sheremet-va, line 4:
```suggestion
export const defaultExclude = ['**/node_modules/**', '**/dist/**', '**/cypress/**', '**/.{idea,git,cache,output,temp}/**', '{vite,vitest}.config.*']
```
I think we can actually add more configs, like in `defaultCoverageExcludes`

sheremet-va (changes requested):

packages/vitest/src/defaults.ts
haikyuu: Done, I also added it to the docs since it was missing

packages/vitest/src/defaults.ts
```diff
@@ -23,6 +22,7 @@ const defaultCoverageExcludes = [
   '**/__tests__/**',
   '**/{karma,rollup,webpack,vite,vitest,jest,ava,babel,nyc,cypress}.config.{js,cjs,mjs,ts}',
   '**/.{eslint,mocha,prettier}rc.{js,cjs,yml}',
+  '{vite,vitest}.config.*',
```
sheremet-va, line 25: No, what I meant is to also put files from `:23` line in `defaultExclude` 😄 

23 line already has vite and vitest configs :)

sheremet-va (changes requested):

haikyuu:
Seems like the macos CI machine needs to upgrade its specs since eve≠the linux one is almost 2x faster. And it's probably why the test times out.

antfu (approved):

PR #2209, opened 2022-10-25
feat: add runAllTimersAsync from sinonjs

### LINKED ISSUE #1804: Add async method variants from sinonjs/fake-timers to vitest

fake-timers provides an `await clock.runAllAsync()` call as an alternative to `clock.runAll()` (https://github.com/sinonjs/fake-timers#clockrunall--await-clockrunallasync) which makes it significantly easier to run all timers in a more complex asynchronous context (e.g. timers called by promises that are called by timers). Without it, it can be really difficult to actually have all timers run when those timers are initiated by other promises.

See Stack Overflow discussions such as https://stackoverflow.com/questions/51126786/jest-fake-timers-with-promises and https://stackoverflow.com/questions/52177631/jest-timer-and-promise-dont-work-well-settimeout-and-async-function to see scenarios where this is an issue.

For a similar reason, `await clock.tickAsync`, `await clock.nextAsync`, `await clock.runToLastAsync` would ideally also be exposed in some way.

I'm currently having to import sinon-js/fake-timers manually to provide this functionality, and I'd prefer to stay within vitest, especially since it's using fake-timers under the hood.

### Suggested solution

Adding something like:

```ts
runAllTimersAsync(): Promise<void> {
    if (this._checkFakeTimers())
      this._clock.runAllAsync()
}
```

to `packages/vitest/src/integrations/mock/timers.ts` for each associated method (and tests as needed).

### Alternative

_No response_

### Additional context

_No response_

### Validations

guillaumeduboc:
close #1804

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -2553,6 +2553,19 @@ Vitest provides utility functions to help you out through it's **vi** helper. Yo
   vi.advanceTimersByTime(150)
   ```
 
+### vi.advanceTimersByTimeAsync
+
+- **Type:** `(ms: number) => Promise<Vitest>`
+
+  Works just like `runAllTimersAsync`, but will end after passed milliseconds. This will include asynchronously set timers. For example this will log `1, 2, 3` and will not throw:
+
+  ```ts
+  let i = 0
+  setInterval(() => Promise.resolve().then(() => console.log(++i)), 50)
+
+  await vi.advanceTimersByTimeAsync(150)
+  ```
+
 ### vi.advanceTimersToNextTimer
 
 - **Type:** `() => Vitest`
@@ -2566,6 +2579,21 @@ Vitest provides utility functions to help you out through it's **vi** helper. Yo
   vi.advanceTimersToNextTimer() // log 1
     .advanceTimersToNextTimer() // log 2
     .advanceTimersToNextTimer() // log 3
+  ```
+
+### vi.advanceTimersToNextTimerAsync
+
+- **Type:** `() => Promise<Vitest>`
+
+  Will call next available timer even if it was set asynchronously. Useful to make assertions between each timer call. You can chain call it to manage timers by yourself.
+
+  ```ts
+  let i = 0
+  setInterval(() => Promise.resolve().then(() => console.log(++i)), 50)
+
+  vi.advanceTimersToNextTimerAsync() // log 1
+    .advanceTimersToNextTimerAsync() // log 2
+    .advanceTimersToNextTimerAsync() // log 3
   ```
 
 ### vi.getTimerCount
@@ -3014,6 +3042,28 @@ IntersectionObserver === undefined
   setInterval(() => console.log(++i), 50)
 
   vi.runOnlyPendingTimers()
+  ```
+
+### vi.runOnlyPendingTimersAsync
+
+- **Type:** `() => Promise<Vitest>`
+
+  This method will asynchronously call every timer that was initiated after `vi.useFakeTimers()` call, even asynchronous ones. It will not fire any timer that was initiated during its call. For example this will log `2, 3, 3, 1`:
+
+  ```ts
+  setTimeout(() => {
+    console.log(1)
+  }, 100)
+  setTimeout(() => {
+    Promise.resolve().then(() => {
+      console.log(2)
+      setInterval(() => {
+        console.log(3)
+      }, 40)
+    })
+  }, 10)
+
+  await vi.runOnlyPendingTimersAsync()
   ```
 
 ### vi.setSystemTime
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -3017,15 +3017,11 @@ IntersectionObserver === undefined
 - **Type:** `() => Promise<Vitest>`
 
   This method will asynchronously invoke every initiated timer until the timers queue is empty. It means that every timer called during `runAllTimersAsync` will be fired even asynchronous timers. If you have an infinite interval,
-  it will throw after 10 000 tries. For example this will log `2, 1`:
+  it will throw after 10 000 tries. For example this will log `result`:
 
   ```ts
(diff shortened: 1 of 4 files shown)


sheremet-va:
I see that `throws before allowing infinite recursion` fails on windows 🤔

guillaumeduboc:
> I see that `throws before allowing infinite recursion` fails on windows 🤔

hey @sheremet-va 👋 
I fixed the test

I was thinking I could add `tickAsync`, `nextAsync` and `runToLastAsync` as suggested in the issue
should I do it in this PR as well ?

sheremet-va:
> I was thinking I could add `tickAsync`, `nextAsync` and `runToLastAsync` as suggested in the issue
> should I do it in this PR as well ?

Sure, we can do that. Don't forget about docs though 😄

guillaumeduboc:
Sorry for the delay, my PR is almost done
I stumbled upon an issue with `@sinonjs/fake-timers` as there `runToLastAsync` was responsible for some timeouts in my tests

when https://github.com/sinonjs/fake-timers/pull/451 is merged and published I can bump the `@sinonjs/fake-timers` version and it should be good

guillaumeduboc:
hey @sheremet-va 👋 

Now that the sinonjs/fake-timers issue is fixed, I finished my PR and updated the docs
wdyt ?

docs/api/index.md
```diff
@@ -2964,6 +2992,25 @@ IntersectionObserver === undefined
...
+### vi.runAllTimersAsync
+
+- **Type:** `async() => Promise<Vitest>`
+
+  This method will asynchronously invoke every initiated timer until the timers queue is empty. It means that every timer called during `runAllTimersAsync` will be fired even asynchronous timers. If you have an infinite interval,
+  it will throw after 10 000 tries. For example this will log `2, 1`:
+
+  ```ts
+  setTimeout(async () => {
+    await setTimeout(() => {
```
sheremet-va, line 3004: this has no effect, is it? (unless imported form `timers/promises`, which is not specified in this code block)

docs/api/index.md
```diff
@@ -2533,6 +2533,19 @@ Vitest provides utility functions to help you out through it's **vi** helper. Yo
   vi.advanceTimersByTime(150)
   ```
 
+### vi.advanceTimersByTimeAsync
+
+- **Type:** `async (ms: number) => Promise<Vitest>`
```
sheremet-va, line 2538: Don't need to specify `async`

```suggestion
- **Type:** `(ms: number) => Promise<Vitest>`
```

docs/api/index.md
```diff
@@ -2548,6 +2561,21 @@ Vitest provides utility functions to help you out through it's **vi** helper. Yo
     .advanceTimersToNextTimer() // log 3
   ```
 
+### vi.advanceTimersToNextTimerAsync
+
+- **Type:** `async () => Promise<Vitest>`
```
sheremet-va, line 2566: Don't need to specify `async`

```suggestion
- **Type:** `() => Promise<Vitest>`
```

docs/api/index.md
```diff
@@ -2964,6 +2992,25 @@ IntersectionObserver === undefined
   vi.runAllTimers()
   ```
 
+### vi.runAllTimersAsync
+
+- **Type:** `async() => Promise<Vitest>`
```
sheremet-va, line 2997: Don't need to specify `async`

```suggestion
- **Type:** `() => Promise<Vitest>`
```

docs/api/index.md
```diff
@@ -2977,6 +3024,28 @@ IntersectionObserver === undefined
   vi.runOnlyPendingTimers()
   ```
 
+### vi.runOnlyPendingTimersAsync
+
+- **Type:** `async () => Promise<Vitest>`
```
sheremet-va, line 3029: Don't need to specify `async`

```suggestion
- **Type:** `() => Promise<Vitest>`
```

test/core/test/timers.test.ts
```diff
@@ -639,6 +920,113 @@ describe('FakeTimers', () => {
...
+      global.setTimeout(spies[0], 10)
+      global.setTimeout(spies[1], 50)
+
+      await timers.runOnlyPendingTimersAsync()
+
+      expect(spies[0]).toBeCalled()
+      expect(spies[1]).toBeCalled()
+    })
+
+    it.only('runs all timers in order', async () => {
```
sheremet-va, line 946: Please, remove `.only` here 😄

sheremet-va (changes requested):
Looks great! Thank you for your contribution. Have some small questions for docs though.

docs/api/index.md
guillaumeduboc: what do you mean it has no effect ? the `await` ?

sheremet-va: Yes, the await

guillaumeduboc: it actually has an impact
I can't find much documentation on this but having a `Promise` in a `setTimeout` has a different behavior from having a synchronous callback

the impact on timers is that this code would not work for `runAllTimers`

guillaumeduboc: okay this was not clear to me at all

I've been playing with these a bit and my understanding is that it doesn't change anything in node timers, but it changes in sinonjs/fake-timers

for their synchronous api, an asynchronous callback in a `setTimeout` will block the timer

```typescript
vi.useFakeTimers()

setTimeout(async () => {
  console.log(1)
  await Promise.resolve(true) // this promise will block the timer
  
  console.log(2) // happens after termination
}, 100)

vi.runAllTimers()
```

so you are right, for node it makes no difference, but for our fake timers it does

I will update the example to better reflect on that

sheremet-va (approved):

antfu (approved):

sheremet-va (approved):

PR #2226, opened 2022-10-28
fix: improve compatibility with native Node CJS resolution

### LINKED ISSUE #2219: Unable to import commonjs file with default export
### Describe the bug

The behavior between Node.js and vitest differs when importing a commonjs file which has a `module.exports.default` defined.

```js
//file.cjs
exports.default = 1;
```

```js
import x from 'file.cjs'
```

In Nodejs x is `{ default: "value" }`
in vitest x is `"value"` (the default export is automatically resolved)

As a workaround, I can add this snippet into my source code to make vitest work, but I really do not want to do this long term:

```js
const x = x.default ?? x
```

### Reproduction

https://github.com/danez-labs/vitest-default-commonjs

clone and run `npm i && npm test`

### System Info

```shell
System:
    OS: macOS 12.6.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 329.33 MB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.10.0 - ~/.asdf/installs/nodejs/18.10.0/bin/node
    npm: 8.19.2 - ~/.asdf/plugins/nodejs/shims/npm
  Browsers:
    Safari: 16.1
  npmPackages:
    vite: ^3.2.0 => 3.2.0 
    vitest: ^0.24.3 => 0.24.3
```

### Used Package Manager

npm

### Validations

sheremet-va:
Fixes #2219

### DIFF
--- a/packages/vite-node/src/client.ts
+++ b/packages/vite-node/src/client.ts
@@ -247,24 +247,21 @@ export class ViteNodeRunner {
         return Reflect.get(exports, p, receiver)
       },
       set(_, p, value) {
-        // Node also allows access of named exports via exports.default
-        // https://nodejs.org/api/esm.html#commonjs-namespaces
-        if (p !== 'default') {
-          if (!Reflect.has(exports, 'default'))
-            exports.default = {}
+        if (!Reflect.has(exports, 'default'))
+          exports.default = {}
 
-          // returns undefined, when accessing named exports, if default is not an object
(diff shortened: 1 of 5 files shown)


Uzlopak:
I did bisecting, and this PR resulted in regressions.
https://github.com/fastify/fastify-autoload/pull/277

Probably also #2320

Issue #2245, opened 2022-10-31
v0.24.4: Unable to load custom reporters by package name


AriPerkkio:
### Describe the bug

Likely related to #2184 (👋 @poyoho).

Custom reporters cannot be loaded by using the package name in `vitest@0.24.4`. There are no errors or any logging printed, the program just exits. Everything works fine with `vitest@0.24.3`.

As this has broken before #1194, I think it is time to add test cases for this. 

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-hubyyo

Stackblitz is working weirdly now so here is minimal vitest config also:

```ts
export default defineConfig({
  test: {
    outputFile: 'report.xml',

    // Works on 0.24.3, breaks on 0.24.4
    reporters: ['vitest-sonar-reporter'],

    // Works on 0.24.4
    // reporters: ['node_modules/vitest-sonar-reporter/dist/index'],
  },
});
```

### System Info

```shell
Stackblitz, Github Codespaces
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

poyoho:
I add the test case in #2249, but it seems to ok?

PR #2249, opened 2022-10-31
fix: load deps reporter

poyoho:
fix: #2245

### DIFF
--- a/packages/vitest/src/node/reporters/utils.ts
+++ b/packages/vitest/src/node/reporters/utils.ts
@@ -1,5 +1,4 @@
 import type { ViteNodeRunner } from 'vite-node/client'
-import { isAbsolute } from 'pathe'
 import type { Reporter } from '../../types'
 import { BenchmarkReportsMap, ReportersMap } from './index'
 import type { BenchmarkBuiltinReporters, BuiltinReporters } from './index'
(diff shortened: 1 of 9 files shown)


sheremet-va:
Doesn't seem like it works? 

It can't work, because `/@fs/` expects absolute path to be after it, and package name is not an absolute path

poyoho:
> Doesn't seem like it works?
> 
> It can't work, because `/@fs/` expects absolute path to be after it, and package name is not an absolute path

hhhh only work in macos 🤦

AriPerkkio (approved):
Fix verified with bug report's minimal reproduction setup by applying changes under `node_modules/vitest/dist`. Also the added test case seems to trigger the expected condition. Thanks @poyoho! 💯 

However the way how vitest process silently exits in `v0.24.4` is not ideal. I would rather see an error message indicating loading reporter failed. I wonder why this is not visible for users.

https://github.com/vitest-dev/vitest/blob/7f55c54e51be24107d6591761b586dc33890cade/packages/vitest/src/node/reporters/utils.ts#L11-L13

sheremet-va:
I think we can just revert the change altogether (revert #2184 I mean). The problem was with `toFilePath`, and we fixed it by checking existence of folder.

poyoho:
I add the no base on root reporter test case, it seems to work well too. 😊

sheremet-va (approved):

Issue #2256, opened 2022-11-02
Vitest doesn't display build errors originating from plugin correctly (Vite does)


haikyuu:
### Describe the bug

Vitest doesn't pick the `frame` property if present in the error originating from a plugin. 
Vite picks it up as a `build` error and displays it using [`outputBuildError`](https://github.com/vitejs/vite/blob/c53ffec3465d2d28d08d29ca61313469e03f5dd6/packages/vite/src/node/build.ts#L523-L534).
### Reproduction

https://stackblitz.com/edit/vitejs-vite-7459gf?file=vite.config.js

### System Info

```shell
*
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #2257, opened 2022-11-02
feat: display error frame if present (like Vite)

haikyuu:
This PR fixes https://github.com/vitest-dev/vitest/issues/2256

It displays the frame if present in the error instead of generating one. This matches Vite's behaviour


sheremet-va:
Can you add a test case to `tests/stacktraces`? 😄

haikyuu:
Need to update the tests. This is a false positive

haikyuu:
@sheremet-va you can take another look now

sheremet-va (approved):

dammy001 (approved):

antfu (approved):

PR #2274, opened 2022-11-04
wip - added lit property update test

christophe-g:


sheremet-va:
Test is passing 🤔

sheremet-va:
Lit recommends testing in the browser - for that purpose we have an example of using Lit in browser mode.

PR #2275, opened 2022-11-04
fix(coverage): `istanbul` provider to preserve implicit else

### LINKED ISSUE #1887: Istanbul code coverage differs from nyc result
### Describe the bug

I tried running Vitest + Istanbul and the result seems to differ from using Istanbul (+ Mocha, but I also tried Cypress) itself.

- Vitest seems to output a Linux-style path for coverage-final.json, even when running on Windows, this breaks NYC merge
- The branch coverage seems to differ
- (_coverageSchema, hash, contentHash are missing, but they are probably not important)

image
image
First image: Mocha + NYC, second image: Vitest

I'm not 100% sure if this a Istanbul or Vitest issue

### Reproduction

https://github.com/marvin-j97/vitest-windows-path

(I also have a more complex repo trying to merge Vitest + Cypress Typescript coverage reports: https://github.com/marvin-j97/good-next-template/tree/dev)

### System Info

```shell
System:
    OS: Windows 10 10.0.19044
    CPU: (8) x64 Intel(R) Core(TM) i7-7700K CPU @ 4.20GHz
    Memory: 8.75 GB / 15.96 GB
  Binaries:
    Node: 16.14.2 - C:\Program Files\nodejs\node.EXE
    npm: 8.5.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (104.0.1293.54)
  npmPackages:
    vitest: 0.22.1 => 0.22.1
```

### Used Package Manager

yarn

### Validations

### LINKED ISSUE #2239: Implicit branches are not considered in coverage
### Describe the bug

Code with implicit branches is not detected in coverage. Consider the following code:

```javascript
function fn(param) {
  if (param) {
	// If param is falsy, this if never runs
  }
}
```

If test code is only executing `fn` with a truthy parameter the implicit else branch is never tested, thus should not count towards covered branches. Interestingly enough, coverage still reports 100% for branch coverage.

This impacts both coverage providers but in case of c8 it is a known issue. With istanbul it is expected to report the correct 50% branch coverage metric, yet running under vitest the problem persists. Read reproduction for details.

### Reproduction

Consider the following code ([origin](https://gist.github.com/mikeal/60aea513165c35ac438f6a35f0727a75)):
```javascript
// cov.js
export const fn = y => {
  if (typeof y !== 'object') {
    y = 'yup'
  }
  console.log(y)
}
```
```javascript
// cov.spec.js
import { fn } from "cov"

test("implicit branch coverage bug", () => {
  const x = 'asdf'
  fn(x)
});
```

The coverage report using `@vitest/coverage-c8` reports this as 100% branch coverage, due to a known issue with v8 (bcoe/c8#227 and [this](https://bugs.chromium.org/p/v8/issues/detail?id=10627) v8 ticket). However, running a sligthly different version of this code [from the gist I linked above](https://gist.github.com/mikeal/60aea513165c35ac438f6a35f0727a75) in commonjs syntax through `nyc node cov.cjs` prints 50%, correctly. (ESM code causes `nyc` to have a blank output but that is also already a known issue).

Interestingly, switching the coverage provider to `@vitest/coverage-istanbul` is still impacted, showing full coverage mistakenly.

```
% Coverage report from istanbul
All files |     100 |      100 |     100 |     100 |                  
 cov.js   |     100 |      100 |     100 |     100 |                  
```

### System Info

```shell
system: Windows 11 Build 10.0.22621
node: v19.0.0
pnpm: v6.32.9
nyc: v15.1.0
vitest: v0.24.3
```

### Used Package Manager

pnpm

### Validations

AriPerkkio:
- Fixes #1887.
- Fixes #2239.
- Upstream issue https://github.com/istanbuljs/istanbuljs/issues/705.
- Upstream fix https://github.com/istanbuljs/istanbuljs/issues/706.

Adds a work-around for implicit else's branch coverages. There is a PR for the upstream issue but as `istanbul` packages do not update that often it might be worth to apply this work-around for now. This work-around is compatible with the upstream patch but should be removed once it is merged and released. 

The upstream issue has minimal reproduction setup which mimics vitest (esbuild, TS, istanbul packages).

### DIFF
--- a/packages/coverage-istanbul/src/provider.ts
+++ b/packages/coverage-istanbul/src/provider.ts
@@ -108,6 +108,8 @@ export class IstanbulCoverageProvider implements CoverageProvider {
     if (this.options.all)
       await this.includeUntestedFiles(mergedCoverage)
 
+    includeImplicitElseBranches(mergedCoverage)
+
     const sourceMapStore = libSourceMaps.createSourceMapStore()
     const coverageMap: CoverageMap = await sourceMapStore.transformCoverage(mergedCoverage)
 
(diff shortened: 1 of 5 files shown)


AriPerkkio:
Windows CI is failing, I'll check that later.

`D:\a\vitest\vitest\test\coverage-test/src/implicitElse.ts` vs `D:/a/vitest/vitest/test/coverage-test/src/implicitElse.ts` 😞

sheremet-va (approved):

PR #2280, opened 2022-11-05
fix: test.each() in vitest differs from jest

### LINKED ISSUE #2278: test.each() in vitest differs from jest
### Describe the bug

There's an difference between the specifications of `jest-each` and `vitest`.

With `jest-each`:
- *when first item is an array* then all arrays are parameter lists.
- *when first item is not an array*, then all arrays are arrays.
 
With `vitest`:
- *if any item is an array* then its content is a parameter list.

## analysis
This test in the `vitest` specs fails on `jest`:
```js
test.each([
  null,
  [null], // <-- vitest sees `null`, jest sees `[null]`
])('null is null', (value) => {
  expect(value).toBe(null)
})
```

### Reproduction

This repo illustrate the difference in behavior:
https://github.com/jf-paradis/vitest-each-demo

```bash
git clone https://github.com/jf-paradis/vitest-each-demo
cd vitest-each-demo
npm install
npm unit:jest // to run tests under jest
npm unit:vitest // to run tests under vitest
```

### System Info

```shell
Binaries:
    Node: 16.16.0 - ~/.nvm/versions/node/v16.16.0/bin/node
  npmPackages:
    vitest: ^0.24.5 => 0.24.5
```

### Used Package Manager

npm

### Validations

jf-paradis:
- Fixes #2278 

Previously:
- every test case was tested individually to see if it's an array,
- if so, the test case was spread as test parameters  (in the test callback function).

This PR makes `test.each()` follow the same algorithm as `jest-each`:
- the array of test cases is first evaluated to see it's a table (an array of arrays),
- if so, then all test cases are spread as test parameters, otherwise not.

This means for example that a single array in a series of test cases is not spread as test parameters.

Note that Jest pre-normalize the table, which we can still can avoid here, and process each row on the fly.
https://github.com/facebook/jest/blob/main/packages/jest-each/src/table/array.ts#L49

### DIFF
--- a/packages/vitest/src/runtime/suite.ts
+++ b/packages/vitest/src/runtime/suite.ts
@@ -232,9 +232,10 @@ function createTest(fn: (
   testFn.each = function<T>(this: { withContext: () => TestAPI }, cases: ReadonlyArray<T>) {
     const test = this.withContext()
 
+    const isTable = cases.every(Array.isArray)
     return (name: string, fn: (...args: T[]) => void, options?: number | TestOptions) => {
       cases.forEach((i, idx) => {
-        const items = Array.isArray(i) ? i : [i]
+        const items = isTable ? i as T[] : [i]
         test(formatTitle(name, items, idx), () => fn(...items), options)
       })
     }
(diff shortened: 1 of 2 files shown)


sheremet-va:
We already have an open PR #2064

antfu:
Thank you for working on this! Close in favor of https://github.com/vitest-dev/vitest/pull/2064

PR #2281, opened 2022-11-06
fix: remove undefined values in options

### LINKED ISSUE #2153: "Patterns must be a string" error when passing `undefined` to `include` option
### Describe the bug

The types for the `include` and `exclude` options allow for passing `undefined` as the value.
This is very nice since it allows for applying the option conditionally very easily.
```ts
{
  include: testSSR ? ["server.test.{ts,tsx}"] : undefined
}
```
But this error appears in the console when trying to do this
```
TypeError: Patterns must be a string (non empty) or an array of strings
```
As a walkaround, I can do this, but it doesn't seem intentional...
```ts
{
  ...(testSSR && { include: ["server.test.{ts,tsx}"] })
}
```

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-atexbn?file=vite.config.ts

### System Info

```shell
System:
    OS: Windows 10 10.0.19041
    CPU: (8) x64 Intel(R) Core(TM) i5-8265U CPU @ 1.60GHz
    Memory: 920.68 MB / 7.79 GB
  Binaries:
    Yarn: 1.22.19 - C:\Program Files\nodejs\yarn.CMD
    npm: 8.19.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.423.0), Chromium (106.0.1370.42)
    Internet Explorer: 11.0.19041.1
```

### Used Package Manager

pnpm

### Validations

g4rry420:

Resolves #2153
I have added a `removeUndefinedValues` function which removes any undefined values passed by the user from userConfig so that default values can be applied.

### DIFF
--- a/packages/vitest/src/node/plugins/index.ts
+++ b/packages/vitest/src/node/plugins/index.ts
@@ -2,7 +2,7 @@ import type { UserConfig as ViteConfig, Plugin as VitePlugin } from 'vite'
 import { relative } from 'pathe'
 import { configDefaults } from '../../defaults'
 import type { ResolvedConfig, UserConfig } from '../../types'
-import { deepMerge, ensurePackageInstalled, notNullish } from '../../utils'
+import { deepMerge, ensurePackageInstalled, notNullish, removeUndefinedValues } from '../../utils'
 import { resolveApiConfig } from '../config'
 import { Vitest } from '../core'
 import { generateScopedClassName } from '../../integrations/css/css-modules'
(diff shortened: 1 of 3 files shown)


packages/vitest/src/node/plugins/index.ts
```diff
@@ -12,6 +12,14 @@ import { MocksPlugin } from './mock'
 import { CSSEnablerPlugin } from './cssEnabler'
 import { CoverageTransform } from './coverageTransform'
 
+function removeUndefinedValues(obj: Record<string, any>) {
```
dammy001, line 15: Can you move this to `utils/base.ts` file

antfu (approved):

dammy001 (approved):

PR #2292, opened 2022-11-08
fix: default reporter regression

antfu:


packages/vitest/src/typecheck/typechecker.ts
```diff
@@ -236,9 +236,6 @@ export class Typechecker {
   }
 
   public getTestFiles() {
-    return Object.values(this._tests || {}).map(({ file }) => ({
-      ...file,
-      result: undefined,
-    }))
+    return Object.values(this._tests || {}).map(i => i.file)
```
sheremet-va, line 239: This was on purpose, but I don't remember why. Does it change the output?

sheremet-va (changes requested):

packages/vitest/src/typecheck/typechecker.ts
sheremet-va: I think it's fine

PR #2294, opened 2022-11-08
feat: allow hooks to be executed in a stack or list

### LINKED ISSUE #2279: beforeAll and beforeEach run in parallel making it hard to compose hooks
### Describe the bug

All beforeAll and beforeEach hooks are run in parallel which means a hook can't depend on results of an earlier hook.

For example I defined a `allWithTmp()` hook to create a temp folder, and I wanted to create a `allWithCert()` hook that would create a self signed cert in the tmp directory. I expected hooks to run sequentially in order.

```typescript
allWithTmp()
allWithCert()

function allWithTmp() {
  beforeAll(async (ctx) => {
    const tmp = ctx.tmp = await mkdtmp('suite-')
    return () => rm(tmp, { recursive: true }) 
  })
})

function allWithCert() {
  beforeAll(async (ctx) => {
    // join throws an error because ctx.tmp is undefined.
    await makeCert(join(ctx.tmp, 'cert.pem'))
  })
})

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-jyerpn?file=test/basic.test.ts

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    vite: latest => 3.2.2 
    vitest: latest => 0.24.5
```

### Used Package Manager

npm

### Validations

sheremet-va:
Closes #2279

This PR adds an option `sequence.hooks` with possible values:

- `stack` will order "after" hooks in reverse order, "before" hooks will run in the order they were defined
- `list` will order all hooks in the order they are defined
- `parallel` will run hooks in a single group in parallel (hooks in parent suites will still run before the current suite's hooks)

I want to release this in 0.25.1 or 0.25.2, so we don't change the default behaviour, since it would be a breaking change.

@kristiandupont can you review this? It's possible that I misunderstood the idea

### DIFF
--- a/test/core/test/hooks-stack.test.ts
+++ b/test/core/test/hooks-stack.test.ts
@@ -1,0 +1,47 @@
+import * as vitest from 'vitest'
+import { describe, expect, test, vi } from 'vitest'
+
+vi.setConfig({
+  sequence: {
+    hooks: 'stack',
+  },
+})
+
+const hookOrder: number[] = []
+function callHook(hook: 'beforeAll' | 'beforeEach' | 'afterAll' | 'afterEach', order: number) {
+  vitest[hook](() => {
+    hookOrder.push(order)
+  })
+}
+
+describe('hooks are called sequentially', () => {
+  callHook('beforeAll', 1)
+  callHook('beforeAll', 2)
+  callHook('beforeAll', 3)
+
+  callHook('afterAll', 4)
+  // will wait for it
+  vitest.afterAll(async () => {
+    await Promise.resolve()
+    hookOrder.push(5)
+  })
+  callHook('afterAll', 6)
+
+  callHook('beforeEach', 7)
+  callHook('beforeEach', 8)
+  callHook('beforeEach', 9)
+
+  callHook('afterEach', 10)
+  callHook('afterEach', 11)
+  callHook('afterEach', 12)
+
+  test('before hooks pushed in order', () => {
+    expect(hookOrder).toEqual([1, 2, 3, 7, 8, 9])
+  })
+})
+
+describe('previous suite run all hooks', () => {
+  test('after all hooks run in reverse order', () => {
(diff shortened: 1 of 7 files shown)


antfu (approved):

test/core/test/hooks-stack.test.ts
```diff
@@ -0,0 +1,47 @@
...
+})
+
+const hookOrder: number[] = []
+function callHook(hook: 'beforeAll' | 'beforeEach' | 'afterAll' | 'afterEach', order: number) {
+  vitest[hook](() => {
+    hookOrder.push(order)
+  })
+}
+
+describe('hooks are called sequentially', () => {
```
kristiandupont, line 17: This describes correct behavior and it works as far as I can tell. A possible change could be to write it like this:
```ts
describe('hooks are called sequentially', () => {
  callHook('beforeAll', 1)
  callHook('beforeAll', 2)
  callHook('beforeAll', 3)

  callHook('afterAll', 4)
  // will wait for it
  vitest.afterAll(async () => {
    await Promise.resolve()
    hookOrder.push(5)
  })
  callHook('afterAll', 6)

  callHook('beforeEach', 7)
  callHook('afterEach', 8)

  callHook('beforeEach', 9)
  callHook('afterEach', 10)

  callHook('beforeEach', 11)
  callHook('afterEach', 12)

  test('before hooks pushed in order', () => {
    expect(hookOrder).toEqual([1, 2, 3, 7, 9, 11])
  })
})

describe('previous suite run all hooks', () => {
  test('after all hooks run in reverse order', () => {
    expect(hookOrder).toEqual([1, 2, 3, 7, 9, 11, 12, 10, 8, 6, 5, 4])
  })
})
```

This emphasizes the feature that I am looking for: the ability to group `before*` and `after*` hooks so they can be extracted into functions that can be reused. But definitely not a show-stopper.

kristiandupont (approved):

PR #2346, opened 2022-11-17
chore: remove cross-env

btea:


kleinfreund:
Just removing cross-env will break scripts using environment variables on all Windows environments.

sheremet-va:
> Just removing cross-env will break scripts using environment variables on all Windows environments.

shell-emulator does the same in pnpm config, so this PR will not break anything

sheremet-va (approved):

PR #2359, opened 2022-11-21
fix: Detect tests in folders starting with `.` (fix #2344)

### LINKED ISSUE #2344: Tests inside folders starting with `.` is ignored
### Describe the bug

I was adding some tests to files inside `.vitepress` folder for MermaidJS, but those tests are ignored. 

Two possible solutions:
1. Add `dot: true` to the globOptions [here](https://github.dev/vitest-dev/vitest/blob/3038991347500f1336089698a3517ee5c5a5b4f3/packages/vitest/src/node/core.ts#L556-L560). (Easy)
2. Expose `globOptions` as a parameter.

I can raise the PR for whichever is the best fix.

### Reproduction

https://github.com/mermaid-js/mermaid/pull/3797/files#diff-1b9f393bba6c20785279cedd5f638becc35225e879f488fbf4d79a4b56a084dc

### System Info

```shell
System:
    OS: macOS 12.3
    CPU: (10) arm64 Apple M1 Pro
    Memory: 73.88 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 19.1.0 - ~/.volta/tools/image/node/19.1.0/bin/node
    Yarn: 1.22.19 - ~/.volta/tools/image/yarn/1.22.19/bin/yarn
    npm: 8.19.3 - ~/.volta/tools/image/node/19.1.0/bin/npm
  Browsers:
    Brave Browser: 106.1.44.112
    Chrome: 107.0.5304.110
    Firefox: 106.0.5
    Safari: 15.4
```

### Used Package Manager

pnpm

### Validations

sidharthv96:
I have added a test in a folder starting with `.` and manually verified that it's picked up. 
image

But it's not the right test as it won't fail even if `dot` is `false`. 

Is there an example test I can refer to ensure the path is picked up? Or somehow verify that that exact test was run?

fixes #2344

### DIFF
--- a/packages/vitest/src/node/core.ts
+++ b/packages/vitest/src/node/core.ts
@@ -553,8 +553,9 @@ export class Vitest {
   async globTestFiles(filters: string[] = []) {
     const { include, exclude, includeSource } = this.config
 
-    const globOptions = {
+    const globOptions: fg.Options = {
       absolute: true,
+      dot: true,
       cwd: this.config.dir || this.config.root,
       ignore: exclude,
     }
--- a/test/core/test/.dot-folder/dot-test.test.ts
+++ b/test/core/test/.dot-folder/dot-test.test.ts
@@ -1,0 +1,5 @@
+import { expect, test } from 'vitest'
+
+test('check that tests inside folder with . is run', () => {
+  expect(true).toBe(true)
+})
--- a/test/fails/fixtures/.dot-folder/dot-test.test.ts
+++ b/test/fails/fixtures/.dot-folder/dot-test.test.ts
@@ -1,5 +1,5 @@
 import { expect, test } from 'vitest'
 
 test('check that tests inside folder with . is run', () => {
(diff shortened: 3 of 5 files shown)


sheremet-va:
> But it's not the right test as it won't fail even if dot is false.

You can add a failing test inside `.dot-folder` in `test/failing/fixtures`.

sidharthv96:
Had to tweak the runner a bit, but now it works. Thanks!

sheremet-va (approved):

chrisbbreuer:
@sidharthv96 I just came across this issue as well, so thanks for fixing it!

Does the VS Code extension work properly for you within these dot folders? As in, do these actions appear for you in your editor? I am struggling to get them to show up and am wondering whether I may be running into a bug or misconfiguration haha

image

sidharthv96:
image

Seems like it's skipped in the extension. The icons don't come for me.

PR #493, opened 2022-01-09
feat(ui): add summary

userquin:
This PR includes:
- fix scrolls on navigation and suite panel
- adds a new panel to add report state (rn only the global state)
- adds current module again to control the ui summary
- adds show ui summary logic on vitest logo

PR #2364, opened 2022-11-22
feat(ui): add coverage

### LINKED ISSUE #2336: Coverage viewer in Vitest UI

As a developer building tests using Vitest I want to keep an eye on the test coverage to make sure I'm hitting all the code paths I want to test so that I can write extra tests should coverage not be sufficient.

Currently in Vitest UI there is no information about code coverage visible except the coverage information in the CLI, which is not very detailed.

image

### Suggested solution

My proposal is a new tab or section in the Vitest UI that shows a project's source code and the coverage like Istanbul would render in their HTML output.

image

With maybe a bit of styling to fit the Vitest UI theme ;)

### Alternative

- The UI could also host the Istanbul coverage HTML files from the `coverage` folder whenever Vitest UI is running. Perhaps a button in the Vitest UI could link to the coverage pages.

- My current 'alternative' is to start a web server through my IDE that hosts the `coverage` folder and I have that open in another browser tab. Opening `coverage/index.html` directly in the browser also works, but then it does not auto-update like it does when hosting through VSCode.

### Additional context

_No response_

### Validations

userquin:
This PR adds the coverage folder in a new iframe inside `Coverage` component:
- coverage css files not being served
- vite tranforms coverage html pages, I need to find a way to exclude them
- fix coverage error (check the screenshot below): I guess the error comes when Vite injects `@vite/client` script

To test the coverage run `pnpm run build && pnpm run test` from root folder and from another shell run `cd packages/ui && nr dev:ui`.

/cc @edimitchel : from your comment here https://github.com/vitest-dev/vitest/pull/493#issuecomment-1012659202

closes #2336 

imagen


userquin:
It is working but only with bundled ui, using it on dev I cannot make it working (css not being loaded):
- `pnpm run install && pnpm run build` from root
- `cd test/core && pnpm vitest --ui`

We also need to disable coverage button while running and adding small delay once tests run finish.

The css styles using `istambul` in dark mode needs a review: https://streamable.com/8jsoac

userquin:
Lastest changes: https://streamable.com/3ek8l0

userquin:
I need to add another script to enable coverage, rn enabled in tests/core.

edimitchel:
I wondering, won't you able to reuse raw data and attach coverage information to each test suite and show the coverage on suite/test items list, and details put in a test details tab?

userquin:
> I wondering, won't you able to reuse raw data and attach coverage information to each test suite and show the coverage on suite/test items list, and details put in a test details tab?

Yeah, I'll work next weekend on this - current state was simple, just adding a middleware and some client stuff, your proposal requires a bit more work

Issue #2316, opened 2022-11-11
Coverage report is not only generated for changed files in watch mode


netcoding87:
### Describe the bug

As a developer using Vitest I want to coverage report in watch mode only for files affected by code changes instead of all so that I can easily see how my changes affected the code.

I know this behavior from using `jest` on projects setup with `cra`. In my current project with more than 100 files it is very frustrating to have to search out the files from the report after every run...

### Reproduction

Repository to investigate: https://github.com/netcoding87/vitest-coverage (see `README.md`)

### System Info

```shell
System:
    OS: macOS 12.6.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 246.67 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.13.2 - ~/.nvm/versions/node/v16.13.2/bin/node
    npm: 8.1.2 - ~/.nvm/versions/node/v16.13.2/bin/npm
  Browsers:
    Chrome: 107.0.5304.110
    Firefox: 106.0.3
    Safari: 16.1
  npmPackages:
    @vitejs/plugin-react: ^2.2.0 => 2.2.0 
    vite: ^3.2.3 => 3.2.3 
    vitest: ^0.25.1 => 0.25.1
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

DevathiMahesh:
Yes. Even i am facing the same issue. Its running for all the files in the codebase.......

AriPerkkio:
[...]

PR #2385, opened 2022-11-28
feat: Istanbul Coverage only for changed Files

g4rry420:

I have added `watchAll` coverage config option, which by defaults provides coverage for all files but if it is `false`, then it only show coverage for changed file.
I have also added an example `react-istanbul-coverage` folder to test out these changes. Please, let me  know if you see any issues :)

Resolves: #2316


sheremet-va:
What's the difference between "all" and why do we need a separate option?

@AriPerkkio can you take a look, when you have time?

g4rry420:
The only reason, I have added `coverage.watchAll` option so that if a user wants to see all files coverages even when a single file has changed

packages/coverage-istanbul/src/provider.ts
```diff
@@ -105,6 +105,11 @@ export class IstanbulCoverageProvider implements CoverageProvider {
       return map
     }, {})
 
+    if (!this.ctx.isFirstRun && !this.ctx.config.coverage.watchAll && this.ctx.changedTests.size > 0) {
+      const changedTests = Array.from(this.ctx.changedTests)
+      mergedCoverage.filter(key => key === changedTests[0].replace('.test', ''))
```
AriPerkkio, line 110: This assumes that the tested file only imports files with similar names, right? So after you make change to `math.test.ts`, you will only see `math.ts`in report. You won't see `random-number-generator.ts` that it might import as well.

Also making changes to `math.spec.ts` breaks this logic as the hard-coded `.test` pattern is used.

packages/coverage-istanbul/src/provider.ts
```diff
@@ -105,6 +105,11 @@ export class IstanbulCoverageProvider implements CoverageProvider {
       return map
     }, {})
 
+    if (!this.ctx.isFirstRun && !this.ctx.config.coverage.watchAll && this.ctx.changedTests.size > 0) {
+      const changedTests = Array.from(this.ctx.changedTests)
```
AriPerkkio, line 109: The `changedTests` cannot be used for resolving changed files, as changed tests can import all kinds of files. Instead, if we really want to include only the changed files, we'll need to see if Vite exposes names of the changed files in HMR-api, or something similar.

packages/coverage-istanbul/src/provider.ts
g4rry420: Yes, this is correct.

packages/coverage-istanbul/src/provider.ts
g4rry420: Thank you for your feedback. I look into more options for this.

AriPerkkio:
@g4rry420 let's see if simply enabling `coverage.cleanOnRerun` is enough: https://github.com/vitest-dev/vitest/issues/2316#issuecomment-1329358821

That option could be enabled by default when CoverageReporter (c8, istanbul) is created, and watch mode is enabled. Also I think `coverage.all` could be disabled when watch mode is enabled.

AriPerkkio:
Alright, based on the 👍's on https://github.com/vitest-dev/vitest/issues/2316#issuecomment-1329358821 from both of the issue reporters, I guess we can proceed with the fixes. 

All that is needed is to set a few defaults:
 - default value for `coverage.cleanOnRerun` should be `true`, when watch mode is enabled
 - default value for `coverage.all` should be `false`, when watch mode is enabled
 
If watch mode is disabled, default values should be as before. I think these changes can be considered as a regular bug fix.

packages/vitest/src/node/config.ts
```diff
@@ -92,6 +92,11 @@ export function resolveConfig(
     mode,
   } as ResolvedConfig
 
+  if (resolved.watch) {
+    resolved.coverage.cleanOnRerun = true
+    resolved.coverage.all = false
```
AriPerkkio, line 97: Does this override user's configurations, or simply set default values?

I'm not sure if we should override the values, as users are then unable to configure these at all in watch mode.

g4rry420:
@AriPerkkio By default, the `watch=true` mode, so I have also set the `clearOnReRun=true` and `all` is already false.
Do we need to set `clearOnReRun=false` when user set `watch=false` ?

packages/vitest/src/node/config.ts
g4rry420: This does overrides the user config, I have modified the changed, after new commits. Thanks for pointing these out.

AriPerkkio (approved):
> Do we need to set clearOnReRun=false when user set watch=false ?

Looks like the `cleanOnReRun` is only used by the watch mode, so no need to disable it in run mode. Also no need to explicitly look for `watch=true` - it can be always enabled by default, exactly as you've done here. 👍

sheremet-va (changes requested):
Why are there changes to lockfile?

g4rry420:
Not sure about, but when I merge the latest changes from `main` branch and run `pnpm install`, these changes come up and I committed them.

sheremet-va:
> Not sure about, but when I merge the latest changes from `main` branch and run `pnpm install`, these changes come up and I committed them.

Please, don't commit lockfile.

sheremet-va (approved):

PR #2391, opened 2022-11-29
fix(mocker): set cache before mocking to allow circular dependencies

### LINKED ISSUE #2381: vi.mock() not working when there is an import loop
### Describe the bug

When using `vi.mock` on a module that is part of an import loop and is imported in the test there is an error displayed.

```
TypeError: default is not a function
 ❯ Module.__vite_ssr_exports__.default ../../src/index.js:8:15
```

I was debugging this for a while and only realized that this is an import loop when I saw the contents of the `moduleCache`:

```js
{
    importers: Set(1) {
      '/Users/danieltschinder/Projects/github/react-docgen/packages/react-docgen/src/utils/__tests__/getMemberValuePath-test.ts'
    },
    promise: Promise { <pending> },
    code: 'removed',
    exports: [Object: null prototype] [Module] {}
}
```

this is from `console.log(mod)` after [this line](https://github.com/vitest-dev/vitest/blob/main/packages/vitest/src/runtime/mocker.ts#L364).
So the promise is still pending, but vitest already mocks the exports which are not yet loaded.
Not sure if this helps, but that is what I found.

### Reproduction

1. `git clone https://github.com/danez-labs/vitest-default-commonjs.git`
2. npm i
3. npm test

### System Info

```shell
Tested on gitpod
  System:
    OS: Linux 5.15 Ubuntu 20.04.5 LTS (Focal Fossa)
    CPU: (16) x64 AMD EPYC 7B13
    Memory: 37.65 GB / 62.80 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Yarn: 1.22.19 - ~/.nvm/versions/node/v16.18.1/bin/yarn
  npmPackages:
    vite: ^3.2.4 => 3.2.4 
    vitest: ^0.25.3 => 0.25.3
```

### Used Package Manager

npm

### Validations

danez:
Fixes #2381

This is an attempt to fix the issue described in #2381.

As suggested by @sheremet-va I removed the `moduleCache` Handling and only used `this.request`
```diff
-const mod = this.moduleCache.get(cacheKey)?.exports || await this.request(dep)
+const mod = await this.request(dep)
```

This alone did not fix the problem, as [this block](https://github.com/vitest-dev/vitest/blob/main/packages/vite-node/src/client.ts#L204-L209) kicks in for the second request to the module and at that point we already have the Module-Symbol set on exports.

In order to fix this, it is necessary to already assign an empty exports object in the module cache for the mocked module. This way the second call to `requestWithMock()` would return this empty object, which later gets filled once the first call resolves.

This also mimics [what Node.js does](https://nodejs.org/api/modules.html#modules_cycles) for circular dependencies and as long as the imported value isn't directly used in the module scope everything is fine.

### DIFF
--- a/packages/vitest/src/runtime/mocker.ts
+++ b/packages/vitest/src/runtime/mocker.ts
@@ -1,7 +1,7 @@
 import { existsSync, readdirSync } from 'fs'
 import { isNodeBuiltin } from 'mlly'
 import { basename, dirname, extname, join, resolve } from 'pathe'
-import { normalizeRequestId, pathFromRoot, toFilePath } from 'vite-node/utils'
+import { normalizeRequestId, pathFromRoot } from 'vite-node/utils'
 import type { ModuleCacheMap } from 'vite-node/client'
 import { getAllMockableProperties, getType, getWorkerState, mergeSlashes, slash } from '../utils'
 import { distDir } from '../constants'
@@ -195,7 +195,7 @@ export class VitestMocker {
     return existsSync(fullPath) ? fullPath : null
   }
 
-  public mockObject(object: Record<string | symbol, any>) {
+  public mockObject(object: Record<Key, any>, mockExports: Record<Key, any> = {}) {
     if (!VitestMocker.spyModule) {
       throw new Error(
         'Error: Spy module is not defined. '
@@ -275,7 +275,7 @@ export class VitestMocker {
       }
     }
 
-    const mockedObject: Record<Key, any> = {}
+    const mockedObject: Record<Key, any> = mockExports
     mockPropertiesOf(object, mockedObject)
 
     // Plug together refs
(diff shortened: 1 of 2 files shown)


packages/vitest/src/runtime/mocker.ts
```diff
@@ -360,10 +360,12 @@ export class VitestMocker {
...
         return cache.exports
-      const cacheKey = toFilePath(dep, this.root)
-      const mod = this.moduleCache.get(cacheKey)?.exports || await this.request(dep)
-      const exports = this.mockObject(mod)
+
+      const exports = {}
+      // Assign the empty exports object early to allow for cycles to work. The object will be filled by mockObject()
       this.moduleCache.set(mockPath, { exports })
+      const mod = await this.request(dep)
+      this.mockObject(mod, exports)
```
danez, line 368: I opted to provide the exports object to `mockObject()` to prevent having to copy the object once more here.

So this could also be 
```js
Object.assign(exports, this.mockObject(mod))
```

sheremet-va (approved):

sheremet-va:
Thank you!

Issue #2352, opened 2022-11-19
Vitest cli not working


WolfgangVing:
### Describe the bug

After following the website instruction, and tried to run `vitest --config ./path/to/vitest.config.ts`  I get , and for every vitest cli command is the same.
```PowerShell
vitest: The term 'vitest' is not recognized as a name of a cmdlet, function, script file, or executable program.                      
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
```

Before I actually used a template to start the project, then installed tailwind and chakra ui, at last vitest.

### Reproduction

https://github.com/WolfgangVing/website-3

### System Info

```shell
System:
  OS: Windows 10 10.0.19044
  CPU: (4) x64 AMD A10-9700 RADEON R7, 10 COMPUTE CORES 4C+6G
  Memory: 1.70 GB / 7.44 GB
Binaries:
  Node: 16.13.1 - C:\Program Files\nodejs\node.EXE
  Yarn: 3.2.4 - ~\AppData\Roaming\npm\yarn.CMD
  npm: 8.1.2 - C:\Program Files\nodejs\npm.CMD
Browsers:
  Edge: Spartan (44.19041.1266.0), Chromium (107.0.1418.52)
  Internet Explorer: 11.0.19041.1566
npmPackages:
  @vitejs/plugin-legacy: ^2.3.1 => 2.3.1
  vite: 3.2.4 => 3.2.4
  vitest: ^0.25.2 => 0.25.2
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

azaleta:
what's `vitest --config ./path/to/vitest.config.ts`
[...]

PR #2397, opened 2022-11-30
docs: install global

poyoho:
fix: #2352

### DIFF
--- a/docs/guide/index.md
+++ b/docs/guide/index.md
@@ -35,7 +35,7 @@ Vitest requires Vite >=v3.0.0 and Node >=v14
 
 It is recommended that you install a copy of `vitest` in your `package.json`, using one of the methods listed above. However, if you would prefer to run `vitest` directly, you can use `npx vitest` (the `npx` command comes with npm and Node.js).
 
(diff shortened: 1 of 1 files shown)


TheJaredWilcurt:
I added a comment to the issue, speaking against this direction.

**TLDR:** Instead of recommending `npm install -g vitest` (considered an anti-pattern), instead recommend `npx vitest`, or better, just installing the package in your repo.

docs/guide/index.md
```diff
@@ -33,6 +33,21 @@ pnpm add -D vitest
...
+```
+or with yarn
+```bash
+yarn add -g vitest
+```
+or with pnpm
+```bash
+pnpm add -g vitest
+```
+
```
TheJaredWilcurt, line 50: Based on the discussion in the original ticket, perhaps this would be an acceptable addition to the documentation:

```suggestion
It is recommended that you install a copy of `vitest` in your `package.json`, using one of the methods listed above. However, if you would prefer to run `vitest` directly, you can use `npx vitest` (the `npx` command comes with npm and Node.js).

```

sheremet-va (changes requested):
I don't really think this PR can close #2352. This is a good information to have, but it doesn't really help understanding how it works? We should add a paragraph or something, so people don't create these kind of PRs.

poyoho:
should we need to tell people what diff of `vitest` or `npx vitest`?

`vitest` will use the global vitest, the `npx vitest` should use the local package.

sheremet-va (approved):

Issue #2046, opened 2022-09-15
Add an --exact flag to vitest cli


haikyuu:
### Clear and concise description of the problem

Let's take the following directory for example:
```
timeout.test.ts
each-timeout.test.ts
```
When running `vitest timeout.test.ts`, both test files will run. And that's sometimes an unwanted behaviour.

### Suggested solution

Add a `vitest run --exact` flag that is meant to receive the exact path to the file to test.

### Additional context

Here is an example in Vitest tests themselves that I encountered. And that was a tricky thing to discover 😅

https://github.com/vitest-dev/vitest/blob/95da4d66933db46da368d5d66f422afd26fba4c1/test/fails/test/runner.test.ts

The test runs multiple tests that are supposed to fail for a specific reason. And that reason is save in a snapshot. So when running two tests at the same time instead of one, the wrong reason is picked and the test fails **sometimes**, other times ... it works, depending on the first test to have completed

```diff
  const files = await fg('*.test.ts', { cwd: root })

  for (const file of files) {
    it(file, async () => {
-     await execa('npx', ['vitest', 'run', file])
+     await execa('npx', ['vitest', 'run', '--exact', file])
    })
}
```
That would solve it without having to check 10 files and see if there's an overlap

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

sheremet-va:
You can just pass down whole file path to Vitest:

```shell
npx vitest /tests/file.test.ts 
```

PR #2404, opened 2022-12-01
feat: exact match file basename filter

poyoho:
fix: #2046

### DIFF
--- a/docs/guide/cli.md
+++ b/docs/guide/cli.md
@@ -79,7 +79,7 @@ vitest related /src/index.ts /src/hello-world.js
 | `--inspect` | Enables Node.js inspector |
 | `--inspect-brk` | Enables Node.js inspector with break |
 | `-h, --help` | Display available CLI options |
-| `--exact-filter` | exact match file basename to test |
+| `--exact-filter` | Will only run tests that match the filter exactly. You need to pass down full paths, including root |
 
 ### changed
 
--- a/docs/guide/cli.md
+++ b/docs/guide/cli.md
@@ -79,7 +79,7 @@ vitest related /src/index.ts /src/hello-world.js
 | `--inspect` | Enables Node.js inspector |
 | `--inspect-brk` | Enables Node.js inspector with break |
 | `-h, --help` | Display available CLI options |
-| `--extra-filter` | extra match file basename to test |
+| `--exact-filter` | exact match file basename to test |
 
 ### changed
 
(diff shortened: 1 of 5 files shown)


sheremet-va (changes requested):
Why is it named extra?

poyoho:
Is mean extra match the file basename? @sheremet-va 😁

sheremet-va:
> Is mean extra match the file basename? @sheremet-va 😁

Hm? Issue mentioned the name `--exact`, which is much more descriptive than `--extra`. I would even consider `--exact-filter`

packages/vitest/src/node/core.ts
```diff
@@ -22,6 +22,13 @@ import { VitestCache } from './cache'
 
 const WATCHER_DEBOUNCE = 100
 
+function isMatchTest(a: string, b: string, extra: boolean): boolean {
```
dammy001, line 25: i think we should move this method to utils directory

dammy001:
We should document this after

docs/guide/cli.md
```diff
@@ -79,6 +79,7 @@ vitest related /src/index.ts /src/hello-world.js
 | `--inspect` | Enables Node.js inspector |
 | `--inspect-brk` | Enables Node.js inspector with break |
 | `-h, --help` | Display available CLI options |
+| `--extra-filter` | extra match file basename to test |
```
sheremet-va, line 82: Why extra? "extra" implies "more", here we are limiting it. It should be "exact"

sheremet-va (changes requested):

docs/guide/cli.md
poyoho: oh! sorry

docs/guide/cli.md
```diff
@@ -79,6 +79,7 @@ vitest related /src/index.ts /src/hello-world.js
 | `--inspect` | Enables Node.js inspector |
 | `--inspect-brk` | Enables Node.js inspector with break |
 | `-h, --help` | Display available CLI options |
+| `--exact-filter` | exact match file basename to test |
```
sheremet-va, line 82:
```suggestion
| `--exact-filter` | Will only run tests that match the filter exactly. You need to pass down full paths, including root |
```

sheremet-va (changes requested):

packages/vitest/src/node/cli-api.ts
```diff
@@ -13,6 +13,12 @@ export interface CliOptions extends UserConfig {
    * Override the watch mode
    */
   run?: boolean
+  /**
+   * extra match file to test
```
sheremet-va, line 17:
```suggestion
   * Will only run tests that match the filter exactly. You need to pass down full paths, including root
```

sheremet-va (approved):

sheremet-va:
Now thinking about it, I don't think we need this feature. Users can just pass down full path to filter, and it will work.

poyoho:
> Now thinking about it, I don't think we need this feature. Users can just pass down full path to filter, and it will work.

like `base.spec.ts`

Issue #2403, opened 2022-12-01
accept formatting function for `name` in `test.each` and `test.describe`


scarf005:
### Clear and concise description of the problem

```ts
test.each`
  widget     | res
  ${mock[0]} | ${[pos(0, 0)]}
  ${mock[1]} | ${[pos(1, 0), pos(2, 0)]}
  ${mock[2]} | ${[pos(0, 1), pos(0, 2)]}
  ${mock[3]} | ${[pos(3, 1), pos(3, 2), pos(4, 1), pos(4, 2)]}
`('coordsOf($widget) -> [$res]', ({ widget, res }: Params & { res: Pos[] }) =>
  expect(coordsOf(widget)).toEqual(res)
)
```
is rendered to:

![image](https://user-images.githubusercontent.com/54838975/204952127-436816b6-1822-4e7f-9696-ce6ea7ab8c4a.png)

when each provided inputs get complex and require multiple objects, it's nearly impossible to format them.

### Suggested solution

accept functions as input, such that above test could be rewritten as
```ts
const fmt = (val: { x: number; y: number }) => `(${val.x},${val.y})`

test.each([
  { a: { x: 1, y: 1}, b: { x: 1, y: 1}, expected: { x: 2, y: 2} },
  ...
])(
  ({ a, b, expected }) => `${fmt(a)} + ${fmt(b)} = ${fmt(expected)}`, 
  ({ a, b, expected }) => expect(a + b).toBe(expected)
)
```
which could correctly render as: `(1, 1) + (1, 1) = (2, 2)`

### Alternative

nested object properties access. for example:

(using example syntax `${}`)
```ts
test.each([
  { a: { x: 1, y: 1}, b: { x: 1, y: 1}, expected: { x: 2, y: 2} },
  ...
])('(${a.x}, ${a.y}) + (${b.x}, ${b.y} ) -> (${expected.x}, ${expected.y})', ({ a, b, expected }) => {
  expect(a + b).toBe(expected)
})
```
however this approach quickly gets repetive.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #2405, opened 2022-12-01
feat: format test tablename

poyoho:
fix: #2403

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -244,6 +244,24 @@ You cannot use this syntax, when using Vitest as [type checker](/guide/testing-t
   // ✓ add(2, 1) -> 3
   ```
 
+  You can also access Object attributes with `.`, if you are using objects as arguments:
+
+    ```ts
+    test.each`
+    a               | b      | expected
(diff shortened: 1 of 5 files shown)


sheremet-va (changes requested):
Let’s just use dot notation

poyoho:
emm.. only apply with string temlplate ?

sheremet-va:
> emm.. Do you mean only apply with string temlplate ?

Yeah, I think we should stick with string names: `$name.name.name`

poyoho:
Done!

sheremet-va:
> Done!

Is it? I see `it` still expects a function in this PR. The test name should always be a string, let's not change that. People using current syntax expect `$name.value` to return something, they don't expect to use a function. So if we are implementing this feature, it should be implemented like this:

```ts
it.each`
 name1 | name2
 ${{ val: '1' }} ${{ val: '2' }}
`('val1 - $name1.val, val2 - $name2.val')
```

poyoho:
@sheremet-va done! 😊

sheremet-va (approved):

PR #2412, opened 2022-12-02
docs(coverage): overall improvements

AriPerkkio:
- adds **Ignoring code** section to guides
- adds clear descriptions for each coverage option
- ordered coverage options in a logical order

There have been many questions related to coverage options on Vitest Discrod and here on Github issues. Currently the available options are unclear as some options of both providers are not included in documentation. Instead of stating `C8Options`, the documentation should describe all `c8` options that are supported.

### DIFF
--- a/docs/guide/coverage.md
+++ b/docs/guide/coverage.md
@@ -88,3 +88,24 @@ export default defineConfig({
 ```
 
 Please refer to the type definition for more details.
+
+## Ignoring code
+
+Both coverage providers have their own ways how to ignore code from coverage reports.
+
+- `c8`: https://github.com/bcoe/c8#ignoring-uncovered-lines-functions-and-blocks
+- `ìstanbul` https://github.com/istanbuljs/nyc#parsing-hints-ignoring-lines
+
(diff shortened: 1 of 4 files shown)


docs/guide/coverage.md
```diff
@@ -88,3 +88,24 @@ export default defineConfig({
...
+For `istanbul` provider you can include a `@preserve` keyword in the ignore hint.
+Beware that these ignore hints may now be included in final production build as well.
+
+```diff
+-/* istanbul ignore if */
++/* istanbul ignore if -- @preserve */
+if (condition) {
+```
+
+Unfortunately this does not work for `c8` at the moment.
```
AriPerkkio, line 111: The reason is that these regexps simply don't recognize comments if there is something between the keyword and closing tag.
https://github.com/istanbuljs/v8-to-istanbul/blob/b5ecd8242cf64389c9a656d78530bd0428edf195/lib/source.js#L54-L69

sheremet-va (approved):

sheremet-va:
Thank you!

docs/guide/coverage.md
AriPerkkio: Upstream PR open: https://github.com/istanbuljs/v8-to-istanbul/pull/203

PR #2421, opened 2022-12-03
feat: format test objects (fix #2420)

### LINKED ISSUE #2420: render objects as `{ key: value }` instead of `[object Object]` in test.each

#2405 still does not fix that it's really, really hard to pretty-print nested objects.

I'm interested in submitting PR for this issue.

### Suggested solution

use some existing libraries such as https://www.npmjs.com/package/js-object-pretty-print
```ts
const foo = {
	bar: { x: 1, y: [1, 2, 3] },
	baz: [1, 2, [1, 2, 3]]
}
```
and the following would be interpretted as:
- `$foo`: `{ bar: { x: 1, y: [1, 2, 3] }, baz: [1, 2, [1, 2, 3]] }`
- `$foo.bar`: `{ x: 1, y: [1, 2, 3] }`

### Alternative

use the same formatting method for `%o` for `$`.

### Additional context
even after #2405, `$foo.bar.x $foo.bar.y $foo.baz.qux[1] $foo.baz.qux[2] $foo.baz.qux[3]` is very hard to format and maintain.

### Validations

scarf005:
Resolves #2420

image

this PR enhances `formatTitle` function by pretty-formatting individual objects.
also added test suite for `objectAttr` function.

### DIFF
--- a/packages/vitest/src/runtime/suite.ts
+++ b/packages/vitest/src/runtime/suite.ts
@@ -281,7 +281,7 @@ function formatTitle(template: string, items: any[], idx: number) {
   return formatted
 }
 
-function formatObject(value: unknown, maxDepth = 4): string {
+function formatObject(value: unknown, maxDepth = 3, maxWidth = 2): string {
   const formatFunctions = {
     test: (val: unknown) => typeof val === 'function',
     serialize: (val: Function) => val.toString(),
@@ -291,6 +291,7 @@ function formatObject(value: unknown, maxDepth = 4): string {
     plugins: [formatFunctions],
     escapeRegex: true,
     maxDepth,
+    maxWidth,
     min: true,
   })
 }
--- a/packages/vitest/src/runtime/suite.ts
+++ b/packages/vitest/src/runtime/suite.ts
@@ -1,4 +1,6 @@
 import util from 'util'
+import { serialize } from 'v8'
+import { format as prettyFormat } from 'pretty-format'
 import type { BenchFunction, BenchOptions, Benchmark, BenchmarkAPI, File, RunMode, Suite, SuiteAPI, SuiteCollector, SuiteFactory, SuiteHooks, Task, Test, TestAPI, TestFunction, TestOptions } from '../types'
 import { getWorkerState, isObject, isRunningInBenchmark, isRunningInTest, noop, objectAttr } from '../utils'
 import { createChainable } from './chain'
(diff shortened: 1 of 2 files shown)


poyoho:
I think function should keep the origin output.

poyoho (approved):

packages/vitest/src/runtime/suite.ts
```diff
@@ -273,13 +274,27 @@ function formatTitle(template: string, items: any[], idx: number) {
...
-      return objectAttr(items[0], key)
-    })
+    formatted = formatted.replace(/\$([$\w_.]+)/g,
+      (_, key) => formatObject(objectAttr(items[0], key)),
+    )
   }
   return formatted
 }
 
+function formatObject(value: unknown, maxDepth = 4) {
```
dammy001, line 284: Can you add the return type here?

scarf005: sure, added.

packages/vitest/src/runtime/suite.ts
```diff
@@ -273,13 +274,27 @@ function formatTitle(template: string, items: any[], idx: number) {
...
   return formatted
 }
 
+function formatObject(value: unknown, maxDepth = 4): string {
+  const formatFunctions = {
+    test: (val: unknown) => typeof val === 'function',
+    serialize: (val: Function) => val.toString(),
+  }
+
+  return prettyFormat(value, {
```
sheremet-va, line 290: I think it's not that readable having your whole object in the test tile. Maybe we should use chai's `objDisplay`?

sheremet-va (changes requested):

packages/vitest/src/runtime/suite.ts
scarf005: does vitest ship with chai? If then, I'll switch to it.

sheremet-va: Yes, it uses chai for assertions.

scarf005: ok. tho i think showing full object won't be much of a problem since
1) it has depth limit (currently 3, could be reduced to be lower)
2) you can only print out individual property using `$foo.bar` notation.

sheremet-va: Ok, I checked it myself, and `objDisplay` returns a string. Please, use this function.

sheremet-va: Although inspect also rendered object the same way as `objDisplay` does.

scarf005: alright, changed it to use `objDisplay`. also posted updated screenshot which used `objDisplay` at pr body.
apparently `objDisplay`'s type signature returned void because it lacked `@returns` parameter in jsdoc. filed a PR in chaijs/chai#1490.

sheremet-va: Great 👍🏻

sheremet-va (approved):

PR #2430, opened 2022-12-05
fix: escape string

poyoho:
fix: #2426 

config avoid escapeString

https://github.com/facebook/jest/blob/main/packages/pretty-format/src/index.ts#L147-L151


poyoho:
image
image

sheremet-va (approved):

PR #2431, opened 2022-12-05
feat!: make web-worker implementation more compatible with spec

### LINKED ISSUE #2399: @vitest/web-worker does not structurally clone its postMessage `message` argument
### Describe the bug

First, thanks for this package -- it's been super helpful for me.

Second, when passing values to a worker's `postMessage()`, the specification states that they are copied via a structural clone algorithm. ([Spec](https://developer.mozilla.org/en-US/docs/Web/API/Worker/postMessage)). This only really matters when passing non-primitive values, e.g. `{ prop: value }`.

At present, this wrapper *does not* clone the data, meaning that if the Worker mutates the data directly, it gets reflected in the main thread's data. The fix is relatively simple:
https://github.com/vitest-dev/vitest/blob/main/packages/web-worker/src/pure.ts#L35

Instead of
```
return (this.callbacks[event] || []).map(fn => fn(...data))
```

I imagine something like this would do the trick:
```
const clone = structuredClone(data);
return (this.callbacks[event] || []).map(fn => fn(...clone))
```
(there is also a whole thing about `transfers` of the properties, but that shouldn't be necessary in normal testing scenarios).

`structuredClone` is available in Node 17, but it seems Vite's compatibility starts with Node 14? I'm not an expert in making build systems play nice with polyfills (otherwise I would have just filed the PR...): https://developer.mozilla.org/en-US/docs/Web/API/structuredClone

### Reproduction

My first time sharing something on stackblitz... hope it works. `worker.js` is just the barebones worker, and the testsuite illustrates the copying problem.

https://stackblitz.com/edit/vitest-dev-vitest-e5mxcm

### System Info

```shell
This is just the output from running that command within StackBlitz (this should be irrelevant, but for completeness here it is):

  System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    vite: latest => 3.2.4 
    vitest: latest => 0.25.3
```

### Used Package Manager

npm

### Validations

### LINKED ISSUE #2457: Add SharedWorker support

I would like to unit test my shared worker. I noticed the web worker package, and figured there wouldn't be to much of difference between the two.

### Suggested solution

  Biggest thing would be that there is a connect event and then you listen onto the port  in the event as shown on [MDN](https://developer.mozilla.org/en-US/docs/Web/API/SharedWorker)

### Alternative

_No response_

### Additional context

_No response_

### Validations

sheremet-va:
Fixes #2399
Closes #2457


antfu:
I am not very acknowledged of web-worker, you may merge if you think it's the right thing to do.

Issue #1716, opened 2022-07-25
HTML reporter


flunderpero:
### Clear and concise description of the problem

`@vitest/ui` already creates a beautiful and very useful HTML report. We would like to have something similar created in our CI.

### Suggested solution

Make `@vitest/ui` also a regular reporter.

### Alternative

We can always convert the Junit output to HTML, but the UI of `@vitest/ui` is so much better.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

antfu:
Sounds like a good idea. To make it possible, it might require some bug refactoring to UI to allow reading from static data instead of the WebSocket RPC.

PR #2444, opened 2022-12-06
feat: html report

poyoho:
fix: #1716


packages/ui/vite.report.config.ts
```diff
@@ -0,0 +1,21 @@
+import { defineConfig } from 'vite'
+import defu from 'defu'
```
sheremet-va, line 2: There is `mergeConfig` in `vite`

packages/vitest/src/node/reporters/html.ts
```diff
@@ -0,0 +1,79 @@
...
+      result.files.map(async (file) => {
+        result.moduleGraph[file.filepath] = await getModuleGraph(this.ctx, file.filepath)
+      }),
+    )
+    await this.writeReport(stringify(result))
+  }
+
+  /**
+   * Writes the report to an output file if specified in the config,
+   * or logs it to the console otherwise.
```
sheremet-va, line 46: I don't think this doc is true. It doesn't write it to the console, and I don't think if it should

packages/vitest/src/types/reporter.ts
```diff
@@ -7,6 +7,7 @@ export interface Reporter {
   onPathsCollected?: (paths?: string[]) => Awaitable<void>
   onCollected?: (files?: File[]) => Awaitable<void>
   onFinished?: (files?: File[], errors?: unknown[]) => Awaitable<void>
+  onFinally?: () => Awaitable<void>
```
sheremet-va, line 10: Why do we need another hook? Why is `onFinished` not enough?

sheremet-va (changes requested):
Thank you for your great work! But I have some small questions.

sheremet-va:
We should also probably disable editing in "code" tab, when UI is in report mode.

packages/vitest/src/types/reporter.ts
poyoho: I need to await `this.state.finishCollectingPaths()` the `onFinished` hook will go into endless loop

packages/vitest/src/node/reporters/html.ts
poyoho: yes. Copy it from json report. forget to delete the comment. 🤦

poyoho:
> We should also probably disable editing in "code" tab, when UI is in report mode.

Yes all the `send` rpc control should be hidden?

poyoho:
Add the `app-version` in ui app for debug ui version.

packages/ui/client/components/views/ViewModuleGraph.vue
```diff
@@ -1,214 +1,214 @@
-<script setup lang="ts">
```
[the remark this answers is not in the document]
poyoho: will done in #2449. don't This pr is not relate with this modify.

packages/vitest/src/types/reporter.ts
sheremet-va: Should we move `onFinished` then?

poyoho: ok

packages/vitest/src/node/core.ts
```diff
@@ -336,14 +336,14 @@ export class Vitest {
...
         process.exitCode = 1
 
-      if (!this.config.browser)
-        await this.report('onFinished', files, this.state.getUnhandledErrors())
       this.cache.results.updateResults(files)
       await this.cache.results.writeToCache()
     })()
-      .finally(() => {
+      .finally(async () => {
         this.runningPromise = undefined
```
sheremet-va, line 343: We should move `this.runningPromise` after `onFinished`

poyoho: good catch

poyoho:
Should we need to dump the file content in report by next step? But it is a bit large. 🥹

sheremet-va (changes requested):

packages/ui/client/main.ts
```diff
@@ -13,3 +14,5 @@ Object.entries(directives).forEach(([name, directive]) => {
 })
 
 app.mount('#app')
+
+document.querySelector('#app')?.setAttribute('app-version', version)
```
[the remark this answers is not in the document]
poyoho: because the ui dist is copy to the report. If had any question, the version key can position the ui verison?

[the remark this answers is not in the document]
poyoho: It just a mark. 😀

sheremet-va:
> Also need to update documentation with a separate segment for UI html reporter, mentioning how to preview generated report.

Any plans to implement this?

poyoho:
> > I think we should disable file content preview in report (because it's giant), but still keep module graph tree somehow.
> 
> > Also need to update documentation with a separate segment for UI html reporter, mentioning how to preview generated report.
> 
> Any plans to implement this?

It done?

packages/vitest/package.json
```diff
@@ -76,9 +76,10 @@
     "node": ">=v14.16.0"
   },
   "scripts": {
-    "build": "rimraf dist && rollup -c",
+    "build": "rimraf dist && rollup -c && pnpm copy",
     "dev": "NODE_OPTIONS=\"--max-old-space-size=8192\" rollup -c --watch -m inline",
-    "prepublishOnly": "pnpm build"
+    "prepublishOnly": "pnpm build",
+    "copy": "esno scripts/copy-ui-to-vitest.ts"
```
sheremet-va, line 82: We should not publish @vitest/ui inside vitest.

packages/vitest/src/node/reporters/html.ts
```diff
@@ -0,0 +1,78 @@
+import { existsSync, promises as fs } from 'fs'
```
sheremet-va, line 1: Reporter should be part of @vitest/ui package.

sheremet-va (changes requested):

packages/vitest/src/node/reporters/html.ts
poyoho: It mean `@vitest/ui/reporter` or copy the ui dist from ui lib?

sheremet-va: I mean this file should be inside `@vitest/ui` package. We can add custom logic for html reporter to `createReporters` utility

sheremet-va:
I added more documentation and changed how `outputFile` works, so it actually makes sense for end user.

We show "log" tab. But do we actually store logs? I think it might be nice to store minified logs (with maximum chars). They can be intercepted with `onUserConsoleLog` in reporter.

Screenshot 2022-12-16 at 17 02 20

Also, is it possible to replace paths in html, and don't actually copy the whole UI? Or is it not practical?

poyoho:
I update the ui remove generate one dist for report mode. 

And I think the test need to reset. The snapslot is always diff.
image

sheremet-va (changes requested):
Do you plan to add logs or do you want to hide tab in the UI? I would prefer showing logs.

poyoho:
showing logs?  I forgot test this feature. But it seems no hard to do.

sheremet-va:
> showing logs?

Yeah, I mentioned it a few comments before.

Issue #2411, opened 2022-12-01
Support JSON config file


sagargurtu:
### Clear and concise description of the problem

Support `vitest.config.json` file similar to Jest.

A JSON config file will be helpful in cases where we need to manipulate test configuration dynamically.

### Suggested solution

N/A

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

sheremet-va:
You can always create your own JSON config, and import it inside vitest.config:

```json
{
  "//": "my.custom.config.json",
  "port": 1000
}
```

```ts
import customConfig from './my.custom.config.json'

export default {
  server: {
    port: customConfig.port,
  }
}
```

sagargurtu:
That's a nice workaround. However, we'd need to maintain two files for test configuration. In a monorepo with 'x' packages, this will mean '2x' config files.

PR #2452, opened 2022-12-07
feat: support json config file

poyoho:
fix: #2411


packages/vitest/src/node/create.ts
```diff
@@ -11,10 +12,16 @@ export async function createVitest(mode: VitestRunMode, options: UserConfig, vit
...
 
-  const configPath = options.config
+  let configPath = options.config
     ? resolve(root, options.config)
     : await findUp(configFiles, { cwd: root } as any)
 
+  let jsonConfig = {}
+  if (configPath && configPath.endsWith('json')) {
+    jsonConfig = JSON.parse(readFileSync(configPath, 'utf-8'))
+    configPath = undefined
```
sheremet-va, line 22: This will use several configs (json and vite.config), if you have separate configs for Vite and Vitest, because undefined `configPath` will force Vite to use close one.

It's better to add this support to Vite directly, which doesn't really make sense, because you can't add plugins there.

sheremet-va (changes requested):

PR #2463, opened 2022-12-08
feat!: rewrite how vite-node resolves id

### LINKED ISSUE #2498: importActual hanging with mocked import of circular dependencies
### Describe the bug

Vitest is hanging when using `vi.mock()` along with `vi.importActual()` on a circular dependency.
The same test works in Jest - `jest.mock()` along with `jest.requireActual()`.

There is already a similar(maybe slightly different) [issue](https://github.com/vitest-dev/vitest/issues/546) that appears to be resolved so I decided to create a new one.

### Reproduction

 https://stackblitz.com/edit/vitest-circular-deps-repro?file=main.test.js

### System Info

```shell
Repro in stackblitz
```

### Used Package Manager

npm

### Validations

### LINKED ISSUE #2501: Imports of dependencies of dependencies not found `vitest
### Describe the bug

After upgrading from `0.25.5` to any more recent version up to the current `0.25.8` I'm getting import errors of the dependencies of those imported packages. This isn't happening with all of them, but for example it is with Sentry.

```
        "@sentry/svelte": "^7.26.0",
        "@sentry/tracing": "^7.26.0",
```

Works fine during runtime, however when running tests, I get the message:

```
Error: Failed to load url @sentry/browser (resolved id: @sentry/browser). Does the file exist?
```

This is only silenced by manually adding `@sentry/browser` as a direct dependency, however then the next messages is about `@sentry/util` and so on.

### Reproduction

Add `sentry/browser` as a dependency and try to write a test with it.

### System Info

```shell
System:
    OS: macOS 12.6.1
    CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
    Memory: 145.65 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.18.0 - ~/.nvm/versions/node/v16.18.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v16.18.0/bin/yarn
    npm: 8.19.2 - ~/.nvm/versions/node/v16.18.0/bin/npm
  Browsers:
    Brave Browser: 108.1.46.134
    Chrome: 108.0.5359.98
    Firefox: 105.0.3
    Safari: 16.1
  npmPackages:
    vite: ^4.0.1 => 4.0.1
    vitest: 0.25.5 => 0.25.5
```
```


### Used Package Manager

pnpm

### Validations

### PR DESCRIPTION
**Pull request by sheremet-va**
This PR adds resolveUrl step before calling `fetchModule`. This should help reduce the number of times we are resolving ID (which can be 2-4 times for the same ID, including when mocking and to get actual filepath for coverage, and if module is ambiguous), and also helps with consistency in the cache, since we always normalize paths.

@antfu need your review, before we can merge this.

I checked it shouldn't affect Histoire, because they are using `resolveId`, but it can give errors in Nuxt, so it's better to check, if everything works fine here. For example, this can add inconsistencies in cache, because it depended on `toFilePath` previously, but Nuxt doesn't return anything in its `resolveId` hook, so we use ID for cache there, so maybe you need to move `toFilePath` to Nuxt? (side note: until this PR, Nuxt tries to get importers in its `resolveId` option, but it's actually wasn't called most of the time - only for ambiguous modules)

This PR also fixes bug, when `__dirname` was normalized on Windows (had posix delimiter), but `__filename` had Windows path delimiter.

Closes #2501
Fixes #2498

### CODE REVIEW

File: packages/vite-node/src/client.ts [hunk 1]
```diff
@@ -187,79 +191,63 @@ export class ViteNodeRunner {
...
+    }
 
     const requestStubs = this.options.requestStubs || DEFAULT_REQUEST_STUBS
     if (id in requestStubs)
       return requestStubs[id]
 
     // eslint-disable-next-line prefer-const
-    let { code: transformed, externalize } = await this.options.fetchModule(resolvedId || dep)
-
-    // in case we resolved fsPath incorrectly, Vite will return the correct file path
```
sheremet-va, line 247: This is the reason why I decided to rewrite this mess with IDs 😄

sheremet-va:
@antfu really want to know your stance on this before I proceed deeper 😄

sheremet-va:
@AriPerkkio hey, after this update coverage started failing. Can you help me identify the problem?

antfu (commented):
This seems to make the id handling much clear. I think it's a good initiative!

I would need to check and ensure that it works in Nuxt. If so, I think we could proceed.

AriPerkkio:
> [...] coverage started failing. Can you help me identify the problem?

Looks like you accidently fixed something. :smile:
This is likely somehow related to how Vitest's C8 provider reads the source maps.

<details>
  <summary>Counter.vue</summary>

https://github.com/vitest-dev/vitest/blob/0967b24bc4e50c696bd5a2c9387d432d8b9de51b/test/coverage-test/src/Counter/Counter.vue#L8

Earlier coverage report of this branch stated that it started at column 18, and **ended at column 13**. So it ended before it started.
Now the branch ending seems better. The column position is still off, but that's out of scope. Istanbul works better for Vue anyways.

```diff
  "start": {
    "column": 18,
    "line": 8,
  },
  "end": {
-   "column": 13,
+   "column": 25,
    "line": 8,
  }
```

</details>

<details>

  <summary>utils.ts</summary>

https://github.com/vitest-dev/vitest/blob/0967b24bc4e50c696bd5a2c9387d432d8b9de51b/test/coverage-test/src/utils.ts#L14-L19

This seems much better now. Earlier the branch ending leaked outside of the if-block. Now it ends properly on the line.

Before:

```json
{
"start": {
  "column": 4,
  "line": 16,
},
"end": {
  "column": 9,
  "line": 18,
}
```

```ts
14 | export function sqrt(a: number) {
15 |   if (a < 0)
16 |     return Number.NaN // This should not be covered
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
17 |
18 |   return Math.sqrt(a)
   |   ^^^^^^
19 | }
```

Now: 

```diff
{
  "start": {
    "column": 4,
    "line": 16,
  },
  "end": {
-   "column": 9,
+    "column": 51,
-    "line": 18,
+    "line": 16,
  },
}
```

```ts
14 | export function sqrt(a: number) {
15 |   if (a < 0)
16 |     return Number.NaN // This should not be covered
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
17 |
18 |   return Math.sqrt(a)
19 | }
```

Also the statement counter shows uncovered line now correctly:

```diff
 "path": "<process-cwd>/src/utils.ts",
 "s": {
   "0": 1,
   "1": 1,
   ...
-  "16": 1,
+  "16": 0,
```

</details>

You can safely update the snapshots using this flag and running `pnpm test` in the `test/coverage-test` directory.

https://github.com/vitest-dev/vitest/blob/0967b24bc4e50c696bd5a2c9387d432d8b9de51b/test/coverage-test/testing.mjs#L3-L4

sheremet-va:
> Looks like you accidently fixed something. 😄

This is great news 😄

sheremet-va:
> I would need to check and ensure that it works in Nuxt. If so, I think we could proceed.

I came up with a solution that should make this changes backward compatible with Nuxt. Please, check that you are using latest version of this branch, when checking compatibility 😄

antfu (approved):
Great job! 👍 Thanks

PR #2466, opened 2022-12-08
fix(coverage-istanbul): clear coverage map after use

AriPerkkio:
Adds snapshot test for `istanbul` provider's coverage result. This should provide confidence for future changes. I'll do same for `c8` some day.

While building the test case I encountered a bug:
When coverage is run with `threads: false`, it is possible that statement counts of coverage report are duplicated. The `coverageProvider.takeCoverage()` (Vitest's internal API, not V8) is called twice without resetting it in between. This can be reproduced in Vitest repository - revert the fix and run `test/coverage-test` tests.

### DIFF
--- a/packages/coverage-istanbul/src/index.ts
+++ b/packages/coverage-istanbul/src/index.ts
@@ -7,5 +7,11 @@ export async function getProvider() {
 
 export function takeCoverage() {
   // @ts-expect-error -- untyped global
-  return globalThis[COVERAGE_STORE_KEY]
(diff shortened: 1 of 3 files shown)


AriPerkkio:
Windows CI failing

```diff
test/coverage-test test:   -     "path": "<root>/src/utils.ts",
test/coverage-test test:   +     "path": "D:/a/vitest/vitest/test/coverage-test/src/utils.ts",
```

sheremet-va (approved):

PR #2470, opened 2022-12-08
feat: support Vite 4

sheremet-va:


packages/vite-node/package.json
```diff
@@ -80,7 +80,7 @@
     "pathe": "^0.2.0",
     "source-map": "^0.6.1",
     "source-map-support": "^0.5.21",
-    "vite": "^3.0.0"
```
Akryum, line 83: Shouldn't this be moved to `peerDependencies`?

sheremet-va: Hm, technically you don't need vite for `ViteNodeRunner` 🤔 

But vite-node can also be used as CLI, where it is required, so

docs/package.json
```diff
@@ -24,9 +24,9 @@
     "https-localhost": "^4.7.1",
     "unocss": "^0.46.0",
     "unplugin-vue-components": "^0.22.9",
-    "vite": "^3.2.3",
-    "vite-plugin-pwa": "0.13.1",
-    "vitepress": "1.0.0-alpha.27",
+    "vite": "^4.0.0",
+    "vite-plugin-pwa": "0.13.3",
```
userquin, line 28: pwa plugin not yet ready for Vite 4: I've sent rn the PR to the repo, maybe tmr I can release it

sheremet-va: Docs appear to work fine anyway. We can merge this now, and update plugin later.

DenebTM:
I have found this to cause a type mismatch in vite.config.ts when used together with vite@3.2.5 and vite-plugin-react@3.0.0, due to (I'm pretty sure) a difference in the `PluginOptions` interface between Vite 3 and 4.

Please see the attached output of `npx tsc vite.config.ts`: [tsc.txt](https://github.com/vitest-dev/vitest/files/10201058/tsc.txt)

I can produce a minimal reproducible example if need be.

Issue #2190, opened 2022-10-22
Setup file(s) end up in the coverage report


KingSora:
### Describe the bug

After running the tests the coverage report contains the setup files which was provided in the config. This happens only on my linux machine and not on my windows machine.

Linux output:
```
-------------------------|---------|----------|---------|---------|-------------------
File                     | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
-------------------------|---------|----------|---------|---------|-------------------
All files                |     100 |      100 |     100 |     100 |
 home/projB/local/vitest |     100 |      100 |     100 |     100 |
  vitest.setup.js        |     100 |      100 |     100 |     100 |
 src                     |     100 |      100 |     100 |     100 |
  MyComponent.vue        |     100 |      100 |     100 |     100 |
-------------------------|---------|----------|---------|---------|-------------------
```

Windows output:
```
-----------------|---------|----------|---------|---------|-------------------
File             | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
-----------------|---------|----------|---------|---------|-------------------
All files        |     100 |      100 |     100 |     100 |
 MyComponent.vue |     100 |      100 |     100 |     100 |
-----------------|---------|----------|---------|---------|-------------------
```

As a  workaround I've added the setup file(s) and the `include` globs to the `exclude` array of the `config.coverage` object. 
**Side note:** If I would add only the setup file(s) to the `config.coverage.exclude` array the `.test` files would show up in the coverage report (which is also a bug I suppose). - Thats why I've also added the `include` globs to it.

### Reproduction

Here is the minimal reproduction: https://stackblitz.com/edit/vitest-dev-vitest-dboudc 

> __Note__: **You have to download the project and run `npm i && npm run test` because for some reason the coverage report is always empty on Stackblitz.**

### System Info

```shell
## On Wndows: 

System:
    OS: Windows 10 10.0.22000
    CPU: (16) x64 12th Gen Intel(R) Core(TM) i7-1260P
    Memory: 2.98 GB / 15.69 GB
  Binaries:
    Node: 16.17.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 8.15.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22000.120.0), Chromium (106.0.1370.52)
[...]

PR #2489, opened 2022-12-12
fix: Removed setup files from coverage by adding them on coverage.exclude

g4rry420:

Closes: #2190


sheremet-va (changes requested):
1. Please, add a test for this.
2. We should not modify anything inside `.map`

PR #2551, opened 2022-12-21
feat(mock): expose a `importOriginal` helper to the factory

antfu:
This could be helpful to avoid duplicating the path with `vi.importActual` to extend the existing imports.


sheremet-va (approved):

dammy001 (approved):

Maxim-Mazurok:
This does seem like a pretty great feature, however, this doesn't seem to be included in v0.26.2, yet I can already see this on the docs website for this version: https://vitest.dev/api/#vi-mock

Perhaps docs should only be published when the actual change is published.

PR #2558, opened 2022-12-23
docs: add describe.skipIf documentation

RobbeCl:
SkipIf was not yet in the api for describe

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -499,8 +499,7 @@ When you use `test` or `bench` in the top level of file, they are collected as p
 
 ### describe.skipIf
 
-- **Type:** `(condition: any) => Test`
-- **Alias:** `it.skipIf`
+- **Type:** `(condition: any) => void`
 
   In some cases you might run suites multiple times with different environments, and some of the suites might be environment-specific. Instead of wrapping the suite with `if`, you can use `describe.skipIf` to skip the suite whenever the condition is truthy.
 
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -501,7 +501,7 @@ When you use `test` or `bench` in the top level of file, they are collected as p
 
 - **Type:** `(condition: any) => void`
 
-  In some cases you might run suites multiple times with different environments, and some of the suites might be environment-specific. Instead of wrapping the suite with `if`, you can use `describe.skipIf` to skip the suite whenever the condition is truthy.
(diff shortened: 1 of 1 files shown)


docs/api/index.md
```diff
@@ -497,6 +497,26 @@ When you use `test` or `bench` in the top level of file, they are collected as p
   })
   ```
 
+### describe.skipIf
+
+- **Type:** `(condition: any) => void`
+
+  In some cases you might run suites multiple times with different environments, and some of the suites might be environment-specific. Instead of wrapping the suite with `if`, you can use `describe.skipIf` to skip the suite whenever the condition is truthy.
```
dammy001, line 504:
```suggestion
  In some cases, you might run suites multiple times with different environments, and some of the suites might be environment-specific. Instead of wrapping the suite with `if`, you can use `describe.skipIf` to skip the suite whenever the condition is truthy.
```

docs/api/index.md
```diff
@@ -497,6 +497,26 @@ When you use `test` or `bench` in the top level of file, they are collected as p
...
+
+  const isDev = process.env.NODE_ENV === 'development'
+
+  describe.skipIf(isDev)('prod only test', () => {
+    // this test only runs in production
+  })
+  ```
+
+::: warning
+You cannot use this syntax, when using Vitest as [type checker](/guide/testing-types).
```
dammy001, line 517:
```suggestion
You cannot use this syntax when using Vitest as [type checker](/guide/testing-types).
```

docs/api/index.md
```diff
@@ -497,6 +497,26 @@ When you use `test` or `bench` in the top level of file, they are collected as p
...
+- **Type:** `(condition: any) => void`
+
+  In some cases you might run suites multiple times with different environments, and some of the suites might be environment-specific. Instead of wrapping the suite with `if`, you can use `describe.skipIf` to skip the suite whenever the condition is truthy.
+
+  ```ts
+  import { assert, test } from 'vitest'
+
+  const isDev = process.env.NODE_ENV === 'development'
+
+  describe.skipIf(isDev)('prod only test', () => {
```
dammy001, line 511: Why do we need to run tests in prod?

RobbeCl: It is just a copy paste from the skipIf method on `it` (And it is just an example tho :-) )

sheremet-va (approved):

PR #2560, opened 2022-12-23
fix: start tracking module resolution as soon as possible for easier tracking

### LINKED ISSUE #2552: Dynamic imports started failing in 0.26.0
### Describe the bug

I'm having problems with dynamic imports
* everything was working in 0.25.8
* everything is working in 0.26.2 when I run **individual** test files (`npx vitest run individualFile.test.js`)

The problem happens when I test multiple test files that have modules which dynamically import other modules:
```js
// For example
await import(`../../${sth}/${sthElse}.js`)
```

The problem doesn't happen all the time - sometimes the tests pass, sometimes they fails and it's not always the exact same tests that fail. It feels like there's some sort of race condition between the workers when they start loading dynamic modules.

I made sure to await in all places 
```js
test('sth', async () => {
  // render a component that has `async created()` that load dynamic modules 

  await flushPromises() 
  await vi.dynamicImportSettled()
  
  // expect the side effects of the dynamic imports to have happened
})  
```

### Reproduction

It's difficult to put together a reproduction because of the complexity of the project.
Any tips that would help narrow this down are welcome. 🙏 

I tried creating a repro but it's still not ready - couldn't reproduce the flaky failures.

However I bumped into **an interesting problem which may or may not be related**:
* starting the tests in watch mode works `npx vitest`
* running the tests once hangs (CPU 100%) `npx vitest run`

https://stackblitz.com/edit/vitest-dynamic-imports-issues

### System Info

```shell
System:
    OS: macOS 13.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 1.34 GB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.19.0 - ~/.volta/tools/image/node/16.19.0/bin/node
    Yarn: 1.22.18 - ~/.volta/tools/image/yarn/1.22.18/bin/yarn
    npm: 8.19.3 - ~/.volta/tools/image/node/16.19.0/bin/npm
  Browsers:
    Chrome: 108.0.5359.124
    Firefox: 107.0.1
    Safari: 16.2
  npmPackages:
    @vitejs/plugin-vue: ^3.2.0 => 3.2.0
    vite: ^3.2.5 => 3.2.5
    vitest: ^0.26.2 => 0.26.2
```

### Used Package Manager

npm

### Validations

sheremet-va:
Fixes #2552

I am not sure, if we should reuse `moduleCache` here, @antfu what do you think? I initially wanted to use original ID and update mod path to `fsPath`, but it complicates things too much, I think.

### DIFF
--- a/packages/vite-node/src/client.ts
+++ b/packages/vite-node/src/client.ts
@@ -64,21 +64,30 @@ export class ModuleCacheMap extends Map<string, ModuleCache> {
     return this
   }
 
+  setByModuleId(modulePath: string, mod: ModuleCache) {
+    return super.set(modulePath, mod)
+  }
+
   set(fsPath: string, mod: ModuleCache) {
-    fsPath = this.normalizePath(fsPath)
-    return super.set(fsPath, mod)
+    return this.setByModuleId(this.normalizePath(fsPath), mod)
+  }
+
+  getByModuleId(modulePath: string): ModuleCache {
+    if (!super.has(modulePath))
+      super.set(modulePath, {})
+    return super.get(modulePath)!
   }
 
   get(fsPath: string): ModuleCache {
-    fsPath = this.normalizePath(fsPath)
-    if (!super.has(fsPath))
-      super.set(fsPath, {})
-    return super.get(fsPath)!
+    return this.getByModuleId(this.normalizePath(fsPath))
+  }
+
+  deleteByModuleId(modulePath: string): boolean {
+    return super.delete(modulePath)
   }
 
   delete(fsPath: string) {
-    fsPath = this.normalizePath(fsPath)
-    return super.delete(fsPath)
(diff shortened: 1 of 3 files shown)


packages/vite-node/src/client.ts
```diff
@@ -215,6 +215,18 @@ export class ViteNodeRunner {
     return [resolvedId, fsPath]
   }
 
+  async resolveUrl(id: string, importee?: string) {
+    const resolveKey = `resolve:${id}`
+    // put info about new import as soon as possible, so we can start tracking it
+    this.moduleCache.set(resolveKey, { resolving: true })
```
antfu, line 221: ModuleCache's `set` will apply a normalization. Ideally, in this case, we could expose `_set` `_delete` without normalization to be used here, and avoid extra normalization costs.

PR #2248, opened 2022-10-31
fix: fix stack trace in errors and console

sheremet-va:
Fixes #2231
Fixes #1331
Fixes #2211

PR #2563, opened 2022-12-25
fix(api): make api parse error stacks and return sourcePos in onTaskUpdate

adrienbaron:
Starting with [this PR](https://github.com/vitest-dev/vitest/pull/2248) included in [Vitest 0.25.0](https://github.com/vitest-dev/vitest/releases/tag/v0.25.0), Vitest api hasn't been parsing and returning the stacks when a test fails.

This is due to the removal of [this piece of code](https://github.com/vitest-dev/vitest/pull/2248/files#diff-d6bd07e0f22c35d34b1ed42ab42349ba26e7a1a063d3880961b4317ccd3c0b93L157).

A side effect of this is that the VSCode extension for Vitest doesn't show the error at the line where the assertion failure occurs, but rather on the failing test itself.

**Here is the bug in action:**

https://user-images.githubusercontent.com/4410247/209474578-402a6b6a-05d6-4dac-aa0f-d2e98f4a6df3.mov

As you can see, the error displays on the `it` block itself rather than on the the expect.

With this fix, we reintroduce parsing the error stack before sending it out. We also add the `sourcePos` attribute in the stack,  which is simply the position of the error as the position is now always the correct one (as sourcemaps are loaded in the Node environment).

**This is what it looks like with the fix:**

https://user-images.githubusercontent.com/4410247/209474618-900d57ae-64e5-48ae-aa1a-d4147967c056.mov

### DIFF
--- a/packages/vitest/src/api/setup.ts
+++ b/packages/vitest/src/api/setup.ts
@@ -1,4 +1,5 @@
 import { promises as fs } from 'node:fs'
+
 import type { BirpcReturn } from 'birpc'
 import { createBirpc } from 'birpc'
 import { parse, stringify } from 'flatted'
@@ -8,6 +9,7 @@ import { API_PATH } from '../constants'
 import type { Vitest } from '../node'
 import type { File, ModuleGraphData, Reporter, TaskResultPack, UserConsoleLog } from '../types'
 import { getModuleGraph } from '../utils'
+import { parseStacktrace } from '../utils/source-map'
 import type { TransformResultWithSource, WebSocketEvents, WebSocketHandlers } from './types'
 
 export function setup(ctx: Vitest) {
@@ -120,6 +122,11 @@ class WebSocketReporter implements Reporter {
   async onTaskUpdate(packs: TaskResultPack[]) {
     if (this.clients.size === 0)
       return
+
+    packs.forEach(([, result]) => {
+      if (result?.error)
+        result.error.stacks = parseStacktrace(result.error)
+    })
 
     this.clients.forEach((client) => {
       client.onTaskUpdate?.(packs)
--- a/packages/vitest/src/node/error.ts
+++ b/packages/vitest/src/node/error.ts
@@ -1,6 +1,6 @@
 /* eslint-disable prefer-template */
 import { existsSync, readFileSync } from 'fs'
-import { join, normalize, relative } from 'pathe'
+import { normalize, relative } from 'pathe'
 import c from 'picocolors'
 import cliTruncate from 'cli-truncate'
 import type { ErrorWithDiff, ParsedStack, Position } from '../types'
@@ -12,12 +12,6 @@ import { type DiffOptions, unifiedDiff } from '../utils/diff'
 import type { Vitest } from './core'
 import { divider } from './reporters/renderers/utils'
 import type { Logger } from './logger'
-
-export function fileFromParsedStack(stack: ParsedStack) {
-  if (stack?.sourcePos?.source?.startsWith('..'))
-    return join(stack.file, '../', stack.sourcePos.source)
-  return stack.file
-}
 
 interface PrintErrorOptions {
   type?: string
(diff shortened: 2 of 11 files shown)


sheremet-va (changes requested):
There is no reason to parse stack trace anymore, because Node should throw error with correct stack trace. VS Code extension should be updated to accommodate this change, not Vitest.

packages/vitest/src/utils/source-map.ts
```diff
@@ -83,6 +83,11 @@ export function parseStacktrace(e: ErrorWithDiff, full = false): ParsedStack[] {
         file,
         line: parseInt(lineNumber),
         column: parseInt(columnNumber),
+        sourcePos: {
```
sheremet-va, line 86: After a little but of retrospective, I think the change in `api/setup` can stay, but this change is useless in the grand scheme of things, and VS Code extension should rewrite their code.

packages/vitest/src/api/setup.ts
```diff
@@ -121,6 +123,11 @@ class WebSocketReporter implements Reporter {
     if (this.clients.size === 0)
       return
 
+    packs.forEach(([, result]) => {
+      if (result?.error)
```
sheremet-va, line 127: Error can be anything, there should be some kind of check.

sheremet-va (changes requested):

packages/vitest/src/utils/source-map.ts
adrienbaron: Fair that make sense 👍, I can a PR there to make it default to use the actual values if `sourcePos` is undefined 👍. We probably should remove `sourcePos` from the type though, from what I've seen across the repo it doesn't seem to be set anywhere anymore.

Do you want me to open a PR that removes it accross the codebase?

packages/vitest/src/api/setup.ts
adrienbaron: According to TS it can only be an `ErrorWithDiff` here 👍. Are there cases where at runtime it doesn't actually match this 🙂? If so we probably should update the types 🙂

image

packages/vitest/src/utils/source-map.ts
adrienbaron: Just updated the PR to not include that change 👍!

sheremet-va: > Fair that make sense 👍, I can a PR there to make it default to use the actual values if `sourcePos` is undefined 👍. We probably should remove `sourcePos` from the type though, from what I've seen across the repo it doesn't seem to be set anywhere anymore.
> 
> Do you want me to open a PR that removes it accross the codebase?

You are right, we don't use it anymore. If it's ok with you, we can remove `sourcePos` in this pr.

adrienbaron: Done ✅!

packages/vitest/src/node/error.ts
```diff
@@ -64,14 +58,13 @@ export async function printError(error: unknown, ctx: Vitest, options: PrintErro
...
-    printStack(ctx, stacks, nearest, errorProperties, (s, pos) => {
+    printStack(ctx, stacks, nearest, errorProperties, (s) => {
       if (showCodeFrame && s === nearest && nearest) {
-        const file = fileFromParsedStack(nearest)
         // could point to non-existing original file
         // for example, when there is a source map file, but no source in node_modules
-        if (nearest.file === file || existsSync(file)) {
-          const sourceCode = readFileSync(file, 'utf-8')
-          ctx.logger.error(c.yellow(generateCodeFrame(sourceCode, 4, pos)))
+        if (existsSync(nearest.file)) {
```
sheremet-va, line 65: This check can also be removed, because it's already done on line 45. It was redone again in the previous iteration because of `fileFromParsedStack`

sheremet-va (changes requested):

packages/vitest/src/node/error.ts
adrienbaron: Oh nice catch! will remove 👍

sheremet-va (approved):

packages/vitest/src/node/error.ts
```diff
@@ -213,12 +200,11 @@ function printStack(
 export function generateCodeFrame(
   source: string,
   indent = 0,
-  start: number | Position = 0,
-  end?: number,
+  pos: Position,
```
sheremet-va, line 203: Oh. This is a stack, not a position. Let's also remove `Position` type, because we don't set it anywhere, and it uses `source` property instead of `file`, which might be confusing.

sheremet-va (changes requested):

packages/vitest/src/utils/source-map.ts
adrienbaron: Should be good now! A test isn't passing but I think it's unrelated to this PR (some web worker test failing to start on Windows)

packages/vitest/src/node/error.ts
adrienbaron: Done ✅!

packages/vitest/src/node/error.ts
adrienbaron: Yeah, I thought the same, but was wary of removing Position as it was used somewhere else. Just refactored `Position` out 🙇‍♂️!

sheremet-va (approved):

Issue #2537, opened 2022-12-20
Double reporters with config and cli parameter


polRk:
### Describe the bug

In console prints two reports if one define in vitest.config.ts and another define via cli arg `--reporter=verbose`

```output
$ npx vitest --reporter=verbose

 DEV  v0.16.0 ~/packages/h


 DEV  v0.16.0 ~/packages/h

 ✓ src/h.test.ts (5)
 ✓ src/h.test.ts (5)
   ✓ h (5)
     ✓ should render open and close tags
     ✓ should render attributes
     ✓ should escape double quote in attribute value
     ✓ should stringify attribute value
     ✓ should render body

Test Files  1 passed (1)
     Tests  5 passed (5)
      Time  484ms (in thread 1ms, 48363.75%)


Test Files  1 passed (1)
     Tests  5 passed (5)
      Time  484ms (in thread 1ms, 48400.27%)


 PASS  Waiting for file changes...
       press h to show help, press q to quit

 PASS  Waiting for file changes...
       press h to show help, press q to quit
```

### Reproduction

```
import { defineConfig } from 'vitest/config'

export default defineConfig({
	test: {
		reporters: 'default',
	},
})
```

npx vitest --reporter=verbose

### System Info

```shell
System:
    OS: macOS 12.6.1
    CPU: (8) arm64 Apple M1
    Memory: 110.00 MB / 8.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 19.0.0 - ~/.nvm/versions/node/v19.0.0/bin/node
    npm: 9.0.1 - ~/.nvm/versions/node/v19.0.0/bin/npm
  Browsers:
    Chrome: 108.0.5359.124
    Safari: 16.2
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
[...]

PR #2573, opened 2022-12-28
fix(cli): allow overrides reporter via cli option

mysteryven:
fix: #2537

### DIFF
--- a/packages/vitest/src/node/config.ts
+++ b/packages/vitest/src/node/config.ts
@@ -201,11 +201,6 @@ export function resolveConfig(
     // @ts-expect-error from CLI
     const reporters = resolved.reporter ?? resolved.reporters
     resolved.reporters = Array.from(new Set(toArray(reporters))).filter(Boolean)
-    resolved.reporters = reporter
-      ? toArray(reporter)
-      : Array.from(new Set(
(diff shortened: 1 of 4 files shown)


packages/vitest/src/node/config.ts
```diff
@@ -198,11 +198,13 @@ export function resolveConfig(
     resolved.related = toArray(options.related).map(file => resolve(resolved.root, file))
 
   if (mode !== 'benchmark') {
-    resolved.reporters = Array.from(new Set([
-      ...toArray(resolved.reporters),
-      // @ts-expect-error from CLI
-      ...toArray(resolved.reporter),
-    ])).filter(Boolean)
+    // @ts-expect-error from CLI
+    const reporter = resolved.reporter
```
sheremet-va, line 202:
```suggestion
    const reporters = resolved.reporter ?? resolved.reporters
    resolved.reporters = Array.from(new Set(
      toArray(resolved.reporters)),
    ).filter(Boolean)
```

sheremet-va (changes requested):

mysteryven:
The CI is always down on ubuntu, I am trying to solve it, please wait some time :)

sheremet-va (approved):

PR #2574, opened 2022-12-29
fix: remove setup files from coverage

g4rry420:

Ref: https://github.com/vitest-dev/vitest/pull/2489
Closes: #2190


packages/vitest/src/node/config.ts
```diff
@@ -190,6 +190,7 @@ export function resolveConfig(
         ?? resolve(resolved.root, file),
     ),
   )
+  resolved.coverage.exclude.push(...resolved.setupFiles.map(file => file.replace(`${resolved.root}/`, '')))
```
sheremet-va, line 193: Maybe it's better to use `relative` from `pathe` module?

sheremet-va (changes requested):
Please, add an actual test case to `test/coverage`. Current test is pretty meaningless.

g4rry420:
Thank you for suggesting the use of `relative` from `pathe` module. I have added that.
For testing, honestly, I am not really good with it. Can you please suggest me what should I add in it ?

sheremet-va:
> Thank you for suggesting the use of `relative` from `pathe` module. I have added that.
> For testing, honestly, I am not really good with it. Can you please suggest me what should I add in it ?

There is `test/coverage-test` folder. You need to add a setup file and check that it's not in the report.

sheremet-va (changes requested):
I don't see a check inside coverage test that setup file is not in a report?

g4rry420:
I have added the test. Please, let me know if you see any issues :)

g4rry420:
One thing, I have noticed is that, if for some reason, a user intentionally wants to include the `setup` file in the coverage, then they can't do that (even if they mention that file in the `coverage.include` option). I believe, this is expected.

sheremet-va (approved):

AriPerkkio (changes requested):
Currently the test case does not fail if the implementation is commented out. Let's add some more checks there:

- Keep the current `setup.ts` and its test assertions as is
- Create `test/coverage-test/src/another-setup.ts`
- Use `src/another-setup.ts` as second setupFile in configuration
- Add assertions to the test case to check that it does not end up in the coverage report folder

Final test case should assert that both setupFiles are not present in coverage report.

AriPerkkio (approved):

sheremet-va (approved):
@AriPerkkio should we add a test that file imported in setupFile is still inside coverage report? This is expected behaviour, right?

```ts
// ./utils.ts
export function defineGlobal() {
  globalThis.testing = true
}
```

```ts
// ./vitest.setup.ts
import { defineGlobal } from './utils.ts'
defineGlobal()
```

Setup file should not be in the report, but utils should?

sheremet-va (changes requested):
^ See comment above

AriPerkkio:
> This is expected behaviour, right?

The imported files will end up in coverage report and we should not try to change that - it would be quite complex. I don't think adding test case for validating it is that important.

The current changes look good to me, and tests are covering the functionality well.

sheremet-va (approved):

PR #2582, opened 2022-12-30
fix: type issue with spyOn method (fix #2365)

### LINKED ISSUE #2365: Issue with `vi.spyOn` while trying to mock classes
### Describe the bug

I'm getting weird behavior with `vi.spyOn(...).mockImplementation[Once](...)` when spying on a class. I created a minimal example which follows the [example in the docs](https://vitest.dev/guide/mocking.html#cheat-sheet:~:text=have%20returned%20object-,Example%20with%20vi.spyOn,-%3A) almost exactly.

Typescript is giving me an error on the following snippet of code:
```typescript
import * as exports from '../src/foo'
vi.spyOn(exports, 'Foo').mockImplementation(() => {
                         ^ typescript: Property 'mockImplementation' does not exist on type 'never'. [2339]
  const Foo = vi.fn()
  Foo.prototype.hello = vi.fn(() => 'world!')
  return { Foo }
})
```

Vitest happily runs the test anyway and fails with the following error:
```text
TypeError: mock.hello is not a function
 ❯ test/foo.test.ts:13:17
     11|
     12|     const mock = new exports.Foo()
     13|     expect(mock.hello('asdf')).toEqual('hello world!')
       |                 ^
     14|   })
     15| })
```

Not ruling out user error, but I'm fairly certain the test is set up correctly. I even tried a few variations of the mocked function. Typescript says `vi.spyOn(...)` is returning the `never` type, which I think indicates `mockImplementation()` shouldn't be callable and therefore shouldn't cause any side effects. However, the `Foo` class still seems to get modified with the removal of its `hello()` method. I can get around the problem by mocking the entire class, but the docs seem to indicate the repro should work.

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-avbfj9?file=test/foo.test.ts

### System Info

```shell
System:
    OS: macOS 13.0.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 3.16 GB / 32.00 GB
    Shell: 3.5.1 - /opt/homebrew/bin/fish
  Binaries:
    Node: 16.17.1 - ~/.asdf/installs/nodejs/lts/bin/node
    Yarn: 2.4.3 - ~/.nvm/versions/node/v16.3.0/bin/yarn
    npm: 8.15.0 - ~/.asdf/plugins/nodejs/shims/npm
  Browsers:
    Chrome: 107.0.5304.110
    Firefox: 107.0
    Safari: 16.1
  npmPackages:
    vitest: ^0.25.2 => 0.25.2
```

### Used Package Manager

npm

### Validations

samkevin1:
I believe the problem was that classes were getting recognized as constructor methods, so it was getting typed incorrectly. Changing the order to check if it's a Class first fixes it.

Related to: #2365


sheremet-va (changes requested):
Can you please add a test case for this?

samkevin1:
Added some test cases @sheremet-va. Can you take a look?

sheremet-va (approved):

sheremet-va (changes requested):
You have conflicts now, spy was moved into a separate package

samkevin1:
Fixed the conflicts

sheremet-va (approved):

Issue #2532, opened 2022-12-19
Watch mode unresponsive to monorepo dependency changes


luxaritas:
### Describe the bug

I'm attempting to use Vitest in a monorepo, and noticed there are issues getting vitest to re-run tests in watch mode when in-repo dependencies change. I've identified a few specific limitations (if these should be split out into separate issues, let me know):
1) `**/dist/**` needs to be excluded from `watchExclude`. This is fine, however I'd expect the default behavior to be smart enough to handle a monorepo where packages are built separately and imported into the package being tested
2) `deps.inline` has to be set to true for changes to be picked up. Again, this is a reasonable workaround, but since vite "just works" for HMR and watch mode to work correctly, I wouldn't expect it to be necessary for vitest.
3) When rebuilding the dependency with Vite, tests are not rerun - even though if I manually change the file in dist, they are. Additionally, building with Vite then trying to change the file in dist (without restarting the test watcher) also does not cause a test rerun. I tried tweaking the chokidar options via vite's server.watch parameter (namely `awaitWriteFinish` and `atomic`) as that seemed like it could affect this, but there was no change in behavior.

### Reproduction

Reproduce via https://stackblitz.com/edit/vitest-dev-vitest-fttlzd (ensure you have two terminals available). Stop any watch processes and reset file states between tests.

1) 1) Run `turbo run --cwd packages/packageb build:watch` in one terminal
   2) Change packages/packagea/src/lib.ts
   3) Run `turbo run --cwd packages/packagea build` in the other terminal
   4) Note packages/packageb/dist/index.js is updated accordingly
2) 1) Run `turbo run --cwd packages/packageb test` in one terminal
   2) Change packages/packagea/dist/src/lib.ts
   3) Note the test watcher updates accordingly
3) 1) Run `turbo run --cwd packages/packageb test` in one terminal
   2) Change packages/packagea/src/lib.ts
   3) Run `turbo run --cwd packages/packagea build` in the other terminal
   3) Note the test watcher does not update
4) 1) Comment out `deps: { inline: true },` (and/or `watchExclude: [],`) in packages/packageb/vite.config.ts
   2) Run `turbo run --cwd packages/packageb test` in one terminal
   3) Change packages/packagea/dist/src/lib.ts
   4) Note the test watcher does not update

Also note that if you run test 3 then 2 without stopping the test watcher, 2 will also fail to update.

### System Info

```shell
System:
  OS: Linux 5.0 undefined
[...]

PR #2592, opened 2023-01-02
fix: rerun tests affected by file changed by build script

luxaritas:
This PR addresses point 3 in #2532.

For some reason, when rebuilding a dependency with vite, the vitest watcher emitted an unlink event, but no add or change event. This change allows vitest's behavior to be consistent with vite's HMR and build watching. Additionally, strictly speaking, it seems sensible to rerun tests that include a removed file somewhere in its dependency chain.

If a test is needed/desired, I'd need some guidance on how to test watch mode behavior.


packages/vitest/src/node/core.ts
```diff
@@ -462,6 +462,9 @@ export class Vitest {
     }
     const onUnlink = (id: string) => {
       id = slash(id)
+      const needsRerun = this.handleFileChanged(id)
+      if (needsRerun)
+        this.scheduleRerun(id)
```
sheremet-va, line 467: This should be done only if source file was deleted (should be `else if` on line 476)

sheremet-va (changes requested):
Test can be written in test/watch

sheremet-va:
> strictly speaking, it seems sensible to rerun tests that include a removed file somewhere in its dependency chain

If the file was removed in the change, then it will probably break on rerun, so I don't see a lot of reason to rerun it - better to wait for the change even in that case

PR #2595, opened 2023-01-02
chore: don't bundle cac and picocolors

sheremet-va:
Do not bundle cac and picocolors, because they are used between our different packages and increase the overall size.


dammy001 (approved):
one test is failing

antfu (approved):

Issue #2299, opened 2022-11-08
Run `vitest typecheck` together with normal test in one `watch`


Shinigami92:
### Clear and concise description of the problem

- As a user
- I want to run `vitest typecheck` and `vitest` in just one command (e.g. just `vitest` or `vitest --with-typecheck`)
- so that I don't need to run e.g. two separate terminal windows in my VSCode editor in parallel to run all tests while I'm working on my project

### Suggested solution

Run `typecheck` by default alongside the normal running tests by default.
(I use `*.test-d.ts` as recommended in the docs)

### Alternative

Let us use a flag for running with `typecheck` via `--with-typecheck` (alternative wording welcome)

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

bugeats:
This could also be addressed by adding a shell command hook to run before reload. Ex:

```bash
vitest --watch --before-run "tsc --noEmit"
```

In my case I have multiple things I'd also like to do on re-test: clear my tmux buffer, run lints, etc.

posva:
It would also be nice if they appear in the UI and are reported through the API

PR #2597, opened 2023-01-02
refactor: make ids in typecheck suites more reliable, store errors in array

sheremet-va:
In preparation for #2299

### DIFF
--- a/packages/vitest/src/typecheck/typechecker.ts
+++ b/packages/vitest/src/typecheck/typechecker.ts
@@ -187,7 +187,7 @@ export class Typechecker {
     const { config, path } = await getTsconfig(root, typecheck)
 
     this.tempConfigPath = path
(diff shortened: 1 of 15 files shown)


packages/vitest/src/typecheck/typechecker.ts
```diff
@@ -179,11 +180,22 @@ export class Typechecker {
...
   }
 
+  public async prepare() {
+    const { root, typecheck } = this.ctx.config
+    await this.ensurePackageInstalled(root, typecheck.checker)
+
+    const { config, path } = await getTsconfig(root, typecheck)
+
+    this.tempConfigPath = path
+    this.allowJs = config.allowJs || typecheck.allowJs || false
```
sheremet-va, line 190:
```suggestion
    this.allowJs = typecheck.allowJs || config.allowJs || false
```

Issue #2477, opened 2022-12-09
use reporter in ci as default reporter


trim21:
### Clear and concise description of the problem

`config.test.reporters` has many available options but non of them is same with current reporter in ci.

I really like the reporter in CI and it it possible to set it as default reporter instead of setting env `CI=true`?

### Suggested solution

add an reporter `default-ci` or some similar name

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

alok-runner:
In a similar vein, verbose with `CI=true` is also good. Especially as a potential "workaround" for #2602

adam-arthur:
Confirmed that the CI reporter is not subject to the degraded performance of other console reporters. 

Thanks! Will use that for my use case now

PR #2612, opened 2023-01-04
feat: add basic reporter which allow user to use reporter in ci

trim21:
Also appending only reporter.

close https://github.com/vitest-dev/vitest/issues/2477


Shinigami92:
@antfu As we have issues with `clearScreen` and getting stuck with it in @faker-js, we are interested in this `basic` reporter

packages/vitest/src/node/reporters/basic.ts
```diff
@@ -0,0 +1,27 @@
+import { DefaultReporter } from './default'
+import type { UserConsoleLog } from '#types'
+
+export class BasicReporter extends DefaultReporter {
```
sheremet-va, line 4: Maybe it should just be extended from a BaseReporter? You don't need to override methods then.

sheremet-va (changes requested):

packages/vitest/src/node/reporters/basic.ts
Shinigami92: On the other side it makes the intend more clean and whenever something changes in `DefaultReporter` this change will go also into `BasicReporter`

sheremet-va: I don't think these two reporters share anything in common. Rendering of "simple" reporter is coded in "base" reporter. Default one just defines its own "live" reporting, and fallbacks to "base", when TTY is not supported.

sheremet-va: I would even add an explicit getter/property to toggle this behaviour instead of relying on "isTTY".

adam-te (approved):
Would be very useful,

sheremet-va:
On my machine this is running 2 seconds more than if I run with `CI=true`. Maybe we need to optimise something else?

trim21:
there is also a similar `isTTY` check inside logger, which I didn't change, it's far too complex.

sheremet-va:
> @sheremet-va need to override `onWatcherRerun` because base reporter may call `clearScreen`

This is desirable behaviour in watch mode.

trim21:
> > @sheremet-va need to override `onWatcherRerun` because base reporter may call `clearScreen`
> 
> This is desirable behaviour in watch mode.

this also cause perf problme ... https://github.com/vitest-dev/vitest/pull/2612#issuecomment-1377797116

sheremet-va:
> there is also a similar isTTY check inside logger, which I didn't change, it's far too complex

I don't see any checks in the logger.

trim21:
> > this also cause perf problme ... [#2612 (comment)](https://github.com/vitest-dev/vitest/pull/2612#issuecomment-1377797116)
> 
> We cannot just remove a feature. Create an issue describing what's happening. Current behaviour is desirable and consistent.

It didn't change any exsiting reporter, this is a new repoter, it didn't remove any feature vitest already have, clear screen on default any existing reporter still work.

sheremet-va:
> It didn't change any exsiting reporter, this is a new repoter, it didn't remove any feature vitest already have, clear screen on default or any existing reporter still work.

Clear screen is staying until we have an issue about this feature. Basic reporter imitates default reporter. If there are any issue with clearScren, open a separate issue.

sheremet-va:
> Clear screen is staying until we have an issue about this feature. Basic reporter imitates default reporter. If there are any issue with clearScren, open a separate issue.

And just as a reminder, Vitest uses `clearScreen` option that you can disable in the config.

Shinigami92:
ping, any news here? we would like to use it in Faker

trim21:
> ping, any news here? we would like to use it in Faker

have you try this branch in your repo? Does it fix your performance issue?

Shinigami92:
Switching to it via https://github.com/faker-js/faker/pull/1763
Works like a charm 👍

Issue #2614, opened 2023-01-05
Rerun setupFiles in watch mode


Maxim-Mazurok:
### Clear and concise description of the problem

I'm mocking http requests using [msw](https://mswjs.io/) and as [suggested in vitest docs](https://vitest.dev/guide/mocking.html) I'm putting my handlers in setup files. 
While running vitest in watch mode I want to edit handlers and to have these changes applied before rerun of tests. However changes to setup files aren't being picked up in watch mode even if I forcefully rerun all tests. I need to restart vitest for changes to apply.

### Suggested solution

Rerun all setupFiles in watch mode, or add an option to do that on a per-file basis.

### Alternative

An alternative is probably using something like nodemon and then run `vitest --run`, but it probably isn't really worth the trouble, hopefully, I won't have to make many changes to mocked API once it's done.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

mysteryven:
> I won't have to make many changes to mocked API once it's done

The `setupFiles` will be run before each test file and not changes often, what about trigger server restart when `setupFiles` content changed just like config file, is it a good implementation ?

Maxim-Mazurok:
> > I won't have to make many changes to mocked API once it's done
> 
> The `setupFiles` will be run before each test file and not changes often, what about trigger server restart when `setupFiles` content changed just like config file, is it a good implementation ?

This does sound better to me. I don't expect these files to change as much as the actual tests during active development, so rerunning them every time might be a bit of a waste, so your suggested optimisation does sound good to me.

sheremet-va:
I think we can just add setupFiles to `forceRerunTriggers`. it will rerun every test, when file changes.

mysteryven:
> I think we can just add setupFiles to `forceRerunTriggers`. it will rerun every test, when file changes.

Great! Thanks for your advice, I will try to solve it.

Maxim-Mazurok:
> it will rerun every test, when file changes.

[...]

PR #2625, opened 2023-01-09
Fix/rerun setup files

mysteryven:
Fix: #2614 

1. I am sorry about I don't know how to mock watch mode by `execa`, so I used tests similar with [test/related](https://github.com/vitest-dev/vitest/tree/main/test/related). Maybe it's not intuitive, but it can tell me setup files have been added into `forceRerunTriggers`. I will be really happy to know whether there are some better way to test it.

2. Do I need to modify the document of `forceRerunTriggers`'s [default values](https://vitest.dev/config/) ? 

```diff
-['**/package.json/**', '**/vitest.config.*/**', '**/vite.config.*/**']
+['**/package.json/**', '**/vitest.config.*/**', '**/vite.config.*/**', '/path/to/setupFiles']
```


sheremet-va (changes requested):
Thank you for your PR! Please, remove pnpm-lock file and add a note to docs that changing setup file will trigger rerun of all user tests.

mysteryven:
> Thank you for your PR! Please, remove pnpm-lock file and add a note to docs that changing setup file will trigger rerun of all user tests.

Done.

sheremet-va (approved):

Maxim-Mazurok:
I can confirm it's working in `0.27.2`, thank you!

For what it's worth, I've switched from manually defining restHandlers for MSW in setup file to using `msw-snapshot` package from NPM, which allows to basically cache actual API responses as snapshots, and then use them in tests. I think it's a much nicer approach, allowing for an easy and good contract.

PR #2626, opened 2023-01-09
fix(coverage): validate extension, when reporting c8 coverage

sheremet-va:
CSS files are not empty anymore, so coverage started failing.

This filters out `.css` virtual files.

### DIFF
--- a/packages/coverage-c8/package.json
+++ b/packages/coverage-c8/package.json
@@ -46,6 +46,7 @@
     "vitest": "workspace:*"
   },
   "devDependencies": {
+    "pathe": "^0.2.0",
     "vite-node": "workspace:*"
   }
 }
--- a/packages/coverage-c8/src/provider.ts
+++ b/packages/coverage-c8/src/provider.ts
@@ -2,7 +2,7 @@ import { existsSync, promises as fs } from 'fs'
 import _url from 'url'
 import type { Profiler } from 'inspector'
 import { takeCoverage } from 'v8'
-import { resolve } from 'pathe'
+import { extname, resolve } from 'pathe'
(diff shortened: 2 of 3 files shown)


AriPerkkio (approved):
I did some debugging and verified that this indeed is caused by the changes in https://github.com/vitest-dev/vitest/pull/2617/files#diff-4889d8f7f93f1cb6cc38e3ae15f827dd8020a0acebe52a9b95d796b216c9ef0dR73.

This condition is no longer triggered for these `file.vue?someQuery...` 
requests but I think it should be left there as it is now. It might catch similar cases in future.

https://github.com/vitest-dev/vitest/blob/fed1cd605fb5e7a47c70519290c8e4f04a9e3449/packages/coverage-c8/src/provider.ts#L77-L83

One small refactoring suggestion for moving the parsing of `file` into a single place but good to go as is.

```diff
  65 |    const filepath = file.split('?')[0]
+    |    const url = _url.pathToFileURL(filepath).href
  66 |
  67 |    return {
  68 |      filepath,
- 69 |      url: _url.pathToFileURL(filepath).href,
- 70 |      id: file,
+ 69 |      url,
+ 70 |      extension: extname(file) || extname(url),
  71 |      map: result.map,
  72 |      source: result.code,
  73 |    }
  74 | })
  75 |  .filter((entry) => {
  76 |    if (!entry)
  77 |      return false
  78 |
- 79 |    const extension = extname(entry.id) || extname(entry.url)
  80 |
- 81 |    if (!extensions.includes(extension))
+ 81 |    if (!extensions.includes(entry.extension))
  82 |      return false
```

AriPerkkio (approved):

PR #2633, opened 2023-01-10
feat: display running processes, if vitest closes with timeout

sheremet-va:


antfu:
Should we make this more general to avoid vendor lock-in? Like hooks before Vitest starts and after ended, etc. I am not sure if there is a case, but maybe someone might want to use a different tracker?

sheremet-va:
> Should we make this more general to avoid vendor lock-in? Like hooks before Vitest starts and after ended, etc. I am not sure if there is a case, but maybe someone might want to use a different tracker?

And what API do you have in mind?

antfu:
Maybe:

```ts
defineConfig({
  test: {
    // similar to Vite Plugin
    extensions: [{
      name: 'why-is-node-running',
      onStart() {
      },
      onFinished(result) {
      },
      onProcessTimeout {
      }
    }]
  }
}
```

We may still have this "extension" built in if you think this is a common usage, but having the infra will help others to integrate other features to Vitest I suppose

sheremet-va:
> We may still have this "extension" built in if you think this is a common usage

Yes, I think it's a good idea to give better error log, when something goes wrong, if we can.

> Maybe

How is it different from reporters then?

antfu:
Or maybe we could have a "hanging-process" reporter and reuse the API? It's reporting the info anyway so I think it could fit into the category

sheremet-va:
> Or maybe we could have a "hanging-process" reporter and reuse the API? It's reporting the info anyway so I think it could fit into the category

Moved to reporter

PR #2639, opened 2023-01-10
Update docs on mocking

ffxsam:
This really threw me off. `mockReset` didn't do the trick, and let mock data bleed between tests. However, setting `restoreMocks: true` in my vitest config solved the issue, so the docs should mention `mockRestore` instead of `mockReset`.

### DIFF
--- a/docs/guide/mocking.md
+++ b/docs/guide/mocking.md
@@ -7,7 +7,7 @@ title: Mocking | Guide
 When writing tests it's only a matter of time before you need to create a "fake" version of an internal — or external — service. This is commonly referred to as **mocking**. Vitest provides utility functions to help you out through its **vi** helper. You can `import { vi } from 'vitest'` or access it **globally** (when [global configuration](/config/#globals) is **enabled**).
 
 ::: warning
-Always remember to clear or restore mocks before or after each test run to undo mock state changes between runs! See [`mockReset`](/api/#mockreset) docs for more info.
+Always remember to clear or restore mocks before or after each test run to undo mock state changes between runs! See [`mockRestore`](/api/#mockrestore) docs for more info.
 :::
 
 If you wanna dive in head first, check out the [API section](/api/#vi) otherwise keep reading to take a deeper dive into the world of mocking.


sheremet-va:
> This really threw me off. `mockReset` didn't do the trick, and let mock data bleed between tests. However, setting `restoreMocks: true` in my vitest config solved the issue, so the docs should mention `mockRestore` instead of `mockReset`.

Can you explain more? What mock data? What were you trying to do? `mockRestore` restores everything to original, I am not sure that's what everyone wants. For example. this might break your tests with `vi.mock`. Usually `mockReset` is enough.

ffxsam:
The following test (indicated below via comment) was breaking until I set `restoreMocks: true` in the config.

```ts
import { describe, it, expect, vi, type Mock } from 'vitest';
import {
  addToBrowser,
  getFromBrowser,
  removeFromBrowser,
} from '../local-storage';

vi.stubGlobal('localStorage', {
  getItem: vi.fn(),
  setItem: vi.fn(),
});

describe('addToBrowser', () => {
  it('checks for existing store', () => {
    (localStorage.getItem as Mock).mockReturnValueOnce(null);
    addToBrowser('mykey', 'myvalue');

    expect(localStorage.getItem).toHaveBeenCalledWith('reelcrafter2-static');
  });

  it('adds to existing store', () => {
    (localStorage.getItem as Mock).mockReturnValueOnce('{"mykey":"myvalue"}');

    addToBrowser('mykey2', 'myvalue2');

    expect(localStorage.setItem).toHaveBeenCalledWith(
      'reelcrafter2-static',
      '{"mykey":"myvalue","mykey2":"myvalue2"}'
    );
  });

  // ** BREAKS ** //
  it('overwrites existing key', () => {
    (localStorage.getItem as Mock).mockReturnValueOnce('{"mykey":"myvalue"}');
    (localStorage.setItem as Mock).mockClear();

    addToBrowser('mykey', 'myvalue2');

    expect(localStorage.setItem).toHaveBeenCalledWith(
      'reelcrafter2-static',
      '{"mykey":"myvalue2"}'
    );
  });
});
```

I had [posted about this on StackOverflow](https://stackoverflow.com/questions/75066320/unit-testing-with-vue-vitest-mock-call-values-are-being-preserved-across-ass) if you want more details.

Mock invocation data was bleeding from the previous test until I set `restoreMocks: true`.

sheremet-va:
> Mock invocation data was bleeding from the previous test until I set restoreMocks: true.

In you stack overflow question you mention that you had `clearMocks`. It only clears the state (mock.calls), and doesn't clear implementations. Did you try enabling `resetMocks` instead? Or maybe it's something different entirely, since I cannot reproduce your error here: https://stackblitz.com/edit/vitest-dev-vitest-y1wzx8?file=test%2Fbasic.test.ts,vite.config.ts&initialPath=__vitest__

ffxsam:
Wow, I'm stumped. I tried `clearMocks: true` again and this time it worked. Actually, removing any mock reset/clear/restore options is fine suddenly. This makes little sense; I wonder if maybe something was cached?

ffxsam:
@sheremet-va Thanks for spending time on this and sorry it was a waste of time!

PR #2645, opened 2023-01-11
fix: terminate workers, when closing process

sheremet-va:
Potential fix for #2008?

### DIFF
--- a/packages/vitest/src/node/pool.ts
+++ b/packages/vitest/src/node/pool.ts
@@ -127,7 +127,9 @@ export function createPool(ctx: Vitest): WorkerPool {
 
   return {
     runTests: runWithFiles('run'),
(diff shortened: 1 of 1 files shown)


sheremet-va:
`pool.destroy()` throws A LOT of error, when calling `pool.destroy()`, when worker started but didn't run code yet:

Terminating workers with `worker.terminate()` doesn't seem to throw errors in this case. It seems like tinypool error, @Aslemammad:

0. Check that you call `pool.destroy()` in close
1. Run `pnpm run test basic`
2. When you see banner with "VITEST" and version, abort the process with `Ctrl+C`
3. See a lot of errors

Aslemammad:
@sheremet-va Thank you so much, could you please create an issue on tinypool so I'd work on it on weekedn? Thank you so much.

PR #2650, opened 2023-01-12
fix: trim input filename and test name

btea:


packages/vitest/src/node/core.ts
```diff
@@ -364,13 +364,13 @@ export class Vitest {
   }
 
   async changeNamePattern(pattern: string, files: string[] = this.state.getFilepaths(), trigger?: string) {
-    this.config.testNamePattern = pattern ? new RegExp(pattern) : undefined
+    this.config.testNamePattern = pattern.trim() ? new RegExp(pattern) : undefined
```
sheremet-va, line 367: Shouldn't trimmed patter be used in `new RegExp`? Also maybe it's better to trim it before `changeNamePattern` is called? The same for `changeFilenamePattern`

sheremet-va (changes requested):

packages/vitest/src/node/core.ts
btea: Maybe here?
https://github.com/vitest-dev/vitest/blob/d6225764856dd62cb1b21ec5bf4619052609c244/packages/vitest/src/node/stdin.ts#L98-L99

sheremet-va: Yes

btea: Updated.

sheremet-va (approved):

PR #2660, opened 2023-01-13
fix(typecheck): store tmp tsconfig close to original one

sheremet-va:


AriPerkkio:
> Could not find a declaration file for module 'vitest'. 'D/a/vitest/vitest/packages/vitest/dist/index.js'

This error message seems to be correct actually. On Windows CI the rollup build is unable to build `.d.ts` files. I'm able to fix this using `relative` from `pathe`. I'll check if this fixes the typecheck errors.

https://github.com/vitest-dev/vitest/blob/24d6380951b3e324016ea32f7342b6fa80d195a7/packages/vitest/rollup.config.js#L109

PR #2664, opened 2023-01-14
chore: update vite as peerDeps

cqh963852:
Considering Vite as peerDeps?

These modifications may cause people who have not explicitly installed vite to be prompted that **vite** cannot be found


sheremet-va:
Vitest can be installed in non-vite projects, so I don't think this is a good idea. Vite-node CLI also cannot be used without Vite.

PR #2670, opened 2023-01-15
chore: add tsup and unbuild config to default excludes

dammy001:


antfu (approved):

packages/vitest/src/defaults.ts
```diff
@@ -1,7 +1,7 @@
 import type { BenchmarkUserOptions, ResolvedCoverageOptions, UserConfig } from './types'
 
 export const defaultInclude = ['**/*.{test,spec}.{js,mjs,cjs,ts,mts,cts,jsx,tsx}']
-export const defaultExclude = ['**/node_modules/**', '**/dist/**', '**/cypress/**', '**/.{idea,git,cache,output,temp}/**', '**/{karma,rollup,webpack,vite,vitest,jest,ava,babel,nyc,cypress}.config.*']
+export const defaultExclude = ['**/node_modules/**', '**/dist/**', '**/cypress/**', '**/.{idea,git,cache,output,temp}/**', '**/{karma,rollup,webpack,vite,vitest,jest,ava,babel,nyc,cypress,tsup,build}.config.*']
```
sheremet-va, line 4: I think we also need to update documentation?

sheremet-va (changes requested):

packages/vitest/src/defaults.ts
dammy001: @dammy001 docs updated

sheremet-va: There are also changes to coverage excludes, which are not documented

dammy001: @sheremet-va updated now.

sheremet-va (approved):

PR #2676, opened 2023-01-16
ci: cache cypress, playwright, puppeteer

AriPerkkio:
Adds caching for browser testing libraries binaries. 

Vite has a bit similar setup for Playwright but I think using Node for resolving the version number is more readable. https://github.com/vitejs/vite/blob/f2ee219e2f4d1a6ac09221dfb7807a684847c595/.github/workflows/ci.yml#L85-L106

Each binary is in their own caches, using OS and library version number as cache keys.

image

### DIFF
--- a/.github/actions/setup-and-cache/action.yml
+++ b/.github/actions/setup-and-cache/action.yml
@@ -1,0 +1,63 @@
+name: Setup and cache
+description: Setup for node, pnpm and cache for browser testing binaries
+inputs:
+  node-version:
+    required: false
+    description: Node version for setup-node
+    default: 18.x
+
+runs:
+  using: composite
+
+  steps:
+    - name: Install pnpm
+      uses: pnpm/action-setup@v2
+
+    - name: Set node version to ${{ inputs.node-version }}
+      uses: actions/setup-node@v3
+      with:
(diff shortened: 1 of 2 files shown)


AriPerkkio:
It's also possible to cache the pnpm store, but based on my testing it seems to be slow on Windows. I think I've seen similar slowness whenever moving lots of small files (`node_modules`) on Windows file systems. 

```yml
    - name: Resolve pnpm store
      id: resolve-pnpm-store
      shell: bash
      run: |
        echo "PNPM_STORE=$(pnpm store path)" >> $GITHUB_OUTPUT

    - name: Cache pnpm store ${{ steps.resolve-pnpm-store.outputs.PNPM_STORE }}
      uses: actions/cache@v3
      id: pnpm-store-cache
      with:
        path: ${{ steps.resolve-pnpm-store.outputs.PNPM_STORE }}
        key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
        restore-keys: |
          ${{ runner.os }}-pnpm-

```

sheremet-va (approved):
Thank you! I think this will be useful.

PR #2681, opened 2023-01-16
docs: add fs-extra to docs dependencies

sheremet-va:


userquin:
I'll send a new PR to use `@vite-pwa/vitepress`

PR #2682, opened 2023-01-16
revert!: don't rely on environment for interopDefault

### LINKED ISSUE #2610: Interop default in TypeScript files
### Describe the bug

In the latest version we disabled `dips.interopDefault` for `node` environments, because we assume users are using Node ESM, but TypeScript can be bundled either way and doesn't actually follow ESM spec.

Because of this we should interop default, if code has tsconfig that doesn't bundle ESM.

### Reproduction

https://github.com/c100k/vitest-0.26.x-breaking-changes

### System Info

```shell
Any
```

### Used Package Manager

npm

### Validations

sheremet-va:
Most of the users don't actually test Node.js applications, so it becomes cumbersome to setup Vitest.

I think we might return to this idea in the future, but with more checks. We should also probably change module resolution to be actually compatible with Node. Until then this is just a distraction.

Closes #2610


antfu (approved):
Not 100% sure about the context, but it seems to make sense.

sheremet-va:
> Not 100% sure about the context, but it seems to make sense.

The initial idea was to follow Node.js spec for non-Node environments. But we forgot about auto interop in TS files. Since this affects most people, I decided to rollback this change, until we can reliably disable it.

Issue #2658, opened 2023-01-13
Support configuring coverage reporter options


AriPerkkio:
### Clear and concise description of the problem

Add support for configuring various options of different Istanbul reporters. This applies to both `@vitest/coverage-c8` and `@vitest/coverage-istanbul`. 

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/276d95e4304b3670eaf6e8e5a7ea9e265a14e338/types/istanbul-reports/index.d.ts

Reporters have various options that solve different issues, e.g. using relative paths in monorepo setups: https://github.com/vitest-dev/vitest/issues/1902#issuecomment-1371912991.

### Suggested solution

Similar API as Jest has https://jestjs.io/docs/configuration#coveragereporters-arraystring--string-options.

Vitest's documentation could look like this:

> #### reporter
> 
> - **Type:** `string | string[] | [string, {}][]`
> - **Default:** `['text', 'html', 'clover', 'json']`
> - **Available for providers:** `'c8' | 'istanbul'`
> - **CLI:** `--coverage.reporter=<reporter>`, `--coverage.reporter=<reporter1> --coverage.reporter=<reporter2>`
> 
> Coverage reporters to use. See [istanbul documentation](https://istanbul.js.org/docs/advanced/alternative-reporters/) for detailed list of all reporters. See [`@types/istanbul-reporter`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/276d95e4304b3670eaf6e8e5a7ea9e265a14e338/types/istanbul-reports/index.d.ts) for details about reporter specific options.
> 
> The reporter has three different types:
> 
> - A single reporter: `{ reporter: 'html' }`
> - Multiple reporters without options: `{ reporter: ['html', 'json'] }`
> - A single or multiple reporters with reporter options:
>   
>   ```ts
>   {
>     reporter: [
>       ['lcov', { 'projectRoot': './src' }],
>       ['json', { 'file': 'coverage.json' }],
>       ['text']
>     ]
>   }
>   ```

### Additional context

- Requires unreleased `c8@7.13.0`.
- Requires refactoring Vitest's internal coverage typings first

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
[...]

PR #2690, opened 2023-01-16
feat(coverage): add support for coverage reporter options

AriPerkkio:
- Closes https://github.com/vitest-dev/vitest/issues/2658
- Requires `c8@7.13.0` https://github.com/bcoe/c8/pull/428 ✅ 

The reporter options API has strict typings, but IDE intellisense won't be able to narrow the prompted options. Typecheck works correctly though. https://github.com/microsoft/TypeScript/issues/43824


packages/vitest/package.json
```diff
@@ -108,6 +108,7 @@
   "dependencies": {
     "@types/chai": "^4.3.4",
     "@types/chai-subset": "^1.3.3",
+    "@types/istanbul-reports": "^3.0.1",
```
sheremet-va, line 111: We usually inline external package types (see jsdom). I personally don't know if we need to, we can type it as `any` for the time being if there are no better solutions. Or we can also augment types in coverage packages without installing it as the Vitest dependency.

AriPerkkio: I see, the rollup build already inlines/bundles external typings. I'll move this into `devDependencies` instead.

packages/coverage-c8/src/provider.ts
```diff
@@ -179,3 +187,24 @@ function resolveC8Options(options: CoverageC8Options, root: string): Options {
 
   return resolved
 }
+
+function resolveReporters(configReporters: NonNullable<CoverageC8Options['reporter']>): Options['reporter'] {
```
AriPerkkio, line 191: This is now duplicated in both providers. If it's moved to `vitest/src/utils/config-helpers` and imported by the coverage packages, that's a breaking change since coverage packages expect new import from Vitest. 

Should these be in `@vitest/utils` instead, and coverage packages should declare that one as `dependency`? The utils package would then need some type imports from `vitest` package... 🙃

AriPerkkio: #2886 introduces `@vitest/utils/coverage` which would be perfect place for this. We can move this function there later once it is merged.

sheremet-va (approved):

Issue #2651, opened 2023-01-12
HTML test reporter - relative path for assets


ben-reitz:
### Clear and concise description of the problem

As part of the CI/CD pipeline at our company, we upload the test report folder to a large repository of test reports where the static assets are automatically served on a standard path related to which version of the tests were run. For example:

`https://test-reports-url.com/repo_name/branch_name/build_number/<folder_of_report_assets>`

This approach needs the folder of static assets for the html report to be referenced from the `index.html` with a relative path (`assets/index-21883a1b.js`) rather than the current fixed path (`/__vitest__/assets/index-21883a1b.js`) so they can be served on a URL like the one shown above.

### Suggested solution

When generating the HTML test report, output the paths with the expectation that the HTML report could be served from any URL and don't mandate the fixed `root.com/__vitest__/report` URL format.

### Alternative

Provide a configuration option to choose how the report should be generated.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

mysteryven:
Since add a new option should be seriously considered, what about just set [`base`](https://vitejs.dev/config/shared-options.html#base) from `'/__vitest__'` to `'./'` in the generated HTML report ?

PR #2695, opened 2023-01-17
feat: support relative path in html report

poyoho:
fix: #2651

setting the `vue-router` base on `/__vitest__/` when is not the report mode.
setting the `vue-router` base on `./` when is not the report mode.

I test this in test/reporter and `vitest --ui` is ok.

### DIFF
--- a/packages/ui/client/global-setup.ts
+++ b/packages/ui/client/global-setup.ts
@@ -3,6 +3,7 @@
 import { createRouter as _createRouter, createWebHistory } from 'vue-router'
 import FloatingVue, { VTooltip } from 'floating-vue'
 import routes from 'virtual:generated-pages'
+import { isReport } from '~/composables/client'
 import 'd3-graph-controller/default.css'
 import 'splitpanes/dist/splitpanes.css'
 import '@unocss/reset/tailwind.css'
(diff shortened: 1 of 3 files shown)


sheremet-va (approved):

sheremet-va:
> I test this in test/reporter and vitest --ui is ok.

Reporter gives me empty page with errors:

Screenshot 2023-01-18 at 14 45 46

Please, write an actual test to see that it works. This PR will not be merged until then.

sheremet-va (changes requested):
See for the comment above ^

poyoho:
I forgot update the tips. I think you should build all the packages and run `npx vite preview --outDir html`.

And how can I add the test for running web? 🤔

sheremet-va:
> And how can I add the test for running web? 🤔

You can try running `vitest` command and check that everything is ok with playwright.

sheremet-va:
> I forgot update the tips. I think you should build all the packages and run npx vite preview --outDir html.

We also have documentation.

poyoho:
I add the e2e test on #2710

sheremet-va:
By the way, after I merged test PR I noticed that Windows tests are failing. Can you look into it?

poyoho:
ok.

poyoho:
image

I run the test in my windows virtual machine is ok. 🤦

sheremet-va (changes requested):
Since you added a base for testing UI, I think we need a new test here for the relative path.

sheremet-va:
> Since you added a base for testing UI, I think we need a new test here for the relative path.

Although since this is a default behaviour now and we don't give an option to use the previous one, it should be fine.

sheremet-va (approved):

sheremet-va:
@poyoho types are failing

sheremet-va (approved):

PR #2701, opened 2023-01-18
fix: ENOENT assets when htmldir exists

everett1992:
if htmlDir already exists the previous code would not create the assets dir, this can cause ENOENT errors whenever htmlDir exists, but assets does not - for instance when outputing junit and html to the same parent folder

```
vitest --reporter junit html --outputFile.html tests/ --outputFile.junit tests/junit.xml
```

I don't understand why the previous code was checking existsSync before mkdir -p. My best guess is performance, but it's not going to have a significant effect.


sheremet-va (approved):

sheremet-va:
CI is failing

PR #2704, opened 2023-01-18
docs(cli): suggest `--run` flag when using `vitest related` with `lint-staged`

mkermani144:
Without passing `--run` to `vitest related`, if it runs in non-ci environments, it causes the `lint-staged` command not to exit because of `vitest` watch mode. In the docs, suggest using the mentioned xargument to prevent probable issues.

### DIFF
--- a/docs/guide/cli.md
+++ b/docs/guide/cli.md
@@ -40,6 +40,15 @@ Useful to run with [`lint-staged`](https://github.com/okonet/lint-staged) or wit
 vitest related /src/index.ts /src/hello-world.js
 ```
 
+Please note that if you want to use it with `lint-staged`, you probably should also pass `--run` so that `lint-staged` command can exit normally.
+
+```js
+// .lintstagedrc.js
+export default {
+  '*.{js,ts}': 'vitest related --run',
+};
+```
+
 ## Options
 
 | Options       |               |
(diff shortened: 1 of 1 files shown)


docs/guide/cli.md
```diff
@@ -40,6 +40,15 @@ Useful to run with [`lint-staged`](https://github.com/okonet/lint-staged) or wit
 vitest related /src/index.ts /src/hello-world.js
 ```
 
+Please note that if you want to use it with `lint-staged`, you probably should also pass `--run` so that `lint-staged` command can exit normally.
```
dammy001, line 43: I'm not sure there's a need to document this because `vitest run` is already documented. @sheremet-va what do you think?

sheremet-va: Vitest detects CI automatically. I think #2705 should help, so there is no need to pass down `--run` manually.

mkermani144: @sheremet-va It's not just about ci. `lint-staged` may be used as a git hook, which normally doesn't run in a ci.

sheremet-va: I updated the wording. @dammy001 can you review it?

dammy001 (approved):

sheremet-va (approved):

PR #2705, opened 2023-01-18
feat: improve "isCI" check

sheremet-va:


antfu (approved):

antfu:
https://github.com/unjs/std-env this could also be an option just FYI

AriPerkkio:
> https://github.com/unjs/std-env this could also be an option just FYI

+1 for `std-env`. It seems to be able to [detect Stackblitz/webcontainers](https://github.com/unjs/std-env/blob/790972a039cd8c28978476b51846ac0086b08367/src/providers.ts#L74-L80) as well. I've been looking for a way to detect such environments, so that in `@vitest/coverage-c8` we could print info message indicating V8 coverage won't work. From time to time users report that `c8` doesn't work on Stackblitz.

I can do that check in separate PR if `std-env` is chosen.

PR #2710, opened 2023-01-18
test: ui e2e

poyoho:
feat e2e test for ui

ref: #2695


sheremet-va:
Doesn't seem like it's working:

Screenshot 2023-01-19 at 09 45 32

poyoho:
@sheremet-va it's working now.

test/ui/test/html-report.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+    root,
+    `npx vite preview --outDir html --strict-port --base /__vitest__/ --port ${port}`,
+    `http://localhost:${port}/__vitest__/`,
+  )
+
+  return exit
+})
+
+it('should load report', async () => {
+  expect((await (await page.$('#app'))?.innerHTML() || '').length).not.toBe(0)
```
sheremet-va, line 29: can we add a more elaborate check here? maybe `toMatch` and check for some HTML there

sheremet-va (changes requested):

PR #2714, opened 2023-01-19
feat: new `environmentMatchGlobs` option to auto infer env based on glob

antfu:
This could enable convetions based env setup and allow conexist of different environments esier.


packages/vitest/src/types/config.ts
```diff
@@ -138,6 +138,22 @@ export interface InlineConfig {
...
+   * @default []
+   * @example [
+   *   // all tests in tests/dom will run in jsdom
+   *   ['tests/dom/**', 'jsdom'],
+   *   // all tests in tests/ with .edge.test.ts will run in edge-runtime
+   *   ['**\/*.edge.test.ts', 'edge-runtime'],
+   *   // ...
+   * ]
+   */
+  environmentMatchGlobs?: [string, BuiltinEnvironment | string][]
```
sheremet-va, line 155: There is `VitestEnvironment` type to allow better TS hints

sheremet-va (changes requested):

sheremet-va:
@antfu I forgot, but can you add documentation for this feature?

PR #2718, opened 2023-01-19
Add `calls` to SpyInstance

erezarnon:


sheremet-va:
`calls` is available on `mocks`. This calls is part of tinyspy API, and it's not public.

kalvenschraut:
@sheremet-va  if it is not public then why is it referenced in the docs?
https://vitest.dev/api/mock.html#mock-calls

I am trying to use it to assert the function has been called at least x time. There doesn't seem to be an existing assertions for this and accessing calls directly provides greater flexibility. While I can do so, typescript complains.

sheremet-va:
> @sheremet-va if it is not public then why is it referenced in the docs? https://vitest.dev/api/mock.html#mock-calls
> 
> I am trying to use it to assert the function has been called at least x time. There doesn't seem to be an existing assertions for this and accessing calls directly provides greater flexibility. While I can do so, typescript complains.

spy.mock.calls is documented, not spy.calls

and there is an assertion for this:

expect(spy).toHaveBeenCalledTimes(5)

kalvenschraut:
Ah alright it being at spi.calls was causing me since my test passed using it, but typescript complained.

I thought mock meant the spy instance at first when looking at the documentation there.

That assertion is exact times I just wanted one saying greater than or equal basically, but the below works fine now the confusion was cleared up.

image

PR #2721, opened 2023-01-20
feat!: move test runner into a separate package

### LINKED ISSUE #1470: Custom test runner

sheremet-va:
This should make extending runners easier. This change also allows disabling auto inlining of the Vitest package (we only inline a single file that does "import").

Also browser package can now directly import runner.

Consequences of this PR:
- Userland can now import "vitest" directly in packages
- Snapshot testing is supported in the browser
- Tests are running ~20% faster, because we don't inline vitest anymore (I tested only test/core for this claim)
- Browser testing works on initial load, bot doesn't reload on restart
- Docs has "advanced" section with `startVitest` and new runner API
- @vitest/* packages are no longer bundled with Vitest
- Userland can define their own "test" function (like benchmark) that will be collected with `describe` and default `@vitest/runner`

Closes #1470


sheremet-va:
@AriPerkkio coverage tests are failing again, I am not sure what can cause this here, can you have a look in your free time, please? 👀 

@Aslemammad @userquin I broke browser package, and not sure how you envision it with this change. The idea right now is to use `@vitest/runner` directly. Do you think you can fix it in this PR?

AriPerkkio:
> coverage tests are failing again

Quick look shows that sourcemaps are a bit off.  Maybe the "magic number" given as offset to sourcemaps in c8 provider is now different? I can debug further in two days. 

It's good that there are now tests validating this, and failing for a real reason. 👍

sheremet-va:
> > coverage tests are failing again
> 
> 
> 
> Quick look shows that sourcemaps are a bit off.  Maybe the "magic number" given as offset to sourcemaps in c8 provider is now different? I can debug further in two days. 
> 
> 
> 
> It's good that there are now tests validating this, and failing for a real reason. 👍 

Yes! You a probably right, I removed mocker from context. Thank you 🙏

sheremet-va:
`--browser` is fixed. Also added possibility to use snapshot tests in browser 😄

sheremet-va:
@antfu this is ready for review. Don't be scared by the numbers, most of it is lockfile 😄

Aslemammad:
@sheremet-va That'd improve the experience of the browser, thank you so much Vladimir.

antfu (approved):
The pref improvement is sweet!

I am not able to review line by line 😅. I think this is a good move, and I trust your implementation 👍

docs/advanced/runner.md
```diff
@@ -0,0 +1,148 @@
...
+```
+
+When initiating this class, Vitest passes down Vitest config, - you should expose it as a `config` property.
+
+::: warning
+`importFile` method in your custom runner must be inlined in `deps.inline` config option, if you call Node `import` inside.
+:::
+
+::: tip
+Snapshot support and some other features depend on the runner. If you don't want to lose it, you can extend your runner from `VitestTestRunner` imported from `vitest/runners`. It also exposes `BenchmarkNodeRunner`, if you want to extend benchmark functionality.
```
snyamathi, line 93: Is there a more concrete example of this?

```js
import { VitestRunner } from '@vitest/runner';
         ^^^^^^^^^^^^
SyntaxError: The requested module '@vitest/runner' does not provide an export named 'VitestRunner'
```

I don't see VitestRunner exported in https://unpkg.com/@vitest/runner@0.28.4/dist/index.js

n3ur0n:
Hello @sheremet-va! What's the state of that PR? Is it possible to run a custom test runner, mentioned in #1470?

Issue #2716, opened 2023-01-19
TypeError: Cannot destructure property 'allTestsRun' of 'undefined' as it is undefined.


AngelAlexQC:
### Describe the bug

When i try setup vitest & test coverage with c8 i have this error:
![image](https://user-images.githubusercontent.com/9358510/213518409-fb345e23-44c9-40ea-86d1-9498d65aeb29.png)

This error its only with version 0.27.2, when i downgraded to 0.27.1 this works:
![image](https://user-images.githubusercontent.com/9358510/213518793-644dae75-30fa-4b92-bba4-3022fa334461.png)

### Reproduction

![image](https://user-images.githubusercontent.com/9358510/213518879-ec15fe13-c305-42c4-bac0-37871190146f.png)

### System Info

```shell
System:
    OS: Linux 5.15 Ubuntu 22.04.1 LTS 22.04.1 LTS (Jammy Jellyfish)
    CPU: (8) x64 Intel(R) Core(TM) i7-3770 CPU @ 3.40GHz
    Memory: 3.39 GB / 7.73 GB
    Container: Yes
    Shell: 5.8.1 - /usr/bin/zsh
  Binaries:
    Node: 18.12.1 - ~/.nvm/versions/node/v18.12.1/bin/node
    npm: 9.2.0 - ~/.nvm/versions/node/v18.12.1/bin/npm
  Browsers:
    Chrome: 108.0.5359.124
  npmPackages:
    @vitejs/plugin-react: 2.1.0 => 2.1.0 
    vitest: 0.25.8 => 0.25.8
```

### Used Package Manager

pnpm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

[...]

PR #2725, opened 2023-01-21
chore: bug reports to include `@vitest/*` packages versions

AriPerkkio:
Include `@vitest/*` packages in `envinfo`. Helps narrowing down issues like https://github.com/vitest-dev/vitest/issues/2716 where users have different versions of `vitest` and other `@vitest/*` packages running.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -26,7 +26,7 @@ body:
     id: system-info
     attributes:
       label: System Info
-      description: Output of `npx envinfo --system --npmPackages '{vitest,vite,@vitejs/*}' --binaries --browsers`
+      description: Output of `npx envinfo --system --npmPackages '{vitest,@vitest/*,vite,@vitejs/*}' --binaries --browsers`
       render: shell
       placeholder: System, Binaries, Browsers
     validations:


AriPerkkio:
The checklist could also include following:

```yaml
- label: Check that the versions of `vitest` and other `@vitest/*` packages are identical
  required: true
```

Any thoughts?

sheremet-va:
> The checklist could also include following:
> 
> ```yaml
> - label: Check that the versions of `vitest` and other `@vitest/*` packages are identical
>   required: true
> ```
> 
>  
> Any thoughts?

Maybe instead we should change `peerDependencies` from `*` to `workspace:*` (get's replaced with current version, when published)?

AriPerkkio:
Yes, `peerDependencies` is the correct way to solve these conflicting versions. Looks like the coverage packages are not included as optional peer dependencies in `vitest` package's `package.json` at all.

Should the `@vitest/*` packages also have `vitest` as **required** peer dependency? Currently they list `vitest` as regular production dependency instead.

sheremet-va:
> Should the @vitest/* packages also have vitest as required peer dependency? Currently they list vitest as regular production dependency instead.

I think only coverage packages depend on Vitest. Others can be used without Vitest at all (except for UI maybe).

antfu (approved):

AriPerkkio:
I think the peer dependencies should be set as following, using `0.27.2` version as example:

```json
{
  "name": "vitest",
  "version": "0.27.2",
  "peerDependencies": {
    "@vitest/coverage-c8": "0.27.2",
    "@vitest/coverage-istanbul": "0.27.2"
  },
  "peerDependenciesMeta": {
    "@vitest/coverage-c8": {
      "optional": true
    },
    "@vitest/coverage-istanbul": {
      "optional": true
    }
  }
}
```

```json
{
  "name": "@vitest/coverage-c8",  "//": "Same for @vitest/coverage-istanbul package",
  "version": "0.27.2",
  "dependencies": {
    "//": "No vitest here. This package should only be used in a project that has vitest as dependency. Do not bring it with this package."
  },
  "peerDependencies": {
    "vitest": "0.27.2"
  },
  "peerDependenciesMeta": {
    "vitest": {
      "optional": false, "//": "Default value so can be omitted"
    }
  }
}
```

Currently the `@vitest/coverage-x` packages declare `vitest` as production dependency. They should not. Currently it's possible to cause conflicts where `@vitest/coverage-x` packages bring in their own `vitest` instead of utilizing the peer dependency:

```bash
$ npm -D -E install vitest@0.26.0
$ npm i -D -E @vitest/coverage-istanbul@0.27.2
$ npm ls vitest
repro@1.0.0 /Users/x/y/repro
├─┬ @vitest/coverage-istanbul@0.27.2
│ └── vitest@0.27.2
└── vitest@0.26.0

# Check from file system as well:
$ grep version node_modules/vitest/package.json 
  "version": "0.26.0",
$ grep version node_modules/@vitest/coverage-istanbul/package.json
  "version": "0.27.2",
```

sheremet-va:
> I think the peer dependencies should be set as following, using 0.27.2 version as example:

I don't think versions should be hardcoded, `*` is enough. Also `vitest` is not an optional peer dependency for coverage providers.

AriPerkkio:
> Also `vitest` is not an optional peer dependency for coverage providers.

Yes exactly, this is what the `peerDependenciesMeta.vitest.optional: false` above means.

> I don't think versions should be hardcoded, `*` is enough.

Using `*` is fine if the packages are indeed guaranteed to work with mixed versions. I think I'll just need to be more careful in future and not introduce breaking changes between `vitest` and `@vitest/*` packages like below:

- `@vitest/coverage-*` packages require `vitest` version which exports `coverageConfigDefaults`. They'll break if `vitest` is not updated as well:
  - https://github.com/vitest-dev/vitest/commit/1503e15d0897389440352b58adc109fe068d80e0#diff-ae7036446bf57196f2cd8050393760894902442f7568db2938587837d0edb5a2R28
- `@vitest/coverage-*` packages expect `vitest` to pass `reportContext` as argument:
  - https://github.com/vitest-dev/vitest/commit/85096281aad680a0e108554738ac395ef8c80586#diff-e1a511166fbfd52ea3880685b345f4d24b827c3c8c9a8639f2eb51478a0f6f36R16
  - This was made backwards compatible afterwards in https://github.com/vitest-dev/vitest/commit/4cd1e5d7fb5824b69e59c874362dc078bab4bc50

If the `peerDependencies` was used to enforce versions these conflicts would come up when updating the packages.

PR #2747, opened 2023-01-25
fix(vite-node): don't cache modules with `timestamp: 0`

danielroe:
**Resolves**:

https://github.com/nuxt/nuxt/issues/15651
https://github.com/nuxt/nuxt/pull/18447

There may be a legitimate reason to use cache for `timestamp: 0`; if so, please feel free to resolve differently.

### DIFF
--- a/packages/vite-node/src/server.ts
+++ b/packages/vite-node/src/server.ts
@@ -130,7 +130,7 @@ export class ViteNodeServer {
     const module = this.server.moduleGraph.getModuleById(id)
     const timestamp = module ? module.lastHMRTimestamp : null
     const cache = this.fetchCache.get(filePath)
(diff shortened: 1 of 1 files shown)


sheremet-va (approved):

sheremet-va:
@antfu Oh I remember why I added this. It never returns `cache`, if the file is not modified. We need a better handling

sheremet-va:
@danielroe what is the actual problem here? `cache.timestamp >= timestamp` should be enough, isn't it? It means we processed the file before? Maybe the dependency chain doesn't update correctly on your side?

I think the previous solutions is the correct one.

danielroe:
Interesting. We are getting `timestamp: 0` for virtual files that have been invalidated (ie. we've manually invalidated them). I'd definitely not rule out an issue elsewhere down the line though.

PR #2749, opened 2023-01-25
feat: allow `config` option to be false

antfu:


sheremet-va (changes requested):
Please, don't forget to update documentation.

sheremet-va:
The fastest hand in the world 🔫

antfu:
We don't have a docs entry for this it seems

antfu:
Considering this is probably not a user-faced API, I think jsdocs would be enough for this.

sheremet-va:
> Considering this is probably not a user-faced API, I think jsdocs would be enough for this.

We mention config at the top of "Configuring Vitest" section. I think all available API should be documented, even for developers. (It can be mention that usually this is option is not needed)

PR #2755, opened 2023-01-26
docs(coverage): c8 ignore hints actually work

AriPerkkio:
When writing this part of the guide initially I thought c8's ignore hints would not work, as their regexp is very strict and does not allow additional content such as `@preseve` keywords.

The ignore hints don't work with `@vitest/coverage-istanbul` since esbuild strips the comments before code is instrumented. Istanbul uses ignore hints to check whether coverage counters should be placed in given code block.

With `@vitest/coverage-c8` the coverage filtering is done after the coverage report has been collected. The exclusion is done by comparing results to the actual sources where original comments are in place. `v8-to-istanbul` used by `c8` uses ignore hints to see whether uncovered lines should be marked as covered.

### DIFF
--- a/docs/guide/coverage.md
+++ b/docs/guide/coverage.md
@@ -124,4 +124,10 @@ Beware that these ignore hints may now be included in final production build as 
 if (condition) {
 ```
 
-Unfortunately this does not work for `c8` at the moment.
+For `c8` this does not cause any issues. You can use `c8 ignore` comments with Typescript as usual:
+
+<!-- eslint-skip -->
+```ts
+/* c8 ignore next 3 */
+if (condition) {
+```
--- a/test/coverage-test/coverage-report-tests/__snapshots__/c8.report.test.ts.snap
+++ b/test/coverage-test/coverage-report-tests/__snapshots__/c8.report.test.ts.snap
@@ -2515,8 +2515,8 @@ exports[`c8 json report 1`] = `
     "f": {
       "0": 1,
       "1": 1,
-      "10": 0,
-      "11": 0,
+      "10": 1,
+      "11": 1,
       "2": 2,
       "3": 2,
       "4": 0,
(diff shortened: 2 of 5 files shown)


AriPerkkio:
The c8 tests seem really unstable now on CI. Randomly failing on both `ubuntu-latest, 16` and `ubuntu-latest, 18`. I'm unable to reproduce this locally.

I'll do some more testing on CI and try to find root cause.

AriPerkkio:
After rebasing latest `main` this seems to be working fine. I think #2758 fixed the issue. Before #2758 the worker from earlier test run was interfering the next run, as coverage tests run two test runs in loop. This only happend on heavy CPU load. The #2758 doesn't still terminate workers early enough to stop writing unnecessary V8 reports but seems to help a lot.

With `0.28.3` I'm also seeing some really good performance and stability improvements. I've been testing the `test/coverage-test` C8 tests in 100x loop to reproduce these CI crashes locally.

```diff
async function runTests() {
+ for (const [round] of Array(100).fill(null).entries()) {
+   console.log(`Round #${1 + round}`)
    for (const threads of [true, false]) {
      for (const [directory, config] of configs) {
        await startVitest('test', [directory], {
```

100x rounds of `pnpm test:c8` with version

- `0.28.3`: 2m29.773s, does not crash with this PRs changes
- `0.28.2, 0.28.1, 0.28.0`: 
  - Without changes of this PR 3m34.741s
  - With this PR's changes crashes after 3-7 rounds
- `0.27.3`: 
  - Crashes even without the changes of this PR after 71 rounds. **Even that takes 3m50.894s.** So 100x rounds would be somewhere around 4m30s maybe?
  - With changes of this PR crashes after 3 rounds.

dammy001 (approved):

sheremet-va (approved):

PR #2772, opened 2023-01-30
feat!: use child_process when --no-threads is used

### LINKED ISSUE #740: Running multiple test files crashes when canvas is installed (Error: Module did not self-register: '.../canvas/build/Release/canvas.node'.)
### Describe the bug

I’ve started exprimenting with Vitest and got all tests to pass individually; however, running `npx vitest run .` (or any set of test files together like `npx vitest run src/*.test.js`) crashes with the following error:

```
 RUN  /home/phil/dev/packages/vue-accessible-color-picker

node:internal/process/promises:265
            triggerUncaughtException(err, true /* fromPromise */);
            ^

Error: Module did not self-register: '/home/phil/dev/packages/vue-accessible-color-picker/node_modules/canvas/build/Release/canvas.node'.
    at Object.Module._extensions..node (node:internal/modules/cjs/loader:1185:18)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (/home/phil/dev/packages/vue-accessible-color-picker/node_modules/canvas/lib/bindings.js:3:18)
    at Module._compile (node:internal/modules/cjs/loader:1103:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1155:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
```

I’ve removed the `node_modules` directory contents several times, re-installed the dependencies, and the issue still happens. I’m writing this only because that’s the only tangible workaround I found for this problem.

### Reproduction

```
git clone https://github.com/kleinfreund/vue-accessible-color-picker.git
cd vue-accessible-color-picker
npm install

npm test
# or
npx vitest run .
```

**Expected behavior**:

Test suite runs with all tests passing.

**Actual behavior**:

Vitest crashes and no tests are run.

### Workaround: turn off threads

In https://github.com/vitest-dev/vitest/issues/740#issuecomment-1042648373, a workaround was presented. Turning off [threads](https://vitest.dev/config/#threads) stops the issue from happening. Using the following configuration, the tests start working:

```js

import { defineConfig } from 'vite'
import Vue from '@vitejs/plugin-vue'

export default defineConfig({
  plugins: [
    Vue(),
  ],
  test: {
    threads: false,
    environment: 'jsdom',
  },
})
```

### System Info

```shell
System:
    OS: Linux 5.13 Ubuntu 20.04.3 LTS (Focal Fossa)
    Container: Yes
    Shell: 5.8 - /usr/bin/zsh
  Binaries:
    Node: 16.14.0 - /usr/bin/node
    Yarn: 1.22.5 - ~/.yarn/bin/yarn
    npm: 8.5.0 - ~/dev/packages/vue-accessible-color-picker/node_modules/.bin/npm
  npmPackages:
    @vitejs/plugin-vue: ^2.2.0 => 2.2.0 
    vite: ^2.8.1 => 2.8.1 
    vitest: ^0.3.4 => 0.3.4
```

### Used Package Manager

npm

### Validations

### LINKED ISSUE #1436: process.chdir() is not supported in workers (--no-threads doesn't work)
### Describe the bug

When I try to call `process.chdir` in a test, I get "TypeError: process.chdir() is not supported in workers". From #566 I understand the solution is passing `--no-threads` but it doesn't seem to work.

### Reproduction

https://stackblitz.com/edit/node-ve8ywd?file=package.json

### System Info

```shell
System:
  OS: Linux 5.4 Linux Mint 20.3 (Una)
  CPU: (8) x64 Intel(R) Core(TM) i7-3630QM CPU @ 2.40GHz
  Memory: 3.70 GB / 15.52 GB
  Container: Yes
  Shell: 5.0.17 - /bin/bash
Binaries:
  Node: 18.2.0 - /usr/local/bin/node
  Yarn: 1.22.11 - ~/.npm-global/bin/yarn
  npm: 8.5.1 - ~/.npm-global/bin/npm
Browsers:
  Chrome: 102.0.5005.61
  Firefox: 101.0
npmPackages:
  vitest: ^0.13.1 => 0.13.1
```

### Used Package Manager

pnpm

### Validations

### LINKED ISSUE #1544: Unable to bundle CDK lambdas when executed through vitest
### Describe the bug

Using vitest to test CDK stacks fails when the stack constructs a nodes lambda. When using `jest` the tests run without issue. I have no idea if the is a problem with CDK or vitest, or something else entirely.

```
Failed to bundle asset MyTestStack/my-lambda/Code/Stage, bundle output is located at /private/var/folders/wf/0n_3mjr91114bj87w744lyvm0000gn/T/cdk.outRqpmmh/bundling-temp-31e6d62c5dd7767fc416ae362fc44caa01c8c1f300c419c61d1f3663ad52c8ff-error: TypeError [ERR_INVALID_ARG_VALUE]: The argument 'stdio' is invalid. Received WritableWorkerStdio {
  _writableState: WritableState {
    objectMode: false,
    highWaterMark: 16384,
    finalCalled: false,...
 ❯ AssetStaging.bundle node_modules/aws-cdk-lib/core/lib/asset-staging.js:2:584
 ❯ AssetStaging.stageByBundling node_modules/aws-cdk-lib/core/lib/asset-staging.js:1:4201
 ❯ stageThisAsset node_modules/aws-cdk-lib/core/lib/asset-staging.js:1:1609
 ❯ Cache.obtain node_modules/aws-cdk-lib/core/lib/private/cache.js:1:242
 ❯ new AssetStaging node_modules/aws-cdk-lib/core/lib/asset-staging.js:1:2004
 ❯ new Asset node_modules/aws-cdk-lib/aws-s3-assets/lib/asset.js:1:736
 ❯ AssetCode.bind node_modules/aws-cdk-lib/aws-lambda/lib/code.js:1:4731
 ❯ new Function node_modules/aws-cdk-lib/aws-lambda/lib/function.js:1:2803
 ❯ new NodejsFunction node_modules/aws-cdk-lib/aws-lambda-nodejs/lib/function.js:1:1171
```

### Reproduction

I have created a repository that demonstrates the issue. For comparison, I have added one test that runs using vitest, and another test that runs using jest.

https://github.com/Poliziano/cdk-vitest-bug

### System Info

```shell
System:
  OS: macOS 12.4
  CPU: (10) arm64 Apple M1 Pro
  Memory: 261.09 MB / 16.00 GB
  Shell: 5.8.1 - /bin/zsh
Binaries:
  Node: 16.13.2 - ~/Library/pnpm/node
  npm: 8.1.2 - ~/Library/pnpm/npm
Browsers:
  Safari: 15.5
npmPackages:
  vitest: ^0.16.0 => 0.16.0
```

### Used Package Manager

npm

### Validations

sheremet-va:
Closes #1436
Closes #740
Closes #1544

This is a breaking change, because `--no-threads` no longer runs tests inside a worker (the name was misleading). After this PR is merged, disabling threads will force Vitest to use `child_process` API to run tests inside a different process. This might cause performance degradation, so there is a fallback to the previous behaviour.

To restore the previous behavior, use this configuration:

```ts
export default {
  test: {
    singleThread: true,
  }
}
```

Or run Vitest with `--single-thread` option:

```sh
vitest --single-thread
```

> **Note**
> Both `--single-thread` and `--no-threads` options group tests by environment and run only those tests in the same process/thread. Vitest can still run tests with several processes/threads if you use multiple environments.


sheremet-va:
@AriPerkkio can you help me identify why coverage tests are failing for this PR? 🤔

AriPerkkio:
This will also enable Vitest usage with `aws-cdk` testing: https://github.com/vitest-dev/vitest/issues/1544#issuecomment-1193938332

packages/vitest/src/node/pool.ts
```diff
@@ -39,51 +42,69 @@ export function createPool(ctx: Vitest): WorkerPool {
...
 
-  options.env = {
-    TEST: 'true',
-    VITEST: 'true',
-    NODE_ENV: ctx.config.mode || 'test',
-    VITEST_MODE: ctx.config.watch ? 'WATCH' : 'RUN',
-    ...process.env,
-    ...ctx.config.env,
-  }
+  options.env = env
```
AriPerkkio, line 107: The C8 provider's `onBeforeFilesRun` modifies main thread's `process.env` which is then copied passed to workers. Use following to get coverage working again.

```ts
  ctx.coverageProvider?.onBeforeFilesRun?.()

  options.env = {
    ...env,
    ...process.env
  }
```

This is quite hacky as there is such side effect, and I've been planning to make it more visible. You are currently seeing effects of this bug/feature: https://github.com/nodejs/node/issues/46378

sheremet-va: Hm, I see... We should probably rename the hook then 🤔"before field run" is not really what is happening here I think 🤔

purepear:
Just thinking out loud here :) Instead of introducing `singleThread`, would it make sense to reuse `threads` by making it `boolean | string`? 🤔  (I'm probably missing something)
```
--threads // the default behaviour, same as "--threads true"
--threads false // use child_process, maybe same as "--threads 0"?
--threads 1 (same as setting minThreads:1 and maxThreads: 1, and it can work with any number)
```

sheremet-va:
> Just thinking out loud here :) Instead of introducing `singleThread`, would it make sense to reuse `threads` by making it `boolean | string`? 🤔 (I'm probably missing something)
> 
> ```
> --threads // the default behaviour, same as "--threads true"
> --threads false // use child_process, maybe same as "--threads 0"?
> --threads 1 // same as setting minThreads:1 and maxThreads: 1, and it can work with any number "--threads 5"
> ```

It requires other options to be set, when initiating Tinypool, not just max/min. I also don’t like having ambiguous option, new defaults are more explicit and don’t conflict with our other options.

Issue #2442, opened 2022-12-06
vitest warns about obsolete snapshots for from skipped tests


dominikg:
### Describe the bug

vite-plugin-svelte uses `describe.runIf` to conditionally run some e2e tests only during `vite dev` or `vite build`.

One of these uses snapshots, and when the runIf condition isn't met, vitest logs that the associated snapshots are obsolete.

while true for the current run, they are still needed for others.

similar to #2174 

### Reproduction

code here: https://github.com/sveltejs/vite-plugin-svelte/blob/main/packages/e2e-tests/import-queries/__tests__/import-queries.spec.ts#L94

log output: https://github.com/sveltejs/vite-plugin-svelte/actions/runs/3628891097/jobs/6120463873#step:9:60

### System Info

```shell
linux, 
node18
vite: ^3.2.5 => 3.2.5 
vitest: ^0.25.4 => 0.25.4
```

### Used Package Manager

pnpm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

sheremet-va:
> Hey! Can I try to work on this one?

Yes

samkevin1:
I'm a bit confused, is the log the problem or are these snapshots not being used?

dominikg:
the snapshots are used conditionally, the warning is logged when the condition leads to skipped tests. but removing the snapshots would cause errors for the test that uses them with the negated condition.

dominikg:
the value of `isBuild` in the code linked in the og report is controlled via an env var: https://github.com/sveltejs/vite-plugin-svelte/blob/360171392990cd907c87b6ba16e17d7698fe9f8e/package.json#L11
and can be both true or false.

samkevin1:
[...]

PR #2782, opened 2023-01-31
fix: removing obsolete warning log for snapshots

samkevin1:
This removes the `obsolete` log for snapshots that are unchecked.

Related to #2442.

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -639,6 +639,24 @@ When running concurrent tests, Snapshots and Assertions must use `expect` from t
 You cannot use this syntax, when using Vitest as [type checker](/guide/testing-types).
 :::
 
+### describe.sequential
+
+- **Type:** `(name: string | Function, fn: TestFunction, options?: number | TestOptions) => void`
+
+  `describe.sequential` in a suite marks every test as sequential. This is useful if you want to run tests in sequential within `describe.concurrent` or with the `--sequence.concurrent` command option.
+
+  ```ts
+  describe.concurrent('suite', () => {
+    test('concurrent test 1', async () => { /* ... */ })
+    test('concurrent test 2', async () => { /* ... */ })
+
+    describe.sequential('', () => {
+      test('sequential test 1', async () => { /* ... */ })
+      test('sequential test 2', async () => { /* ... */ })
+    })
+  })
+  ```
+
 ### describe.shuffle
 
 - **Type:** `(name: string | Function, fn: TestFunction, options?: number | TestOptions) => void`
--- a/packages/runner/src/suite.ts
+++ b/packages/runner/src/suite.ts
@@ -53,7 +53,7 @@ export function createSuiteHooks() {
 }
 
 // implementations
-function createSuiteCollector(name: string, factory: SuiteFactory = () => { }, mode: RunMode, concurrent?: boolean, shuffle?: boolean, each?: boolean, suiteOptions?: TestOptions) {
+function createSuiteCollector(name: string, factory: SuiteFactory = () => { }, mode: RunMode, concurrent?: boolean, sequential?: boolean, shuffle?: boolean, each?: boolean, suiteOptions?: TestOptions) {
   const tasks: (Test | TaskCustom | Suite | SuiteCollector)[] = []
   const factoryQueue: (Test | Suite | SuiteCollector)[] = []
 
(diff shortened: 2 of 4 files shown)


sheremet-va (changes requested):
Vitest should still display obsolete snapshots, it should not display skipped ones as obsolete.

samkevin1:
Got it, will change

sheremet-va:
> Vitest should still display obsolete snapshots, it should not display skipped ones as obsolete.

I think I might've phrased this in a wrong way. We don't want Vitest to notify users about skipped snapshots because this information is irrelevant. If we skipped a test, we don't want to remove these snapshots and don't want to show them alongside REALLY obsolete ones. But if we detected actually obsolete snapshots, we want to display them.

So, there are a few cases:

1. Vitest run a test, and all snapshots exist
   - Don't show any warnings
 2. Vitest skipped a test
    - Ignore snapshots inside, do not give a warning that there are obsolete snapshots for this test
 3. Vitest run a test, and found obsolete snapshots
    - Show a warning about obsolete snapshots

samkevin1:
@sheremet-va can you rerun the failing test? Locally it doesn't break

packages/vitest/src/integrations/snapshot/manager.ts
```diff
@@ -70,7 +71,8 @@ export function addSnapshotResult(summary: SnapshotSummary, result: SnapshotResu
     })
   }
 
+  summary.skipped += result.skipped
```
sheremet-va, line 74: We don't care for skipped snapshots. We already show that a test was skipped, it should be enough. 

We shouldn't show a user anything about skipped tests.

packages/vitest/src/integrations/snapshot/client.ts
```diff
@@ -26,31 +26,32 @@ interface AssertOptions {
...
 export class SnapshotClient {
   test: Test | undefined
+  suite: Suite | undefined
   snapshotState: SnapshotState | undefined
   snapshotStateMap = new Map<string, SnapshotState>()
 
-  async setTest(test: Test) {
-    this.test = test
+  async setTaskBase(test: Test, suite?: Suite) {
+    suite ? this.suite = suite : this.test = test
```
sheremet-va, line 34: Why do we need a suite here? Snapshots are test specific, they should not rely on a suite.

packages/vitest/src/integrations/snapshot/port/state.ts
```diff
@@ -102,6 +104,15 @@ export default class SnapshotState {
     })
   }
 
+  markSnapshotsAsSkippedForTest(testName: string): void {
+    this._uncheckedKeys.forEach((uncheckedKey) => {
+      if (keyToTestName(uncheckedKey).includes(testName)) {
```
sheremet-va, line 109: I feel like the safer way would be to use a regular expression here.

packages/vitest/src/runtime/runners/test.ts
```diff
@@ -28,7 +28,20 @@ export class VitestTestRunner implements VitestRunner {
     await this.snapshotClient.saveCurrent()
   }
 
+  async onBeforeRunSuite(suite: Suite) {
```
sheremet-va, line 31: What is happening here?

packages/vitest/src/runtime/runners/test.ts
```diff
@@ -28,7 +28,20 @@ export class VitestTestRunner implements VitestRunner {
...
+      return
+    }
+
+    clearModuleMocks(this.config)
+    await this.snapshotClient.setTaskBase({} as Test, suite)
+  }
+
   onAfterRunSuite(suite: Suite) {
+    if (suite.mode === 'skip')
+      this.snapshotClient.skipSuiteSnapshots(suite)
```
sheremet-va, line 43: We should be able to mark skipped snapshots inside a test itself. 

```ts
getTasks(suite).forEach((test) => client.skipTestSnapshots(test))
```

packages/vitest/src/integrations/snapshot/client.ts
```diff
@@ -61,6 +62,11 @@ export class SnapshotClient {
     this.snapshotState?.markSnapshotsAsCheckedForTest(test.name)
```
sheremet-va, line 62: Just use this to mark all tests, don't reinvent the wheel

sheremet-va (changes requested):

PR #2792, opened 2023-02-02
fix: duplicate callbacks in vite-node HMR

jgoux:
When using vite-node HMR, the callbacks registered with:

```ts
if (import.meta.hot) {
  import.meta.hot.on("vite:beforeFullReload", () => {
    console.log("full reload");
    server.close();
  });
}
```

are duplicated/accumulated on each reload (logs below after 4 saves):

```bash
> vite-node --watch src/index.ts

listening on port 4000
full reload
listening on port 4000
full reload
full reload
listening on port 4000
full reload
full reload
full reload
listening on port 4000
full reload
full reload
full reload
full reload
listening on port 4000
```

This change ensures that if for the same event, two callbacks are identical, the old one is replaced with the new one.

We can't use reference identity check unfortunately as the identity of the callback change between reloads, hence the usage of `.toString()`.

### DIFF
--- a/packages/vite-node/src/hmr/hmr.ts
+++ b/packages/vite-node/src/hmr/hmr.ts
@@ -292,6 +292,9 @@ export function createHotContext(
     ): void {
       const addToMap = (map: Map<string, any[]>) => {
         const existing = map.get(event) || []
+        const existingCbIndex = existing.findIndex(_cb => _cb.toString() === cb.toString())
+        if (existingCbIndex > -1)
+          existing.splice(existingCbIndex, 1)
         existing.push(cb)
         map.set(event, existing)
       }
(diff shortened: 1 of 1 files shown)


sheremet-va (changes requested):
Maybe it's better to remove callbacks when a module is invalidated? This doesn't seem very stable.

jgoux:
> Maybe it's better to remove callbacks when a module is invalidated? This doesn't seem very stable.

Sure, whatever works best. I don't have much knowledge about the internals of vite/vite-node. How should I do that?

sheremet-va:
>  I don't have much knowledge about the internals of vite/vite-node. 

So do I. I assume it should be possible to detect what module added a listener and remove it if it was invalidated.

sheremet-va:
> So do I. I assume it should be possible to detect what module added a listener and remove it if it was invalidated.

I think we can also just safely clear all "vite:onBeforeReload" listeners, because it assumes that everything will be reloaded

sheremet-va (approved):

Issue #2554, opened 2022-12-23
Enable vite-node to run via a shebang


mrmckeb:
### Clear and concise description of the problem

Using vite-node with a shebang enables no-build workflows for utilities in monorepos, which can reduce complexity and improves DX.

Usage example:
```json
{
  "scripts": {
    "start": "my-utility run --fast"
  }
}
```

This is supported in `ts-node`:
https://typestrong.org/ts-node/docs/usage/#shebang

### Suggested solution

As I understand it, `vite-node` would need two changes when running scripts this way:
- The `--` separator shouldn't be required, and all args should be considered args of the calling script.
- The `tsconfig.json` for the script should be used.

### Additional context

It looks like `ts-node` achieved this by adding a --script-mode and a new bin script.
https://github.com/TypeStrong/ts-node/commit/1ad44bf82b7e33a4dab8fcb997c891636309e89a

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

jgoux:
I was able to have something working writing my shebang like this:

`#!/usr/bin/env vite-node . --`

But it's definitely a hack, and I'm not sure which `tsconfig.json` file it's using actually 🤔

Also, another issue I have is that it seems like scripts run this way are cached somewhere because if I edit the source and re-run the script it executes the old content. ⁉️ 

This would definitely push `vite-node` toward being a great alternative to `ts-node` or `tsx`!

PR #2793, opened 2023-02-02
feat: add --script option to vite-node

jgoux:
In order to fix https://github.com/vitest-dev/vitest/issues/2554, I added a `--script` option to `vite-node` so it can be used in shebang! 🎊 

The code is a bit weird because `cac` does not interpret options passed before positional arguments correctly (see [here](https://github.com/cacjs/cac/issues/25)) so I had to play with this and `process.argv` ordering to have it right.

Another limitation in this implementation is that passing other options in the shebang to `vite-node` won't work, as I expect exact indexes in order to figure out what are the arguments to forward.

It's not perfect but I think it's good enough to unblock people like me as a start. 👍

### DIFF
--- a/packages/vite-node/README.md
+++ b/packages/vite-node/README.md
@@ -72,7 +72,7 @@ $ ./file.ts hello
 argv: [ 'hello' ]
 ```
 
-Note that when using the `--script` option, Vite Node forwards every arguments and options to the script to execute, even the one supported by Vite Node itself.
+Note that when using the `--script` option, Vite Node forwards every argument and option to the script to execute, even the one supported by Vite Node itself.
 
 ## Programmatic Usage
 
--- a/packages/vite-node/README.md
+++ b/packages/vite-node/README.md
@@ -50,7 +50,7 @@ Note that for options supporting RegExps, strings passed to the CLI must start _
 
 ### Hashbang
 
-If you prefer to write scripts that doesn't need to be passed into tsx, you can declare it in the [hashbang](https://bash.cyberciti.biz/guide/Shebang).
+If you prefer to write scripts that doesn't need to be passed into Vite Node, you can declare it in the [hashbang](https://bash.cyberciti.biz/guide/Shebang).
 
 Simply add `#!/usr/bin/env vite-node --script` at the top of your file:
 
@@ -66,13 +66,13 @@ And make the file executable:
 chmod +x ./file.ts
 ```
 
-Now, you can run the file without passing it into vite-node:
+Now, you can run the file without passing it into Vite Node:
 ```sh
 $ ./file.ts hello
 argv: [ 'hello' ]
 ```
 
-Note that when using the `--script` option, `vite-node` forwards every arguments and options to the script to execute, even the one supported by `vite-node` itself.
+Note that when using the `--script` option, Vite Node forwards every arguments and options to the script to execute, even the one supported by Vite Node itself.
 
 ## Programmatic Usage
 
--- a/packages/vite-node/README.md
+++ b/packages/vite-node/README.md
@@ -48,6 +48,32 @@ npx vite-node --options.deps.inline="module-name" --options.deps.external="/modu
 
 Note that for options supporting RegExps, strings passed to the CLI must start _and_ end with a `/`;
 
+### Hashbang
+
+If you prefer to write scripts that doesn't need to be passed into tsx, you can declare it in the [hashbang](https://bash.cyberciti.biz/guide/Shebang).
+
+Simply add `#!/usr/bin/env vite-node --script` at the top of your file:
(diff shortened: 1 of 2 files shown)


packages/vite-node/src/cli.ts
```diff
@@ -17,31 +17,40 @@ cli
...
+  scriptMode?: string
   config?: string
   watch?: boolean
   options?: ViteNodeServerOptionsCLI
   '--'?: string[]
 }
 
 async function run(files: string[], options: CliOptions = {}) {
+  if (options.scriptMode)
+    files = [options.scriptMode]
```
sheremet-va, line 41: Script mode can only be used with one file? What if user calls it like this: `vite-node file --script-mode`?

sheremet-va (changes requested):

packages/vite-node/src/cli.ts
jgoux: If we want to support that I'm afraid we have to drop `cac` and use something better at parsing args.
Right now `--script-mode` would only be used in shebang as you can already do `vite-node file --` if you want to call it from a shell.

sheremet-va: We can just add a validation here ourselves

jgoux:
It seems like the bug in `cac` is when parsing boolean options containing a "-". If I rename the option "script" it's recognized as a boolean no matter if there is a positional argument before or after it, it simplifies the logic!

It now works in any combination! 🎊 

The rule is, if `--script` is received, we extract the first positional argument as the script to run, and we forward everything else to this script. So if people are making a script that could potentially receive `--watch`, it's not interpreted by `vite-node` anymore and is just forwarded.

The only exception being the `--script` argument which can't be forwarded for obvious reasons. 😅

sheremet-va:
> It seems like the bug in `cac` is when parsing boolean options containing a "-". If I rename the option "script" it's recognized as a boolean no matter if there is a positional argument before or after it, it simplifies the logic!
> 
> 
> 
> It now works in any combination! 🎊 
> 
> 
> 
> The rule is, if `--script` is received, we extract the first positional argument as the script to run, and we forward everything else to this script. So if people are making a script that could potentially receive `--watch`, it's not interpreted by `vite-node` anymore and is just forwarded.
> 
> 
> 
> The only exception being the `--script` argument which can't be forwarded for obvious reasons. 😅 

Can you add documentation to README with example and caveats?

sheremet-va (approved):
Thank you, LGTM 👍🏻

jgoux:
@sheremet-va What is missing in order to merge this PR? I'm still unsure why is the `CI / test (windows-latest, 18)` fails.

packages/vite-node/README.md
```diff
@@ -48,6 +48,32 @@ npx vite-node --options.deps.inline="module-name" --options.deps.external="/modu
 
 Note that for options supporting RegExps, strings passed to the CLI must start _and_ end with a `/`;
 
+### Hashbang
+
+If you prefer to write scripts that doesn't need to be passed into Vite Node, you can declare it in the [hashbang](https://bash.cyberciti.biz/guide/Shebang).
```
dammy001, line 53:
```suggestion
If you prefer to write scripts that don't need to be passed into Vite Node, you can declare it in the [hashbang](https://bash.cyberciti.biz/guide/Shebang).
```

packages/vite-node/README.md
```diff
@@ -48,6 +48,32 @@ npx vite-node --options.deps.inline="module-name" --options.deps.external="/modu
...
+chmod +x ./file.ts
+```
+
+Now, you can run the file without passing it into Vite Node:
+```sh
+$ ./file.ts hello
+argv: [ 'hello' ]
+```
+
+Note that when using the `--script` option, Vite Node forwards every arguments and options to the script to execute, even the one supported by Vite Node itself.
```
dammy001, line 75:
```suggestion
Note that when using the `--script` option, Vite Node forwards every argument and option to the script to execute, even the one supported by Vite Node itself.
```

packages/vite-node/src/cli.ts
```diff
@@ -16,33 +16,42 @@ cli
   .option('-r, --root <path>', 'Use specified root directory')
   .option('-c, --config <path>', 'Use specified config file')
   .option('-w, --watch', 'Restart on file changes, similar to "nodemon"')
+  .option('--script', 'Use vite-node as a script runner')
```
dammy001, line 19: Can we have shorthand here too? @sheremet-va what do you think?

jgoux: I avoided the shorthand because it would deny forwarding `-s` to the executed script which is very common. The more we can pass to the executed script the better?

PR #2786, opened 2023-02-01
feat!: coverage-c8 to use V8 profiler directly instead of `NODE_V8_COVERAGE`

AriPerkkio:
Breaking changes:
- `vitest` and `@vitest/coverage-c8` versions must match as the `CoverageProviderModule` introduces new APIs and c8-provider starts to use these. 
- Removes `CoverageProvider.onBeforeFilesRun`. Custom coverage providers should use  `CoverageProviderModule.startCoverage` instead.

Currently Vitest is using [`NODE_V8_COVERAGE`](https://github.com/nodejs/node/blob/main/doc/api/cli.md#node_v8_coveragedir) environment variable to generate V8 coverage reports for `@vitest/coverage-c8`. We've used this environment variable to trigger Node to start writing JSON reports to `<reports-dir>/tmp`. 

These reports contain coverage results of all run code, including Node's internal `node:<module>` and unnecessary third party `**/node_modules/**` packages. In addition the results contain [`source-map-cache` information](https://github.com/nodejs/node/blob/main/doc/api/cli.md#source-map-cache) which is completely irrelevant for Vitest. There is so much data on these JSON reports. In #2108 user reported **435 MB**  `coverage/tmp` directory on CI. 

Also using `NODE_V8_COVERAGE` with `worker_threads` has some weird features: https://github.com/nodejs/node/issues/46378. I think it's meant more for running a single script instead of to be utilized by test runners. 

This PR replaces `process.env.NODE_V8_COVERAGE` usage by controlling V8 Profiler directly using `node:inspect`. We don't enable unncessary `source-map-cache` and filter the ScriptCoverage results early so the amount of stored data is reduced a lot. Results are stored in memory instead of file system - exactly as `@vitest/coverage-istanbul` does. This reduces amount of file system handles. Since we now know what files are in the report, we can reduce the amount of Vite source map handling in [`provider.ts`](https://github.com/vitest-dev/vitest/blob/4fc492c61d6e01bc34bc807a6b57197a1914f823/packages/coverage-c8/src/provider.ts#L74-L76). 

This should help with flaky coverage test on CI since workers no more write to `<reports-dir>/tmp` when they exit.

Issue #2028, opened 2022-09-12
Unexpected Istanbul code coverage report


kylerush:
### Describe the bug

In my experience Vitest produces an unexpected code coverage report with Istanbul. In general, when I have only unit tests in my applicaiton, Vitest+ Istanbul outputs a perfect code coverage report. However, if I do anything more complex, the code coverage report is incorrect and unexplainable.

For example, I a have a [Fastify](https://www.fastify.io/) application. The moment I do testing with Fastify (i.e. not strict unit tests), the code coverage go haywire. I am comparing Vitest + Istanbul to [Node TAP](https://node-tap.org/) (NYC + Istanbul).

Coverage report with Node TAP:
```
-----------------|---------|----------|---------|---------|-------------------
File             | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
-----------------|---------|----------|---------|---------|-------------------
All files        |    92.3 |       75 |   88.88 |   91.66 |                   
 src             |     100 |      100 |     100 |     100 |                   
  app.ts         |     100 |      100 |     100 |     100 |                   
 src/lib         |    87.5 |       75 |     100 |    87.5 |                   
  foo-bar.ts     |     100 |      100 |     100 |     100 |                   
  hello-world.ts |      75 |       50 |     100 |      75 | 5                 
 src/routes      |     100 |      100 |     100 |     100 |                   
  root.ts        |     100 |      100 |     100 |     100 |                   
 src/routes/foo  |      80 |      100 |   66.66 |      75 |                   
  index.ts       |      80 |      100 |   66.66 |      75 | 6                 
-----------------|---------|----------|---------|---------|-------------------
```

Coverage report with Vitest + Istanbul:
```
-----------------|---------|----------|---------|---------|-------------------
File             | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
-----------------|---------|----------|---------|---------|-------------------
All files        |   13.63 |        0 |    12.5 |   13.63 |                   
 src             |   27.27 |      100 |      50 |   27.27 |                   
  app.ts         |     100 |      100 |     100 |     100 |                   
  server.ts      |       0 |      100 |       0 |       0 | 2-12              
 src/lib         |       0 |        0 |       0 |       0 |                   
  foo-bar.ts     |       0 |        0 |       0 |       0 | 2-5               
  hello-world.ts |       0 |        0 |       0 |       0 | 2-5               
[...]

PR #2795, opened 2023-02-02
support inline source map

oliverzy:
Some server side framework like egg.js relies on ts-node to handle TS files.

When using vitest to test such projects, I found the source map is completely missing, because the ts file is not loaded by vitest.

This PR fixes source map issue for such projects that rely on ts-node  runtime transpiler.

### DIFF
--- a/packages/coverage-c8/src/provider.ts
+++ b/packages/coverage-c8/src/provider.ts
@@ -135,13 +135,13 @@ export class C8CoverageProvider implements CoverageProvider {
     // that we add in packages/vite-node/src/client.ts:114 (vm.runInThisContext)
     // TODO: Include our transformations in sourcemaps
     const offset = 185
-
+    const originalGetSourceMap = report._getSourceMap;
     report._getSourceMap = (coverage: Profiler.ScriptCoverage) => {
       const path = _url.pathToFileURL(coverage.url.split('?')[0]).href
       const data = sourceMapMeta[path]
 
       if (!data)
-        return {}
+        return originalGetSourceMap.call(report, coverage)
 
       return {
         sourceMap: {
--- a/packages/vite-node/src/source-map.ts
+++ b/packages/vite-node/src/source-map.ts
@@ -42,6 +42,7 @@ export function installSourcemapsSupport(options: InstallSourceMapSupportOptions
   install({
     environment: 'node',
     handleUncaughtExceptions: false,
+    hookRequire: true,
     retrieveSourceMap(source) {
       const map = options.getSourceMap(source)
       if (map) {


sheremet-va:
Please, give example, and write a test for this use case.

packages/coverage-c8/src/provider.ts
```diff
@@ -135,13 +135,13 @@ export class C8CoverageProvider implements CoverageProvider {
...
     const offset = 185
-
+    const originalGetSourceMap = report._getSourceMap;
     report._getSourceMap = (coverage: Profiler.ScriptCoverage) => {
       const path = _url.pathToFileURL(coverage.url.split('?')[0]).href
       const data = sourceMapMeta[path]
 
       if (!data)
-        return {}
+        return originalGetSourceMap.call(report, coverage)
```
AriPerkkio, line 144: Is this needed? If the source maps are not found from `ctx.vitenode.fetchCache`, how would they be present on V8 coverage reports? All files present in V8 reports should have been loaded through vite-node. 

In #2786 the [`source-map-cache`](https://github.com/nodejs/node/blob/main/doc/api/cli.md#source-map-cache) will be disabled completely. C8's `_getSourceMap` won't find source maps in reports then.

oliverzy:
OK, I’ll extract a small example for this case. I’m using Vitest in a big Node project to replace Jest

packages/coverage-c8/src/provider.ts
oliverzy: The server side framework will load our code in the runtime. And the entry point is the server side framework, not our code. Therefore it is handled as external. 

And since ts-node is used as a require hook, our code actually is processed by ts-node, to-node will generate inline sourcemap and c8 can pick it up.

oliverzy:
@sheremet-va @AriPerkkio Please see https://github.com/oliverzy/eggjs-vitest for minimal example. I have put the patch in the repo for quick verification.

packages/coverage-c8/src/provider.ts
AriPerkkio: This sounds similar as Fastify which was discussed here: https://github.com/vitest-dev/vitest/issues/2028.

I'll do some testing with the reproduction case. Maybe with #2786 this would require `--enable-source-maps` or `process.setSourceMapsEnabled(true)` called manually.

oliverzy: Yes, it's similar. Egg.js uses `require` to load our code.

https://github.com/eggjs/egg-core/blob/5cc402347cde8c672a1a448eac7df7b670464df0/lib/utils/index.js#L27

AriPerkkio: With #2786 this will break. I'm not sure where the source maps could be loaded from.

oliverzy: As far as I can tell, vitest doesn't use a require hook, the require hook is synchronous, it may not possible to be handled by vitest.

Maybe we can expose this as an option that allow use to decide according to his project.

sheremet-va: > I'll do some testing with the reproduction case. Maybe with https://github.com/vitest-dev/vitest/pull/2786 this would require `--enable-source-maps` or process.setSourceMapsEnabled(true) called manually.

I don't think `--enable-source-maps` works inside VM, although since we don't process `require` it shouldn't matter because they are loaded by Node.js itself.

> As far as I can tell, vitest doesn't use a require hook

Yes, Vitest doesn't intercept `require` calls and delegates them to Node via `createRequire`. There are no plans to support `require`, Vitest's goal is to be ESM first, and supporting `require` will introduce a lot of complexity. There are better tools to use if you rely on CJS.

oliverzy: Well, the fix in this PR is to enable vitest to support just in time transpiler like ts-node/tsx. It has been supported by source-map-support and c8 that vitest depends on, therefore the implementation is quite straightforward.

This feature doesn’t require vitest to intercept require to work.

sheremet-va: Supporting this will imply that Vitest can be used with `require`, which is not true. Some features (like mocking) don't work and are not meant to work (we don't want to receive issues that cannot be resolved). I specifically left out `hookRequire`, because we don't support it and we don't need additional overhead. If we will ever support `require`, it should be done in a single action (PR).

I am fine with the coverage change though.

sheremet-va: I am also not against some kind of flag to enable it manually, so users can understand the consequences.

AriPerkkio: Now that #2786 is merged, there will be no `source-map-cache` for `c8` to use. 
https://github.com/bcoe/c8/blob/2a3d0c7347c77165962de8deca214b077c9166e2/lib/report.js#L183-L184

This PR's change has no effect. The source maps generated by `ts-node` need to fetched from somewhere.

oliverzy:
This PR is obosolete since 0.29.0 has changed the source map & v8 coverage implementation a lot.

PR #2817, opened 2023-02-06
fix(coverage): custom providers to work inside worker threads

### LINKED ISSUE #2797: `UserConfigExport['coverage']['provider']` with a object option did not work at all
### Describe the bug

When you defined setting as following

```
const CoverageModule = {
  getProvider: () => { throw 'called' }
}
export default defineConfig({
  test: {
    coverage: {
      provider: CoverageModule
    }
  }
})
```

It always say your provider is `c8`

Note: it is a bug in https://github.com/vitest-dev/vitest/blob/482b72fc5f02da87498a18e0ed844ecb6bc772ca/packages/vitest/src/utils/base.ts#L93-L98

Condition here do no check for `isMergeableObject(target[key])` .
With in this case, cause `deepMerge({ provider: 'c8' }, { provider: CoverageModule })` to result in `{ provider: 'c8' }`

Yet another note:

It would still crash even this fixed because it is unable to send this object to other worker.
This option should probably not be the module instance at first place.

圖片

Yet Yet another note:

It seems worker is used even I run it with `vitest run --no-threads --coverage --no-isolate`, this can not be right...

### Reproduction

Run `npx vitest --run --coverage` in https://stackblitz.com/edit/vitest-dev-vitest-x4h2tg?file=vite.config.ts

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    @vitest/ui: latest => 0.28.3 
    vite: latest => 4.1.1 
    vitest: latest => 0.28.3
```

### Used Package Manager

npm

### Validations

AriPerkkio:
- Fixes #2797
- [x] Fix typings
- [x] Update docs

Custom coverage providers cannot be Javascript objects anymore. The configuration of `test.coverage.provider` needs to be serializable so that it can be passed to `worker_threads`. Now custom providers are path to a file on file system.
 
❗ This introduces breaking change to `coverage.provider` API, but it seems that it has never really worked for custom providers.

### DIFF
--- a/packages/vitest/src/types/coverage.ts
+++ b/packages/vitest/src/types/coverage.ts
@@ -53,12 +53,13 @@ export type CoverageReporter =
   | 'text-summary'
   | 'text'
 
-type Provider = 'c8' | 'istanbul' | CoverageProviderModule | undefined
+type Provider = 'c8' | 'istanbul' | 'custom' | undefined
 
 export type CoverageOptions<T extends Provider = Provider> =
-  T extends CoverageProviderModule ? ({ provider: T } & BaseCoverageOptions) :
-    T extends 'istanbul' ? ({ provider: T } & CoverageIstanbulOptions) :
-        ({ provider?: T } & CoverageC8Options)
+  T extends 'istanbul' ? ({ provider: T } & CoverageIstanbulOptions) :
+    T extends 'c8' ? ({ provider: T } & CoverageC8Options) :
+      T extends 'custom' ? ({ provider: T } & CustomProviderOptions) :
+          ({ provider?: T } & (CoverageC8Options))
 
 /** Fields that have default values. Internally these will always be defined. */
 type FieldsWithDefaultValues =
@@ -232,4 +233,9 @@ export interface CoverageC8Options extends BaseCoverageOptions {
    * @default false
    */
   100?: boolean
+}
+
+export interface CustomProviderOptions extends Pick<BaseCoverageOptions, FieldsWithDefaultValues> {
+  /** Name of the module or path to a file to load the custom provider from */
+  customProviderModule: string
 }
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -617,7 +617,7 @@ Isolate environment for each test file. Does not work if you disable [`--threads
 
 ### coverage
 
-You can use [`c8`](https://github.com/bcoe/c8) or [`istanbul`](https://istanbul.js.org/) for coverage collection.
+You can use [`c8`](https://github.com/bcoe/c8), [`istanbul`](https://istanbul.js.org/)  or [a custom coverage solution](/guide/coverage#custom-coverage-provider) for coverage collection.
 
 You can provide coverage options to CLI with dot notation:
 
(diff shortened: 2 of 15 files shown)


packages/vitest/src/types/coverage.ts
```diff
@@ -53,12 +53,15 @@ export type CoverageReporter =
...
+type Provider = BuiltinProviders | CustomProvider | undefined
 
 export type CoverageOptions<T extends Provider = Provider> =
-  T extends CoverageProviderModule ? ({ provider: T } & BaseCoverageOptions) :
-    T extends 'istanbul' ? ({ provider: T } & CoverageIstanbulOptions) :
-        ({ provider?: T } & CoverageC8Options)
+  T extends 'istanbul' ? ({ provider: T } & CoverageIstanbulOptions) :
+    T extends 'c8' ? ({ provider: T } & CoverageC8Options) :
+      T extends CustomProvider ? ({ provider: T } & CustomProviderOptions) :
+          ({ provider?: T } & CoverageC8Options)
```
AriPerkkio, line 64: This part is getting really tricky. As the `provider` can be a known value of `c8` | `istanbul`,  it can also be just a dynamic `string`. The `Omit<string, BuiltinProviders>` doesn't work anymore. [TS playground](https://www.typescriptlang.org/play?#code/FAFwngDgpgBAQgVwJYBsRIHYAUBOB7ANyQBMocBnGAXhgHIBjADlpgB86lyQBDDAIwQpaoSLADCCLngC2uQiTLUYXHJgDmMAGQwA8tKQgAPCvUAaeMjSY5RUhQB8I6DBsKcSxKnTZ8tspXYJKVlfNzYYBAxSADNMKGJgJ3FCMm41KABJLl4BFB0IdDwMShoAb2UZTOz+QXzCjAAuGD48PBQoXhgAXySYMRScNKgxRjqkIpKYcvJKkbGippMMDR7wZyCQGVc7eeKladnJTeldpowEaT5FHuAoAA8IPBwQGDXkglT03fJDABUYe4gKBRSjbRQ0ME4ezUYAwGD-QHA4iUWicHg1IQwAD8MAAFOUIKE7E1-l0tH0BkMsujct8AJQwBqwuHwgF3IEguhMFg4-EwQnyYmssnafofQbpOYFcbFBlMlkshHspGUDZbImKXkEjU4EndclqkKCsj0xnMhUKvkCvw4LF6kUU8VDKX1ch0xLAegTF5onKCfoYWJqJpiz5Qb77ZnWtxNABEvoxsdMzJm0iqNNq0oWrxwCCgyeZAHpCzAAAIgcgAWnu0HoIGrOHw7krlfhAAtOMo23hBMQYINOLAyE2U7NRlnGjBY9xyMQk8AugBuRJe4ovJgBoMhylfCeTUpRnVxpjzuGp4bj+pxmdzgtw4tlivVh5QOsNpswFvtzvkbu9-vcIOAKNk8o5ptSfp5BOTQgLm+YLsuQA)

I think we'll need to use 100% known values for `provider`, and use additional option for custom providers where they can define the path for the module. 

E.g.

```ts
provider: 'custom',
path: 'custom-provider.ts',
```

sheremet-va (approved):

sheremet-va (approved):

PR #2819, opened 2023-02-06
perf: don't inline Vitest entry

sheremet-va:
This gives a small boost to performance, because we only inline a small portion of the codebase. (it was quite small since 0.27.0, but now it's even smaller!)


sheremet-va:
@AriPerkkio I cannot reproduce errors from CI on my machine. Do you think they are just flacky, or I introduced a change that might've broken it? :(

AriPerkkio:
The error seen on Windows CI also comes up in current `main`, so no worries with changes of this PR.

I'm quite confident that #2786 will stop these flaky coverages tests for once.

Issue #2519, opened 2022-12-16
Large diffs do not show


silverwind:
### Describe the bug

vitest does not render large object diffs. The config options `outputDiffLines: Infinity`, and `outputTruncateLength: Infinity` seem to have no effect, it shows this unhelpful message with no diff visible at all:

```
AssertionError: expected [ Array(500) ] to deeply equal [ Array(500) ]
```
### Reproduction

https://github.com/silverwind/vitest-diff

Clone, `npm install` and `npx vitest` yields this unhelpful output:

```
 RUN  v0.25.8 /Users/silverwind/git/vitest-diff

 ❯ test.js (1)
   × diff

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯

 FAIL  test.js > diff
AssertionError: expected [ Array(500) ] to deeply equal [ Array(500) ]
 ❯ test.js:6:13
      4|   const a = JSON.parse(readFileSync(new URL("a.json", import.meta.url), "utf8"));
      5|   const b = JSON.parse(readFileSync(new URL("b.json", import.meta.url), "utf8"));
      6|   expect(a).toEqual(b);
       |             ^
      7| });
      8|



[...]

PR #2828, opened 2023-02-07
feat!: use "concordance" package to display diff instead of using custom diff

### LINKED ISSUE #2764: Extremely slow for huge snapshots mismatch
### Describe the bug

I have a huge snapshot (38k lines JS Map), when I run `vitest` in either `dev` or `run` modes - it takes insane amount of time to get response, one might think that it just hangs.

### Reproduction

My test:
```
import { it, describe, expect } from "vitest";

describe("something", () => {
  it("works", () => {
    expect(something()).toMatchSnapshot();
  });
});
```

My snapshot:
```
// Vitest Snapshot v1

exports[`something > works 1`] = `
Map {
  "Cucumber Lebanese" => [
    {
      "amountPerUnit": 5,
      "dateTime": "2020-01-01T01:01:00.000Z",
      "item": {
        "amount": 1,
        "description": "Cucumber Lebanese",
        "onSpecial": false,
        "productSeries": "Cucumber Lebanese",
        "quantity": 1,
      },
    },
    // repeat 3 times
    }
  ],
  // repeat 1000 times
```

When I run tests, I get this:

```
 FAIL  util.spec.ts > something > works
AssertionError: Snapshot `something > works 1` mismatched
 ❯ util.spec.ts:6:42
      4| describe("something", () => {
      5|   it("works", () => {
      6|     expect(something()).toMatchSnapshot();
       |                                          ^
      7|   });
      8| });
```

and then the process keeps on running while using ~10-20% of CPU.

After a long time I think the process fails with non-zero code without showing diff, but I'm not 100% sure about that.

Might be related to https://github.com/vitest-dev/vitest/issues/2008

### System Info

```shell
System:
    OS: Windows 10 10.0.19044
    CPU: (20) x64 12th Gen Intel(R) Core(TM) i9-12900H
    Memory: 32.59 GB / 63.68 GB
  Binaries:
    Node: 19.3.0 - ~\AppData\Local\nvs\node\19.3.0\x64\node.EXE
    npm: 9.2.0 - ~\AppData\Local\nvs\node\19.3.0\x64\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (108.0.1462.76)
```

### Used Package Manager

npm

### Validations

sheremet-va:
"concordance" takes significantly less time to process and render diff, so it should fix #2764 and not affect #2519

This PR removes these options as they are no longer used:

- outputTruncateLength
- outputDiffLines
- outputDiffMaxLines
- outputDiffMaxSize

The new package can also display differences in strings, which is very nice to have:

Screenshot 2023-02-07 at 22 44 11


sheremet-va:
There is an issue that it prints the whole object, if they are different (https://github.com/concordancejs/concordance/issues/68). I wonder if we should fork it to make this change or process a string afterward (we already split it to count +/-). What do you think, @antfu?

antfu:
Yeah I think we can fork it if it's not pluggable. Probably they would be happy to merge it via a PR as well?

Maxim-Mazurok:
Thank you, I've tried it out and it does seem to solve my issue - the response from vitest is instant in case of failure.

The diff looks a bit weird tho:
image

There I changed "Danone" to "Comone" in the snapshot. At least it is pointing at the correct line and is quick, so that would work for me.

silverwind:
PR description has one error: `outputDiffLines` is still recognized and not removed.

sheremet-va:
> PR description has one error: `outputDiffLines` is still recognized and not removed.

You are correct. Fixed the description.

silverwind:
> > PR description has one error: `outputDiffLines` is still recognized and not removed.
> 
> You are correct. Fixed the description.

Actually incorrect, you removed the option in https://github.com/vitest-dev/vitest/pull/2766#discussion_r1244904120 a few days earlier. I missed it because it was not called out in the changelog and it was sneakily done in a commit named `update all non-major dependencies`.

Please don't do that in the future, it is absolutely hostile towards users to change an interface without calling it out.

PR #2835, opened 2023-02-08
feat(spy): implement mock.withImplementation API

### LINKED ISSUE #2812: Jest compatibility API: mockFn.withImplementation API

Jest added mockFn.withImplementation API in v29.1.0
* PR: https://github.com/facebook/jest/pull/13281
* Docs: https://jestjs.io/docs/mock-function-api#mockfnwithimplementationfn-callback

It accepts a function which should be temporarily used as the implementation of the mock while the callback is being executed.

### Suggested solution

Add this API to vitest, if maintainers think it would be helpful to it's users.

### Alternative

Document how `mockImplementation` or `mockImplementationOnce` can be used in these use cases

### Additional context

This issue was noticed while writing vitest-codemod tests for mock API
https://github.com/trivikr/vitest-codemod/pull/80

### Validations

obadakhalili:
Closes #2812.

This is my first contribution, so I might have missed a few test cases, or missed a better way to implement this. Hopefully, it won't be my last 😄.

**Note**: I don't think that this is the best implementation for the asynchronous version. It would be better if it was explicit that the stub method `.withImplementation` returns a promise when the passed callback is asynchronous, so the user knows that they have to use `async` with it.

I'm happy to make any further changes and change the docs as necessary.


packages/spy/src/index.ts
```diff
@@ -248,6 +249,24 @@ function enhanceSpy<TArgs extends any[], TReturns>(
...
+    stub.mockImplementation(fn)
+
+    if (cb.constructor.name === 'AsyncFunction') {
+      const asyncCb = cb as () => Promise<void>
+
+      asyncCb().then(() => {
+        stub.mockImplementation(originalImplementation)
+      })
+    }
+    else {
```
sheremet-va, line 263: You don't call cb

packages/spy/src/index.ts
```diff
@@ -248,6 +249,24 @@ function enhanceSpy<TArgs extends any[], TReturns>(
     return stub
   }
 
+  stub.withImplementation = (fn, cb) => {
+    const originalImplementation = implementation!
+    stub.mockImplementation(fn)
+
+    if (cb.constructor.name === 'AsyncFunction') {
```
sheremet-va, line 256: You can just check if cb returned a Promise

packages/spy/src/index.ts
```diff
@@ -46,6 +46,7 @@ export interface SpyInstance<TArgs extends any[] = any[], TReturns = any> {
   getMockImplementation(): ((...args: TArgs) => TReturns) | undefined
   mockImplementation(fn: ((...args: TArgs) => TReturns) | (() => Promise<TReturns>)): this
   mockImplementationOnce(fn: ((...args: TArgs) => TReturns) | (() => Promise<TReturns>)): this
+  withImplementation(fn: ((...args: TArgs) => TReturns), cb: () => void): this
```
sheremet-va, line 49: You can add overload for Promise cb

test/core/test/mocked.test.ts
```diff
@@ -169,3 +169,29 @@ describe('mocked function which fails on toReturnWith', () => {
 test('streams', () => {
   expect(exportedStream).toBeDefined()
 })
+
+describe('temporary mock implementation', () => {
+  test('temporary mock implementation works as expected', () => {
+    const mock = vi.fn(() => 1)
+
+    mock.withImplementation(() => 2, () => {
+      expect(mock()).toBe(2)
```
sheremet-va, line 178: if you have expect inside a callback, don't forget to call
`expect.assertions` beforehand

test/core/test/mocked.test.ts
```diff
@@ -169,3 +169,29 @@ describe('mocked function which fails on toReturnWith', () => {
 test('streams', () => {
   expect(exportedStream).toBeDefined()
 })
+
+describe('temporary mock implementation', () => {
```
sheremet-va, line 173: There are not enough tests. What if I used `mockImplementationOnce` before this function, for example? Or any other method for that matter

sheremet-va (changes requested):

mysteryven:
I think the `withImplementation` should take precedence over `implementationOnce`. I created an [issue](https://github.com/facebook/jest/issues/13887) to Jest, and they fixed it now, maybe we can also add it.

test/core/test/mocked.test.ts
```diff
@@ -169,3 +169,80 @@ describe('mocked function which fails on toReturnWith', () => {
...
+
+    expect(mock()).toBe(1)
+  })
+
+  test('temporary mock implementation can be async', async () => {
+    const mock = vi.fn(() => 1)
+
+    expect.assertions(3)
+
+    /* NOTE:
```
obadakhalili, line 220: The below caveat is annoying.

sheremet-va:
I changes types a bit, can you add documentation about this function with examples?

obadakhalili:
I added the docs. I have a few questions though:
1. This now renders `mockImplementationOnce` unnecessary, right?
2. Should we add a test case with `mock.restore/reset`? Maybe here it is enough to test that `getMockImplementation` returns the correct function reference when `withImplementation` is used, and to test those methods (`restore` & `reset`) independently. Or maybe a test with these methods isn't necessary here. What do you think?
3. One way to handle precedence over the `mockImplementationOnce` function is to clear the once implementations temporarily inside the function as opposed to using an extra variable `implementationChangedTemporarily`. Do you think we should do that?
4. The change you did prevents TS users to mock the implementation with a different return value type. What if they need to do this temporarily?
5. Finally, the change you did actually handled the TS caveat and made the `withImplementation` return a promise if the callback is asynchronous. Do you think you can tell me why your solution works but not mine?

A small note regarding your changes, I appreciate your help, though I would also like it to do it by myself so I get to benefit as much as possible. Thanks though 😄

sheremet-va:
> 1. This now renders `mockImplementationOnce` unnecessary, right?

How so? You can always chain them to return the value several times.

> 2. Should we add a test case with `mock.restore/reset`? Maybe here it is enough to test that `getMockImplementation` returns the correct function reference when `withImplementation` is used, and to test those methods (`restore` & `reset`) independently. Or maybe a test with these methods isn't necessary here. What do you think?

Sure, the more tests we have the better.

> 3. One way to handle precedence over the `mockImplementationOnce` function is to clear the once implementations temporarily inside the function as opposed to using an extra variable `implementationChangedTemporarily`. Do you think we should do that?

No, I think it's unnecessarily complex because we will need to restore them afterward.

> 4. The change you did prevents TS users to mock the implementation with a different return value type. What if they need to do this temporarily?

If a function defined its return type, it should always return that type. It doesn't make sense to return anything else there.

> 5. Finally, the change you did actually handled the TS caveat and made the `withImplementation` return a promise if the callback is asynchronous. Do you think you can tell me why your solution works but not mine?

Your order was wrong. `void` is subset of `Promise<void>`, so typescript always resolved it as `void`. If you change the order and define Promise overload first, it will work.

obadakhalili:
> How so? You can always chain them to return the value several times.

Yeah yeah, but I meant that `withImplementation` can do `mockImplementationOnce` work and more. So maybe `mockImplementationOnce` won't be as useful to users as before. Not suggesting to remove it. Just thought I would point it out.

obadakhalili:
Hey @sheremet-va, everything's fine with the PR? are you guys planning to merge it soon?

obadakhalili:
@sheremet-va I re-resolved the conflicts with the pnpm lock file. Is there a reason why you guys are taking long to merge the PR?

sheremet-va (approved):

PR #2836, opened 2023-02-08
feat!: vitest as peer dependency for coverage packages

AriPerkkio:
Breaking changes:
- `@vitest/coverage-c8` and `@vitest/coverage-istanbul` require `vitest` as peer dependency. End users may need to update their `vitest` when updating latest coverage packages.

Some related earlier discussion at #2725.

Currently the coverage packages declare `vitest` as depedency. If user installs `vitest@0.28.0` and `@vitest/coverage-c8@0.28.0`, they'll end up with two copies of `vitest` in `node_modules`, https://github.com/vitest-dev/vitest/pull/2725#issuecomment-1399838852. Note that now that Vitest is in version `0.x.y`, the semver follows some special rules. For example `0.21.0` does not match `^0.20.0`.

>
> ```sh
> $ npm i -D -E vitest@0.26.0
> $ npm i -D -E @vitest/coverage-istanbul@0.27.2
> $ npm ls vitest
> repro@1.0.0 /Users/x/y/repro
> ├─┬ @vitest/coverage-istanbul@0.27.2
> │ └── vitest@0.27.2
> └── vitest@0.26.0
> ```
>

Also when we intentionally introduce breaking changes to the packages, users may run into errors when mixing two incompatible versions. Using `peerDependencies` solves the issue by failing fast when package managers detect incompatibility issues.

### DIFF
--- a/packages/coverage-c8/package.json
+++ b/packages/coverage-c8/package.json
@@ -41,14 +41,17 @@
     "dev": "rollup -c --watch --watch.include 'src/**'",
     "prepublishOnly": "pnpm build"
   },
+  "peerDependencies": {
+    "vitest": ">=0.28.0 <1"
+  },
   "dependencies": {
     "c8": "^7.12.0",
     "picocolors": "^1.0.0",
-    "std-env": "^3.3.1",
-    "vitest": "workspace:*"
+    "std-env": "^3.3.1"
   },
   "devDependencies": {
     "pathe": "^1.1.0",
-    "vite-node": "workspace:*"
+    "vite-node": "workspace:*",
+    "vitest": "workspace:*"
   }
 }
--- a/packages/coverage-istanbul/package.json
+++ b/packages/coverage-istanbul/package.json
@@ -41,14 +41,16 @@
     "dev": "rollup -c --watch --watch.include 'src/**'",
     "prepublishOnly": "pnpm build"
   },
+  "peerDependencies": {
+    "vitest": ">=0.28.0 <1"
+  },
   "dependencies": {
     "istanbul-lib-coverage": "^3.2.0",
     "istanbul-lib-instrument": "^5.2.1",
     "istanbul-lib-report": "^3.0.0",
     "istanbul-lib-source-maps": "^4.0.1",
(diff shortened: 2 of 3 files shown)


packages/coverage-c8/package.json
```diff
@@ -41,14 +41,17 @@
     "dev": "rollup -c --watch --watch.include 'src/**'",
     "prepublishOnly": "pnpm build"
   },
+  "peerDependencies": {
+    "vitest": ">=0.28.0 <1"
```
AriPerkkio, line 45: This means that coverage packages set minimum requirement of current "breaking change" release, and upper limit for v1. This may still cause some incompatibility issues if API is broken from the `vitest` side.

On the other hand, if we use just `"^0.28.0"` we need to bump this version on every minor/breaking release. `0.29.0` does not match `"^0.28.0"`.

Any thoughts, which one should we end up using?

sheremet-va: I think >= is better than using a caret here, because it doesn’t match minor versions, if major is 0. When we release v1, we can update this to use a caret

packages/coverage-istanbul/package.json
```diff
@@ -41,21 +41,24 @@
     "dev": "rollup -c --watch --watch.include 'src/**'",
     "prepublishOnly": "pnpm build"
   },
+  "peerDependencies": {
+    "vitest": ">=0.28.0 <1"
```
dammy001, line 45: same here too @AriPerkkio

packages/coverage-c8/package.json
AriPerkkio: Yes, that sounds good. And now when ever we need to require newer version as peer dependency, we'll just manually bump the version here. For example, after merging #2786 this sound be changed to `">=0.29.0 <1"` before publishing new minor release. 
After v1 this will get easier.

packages/coverage-istanbul/package.json
AriPerkkio: We'll proceed with `>=`.

sheremet-va (approved):

Issue #2849, opened 2023-02-11
expect.toHaveReturned does not return true when a mock function is updated to return value


trivikr:
### Describe the bug

expect.toHaveReturned does not return true when a mock function is updated to return value

### Reproduction

Code:
```js
import { expect, test, vi } from "vitest";

test("toHaveReturned", () => {
  const mockError = new Error("mockError");
  const mockFn = vi.fn(() => { throw mockError; });

  try {
    mockFn();
  } catch (error) {}
  expect(mockFn).not.toHaveReturned();

  const mockReturn = 42;
  mockFn.mockReturnValue(mockReturn);
  const mockOutput = mockFn();
  expect(mockOutput).toBe(mockReturn);
  expect(mockFn).toHaveReturned(); // <-- This should be true.
});
```

Output:
```console
AssertionError: expected "spy" to be successfully called at least once
 ❯ src/__fixtures__/expect/toHaveReturned.output.js:15:18
     13|   const mockOutput = mockFn();
     14|   expect(mockOutput).toBe(mockReturn);
     15|   expect(mockFn).toHaveReturned();
       |                  ^
     16| });

  - Expected   "false"
  + Received   "true"
```

### System Info

```shell
System:
    OS: macOS 13.1
    CPU: (8) arm64 Apple M1 Pro
  Binaries:
    Node: 18.13.0 - ~/Library/Caches/fnm_multishells/43666_1676070339954/bin/node
    npm: 8.19.3 - ~/Library/Caches/fnm_multishells/43666_1676070339954/bin/npm
  npmPackages:
    vitest: ^0.28.4 => 0.28.4
```

### Used Package Manager

yarn

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
[...]

PR #2850, opened 2023-02-11
fix(expect): check for no 'throw' type in toHaveReturned

trivikr:
Fixes: https://github.com/vitest-dev/vitest/issues/2849


sheremet-va:
I don’t think this is correct? Shouldn’t it never threw?

trivikr:
It didn't throw at least once.

From docs:

> This assertion checks if a function has successfully returned a value at least once (i.e., did not throw an error).

https://vitest.dev/api/expect.html#tohavereturned

trivikr:
This issue was noticed while writing tests for vitest-codemod in https://github.com/trivikr/vitest-codemod/pull/102

sheremet-va (approved):

PR #2854, opened 2023-02-12
feat: add "strictESM" option to Node environment

### LINKED ISSUE #2198: Option to turn off strict mode

Currently, strict mode is forced on instead of being an option, making it not usable for certain types of codebases.

### Suggested solution

config toggle to turn off strict mode.

### Alternative

_No response_

### Additional context

_No response_

### Validations

### LINKED ISSUE #2841: Vitest incorrectly injects __dirname when type is module
### Describe the bug

I'm in the process of migrating a large codebase from CJS to ESM. One of the main "gotchas" is 
 that there is [No __filename or __dirname](https://nodejs.org/api/esm.html#no-__filename-or-__dirname).

I'm getting a lot of tests which incorrectly pass when they should actually fail because Vitest seems to inject the `__dirname` into the ES Module. This would be fine for CommonJS but is not correct for ESM.

Please see the minimal example enclosed where `npx vitest run` will pass, but `node index.js` will fail.

It actually looks like all of the [CJS Module Scope](https://nodejs.org/api/modules.html#the-module-scope) is available in ESM. This is not correct and will cause a lot of issues for developers who want to accurately test against an ESM environment.

Screenshot 2023-02-08 at 10 48 12 AM

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-ixfqyt?file=index.js

```
$ cat index.js
export default __dirname;

$ jq '.type' package.json 
"module"

$ npx vitest run
 RUN  v0.28.4 /home/projects/vitest-dev-vitest-ixfqyt

 ✓ index.test.js (1)

 Test Files  1 passed (1)
      Tests  1 passed (1)
   Start at  09:37:40
   Duration  4.72s (transform 136ms, setup 0ms, collect 21ms, tests 3ms)

$ node index.js
Error: __dirname is not defined
    at file:///home/projects/vitest-dev-vitest-ixfqyt/index.js
    at async ESMLoader.import (https://vitestdevvitestixfqyt-psy2.w-credentialless.staticblitz.com/blitz.2913a9f2fedb71678a44d5a6b5e0b08a6f00d17f.js:6:1209048)
    at async i.loadESM (https://vitestdevvitestixfqyt-psy2.w-credentialless.staticblitz.com/blitz.2913a9f2fedb71678a44d5a6b5e0b08a6f00d17f.js:6:246622)
    at async handleMainPromise (https://vitestdevvitestixfqyt-psy2.w-credentialless.staticblitz.com/blitz.2913a9f2fedb71678a44d5a6b5e0b08a6f00d17f.js:6:989055)
```

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    vitest: * => 0.28.4
```

### Used Package Manager

npm

### Validations

sheremet-va:
Closes #2841
Closes #2198

TODO:
- [ ] Tests
- [ ] Docs
- [ ] Examples
- [ ] Better error messages


sheremet-va:
@antfu do you approve of this idea as a whole? It bypasses Vite resolution, because it's unrestrictive, and I don't want to add this complexity to Vite itself. I think we should have some kind of a strict mode, because of this line in our documentation:

> Vitest aims to position itself as the Test Runner of choice for Vite projects, and as a solid alternative even for projects not using Vite.

Currently, Vitest is too permissive and will not fail, when it would've in Node.js, which is a problem for a tool that aims to eliminate such issues. Also, with our current approach, we can give useful error messages when someone uses a syntax when it's not supported (supported in this PR partially).

antfu:
While I think it make sense to have a strict mode, looking into the code, am I not sure if the complexity introduced to support it is worth it. I am a bit worried about if having this would create more problems as we will be in a middle ground that not 100% align with Node nor 100% align with Vite.

I don't have a strong opinion to push it back tho. We can still have it if you think it's worth it.

sheremet-va:
> I not sure if the complexity introduced to support it is worth it.

I also do not like introduced complexity. But I think it can be fixed with a refactor, this PR is a proof of concept currently.

> I am a bit worried about if having this would create more problems as we will be in a middle ground that not 100% align with Node nor 100% align with Vite.

Not sure about 100% with Vite statement here 🤔 Strict ESM should not be 100% aligned with Vite, because it has its own module resolution and transform pipeline.

About 100% with Node: there is a need to have our own module mechanism to allow mocking with `vi.mock`. Everything else (including resolution rules) should be aligned with Node. I personally think that it's better to use Node.js built-in test runner or runners like ava that just run your code, but neither of them support module mocking. And this is where Vitest comes in.

> We can still have it if you think it's worth it.

I think it's worth it, but it's a hard task. The main problem is that we don't use Vite plugin pipeline here at all, but we still have these options in config file 🤷🏻 Which will be confusing for people.

antfu:
To me, the important part of Vitest is it uses the Vite pipeline. If we are going not to use the Vite pipeline, I would honestly consider it out-of-scope of Vitest. I think we could have a mode to be strict on global variables like `__dirname` etc. But fully aligning with Node's behavior might belong to another test runner IMO.

sheremet-va:
> But fully aligning with Node's behavior might belong to another test runner IMO.

Then we should not advertise Vitest as a "solid alternative even for projects not using Vite.", and have a badge somewhere to recommend other test runners.

sheremet-va:
> To me, the important part of Vitest is it uses the Vite pipeline. If we are going not to use the Vite pipeline, I would honestly consider it out-of-scope of Vitest. I think we could have a mode to be strict on global variables like `__dirname` etc. But fully aligning with Node's behavior might belong to another test runner IMO.

We can also try to add configurations to Vite/improve existing ones so they align with how Node ESM works (need for extensions for example).

antfu:
To me, I use Vitest to test a lot of Node projects, and I am completely fine with how it works right now. I think the statement itself has no problem - it's true you can test non-Vite projects. I think every tool has its own behavior and own caveat, not strictly aligning with Node's behavior doesn't change it. Probably only the node's builtin test can represent the 100% node behavior. Being able to mock a module might be considered a misalignment of ESM spec if you go very strict.

sheremet-va:
> To me, I use Vitest to test a lot of Node projects, and I am completely fine with how it works right now. I think the statement itself has no problem - it's true you can test non-Vite projects. I think every tool has its own behavior and own caveat, not strictly aligning with Node's behavior doesn't change it.

You might be fine, but your code doesn't actually represent how it works for your users. The easiest example is misconfigured package.json. If your test Node.js code, it should at least fail where it suppose to, otherwise, there is no point in the tool.

>  Probably only the node's builtin test can represent the 100% node behavior.

We don't need to 100% follow it, but we should behave as closely to it as possible.

>  Being able to mock a module might be considered a misalignment of ESM spec if you go very strict.

I do not like module mocking, but people seem to enjoy it, even though it's very "magical" and sometimes hard to predict how it behaves. But it is an intentional deviation from the spec that people expect. When their package works in Vitest, but doesn't work for users - that's a problem. I for example would like to test Vitest's node entry point, but I cannot do that because Vitest will always work, but it won't work for actual users.

redonkulus:
@sheremet-va @antfu still think this will get in? any major blockers?

sheremet-va:
> any major blockers?

only time

sheremet-va:
Superseded by #3995

PR #2860, opened 2023-02-13
feat!: always run separate environments in isolation

sheremet-va:
In preparation for #2854

Tests are no longer running in the same worker if they don't share the same environment. This means that they don't share the same module cache and global scope, which should help with isolation. After PR is merged, tests in some projects will take longer to run, but the difference should be insignificant.

This also allows us to add "browser"/"node" conditions based on the environment in the future (we can safely pass `--conditions` to "worker" option).


antfu (approved):
Yeah, make sense 👌

PR #2867, opened 2023-02-14
feat: add a link to the comment at the top of the snapshot file

### LINKED ISSUE #2864: Add link to snapshot testing in file snapshots

This suggestion comes from the helpful link to snapshot testing which Jest adds.

Example code:
```js
test("toMatchSnapshot", () => {
  expect("foobar".toUpperCase()).toMatchSnapshot();
});
```

The snapshot on running jest
```console
// Jest Snapshot v1, https://goo.gl/fbAQLP

exports[`toMatchSnapshot 1`] = `"FOOBAR"`;
```

The link `https://goo.gl/fbAQLP` in the top level comment goes to https://jestjs.io/docs/snapshot-testing

No such link is present in Vitest snapshots
```console
// Vitest Snapshot v1

exports[`toMatchSnapshot 1`] = `"FOOBAR"`;
```

### Suggested solution

Add link to snapshot testing in the top-level comment of Vitest snapshots

```console
// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html

exports[`toMatchSnapshot 1`] = `"FOOBAR"`;
```

### Alternative

_No response_

### Additional context

_No response_

### Validations

btea:
close #2864

### DIFF
--- a/packages/vitest/src/integrations/snapshot/port/utils.ts
+++ b/packages/vitest/src/integrations/snapshot/port/utils.ts
@@ -20,7 +20,7 @@ import { getSerializers } from './plugins'
 
 export const SNAPSHOT_VERSION = '1'
 
-const writeSnapshotVersion = () => `// Vitest Snapshot v${SNAPSHOT_VERSION}`
+const writeSnapshotVersion = () => `// Vitest Snapshot v${SNAPSHOT_VERSION}, https://vitest.dev/guide/snapshot.html`
 
 export const testNameToKey = (testName: string, count: number): string =>
   `${testName} ${count}`


trivikr:
Should this be a breaking change?

The snapshot tests will not fail, when comments are updated in snapshot files.

sheremet-va:
> Should this be a breaking change?
> 
> The snapshot tests will not fail, when comments are updated in snapshot files.

Yes, but this will cause every snapshot to be updated.

dominikg:
imho this should not have been added during a regular test run but only when snapshots are being updated via cli flag

Right now this creates a dirty git state after tests have been run which can break tools that test for clean state and also accidentally add these changes to commits where they don't belong if a dev goes `git commit -am "added all files because i didn't expect test to modify files under version control"`

Issue #2870, opened 2023-02-14
vi.doMock does not override previous mock


trivikr:
### Describe the bug

~~vi.resetModules does not reset mocks created with doMock~~
vi.doMock does not override previous mock

### Reproduction

Example source file:
```js
const getToken = () => 'token';

module.exports = { getToken };
```

Example test file:
```js
// beforeEach(() => {
//  vi.resetModules();
// });

test('getToken 1', async () => {
  vi.doMock('./getToken', () => ({
    getToken: () => 'token1',
  }));
  const { getToken } = await import('./getToken');
  expect(getToken()).toBe('token1');
});

test('getToken 2', async () => {
  vi.doMock('./getToken', () => ({
    getToken: () => 'token2',
  }));
  const { getToken } = await import('./getToken');
  expect(getToken()).toBe('token2'); // <-- This should be 'token2'
});
```

Error thrown:
```console
 FAIL  doMock.spec.js > getToken 2
AssertionError: expected 'token1' to be 'token2' // Object.is equality
 ❯ doMock.spec.js:18:22
     16|   }));
     17|   const { getToken } = await import('./getToken');
     18|   expect(getToken()).toBe('token2');
       |                      ^
     19| });

  - Expected   "token2"
  + Received   "token1"
```

### System Info

```shell
System:
    OS: macOS 13.1
    CPU: (8) arm64 Apple M1 Pro
  Binaries:
    Node: 18.13.0 - ~/Library/Caches/fnm_multishells/65073_1676392486717/bin/node
    Yarn: 3.4.1 - ~/Library/Caches/fnm_multishells/65073_1676392486717/bin/yarn
    npm: 8.19.3 - ~/Library/Caches/fnm_multishells/65073_1676392486717/bin/npm
  npmPackages:
    vitest: ^0.28.5 => 0.28.5
```

### Used Package Manager

yarn

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
[...]

PR #2872, opened 2023-02-15
fix: cleanup last mocked cache when call vi.doMock

mysteryven:
fix: #2870 

[unmock-import.test.ts](https://github.com/vitest-dev/vitest/blob/609185bdc5027fba743ddbd5cb942faaed277cb8/test/core/test/unmock-import.test.ts#L29) has covered the changes for `vi.doUnmock`, so I only add tests for `vi.doMock`.

### DIFF
--- a/docs/api/vi.md
+++ b/docs/api/vi.md
@@ -221,13 +221,11 @@ increment(1) === 2
 let mockedIncrement = 100
 
 beforeEach(() => {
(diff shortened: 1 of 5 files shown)


mysteryven:
The CI is broken, I am trying to fix it, so convert it to draft.

mysteryven:
Maybe the error occurred in CI is not influenced by me.

sheremet-va (approved):

PR #2877, opened 2023-02-16
feat: add test seed to banner

### LINKED ISSUE #2876: seed not printed when using --sequence.shuffle
### Describe the bug

When using `--sequence.shuffle`, the default seed's value is not printed to stdout, making it impossible to reproduce a failing run by supplying a value to `--sequence.seed`

### Reproduction

Run vitest with `--sequence.shuffle`

### System Info

```shell
System:
    OS: macOS 13.2
    CPU: (10) arm64 Apple M1 Pro
    Memory: 103.78 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.13.0 - ~/.nvm/versions/node/v18.13.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v18.13.0/bin/yarn
    npm: 8.19.3 - ~/.nvm/versions/node/v18.13.0/bin/npm
  Browsers:
    Chrome: 110.0.5481.100
    Firefox: 110.0
    Safari: 16.3
  npmPackages:
    vitest: ^0.28.3 => 0.28.3
```

### Used Package Manager

pnpm

### Validations

btkostner:
This adds the sequence seed to the log banner (if it's set). My terminal color probably isn't the best, but this is what it looks like:

image

Closes #2876


packages/vitest/src/node/logger.ts
```diff
@@ -105,6 +105,9 @@ export class Logger {
 
     this.log(`${c.inverse(c.bold(mode))} ${versionTest} ${c.gray(this.ctx.config.root)}`)
 
+    if (this.ctx.config.sequence.seed)
```
sheremet-va, line 108: This value can only be set manually, so it doesn't really change anything:

https://github.com/vitest-dev/vitest/blob/9e03f2b8adadf72999c0a633c1960fa4495be0f1/packages/vitest/src/node/sequencers/RandomSequencer.ts#L8

There are also random seeds in `describe.shuffle`.

We should probably define seed somewhere if it's empty (even if we are not running in "shuffle" mode). Maybe here:

https://github.com/vitest-dev/vitest/blob/9e03f2b8adadf72999c0a633c1960fa4495be0f1/packages/vitest/src/node/core.ts#L135

This will also make `describe.shuffle` more deterministic. The problem now is - when to show the seed? If we only show the seed in shuffle mode, then we can't see it, when `describe.shuffle` is used somewhere. Maybe we should show the seed close to the shuffled `describe` in the test overview, if we are not in the shuffle mode? And if we are in the shuffle mode, let's show the seed, like you do here.

sheremet-va (changes requested):

packages/vitest/src/node/logger.ts
btkostner: We can make it more deterministic and set it in the `getSerializableConfig`, but I think the easiest option for logging is to just always print it in the header.

sheremet-va: This PR will not print the seed that is generated automatically. Which is what issue is asking to do.

btkostner: I must be missing something. The latest commit should have fixed that. From what I'm reading in the code:

```mermaid
graph TD
	Vitest[Vitest class gets created]
	Vitest --> setServer[setServer is ran that resolves config]
    setServer --> seed[Seed is set if it's not defined in the user given configuration]
    seed --> printBanner[printBanner is called to print the seed]
    printBanner --> RandomSequencer[RandomSequencer uses the configured seed to shuffle]
    printBanner --> runSuite[if suite.shuffle is set, runSuite will use the configured sequence seed to shuffle]
```

sheremet-va: Yes, sorry, I didn't saw the commit, I was answering from my phone.

Still, I don't think we should show the seed every time, "shuffle" mode is not used much and seeing a seed every time will only cause confusion. 

The other thing is the place you define it - it means the seed will not change in watch mode between reruns, which is not desirable. This is why I pointed you to `getSerializableConfig` - it's called every time there is a rerun.

sheremet-va: >  This is why I pointed you to getSerializableConfig - it's called every time there is a rerun. You should also just redefine it. when there is a rerun.

It seems that this was not a good idea in the end. I will revert the changes to declare seed when the config is resolved.

sheremet-va (approved):

PR #2886, opened 2023-02-19
feat(coverage): automatic threshold updating

AriPerkkio:
- Closes https://github.com/vitest-dev/vitest/issues/1241
- Adds new exported module to `vitest/coverage` entrypoint.

.
- [x] Manual testing
- [x] New test cases to `test/coverage-test` if possible
- [x] Documentation

PR #2889, opened 2023-02-20
fix: reload changed configuration file on watch mode

AriPerkkio:
- Fixes issues where configuration file changes were not reloaded in watch mode, https://github.com/vitest-dev/vitest/pull/2886#issuecomment-1436519446

The `options` here is mutated by multiple Vite hooks.

https://github.com/vitest-dev/vitest/blob/b67a5fbdf9940191573c5f90d94be7a2ecb26a36/packages/vitest/src/node/plugins/index.ts#L155-L161

Once it has been initialized by first server setup, the `deepMerge` will take previous values of `options` instead of using the fresh `viteConfig`. This PR will restore the `options` back to initial values when in watch mode.

Tested watch mode locally in linked project by changing `test.reporters` and `test.coverage.provider`. I don't think Vitest has any watch mode related tests set so just manual testing for now.

### DIFF
--- a/packages/vitest/src/node/plugins/index.ts
+++ b/packages/vitest/src/node/plugins/index.ts
@@ -14,6 +14,8 @@ import { CSSEnablerPlugin } from './cssEnabler'
 import { CoverageTransform } from './coverageTransform'
 
 export async function VitestPlugin(options: UserConfig = {}, ctx = new Vitest('test')): Promise<VitePlugin[]> {
+  const userConfig = deepMerge({}, options) as UserConfig
+
   const getRoot = () => ctx.config?.root || options.root || process.cwd()
 
   async function UIPlugin() {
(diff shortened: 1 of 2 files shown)


packages/vitest/src/node/plugins/index.ts
```diff
@@ -34,6 +36,11 @@ export async function VitestPlugin(options: UserConfig = {}, ctx = new Vitest('t
         this.meta.watchMode = false
       },
       config(viteConfig: any) {
+        if (options.watch) {
```
sheremet-va, line 39: Why `options.watch`? watch mode can be configured in the config

sheremet-va (changes requested):
Also I don't really understand what is happening here, can you add a comment about why a reset is needed?

AriPerkkio:
> Also I don't really understand what is happening here

The `VitestPlugin` is created only once. Its hooks are called first when server is initially set up, and again always when watch mode re-triggers it. As the hooks are modifying the given `options`, the `deepMerge` on the shown lines will behave differently when watch mode re-triggers. 

https://github.com/vitest-dev/vitest/blob/b67a5fbdf9940191573c5f90d94be7a2ecb26a36/packages/vitest/src/node/plugins/index.ts#L155-L161

Here on the first run the `options` is the `CliOptions`. On second run it is the complete merged configuration of the first run, as `VitestPlugin` is not re-created. As `options` is the last argument, it has priority over the `viteConfigTest`. So the fresh values from `viteConfigTest` are overwritten by earlier run's `options`. 

To be honest it's quite difficult to follow the flow here as all hooks are sharing the mutable `options` and modifying it. Does the explanation make any sense? 😅

sheremet-va:
Conflict 👀

sheremet-va (approved):

PR #2893, opened 2023-02-21
feat(vite-node): cli option for vite mode

abarke:
Add CLI support for setting the vite mode e.g.

```bash
vite-node -m bar foo.ts
```

Tested using:

```bash
echo "console.log('mode', import.meta.env.MODE)" > foo.ts && node dist/cli.mjs -m bar foo.ts && rm foo.ts
echo "console.log('mode', import.meta.env.MODE)" > foo.ts && node dist/cli.mjs --mode bar foo.ts && rm foo.ts
```

Side-catch: I couldn't commit without resolving the eslint errors so I fixed those too:

```bash
121:38  error  Unexpected mix of '&&' and '?:'. Use parentheses to clarify the intended order of operations  no-mixed-operators
122:13  error  Unexpected mix of '&&' and '?:'. Use parentheses to clarify the intended order of operations  no-mixed-operators
127:36  error  Unexpected mix of '&&' and '?:'. Use parentheses to clarify the intended order of operations  no-mixed-operators
128:11  error  Unexpected mix of '&&' and '?:'. Use parentheses to clarify the intended order of operations  no-mixed-operators
```

### DIFF
--- a/packages/vite-node/src/cli.ts
+++ b/packages/vite-node/src/cli.ts
@@ -15,6 +15,7 @@ cli
   .version(version)
   .option('-r, --root <path>', 'Use specified root directory')
   .option('-c, --config <path>', 'Use specified config file')
+  .option('-m, --mode <mode>', 'Set env mode')
   .option('-w, --watch', 'Restart on file changes, similar to "nodemon"')
   .option('--script', 'Use vite-node as a script runner')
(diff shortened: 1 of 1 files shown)


sheremet-va (approved):

abarke:
Thanks @sheremet-va 🚀

PR #2894, opened 2023-02-21
docs: add WebdriverIO to framework comparison section

christian-bromann:
I know it is not the job for Vitest docs to create an exhaustive list of framework competitors, however I think adding WebdriverIO as an alternative for Cypress makes sense as it promotes the use of web standards and gives an alternative that is not driven by a cooperation.


sheremet-va (approved):

sheremet-va:
Thank you for this comprehensive comparison 🙏🏻 LGMT 👍🏻

PR #2905, opened 2023-02-23
feat: use custom source-map-support implementation

sheremet-va:
Replacing `source-map-support` package with our custom implementation. Also replacing `source-map` package with `@jridgewell/trace-mapping`. This should reduce the size of the package considerably and make `new Error().stack` a bit faster.


sheremet-va:
We might also publish this as a separate package because it will be moved to Vite at some point for source map support.

antfu (approved):

Issue #579, opened 2022-01-19
Vitest runs tests 3x slower than Jest with threads: true (default) setting


dmLinker:
### Describe the bug

With the latest Vitest/Vite dependencies, Vitest is ~3 times slower than Jest in https://github.com/EvHaus/jest-vs-jasmine (more or less like a typical front-end repo) with the  `threads: true` (default) setting.

```
Benchmark 2: yarn workspace jest test
  Time (mean ± σ):     34.704 s ±  2.548 s    [User: 91.246 s, System: 16.712 s]
  Range (min … max):   32.017 s … 41.447 s    10 runs
 
Benchmark 3: yarn workspace vitest test --threads=true
  Time (mean ± σ):     99.077 s ±  2.015 s    [User: 185.493 s, System: 40.303 s]
  Range (min … max):   97.352 s … 103.406 s    10 runs
```

Related comment https://github.com/vitest-dev/vitest/issues/229#issuecomment-1003235680. That issue was closed but `threads: true` is still 3-4 slower than Jest after the issue was closed and the repro dependencies updated).

The issue was raised in [Discord chat as well](https://discord.com/channels/917386801235247114/926890821364621342/932577669751648266).

With `threads: false` Vitest is ~2 times faster, but the cost of it - there's no real isolation. With `threads: false` it's unfortunately not even close what Jest is doing in terms or resetting state, so not even sure if makes sense to compare `threads: false` with Jest (maybe threads: false should be compared with `uvu` or other "run-once-and-forget" runners with no real isolation). In a variety of projects that we have worked on, even with pure stateless components, the proper isolation is a super important concern. Especially in watch mode where introducing some unwanted state/mocking or polluting global/process state is simply an expected thing to happen due to the nature of various incremental changes (that are not always good or final) that simply break not properly isolated watch mode. 

### Reproduction

```
git clone https://github.com/EvHaus/jest-vs-jasmine
```

```
yarn
```

(if required, [install `hyperfine`](https://github.com/sharkdp/hyperfine#installation))

```
hyperfine --warmup 1 'yarn workspace jest test' 'yarn workspace vitest test --threads=true'
```

### System Info

```shell
System:
    OS: macOS 11.2.3
    CPU: (16) x64 Intel(R) Xeon(R) W-2140B CPU @ 3.20GHz
    Memory: 25.41 MB / 32.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
[...]

PR #2912, opened 2023-02-24
feat: add an option to enable Vite optimizer

sheremet-va:
This PR adds `deps.experimentalOptimizer` option to enable Vite optimizer. It bundles dependencies which makes them faster to import.

Benchmarks on M1 Macbook Air from #579:
Before:
Screenshot 2023-02-24 at 10 38 39
After (notice "collect"):
> First run:
Screenshot 2023-02-24 at 14 19 49

> Second run:
Screenshot 2023-02-24 at 10 39 05

There are a few caveats though. The first run will be slower than subsequent runs because we need to wait until dependencies are bundled (it's still faster than without the optimizer). So it's strongly recommended to have a shared cache for `node_modules/.vitest` folder (or the one you specify in `config.cacheDir`) in your CI environment. This optimization works only in `jsdom`/`happy-dom` environments. `node` and `edge` environments use Vite's SSR optimizer.

### DIFF
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -91,6 +91,28 @@ Files to exclude from the test run, using glob pattern.
 
 Handling for dependencies resolution.
 
+#### deps.experimentalOptimizer
+
+- **Type:** `DepOptimizationConfig & { enabled: boolean }`
+- **Version:** Vitets 0.29.0
+- **See also:** [Dep Optimization Options](https://vitejs.dev/config/dep-optimization-options.html)
+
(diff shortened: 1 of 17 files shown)


sheremet-va:
What is fascinating here is that most of the time in "collect" phase is just importing "react" 🤔 If I exclude react from optimization, it takes 10 seconds more in the "collect" section which results in 2 more seconds overall

antfu (approved):
🔥 LGTM! We could roll it out and see how it benefits real-world projects!

Issue #2897, opened 2023-02-22
UI never connects when using run


Weetbix:
### Describe the bug

The never gets past the "connecting" stage when running with `vitest run`, but works fine with `vitest watch`.

Not sure if this is so much a bug, or just unsupported. But I cant see why it shouldnt work, and for larger test suites that can take minutes to run it makes sense to provide the UI.

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-shyza9?file=package.json

### System Info

```shell
System:
    OS: macOS 13.2.1
    CPU: (8) arm64 Apple M1
    Memory: 138.16 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.13.2 - ~/.nvm/versions/node/v16.13.2/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v16.13.2/bin/yarn
    npm: 8.1.2 - ~/.nvm/versions/node/v16.13.2/bin/npm
  Browsers:
    Firefox: 97.0.1
    Safari: 16.3
  npmPackages:
    @vitejs/plugin-vue: 4.0.0 => 4.0.0
    @vitest/ui: ^0.28.5 => 0.28.5
    vite: ^4.1.4 => 4.1.4
    vitest: ^0.28.5 => 0.28.5
```

### Used Package Manager

yarn

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

sheremet-va:
It's not supported for "run" mode. You can use html reporter instead to get generated HTML page with the same results.

Weetbix:
Hey :) Maybe this should be a feature request then?

[...]

PR #2915, opened 2023-02-24
chore: not inject UIPlugin if disable watch mode

mysteryven:
close: #2897

I run the [reproduction](https://stackblitz.com/edit/vitest-dev-vitest-shyza9?file=package.json) in my local, and found it maybe a little annoying about opening a discounted window when run `vitest run --ui`, so I disable the `UIPlugin` injection in run mode.

I am sorry for I don't really know add which kind of test suits for this.

### DIFF
--- a/packages/vitest/src/node/plugins/index.ts
+++ b/packages/vitest/src/node/plugins/index.ts
@@ -17,6 +17,7 @@ export async function VitestPlugin(options: UserConfig = {}, ctx = new Vitest('t
   const getRoot = () => ctx.config?.root || options.root || process.cwd()
 
   async function UIPlugin() {
+    console.log(options.watch)
     await ensurePackageInstalled('@vitest/ui', getRoot())
     return (await import('@vitest/ui')).default(options.uiBase)
   }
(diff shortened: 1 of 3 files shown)


packages/vitest/src/node/plugins/index.ts
```diff
@@ -207,7 +207,7 @@ export async function VitestPlugin(options: UserConfig = {}, ctx = new Vitest('t
       : []),
     ...CSSEnablerPlugin(ctx),
     CoverageTransform(ctx),
-    options.ui
+    (options.ui && options.watch !== false)
```
sheremet-va, line 210: `watch` can be changed in the config file. These options are from the CLI

sheremet-va (changes requested):
#2897 expects that `html` reporter will be used when `--ui` flag is used in `--run` mode

Weetbix:
> https://github.com/vitest-dev/vitest/issues/2897 expects that html reporter will be used when --ui flag is used in --run mode

I was expecting that when `--ui` flag is used with `--run`, the ui server starts up and shows the test results as they are being ran. After the run the page would become static (as vitest and the ui server shuts down). My test suite is large and takes about 2 minutes to run, but I still want to see the results and any errors "live" as they are running.

This PR is still an improvement though 👍 , looks like it removes the UI server and instead generates a `html` report at the end of the test run. I think this means less confusion, but for me I will not use this report at the end of the test run.

mysteryven:
> the ui server starts up and shows the test results as they are being ran.

Run `npx vite preview --base __vitest__ --outDir ./html` manually can preview it as a static server in run mode, but I think you know this.

PR #2946, opened 2023-03-01
docs: update coverage file to specify file

ryangrunest:


sheremet-va:
What is the error here? `vite.config.ts` is a valid config name

PR #2963, opened 2023-03-05
chore: use node: protocol when importing built-in modules

code-ManL:
`import fs from 'fs'`

->

`import fs from 'node:fs'`


danez:
Vitest does support node 14.16.0 which does not support the `node:` protocol. I'm not sure how vitest is bundled so maybe this is handled when bundling.

sheremet-va:
We have a eslint rule that does this now.

Issue #1867, opened 2022-08-17
Cannot Debug Test Files with `--inspect-brk` Flag


alexlafroscia:
### Describe the bug

With the new `--inspect` and `--inspect-brk` flags, I expected to have a way for Vitest to debug test files without a tool like VSCode or IntelliJ. Unfortunately, that was not the case!

With a `debugger` in my test file, the following command does _not_ pause on the `debugger` statement

```
npx vitest --inspect-brk --threads false run whatever-filter
```

A `debugger` in my config file is paused on just fine, but ones in tests (which are more likely what folks want to debug) are not respected.

The [commit that introduces the flags](https://github.com/vitest-dev/vitest/commit/ea80f2bf) doesn't specifically _say_ that the flag does or doesn't allow for debugging tests, so I suppose this expectation is an assumption on my part, but I believe that this flag _not_ allowing child processes to be debugged would surprise users.

Adding the `--threads false` configuration was an attempt to avoid spawning child processes, like Jest's `--runInBand`, but that doesn't resolve the issue.

The only way I have found to debug a test file is with a tool that automatically allows child processes to be debugged, such as `ndb` or the VSCode debugging tools (through their "launch" config or integrated JavaScript debugging terminal). While this is fine for some users (myself included!) I am not sure how to advise team members that do not use these tools on how to debug test files.

### Reproduction

This repo contains a reproduction environment, as well as instructions

https://github.com/alexlafroscia/__vitest-inspect-brk

### System Info

```shell
System:
    OS: macOS 12.5
    CPU: (16) x64 Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
    Memory: 29.96 GB / 64.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.13.2 - ~/.volta/tools/image/node/16.13.2/bin/node
    Yarn: 1.22.17 - ~/.volta/tools/image/yarn/1.22.17/bin/yarn
    npm: 8.1.2 - ~/.volta/tools/image/node/16.13.2/bin/npm
    Watchman: 4.9.0 - /opt/brew/bin/watchman
  Browsers:
    Chrome: 104.0.5112.79
    Safari: 15.6
    Safari Technology Preview: 16.0
  npmPackages:
    vitest: ^0.22.0 => 0.22.0
```

### Used Package Manager

npm

### Validations

[...]

PR #2983, opened 2023-03-08
fix: --inspect to work inside workers

AriPerkkio:
- Fixes https://github.com/vitest-dev/vitest/issues/1867

This solution doesn't utilize NodeJS **CLI arguments** `--inspect` or `--inspect-brk`, but mimics how they work. 
We can only open inspector in a single `worker_thread` or `child_process` at a time so `threads: false` or `singleThread: true` is required. IDEs may do their own magic which allows debugging multiple `worker_threads`, e.g. VSCode's `autoAttachChildProcesses`.

When these flags are used, it's not possible to debug the main thread. But Vitest users should never need to do that, right?

Minimal demo for testing inspector with `worker_threads` and `child_process`: https://gist.github.com/AriPerkkio/ffd517437c50e64ccdc875b5e3553089

### DIFF
--- a/docs/guide/debugging.md
+++ b/docs/guide/debugging.md
@@ -45,3 +45,19 @@ JavaScript file | ./node_modules/vitest/vitest.mjs
 Application parameters | run --threads false
 
 Then run this configuration in debug mode. The IDE will stop at JS/TS breakpoints set in the editor.
+
+## Node Inspector, e.g. Chrome DevTools
+
+Vitest also supports debugging tests without IDEs. However this requires that tests are not run parallel. Use one of the following commands to launch Vitest.
+
+```sh
+# To run in a single worker
+vitest --inspect-brk --single-thread
+
+# To run in a child process
+vitest --inspect-brk --no-threads
+```
+
+Once Vitest starts it will stop execution and waits for you to open developer tools that can connect to [NodeJS inspector](https://nodejs.org/en/docs/guides/debugging-getting-started/). You can use Chrome DevTools for this by opening `chrome://inspect` on browser.
+
+In watch mode you can keep the debugger open during test re-runs by using the `--single-thread --isolate false` options.
--- a/packages/vitest/src/runtime/child.ts
+++ b/packages/vitest/src/runtime/child.ts
@@ -6,6 +6,7 @@ import type { RuntimeRPC } from '../types/rpc'
 import type { ChildContext } from '../types/child'
 import { mockMap, moduleCache, startViteNode } from './execute'
 import { rpcDone } from './rpc'
+import { setupInspect } from './inspector'
 
 function init(ctx: ChildContext) {
   const { config } = ctx
@@ -58,10 +59,17 @@ function unwrapConfig(config: ResolvedConfig) {
 }
 
 export async function run(ctx: ChildContext) {
-  init(ctx)
-  const { run, executor } = await startViteNode(ctx)
-  await run(ctx.files, ctx.config, ctx.environment, executor)
-  await rpcDone()
+  const inspectorCleanup = setupInspect(ctx.config)
+
+  try {
+    init(ctx)
+    const { run, executor } = await startViteNode(ctx)
+    await run(ctx.files, ctx.config, ctx.environment, executor)
+    await rpcDone()
+  }
+  finally {
+    inspectorCleanup()
+  }
 }
 
 const procesExit = process.exit
--- a/packages/vitest/src/node/cli-wrapper.ts
+++ b/packages/vitest/src/node/cli-wrapper.ts
@@ -11,8 +11,6 @@ const ENTRY = new URL('./cli.js', import.meta.url)
 
 /** Arguments passed to Node before the script */
 const NODE_ARGS = [
-  '--inspect',
-  '--inspect-brk',
   '--trace-deprecation',
   '--experimental-wasm-threads',
   '--wasm-atomics-on-non-shared-memory',
--- a/packages/vitest/src/node/config.ts
+++ b/packages/vitest/src/node/config.ts
@@ -106,6 +106,13 @@ export function resolveConfig(
     resolved.shard = { index, count }
   }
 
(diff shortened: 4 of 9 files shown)


docs/guide/debugging.md
```diff
@@ -45,3 +45,17 @@ JavaScript file | ./node_modules/vitest/vitest.mjs
...
+
+```sh
+# To run in a single worker
+vitest --inspect-brk --singleThread
+
+# To run in a child process
+vitest --inspect-brk --no-threads
+```
+
+Once Vitest starts it will stop execution and waits for you to open developer tools that can connect to [NodeJS inspector](https://nodejs.org/en/docs/guides/debugging-getting-started/). You can use Chrome DevTools for this by opening `chrome://inspect` on browser.
```
AriPerkkio, line 61: @alexlafroscia any thoughts about this part?

Did you expect this kind of workflow in #1867?

docs/guide/debugging.md
```diff
@@ -45,3 +45,17 @@ JavaScript file | ./node_modules/vitest/vitest.mjs
 Application parameters | run --threads false
 
 Then run this configuration in debug mode. The IDE will stop at JS/TS breakpoints set in the editor.
+
+## Node Inspector, e.g. Chrome DevTools
+
+Vitest also supports debugging tests without IDEs. However this requires that tests are not run parallel. Use one of the following commands to launch Vitest.
```
sheremet-va, line 51: Should Vitest fail with a nice error message, if it's not running with either of these? Validation can be done in `node/config.ts`

docs/guide/debugging.md
```diff
@@ -45,3 +45,17 @@ JavaScript file | ./node_modules/vitest/vitest.mjs
...
+
+```sh
+# To run in a single worker
+vitest --inspect-brk --singleThread
```
sheremet-va, line 55: Let's keep it consistent 

```suggestion
vitest --inspect-brk --single-thread
```

docs/guide/debugging.md
AriPerkkio: Added validation for this. Looks like `node:test` has similar behaviour: https://github.com/nodejs/node/blob/c733cc0c7f34916dde4808716e32df62f335dc30/lib/internal/main/test_runner.js#L19

I also noticed that the errors thrown in `node/config.ts` are not shown to users. I'll fix that next and add some test cases.

```sh
$ pnpm run test --shard 2 --watch

> example-project@1.0.0 test /x/vitest-example-project
> vitest run "--shard" "2" "--watch"

```

packages/vitest/src/runtime/child.ts
```diff
@@ -58,10 +59,12 @@ function unwrapConfig(config: ResolvedConfig) {
 }
 
 export async function run(ctx: ChildContext) {
+  const inspectorCleanup = setupInspect(ctx.config)
   init(ctx)
   const { run, executor } = await startViteNode(ctx)
   await run(ctx.files, ctx.config, ctx.environment, executor)
   await rpcDone()
+  inspectorCleanup()
```
sheremet-va, line 67: Do we need to clean it up, if run fails? `try/finally`

sheremet-va (changes requested):

alexlafroscia:
One thing I'm thinking about, in terms of these `--inspect` and `--inspect-brk` _looking_ like the normal flags but not actually being passed down to child processes... is there a risk that users might be confused by that at all?

I occurs to me that the flag could really be called _anything_ -- it could be something like

```
vitest --debug
```

or something like that, with this same behavior, but without any mental baggage that folks might bring along with seeing `--inspect` or `--inspect-brk`.

I definitely don't think it's a major risk or anything, and am really happy to see this change land! It just feels like it could be a _little_ confusing. Using something like `--debug` for this test-debugging situation also means that the original `--inspect` behaviors could be maintained, if that was desirable.

AriPerkkio:
> these `--inspect` and `--inspect-brk` _looking_ like the normal flags but not actually being passed down to child processes

The `--inspect` and `--inspect-brk` CLI flags could be used instead of utilizing `node:inspector` directly but it does not work for `worker_threads`. It works fine for `child_process` though. So for interoperability I think it's better to choose one way and use it in both cases.

This works fine:

```js
import { fork } from "node:child_process";

const filename = fileURLToPath(import.meta.url);

if (!process.env.CHILD_PROCESS) {
  fork(filename, { execArgv: ["--inspect-brk"], env: { CHILD_PROCESS: "true" }, });
} else {
  debugger; // Stops in child_process
}
```

This does not work:

```js
import { isMainThread, Worker } from "node:worker_threads";

const filename = fileURLToPath(import.meta.url);

if (isMainThread) {
  new Worker(filename, { execArgv: ["--inspect-brk"] });
} else {
    debugger; // No stop here
}
```

But if `node:inspector` is used inside `worker_threads` it works just fine. To me this looks like a bug or missing feature of NodeJS.

```diff
import { isMainThread, Worker } from "node:worker_threads";
import { fileURLToPath } from "node:url";
+import inspector from "node:inspector";

const filename = fileURLToPath(import.meta.url);

if (isMainThread) {
  new Worker(filename, { execArgv: ["--inspect-brk"] });
} else {
+  if (process.execArgv.includes("--inspect-brk")) {
+    inspector.open();
+    inspector.waitForDebugger();
+  }

  debugger; // Stops in worker thread!
}
```

We could rename these options to `--debug` but I would rather use the names of existing options. I've checked the NodeJS sources and I cannot spot anything else that these options would do. The internal inspector APIs are a bit different but in the end it should work the same way.

alexlafroscia:
Thanks for such a detailed explanation, @AriPerkkio! That sounds reasonable to me

packages/vitest/src/runtime/child.ts
AriPerkkio: Good point. An open inspector can also prevent the worker thread from terminating.

Just to be sure that inspector does not remain open, I'll wrap all these functions inside the `try`.

PR #2989, opened 2023-03-09
docs: update config example

staff0rd:
Include the triple slash referenced just above in the example

### DIFF
--- a/docs/guide/index.md
+++ b/docs/guide/index.md
@@ -50,6 +50,7 @@ One of the main advantages of Vitest is its unified configuration with Vite. If 
 To configure `vitest` itself, add `test` property in your Vite config. You'll also need to add a reference to Vitest types using a [triple slash command](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html#-reference-types-) at the top of your config file, if you are importing `defineConfig` from `vite` itself.
 
 ```ts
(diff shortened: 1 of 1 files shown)


docs/guide/index.md
```diff
@@ -50,6 +50,7 @@ One of the main advantages of Vitest is its unified configuration with Vite. If
 To configure `vitest` itself, add `test` property in your Vite config. You'll also need to add a reference to Vitest types using a [triple slash command](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html#-reference-types-) at the top of your config file, if you are importing `defineConfig` from `vite` itself.
 
 ```ts
+/// <reference types="vitest" />
```
sheremet-va, line 53: You don't need to do that if you are importing from `vitest/config` (according to the previous paragraph). This example imports from there

sheremet-va (changes requested):

docs/guide/index.md
Jannchie: I'm actually confused here too. I think two examples can be provided, one using `/// <reference types="vitest" />` and the other using `import { defineConfig } from 'vitest/config'`.

Jannchie, line 53:
```suggestion
```

PR #2990, opened 2023-03-09
Update jest-utils.ts

lifeiscontent:
don't do property access to check if something isDomNode

this call is breaking the usage of proxy objects in tests, e.g. if I want to make sure that the object I'm testing is only ever accessing keys it defined.

```ts
  it("Should return whatever you pass in", () => {
    const result = fromPartial({ foo: "bar" });

    expect(result).toEqual({ foo: "bar" });
  });

  it("Should throw an error when the test itself accesses a property not defined on the input", () => {
    const func = (input: { id: string }) => {
      return input.id;
    };

    expect(() => func(fromPartial({}))).toThrow();
  });
```

where fromPartial looks like this:

```ts
export const fromPartial = <T extends {}>(mock: PartialDeep<NoInfer<T>>): T => {
  const proxy = new Proxy(mock, {
    get(target, prop, receiver) {
      let value = target[prop as keyof typeof target];
      let v = typeof value === "function" ? value.bind(target) : value;

      console.log({ v, prop }); // this is where I'm seeing nodeType from vitest
      return v;
    },
  });
  return proxy as T;
};
```


packages/expect/src/jest-utils.ts
```diff
@@ -135,10 +135,6 @@ function eq(
   if (typeof a !== 'object' || typeof b !== 'object')
     return false
 
-  // Use DOM3 method isEqualNode (IE>=9)
-  if (isDomNode(a) && isDomNode(b))
```
sheremet-va, line 139: We can't just remove this because someone wants to.

sheremet-va (changes requested):

packages/expect/src/jest-utils.ts
lifeiscontent: @sheremet-va I agree, but wouldn't this be something that should enhance functionality? e.g. if you're using the JSDOM adapter? if you're not writing tests for non-browser APIs, why should it exist?

sheremet-va: There is no JSDOM adapter for `equality`. If you are afraid of the Proxy usage here we can change it in a few other ways:

- Check that global `Element` exists and check `instanceof` on the value.
- Instead of direct access `obj.nodeType` use `'nodeType' in obj && obj.nodeType`

lifeiscontent: @sheremet-va that's a great suggestion :) Updated.

lifeiscontent:
@sheremet-va friendly ping

dammy001 (approved):

PR #3005, opened 2023-03-13
doc(mocking/date): add caution on useFakeTimers

wongjiahau:
I faced the same issue today when I followed the code in the doc, and then my tests started to hang; it took me some time to figure out that `useFakeTimers` was the culprit.

### DIFF
--- a/docs/guide/mocking.md
+++ b/docs/guide/mocking.md
@@ -343,6 +343,11 @@ const mock = vi.fn(() => console.log('executed'))
 
 describe('delayed execution', () => {
   beforeEach(() => {
+    // Note: callng useFakeTimers can causes async function to hang
+    // Don't put this line if you only need to mock the date but not the timer,
+    // References:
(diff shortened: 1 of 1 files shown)


docs/guide/mocking.md
```diff
@@ -343,6 +343,11 @@ const mock = vi.fn(() => console.log('executed'))
 
 describe('delayed execution', () => {
   beforeEach(() => {
+    // Note: callng useFakeTimers can causes async function to hang
```
sheremet-va, line 346: Fake timers cannot cause async functions to hang. Only if the function uses a timer-based resolve.

sheremet-va (changes requested):

docs/guide/mocking.md
wongjiahau: Out of topic: anyway, I have given up using mock dates, as it doesn't play well with integration test, for example Google Auth failed after the date object is mocked.

PR #3006, opened 2023-03-13
perf(reporters): overall improvements

### LINKED ISSUE #2602: Severe performance regression when reporter with terminal output is enabled
### Describe the bug

Problem:
Full test suite performance is 10x slower when enabling a reporter with terminal output (e.g. `default`, `verbose`, `dot`).

Details:
In converting our unit tests (~8000) from karma to vitest I noticed performance of vitest was significantly worse than karma. After a lot of debugging, I came to find that the `wrapAnsi(` function call, used in the vitest `log-update` dependency, was the culprit. This call alone takes ~95% of test execution time.

https://github.com/sindresorhus/log-update/blob/main/index.js#L54

image

Workaround:
Tests run 10-20x faster if the wrapAnsi line in the `vitest/dist/chunk-snapshot-manager.js` file is commented out. Console output continues to look normal and functional with ansi wrapping disabled

Vitest entrypoint into `log-update` dependency:
https://github.com/vitest-dev/vitest/blob/main/packages/vitest/src/node/logger.ts#L13

image

Thoughts:
Its bizarre that terminal output logic is reducing test performance by an order of magnitude. Ideally the performance would be optimized here, but perhaps terminal logic can be moved to a worker/sub-process?

### Reproduction
TODO: Will add soon. Should be able to reproduce on any large test suite via enabling/disabling `default` or `dot` reporter though.

### System Info

```shell
System:
    OS: macOS 13.0.1
    Memory: 72.25 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.7.0 - ~/.nvm/versions/node/v18.7.0/bin/node
    npm: 8.15.0 - ~/.nvm/versions/node/v18.7.0/bin/npm
  Browsers:
    Brave Browser: 108.1.46.140
    Chrome: 108.0.5359.124
    Firefox: 102.3.0
    Safari: 16.1
  npmPackages:
    vite: ^4.0.0 => 4.0.0 
    vitest: ^0.26.3 => 0.26.3
```

### Used Package Manager

npm

### Validations

AriPerkkio:
- Fixes https://github.com/vitest-dev/vitest/issues/2602

Reduces the workload of expensive `renderTree` by bailing out once enough lines to fill the screen has been constructed. I think this fixes #2602 completely as `wrap-ansi` and related dependencies of `log-update` are not called that much anymore. The [`log-update` already limits the printed rows to fit the screen](https://github.com/sindresorhus/log-update/blob/91822b659b249238602e8a69394fdcef63231f5c/index.js#L20-L33) but as it's using `slice-ansi` for this it is slow. We can help it by reducing the passed rows ourselves as we already know "what makes a row" and don't need to care about ANSIs when reducing the rows.  

Using the reproduction case from https://github.com/vitest-dev/vitest/issues/579#issuecomment-1462503600 as reference with Apple M2 Air:

- Vitest `0.29.2` with `--isolate false`: 5667.68s
- Jest without Babel: 46.271s
- Vitest with PR fix applied and `--isolate false`: 7.98s 
- Vitest with PR fix applied, `--isolate false` and latest `vite`: 4.63s

Before:
image

After:
image

In the __After__ picture its seen that there is a performance issue on Vite's side. I'm working on fixing that one and with my current WIP changes the tests run 2.7s faster, in 5.25s with the fix applied. 
image

Before merging:
- [x] More manual testing
- [x] Check other test reporters as well, e.g. `dot`

### DIFF
--- a/packages/vitest/src/node/reporters/benchmark/table/tableRender.ts
+++ b/packages/vitest/src/node/reporters/benchmark/table/tableRender.ts
@@ -101,8 +101,8 @@ function renderBenchmark(task: Benchmark, tasks: Task[]): string {
   ].join('  ')
 }
 
-export function renderTree(tasks: Task[], options: ListRendererOptions, level = 0) {
-  let output: string[] = []
+export function renderTree(tasks: Task[], options: ListRendererOptions, level = 0): string {
+  const output: string[] = []
 
   let idx = 0
   for (const task of tasks) {
@@ -154,7 +154,7 @@ export function renderTree(tasks: Task[], options: ListRendererOptions, level = 
 
     if (task.type === 'suite' && task.tasks.length > 0) {
       if (task.result?.state)
-        output = output.concat(renderTree(task.tasks, options, level + 1))
+        output.push(renderTree(task.tasks, options, level + 1))
     }
     idx++
   }
--- a/packages/vitest/src/node/reporters/renderers/dotRenderer.ts
+++ b/packages/vitest/src/node/reporters/renderers/dotRenderer.ts
@@ -7,25 +7,64 @@ export interface DotRendererOptions {
   logger: Logger
 }
 
-const check = c.green('·')
-const cross = c.red('x')
-const pending = c.yellow('*')
-const skip = c.dim(c.gray('-'))
+interface Icon { char: string; color: (char: string) => string }
 
-function render(tasks: Task[]) {
+const check: Icon = { char: '·', color: c.green }
+const cross: Icon = { char: 'x', color: c.red }
+const pending: Icon = { char: '*', color: c.yellow }
+const skip: Icon = { char: '-', color: (char: string) => c.dim(c.gray(char)) }
+
+function getIcon(task: Task) {
+  if (task.mode === 'skip' || task.mode === 'todo')
+    return skip
+  switch (task.result?.state) {
+    case 'pass':
+      return check
+    case 'fail':
+      return cross
+    default:
+      return pending
+  }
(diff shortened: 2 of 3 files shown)


AriPerkkio:
The reason why `renderTree` is slow seems to be these lines:

https://github.com/vitest-dev/vitest/blob/1fe8286c790cef9063b1f5f8147f6d41fc14bf79/packages/vitest/src/node/reporters/renderers/listRenderer.ts#L143-L150

Just by replacing the `output = output.concat(<fn>)` to `output.push(<fn>)` the `renderTree` will reduce from 1200ms to 11ms when called **once** with 5000 tasks. 

image

image

In overall the whole test run drops ~1300ms more. Now the total test run of reproduction setup is at 4.63s.

sheremet-va:
> Just by replacing the output = output.concat(<fn>) to output.push(<fn>) the renderTree will reduce from 1200ms to 11ms when called once with 5000 tasks.

Yes, `concat` is evil!

AriPerkkio:
`dot` reporter:
- Passing 5000 `c.green('·')` strings to `log-update` is bad for performance. `checkAnsi` + `sliceAnsi` take most of the execution time.
- Passing a single `c.green('·'.repeat(5000))` to `log-update` performs fine

I've refactored dot renderer to group sibling task icons into a single ANSI color code block.

sheremet-va (approved):
LGTM 👍🏻 Thank you!

dammy001 (approved):

Issue #1771, opened 2022-08-02
consider --timeout in CLI options for use with VSCode debugger


panick-carousel:
### Clear and concise description of the problem

When using the debugger in vscode, our tests keep timing out, but we don't want to have to set the timeout in the config file because this works great for our CI setup.

### Suggested solution

Ideally at invocation time (ie: from the launcher) we could set a timeout so that when we hit a breakpoint, the delay will not cause the test to fail and therefore not proceed any further down the codepath we are trying to troubleshoot.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

sheremet-va:
Sorry, I don't understand what you want. Do you need timeout after tests are run, or before Vitest runs tests?

panick-carousel:
I am requesting to be able to override the config file timeout setting by passing a cli option. Specifically this setting: https://vitest.dev/config/#testtimeout

This means vscode debugger becomes usable easily.

larswww:
Having the same issue. I am debugging the code and due to time spent paused on a breakpoint, I hit the 5000ms default limit. I don't want to increase the timeout because when I run the tests without the debugger, then I want it to be limited. 

Perhaps an interim solution could be to set timeout to infinite, based on process.env?

mekapotulas:
+1 facing this issue as well

PR #3019, opened 2023-03-17
feat: `--test-timeout` CLI argument

AriPerkkio:
- Closes https://github.com/vitest-dev/vitest/issues/1771

Exposes `testTimeout` configuration option to be configured via CLI, e.g. `vitest run --test-timeout=50000`.

### DIFF
--- a/packages/vitest/src/node/cli.ts
+++ b/packages/vitest/src/node/cli.ts
@@ -48,6 +48,7 @@ cli
   .option('--no-color', 'Removes colors from the console output')
   .option('--inspect', 'Enable Node.js inspector')
   .option('--inspect-brk', 'Enable Node.js inspector with break')
+  .option('--test-timeout <time>', 'Default timeout of a test in milliseconds (default: 5000)')
   .help()
 
 cli
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -536,6 +536,7 @@ This can improve performance in some cases, but might cause segfault in older No
 
 - **Type:** `number`
 - **Default:** `5000`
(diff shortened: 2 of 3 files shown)


packages/vitest/src/node/cli.ts
```diff
@@ -48,6 +48,7 @@ cli
   .option('--no-color', 'Removes colors from the console output')
   .option('--inspect', 'Enable Node.js inspector')
   .option('--inspect-brk', 'Enable Node.js inspector with break')
+  .option('--test-timeout', 'Default timeout of a test in milliseconds')
```
sheremet-va, line 51: 1. Does `cac` know that it's a number?
2. Should we specify default in brackets like we do for other options?

sheremet-va (changes requested):

packages/vitest/src/node/cli.ts
AriPerkkio: 2: Done

1: It works when a number is passed:
```
$ pnpm run test --test-timeout 1
> vitest run "--test-timeout" "1"

...

 FAIL  test/animals.test.ts > Dog says woof
Error: Test timed out in 1ms.
```

But it actually accepts anything:
```
$ pnpm run test --test-timeout some-text
> vitest run "--test-timeout" "some-text"

...

 FAIL  test/animals.test.ts > Dog says woof
Error: Test timed out in some-textms.
```

But even the existing `--shard` option accepts incorrect type so I guess there is no checking for types in place:
```
$ pnpm run test --shard 1
> vitest run "--shard" "1"

TypeError: options.shard.split is not a function or its return value is not iterable
```

PR #3020, opened 2023-03-17
fix: show correct line numbers in stack trace when using vi.resetModules()

### LINKED ISSUE #3001: Wrong line numbers in stack trace when using `vi.resetModules()`
### Describe the bug

Hi, my team noticed an issue with line numbers in our errors being off when we upgraded from `vitest` v0.24.4 to v0.28.5 (and the problem still seems to be present in the latest release). It looks like the problem was introduced in the v.0.25.0 release because anything before that still reports the correct line numbers.

Through some trial and error, I narrowed down the cause to be using `vi.resetModules()` in any test. That causes the line numbers in the stack trace to be reported incorrectly. We use `vi.resetModules()` in many `beforeEach()` hooks in our tests to isolate state between tests.

Here's an example with a basic failing test that includes `vi.resetModules()` and incorrect line number: https://stackblitz.com/edit/vitest-dev-vitest-vsb2zh?file=test/basic.test.ts

If you comment out `vi.resetModules()`, the failing test will report the line numbers correctly in the stack trace.

Thanks and let me know if you need any other information.

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-vsb2zh?file=test/basic.test.ts

### System Info

```shell
System:
    OS: macOS 12.6.3
    CPU: (8) x64 Intel(R) Core(TM) i7-1068NG7 CPU @ 2.30GHz
    Memory: 19.06 MB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.19.0 - ~/n/bin/node
    Yarn: 1.22.17 - ~/n/bin/yarn
    npm: 8.15.0 - /usr/local/bin/npm
  Browsers:
    Brave Browser: 110.1.48.164
    Chrome: 110.0.5481.177
    Firefox: 110.0.1
    Safari: 16.1
  npmPackages:
    @vitejs/plugin-basic-ssl: 1.0.1 => 1.0.1
    @vitejs/plugin-vue: 4.0.0 => 4.0.0
    @vitest/coverage-c8: 0.29.2 => 0.29.2
    @vitest/coverage-istanbul: 0.29.2 => 0.29.2
    vite: 4.0.4 => 4.0.4
    vitest: 0.29.2 => 0.29.2
```

### Used Package Manager

pnpm

### Validations

sheremet-va:
Fixes #3001

### DIFF
--- a/packages/vite-node/src/client.ts
+++ b/packages/vite-node/src/client.ts
@@ -92,6 +92,14 @@ export class ModuleCacheMap extends Map<string, ModuleCache> {
     return this.deleteByModuleId(this.normalizePath(fsPath))
   }
 
+  invalidateModule(mod: ModuleCache) {
+    delete mod.evaluated
+    delete mod.resolving
+    delete mod.promise
+    delete mod.exports
+    return true
+  }
+
   /**
    * Invalidate modules that dependent on the given modules, up to the main entry
    */
--- a/packages/vitest/src/utils/index.ts
+++ b/packages/vitest/src/utils/index.ts
@@ -31,10 +31,10 @@ export function resetModules(modules: ModuleCacheMap, resetMocks = false) {
     // don't clear mocks
     ...(!resetMocks ? [/^mock:/] : []),
   ]
-  modules.forEach((_, path) => {
+  modules.forEach((mod, path) => {
     if (skipPaths.some(re => re.test(path)))
       return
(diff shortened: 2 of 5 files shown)


jdk2pq:
Thank you for a quick turnaround on this fix @sheremet-va! My team and I really appreciate it! 🚀

PR #3026, opened 2023-03-18
fix: handle cloning proxied classes w/ enumerable getters

tgriesser:
Fixes a very, very edge case I ran into with Vitest's `clone`, where a class is proxied in a way that exposes its getters as enumerable properties.

```ts
const dataSym = Symbol.for('MyClass.data')

class MyClass {
  constructor(obj) {
    this[dataSym] = obj
    return new Proxy(this, {
      ownKeys() {
        return Reflect.ownKeys(obj);
      },
      getOwnPropertyDescriptor(target, p) {
        return {
          ...Reflect.getOwnPropertyDescriptor(obj, p),
          enumerable: true,
        };
      }
    });
  }
  
  get a() {
    return this[dataSym].a
  }
  
  get b() {
    return this[dataSym].b    
  }
}
```

```ts
deepClone(new MyClass({a: 1, b: 2}))
```

Previously:

```
TypeError: Cannot set property a of #<MyClass> which has only a getter
```

New Behavior: clones properly

### DIFF
--- a/packages/utils/src/helpers.ts
+++ b/packages/utils/src/helpers.ts
@@ -88,7 +88,7 @@ export function clone<T>(val: T, seen: WeakMap<any, any>): T {
     const props = getOwnProperties(val)
     for (const k of props) {
       Object.defineProperty(out, k, {
-        ...Object.getOwnPropertyDescriptor(val, k),
+        enumerable: true,
         value: clone((val as any)[k], seen),
       })
     }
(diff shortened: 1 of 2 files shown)


sheremet-va:
This will fail, if descriptor has only getter: 

```js
Object.defineProperty(objB, 'getter', {
  configurable: true,     
  enumerable: false,
  get: () => 1,
})
```

PR #3030, opened 2023-03-18
chore: fix typescript performance

sheremet-va:
I noticed we have a poor TS performance in the repo for quite some time now. I decided to check the performance on PC with 64GB, and it took a few seconds to update autocomplete 👀 

`tsc --generateTrace` showed that `deepMerge` takes _a lot of_ time to process in the `pools/child.ts` file:

image

Now it's a lot better:

image

We don't really use the `deepMerge` type merging, so I decided to just remove it.

### DIFF
--- a/packages/vitest/src/node/core.ts
+++ b/packages/vitest/src/node/core.ts
@@ -133,7 +133,7 @@ export class Vitest {
   }
 
   getSerializableConfig() {
-    return deepMerge<ResolvedConfig>({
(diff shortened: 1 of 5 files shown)


AriPerkkio:
Great finding! I've also noticed that sometimes the IDE intellisense takes a while and can even stop working on a 16GB RAM. Restaring VSCode's Typescript server has restored the intellisense. Hopefully this fixes it.

PR #3031, opened 2023-03-19
feat: process timeout to log names of stuck test files

AriPerkkio:
- Related to #2008
- Enabled in https://github.com/tinylibs/tinypool/issues/49

Include names of the test files that prevent Worker from terminating. This will help users to identify which test files are causing their Vitest process to hang. **This does not fix** process hangs but just helps debugging them. 

For example with `trpc` repository there are two test cases which can make the process hang. If these two test cases are excluded from the test run, the process never hangs. I'm not familiar with trpc's codebase but I was able to spot that those two test cases are the only ones that use Miniflare which could be the root cause (?) in their case. Ideally this feature would allow users to spot these kind of differences from their test cases and narrow down the root cause. 

Example log from trpc repository:

```
 Test Files  124 passed (124)
      Tests  613 passed | 2 skipped (615)
   Start at  10:06:39
   Duration  23.23s (transform 1.78s, setup 10.16s, collect 42.84s, tests 31.86s)

close timed out after 10000ms
Failed to terminate worker while running /x/y/z/trpc/packages/tests/server/interop/adapters/fetch.test.ts.
```

### DIFF
--- a/packages/vitest/src/node/core.ts
+++ b/packages/vitest/src/node/core.ts
@@ -597,6 +597,7 @@ export class Vitest {
     setTimeout(() => {
       this.report('onProcessTimeout').then(() => {
         console.warn(`close timed out after ${this.config.teardownTimeout}ms`)
+        this.state.getProcessTimeoutCauses().forEach(cause => console.warn(cause))
         process.exit()
       })
     }, this.config.teardownTimeout).unref()
--- a/packages/vitest/package.json
+++ b/packages/vitest/package.json
@@ -137,7 +137,7 @@
     "std-env": "^3.3.1",
     "strip-literal": "^1.0.0",
     "tinybench": "^2.3.1",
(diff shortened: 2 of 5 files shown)


packages/vitest/src/node/core.ts
```diff
@@ -42,6 +42,7 @@ export class Vitest {
   filenamePattern?: string
   runningPromise?: Promise<void>
   closingPromise?: Promise<void>
+  processTimeoutCauses: string[] = []
```
AriPerkkio, line 45: Is there any better place than this `ctx.processTimeoutCauses` to store these errors? I thought about `ctx.state.processTimeoutCauses` but I wasn't sure that was right either.

sheremet-va: State is probably better

Issue #2859, opened 2023-02-13
toMatchObject(): Error: [vitest] No "Symbol(Symbol.toStringTag)" export is defined


yuliankarapetkov:
### Describe the bug

I am mocking a module:

```
vi.mock("country-region-data", () => {
    return {
        allCountries: ["UK"],
        UK: "UK",
    }
})
```

Then, testing it:
```
// getCountries just imports the module - import("country-region-data")
const result = await getCountries()

expect(result).toMatchObject({
  allCountries: ["UK"],
  UK: "UK",
})
```

When I run the test, I get the following error:
```
Error: [vitest] No "Symbol(Symbol.iterator)" export is defined on the "country-region-data" mock. 
Did you forget to return it from "vi.mock"?
```

If I spread the object in `expect` (`expect({ ...result })`), the test succeeds. 

I think it shouldn't expect an iterator in this case.

### Reproduction

https://stackblitz.com/edit/node-pnfxmw?file=test%2Fcountries.test.ts,package.json

### System Info

```shell
System:
    OS: macOS 13.1
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 1.06 GB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 19.2.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 8.13.1 - 
  Browsers:
    Brave Browser: 110.1.48.158
    Chrome: 109.0.5414.119
    Firefox: 105.0.1
    Safari: 16.2
  npmPackages:
    vite: ^4.1.1 => 4.1.1 
    vitest: ^0.28.4 => 0.28.4
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
[...]

PR #3036, opened 2023-03-20
fix: mocking value proxy filter Symbol static properties

ChpShy:
Fix: #2859


sheremet-va:
There is a list of well known symbols. Maybe we can just add them? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol#static_properties

Also, please use formatter

sheremet-va (approved):

PR #3040, opened 2023-03-20
feat: implement `istanbul` coverage support for browser testing

### LINKED ISSUE #3086: In browser mode, aliases in the test section of vite.config are ignored
### Describe the bug

When using browser mode, aliases that are defined in the vite config's resolve.alias section work as expected but aliases defined in test.alias throw an error as if they are packages that have not been installed.

vite config example:
```ts

import { resolve } from 'path';

function pathResolve(dir: string) {
  return resolve(process.cwd(), '.', dir);
}

export default defineConfig({
  resolve: {
    // works
    alias: [{ find: /@src\//, replacement: pathResolve('src') + '/' }],
  },
  test: {
    browser: { enabled: true },
    // fails
    alias: [{ find: /@tests\//, replacement: pathResolve('test') + '/' }],
  },
});
```

output example: 
```
Error: The following dependencies are imported but could not be resolved:

  @tests/testUtils (imported by path/to/the/test.ts)

Are they installed?
```

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-rb4nqu?file=vite.config.ts

### System Info

```shell
System:
    OS: Windows 10 10.0.19042
    CPU: (8) x64 Intel(R) Core(TM) i5-10210U CPU @ 1.60GHz
    Memory: 7.92 GB / 15.78 GB
  Binaries:
    Yarn: 1.22.19 - C:\Program Files\nodejs\yarn.CMD
    npm: 8.15.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (111.0.1661.54)
    Internet Explorer: 11.0.19041.1202
  npmPackages:
    @vitest/browser: ^0.29.7 => 0.29.7
    @vitest/ui: latest => 0.29.7
    vite: latest => 4.2.1
    vitest: latest => 0.29.7
```

### Used Package Manager

npm

### Validations

sheremet-va:
This PR adds coverage support for `istanbul` provider. It is currently not possible to use `c8` as a coverage provider when running in the browser. 

Also closes #3086

### DIFF
--- a/test/watch/test/file-watching.test.ts
+++ b/test/watch/test/file-watching.test.ts
@@ -65,7 +65,7 @@ test('editing config file reloads new changes', async () => {
 })
 
 describe('browser', () => {
-  test('editing source file triggers re-run', async () => {
+  test.runIf((process.platform !== 'win32'))('editing source file triggers re-run', async () => {
     const vitest = await startWatchMode('--browser.enabled', '--browser.headless', '--browser.name=chrome')
 
     writeFileSync(sourceFile, editFile(sourceFileContent), 'utf8')


AriPerkkio:
I'm really excited about these browser testing features! 


I'll need to check how WebDriverIO can be setup on macos before looking at the PR's code:

```
> vitest run --browser chrome

Error: Failed to find a Chrome installation via:
$ which chrome: not found: chrome
$ which chromium: not found: chromium
$ which google-chrome: not found: google-chrome
```

> I don't think it's possible to make `c8`

In #2786 we removed most of the Node related APIs from `c8` provider. Same V8 profiler commands work on Chromium based browsers when they are spawned with certain exec args. But let's first focus on the `istanbul` provider and see if `c8` for Chromium browsers is worth the work later.

sheremet-va:
> I'll need to check how WebDriverIO can be set up on MacOS before looking at the PR's code:

I would recommend running `pnpm coverage` and setup browser of your choice in the config. It should not matter.

I managed to make coverage work with `--isolate` flag, but if I run all tests it seems there is some kind of a race condition that makes coverage not work.

AriPerkkio:
> some kind of a race condition that empties coverage before the function is actually running

This might be causing that:

https://github.com/vitest-dev/vitest/blob/be344518b48e56c29127ff61b728be4281345836/packages/coverage-istanbul/src/index.ts#L12-L14

packages/coverage-c8/src/index.ts
```diff
@@ -1,6 +1,11 @@
-export * from './takeCoverage'
+import * as coverage from './takeCoverage'
 
-export async function getProvider() {
-  const { C8CoverageProvider } = await import('./provider')
-  return new C8CoverageProvider()
+export default {
+  ...coverage,
+  async getProvider() {
```
AriPerkkio, line 5: We should be throwing an error if this is called in `--browser` mode. Is there some global flag that could be used easily?

```ts
if (globalThis.__vitest_browser_or_something__)
  throw new Error('@vitest/c8 does not work with --browser (yet). Use @vitest/istanbul instead.')
```

sheremet-va: No, there is no flag currently. But I would probably throw error sooner on the node side with `ctx.isBrowserEnabled` check.

packages/browser/src/client/main.ts
```diff
@@ -80,14 +80,27 @@ async function runTests(paths: string[], config: any) {
...
+  }
 
   if (!runner) {
     const { VitestTestRunner } = await importId('vitest/runners') as typeof import('vitest/runners')
-    const BrowserRunner = createBrowserRunner(VitestTestRunner)
+    const BrowserRunner = createBrowserRunner(VitestTestRunner, { takeCoverage: () => takeCoverageInsideWorker(config.coverage, executor) })
     runner = new BrowserRunner({ config, browserHashMap })
   }
 
+  await startCoverageInsideWorker(config.coverage, executor)
```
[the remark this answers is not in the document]
sheremet-va: Feel free to edit PR as you see fit. I also did not start with the tests yet, if you want to touch those.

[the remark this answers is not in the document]
AriPerkkio: Sure, I can make the changes and add tests. Should I have write access rights to your Git remote? Typically Github project owners can write to other developer's remotes when someone opens a PR to their project. But I think I don't have required rights to your Git remote:

```
~/Git/vitest (feat/browser-coverage-support) $ git push
ERROR: Permission to sheremet-va/vitest.git denied to AriPerkkio.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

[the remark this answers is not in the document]
sheremet-va: Hm, it says you should be allowed 🤔 

Screenshot 2023-03-23 at 15 38 13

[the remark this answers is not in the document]
sheremet-va: I've added you as a collaborator, maybe it will work 🤷🏻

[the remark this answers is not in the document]
AriPerkkio: That worked, thanks! 
I think my access rights for this project are `collaborator` in Github terms. `maintainer` is a separate one.

test/coverage-test/test/coverage.test.ts
```diff
@@ -1,8 +1,11 @@
 import { expect, test } from 'vitest'
-import { pythagoras } from '../src/index.mjs'
 import { implicitElse } from '../src/implicitElse'
 import { useImportEnv } from '../src/importEnv'
 
+// TODO Fix: Browser fails to load if extensions ".mjs" is used
+// @ts-expect-error -- extension
+import { pythagoras } from '../src/index'
+
```
AriPerkkio, line 8: I'll look into this next. Any pointers are appreciated. 🙌  
Do we set any specific settings on Node side for this that is not included when running in browser?


```
// Filename on filesystem is "../src/index.mts"
import { pythagoras } from '../src/index.mjs'
```

```
12:34:25 PM [vite] Internal server error: Failed to resolve import "../src/index.mjs" from "test/coverage.test.ts?v=1679654065071". Does the file exist?
  Plugin: vite:import-analysis
  File: /x/vitest/test/coverage-test/test/coverage.test.ts?v=1679654065071:7:27
  2  |  import { implicitElse } from "../src/implicitElse";
  3  |  import { useImportEnv } from "../src/importEnv";
  4  |  import { pythagoras } from "../src/index.mjs";
     |                              ^
  5  |  test("Math.sqrt()", async () => {
  6  |    expect(pythagoras(3, 4)).toBe(5);
      at formatError (file:///x/vitest/node_modules/.pnpm/vite@4.0.0_@types+node@18.7.13/node_modules/vite/dist/node/chunks/dep-ed9cb113.js:42301:46)
      at TransformContext.error (file:///x/vitest/node_modules/.pnpm/vite@4.0.0_@types+node@18.7.13/node_modules/vite/dist/node/chunks/dep-ed9cb113.js:42297:19)
      at normalizeUrl (file:///x/vitest/node_modules/.pnpm/vite@4.0.0_@types+node@18.7.13/node_modules/vite/dist/node/chunks/dep-ed9cb113.js:40600:33)
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
      at async TransformContext.transform (file:///x/vitest/node_modules/.pnpm/vite@4.0.0_@types+node@18.7.13/node_modules/vite/dist/node/chunks/dep-ed9cb113.js:40753:47)
      at async Object.transform (file:///x/vitest/node_modules/.pnpm/vite@4.0.0_@types+node@18.7.13/node_modules/vite/dist/node/chunks/dep-ed9cb113.js:42572:30)
      at async loadAndTransform (file:///x/vitest/node_modules/.pnpm/vite@4.0.0_@types+node@18.7.13/node_modules/vite/dist/node/chunks/dep-ed9cb113.js:40378:29)
```

sheremet-va: > Do we set any specific settings on Node side for this that is not included when running in browser?

No, it should work out of the box as far as I'm concerned. You can try doing a reproduction in https://vite.new with a fresh VIte project

sheremet-va: I don't think they are necessary anymore, since we reload the page on watch change. And these were there to allow revalidation.

AriPerkkio: After some testing it seems that we need these `v` query parameters. In watch mode the changes are not reloaded on re-runs. I guess even the browser cache could be serving the old file.

We can leave this TODO here as is and come back to it later. After looking at this more think this is a bug in `vite:resolve` plugin - I'll keep debugging that one and see if fix could be applied later.

sheremet-va: > We can leave this TODO here as is and come back to it later

We can also just strip `v` in `resolveId` hook before `vite:resolve` kicks in.

packages/browser/src/node/index.ts
```diff
@@ -18,6 +18,16 @@ export default (base = '/'): Plugin[] => {
     {
       enforce: 'pre',
       name: 'vitest:browser',
+      async resolveId(id, importer, options) {
+        if (id.includes('?'))
+          // Work-around for vite:resolve. The "import('./file.mjs?v=123')" fails to load "./file.mts" and throws error.
```
sheremet-va, line 23: I think this might introduce bugs with other plugins that rely on the query. We should only remove `?v=` here I think.

AriPerkkio: Good catch. I'll change this to `.split('?v=')[0]` instead. There's still a small change that some library is using same query parameter and we end up removing that.

sheremet-va: Technically Vite also uses `v=` query for optimized dependencies. And this might actually be the reason it doesn't work 🤔 Did you try changing the query to something else? Like `vitest=` or something

AriPerkkio: I'll change this to `?vitest=` so that we'll prevent conflicting with possible other queries. 

This got accidentally fixed in Vite but is still unreleased: https://github.com/vitejs/vite/pull/12571.

packages/coverage-c8/src/index.ts
AriPerkkio: The config checking will now throw error: 752fcd2

packages/vitest/src/node/config.ts
```diff
@@ -113,6 +113,9 @@ export function resolveConfig(
     }
   }
 
+  if (resolved.coverage.enabled && resolved.coverage.provider === 'c8' && resolved.browser.enabled)
```
sheremet-va, line 116: It's possible to run tests in a browser only for a subset of files. That's why there is a `ctx.isBrowserEnabled`

AriPerkkio: Do you mean a case where the `poolMatchGlobs` would be used?

Should we in those cases print a warning message that `c8` is simply ignore in browser tests and still run the tests. In Node tests it would be enabled as before.

sheremet-va: Yes, I mean both. `browser` can be disabled, but might still have browser tests in `poolMatchGlobs`, and vice versa.

AriPerkkio: The `poolMatchGlobs` is now detected in [ae63ed0](https://github.com/vitest-dev/vitest/pull/3040/commits/ae63ed0865e70f7742fa51f3ce6b0a62a7f15174).

packages/browser/src/node/index.ts
AriPerkkio: Looks like this actually breaks watch mode. We'll need to use `?v=` query and pass that to Vite. Without the query parameter, or even with `?vitest=` query, the file updating triggers the re-load but file content is not updated. 

I'll start by adding a test case for watch mode so that this does not break again.

test/watch/test/file-watching.test.ts
```diff
@@ -63,3 +71,19 @@ test('editing config file reloads new changes', async () => {
     expect(vitest.getOutput()).toMatch('ok 2')
   })
 })
+
+describe('browser', () => {
+  test('editing source file triggers re-run', async () => {
```
AriPerkkio, line 76: This test fails for some reason. I've been looking into it for a while but have no idea what's going on.

It works fine when run without `--browser` flags:

```sh
$ pnpm run test run --root fixtures

> @vitest/test-watch@ test /x/vitest/test/watch
> vitest "run" "--root" "fixtures"


 RUN  v0.29.7 /x/vitest/test/watch/fixtures

 ✓ math.test.ts (1)
   ✓ sum
 ✓ example.test.ts (1)
   ✓ getHello

 Test Files  2 passed (2)
```

But with `--browser` it does not respect `--root` when collecting tests. And somehow something ends up loading  test case from `floating-vue` dependency.

```sh
$ pnpm run test run --root fixtures --browser.enabled --browser.name=chrome

> @vitest/test-watch@ test /x/vitest/test/watch
> vitest "run" "--root" "fixtures" "--browser.enabled" "--browser.name=chrome"


 RUN  v0.29.7 /x/vitest/test/watch/fixtures
      Browser runner started at http://localhost:63315

1:58:28 PM [vite] Internal server error: Failed to resolve import "@peeky/test" from "../../node_modules/.pnpm/floating-vue@2.0.0-y.0_vue@3.2.45/node_modules/floating-vue/src/directives/v-tooltip.spec.ts?vitest=1679914708888". Does the file exist?

 ❯ ../node_modules/@vitest/browser/node_modules/@vitest/ui/node_modules/floating-vue/src/directives/v-tooltip.spec.ts (0)
 ✓ example.test.ts (1)
   ✓ getHello
 ❯ ../test/stdout.test.ts (0)
 ✓ math.test.ts (1)
   ✓ sum
 ❯ ../test/file-watching.test.ts (0)
 ❯ ../test/stdin.test.ts (0)

 Test Files  2 passed (2)
      Tests  2 passed (2)
   Start at  13:58:27
   Duration  2.19s (transform 0ms, setup 0ms, collect 111ms, tests 12ms)

```

sheremet-va: Did you try making `root` an absolute file? `resolve(__dirname, 'fixtures')` for example

sheremet-va: It is probably bundled somehow and imported somewhere 🤔 Looking at `Test Files`, it says `2 passed (2)` - so only those were passed down to the browser, but others were imported from the outside somehow during the import of `math` and `example`

AriPerkkio: The [adfaea2](https://github.com/vitest-dev/vitest/pull/3040/commits/adfaea20889d5575ea8f5c747f80a329f35addb7) fixes this.

sheremet-va:
@AriPerkkio seems like test/watch is failing 🤔

AriPerkkio:
Either the watch mode is broken on Windows or the Windows CI just cannot run the test case.

It fails even when CI runs only `test/watch` set, so it shouldn't be caused by high CPU load or anything. https://github.com/AriPerkkio/vitest/actions/runs/4540634358/jobs/8001791530

Should we just disable the test on Windows CI?

sheremet-va:
> Should we just disable the test on Windows CI?

We do this for other execa tests, so I don't see the reason not to 🤔

packages/coverage-c8/src/index.ts
```diff
@@ -1,6 +1,11 @@
-export * from './takeCoverage'
+import * as coverage from './takeCoverage'
 
-export async function getProvider() {
```
AriPerkkio, line 3: This removal of named export `getProvider` caused a breaking change. The `@vitest/coverage-istanbul` is still exporting it as both named and default. Was this intentional @sheremet-va? 

There are some reports where users are using different versions of `vitest` and `@vitest/coverage-c8` and the API breaks:
- https://github.com/vitest-dev/vitest/issues/3125
- https://github.com/vitest-dev/vitest/issues/3118

sheremet-va: Yeah, not intentional. Should've either not remove it or update `peerDependencies`. The next version is locked to be 0.30.0 though, so we will have to wait for a fix, but PR is welcome.

Issue #3052, opened 2023-03-21
typing of expect.objectContaining not generic


ArtyMaury:
### Describe the bug

expect.objectContaining doesn't allow to type the argument.

### Reproduction

`expect.objectContaining( { } /*any*/)`

### System Info

```shell
System:                                                
    OS: Windows 10 10.0.19045                            
    CPU: (8) x64 Intel(R) Core(TM) i5-8400H CPU @ 2.50GHz
    Memory: 1.68 GB / 15.82 GB                           
  Binaries:                                              
    Node: 16.17.1 - ~\Documents\Programmes\NodeJs\node.EXE
    Yarn: 1.22.19 - ~\Documents\Programmes\NodeJs\yarn.CMD
    npm: 8.19.2 - ~\Documents\Programmes\NodeJs\npm.CMD
  Browsers:
    Chrome: 111.0.5563.65
    Edge: Spartan (44.19041.1266.0), Chromium (111.0.1661.44)
    Internet Explorer: 11.0.19041.1566
  npmPackages:
    @vitejs/plugin-react: ^3.1.0 => 3.1.0
    vite: ^4.2.1 => 4.2.1
    vitest: ^0.29.7 => 0.29.7
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #3053, opened 2023-03-21
fix: add generic to expect.objectContaining type

ArtyMaury:
closes https://github.com/vitest-dev/vitest/issues/3052


sheremet-va (changes requested):
Why do we need so many fromatting edits for such a small change?

ArtyMaury:
Sorry, I forgot to disable my formatter, should be better now

Issue #2954, opened 2023-03-03
Provide option to disable test case printf truncation


jasperdunn:
### Clear and concise description of the problem

As a developer using Vitest I want to be able to use `test.each` in a way that helps me to understand what tests are breaking without losing some of the description.

I am using test cases via `test.each` to reduce the amount of test boilerplate code.

However when using `$description` I am getting only a truncated version of the text, which doesn't help determine which test is breaking.

Each test case object has `expected: ReturnType<Fn>` and `description: string` properties;

```
describe('getTimelineRange', () => {
  test('throws error when there are no rows', () => {
    expect(() => getTimelineRange([])).toThrowError();
  });

  test.each(getTimelineRangeCases)('$description', ({rows, expected}) => {
    expect(getTimelineRange(rows)).toEqual(expected);
  });
});
```

I am getting this as my test output:

<img width="387" alt="Screen Shot 2023-03-03 at 10 03 12 am" src="https://user-images.githubusercontent.com/26566353/222594628-fb02bd61-0f31-4bcc-8994-c808aae54fdc.png">

I'd prefer to get the full description, instead of a truncated version.

### Suggested solution

Could a setting be exposed to disable or set the truncation size?

### Alternative

For now I am considering either:

Prefixing a number
<img width="387" alt="Screen Shot 2023-03-03 at 10 11 25 am" src="https://user-images.githubusercontent.com/26566353/222596851-088dad17-d15c-4db6-911e-88a622d2e8a5.png">

or

Returning the entire object as a string via `%s` -> `test.each(getTimelineRangeCases)('%s', ({rows, expected}) => {`
<img width="793" alt="Screen Shot 2023-03-03 at 10 12 26 am" src="https://user-images.githubusercontent.com/26566353/222597086-c8769696-f167-40de-bd41-6cf042b8d4fb.png">

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

sheremet-va:
This is configurable with chai config: https://github.com/vitest-dev/vitest/issues/2791#issuecomment-1413347800

```ts
import { chai } from 'vitest'
[...]

Issue #2791, opened 2023-02-02
toMatch output is truncated


Shinigami92:
### Describe the bug

While using e.g. `toMatch` the expected result is truncated even when setting `outputTruncateLength` to 1000

```node
 FAIL  test/git.spec.ts > git > random seeded tests for seed 3694299162975756 > commitEntry > should return a valid random commitEntry
AssertionError: expected 'Author: Reynold71 <Reynold_Walsh57@ya…' to match /^Author: [\w_\. ]+ \<[\w\.]+@[\w\.]+\>$/
 ❯ test/git.spec.ts:62:30
     60|             expect(parts[5]).toMatch(/^\s{4}.+$/);
     61|           } else {
     62|             expect(parts[1]).toMatch(/^Author: [\w_\. ]+ \<[\w\.]+@[\w\.]+\>$/);
       |                              ^
     63|             expect(parts[2]).toMatch(/^Date: .+$/);
     64|             expect(parts[3]).toBe('');
```

It might be somewhat related to #2448
But there is no expected vs actual comparison/diff, so it's harder here to find the problem of the failing test

### Reproduction

will add if really needed

### System Info

```shell
System:
    OS: macOS 13.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 3.49 GB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.4.0 - ~/.nvm/versions/node/v18.4.0/bin/node
    npm: 8.12.1 - ~/.nvm/versions/node/v18.4.0/bin/npm
  Browsers:
    Chrome: 109.0.5414.119
    Safari: 16.2
  npmPackages:
    @vitest/coverage-c8: ~0.28.3 => 0.28.3 
    @vitest/ui: ~0.28.3 => 0.28.3 
    vite: ~4.0.4 => 4.0.4 
    vitest: ~0.28.3 => 0.28.3
```

### Used Package Manager

pnpm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
[...]

PR #3066, opened 2023-03-23
feat: add chai config

btea:
rel #2954

### DIFF
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -1389,8 +1389,8 @@ The number of milliseconds after which a test is considered slow and reported as
 
 ### chaiConfig
 
-- **Type:** `{ includeStack?, showDiff?, truncateThreshold?, useProxy?, proxyExcludedKeys? }`
-- **Default:** `{ includeStack: false, showDiff: true, truncateThreshold: 40, useProxy: true, proxyExcludedKeys: ['then', 'catch', 'inspect', 'toJSON'] }`
+- **Type:** `{ includeStack?, showDiff?, truncateThreshold? }`
+- **Default:** `{ includeStack: false, showDiff: true, truncateThreshold: 40 }`
 
 Equivalent to [Chai config](https://github.com/chaijs/chai/blob/4.x.x/lib/chai/config.js). 
 
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -1392,7 +1392,7 @@ The number of milliseconds after which a test is considered slow and reported as
 - **Type:** `{ includeStack?, showDiff?, truncateThreshold?, useProxy?, proxyExcludedKeys? }`
 - **Default:** `{ includeStack: false, showDiff: true, truncateThreshold: 40, useProxy: true, proxyExcludedKeys: ['then', 'catch', 'inspect', 'toJSON'] }`
 
-Equivalent to [chai config](https://github.com/chaijs/chai/blob/4.x.x/lib/chai/config.js). You can override it in your `vitest.config.ts` file.
+Equivalent to [Chai config](https://github.com/chaijs/chai/blob/4.x.x/lib/chai/config.js). 
 
 #### chaiConfig.includeStack
 
@@ -1406,7 +1406,7 @@ Influences whether stack trace is included in Assertion error message. Default o
 - **Type:** `boolean`
 - **Default:** `true`
 
-influences whether or not the `showDiff` flag should be included in the thrown AssertionErrors. `false` will always be `false`; `true` will be true when the assertion has requested a diff be shown.
+Influences whether or not the `showDiff` flag should be included in the thrown AssertionErrors. `false` will always be `false`; `true` will be true when the assertion has requested a diff to be shown.
 
 #### chaiConfig.truncateThreshold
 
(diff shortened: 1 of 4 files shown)


docs/config/index.md
```diff
@@ -1386,3 +1386,45 @@ Path to custom tsconfig, relative to the project root.
 - **Default**: `300`
 
 The number of milliseconds after which a test is considered slow and reported as such in the results.
+
+### chaiConfig
+
+- **Type:** `{ includeStack?, showDiff?, truncateThreshold?, useProxy?, proxyExcludedKeys? }`
+- **Default:** `{ includeStack: false, showDiff: true, truncateThreshold: 40, useProxy: true, proxyExcludedKeys: ['then', 'catch', 'inspect', 'toJSON'] }`
+
+Equivalent to [chai config](https://github.com/chaijs/chai/blob/4.x.x/lib/chai/config.js). You can override it in your `vitest.config.ts` file.
```
sheremet-va, line 1395:
```suggestion
Equivalent to [Chai config](https://github.com/chaijs/chai/blob/4.x.x/lib/chai/config.js).
```

This is implied.

docs/config/index.md
```diff
@@ -1386,3 +1386,45 @@ Path to custom tsconfig, relative to the project root.
...
+- **Default:** `false`
+
+Influences whether stack trace is included in Assertion error message. Default of false suppresses stack trace in the error message.
+
+#### chaiConfig.showDiff
+
+- **Type:** `boolean`
+- **Default:** `true`
+
+influences whether or not the `showDiff` flag should be included in the thrown AssertionErrors. `false` will always be `false`; `true` will be true when the assertion has requested a diff be shown.
```
sheremet-va, line 1409:
```suggestion
Influences whether or not the `showDiff` flag should be included in the thrown AssertionErrors. `false` will always be `false`; `true` will be true when the assertion has requested a diff to be shown.
```

docs/config/index.md
```diff
@@ -1386,3 +1386,45 @@ Path to custom tsconfig, relative to the project root.
...
+
+#### chaiConfig.truncateThreshold
+
+- **Type:** `number`
+- **Default:** `40`
+
+Sets length threshold for actual and expected values in assertion errors. If this threshold is exceeded, for example for large data structures, the value is replaced with something like `[ Array(3) ]` or `{ Object (prop1, prop2) }`. Set it to `0` if you want to disable truncating altogether.
```
sheremet-va, line 1416:
```suggestion
Sets length threshold for actual and expected values in assertion errors. If this threshold is exceeded, for example for large data structures, the value is replaced with something like `[ Array(3) ]` or `{ Object (prop1, prop2) }`. Set it to `0` if you want to disable truncating altogether.

This config option affects truncating values in `test.each` titles and inside the assertion error message.
```

docs/config/index.md
```diff
@@ -1386,3 +1386,45 @@ Path to custom tsconfig, relative to the project root.
...
+
+#### chaiConfig.useProxy
+
+- **Type:** `boolean`
+- **Default:** `true`
+
+Defines if chai will use a Proxy to throw an error when a non-existent property is read, which protects users from typos when using property-based assertions.
```
sheremet-va, line 1423: This can be removed, because Vitest requires Proxy to exist, so the config is irrelevant.

docs/config/index.md
```diff
@@ -1386,3 +1386,45 @@ Path to custom tsconfig, relative to the project root.
...
+
+#### chaiConfig.proxyExcludedKeys
```
sheremet-va, line 1425: This can be removed, because Vitest requires Proxy to exist, so the config is irrelevant.

packages/vitest/src/integrations/chai/index.ts
```diff
@@ -78,3 +78,7 @@ Object.defineProperty(globalThis, GLOBAL_EXPECT, {
 
 export { assert, should } from 'chai'
 export { chai, globalExpect as expect }
+export const setupConfig = (config: chaiConfig) => {
```
sheremet-va, line 81:
```suggestion
export const setupChaiConfig = (config: chaiConfig) => {
```

packages/vitest/src/integrations/chai/index.ts
```diff
@@ -78,3 +78,7 @@ Object.defineProperty(globalThis, GLOBAL_EXPECT, {
...
+  Object.assign(chai.config, config)
+}
+export type chaiConfig = Partial<typeof chai.config>
```
sheremet-va, line 84:
```suggestion
export type ChaiConfig = Partial<typeof chai.config>
```

packages/vitest/src/types/config.ts
```diff
@@ -583,6 +584,12 @@ export interface InlineConfig {
    * Requires `singleThread: true` OR `threads: false`.
    */
   inspectBrk?: boolean
+
+  /**
+   * chai config
```
sheremet-va, line 589:
```suggestion
   * Modify default Chai config. Vitest uses Chai for `expect` and `assert` matchers.
```

sheremet-va (changes requested):

docs/config/index.md
btea: Means `useproxy` and `proxyExcludedKeys` two options need to be removed?

sheremet-va: Yes

btea: Done.

sheremet-va (approved):

sheremet-va:
> Looks like #2791  and #2954 can be closed.

If you can add a note to documentation in the test.each section, we can close them. Also in expect.toMatch section I think

btea:
image

Like this ？

sheremet-va:
> image
> 
> Like this ？

Like this

PR #3067, opened 2023-03-23
chore: update snapshot url

btea:


btea:
Hello, you mean that I only need to update locally and don't need to commit to a remote repository, right?

btea:
Ok, I understand, I didn't notice that there are multiple project snapshots that have not been updated before, I will check other tests, if there are still project snapshots that have not been updated, I will commit to this pr.

btea:
The snapshots of all projects should now be updated.

PR #3076, opened 2023-03-24
fix: spy on popup apis like alert

### LINKED ISSUE #3061: bug (browser-runner): alert stales test execution
### Describe the bug

Heyho 👋 

since I discovered this edge case in WebdriverIO I thought I would report it in Vitest as well. If `alert` is called, all JS execution is stopped including the communication with the Vite server.

I am thinking about the following options:
- throw an error and advise users to mock this method
- auto mock it as part of the test setup

What do you think? 

### Reproduction

Given the following test:

```ts
import { test } from 'vitest'

test('haha', () => {
    alert("UOOO")
    console.log('Hello World')
})
```

The console command is never executed.

### System Info

```shell
System:
    OS: macOS 12.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 711.84 MB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Yarn: 1.22.19 - ~/.nvm/versions/node/v16.18.1/bin/yarn
  Browsers:
    Brave Browser: 111.1.49.120
    Chrome: 111.0.5563.64
    Firefox: 109.0
    Firefox Nightly: 111.0a1
    Safari: 15.2
```

### Used Package Manager

npm

### Validations

Aslemammad:
Resolves #3061 

We try to spy the `alert`, `confirm` and `prompt` so we do not let the browser stop and not send any event to vitest. We return null and show a warning to the user, so it'd be better if the user mock them.


packages/browser/src/client/popup.ts
```diff
@@ -0,0 +1,22 @@
...
+${name}(${formatedParams})
+expect(${name}).toHaveBeenCalled()
+\`\`\``)
+  return null
+}
+
+export const setupPopupSpy = () => {
+  globalThis.alert = showPopupWarning('alert')
+  // @ts-expect-error mocking native popup apis
+  globalThis.confirm = showPopupWarning('confirm')
```
sheremet-va, line 20: `confirm` always returns `boolean`, so you should update the return value here instead of suppressing typescript error

packages/browser/src/client/popup.ts
```diff
@@ -0,0 +1,22 @@
+const showPopupWarning = (name: string) => (...params: any[]) => {
+  const formatedParams = params.map(p => JSON.stringify(p)).join(', ')
+
+  console.warn(`Vitest encountered a \`${name}\` call with parameters: ${
+      formatedParams}. Vitest cannot handle \`${
+      name}\` by default in the browser, so it returned null. If you need to handle this case, you can mock the \`${
+      name}\` call yourself like this:
+
+\`\`\`
+vi.spyOn(window, '${name}').mockReturnValue('your value')
```
sheremet-va, line 10: The text was just an example. We also need to import `vi` from `vitest` here to show where it comes from to the new users.

I also don't think using "`" does anything here for the preview. Do you have pictures how it looks like in browser and in the terminal? Can we improve the look somehow?

sheremet-va (changes requested):
I also assumed we will have a page in documentation for this types of errors that we can also reference to? Or at least have a link to mocking section?

packages/browser/src/client/popup.ts
Aslemammad: image

packages/browser/src/client/popup.ts
Aslemammad: Yea, tried that first, but thought returning false may be wrong. Let me change that, I think it's better.

packages/browser/src/client/popup.ts
Aslemammad: Yes, I'll update it. but around the `, I think they separate the text from the code properly.

image

Aslemammad: Here's the new version, I also added some docs.

image

Aslemammad: image

packages/browser/src/client/dialog.ts
```diff
@@ -0,0 +1,21 @@
+const showPopupWarning = <T>(name: string, value: T, defaultValue?: T) => (...params: any[]) => {
+  const formatedParams = params.map(p => JSON.stringify(p)).join(', ')
+
+  console.warn(`Vitest encountered a \`${name}(${formatedParams})\` call that it cannot handle by default, so it returned \`${value}\`. Read more in https://vitest.dev/guide/browser#thread-blocking-dialogs.
```
Aslemammad, line 4: @sheremet-va You also need to update the tests for the new text! I can do it but I'm busy with the playwright integration right now.

charlieforward9:
Any chance we can get `window.print` added here quickly? I am waiting on a [`vi.mock` bug](https://github.com/vitest-dev/vitest/issues/7290) to get fixed and could use this default implementation for the time being.

PR #3078, opened 2023-03-25
fix: avoid call stack recursion with large error (fix: #3060)

### LINKED ISSUE #3060: RangeError: Maximum call stack size exceeded - when there are large outputs
### Describe the bug

When there are very large outputs, picocolors causes a stack overflow.

This is reproducible with this stackblitz: [stackblitz](https://stackblitz.com/edit/vitejs-vite-qv6ccg?file=package.json,vitest.config.ts,src%2Findex.test.jsx,src%2Fgrid-example.jsx&terminal=dev) [open, run npm install, then npm run test]

Note the following changes I've made in this stackblitz in order to get a huge output to trigger this:
1. I'm importing a large external library (ag-grid) that I know causes massive DOMs
2. I'm using react-testing-library
3. I've set the DEBUG_PRINT_LIMIT env var to 100000.

This is a common use case because react-testing-library usually restricts its output and it's hard to see where in the DOM an issue exists.

However, in this case, the use of picocolors in displaying the error causes a stack overflow:
...
 ❯ replaceClose node_modules/picocolors/picocolors.js:25:30
 ❯ replaceClose node_modules/picocolors/picocolors.js:25:30
 ❯ replaceClose node_modules/picocolors/picocolors.js:25:30
 ❯ replaceClose node_modules/picocolors/picocolors.js:25:30
 ❯ Object.red node_modules/picocolors/picocolors.js:17:13
 ❯ printErrorMessage node_modules/vitest/dist/chunk-node-pkg.f9696901.js:13922:18
 ❯ printError node_modules/vitest/dist/chunk-node-pkg.f9696901.js:13811:3
 ❯ Logger.printError node_modules/vitest/dist/chunk-node-pkg.f9696901.js:14026:12
 ❯ DefaultReporter.printTaskErrors node_modules/vitest/dist/chunk-node-pkg.f9696901.js:11166:29
 ❯ DefaultReporter.printErrorsSummary node_modules/vitest/dist/chunk-node-pkg.f9696901.js:11111:18
 ❯ DefaultReporter.reportSummary node_modules/vitest/dist/chunk-node-pkg.f9696901.js:11037:16
 ❯ DefaultReporter.onFinished node_modules/vitest/dist/chunk-node-pkg.f9696901.js:10898:16
 ❯ DefaultReporter.onFinished node_modules/vitest/dist/chunk-node-pkg.f9696901.js:12212:17
 ❯ Vitest.report node_modules/vitest/dist/chunk-node-pkg.f9696901.js:15571:5
 ❯ async file:/C:/Users/E586213/git/meiji.ronin/node_modules/vitest/dist/chunk-node-pkg.f9696901.js:15372:7
 ❯ Vitest.runFiles node_modules/vitest/dist/chunk-node-pkg.f9696901.js:15375:12
 ❯ Vitest.start node_modules/vitest/dist/chunk-node-pkg.f9696901.js:15291:5
 ❯ startVitest node_modules/vitest/dist/chunk-node-pkg.f9696901.js:22440:5
 ❯ start node_modules/vitest/dist/cli.js:105:17

### Reproduction

This is reproducible with this stackblitz: [stackblitz](https://stackblitz.com/edit/vitejs-vite-qv6ccg?file=package.json,vitest.config.ts,src%2Findex.test.jsx,src%2Fgrid-example.jsx&terminal=dev) [open, run npm install, then npm run test]

### System Info

```shell
Reproduced on stackblitz:


  System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    vitest: ^0.29.2 => 0.29.7 

But I originally found this issue on a windows machine with similar values as above (yarn 1, node16, vitest 0.29.7)
```

### Used Package Manager

npm

### Validations

nathanmmiller:
Fixes #3060

Full details in original issue. I have tested this with the same code used therein - I did not include that code in here because it relies on too many external libraries to make such a massive error output.


nathanmmiller:
@sheremet-va are you the right person to review this? I see you tagged my original issue. Am keen to close this out as it's preventing me from moving to vitest, which I'd love to do.

nathanmmiller:
> > @sheremet-va are you the right person to review this? I see you tagged my original issue. Am keen to close this out as it's preventing me from moving to vitest, which I'd love to do.
> 
> I don't really understand how this fixes it? Why do we need an `if`? Why can't we always do what is done for `5000` characters?

You mean just replace the current implementation with my change for <5000? I'm happy to do so if you'd prefer that. I assumed (maybe wrongly) there was a reason why the code c.red() the entire error message instead of just the FAIL part. 

It seems to me that logger.error(`${c.red(c.bold(errorName))}: ${error.message}`) is the better code for all instances instead of an if, but I didn't want to just get rid of the existing implementation if there was a good reason for it. Happy to do so though! Lemme know and I can re-raise a PR with that change.

nathanmmiller:
> Ah, now I see 🤦🏻
> 
> I don't know then 🤔 Maybe it's possible to fix the recursion itself? Maybe use tail recursion 🤔

I don't think fixing the underlying recursion will be possible without a change to picocolors, not vitest. The real issue here is that RTL also colors its output - so we end up running a color-fied output through c.red() which then fills the stack with a zillion calls. It seems to only break at >7000 characters (the default output size for RTL actually, so with the default you're usually fine) but I chose 5000 as a safe buffer.

My assumption was:
if output.length > 5000, it's likely that the output comes from something (like RTL) that already has done formatting/coloring, and we shouldn't .red() it anyway.

I understand your hesitance for the if/else - I think if you're not good with that approach I'd rather submit to just not c.red() the error itself at all - and only do the FAIL part - it seems like a "nice to have" (red error details) is not worth a stack overflow (ability to actually see full output from RTL, a very common use case).

Thoughts?

nathanmmiller:
@sheremet-va any tweaks you want me to make? I really do want to use vitest in our project but this issue is a blocker. Thanks!

sheremet-va:
> I don't think fixing the underlying recursion will be possible without a change to picocolors, not vitest

We have a forked version in the monorepo in `utils/src/colors.ts` file. It's used mostly for runtime, not for Node.js logic, but I guess we can set it up here. The only problem is that we need to set it up somewhere, and I'm not sure where.

nathanmmiller:
> > I don't think fixing the underlying recursion will be possible without a change to picocolors, not vitest
> 
> We have a forked version in the monorepo in `utils/src/colors.ts` file. It's used mostly for runtime, not for Node.js logic, but I guess we can set it up here. The only problem is that we need to set it up somewhere, and I'm not sure where.

Is that something I should wait for or is there an action item for me here on this PR?

nathanmmiller:
How about I raise a PR for just the code path with red() only on FAIL and not on the rest of the message (so no if/else) so that we can unblock this issue, and then if there are future changes to be made, we can raise a new issue for them? Would that work @sheremet-va ?

nathanmmiller:
So to be clear @sheremet-va - there's nothing for me to do here and at some point this issue will be closed by a different PR? If so, I shall wait and then migrate our code base once that is done.

sheremet-va:
> So to be clear @sheremet-va - there's nothing for me to do here and at some point this issue will be closed by a different PR? If so, I shall wait and then migrate our code base once that is done.

Tinyrainbow still has the bug. Feel free to fix it there

sheremet-va:
> So to be clear @sheremet-va - there's nothing for me to do here and at some point this issue will be closed by a different PR? If so, I shall wait and then migrate our code base once that is done.

Tinyrainbow still has the bug. Feel free to fix it there

sheremet-va:
> So to be clear @sheremet-va - there's nothing for me to do here and at some point this issue will be closed by a different PR? If so, I shall wait and then migrate our code base once that is done.

Tinyrainbow still has the bug. Feel free to fix it there

nathanmmiller:
@sheremet-va I've played around with tinyrainbow a little and I don't think it's going to be realistically solvable there either - the fundamental problem is that there is - with large pre-colored inputs - a stack overflow no matter what. 

Tail recursion does not solve this issue, it just will kick the can down the road (so an even larger pre-colored input) still will break it) and [most js interpreters can't even optimize](https://kangax.github.io/compat-table/es6/#test-proper_tail_calls_(tail_call_optimisation)) for tail recursion as far as I can tell.

I don't think this is the solution we want, and I think that with [others now facing the same issue](https://github.com/vitest-dev/vitest/issues/3060#issuecomment-1566427664) outside of just RTL, it would be a good idea to use my original PR: a simple test to say "probably color this, but if the input is very large assume it's already colored and don't bother doing so." This would unblock our adoption and solve the issue for other people as well.

I really want to onboard to vitest, but this is a critical blocker for us. I cannot use your library without it being solved.

sheremet-va (approved):

nathanmmiller:
@sheremet-va any way to re-run the CI test that's failing? It does not look to be a real failure:

⎯⎯⎯⎯⎯⎯ Unhandled Error ⎯⎯⎯⎯⎯⎯⎯
# Error: connect ECONNREFUSED 1[27](https://github.com/vitest-dev/vitest/actions/runs/5223659425/jobs/9430830180?pr=3078#step:9:28).0.0.1:65049
#  ❯ TCPConnectWrap.afterConnect [as oncomplete] node:net:1494:16
#

nathanmmiller:
@sheremet-va sorry to bother - when should we anticipate a merge/release? Thanks in advance!

PR #2999, opened 2023-03-10
feat: webdriverio (+ custom providers) integration for browser mode

Aslemammad:
Resolves #2517 

@christian-bromann provided these arguments for migrating to webdriverio:

Migrating to WebDriver has the following advantage:
- user can pick more than just Chrome and Firefox (potentially even testing in mobile environments using additional tooling like Appium)
- using WebDriver provides an environment much closer to reality than using Chrome DevTools (see some investigations I did here https://github.com/webdriverio/component-testing-examples/tree/main/framework-comparisons)
- using WebdriverIO you still can access the Puppeteer API as provided by your original PR
- WebdriverIO (an OpenJS Foundation project) as dependency is safer to use in the long term as it doesn't depend on maintenance funding from a company

So I feel in the long run, webdriverio would be more helpful.

```sh
pnpm vitest --browser 
pnpm vitest --browser=chrome
pnpm vitest --browser=firefox
pnpm vitest --browser=edge
pnpm vitest --browser=safari
```

PR #3079, opened 2023-03-26
feat: playwright as browser provider

### LINKED ISSUE #3056: Implement playwright browser provider

Add built-in support for Playwright alongside the existing webdriver support 

### Suggested solution

Reference https://github.com/vitest-dev/vitest/pull/2999

### Alternative

_No response_

### Additional context

In SvelteKit we offer Vitest and Playwright as defaults that we automatically setup for the user. It would be nice to have a Playwright browser provider to allow usage of Vitest's browser testing feature without needing to additionally download webdriver.

### Validations

Aslemammad:
Resolves #3056 

As mentioned in #2999, it was planned to add a playwright provider so Sveltekit users find it easier to use the browser feature without the need to add another dependency (`webdriverio`) in their project! cc @benmccann

Feel free to sponsor [Me](https://github.com/sponsors/Aslemammad) or the [Vitest org](https://github.com/sponsors/vitest-dev) to keep our work sustainable and make Vitest keep getting better! Thanks

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,7 +111,7 @@ jobs:
     timeout-minutes: 10
 
     env:
-      BROWSER: ${{ matrix.browser.0 }}
+      BROWSER: ${{ matrix.browser[0] }}
     steps:
       - uses: actions/checkout@v3
 
@@ -135,7 +135,7 @@ jobs:
       - name: Test Browser (playwright)
         run: pnpm run browser:playwright:test
         env:
-          BROWSER: ${{ matrix.browser.1 }}
+          BROWSER: ${{ matrix.browser[1] }}
           PLAYWRIGHT: true
 
   test-browser-safari:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,12 +106,12 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        browser: [chrome, firefox, edge]
+        browser: [[chrome, chromium], [firefox, firefox], [edge, chromium]]
 
     timeout-minutes: 10
 
     env:
-      BROWSER: ${{ matrix.browser }}
+      BROWSER: ${{ matrix.browser.0 }}
     steps:
       - uses: actions/checkout@v3
 
@@ -131,6 +131,12 @@ jobs:
 
       - name: Test Browser
         run: pnpm run browser:test
+
+      - name: Test Browser (playwright)
+        run: pnpm run browser:playwright:test
+        env:
+          BROWSER: ${{ matrix.browser.1 }}
+          PLAYWRIGHT: true
 
   test-browser-safari:
     runs-on: macos-latest
@@ -159,3 +165,9 @@ jobs:
 
       - name: Test Browser
         run: sudo pnpm run browser:test
+
+      - name: Test Browser (playwright)
+        run: sudo pnpm run browser:playwright:test
+        env:
+          BROWSER: webkit
+          PLAYWRIGHT: true
--- a/packages/vitest/package.json
+++ b/packages/vitest/package.json
@@ -123,6 +123,9 @@
     "safaridriver": {
       "optional": true
     },
+    "playwright": {
+      "optional": true
+    },
     "@edge-runtime/vm": {
       "optional": true
     }
@@ -186,6 +189,7 @@
     "natural-compare": "^1.4.0",
     "p-limit": "^4.0.0",
     "pkg-types": "^1.0.1",
+    "playwright": "^1.28.0",
     "pretty-format": "^27.5.1",
     "prompts": "^2.4.2",
     "rollup": "^2.79.1",
--- a/packages/vitest/src/node/browser/playwright.ts
+++ b/packages/vitest/src/node/browser/playwright.ts
@@ -1,5 +1,5 @@
 import { promisify } from 'util'
-import type { Browser } from 'webdriverio'
+import type { Page } from 'playwright'
 import type { Awaitable } from '@vitest/utils'
 // @ts-expect-error doesn't have types
 import detectBrowser from 'x-default-browser'
@@ -9,11 +9,10 @@ import type { BrowserProvider } from '../../types/browser'
 import { ensurePackageInstalled } from '../pkg'
 import type { Vitest } from '../core'
 
-export class WebdriverBrowserProvider implements BrowserProvider {
-  private supportedBrowsers = ['firefox', 'chrome', 'edge', 'safari'] as const
-  private cachedBrowser: Browser | null = null
+export class PlaywrightBrowserProvider implements BrowserProvider {
+  private supportedBrowsers = ['firefox', 'webkit', 'chromium'] as const
+  private cachedBrowser: Page | null = null
   private testDefers = new Map<string, ReturnType<typeof createDefer>>()
-  private stopSafari: () => void = () => {}
   private host = ''
   private browser!: typeof this.supportedBrowsers[number]
   private ctx!: Vitest
@@ -31,13 +30,10 @@ export class WebdriverBrowserProvider implements BrowserProvider {
       throw new Error('Cannot detect browser. Please specify it in the config file.')
 
     if (!this.supportedBrowsers.includes(this.browser))
-      throw new Error(`Webdriver provider does not support this browser, and only supports these browsers: ${this.supportedBrowsers.join(', ')}`)
+      throw new Error(`Playwright provider does not support this browser, and only supports these browsers: ${this.supportedBrowsers.join(', ')}`)
 
-    if (!await ensurePackageInstalled('webdriverio', root))
+    if (!await ensurePackageInstalled('playwright', root))
       throw new Error('Cannot find "webdriverio" package. Please install it manually.')
-
-    if (browser === 'safari' && !await ensurePackageInstalled('safaridriver', root))
-      throw new Error('Cannot find "safaridriver" package. Please install it manually.')
   }
 
   private async resolveBrowserName(): Promise<string> {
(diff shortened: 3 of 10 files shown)


packages/vitest/src/node/browser/playwright.ts
```diff
@@ -0,0 +1,119 @@
...
+  private ctx!: Vitest
+
+  async initialize(ctx: Vitest) {
+    this.ctx = ctx
+    this.host = `http://${ctx.config.browser.api?.host || 'localhost'}:${ctx.browser.config.server.port}`
+
+    const root = this.ctx.config.root
+    const browser = await this.getBrowserName()
+
+    this.browser = browser as any
```
sheremet-va, line 27: The `browser` field in Vitest is shared between providers. If we want to go with this approach, we will need to support all fields. Instead of throwing an error, we need to match found browsers with what playwright supports.

sheremet-va (changes requested):

.github/workflows/ci.yml
```diff
@@ -106,12 +106,12 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        browser: [chrome, firefox, edge]
+        browser: [[chrome, chromium], [firefox, firefox], [edge, chromium]]
```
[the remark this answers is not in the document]
Aslemammad: [Webdriver browser, Playwright browser]! Since I couldn't leave the second item empty, I put it there! Let me know if you got another way! Let me test webkit maybe.

test/browser/specs/runner.test.mjs
```diff
@@ -30,7 +30,7 @@ test('logs are redirected to stdout', async () => {
   assert.match(stdout, /hello from console.debug/, 'prints console.debug')
   assert.match(stdout, /{ hello: 'from dir' }/, 'prints console.dir')
   assert.match(stdout, /{ hello: 'from dirxml' }/, 'prints console.dixml')
-  assert.match(stdout, /hello from console.trace\s+\w+/, 'prints console.trace')
+  assert.match(stdout, /hello from console.trace\s*\w*/, 'prints console.trace')
```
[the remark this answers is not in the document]
Aslemammad: It does not work for safari stack trace as ci shows!

test/browser/package.json
```diff
@@ -4,6 +4,7 @@
   "module": "true",
   "scripts": {
     "test": "node --test specs/",
+    "test:playwright": "PLAYWRIGHT=true node --test specs/",
```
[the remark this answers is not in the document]
Aslemammad: correct, let me fix.

test/browser/specs/runner.test.mjs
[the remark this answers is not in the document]
sheremet-va: it should be fixed then and not just hidden. what is the point in tests when you just remove them if they fail?

packages/vitest/src/node/browser/playwright.ts
Aslemammad: I thought of this, but for instance, the playwright does not support anything that web driver supports. Something like this I had in mind:

`chrome`: `chromium`
`firefox`: `firefox`
`safari`: `webkit`

But as I said, it does not support everything (like the edge option), and also the chromium is not the same as chrome, or webkit is not the same as safari, so it'd be a DX problem in that case. I belive throwing error and mentioning the possible options in the docs is a better approach. Let me know.

sheremet-va: If you cannot support single `browser` option, then it should be defined separately for each provider (with a new option). Otherwise the type you defined are not actually true.

Aslemammad: Do you mean a new option like `playwrightBrowser`? I'm not a fan of this and not convinced actually, since I prefer the unified way of doing it in the `browser`! 
But if the type can be changed, I would go with that way. Let me know what types are desired instead!

Aslemammad: And at the end we need to filter or verify the browser option with throwing an error, since the users should expect that for each provider, not every browser is supported.

sheremet-va: > Do you mean a new option like `playwrightBrowser`?

No, not `playwrightBrowser`, but `providerOptions.browser`.

> But if the type can be changed, I would go with that way. Let me know what types are desired instead!

We already support different providers for different features. You can just take a look at how it's implemented in coverage types.

.github/workflows/ci.yml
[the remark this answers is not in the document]
sheremet-va: I don't know how CI works, but I would prefer it to be fixed _somehow_. I don't know _why_ you cannot provide an empty option.

packages/vitest/src/node/browser/playwright.ts
sheremet-va: Also make sure you know what string `x-default-browser` returns. It might not be aligned with what playwright uses.

Aslemammad: Ok, I'll fix it! And actually, I'm against using x-default-browser, since it does not work with linux in my system since chrome is named google-chrome-stable and it gives this to vitest and we cannot recognize it properly! And I believe the same can happen to other systems and browsers.

sheremet-va: I'm ok with removing it and making `browser` property required

packages/vitest/package.json
```diff
@@ -123,6 +123,9 @@
     "safaridriver": {
       "optional": true
     },
+    "playwright": {
```
sheremet-va, line 126: Should also be specified in `peerDependencies` with webdriverio and safaridriver (we missed it last time)

sheremet-va (changes requested):

packages/vitest/src/node/browser/playwright.ts
Aslemammad: I did remove the detect browser and made `name` required. But I couldn't figure out the exact typing and it was really complicated to do so! So I reverted it as it was (`string`) and the browser validation remained there.

Aslemammad: Now we've got a problem, since the name should not be required because the user may want to open the link manually in their favourite browser.

Aslemammad: So by making it required, we kind of remove the `--browser` only flag feature!

test/browser/specs/runner.test.mjs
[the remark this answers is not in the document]
Aslemammad: Fixed it, turns out safari logs files in the stack with @ as a prefix. so the \w won't match it! But now it works.

packages/vitest/src/node/browser/playwright.ts
sheremet-va: Yes, we are removing —browser flag. This is we had autodection before.

Aslemammad: I see, I don't think that's the right move though, because me as a user, most of the time I use the --browser plain flag with my favourite browser! I think it's a desired use case for many!

Aslemammad: That's actually why the browser feature was made initially.

sheremet-va: It doesn't work that way anymore.

Aslemammad: How so? And the thing is, that what (used to?) seperates us from other browser test runners where they run their own browser automatically. with that feature, what we basically did is that we bring the goodies of the both worlds which makes vitest a next level browser test runner. I think we need somehow to support that feature that we used to. Any opinion? 

And let me know what's the next step for this PR so we get it done.

sheremet-va: > How so? 

Did you even run it with `--browser` flag, when you removed auto detection?

> And the thing is, that what (used to?) seperates us from other browser test runners where they run their own browser automatically.

It doesn't separate us from anyone.

> with that feature, what we basically did is that we bring the goodies of the both worlds which makes vitest a next level browser test runner. 

Elaborate "goodies of the both worlds" that this feature brings. And what are those two worlds?

> Any opinion?

Opening by link can be implemented as a separate provider when the browser name is not provided.

Aslemammad: > Did you even run it with --browser flag, when you removed auto detection?

Yes, it runs wd with chrome, the thing is I can still navigate to the url with a custom browser and not an automated one, which is good, but I prefer the plain link sometimes.

> It doesn't separate us from anyone.

How so, does Cypress allow you to open/run your tests by navigating to the url in the browser that you're currently using as a daily driver? or any browser in the space, where it does not limit us to the wd options or playwright options? That's one of the benefits of the initial feature, where you can open the URL anywhere you want! Let alone the control it gives, because sometimes I do not want the app to open a browser for me, and I prefer to open myself in my dailydriver browser.
> Elaborate "goodies of the both worlds" that this feature brings. And what are those two worlds?

The cypress/playwright/automated test running space and the controlled one where the user controls everything (like vitest ui). The goodies I'm talking about is that with the wd/playwright integrations, and the initial feature, Vitest will have the both worlds I mentioned! That's what I believe would seperate us (+ used to seperate us).

> Opening by link can be implemented as a separate provider when the browser name is not provided.

That's good to know, but I expected the integrations to be built upon that, because this feature is simply a vitest browser server that is waiting for a request from the user as it used to be. I'll create an issue for this soon, so we can also implement it. 


Anyways, let me make the name required and discuss things in another issue later.

packages/vitest/src/defaults.ts
```diff
@@ -42,6 +42,7 @@ export const coverageConfigDefaults: ResolvedCoverageOptions = {
 export const browserConfigDefaults = {
   enabled: false,
   headless: isCI,
+  name: 'chrome',
```
Aslemammad, line 45: @sheremet-va Any way I can avoid this? If I don't put it the type complains (which is expected kinda)

packages/vitest/src/node/browser/playwright.ts
Aslemammad: > Opening your tests in any browser is a very bad idea because it doesn't isolate your cookies, extensions, and any configuration.

Correct, but not a very bad idea, the user should be able to run it everywhere I believe, and for me, it has a huge DX benefit, when I run it in my own browser, I know the risks, but still, I prefer it. It's like `--isolate` option is good, but sometimes the user prefers to not have it at all even with the risks, so we have `--no-isolate`.

> I don't think this feature is needed and can only harm you by creating inconsistencies with your CI setup.

For me, it's not always the CI. Sometimes I just want to have a tab alongside other tabs and test my code easily. In the end, the user might (or not) want to run the tests there (CI). If it was always about the CI, then can the user access vitest ui there? what's the point then of this package? What I'm trying to say is that the CI is not the point of this feature.
And the thing is, it's not a new feature, that was how vitest browser used to work and still, this is how the integrations work are built. it's just not straightly available to the user.

> I still don't understand what you mean. What is controlled and why is it not controlled in other test runners?

I do not know actually how to explain this, but I wanted to say I'd like to have a Cypress (browser native test runner) and a Vitest ui (something you open the tab by yourself) in vitest browser at the same time. I do not have any other way to explain it. But You're currently discussing my points correctly, so I guess it's good to stick to the arguments above.

PR #3103, opened 2023-03-29
feat!: add workspace support

### LINKED ISSUE #256: Monorepo first class support

sheremet-va:
Workspace (or "projects") is defined inside `vitest.workspace`, or `vitest.projects` file (supports js/ts/json files):

```ts
import { defineWorkspace } from 'vitest/config'

  'packages/*'
])
```

This file should be in the same directory as the root config file.

Inline projects are also allowed:

```ts
import { defineWorkspace } from 'vitest/config'

  // you can use both at the same time
  'packages/*',
  // you can even run the same tests, but with different configs in the same "vitest" process
  {
    test: {
      name: 'jsdom',
      root: './shared_tests',
      environment: 'jsdom',
      setupFiles: ['./setup.jsdom.ts'],
    },
  },
  {
    test: {
      name: 'node',
      root: './shared_tests',
      environment: 'node',
      setupFiles: ['./setup.node.ts'],
    },
  },
])
```

The top-level config will not run tests, if it's not specified in the workspaces folder as `<rootDir>` or `./*`.

> ⚠️ Warning
> Project configs **do not** inherit properties from the root config. But some CLI arguments **override** every project's config (most notable, `threads`, `isolate`, `mode`, `sequence`, `allowOnly`, `logHeapUsage`, `testTimeout`).

What you cannot do per project:

- use custom reporters (and anything reporter-related), all reporters are defined in the root config file
- use another coverage provider, there should be only one in the root config file 

Everything else can be configured per project. It is advised to use `defineProject` instead of `defineConfig` for these config files for better type hinting.

This release will include quite a few breaking changes to the public Vitest API due to its nature:

- Some methods on `Vitest` instance were moved to `WorkspaceProject`. You can get the list of projects via `Vitest.prototype.projects`. Removed methods:
  - `initBrowserServer`
  - `globFiles`
  - `globAllTestFiles`
  - `filterFiles`
  - `isBrowserEnabled`

- Some methods on Vitest instance were made `private`:
  - `isTargetFile`

- Some new methods
  - `getProjectsByTestFile` returns the list of projects for a given test file. Usually, return a single project, but it is allowed for different projects to include the same set of files
  - `getModuleProjects` returns the list of projects for any file that was referenced during test execution. Also works for test files, but `getProjectsByTestFile` is faster for this use-case
  - `globTestFiles` now returns a list of files in a tuple: `[WorkspaceProject, testFile: string]`
  - `filterTestsBySource` now expects a list of files in a tuple: `[WorkspaceProject, testFile: string]`
 - `TestSequencer` now expects a list of files in a tuple: `[WorkspaceProject, testFile: string]`

Closes #256

- [x] Coverage support
- [x] Tests
- [x] Documentation


sheremet-va:
@antfu PR should be ready for review if you are interested, I will write more documentation tomorrow, and hopefully, release a new minor.

antfu:
For high-level API design and changes listed in the description - they all make sense to me 👍

I will try to find time to review the implementation tomorrow or so. Or if I failed to make it, feel free to proceed without it. We could improve as we go.

antfu (approved):
Looks solid, incredible work! 🔥

We could have it when you think it's ready :)

dammy001:
@sheremet-va let me review the docs and see if there's a need for any addition

sheremet-va:
> @sheremet-va let me review the docs and see if there's a need for any addition

Sure.

sheremet-va:
> @sheremet-va let me review the docs and see if there's a need for any addition

I will merge the PR now. Feel free to open a separate PR with wording changes.

PR #3104, opened 2023-03-29
docs: add the motivation section for more context

Aslemammad:
In this PR we add the motivation section so users understand the reason behind the browser mode feature and why it's different than other environemnts.

### DIFF
--- a/docs/guide/browser.md
+++ b/docs/guide/browser.md
@@ -102,6 +102,3 @@ In this case, Vitest will run in headless mode using the Chrome browser.
 When using Vitest Browser, it's important to note that thread blocking dialogs like `alert` or `confirm` cannot be used natively. This is because they block the web page, which means Vitest cannot continue communicating with the page, causing the execution to hang.
 
 In such situations, Vitest provides default mocks with default returned values for these APIs. This ensures that if the user accidentally uses synchronous popup web APIs, the execution would not hang. However, it's still recommended for the user to mock these web APIs for better experience. Read more in [Mocking](/guide/mocking).
-
-### Mocking
-Mocking modules is not currently supported in the browser mode, but we are trying our best to deliver this feature and bring the support.
--- a/docs/guide/browser.md
+++ b/docs/guide/browser.md
@@ -6,6 +6,20 @@ title: Browser Mode | Guide
 
 This page provides information about the experimental browser mode feature in the Vitest API, which allows you to run your tests in the browser natively, providing access to browser globals like window and document. This feature is currently under development, and APIs may change in the future.
 
+## Motivation
+
+We developed the Vitest browser mode feature to help improve testing workflows and achieve more accurate and reliable test results. This experimental addition to our testing API allows developers to run tests in a native browser environment. In this section, we'll explore the motivations behind this feature and its benefits for testing.
+
+### Different ways of Testing
+
+There are different ways to test JavaScript code, and each testing approach has its own advantages and disadvantages. Some testing frameworks simulate browser environments in Node.js, while others run tests in real browsers. In this context, jsdom is an example of a testing library that simulates a browser environment, while Cypress is an end-to-end testing tool that allows developers to test their applications in a real browser. 
+
+### The Main caveat
+
+Testing in simulated environments like jsdom or happy-dom have made it easier to test JavaScript programs and provide some level of confidence in test results, making them suitable for many projects. However, it's important to remember that these tools are only simulating a browser environment, so we are not testing in an actual browser. This means that there may still be some discrepancies between the simulated environment and the real environment, potentially leading to false positives or negatives in test results.
+
+To achieve the highest level of confidence in our tests, it's crucial to test in a real browser environment. This is why we developed the browser mode feature in Vitest, allowing developers to run tests natively in a browser and gain more accurate and reliable test results. With browser-level testing, developers can be more confident that their application will work as intended in a real-world scenario.
+
 ## Configuration
 
 To activate browser mode in your Vitest configuration, you can use the `--browser` flag or set the `browser.enabled` field to `true` in your Vitest configuration file. Here is an example configuration using the browser field:
--- a/docs/guide/browser.md
+++ b/docs/guide/browser.md
@@ -102,3 +102,7 @@ In this case, Vitest will run in headless mode using the Chrome browser.
 When using Vitest Browser, it's important to note that thread blocking dialogs like `alert` or `confirm` cannot be used natively. This is because they block the web page, which means Vitest cannot continue communicating with the page, causing the execution to hang.
 
 In such situations, Vitest provides default mocks with default returned values for these APIs. This ensures that if the user accidentally uses synchronous popup web APIs, the execution would not hang. However, it's still recommended for the user to mock these web APIs for better experience. Read more in [Mocking](/guide/mocking).
+
+### Mocking
+Mocking modules is not currently supported in the browser mode, but we are
+trying our best to deliver this feature and bring the support.
--- a/docs/guide/browser.md
+++ b/docs/guide/browser.md
@@ -22,7 +22,7 @@ To achieve the highest level of confidence in our tests, it's crucial to test in
 
 ## Drawbacks
 
-When using Vitest browser as a testing tool, it is important to consider the following drawbacks:
+When using Vitest browser, it is important to consider the following drawbacks:
 
 ### Early Development
 
--- a/docs/guide/browser.md
+++ b/docs/guide/browser.md
@@ -104,5 +104,4 @@ When using Vitest Browser, it's important to note that thread blocking dialogs l
 In such situations, Vitest provides default mocks with default returned values for these APIs. This ensures that if the user accidentally uses synchronous popup web APIs, the execution would not hang. However, it's still recommended for the user to mock these web APIs for better experience. Read more in [Mocking](/guide/mocking).
 
 ### Mocking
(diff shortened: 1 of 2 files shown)


dammy001 (approved):

docs/guide/browser.md
```diff
@@ -6,6 +6,20 @@ title: Browser Mode | Guide
 
 This page provides information about the experimental browser mode feature in the Vitest API, which allows you to run your tests in the browser natively, providing access to browser globals like window and document. This feature is currently under development, and APIs may change in the future.
 
+## Motivation
```
sheremet-va, line 9: This section describes a lot without saying anything. It feels like it was written for a diploma project or by an AI. It also touches on subjects we already discussed at https://vitest.dev/guide/comparisons.html (where it is detailed even more). If you want to add this section, you should also edit comparisons (which should be updated anyway).

docs/guide/browser.md
```diff
@@ -88,3 +102,7 @@ In this case, Vitest will run in headless mode using the Chrome browser.
 When using Vitest Browser, it's important to note that thread blocking dialogs like `alert` or `confirm` cannot be used natively. This is because they block the web page, which means Vitest cannot continue communicating with the page, causing the execution to hang.
 
 In such situations, Vitest provides default mocks with default returned values for these APIs. This ensures that if the user accidentally uses synchronous popup web APIs, the execution would not hang. However, it's still recommended for the user to mock these web APIs for better experience. Read more in [Mocking](/guide/mocking).
+
+### Mocking
+Mocking modules is not currently supported in the browser mode, but we are
```
sheremet-va, line 107: 1. Don't add manual lines. 
2. I don't think we need to specify every feature we are working on in the documentation. The best way to add this note would be in the API section for `vi.mock`

sheremet-va (changes requested):

docs/guide/browser.md
Aslemammad: I believe this is a guide and we need to take this user to the most relevant section step by step! Which I do this by explaining two famous methods of testing and why the first one can not be enough. Which is exactly the motivation behind implementing this feature. 
On the second part, yes, I agree, I can add some to the comparisons, but I do not know what you exactly want by adding to this section! Let me know what should I also add to that guide.

docs/guide/browser.md
Aslemammad: 1- sure.
2- I believe having a place like a browser mode guide where the user sees many of the details is a good thing, it may be a big guide at some point, but I think it's worth it. Let me know your opinion.
At least that's how I expect the users to refer to the guide.

sheremet-va: No one who is using mocking will open this page instead of the API section. I've been fixing documentation for a year now, and I just know that this is not the place to describe the mocking behavior.

It will take a lot of time if you want to describe all features that are not implemented. If you know that something is not working in the browser mode, add a note in the API section.

docs/guide/browser.md
sheremet-va: > I agree, I can add some to the comparisons, but I do not know what you exactly want by adding to this section!

Read the comparison and remove points that are no longer valid. The comparison page was written when browser mode was not implemented, so it is very outdated and misleading.

Aslemammad: > I will note that the browser mode is in very early development, and that I would still professionally recommend people augment their Vitest experience with a standalone browser-side test runner.

I'm actually happy to mention this point.

> I would encourage a production implementation

hmm, I agree. I guess that would help so much with making the browser mode ready for more users.

JessicaSachs: @Aslemammad do you want me to land this PR after working together in Discord?

I'm happy to do the comparisons page.

Aslemammad: If yes, then why not! Let's do it.

sheremet-va: > In this guide, we say why we might want to use the browser mode, and not why we do not need the browser mode.

But you are also listing other methods without giving them much credit, and why they might be used. This guide also doesn't mention drawbacks of browser testing (longer initialization, execution should be on par). I just think if you describe this feature, it should be described with all possible caveats mentioned. We don't have any reason to hide them 🤷🏻

docs/guide/browser.md
```diff
@@ -6,6 +6,20 @@ title: Browser Mode | Guide
 
 This page provides information about the experimental browser mode feature in the Vitest API, which allows you to run your tests in the browser natively, providing access to browser globals like window and document. This feature is currently under development, and APIs may change in the future.
 
+## Motivation
+
+We developed the Vitest browser mode feature to help improve testing workflows and achieve more accurate and reliable test results. This experimental addition to our testing API allows developers to run tests in a native browser environment. In this section, we'll explore the motivations behind this feature and its benefits for testing.
+
+### Different ways of Testing
+
+There are different ways to test JavaScript code, and each testing approach has its own advantages and disadvantages. Some testing frameworks simulate browser environments in Node.js, while others run tests in real browsers. In this context, jsdom is an example of a testing library that simulates a browser environment, while Cypress is an end-to-end testing tool that allows developers to test their applications in a real browser.
```
sheremet-va, line 15: jsdom is not a testing library, it is a spec implementation, and it has nothing to do with testing. So it cannot be compared to Cypress, because these are different tools. Jest with JSDOM and Cypress can be compared, for example

docs/guide/browser.md
```diff
@@ -6,6 +6,20 @@ title: Browser Mode | Guide
...
+
+### Different ways of Testing
+
+There are different ways to test JavaScript code, and each testing approach has its own advantages and disadvantages. Some testing frameworks simulate browser environments in Node.js, while others run tests in real browsers. In this context, jsdom is an example of a testing library that simulates a browser environment, while Cypress is an end-to-end testing tool that allows developers to test their applications in a real browser. 
+
+### The Main caveat
+
+Testing in simulated environments like jsdom or happy-dom have made it easier to test JavaScript programs and provide some level of confidence in test results, making them suitable for many projects. However, it's important to remember that these tools are only simulating a browser environment, so we are not testing in an actual browser. This means that there may still be some discrepancies between the simulated environment and the real environment, potentially leading to false positives or negatives in test results.
```
sheremet-va, line 19: > Testing in simulated environments like jsdom or happy-dom have made it easier to test JavaScript programs and provide some level of confidence in test results, making them suitable for many projects

Not sure about this statement. How did it make it easier? I would assume it just allowed testing in Node.js, which is faster to run than a browser, but for a newby this statement doesn't really say much - "why can't I use JSDOM in a browser?"

If we leave it like this, I think we might get an issue from someone to elaborate on this. And I agree with this, I think this is just too vague.

sheremet-va (changes requested):
I understand that "Motivation" section doesn't have to list other ways of testing or whatnot, but I would really like then to have a separate section that goes more deeply into this and describes the drawbacks of this approach and why Vitest only now implements this feature and not went with it from the beginning.

I am confident that it should be written with this PR, otherwise it will be left as is.

docs/guide/browser.md
Aslemammad: I believe your second batch of requested changes is much more helpful than the previous ones! Let me work on what you actually say now, they are much better than the vague replies.

docs/guide/browser.md
```diff
@@ -6,6 +6,32 @@ title: Browser Mode | Guide
 
 This page provides information about the experimental browser mode feature in the Vitest API, which allows you to run your tests in the browser natively, providing access to browser globals like window and document. This feature is currently under development, and APIs may change in the future.
 
+## Motivation
+
+We developed the Vitest browser mode feature to help improve testing workflows and achieve more accurate and reliable test results. This experimental addition to our testing API allows developers to run tests in a native browser environment. In this section, we'll explore the motivations behind this feature and its benefits for testing.
+
+### Different ways of Testing
+
+There are different ways to test JavaScript code. Some testing frameworks simulate browser environments in Node.js, while others run tests in real browsers. In this context, jsdom is an example of a spec implementation that simulates a browser environment by being used with a test runner like Jest or Vitest, while Cypress is an end-to-end testing tool that allows developers to test their applications in a real browser.
```
christian-bromann, line 15: Is there a reason we have to name Cypress here? There are so many e2e testing frameworks out so I would suggest to either not mentioned a single tool or list a bunch of them. Note that Cypress for example is using WebKit and not Safari which goes against the point of running in a real browser.

docs/guide/browser.md
```diff
@@ -6,6 +6,32 @@ title: Browser Mode | Guide
...
+
+To achieve the highest level of confidence in our tests, it's crucial to test in a real browser environment. This is why we developed the browser mode feature in Vitest, allowing developers to run tests natively in a browser and gain more accurate and reliable test results. With browser-level testing, developers can be more confident that their application will work as intended in a real-world scenario.
+
+## Drawbacks
+
+When using Vitest browser, it is important to consider the following drawbacks:
+
+### Early Development
+
+The browser mode feature of Vitest is still in its early stages of development. As such, it may not yet be fully optimized, and there may be some bugs or issues that have not yet been ironed out. It is recommended that users augment their Vitest browser experience with a standalone browser-side test runner like Cypress.
```
christian-bromann, line 29: Same here

docs/guide/browser.md
Aslemammad: Let's mention bunch of them, could you give me the names? or even push to this PR if you want.

Aslemammad:
@christian-bromann Mentioned other tools! So I guess this PR is ready for a merge.

PR #3113, opened 2023-03-31
feat: gzip html reporter's metadata

7rulnik:
Hi! We are trying to use html reporter in our CI and I mentioned that html.meta.json weighs 35mb for us.
So I think it makes sense to gzip it. After compression with gzip level 9 it weighs 3mb.

### DIFF
--- a/packages/ui/client/composables/client/static.ts
+++ b/packages/ui/client/composables/client/static.ts
@@ -2,6 +2,7 @@ import type { BirpcReturn } from 'birpc'
 import type { VitestClient } from '@vitest/ws-client'
 import type { WebSocketHandlers } from 'vitest/src/api/types'
 import { parse } from 'flatted'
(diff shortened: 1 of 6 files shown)


sheremet-va:
But how would you open it then? There is a note "You can run `npx vite preview --outDir` to see the tests results" in the reporter. It won't work anymore then?

sheremet-va:
> But how would you open it then? 

I see now you unzip it 🤦🏻

7rulnik:
Hmm, interesting. So the reason for failed tests is that `npx vite preview` serves `.gz` files as
```
Content-Type: application/json
Content-Encoding: gzip
```
but for example `npx serve` serves it like
```
Content-Type: application/gzip
```

So I think I can add condition based on content-type. What do you think?

7rulnik:
@sheremet-va I guess it's ready

sheremet-va (approved):

Issue #3004, opened 2023-03-13
Stack trace points to incorrect file on error


DanHulton:
### Describe the bug

When running tests that encounter an error in a required file, the generated trace does not point to the correct file.

Expected:

```
ReferenceError: boop is not defined
 ❯ test/main.test.js:5:12
 > src/main.js:2:9
      1| export function testFunc(x) {
      2|   return boop(x);
       |          ^
      3| }
```

Actual:

```
ReferenceError: boop is not defined
 ❯ Module.testFunc ../../../../../src/main.js:2:3
 ❯ test/main.test.js:5:12
      3| describe('Main test', () => {
      4|   it('should blow up here', () => {
      5|     expect(testFunc(5)).toBe(10);
       |            ^
      6|   });
      7| });
```

### Reproduction

1) Check out https://github.com/DanHulton/vitest-stack-sourcemap-repro
2) `yarn install`
3) `yarn test`

### System Info

```shell
System:
    OS: macOS 13.2.1
    CPU: (8) x64 Apple M1 Pro
    Memory: 19.44 MB / 32.00 GB
    Shell: 3.5.1 - /usr/local/bin/fish
  Binaries:
    Node: 17.3.0 - /usr/local/bin/node
    Yarn: 3.3.1 - /usr/local/bin/yarn
    npm: 8.16.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 110.0.5481.177
    Firefox: 109.0
    Safari: 16.3
  npmPackages:
    vite: latest => 4.1.4
    vitest: latest => 0.29.2
```

### Used Package Manager

yarn

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

obadakhalili:
[...]

PR #3115, opened 2023-04-01
fix: stack trace point to incorrect file (#3004)

ChenKS12138:
# Motivation

- Fixes https://github.com/vitest-dev/vitest/issues/3004

# Test Result

## Before Fix

image

## After Fix

image

### DIFF
--- a/packages/vite-node/src/server.ts
+++ b/packages/vite-node/src/server.ts
@@ -192,7 +192,9 @@ export class ViteNodeServer {
   }
 
   protected async processTransformResult(result: TransformResult) {
-    return withInlineSourcemap(result)
+    return withInlineSourcemap(result, {
+      root: this.server.config.root,
+    })
   }
 
   private async _transformRequest(id: string, customTransformMode?: 'web' | 'ssr') {
--- a/packages/vite-node/src/source-map.ts
+++ b/packages/vite-node/src/source-map.ts
@@ -1,6 +1,7 @@
 import type { TransformResult } from 'vite'
 import type { EncodedSourceMap } from '@jridgewell/trace-mapping'
 import { install } from './source-map-handler'
+import { toFilePath } from './utils'
 
 interface InstallSourceMapSupportOptions {
   getSourceMap: (source: string) => EncodedSourceMap | null | undefined
@@ -13,12 +14,21 @@ const VITE_NODE_SOURCEMAPPING_SOURCE = '//# sourceMappingSource=vite-node'
 const VITE_NODE_SOURCEMAPPING_URL = `${SOURCEMAPPING_URL}=data:application/json;charset=utf-8`
 const VITE_NODE_SOURCEMAPPING_REGEXP = new RegExp(`//# ${VITE_NODE_SOURCEMAPPING_URL};base64,(.+)`)
 
-export function withInlineSourcemap(result: TransformResult) {
+export function withInlineSourcemap(result: TransformResult, options: {
+  root: string // project root path of this resource
+}) {
   const map = result.map
   let code = result.code
 
   if (!map || code.includes(VITE_NODE_SOURCEMAPPING_SOURCE))
     return result
+
+  // sources path from `ViteDevServer` may be not a valid filesystem path (eg. /src/main.js),
+  // so we try to convert them to valid filesystem path
+  map.sources = map.sources.map((one) => {
+    const { exists, path } = toFilePath(one, options.root)
+    return exists ? path : one
+  })
 
   // to reduce the payload size, we only inline vite node source map, because it's also the only one we use
   const OTHER_SOURCE_MAP_REGEXP = new RegExp(`//# ${SOURCEMAPPING_URL}=data:application/json[^,]+base64,(.+)`, 'g')
--- a/packages/vite-node/src/source-map.ts
+++ b/packages/vite-node/src/source-map.ts
@@ -25,9 +25,11 @@ export function withInlineSourcemap(result: TransformResult, options: {
 
   // sources path from `ViteDevServer` may be not a valid filesystem path (eg. /src/main.js),
   // so we try to convert them to valid filesystem path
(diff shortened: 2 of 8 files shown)


sheremet-va:
Stack should already be correct at this stage when we process the error.

ChenKS12138:
> Stack should already be correct at this stage when we process the error.

It this means that `error` which is the params of function `printError` should be corrected before calling function `parseErrorStacktrace`?

Making `stackStr` to be something like  `Module.testFunc (/home/cattchen/codes/github.com/DanHulton/vitest-stack-sourcemap-repro/src/main.js:2:3)`  or  `Module.testFunc (src/main.js:2:3)` , instead of `Module.testFunc (/src/main.js:2:3)`

image

sheremet-va:
> > Stack should already be correct at this stage when we process the error.
> 
> It this means that `error` which is the params of function `printError` should be corrected before calling function `parseErrorStacktrace`?
> 
> Making `stackStr` to be something like `Module.testFunc (/home/cattchen/codes/github.com/DanHulton/vitest-stack-sourcemap-repro/src/main.js:2:3)` or `Module.testFunc (src/main.js:2:3)` , instead of `Module.testFunc (/src/main.js:2:3)`
> 
> image
> 
> # UPDATE:
> It seems that stack has already to be correct at this stage when we process the error in somehow. `/src/main.js` is a valid path for `vite`, but not valid for `node:fs`, it depends on the point of view.
> 
> And We could barely change the `[[FunctionLocation]]` of a variable, which is used to create error stackframe.
> 
> image

Vitest gives full path to modules when processing files, so it should be a fullpath already. If it isn’t, then something is broken sonewhera. This error object is given to other custom reporters, so it should have correct stack trace before that.

ChenKS12138:
> > > Stack should already be correct at this stage when we process the error.
> > 
> > 
> > It this means that `error` which is the params of function `printError` should be corrected before calling function `parseErrorStacktrace`?
> > Making `stackStr` to be something like `Module.testFunc (/home/cattchen/codes/github.com/DanHulton/vitest-stack-sourcemap-repro/src/main.js:2:3)` or `Module.testFunc (src/main.js:2:3)` , instead of `Module.testFunc (/src/main.js:2:3)`
> > image
> > # UPDATE:
> > It seems that stack has already to be correct at this stage when we process the error in somehow. `/src/main.js` is a valid path for `vite`, but not valid for `node:fs`, it depends on the point of view.
> > And We could barely change the `[[FunctionLocation]]` of a variable, which is used to create error stackframe.
> > image
> 
> Vitest gives full path to modules when processing files, so it should be a fullpath already. If it isn’t, then something is broken sonewhera. This error object is given to other custom reporters, so it should have correct stack trace before we print it.

Thanks for your tips! And it seems that function `originalPositionFor` returns the wrong `source`, and it finally leads to prepare wrong data to `Error.prepareStackTrace`.

6f28add6891676a6fdff800adacf11f

packages/vite-node/src/source-map-handler.ts
```diff
@@ -192,8 +192,10 @@ function mapSourcePosition(position: OriginalMapping) {
     // better to give a precise location in the compiled file than a vague
     // location in the original file.
     if (originalPosition.source !== null) {
+      // `originalPositionFor` do not return the expected file source. So use `position.source`
+      // instead of `originalPosition.source`
       originalPosition.source = supportRelativeURL(
-        sourceMap.url, originalPosition.source)
+        sourceMap.url, position.source)
```
sheremet-va, line 198: Doesn't look like this is the fix. `originalPositionFor` should return the correct position. Is something wrong with the map?

sheremet-va (changes requested):

packages/vite-node/src/source-map-handler.ts
ChenKS12138: 1. function `originalPositionFor` works well, but the arguemnt `sourceMap.map` is broken, thus `originalPositionFor ` doesn't return the correct position. It seems that `ViteDevServer` doesn't return the `sources` in absolute path (`/home/cattchen/codes/github.com/DanHulton/vitest-stack-sourcemap-repro/src/main.js` is expected, but finally got `/src/main.js`). I suppose that it's better to use `position.source` than `originalPosition.source`

image

image

2. `sourceMap.url` and `supportRelativeURL` seems work well.

sheremet-va: Let's patch map's `sources` in `packages/vite-node/src/server.ts:195` then. I think `toFilePath` should return the correct path.

ChenKS12138: Thanks for your suggestions! It seems to work well.

image

image

sheremet-va:
Can you add a test case that fails without this patch, but succeeds when the patch is applied?

sheremet-va:
> I have added a test case, and checkout to commits includes `before patch`([d37fa3c](https://github.com/vitest-dev/vitest/commit/d37fa3c75a341db7d135da1d1da9cee7ed9c65b4)) and `after patch`([35a6ef4](https://github.com/vitest-dev/vitest/commit/35a6ef46abd4216efcb2ad024c38676f087be042)) to run test, it works well.
> 
> ```shell
> pnpm build && git rev-parse HEAD && pnpm --filter='./test/vite-node' run test
> ```
> 
> 929eaab2dd80fef1f7094f7c2371f87
> 
> fba969bd40ca31917fa3929a087152a

Yes, but this test doesn’t really test the behavior. Can you add a test where it fails and prints correct stack trace? You can create these conditions in test/stacktrace.

ChenKS12138:
Oh, I see. I have added a new test case in `test/stacktrace`, and it works.

```shell
pnpm build && git rev-parse HEAD && pnpm --filter='./test/stacktraces' run test
```

b8fcc19d7c01430276bf18010974572

1584c99ddcdbc99a78877b1132a63f3

sheremet-va:
Thank you for fixing this!

Issue #2139, opened 2022-10-11
Add `aggregate_failures` option


sheremet-va:
### Clear and concise description of the problem

I think it might be useful to allow aggregating failures inside a single test block.

### Suggested solution

RSpec, Ruby testing tool, has an interesting API called `"aggregate_failures"`. It allows users to group expectations inside a single test instead of aborting the test after the first `expect`. From the [docs](https://relishapp.com/rspec/rspec-core/docs/expectation-framework-integration/aggregating-failures):

> RSpec::Expectations provides [aggregate_failures](https://relishapp.com/rspec/rspec-expectations/docs/aggregating-failures), an API that allows you to group a set of expectations and see all the failures at once, rather than it aborting on the first failure.

> The metadata form is quite convenient, but may not work well for end-to-end tests that have multiple distinct steps. For example, consider a spec for an HTTP client workflow that (1) makes a request, (2) expects a redirect, (3) follows the redirect, and (4) expects a particular response. You probably want the expect(response.status).to be_between(300, 399) expectation to immediately abort if it fails, because you can't perform the next step (following the redirect) if that is not satisfied. For these situations, we encourage you to use the aggregate_failures block form to wrap each set of expectations that represents a distinct step in the test workflow.

This API can also be useful for displaying errors in [type checking context](https://github.com/vitest-dev/vitest/issues/1954), since it's possible to have several type errors inside a single test. Also docs give a lot of scenarios, where this behaviour is useful.

The proposed API is:

```ts
test('name', () => {
  expect(1).toBe(2)
  expect('a').toBe('b')
}, { aggregateFailures: true }) // false by default
```

### Additional context

This API cannot work with 3rd party assertions, since they throw error, different from chai, and cannot be prevented.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

sheremet-va:
Superseded by https://github.com/vitest-dev/vitest/issues/3435

PR #3116, opened 2023-04-01
feat(snapshot): introduce `toMatchFileSnapshot` and auto queuing expect promise

antfu:
For now, both normal snapshot and inline snapshot save the content as a string in JS, which sometimes can be hard to read without syntax highlighting, and with some escape, we need to insert.

`toMatchFileSnapshot` allows storing one snapshot in an explicit file. Which can enable syntax highlighting and language server from the IDE (e.g. you are snapshotting JS codegen, the editor could help you capture syntax errors in your snapshot).

This also supports some colocated test structures, for example `input.txt` and `output.txt` files in each folder.


sheremet-va:
I do like the idea, but I am concerned that it's async. This is the only async matcher that we will have, which will definitely cause confusion and false positive tests. Especially since other snapshot matchers are sync and they also do writing/reading stuff. Some people forget `async` when using `expect().rejects()`... Do you think we can make it sync somehow? For example, we can queue these matchers and do the check when the test function is finished running? Maybe alternatively we can provide the API on a different object? Like `expect.async(file).toMatch...`.

antfu:
Yeah that's what I am worrying about as well. The reason we could make normal snapshot sync is that we know where the snapshot is from the file name and we read it async before the test runs. For here, since the snapshot file path is non-deterministic, we could not pre-read the file. If we queue the async options it could work, but then the `expect` will be non-blocking (and maybe it's not that bad?).

For node, we could use `fs.readFileSync` to make it work, but that will be non-compactible to the browser runner.

Talking about non-blocking, with the similar intention of https://github.com/vitest-dev/vitest/pull/2210, maybe we could have `expect.queued.something()` where we collect the input and execute them with `Promise.allSettled` at the end of the test, which then we could naturally support async and making it explicit that they are non-blocking?

sheremet-va:
> Talking about non-blocking, with the similar intention of #2210, maybe we could have `expect.queued.something()` where we collect the input and execute them with `Promise.allSettled` at the end of the test, which then we could naturally support async and making it explicit that they are non-blocking?

I was thinking about `expect.soft` API like Playwright has: https://playwright.dev/docs/test-assertions#soft-assertions (this would also close #2139)

We can have `toMatchFileSnapshot` exclusively there for now, and extend it to other assertions later (or in the same PR, why not)

antfu:
After a second thought, I do think `.soft` is a good idea that we could have. But having `toMatchFileSnapshot` exclusively on it seems a bit overkill to me, as async is not that unacceptable, and sometimes ppl should have the choice to make it blocking. We also have https://vitest.dev/api/expect.html#resolves which requires(or better to be) await.

So I would prefer to have this feature as-is, and then introduce `.soft` as another feature.

sheremet-va:
> We also have [vitest.dev/api/expect.html#resolves](https://vitest.dev/api/expect.html#resolves) which requires(or better to be) await.
> 
> So I would prefer to have this feature as-is, and then introduce `.soft` as another feature.

Literally, 10 minutes ago someone in the Discord chat had a bug because they forgot to add `await` to it. This will be a big problem if we keep it on `expect`.

sheremet-va:
> Added a mechanism to collect hanging promises and resolve them before the test finished. This should also help the issue of `.resolves` and `.rejects`. So that users could explicitly await to make them blocking, or omit to make them soft.

Not sure how this will help? Vitest now just swallows errors there, no? I don't see any errors being set, and `allSettled` never throws 🤔 Maybe you can add a failing test for this?

antfu:
Yeah, my mistake, I thought `allSettled` would throw automatically. Tests added and now it should work as expected

sheremet-va (approved):

PR #3123, opened 2023-04-03
docs: update extend matcher

bodinsamuel:
Hey team,

- I'm not sure if it was outdated or something else, but the method shown wasn't working. After trying many things, I actually found someone doing that elsewhere on Github.

- A typo on below


sheremet-va:
I think it didn't work because there is no `export {}`, so TypeScript cannot identify it as a module. But your solution seems better.

PR #3137, opened 2023-04-05
build: ui copying to be part of browser's vite build

AriPerkkio:
When using browser feature in local development with `pnpm dev`, the browser package's `pnpm copy` is not run at all. UI in browser is a blank white page as `packages/ui` has not been copied there. 

This PR adds ui copying to be part of `pnpm dev` workflow. It also makes sure monorepo's `pnpm build` command runs `packages/browser` as last step to make sure `packages/ui` has been completely built before copying it. 

Tested manually using locally linked package + `test/browser`.


package.json
```diff
@@ -7,7 +7,7 @@
   "description": "A blazing fast unit test framework powered by Vite",
   "scripts": {
     "ci": "ni && nr typecheck && nr lint && nr build && nr test:all",
-    "build": "pnpm -r --filter='./packages/**' run build && pnpm -C ./packages/browser run copy",
+    "build": "pnpm -r --filter='./packages/**' --filter !browser run build && pnpm --filter browser run build",
```
sheremet-va, line 10: It's already the last one, because it has a UI in dependencies, so this should not be needed

sheremet-va (changes requested):

package.json
AriPerkkio: Great, I'll remove this second step. So it's `pnpm` who knows that `packages/browser` needs to wait for `packages/ui` to finish?

sheremet-va: Yes, it is ordered automatically based on what depends on what

sheremet-va:
While we are here, I noticed some dependencies in the browser should be moved to "dev". From `@vitest/runner` we only need types, so it can be moved. `rollup-plugin-node-polyfills`, `local-pkg`, and `mlly` are not used at all. Can we remove them in this PR, or should we split it?

AriPerkkio:
Let's include it in this PR. I'll go through browser package's dependencies and check how they are used.

sheremet-va:
> @sheremet-va if browser package runs before ui package the browser `dist/client/__vitest__` folder will be empty (I can test this PR on my Windows laptop, current version always build browser before ui package on my local builds).

Please, read the previous conversation: https://github.com/vitest-dev/vitest/pull/3137#discussion_r1158132985

`browser` is always built after the `ui`.

AriPerkkio:
> runs before ui package the browser `dist/client/__vitest__` folder will be empty

There is also a check for this in browser package's build. This will trigger if `packages/browser` is somehow built before `packages/ui`, or that in local environment the developer tries to manually build only the `packages/browser`.

https://github.com/vitest-dev/vitest/blob/b9d6b7ef1cf7292a5bf21fc287687d9306ec3af1/packages/browser/src/client/vite.config.ts#L23-L28

sheremet-va (approved):

PR #3147, opened 2023-04-07
fix: Improve Jest compatibility by widening the return type for tests and hooks to `Assertable<unknown>` from `Assertable<void>` (fix #3146)

### LINKED ISSUE #3146: Types are not actually Jest compatible
### Describe the bug

On the front page of the documentation, it says that it's "Jest Compatible". 

However, the type for a vitest `TestFunction` is `() => Awaitable<void>`. 

But, in Jest the corresponding part of their `ProvidesCallback` type is `() => Promise<unknown>`.

This matters in `beforeEach` and `afterEach`, where you might quite reasonably [call a setup promise which happens to return a value that you don't care about](https://github.com/pact-foundation/pact-js/issues/965#issuecomment-1348185596).

I think there are two possible fixes:

* Widen the type to include `Awaitable<unknown>`
* Update the [migration guide](https://vitest.dev/guide/migration.html#migrating-from-jest) to mention this issue

(side note: I think the note on the front page suggests that it would be fully jest compatible with no changes).

### Reproduction

```
describe("a test that doesn't compile", ()=> {
  it("won't compile",() => {
     return Promise.resolve(2);
  });

  beforeEach(() => {
    // neither will this
    return Promise.resolve("Some setup promise that happens to return a value")
  });
})
```

### System Info

```shell
N/A
```

### Used Package Manager

npm

### Validations

TimothyJones:
This fixes #3146, improving Jest API compatibility by widening the type for hooks and tests to `Assertable<unknown>` from `Assertable<void>`.

I'm new to the source, so I might have misread something. Some comments / callouts:

* Unless I'm misreading it, it seems to me that the `TestFunction` type is actually `Assertable<any>`. I changed the documentation from `Assertable<void>` to `Assertable<unknown>`, but I didn't actually touch the type.
* I also changed the types for `TestEachFunction` and `SuiteEachFunction`. It's possible this is inappropriate.

`pnpm test` passes. However, for me `pnpm test:all` didn't pass before or after this change (but the change didn't affect the way it failed).

### DIFF
--- a/packages/runner/src/types/tasks.ts
+++ b/packages/runner/src/types/tasks.ts
@@ -108,37 +108,37 @@ type ExtractEachCallbackArgs<T extends ReadonlyArray<any>> = {
 interface SuiteEachFunction {
   <T extends any[] | [any]>(cases: ReadonlyArray<T>): (
     name: string,
-    fn: (...args: T) => Awaitable<void>,
+    fn: (...args: T) => Awaitable<unknown>,
   ) => void
   <T extends ReadonlyArray<any>>(cases: ReadonlyArray<T>): (
     name: string,
-    fn: (...args: ExtractEachCallbackArgs<T>) => Awaitable<void>,
+    fn: (...args: ExtractEachCallbackArgs<T>) => Awaitable<unknown>,
   ) => void
   <T>(cases: ReadonlyArray<T>): (
     name: string,
-    fn: (...args: T[]) => Awaitable<void>,
+    fn: (...args: T[]) => Awaitable<unknown>,
   ) => void
 }
 
 interface TestEachFunction {
   <T extends any[] | [any]>(cases: ReadonlyArray<T>): (
     name: string,
-    fn: (...args: T) => Awaitable<void>,
+    fn: (...args: T) => Awaitable<unknown>,
     options?: number | TestOptions,
   ) => void
   <T extends ReadonlyArray<any>>(cases: ReadonlyArray<T>): (
     name: string,
-    fn: (...args: ExtractEachCallbackArgs<T>) => Awaitable<void>,
+    fn: (...args: ExtractEachCallbackArgs<T>) => Awaitable<unknown>,
     options?: number | TestOptions,
   ) => void
   <T>(cases: ReadonlyArray<T>): (
     name: string,
-    fn: (...args: T[]) => Awaitable<void>,
+    fn: (...args: T[]) => Awaitable<unknown>,
     options?: number | TestOptions,
   ) => void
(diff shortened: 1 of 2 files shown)


TimothyJones:
Build failure seems related to timing rather than this PR?

packages/runner/src/types/tasks.ts
```diff
@@ -189,9 +189,9 @@ export type SuiteAPI<ExtraContext = {}> = ChainableSuiteAPI<ExtraContext> & {
   runIf(condition: any): ChainableSuiteAPI<ExtraContext>
 }
 
-export type HookListener<T extends any[], Return = void> = (...args: T) => Awaitable<Return>
+export type HookListener<T extends any[], Return = unknown> = (...args: T) => Awaitable<Return>
 
-export type HookCleanupCallback = (() => Awaitable<unknown>) | void
+export type HookCleanupCallback = (() => Awaitable<unknown>) | unknown
```
sheremet-va, line 194: I am not sure about this one. `unknown` absorbs the callback definition, so it's no longer visible that you need to return a function.

sheremet-va (changes requested):

packages/runner/src/types/tasks.ts
TimothyJones: What would you suggest as an alternative?

sheremet-va:
> Hi team - this review has stalled at a "requests changes" with no requests, and now has conflicts. Do you want it? It would be nice to fix this issue for people who want to port their Jest code.

Vitest behavior is inherently different in this case, so inconsistencies between tools are to be expected. I think every change is fine in this PR except for allowing to return anything inside a hook as I pointed out in the review.

TimothyJones:
> Vitest was never "fully" compatible with jest and will never be. 

Fair enough.

Would you accept a PR that adds this instruction to the migration guide, then? The current instructions to not return from the callback don't cover the case I was trying to fix (which is about returning promises that happen to return values).

I'm not even sure what the expected vitest code would be when that happens. Maybe this?

```
    beforeEach(() =>
      someSetupPromise.then(() => {
        /* empty teardown function for compile reasons */
      }),
    );
```

sheremet-va:
> > Vitest was never "fully" compatible with jest and will never be.
> 
> Fair enough.
> 
> Would you accept a PR that adds this instruction to the migration guide, then? The current instructions to not return from the callback don't cover the case I was trying to fix (which is about returning promises that happen to return values).

Yes, I am open to extending the migration guide.

TimothyJones:
Closing this as it's now conflicting, and sounds like it's not wanted.

Issue #3158, opened 2023-04-09
ReferenceError: AggregateError is not defined on attempting to bump to 0.30.0


trivikr:
### Describe the bug

The following error is thrown when attempting to bump vitest to `v0.30.0`

```console
⎯⎯⎯⎯⎯⎯ Unhandled Error ⎯⎯⎯⎯⎯⎯⎯
ReferenceError: AggregateError is not defined
 ❯ Object.runTests node_modules/vitest/dist/vendor-cli-api.c04eaa34.js:7257:11
 ❯ Object.runTests node_modules/vitest/dist/vendor-cli-api.c04eaa34.js:7396:5
 ❯ async file:/home/runner/work/aws-sdk-js-codemod/aws-sdk-js-codemod/node_modules/vitest/dist/vendor-cli-api.c04eaa34.js:13913:9
 ❯ Vitest.runFiles node_modules/vitest/dist/vendor-cli-api.c04eaa34.js:13927:12
 ❯ Vitest.start node_modules/vitest/dist/vendor-cli-api.c04eaa34.js:13836:5
 ❯ startVitest node_modules/vitest/dist/vendor-cli-api.c04eaa34.js:[20](https://github.com/awslabs/aws-sdk-js-codemod/actions/runs/4652038000/jobs/8232008884?pr=483#step:7:21)784:5
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
 ❯ start node_modules/vitest/dist/cli.js:106:17
```

### Reproduction

GitHub CI failure: https://github.com/awslabs/aws-sdk-js-codemod/actions/runs/4652038000/jobs/8232008884?pr=483

### System Info

The issues happens on macOS as well as GitHub CI.

```shell
  System:
    OS: macOS 13.2.1
    CPU: (8) x64 Apple M1 Pro
    Memory: 40.89 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 14.21.3 - ~/Library/Caches/fnm_multishells/74988_1681066077026/bin/node
    Yarn: 3.5.0 - ~/Library/Caches/fnm_multishells/74799_1681066062748/bin/yarn
    npm: 6.14.18 - ~/Library/Caches/fnm_multishells/74988_1681066077026/bin/npm
[...]

PR #3159, opened 2023-04-10
fix: import performance from perf_hooks in __vitest_worker__

trivikr:
Refs: https://github.com/vitest-dev/vitest/issues/3158


trivikr:
> I don't think it solves #3158 completely

It solves the specific error in facing though.
I changed "Fixed" to "Refs" so that issue with AggregateError is kept open.

Issue #3128, opened 2023-04-03
CLI doesn't currently allow choosing filename test pattern until all tests are completed


ghiscoding:
### Describe the bug

I'm migrating Jest unit tests to Vitest in [Lerna-Lite](https://github.com/lerna-lite/lerna-lite) (a very light fork of Lerna) and I need to use `--no-threads` because a few tests requires the use of `process.chdir()` and because of the no thread, it takes well over a 2min for all tests to complete and during that period the CLI is locked and I cannot use the CLI filename pattern. That's not the case in Jest, I can type in a pattern right after the CLI process starts. I'm on Windows and I know the Lerna/Lerna-Lite tests are running slower on Windows, so the fact that I'm on Windows might be worst for me but still I think the CLI should allow keyboard events even when the unit tests are being performed

### Reproduction

Have a long running unit test suite and while it's running (and not completed), try choosing any CLI options while the test runs and you'll see that you can't. That's also true for all CLI commands, for example you can't Quit, by typing "q", until all unit tests are finished. 

### System Info

```shell
System:
    OS: Windows 10 10.0.19044
    CPU: (12) x64 Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
    Memory: 7.50 GB / 31.75 GB
  Binaries:
    Node: 16.17.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 8.15.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 111.0.5563.147
    Edge: Spartan (44.19041.1266.0), Chromium (111.0.1661.54)
    Internet Explorer: 11.0.19041.1566
  npmPackages:
    @vitejs/plugin-vue: ^4.1.0 => 4.1.0
    @vitest/coverage-c8: ^0.29.8 => 0.29.8
    vite: ^4.2.1 => 4.2.1
    vitest: ^0.29.8 => 0.29.8
```

### Used Package Manager

yarn

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
[...]

Issue #1459, opened 2022-06-09
Bail option to skip all subsequent tests


jgoux:
### Clear and concise description of the problem

Depending on the kind of tests we run, it can be interesting to have a fail-fast feature and exit as soon as a test is failing.

I'm not the only one needing such feature apparently: https://github.com/vitest-dev/vitest/discussions/964

Jest has [bail](https://jestjs.io/fr/docs/cli#--bail) for this purpose.

### Suggested solution

As Vitest tries to smooth the transition from Jest, I think the `bail` option is a sensible name.

**But** Jest implementation is incomplete/weird, it's only skipping the next test suites (describe blocks), but it continues to run the other tests within the same test suite.

I think the default should be to bail by test case and not by test suite.

```
# Skip all the subsequent tests after n test failures (default 1)
--bail <n>
```

I'm not sure what could be the API if we want to deal with both test cases and test suites, maybe adding a `--bail-scope <test|suite>` (default test)? 🤔 

### Additional context

Here are related issues about it:
- https://github.com/facebook/jest/issues/8387
- https://github.com/facebook/jest/issues/6527
- https://github.com/facebook/jest/issues/2867

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

sheremet-va:
Can you explain how you see it? For example, what should happen in those situations:

1. test failed without describe
```ts
test('test 1', () => {/* code that fails */})
test('test 2', () => {/* other test */})
```

2. test failed inside describe
```ts
describe('describe 1', () => {
  test('test 1', () => {/* code that fails */})
  test('test 2', () => {/* other test */})
})

describe('describe 2', () => {
  test('test 1', () => {/* other test */})
  test('test 2', () => {/* other test */})
})
```

3. test failed inside nested describe
```ts
describe('describe 1', () => {
  describe('describe 2', () => {
    test('test 1', () => {/* code that fails */})
    test('test 2', () => {/* other test */})
  })

  test('test 1', () => {/* other test */})
  test('test 2', () => {/* other test */})
})
```

4. test fails in one spec
```ts
// spec1.spec.ts
[...]

PR #3163, opened 2023-04-10
feat(watch): test run cancelling, feat: `--bail` option for cancelling test run

AriPerkkio:
Contains fixes for two separate issues. This PR can be merged with rebase to preserve the commits. 

- Fixes https://github.com/vitest-dev/vitest/issues/3128
- Fixes https://github.com/vitest-dev/vitest/issues/1459

<details>
   <summary>Video: Watch mode cancel</summary>
   
https://user-images.githubusercontent.com/14806298/230854177-9947981a-2aeb-4992-8387-40ea2f08ab36.mov

</details>

<details>
   <summary>Video: --bail option</summary>

https://user-images.githubusercontent.com/14806298/232055152-3293b568-13b8-4ce8-8c4f-d0c751342223.mov

</details>

To test:

- [x] Multithread
- [x] Single thread
- [x] Child process
- [x] Browser

And each combination with
- [x] Isolation on/off
- [x] TTY on/off

### DIFF
--- a/packages/vitest/src/node/pools/threads.ts
+++ b/packages/vitest/src/node/pools/threads.ts
@@ -6,7 +6,7 @@ import { resolve } from 'pathe'
 import type { Options as TinypoolOptions } from 'tinypool'
 import Tinypool from 'tinypool'
 import { distDir } from '../../paths'
-import type { ContextTestEnvironment, ResolvedConfig, RuntimeRPC, Vitest, WorkerContext } from '../../types'
+import type { ContextTestEnvironment, ResolvedConfig, RunnerRPC, RuntimeRPC, Vitest, WorkerContext } from '../../types'
 import type { PoolProcessOptions, ProcessPool, RunWithFiles } from '../pool'
 import { envsOrder, groupFilesByEnv } from '../../utils/test-helpers'
 import { AggregateError, groupBy } from '../../utils/base'
@@ -20,9 +20,10 @@ function createWorkerChannel(project: WorkspaceProject) {
   const port = channel.port2
   const workerPort = channel.port1
 
-  createBirpc<{}, RuntimeRPC>(
+  const rpc = createBirpc<RunnerRPC, RuntimeRPC>(
     createMethodsRPC(project),
     {
+      eventNames: ['onCancel'],
       post(v) {
         port.postMessage(v)
       },
@@ -31,6 +32,8 @@ function createWorkerChannel(project: WorkspaceProject) {
       },
     },
   )
+
+  project.ctx.onCancel(reason => rpc.onCancel(reason))
 
   return { workerPort, port }
 }
@@ -93,6 +96,11 @@ export function createThreadsPool(ctx: Vitest, { execArgv, env }: PoolProcessOpt
         // Worker got stuck and won't terminate - this may cause process to hang
         if (error instanceof Error && /Failed to terminate worker/.test(error.message))
           ctx.state.addProcessTimeoutCause(`Failed to terminate worker while running ${files.join(', ')}.`)
+
+        // Intentionally cancelled
+        else if (ctx.isCancelling && error instanceof Error && /The task has been cancelled/.test(error.message))
+          ctx.state.cancelFiles(files, ctx.config.root)
+
         else
           throw error
       }
@@ -106,6 +114,9 @@ export function createThreadsPool(ctx: Vitest, { execArgv, env }: PoolProcessOpt
     const sequencer = new Sequencer(ctx)
 
     return async (specs, invalidates) => {
+      // Cancel pending tasks from pool when possible
+      ctx.onCancel(() => pool.cancelPendingTasks())
+
       const configs = new Map<WorkspaceProject, ResolvedConfig>()
       const getConfig = (project: WorkspaceProject): ResolvedConfig => {
         if (configs.has(project))
--- a/docs/advanced/runner.md
+++ b/docs/advanced/runner.md
@@ -18,6 +18,13 @@ export interface VitestRunner {
   onCollected?(files: File[]): unknown
 
   /**
+   * Called when test runner should cancel next test runs.
+   * Runner should listen for this method and mark tests and suites as skipped in
+   * "onBeforeRunSuite" and "onBeforeRunTest" when called.
+   */
+  onCancel?(reason: CancelReason): unknown
+
+  /**
    * Called before running a single test. Doesn't have "result" yet.
    */
   onBeforeRunTest?(test: Test): unknown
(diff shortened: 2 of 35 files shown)


packages/runner/src/run.ts
```diff
@@ -105,6 +105,14 @@ async function callCleanupHooks(cleanups: HookCleanupCallback[]) {
 }
 
 export async function runTest(test: Test, runner: VitestRunner) {
+  // @ts-expect-error untyped global
```
sheremet-va, line 108: runner cannot depend on private vitest properties, everything should be provided with the runner class

sheremet-va (changes requested):

packages/runner/src/run.ts
sheremet-va: I don't know the best solution. We can also add a method there specifically for canceling, I suppose.

AriPerkkio: Alright I'm planning a `Runner.onCancel` method that runners can use to decide how they should cancel the runs. This seems to work nice for `worker_threads` and `child_process` pools but I'll need to figure out browser next. 

```ts
class CustomRunner {
  private cancelRun = false;

  /**
   * Called when test runner should cancel next test runs.
   * Runner should listen for this method and mark tests and suites as skipped in
   * "onBeforeRunSuite" and "onBeforeRunTest" when called.
   */
  onCancel() {
    this.cancelRun = true;
  }

  async onBeforeRunTest(test: Test) {
    if (this.cancelRun) test.mode = "skip";
    // ...
  }

  onBeforeRunSuite(suite: Suite) {
    if (this.cancelRun)
      suite.mode = 'skip'
    // ...
  }
}
```

jgoux:
I'm really grateful you're integrating `--bail` into this!

Seeing the demo, is there a possibility to add an option to bail from all the files being tested? So if we test files a, b and c and b is in error, it would stop both a, b and c. This is important to save CI time when we launch everything together. 😁

AriPerkkio:
Yes, that is the current behaviour. 

When a test case fails, the runner will ask main thread the count of currently failed test cases (across all test files) and decides if a bail/cancel should be done. When it sees that enough test cases have failed (`failedCount >= config.bail`):
- runner won't start any new **test cases** from the current test file
- runner informs main thread that bail/cancel should be done
  - main thread won't start any new **test files** in the pool
  - main thread / pool will inform all currently active runners (threads) that they should not start any new **test cases** in their current files
  
In practice the test cancelling means that no new tests should be started. It won't forcefully terminate any on-going test cases as this could cause unexpected side effects. The demo video demonstrates this by using 3000ms slow test cases which might be a bit overkill when highlighting the feature. But it will be identical with Jest, I think.

ghiscoding:
Just to give my point of view in regards to the issue #3128 which I think triggered the opening of this PR (which I'm also highly anticipating). My use case is that I have to use `--no-threads` because I need child process in [Lerna-Lite](https://github.com/lerna-lite/lerna-lite), however with this option it is extremely slow on a Windows platform, it takes 2-3min to complete all tests before I can start using the test pattern of the CLI (by typing "p" in the shell). I didn't have to wait until the end in Jest and that is what I was expecting Vitest to also do (not yet but with this PR it will)... 

long story short, this will be super helpful for long process execution which is especially worst on Windows with `--no-threads`. I also agree with @AriPerkkio that if any tests were started then they should finish their execution but no new tests should start since we interrupted the process which is also what Jest is doing.

out of topic, `--no-threads` is extremely slow on Windows but the equivalent `runInBand` on Jest wasn't that slow. It's so damn slow that I had switch to WSL (Windows Subsystem for Linux) to get back to a decent test time. I did find a few opened issues about this subject but no real answer on to fix this on Windows, any pointer or future PR would be nice.

Cheers and thanks for opening this PR :)

packages/vitest/src/node/pools/threads.ts
```diff
@@ -123,6 +141,12 @@ export function createThreadsPool(ctx: Vitest, { execArgv, env }: PoolProcessOpt
         workspaceMap.set(file, workspaceFiles)
       }
 
+      // TODO: Check and make sure Tinypool cancels tasks gracefully instead of forcefully terminating them.
+      // If not, remove whole signal and rely on runner level cancellation instead.
+      const signal = new EventEmitter()
+      signal.setMaxListeners(ctx.state.getFiles().length)
+      ctx.onCancel(() => signal.emit('abort'))
```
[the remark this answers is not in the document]
Aslemammad: Yes, v0.5.0 published seconds ago.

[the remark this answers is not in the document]
AriPerkkio: Awesome, thanks! 🙌

AriPerkkio:
Ready for review. @sheremet-va as this PR contains two separate features I'll keep the commits ready for rebasing when merging to main - the release changelog will show both features then. 

The `test-browser-safari` CI build failure also happens on main branch. I think it's related to latest `pnpm@8` update. I saw same error locally but fixed it by clearing `packages/*/node_modules`.

sheremet-va (changes requested):
LGTM, but can we add more documentation for the `--bail` feature (and add to CLI args)?

AriPerkkio:
Added flag to CLI argument documentation page.

Also updated the option documentation to describe its use case in more detail. I'm not sure what more to add - writing documentation is hard. 🙃 

https://github.com/vitest-dev/vitest/blob/b046f0c7ac811b36d1c53fe652c5e13637390751/docs/config/index.md?plain=1#L1379-L1387

sheremet-va (approved):

capaj:
the default is not working. I just got this on CI:

```
$ TZ=UTC NODE_ENV=test vitest run --coverage --bail
file:///home/runner/work/secret/graphql/node_modules/cac/dist/index.mjs:413
          throw new CACError(`option \`${option.rawName}\` value is missing`);
                ^

CACError: option `--bail <number>` value is missing
    at Command.checkOptionValue (file:///home/runner/work/secret/graphql/node_modules/cac/dist/index.mjs:413:17)
    at CAC.runMatchedCommand (file:///home/runner/work/secret/graphql/node_modules/cac/dist/index.mjs:599:13)
    at CAC.parse (file:///home/runner/work/secret/graphql/node_modules/cac/dist/index.mjs:537:12)
    at file:///home/runner/work/secret/graphql/node_modules/vitest/dist/cli.js:66:5
    at ModuleJob.run (node:internal/modules/esm/module_job:194:25)

Node.js v18.16.0
```

AriPerkkio:
Default is the value that is used when you don't specify `bail` option at all.

When you wish to change `bail` option, specify a value - exactly as the error message there is instructing.

capaj:
so I need `--bail 1`, ok thanks

gajus:
I don't believe `--bail` works as it is supposed to

https://github.com/vitest-dev/vitest/issues/3372

PR #3185, opened 2023-04-12
docs: improve spyOn example with a contextual example

bencodezen:
As I was reading this example, I noticed that the naming of `obj` is rather generic given that one would expect students to understand what an object is. 

Given that the example talks about apples in a contextual way, the suggestion to replace `obj` with `cart` should help improve comprehension for people analyzing the code example.

Happy to chat more if people have questions!

### DIFF
--- a/docs/api/vi.md
+++ b/docs/api/vi.md
@@ -567,14 +567,14 @@ IntersectionObserver === undefined
 
   ```ts
   let apples = 0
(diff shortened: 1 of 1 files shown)


antfu (approved):

sheremet-va:
Good point! Thank you.

Issue #3129, opened 2023-04-04
Snapshot differs and fail when running all tests vs one test at time


ghiscoding:
### Describe the bug

Migrating from Jest to Vitest (to support ESM), I see that all snapshots created by Vitest have different escaping and fail when running the entire tests but if I run them as a single filename pattern then they always pass (either by filename pattern in the CLI or via the VSCode debugger). When run one file at a time, it always pass and are equivalent to Jest snaphots (escaped with double quotes) but they are very different when running all tests at once (sometime escaped with single quotes & sometime without and sometime with double quotes).

The only thing that I can think of to bypass the problem for now would be to update all snapshots so that it passes in the CI but that is basically inversing the problem, it will then fail when run one at a time. I also think that the one at a time snapshot are the correct ones since they are equivalent to Jest before I started the migration to Vitest. 

### Reproduction

For a reproduction you can do a git checkout of this Lerna-Lite [PR 537](https://github.com/lerna-lite/lerna-lite/pull/537) and run all unit tests (follow [installation](https://github.com/lerna-lite/lerna-lite#installation) instructions). I will merge the PR with the snapshot failures so that it can be inspected and I can continue with my migration to ESM. The only thing that is consistent in all of this is that it always fail when all tests are ran **but** always pass when run one at a time. 

Also worth to know that I need to run Vitest with `--no-threads` since Lerna/Lerna-Lite unit tests are using `process.chdir()` extensively, what I wonder is why it is different when running the test one at a time, technically speaking it is supposed to be isolated because of the `--no-threads` so why does it fail when running all tests? Is there another flag to really isolate each tests? I tried `isolate`, `useAtomic` and `singleThread` but none have any effects and are irrelevant, if I read correctly, when executed with `--no-threads`, so... ?

#### Fails when running all unit tests
![image](https://user-images.githubusercontent.com/643976/229661262-98daea8b-4024-49dd-b463-e34caab4cf23.png)

#### Always pass when running one file at a time
![image](https://user-images.githubusercontent.com/643976/229661301-c97ef9dc-17d4-4031-b737-55c591110e79.png)

### System Info

```shell
System:
    OS: Windows 10 10.0.19045
    CPU: (12) x64 Intel(R) Xeon(R) CPU E5-1650 v2 @ 3.50GHz  
    Memory: 10.87 GB / 31.93 GB
  Binaries:
    Node: 16.20.0 - C:\Program Files\nodejs\node.EXE
[...]

PR #3203, opened 2023-04-17
feat!: support running tests using VM context

sheremet-va:
aka bring all Jest (V8) bugs into Vitest

The idea is to have a separate pool that will spawn workers with tinypool, but they will not be isolated. Isolation is introduced by `vm.createContext` inside the worker. For this, we need to implement our own `import` handling. I think we should keep our `__vitest_ssr_import__` because it is the only way to support both ESM/CJS in the same file in wrongly bundled dependencies.

This is an optional pool! You need to opt-in to use it:

- via CLI argument: `--experimental-vm-threads` (this will throw an error, if you use `--threads` or `--browser` at the same time)
- via config option:
  ```ts
  export default {
    test: {
      experimentalVmThreads: true,
    }
  }
  ```
- run only a subset of files using this pool with `poolMatchGlob`

  ```ts
  export default {
    test: {
      threads: true,
      // all test files will be running like they did before,
      // except for the tests in "fast-leaking-tests" directory 
      poolMatchGlob: [
        ['**/fast-leaking-tests/*.spec.ts', 'experimentalVmThreads'],
      ]
    }
  }
  ```

You also cannot disable isolation when using `experimentalVm`, because this doesn't make sense with this pool. If you don't need isolation, use `threads` or `child_process` directly.

`vm` has a notable memory leak problem, so we are also introducing `experimentalVmWorkerMemoryLimit` option (works like Jest's [workerIdleMemoryLimit](https://jestjs.io/docs/configuration/#workeridlememorylimit-numberstring) to destroy the worker before it reaches this point. It is set to `max_heap/number_of_workers` by default. You might have better performance if you tweak this value manually.

`--experimental-vm-child-process` (`experimentalVmChildProcess`) will be implemented in a separate PR.

Pros:

- Tests go brrrrr (fast)

Cons:

- Vm is not very good with ESM so this introduces a ton of complexity
- We will have this fan-favorite bug: https://github.com/facebook/jest/issues/2549
- And this one: https://github.com/facebook/jest/issues/7874
- Also this bug: https://github.com/nodejs/node/issues/31658
- Basically, all of these issues are now also our issues: https://github.com/nodejs/node/issues/37648

----

Breaking changes:

- the environment is now loaded via the native Node.js mechanism, it should be transformed before the file is passed down to Vitest
- all environments always require `transformMode` (`web` or `ssr`). If transform mode is not specified, it will throw an error
- "VitestExecutor" is moved from "vitest/node" to "vitest/execute" entry point


ghiscoding:
will that help with `--no-threads` (child process) which is extremely slow on Windows? or is that unrelated?

sheremet-va:
> will that help with `--no-threads` (child process) which is extremely slow on Windows? or is that unrelated?

If you need speed, don't use `no-threads`

Current version of this PR uses only workers to run tests in parallel. Will probably add a way to run `child_process` also.

ghiscoding:
> If you need speed, don't use `no-threads`

Yeah I wish but in [Lerna-Lite](https://github.com/lerna-lite/lerna-lite) (a fork of Lerna), it uses a ton of `process.chdir()` and child process and it seems that `--no-threads` is the only way to go about it (as per #1436)
 
> Current version of this PR uses only workers to run tests in parallel. Will probably add a way to run `child_process` also.

that would be really nice and I wonder if that would help with my opened issue #3129 that I still have and found no fix yet (apart from not running the test one at a time)

silverwind:
With this third runner environment option, I wonder if the config should be refactored to replace `threads` with something like `runner: 'worker_threads' | 'child_process' | 'vm'` as I it seems they are mutually exclusive, or can they be combined?

sheremet-va:
> With this third runner environment option, I wonder if the config should be refactored to replace `threads` with something like `runner: 'worker_threads' | 'child_process' | 'vm'` as I it seems they are mutually exclusive, or can they be combined?

Yes, I've been thinking about a similar approach. But technically it should also contain "browser". The list might be:

- `threads`
- `child_process`
- `browser`
- `experimentalVmThreads`
- `experimentalVmChildProcess`

We decided against it for now to keep the same API. The `--experimental-vm-threads --threads` will throw an error.

silverwind:
What's a `browser` environment? Some integration with playwright? 😆

sheremet-va:
> What's a `browser` environment? Some integration with playwright? 😆

https://vitest.dev/guide/browser.html

silverwind:
Thanks, that does sound awesome indeed to be finally able to stop fighting jsdom 👍

packages/vitest/src/runtime/execute.ts
```diff
@@ -32,12 +37,31 @@ let _viteNode: {
 export const moduleCache = new ModuleCacheMap()
 export const mockMap: MockMap = new Map()
 
-export async function startViteNode(ctx: ContextRPC) {
+export async function startViteNode(ctx: ContextRPC, options: ContextExecutorOptions) {
   if (_viteNode)
     return _viteNode
```
AriPerkkio, line 42: Adding a note here while I remember this:

This also caches `rpc` and breaks it when running tests with following conditions:
- `--no-isolate`
- `--single-thread`
- Test files have multiple test environments in use, e.g. `node`, `jsdom`

AriPerkkio: Another founding: Something makes `--no-isolate` stuck even without `--single-thread` and with a single test environment. To reproduce remove this line

https://github.com/sheremet-va/vitest/blob/1b0e6ad5d8cdcf404d8daba01524f1b26111b3ee/test/core/vitest.config.ts#L53-L55

... and run `test/core` tests with `pnpm run test --no-isolate`. Vitest gets stuck after few first initial tests. I have not yet found the root cause but I'm not sure whether it's related to the

sheremet-va: I think it's the same issue. One worker can be used for different tests, but vite node instance is the same.

sheremet-va: Maybe we should just not cache it 🤔

sheremet-va: Currently caching only affects `--no-isolate` 🤔 So we can either disable it or make tinypool smarter to not reuse the same worker for different workspaces. Because different tests will have the same config, but different rpc

sheremet-va: It should run in a separate process if test has different env, env options and or it's in another workspace

AriPerkkio: > But I don't think it's this PR's issue, the same happens in main.

I agree. I can create a bug report for this later. Here's minimal reproduction:

<details>
  <summary>Minimal repro for main branch</summary>

```js
export default defineConfig({
  test: {
    isolate: false,
    singleThread: true,
  },
});
```

```js
/**
 * @vitest-environment node
 */
test("jsdom globals not found", async () => {
  globalThis.__leaking_from_node = "leaking";
  expect(globalThis.__leaking_from_jsdom).toBe(undefined);
});
```

```ts
/**
 * @vitest-environment jsdom
 */
test("Node globals not found", async () => {
  globalThis.__leaking_from_jsdom = "leaking";
  expect(globalThis.__leaking_from_node).toBe(undefined);
});
```

</details>

> So we can either disable it or make tinypool smarter to not reuse the same worker for different workspaces. 

I'm starting to think that we should add support for this in Tinypool. In future we may have other use cases where it would be ideal to use a single pool and be able to recycle a worker - either a new imperative method or through `pool.run` options.

packages/vitest/src/runtime/external-executor.ts
```diff
@@ -0,0 +1,402 @@
...
+      const exports = this.loadCommonJSModule(module, pathUrl)
+      return this.wrapSynteticModule(fileUrl, 'cjs', exports)
+    }
+
+    const inlineCode = this.getIdentifierCode(identifier)
+
+    if (inlineCode || extension === '.mjs')
+      return await this.createEsmModule(fileUrl, inlineCode || await readFile(pathUrl, 'utf8'))
+
+    const pkgData = await this.findNearestPackageData(pathUrl)
```
sheremet-va, line 387: Have a separate worker level cache for this to not call the RPC all the time

packages/vitest/src/runtime/mocker.ts
```diff
@@ -73,7 +81,26 @@ export class VitestMocker {
...
+      return Error
+    return vm.runInContext('Error', context)
+  }
+
+  private createError(message: string) {
+    const Error = this.getErrorConstructor()
+    return new Error(message)
+  }
+
+  private getFinalConstructors(): GlobalConstructors {
```
sheremet-va, line 99: Copy paste the same logic from executor (`this.primitives`), remove `getErrorConstructor`

packages/vitest/src/types/config.ts
```diff
@@ -286,6 +286,24 @@ export interface InlineConfig {
...
+   */
+  experimentalVmThreads?: boolean
+
+  /**
+   * Specifies the memory limit for workers before they are recycled.
+   * If you see your worker leaking memory, try to tinker this value.
+   *
+   * This only has effect on workers that run tests in VM context.
+   */
+  experimentalVmWorkerMemoryLimit?: string | number
```
sheremet-va, line 302: docs

packages/vitest/suppress-warnings.cjs
```diff
@@ -5,6 +5,8 @@ const ignoreWarnings = new Set([
   '--experimental-loader is an experimental feature. This feature could change at any time',
   'Custom ESM Loaders is an experimental feature. This feature could change at any time',
   'Custom ESM Loaders is an experimental feature and might change at any time',
+  'VM Modules is an experimental feature and might change at any time',
+  'VM Modules is an experimental feature. This feature could change at any time',
```
sheremet-va, line 9: Hide `import.meta.resolve` warnings

packages/web-worker/package.json
```diff
@@ -41,7 +41,7 @@
     "typecheck": "tsc --noEmit"
   },
   "peerDependencies": {
-    "vitest": "*"
+    "vitest": ">=0.33.0"
```
sheremet-va, line 44: Bump to 0.34.0

sheremet-va:
@dammy001 can you review docs part?

docs/config/index.md
```diff
@@ -133,9 +133,13 @@ You will not be able to edit your `node_modules` code for debugging, since the c
 #### deps.external
 
 - **Type:** `(string | RegExp)[]`
-- **Default:** `['**/node_modules/**']`
+- **Default:** `[]`
+
+Externalize means that Vite will bypass the package to native Node. Externalized dependencies will not be applied Vite's transformers and resolvers, so they do not support HMR on reload. By default, all packages inside `node_modules` are externalized.
```
dammy001, line 138:
```suggestion
Externalize means that Vite will bypass the package to the native Node. Externalized dependencies will not be applied to Vite's transformers and resolvers, so they do not support HMR on reload. By default, all packages inside `node_modules` are externalized.
```

docs/config/index.md
```diff
@@ -497,6 +502,69 @@ Custom reporters for output. Reporters can be [a Reporter instance](https://gith
...
+:::
+
+### experimentalVmWorkerMemoryLimit
+
+- **Type:** `string | number`
+- **CLI:** `--experimentalVmWorkerMemoryLimit`, `--experimental-vm-worker-memory-limit`
+- **Default:** `1 / CPU Cores`
+- **Version:** Since Vitest 0.34.0
+
+Specifies the memory limit for workers before they are recycled. This value heavily depends on your environment, so it's better to specify it manually instead of relying on default.
```
dammy001, line 542:
```suggestion
Specifies the memory limit for workers before they are recycled. This value heavily depends on your environment, so it's better to specify it manually instead of relying on the default.
```

docs/guide/cli.md
```diff
@@ -69,7 +69,10 @@ Run only [benchmark](https://vitest.dev/guide/features.html#benchmarking-experim
 | `--ui` | Enable UI |
 | `--open` | Open the UI automatically if enabled (default: `true`) |
 | `--api [api]` | Serve API, available options: `--api.port <port>`, `--api.host [host]` and `--api.strictPort` |
-| `--threads` | Enable Threads (default: `true`) |
+| `--threads` | Enable Threads (default: `true`) |   
+| `--single-thread` | Run tests inside a single thread, requires --threads (default: `false`) |
+| `--experimental-vm-threads` | Run tests in a worker pool using VM isolation (default: `false`) |
+| `--experimental-vm-worker-memory-limit` | Set maximum allowed memory for a worker. When reached, new worker will be created instead |
```
dammy001, line 75:
```suggestion
| `--experimental-vm-worker-memory-limit` | Set the maximum allowed memory for a worker. When reached, a new worker will be created instead |
```

packages/vitest/src/integrations/env/edge-runtime.ts
```diff
@@ -5,6 +5,24 @@ import { populateGlobal } from './utils'
...
+        context.Buffer = Buffer
+        return context
+      },
+    })
+    return {
+      getVmContext() {
+        return vm.context
+      },
+      teardown() {
+        // TODO
```
dammy001, line 22: working on this now? @sheremet-va

packages/vitest/src/node/cli.ts
```diff
@@ -23,6 +23,8 @@ cli
   .option('--api [api]', 'Serve API, available options: --api.port <port>, --api.host [host] and --api.strictPort')
   .option('--threads', 'Enabled threads (default: true)')
   .option('--single-thread', 'Run tests inside a single thread, requires --threads (default: false)')
+  .option('--experimental-vm-threads', 'Run tests in a worker pool using VM isolation (default: false)')
+  .option('--experimental-vm-worker-memory-limit', 'Set maximum allowed memory for a worker. When reached, new worker will be created instead')
```
dammy001, line 27:
```suggestion
  .option('--experimental-vm-worker-memory-limit', 'Set the maximum allowed memory for a worker. When reached, a new worker will be created instead')
```

packages/vitest/src/node/cli-api.ts
```diff
@@ -56,6 +56,9 @@ export async function startVitest(
   if (typeof options.browser === 'object' && !('enabled' in options.browser))
     options.browser.enabled = true
 
+  if ('threads' in options && options.experimentalVmThreads)
```
dammy001, line 59: `'threads' in options` checks if the key exists right? How about the value?

sheremet-va: Value doesn't matter, I mention in docs that both threads and no-threads will throw an error

docs/config/index.md
```diff
@@ -497,6 +502,69 @@ Custom reporters for output. Reporters can be [a Reporter instance](https://gith
...
+- **Version:** Since Vitest 0.34.0
+
+Run tests using [VM context](https://nodejs.org/api/vm.html) (inside a sandboxed environment) in a worker pool.
+
+This makes tests run faster, but VM module is unstable when running [ESM code](https://github.com/nodejs/node/issues/37648). Your tests will [leak memory](https://github.com/nodejs/node/issues/33439) - to battle that, consider manually editing [`experimentalVmWorkerMemoryLimit`](#experimentalvmworkermemorylimit) value.
+
+::: warning
+Running code in a sandbox has some advantages (faster tests), but also comes with a number of disadvantages.
+
+- Globals from native modules (`fs`, `path`, etc) are different from the globals in your test environment. It means that error thrown by any of the native modules will be different from the one you are using in your code:
```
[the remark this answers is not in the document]
dammy001: @sheremet-va you're right with the second paragraph. It should be "the ones that were thrown in your code" for clarity.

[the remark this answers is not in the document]
dammy001: "a distinct nature" also means "unique behavior" but it doesn't make any reference so it can confuse. Let's use "unrelated error instances." or "divergent error constructors."?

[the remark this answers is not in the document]
sheremet-va: Can't we use more technical term? This is not a poem. It should be understandable from reading it the first time - error from `fs.readFileSync` won't be instance of Error available in the test context

docs/config/index.md
```diff
@@ -497,6 +502,69 @@ Custom reporters for output. Reporters can be [a Reporter instance](https://gith
...
+Running code in a sandbox has some advantages (faster tests), but also comes with a number of disadvantages.
+
+- The globals within native modules, such as (`fs`, `path`, etc), differ from the globals present in your test environment. As a result, any error thrown by these native modules will reference a different Error constructor compared to the one used in your code:
+
+```ts
+try {
+  fs.writeFile('/doesnt exist')
+}
+catch (err) {
+  consoel.log(err instanceof Error) // false
```
alex1701c, line 525: I am pretty sure this is meant to be `console.log` :)

dammy001: Oh yeah, Thanks

docs/config/index.md
```diff
@@ -500,6 +502,69 @@ Custom reporters for output. Reporters can be [a Reporter instance](https://gith
...
+This makes tests run faster, but the VM module is unstable when running [ESM code](https://github.com/nodejs/node/issues/37648). Your tests will [leak memory](https://github.com/nodejs/node/issues/33439) - to battle that, consider manually editing [`experimentalVmWorkerMemoryLimit`](#experimentalvmworkermemorylimit) value.
+
+::: warning
+Running code in a sandbox has some advantages (faster tests), but also comes with a number of disadvantages.
+
+- The globals within native modules, such as (`fs`, `path`, etc), differ from the globals present in your test environment. As a result, any error thrown by these native modules will reference a different Error constructor compared to the one used in your code:
+
+```ts
+try {
+  fs.writeFile('/doesnt exist')
```
sheremet-va, line 522:
```suggestion
  fs.writeFileSync('/doesnt exist')
```

docs/config/index.md
```diff
@@ -500,6 +502,69 @@ Custom reporters for output. Reporters can be [a Reporter instance](https://gith
...
+}
+catch (err) {
+  console.log(err instanceof Error) // false
+}
+```
+
+- Importing ES modules cache them indefinitely which introduces memory leaks if you have a lot of contexts (or tests). There is no API in Node.js that clears that cache.
```
sheremet-va, line 529:
```suggestion
- Importing ES modules caches them indefinitely which introduces memory leaks if you have a lot of contexts (test files). There is no API in Node.js that clears that cache.
```

packages/vite-node/src/server.ts
```diff
@@ -1,7 +1,7 @@
 import { performance } from 'node:perf_hooks'
 import { existsSync } from 'node:fs'
 import { join, normalize, relative, resolve } from 'pathe'
-import type { TransformResult, ViteDevServer } from 'vite'
+import { type TransformResult, type ViteDevServer } from 'vite'
```
sheremet-va, line 4:
```suggestion
import type { TransformResult, ViteDevServer } from 'vite'
```

PR #3226, opened 2023-04-21
fix(coverage): workspaces c8 source maps

### LINKED ISSUE #3220: Workspaces generate inaccurate c8 coverage reports
### Describe the bug

The C8 test coverage provider reports different statistics and inaccurate line numbers when the reporter runs on a workspace. When I preview the HTML report, the code highlighted is about 10 lines offset from where the actual uncovered code is. The column number offset also appear to shift 1 or 2 characters.

### Reproduction

Vitest hangs when generating coverage in StackBlitz - I will provide a reproduction of this on GitHub when I have a chance.

### System Info

```shell
System:
    OS: macOS 12.6.3
    CPU: (10) arm64 Apple M1 Pro
    Memory: 112.97 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.19.1 - ~/.local/share/homebrew/bin/node
    Yarn: 1.22.19 - ~/.local/share/homebrew/bin/yarn
    npm: 8.19.3 - ~/.local/share/homebrew/bin/npm
  Browsers:
    Firefox: 102.2.0
    Safari: 16.3
  npmPackages:
    @vitest/coverage-c8: ^0.30.1 => 0.30.1 
    vitest: ^0.30.1 => 0.30.1
```

### Used Package Manager

npm

### Validations

AriPerkkio:
- Fixes #3220

### DIFF
--- a/packages/coverage-c8/src/provider.ts
+++ b/packages/coverage-c8/src/provider.ts
@@ -100,8 +100,12 @@ export class C8CoverageProvider extends BaseCoverageProvider implements Coverage
     const sourceMapMeta: Record<SourceMapMeta['url'], MapAndSource> = {}
     const extensions = Array.isArray(this.options.extension) ? this.options.extension : [this.options.extension]
 
+    const fetchCache = this.ctx.projects.map(project =>
+      Array.from(project.vitenode.fetchCache.entries()),
+    ).flat()
+
     const entries = Array
-      .from(this.ctx.vitenode.fetchCache.entries())
+      .from(fetchCache)
       .filter(entry => report._shouldInstrument(entry[0]))
       .map(([file, { result }]) => {
         if (!result.map)
--- a/test/workspaces/coverage-report-tests/check-coverage.test.ts
+++ b/test/workspaces/coverage-report-tests/check-coverage.test.ts
@@ -15,32 +15,8 @@ test('file coverage summary matches', () => {
   const fileCoverage = coverageMap.fileCoverageFor(normalize(resolve('./src/math.ts')))
 
   // There should be 1 uncovered branch and 1 uncovered function. See math.ts.
-  expect(fileCoverage.toSummary()).toMatchInlineSnapshot(`
-    {
-      "branches": {
-        "covered": 3,
-        "pct": 75,
-        "skipped": 0,
-        "total": 4,
-      },
-      "functions": {
-        "covered": 2,
-        "pct": 66.66,
-        "skipped": 0,
-        "total": 3,
-      },
-      "lines": {
-        "covered": 7,
-        "pct": 50,
-        "skipped": 0,
-        "total": 14,
-      },
-      "statements": {
-        "covered": 7,
-        "pct": 50,
-        "skipped": 0,
-        "total": 14,
-      },
-    }
-  `)
+  const { branches, functions } = fileCoverage.toSummary()
+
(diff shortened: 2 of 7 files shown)


packages/coverage-c8/src/provider.ts
```diff
@@ -95,8 +95,12 @@ export class C8CoverageProvider extends BaseCoverageProvider implements Coverage
     const sourceMapMeta: Record<SourceMapMeta['url'], MapAndSource> = {}
     const extensions = Array.isArray(this.options.extension) ? this.options.extension : [this.options.extension]
 
+    const fetchCache = this.ctx.projects.map(project =>
```
sheremet-va, line 98: It's possible that one workspace transforms code not like the other workspace, so there will be several files with the same name, but they might look different (for example, one workspace tests Vue without SSR, and the other Vue with SSR). How does it handle those?

sheremet-va (changes requested):

packages/coverage-c8/src/provider.ts
sheremet-va: > Can you think of any other examples than Vue - something that's easy to set up?

You can create two workspace projects with custom plugins. In the first project, it will return one code, and in the second one - the other.

This can happen for node/jsdom environments for example. "node" uses SSR flag when transforming files. jsdom doesn't use SSR flag.

AriPerkkio: There's now a test case that reproduces this bug. By excluding either `space_1` or `space_3` the final coverage works OK. When both workspaces are included, the coverage of `if (a === 999 && b === 1)` breaks. 

Same bug affects `istanbul` provider as the root cause is in the reporting logic itself.  🙃

AriPerkkio: I think we can get the https://github.com/vitest-dev/vitest/issues/3220 fixed quickly with changes of 66c1b9f27a826d91f479b736d0c6a0ca58a9eeff.

But to support having different kind of code transformations in different workspace projects, and coverage mapped correctly, we would need bigger refactoring. Should that be done in separate PR and land the 66c1b9f27a826d91f479b736d0c6a0ca58a9eeff now?

sheremet-va: I think we can release the fix now, and work on bigger refactoring after that.

test/workspaces/coverage-report-tests/check-coverage.test.ts
```diff
@@ -0,0 +1,46 @@
...
+        "covered": 3,
+        "pct": 75,
+        "skipped": 0,
+        "total": 4,
+      },
+      "functions": {
+        "covered": 2,
+        "pct": 66.66,
+        "skipped": 0,
+        "total": 3,
```
AriPerkkio, line 30: It looks like the function count does not match. Weird that no-one has reported this issue. 
We'll need to add `/* c8 ignore next */` above each `Object.defineProperty(__vite_ssr_exports__, ...`. I'll work on that on separate PR as it's unrelated to this issue.

AriPerkkio: Fixed in https://github.com/vitest-dev/vitest/pull/3241

sheremet-va (approved):

PR #3236, opened 2023-04-23
chore: replace source-map with trace-mapping

sapphi-red:
Vitest currently uses [`source-map@0.6.1` that takes 789kB](https://packagephobia.com/result?p=source-map@0.6.1).
This PR replaces `source-map` with `@jridgewell/trace-mapping`.

`@jridgewell/trace-mapping` is 253kB and Vitest already bundles it because `@ampproject/remapping` depends on it. So this PR will reduce 789kB.
Also [`@jridgewell/trace-mapping` has a better performance than `source-map`](https://github.com/jridgewell/trace-mapping#benchmarks).


antfu (approved):

sheremet-va:
Thank you!

PR #3241, opened 2023-04-24
fix(coverage): c8 to ignore vite's generated helpers

### LINKED ISSUE #3237: `@vitest/coverage-c8` includes generated `__vite_ssr_exports__` helper in coverage
### Describe the bug

When a source file declares named export that is transformed into `__vite_ssr_exports__` by Vite, the `@vitest/coverage-c8` coverage reports increase function count by 1 and consider the `Object.defineProperty(__vite_ssr_exports__, ...` as part of source code. This is due to `Object.defineProperty(__vite_ssr_exports__, ...` being included in the source maps. 

image

### Reproduction

This can be reproduced in Vitest repository's `test/coverage-test` when `c8` is used. All files have function counts that do not match. (https://github.com/vitest-dev/vitest/pull/3226#discussion_r1174016083)

This is also seen when adding an unused `export`. The line that is marked as uncovered contains the `Object.defineProperty(__vite_ssr_exports__, ...` where `get()` is not executed since no one imports that file. This `get() {}` also increases function count by one.

### System Info

```shell
System:
    OS: macOS 13.3.1
    CPU: (8) arm64 Apple M2
    Memory: 56.73 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.14.0 - ~/.nvm/versions/node/v18.14.0/bin/node
    npm: 9.3.1 - ~/.nvm/versions/node/v18.14.0/bin/npm
  Browsers:
    Chrome: 112.0.5615.137
    Safari: 16.4
  npmPackages:
    @vitejs/plugin-vue: latest => 4.1.0 
    @vitest/browser: workspace:* => 0.30.1 
    vite: latest => 4.2.1 
    vitest: workspace:* => 0.30.1
```

### Used Package Manager

pnpm

### Validations

AriPerkkio:
- Fixes #3237

Users may see changes in their coverage summary's `functions` and `branches`. Previously these have included some generated code that should not have ended up in the coverage report. Now the `branch` and `function` counts should correspond the source files.

### DIFF
--- a/packages/coverage-c8/src/provider.ts
+++ b/packages/coverage-c8/src/provider.ts
@@ -1,6 +1,8 @@
 import { existsSync, promises as fs } from 'node:fs'
 import _url from 'node:url'
 import type { Profiler } from 'node:inspector'
+import MagicString from 'magic-string'
+import remapping from '@ampproject/remapping'
 import { extname, resolve } from 'pathe'
 import c from 'picocolors'
 import { provider } from 'std-env'
@@ -17,6 +19,9 @@ import createReport from 'c8/lib/report.js'
 import { checkCoverages } from 'c8/lib/commands/check-coverage.js'
 
 type Options = ResolvedCoverageOptions<'c8'>
+
+// Note that this needs to match the line ending as well
+const VITE_EXPORTS_LINE_PATTERN = /Object\.defineProperty\(__vite_ssr_exports__.*\n/g
 
 export class C8CoverageProvider extends BaseCoverageProvider implements CoverageProvider {
   name = 'c8'
@@ -169,7 +174,7 @@ export class C8CoverageProvider extends BaseCoverageProvider implements Coverage
 
       return {
         sourceMap: {
-          sourcemap: data.map,
+          sourcemap: removeViteHelpersFromSourceMaps(data.source, data.map),
         },
         source: Array(offset).fill('.').join('') + data.source,
       }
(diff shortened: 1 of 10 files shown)


packages/coverage-c8/src/provider.ts
```diff
@@ -193,3 +193,27 @@ export class C8CoverageProvider extends BaseCoverageProvider implements Coverage
...
+  const mappings = map.mappings.split(';')
+  exportIndexes.forEach((index) => {
+    mappings[index] = ''
+  })
+
+  return {
+    ...map,
+    mappings: mappings.join(';'),
+  }
+}
```
AriPerkkio, line 219: This functions looks for lines where `__vite_ssr_exports__` is appended with new getters and removes those lines from the source maps. It's hacky but it shows what should be done.

Any ideas what would the correct way to do this? I don't think `magic-string` supports modifying existing source maps - it simply generates new ones. And I guess there is no way to tell Vite not to include these lines in source maps. 

Also it would be great if we could do this directly in rollup/vite plugins but I don't think Vite exports are available there even in `enforce: 'post'`.

AriPerkkio: I picked only the needed `remapping` from `combineSourcemaps`: https://github.com/ampproject/remapping#multiple-transformations-of-a-file

I checked all files of `test/coverage-test` and it seems to work well. Vue had some issues but eventually got it working as well. Source maps are fun - said no one ever. 🙃

sheremet-va (approved):

Issue #3046, opened 2023-03-20
Implement module mocking in browser mode


christian-bromann:
### Describe the bug

The following test:

```ts
import { test, vi } from 'vitest'
import { v4 } from 'uuid'

vi.mock('uuid', () => ({
    v4: 'I am mocked'
}))

test('I live in the browser', () => {
    console.log('YES', window.navigator.userAgent, v4)
})
```

throws an error:

```
 FAIL  tests/browser.test.js [ tests/browser.test.js ]
TypeError: this._mocker.queueMock is not a function
 ❯ VitestUtils.mock http:/localhost:63315/node_modules/.vite/deps/chunk-DGJ3HNQR.js?v=22cddd7a:10193:18
 ❯ http:/localhost:63315/private/tmp/vitestbrowser/tests/browser.test.js?v=1679345265563:4:4
```

### Reproduction

see above

### System Info

```shell
System:
    OS: macOS 12.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 420.98 MB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 18.7.0 - ~/.nvm/versions/node/v18.7.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v18.7.0/bin/yarn
    npm: 8.15.0 - ~/.nvm/versions/node/v18.7.0/bin/npm
  Browsers:
    Brave Browser: 111.1.49.120
    Chrome: 111.0.5563.64
    Firefox: 109.0
    Firefox Nightly: 111.0a1
    Safari: 15.2
  npmPackages:
    @vitest/browser: ^0.29.7 => 0.29.7
    vitest: ^0.29.7 => 0.29.7
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
[...]

PR #3258, opened 2023-04-25
feat!: update mock implementation to support ESM runtime, introduce "vi.hoisted"

### LINKED ISSUE #3228: Introduce `vi.hoisted` to run code before imports and `vi.mock`

All imports in ESM are hoisted to the top of the file, so any code that is defined before that run _after_ imports and all its side effects. Vitest also has `vi.mock` which behaves in a similar manner, but it is hoisted even before the imports.

Sometimes there are reasons to run code before import happens (for example, to mock time). It also would be nice to have access to variables in `vi.mock` to later reuse them in tests (like setting up a spy).

### Suggested solution

Introduce `vi.hoisted` method. It receives a callback that will run before all the imports and `vi.mock`. You can also return values from this function to access them later.

Examples:

```ts
// mock time before import happens
import { assert } from 'vitest'
import currentTime from './currentTime.ts'

vi.hoisted(() => {
  vi.setSystemTime(new Date(2020, 1, 1, 0, 0, 0))
})

assert.ok(currentTime.getTime() === new Date(2020, 1, 1, 0, 0, 0).getTime())
```

```ts
// setup spies for vi.mock
import { assert } from 'vitest'
import { getTimeFactory } from './getTime.ts'

const mocks = vi.hoisted(() => {
  return { getTime: vi.fn() }
})

vi.mock('./getTime.ts', () => {
  return {
    getTimeFactory: () => mocks.getTime,
  }
})

mocks.getTime.mockReturnValue(0)
const getTime = getTimeFactory()
assert.equals(getTime(), 0)
```

### Alternative

_No response_

### Additional context

_No response_

### Validations

sheremet-va:
Closes #3228

In preparation for #3046

- vi.mock implementation is rewritten to support ESM usage outside of Vitest sandbox in the future
- introduce "vi.hoisted" API to run code before imports are evaluated
- make "vi.dynamicImportSettled" work in browser context
- remove support for "browser" pool in `poolMatchGlobs`. Use workspaces feature to run tests in browser.
- it's now possible to use `vi.spyOn` on an ES module named export in browser environment

For performance reasons I am hoisting `vi.mock` and transform `import` in the same function, so it looks a little bit over-engineered 🤔 If there are better and more performant ways, I am wiling to rewrite this part.


antfu (approved):
I am ok to give it a try :)

antfu:
Love the `vi.hoisted` API btw, I would also find it useful!

jalooc:
Nice one 👍 

Just a little FYI - looks like this poped up under the `v0.30.1` version in the documention:
image

which I believe is not valid, since `vi.hoisted()` was just merged and that version is already a couple of weeks old.

PR #3263, opened 2023-04-26
fix(spy): update to set initial implementation through normal logic (fixes #3260)

### LINKED ISSUE #3260: vi.fn(mockImplementation) does not expose implementation
### Describe the bug

When using `vi.fn()` to create a mock:
```ts
const fnWithMock = vi.fn(() => 1)
```

The provided mock should be obtainable via `getMockImplementation`, but it is not:
```
> fnWithMock.getMockImplementation()
undefined
```

Upon some quick investigation this appears to be because of how the call to `enhanceSpy` from `fn` assumes that no implementation has been provided.

If you set an implementation after the initialisation, then you get the expected behaviour as the setter sets `implementation`.

We could do something like this:
```ts
export function fn<TArgs extends any[] = any[], R = any>(
  implementation?: (...args: TArgs) => R,
): Mock<TArgs, R> {
  const enhancedSpy = enhanceSpy(tinyspy.internalSpyOn({ fn: () => {} }, 'fn'))
  if (implementation)
    enhancedSpy.mockImplementation(implementation)

  return enhancedSpy as Mock
}
``` 

This would prevent breaking how `spyOn` works already, if you're happy with this, I can make a PR.

There are some existing tests in `jest-mock.ts`, however these assert `expect(spy.getMockImplementation()).toBe(undefined)` when it _should_ actually be defined.

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-vuwwus?file=test/fn.test.ts

### System Info

```shell
System:
    OS: macOS 13.3.1
    CPU: (8) arm64 Apple M1
    Memory: 43.36 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 16.19.0 - ~/.asdf/installs/nodejs/lts-gallium/bin/node
    Yarn: 1.22.19 - ~/.asdf/installs/nodejs/lts-gallium/.npm/bin/yarn
    npm: 8.19.3 - ~/.asdf/plugins/nodejs/shims/npm
  Browsers:
    Chrome: 112.0.5615.137
    Firefox Developer Edition: 112.0
    Safari: 16.4
  npmPackages:
    @vitest/browser: workspace:* => 0.30.1 
    @vitest/coverage-c8: workspace:* => 0.30.1 
    @vitest/coverage-istanbul: workspace:* => 0.30.1 
    @vitest/ui: workspace:* => 0.30.1 
    vite: ^4.2.1 => 4.2.1 
    vitest: workspace:* => 0.30.1
```

### Used Package Manager

npm

### Validations

Codex-:
Fixes: #3260

The issue currently with `vi.fn` and providing an initial mock implementation (i.e. `vi.fn(() => true)`) is that the provided implementation does not get set by the `enhanceSpy` wrapping. This means that while `stub` is extended to provide the supporting functions that are added by `enhanceSpy`, the state of `implementation` is left `undefined`.

We could go down the path of adding extra parameters and such to `enhanceSpy`, but if we simply call `mockImplementation` during `vi.fn` we can follow the existing logic and have consistent behaviour, allowing the mocked implementation to follow the behaviour we expect.

I've added a test for both the provided implementation and no-implementation cases, as well as updated the other testing / snapshots to reflect this change.

Would love to hear your thoughts or feedback on this :)

### DIFF
--- a/packages/spy/src/index.ts
+++ b/packages/spy/src/index.ts
@@ -324,5 +324,9 @@ export function fn<TArgs extends any[] = any[], R = any>(
 export function fn<TArgs extends any[] = any[], R = any>(
   implementation?: (...args: TArgs) => R,
 ): Mock<TArgs, R> {
-  return enhanceSpy(tinyspy.internalSpyOn({ fn: implementation || (() => {}) }, 'fn')) as unknown as Mock
+  const enhancedSpy = enhanceSpy(tinyspy.internalSpyOn({ spy: () => {} }, 'spy'))
+  if (implementation)
+    enhancedSpy.mockImplementation(implementation)
+
+  return enhancedSpy as Mock
 }
--- a/test/core/test/jest-mock.test.ts
+++ b/test/core/test/jest-mock.test.ts
@@ -41,6 +41,17 @@ describe('jest mock compat layer', () => {
     expect(Spy.mock.instances).toHaveLength(0)
   })
 
+  it('implementation is set correctly on init', () => {
+    const impl = () => 1
+    const mock1 = vi.fn(impl)
+
+    expect(mock1.getMockImplementation()).toEqual(impl)
+
+    const mock2 = vi.fn()
+
+    expect(mock2.getMockImplementation()).toBeUndefined()
+  })
+
   it('implementation sync fn', () => {
     const originalFn = function () {
       return 'original'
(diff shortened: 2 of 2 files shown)


sheremet-va (changes requested):
Let's rename default `fn` to spy, so it doesn't change snapshots. Otherwise looks good

Codex-:
> Let's rename default `fn` to `spy`, so it doesn't change snapshots. Otherwise looks good

Done :)

sheremet-va (approved):

PR #3274, opened 2023-04-30
fix(vitest): also check for vite relative to vitest package

### LINKED ISSUE #3112: Suggested .vscode/launch.json requires installing vite

Going through the VS Code debugging guide in https://vitest.dev/guide/debugging.html#vscode, when I run _F5_ on a repo that doesn't have `vite` listed as a dependency:

```plaintext
joshgoldberg ~/repos/template-typescript-node-package $  cd /Users/josh/repos/template-typescript-node-package ; /usr/bin/env 'NODE_OPTIONS=--require "/Applications/Visual Studio Code.app/Contents/Resources/app/extensions/ms-vscode.js-debug/src/bootloader
.bundle.js"     --inspect-publish-uid=http' 'VSCODE_INSPECTOR_OPTIONS={"inspectorIpc":"/var/folders/sr/j__6_9gn0dv557z0xf3p7y8h0000gn/T/node-cdp.67533-14.sock","deferredMode":false,"waitForDebugger":"","execPath":"/Users/josh/.nvm/versions/node/v18.7.0/bi
n/node","onlyEntrypoint":false,"autoAttachMode":"always","fileCallback":"/var/folders/sr/j__6_9gn0dv557z0xf3p7y8h0000gn/T/node-debug-callback-00402a5d1b51117e"}' /Users/josh/.nvm/versions/node/v18.7.0/bin/node ./node_modules/vitest/vitest.mjs run src/gree
t.test.ts 
Debugger attached.
 MISSING DEP  Can not find dependency 'vite'

? Do you want to install vite? › (y/N)
```

Screenshot of the first few dozen columns of the previous console log

After accepting with `y` (or running `pnpm install vite` manually), the debug command works.

Reproduction:
* Check out https://github.com/JoshuaKGoldberg/template-typescript-node-package/tree/c122ea1aa2e70ed5ead091b345f2b0a424dcd2bb
* `pnpm install`
* Open [`src/greet.test.ts`](https://github.com/JoshuaKGoldberg/template-typescript-node-package/blob/c122ea1aa2e70ed5ead091b345f2b0a424dcd2bb/src/greet.test.ts)
* Launch the _Debug Current Test File_ command

### Suggested solution

I don't know why this is happening enough to suggest a solution 😞 sorry. `./vitest.mjs` is the `bin` entry in `node_modules/vitest/package.json`.

### Alternative

_No response_

### Additional context

Note that `pnpm test` in that repository works _without_ adding `vitest` as a dependency.

### Validations

JoshuaKGoldberg:
Changes `ensurePackageInstalled` to also check relative to a file within the `vitest` package. That way, packages like `vitest` that maybe installed as nested dependencies can still be resolved by the `require.resolve` inside `isPackageExists`.

This is my first PR in vitest (hello! happy to be here!). Left a couple of questions inline. 😄 

Fixes #3112.

### DIFF
--- a/packages/vitest/src/node/pkg.ts
+++ b/packages/vitest/src/node/pkg.ts
@@ -1,13 +1,16 @@
+import url from 'node:url'
 import c from 'picocolors'
 import { isPackageExists } from 'local-pkg'
 import { EXIT_CODE_RESTART } from '../constants'
 import { isCI } from '../utils/env'
 
+const __dirname = url.fileURLToPath(new URL('.', import.meta.url))
+
 export async function ensurePackageInstalled(
   dependency: string,
   root: string,
 ) {
-  if (isPackageExists(dependency, { paths: [root] }))
+  if (isPackageExists(dependency, { paths: [root, __dirname] }))
     return true
 
   const promptInstall = !isCI && process.stdout.isTTY


packages/vitest/src/node/pkg.ts
```diff
@@ -1,13 +1,16 @@
...
 import { isCI } from '../utils/env'
 
+const __dirname = url.fileURLToPath(new URL('.', import.meta.url))
+
 export async function ensurePackageInstalled(
   dependency: string,
   root: string,
 ) {
-  if (isPackageExists(dependency, { paths: [root] }))
+  if (isPackageExists(dependency, { paths: [root, __dirname] }))
```
JoshuaKGoldberg, line 13: This change means _all_ calls to `ensurePackageInstalled` will also allow packages that are only dependencies of `vitest`, not whoever has installed `vitest`. Is that good?

An alternative could be to allow the `root` param to be `string | string[]`.

packages/vitest/src/node/pkg.ts
```diff
@@ -1,13 +1,16 @@
+import url from 'node:url'
 import c from 'picocolors'
 import { isPackageExists } from 'local-pkg'
 import { EXIT_CODE_RESTART } from '../constants'
 import { isCI } from '../utils/env'
 
+const __dirname = url.fileURLToPath(new URL('.', import.meta.url))
+
 export async function ensurePackageInstalled(
```
JoshuaKGoldberg, line 9: I looked around and couldn't figure out where tests should go for this change, if anywhere. Would you like me to test this in some way?

dammy001: @sheremet-va will look into this.

packages/vitest/src/node/pkg.ts
sheremet-va: This call is needed to check if importing the module will fail. If it's still possible to import (and I assume it is), then it should be fine.

packages/vitest/src/node/pkg.ts
sheremet-va: There are no tests for this feature.

sheremet-va (approved):

PR #3301, opened 2023-05-04
fix: throw an error, if tests are collected with a different vitest version

### LINKED ISSUE #3296: "No test suite found in file" Once again
### Describe the bug

I've seen the error "No test suite found in file" many times over the course of months and it is reported at least twice before without clear resolution or at least a resolution that is helpful for my case:
- #847 
- #2604

sometimes, the error it disappears, just today I though I understood how to solve it but it appeared again. I am not sure why this happens. I was able to lock it down though to always fail with the code provided in the reproduction info block.

### Reproduction

minimal reproduction code:
https://github.com/xania/view/tree/vitest-no-test-suite-found

> pnpm install

image

### System Info

```shell
System:
    OS: Windows 10 10.0.22621
    CPU: (20) x64 12th Gen Intel(R) Core(TM) i9-12900HK
    Memory: 17.43 GB / 31.68 GB
  Binaries:
    Node: 19.3.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 9.6.0 - ~\AppData\Roaming\npm\npm.CMD
  Browsers:
    Chrome: 107.0.5304.122
    Edge: Spartan (44.22621.1555.0), Chromium (112.0.1722.68)
  npmPackages:
    vite: ^4.3.4 => 4.3.4
    vitest: ^0.31.0 => 0.31.0
```

### Used Package Manager

pnpm

### Validations

sheremet-va:
Closes #3296

### DIFF
--- a/packages/runner/rollup.config.js
+++ b/packages/runner/rollup.config.js
@@ -1,5 +1,6 @@
 import { builtinModules } from 'node:module'
 import esbuild from 'rollup-plugin-esbuild'
+import json from '@rollup/plugin-json'
 import dts from 'rollup-plugin-dts'
 import { defineConfig } from 'rollup'
 import pkg from './package.json' assert { type: 'json' }
@@ -21,6 +22,7 @@ const plugins = [
   esbuild({
     target: 'node14',
   }),
+  json(),
 ]
 
 export default defineConfig([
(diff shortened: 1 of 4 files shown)


AriPerkkio (approved):
I don't think there is an easy way to write test case for this yet. Once we have one published release where this version check is in place, we can write a test case for this by installing the older version in a monorepo fixture setup.

PR #3305, opened 2023-05-05
feat(watch): press `r` should rerun current pattern tests

Dunqing:
### Why?

In debugging, we may need to change the `node_modules` files, but it does not automatically trigger a re-run of the test.

>  Besides, I think press `enter` is a better way to execute rerun. but the `enter` has been used in rerun all tests.

### DIFF
--- a/packages/vitest/src/node/stdin.ts
+++ b/packages/vitest/src/node/stdin.ts
@@ -61,11 +61,11 @@ export function registerConsoleShortcuts(ctx: Vitest) {
     if (name === 'u')
       return ctx.updateSnapshot()
     // rerun all tests
-    if (name === 'a')
+    if (name === 'a' || name === 'return')
       return ctx.changeNamePattern('')
     // rerun current pattern tests
-    if (name === 'return')
-      return ctx.rerunFiles(ctx.state.getFilepaths(), 'rerun')
+    if (name === 'r')
+      return ctx.rerunFiles()
     // rerun only failed tests
     if (name === 'f')
       return ctx.rerunFailed()
--- a/packages/vitest/src/node/stdin.ts
+++ b/packages/vitest/src/node/stdin.ts
@@ -13,7 +13,7 @@ const keys = [
   ['t', 'filter by a test name regex pattern'],
   ['q', 'quit'],
 ]
-const cancelKeys = ['space', 'c', ...keys.map(key => key[0])]
+const cancelKeys = ['space', 'c', ...keys.map(key => key[0]).flat()]
 
 export function printShortcutsHelp() {
   stdout().write(
--- a/packages/vitest/src/node/stdin.ts
+++ b/packages/vitest/src/node/stdin.ts
@@ -6,6 +6,7 @@ import type { Vitest } from './core'
 
 const keys = [
   ['a', 'rerun all tests'],
+  ['r', 'rerun current pattern tests'],
   ['f', 'rerun only failed tests'],
   ['u', 'update snapshot'],
   ['p', 'filter by a filename'],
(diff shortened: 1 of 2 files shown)


sheremet-va (changes requested):
Do you think this shortcut should be mentioned anywhere?

Dunqing:
> Do you think this shortcut should be mentioned anywhere?

I searched for a long time without finding any documentation about these shortcuts, until I saw in the source code 😩

It looks like we need to add a `Watch Mode` page to the documentation and add the documentation related to shortcuts here.

AriPerkkio (commented):
> Do you think this shortcut should be mentioned anywhere?

This should be mentioned here:

https://github.com/vitest-dev/vitest/blob/77d071a91f81b6faef28cb30c6c5b617a661459d/packages/vitest/src/node/stdin.ts#L7-L14

Also please add test case here if possible https://github.com/vitest-dev/vitest/blob/main/test/watch/test/stdin.test.ts

test/watch/test/stdin.test.ts
```diff
@@ -17,6 +17,15 @@ test('quit watch mode', async () => {
   await vitest.isDone
 })
 
+test('rerun current pattern tests', async () => {
+  const pattern = 'TEST_NAME_PATTERN'
+  const vitest = await startWatchMode(`-t ${pattern}`)
```
AriPerkkio, line 22: `startWatchMode` accepts an array of strings. A single argument separated with spaces doesn't seem to work through `execa`. Let's change this to array instead.

In the `test/watch/fixtures` there is no test case that would match `TEST_NAME_PATTERN`. Let's change this to `sum` so it matches test case in `fixtures/math.test.ts`.

```suggestion
  const vitest = await startWatchMode('-t', 'sum')
```

test/watch/test/stdin.test.ts
```diff
@@ -17,6 +17,15 @@ test('quit watch mode', async () => {
...
+
+  vitest.write('r')
+
+  await vitest.waitForOutput(`Test name pattern: ${pattern}`)
```
AriPerkkio, line 26: The output is formatted as Regexp:

```suggestion
  await vitest.waitForOutput('Test name pattern: /sum/')
```

test/watch/test/stdin.test.ts
```diff
@@ -17,6 +17,15 @@ test('quit watch mode', async () => {
...
+})
```
AriPerkkio, line 27: Add one additional check to make sure `r` preserved the pattern properly. 

```suggestion
  await vitest.waitForOutput('1 passed')
})
```

packages/vitest/src/node/stdin.ts
```diff
@@ -5,7 +5,8 @@ import { isWindows, stdout } from '../utils'
 import type { Vitest } from './core'
 
 const keys = [
-  ['a', 'rerun all tests'],
+  [['a, return'], 'rerun all tests'],
```
AriPerkkio, line 8: This breaks test run cancelling when pressing `'a'` as `cancelKeys` no longer contains `'a'`:

https://github.com/vitest-dev/vitest/blob/77d071a91f81b6faef28cb30c6c5b617a661459d/packages/vitest/src/node/stdin.ts#L15

AriPerkkio (changes requested):

packages/vitest/src/node/stdin.ts
Dunqing: Thanks!!

sheremet-va (approved):

AriPerkkio (approved):

PR #3306, opened 2023-05-05
fix(coverage): c8 aliased paths

### LINKED ISSUE #3268: Tests report does not cover exported function from typescript.
### Describe the bug

I have a file called `utils.ts` in the folder `src/logic/utils.ts` that has a simple function that adds three numbers.
```ts
export function addThreeNumbers(a: number, b: number, c: number) {
    return a + b + c;
}
```
in the folder `src/__tests__/unit/utils.test.ts` I have a test for this sum function.
```ts
import { describe, it, expect } from 'vitest'
import {addThreeNumbers} from "src/logic/utils";


describe("Sum of three numbers test", () =>{
    it("1 + 2 + 3 = 6",()=>{
        expect(addThreeNumbers(1,2,3)).toBe(6)
    })
})
```

My vitest.config.ts file looks like this
```ts
export default mergeConfig(
    viteConfig,
    defineConfig({
        test: {
            environment: 'jsdom',
            exclude: [...configDefaults.exclude, 'e2e/*'],
            root: fileURLToPath(new URL('./', import.meta.url)),
            coverage: {
                all: true,
                provider: 'c8',
                // move coverage report to ./tests/unit/coverage
                reportsDirectory: './tests/unit/coverage',
                include: ['src/**/*.vue', "src/**/*.ts", "src/**/*.js"],
                exclude: ['src/router', "**/__tests__/**"]
            },
        }
    })
)
```

I run `vitest run --coverage` and I see that `✓ src/__tests__/unit/utils.test.ts (1)` is ran but
in the report `utils.ts` has 0 statements covered.

### Reproduction

I hope the files provided are enough to see if there is either a bug or I have a misconfiguration somewhere.

### System Info

```shell
System:
    OS: Windows 10 10.0.22000
    CPU: (16) x64 AMD Ryzen 7 5700U with Radeon Graphics
    Memory: 5.40 GB / 15.35 GB
  Binaries:
    Node: 18.16.0 - C:\Program Files\nodejs\node.EXE
    npm: 9.6.5 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22000.120.0), Chromium (112.0.1722.58)

    "vitest": "^0.30.1" (from packages.json)
```

### Used Package Manager

npm

### Validations

AriPerkkio:
- Fixes #3268

`vitenode.fetchCache` contains file path without extension when module is imported using alias. This does not end up in coverage report due to missing extension.

### DIFF
--- a/packages/vite-node/src/client.ts
+++ b/packages/vite-node/src/client.ts
@@ -289,7 +289,8 @@ export class ViteNodeRunner {
 
     const modulePath = cleanUrl(moduleId)
     // disambiguate the `<UNIT>:/` on windows: see nodejs/node#31710
-    const href = pathToFileURL(modulePath).href
+    const [moduleUrl] = await this.resolveUrl(modulePath)
+    const href = pathToFileURL(moduleUrl).href
     const meta = { url: href }
     const exports = Object.create(null)
     Object.defineProperty(exports, Symbol.toStringTag, {


packages/vite-node/src/client.ts
```diff
@@ -289,7 +289,8 @@ export class ViteNodeRunner {
 
     const modulePath = cleanUrl(moduleId)
     // disambiguate the `<UNIT>:/` on windows: see nodejs/node#31710
-    const href = pathToFileURL(modulePath).href
+    const [moduleUrl] = await this.resolveUrl(modulePath)
+    const href = pathToFileURL(moduleUrl).href
```
AriPerkkio, line 293: When an aliased modules is loaded the `fsPath` doesn't contain extension at all. @sheremet-va 
 should I make this change to `fsPath` instead so that everywhere it's used would be affected, e.g. stacktraces I think?

With alias, without PR changes:

```js
import { addThreeNumbers } from "src/logic/utils";

{
  __filename: '/x/y/repro/src/logic/utils',
  id: 'src/logic/utils',
  fsPath: 'src/logic/utils'
}
```

Without alias, without PR changes, works OK:

```js
import { addThreeNumbers } from "../../logic/utils";

{
  __filename: '/x/y/repro/src/logic/utils.ts',
  id: '/src/logic/utils.ts',
  fsPath: '/x/y/repro/src/logic/utils.ts'
}
```


With alias, with PR changes:

```js
import { addThreeNumbers } from "src/logic/utils";

{
  __filename: '/x/y/repro/src/logic/utils.ts',
  id: 'src/logic/utils',
  fsPath: 'src/logic/utils'
}
```

sheremet-va: This should not even be allowed. Vitest should fail if the module is not resolved. I am coming up with a fix for this. Some people rely on tsconfig resolution when in practice we don't support it.

AriPerkkio: I see, the reproduction case is not using `alias` in `vitest.config.ts`. By adding following the coverage works OK:

```ts
    alias: {
      src: fileURLToPath(new URL("./src", import.meta.url)),
    },
```

So I guess in your fix Vitest will fail to load module if only `tsconfig.json` has `alias` but `vitest.config.ts` does not?

sheremet-va: We can finally stip supporting this weird edge case: https://github.com/vitest-dev/vitest/pull/3307

sheremet-va: > So I guess in your fix Vitest will fail to load module (or warn?) if only tsconfig.json has alias but vitest.config.ts does not?

People rely on the fact that their modules are relative to the `baseDir` which is never true in Vite. Because of this, we have a lot of issues when the file is not resolved correctly, but it can still fetch module code.

PR #3308, opened 2023-05-05
feat: ensure mocking plugin is the last one

sheremet-va:


antfu (approved):
Pushed a change, and verified it fixing `nuxt-vitest`'s failing

sheremet-va:
> Pushed a change, and verified it fixing `nuxt-vitest`'s failing

It should run after vite-analyzer.

sheremet-va:
@antfu do you want to look into this?

sheremet-va:
@antfu I am not sure about your change. What if another package will also use `order: 'post'`? Isn't it safer to modify plugins?

Currently, it doesn't work because Vite already resolved `vitest` entry and changed all imports. Mocks plugin expects that it didn't happen yet.

huang-julien:
Hi :wave: is there some kind of workaround for this issue ? This is quite blocking for nuxt-vitest

sheremet-va:
> Hi 👋 is there some kind of workaround for this issue ? This is quite blocking for nuxt-vitest

From my point of view, @antfu overtook this PR. So I am personally waiting on his actions/comments 😄

oskarols:
@antfu Anything we can do to help this move forward? 🙏

PR #3307, opened 2023-05-05
feat!: throw an error, if module cannot be resolved

sheremet-va:
Previously, if we could not resolve the path we just passed it down to `server.transformRequest` in hopes that it would work. However, this is not how it actually works in Vite - the frontend part will throw an error if the path is not resolvable during the transform phase.

This leads to several issues:

- People expect that Vitest takes `tsconfig.json` into account, but it doesn't (`src/file` is not resolved into `<root>/src/file.ts`, we just pass it down without an importer to `server.transformRequest` and it is resolved relative to root by accident)
- Coverage receives unresolved URL (`<root>/src/file` is not a valid path, it doesn't have an extension)
- This creates duplicate entries in the module graph for the same module (it's possible to have both `<root>/src/file.ts` and `src/file` in the module graph)

This PR now throws an error, if the module cannot be resolved. This doesn't affect modules that were mocked by `vi.mock` (even if they are not installed).

PR #3312, opened 2023-05-06
fix(vite-node): allow returning id not wrapped in promise

danielroe:
This is a small change to the type of `resolveId`. I believe it is only ever awaited, which means it should be safe to synchronously return an id.

I didn't think it was worth adding `@vitest/utils` to benefit from the `Awaitable` type but happy if you think otherwise.

(Might we also want to allow `void`?)

### DIFF
--- a/packages/vite-node/src/types.ts
+++ b/packages/vite-node/src/types.ts
@@ -41,7 +41,7 @@ export type HotContext = Omit<ViteHotContext, 'acceptDeps' | 'decline'>
 
 export type FetchFunction = (id: string) => Promise<FetchResult>
 
-export type ResolveIdFunction = (id: string, importer?: string) => Promise<ViteNodeResolveId | null>
+export type ResolveIdFunction = (id: string, importer?: string) => Promise<ViteNodeResolveId | null> | ViteNodeResolveId | null
 
 export type CreateHotContextFunction = (runner: ViteNodeRunner, url: string) => HotContext
 
--- a/packages/vite-node/src/types.ts
+++ b/packages/vite-node/src/types.ts
@@ -4,6 +4,7 @@ import type { ModuleCacheMap, ViteNodeRunner } from './client'
 
 export type Nullable<T> = T | null | undefined
 export type Arrayable<T> = T | Array<T>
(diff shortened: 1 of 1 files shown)


antfu (approved):
LGTM. Supporting `undefined` and `void` also makes sense to me.

sheremet-va:
I am not sure why this is needed. The only reason why `resolveId` exists as an option is to support `await import(customVariable)`. ViteNode will throw an error in future versions if it didn't return anything: #3307

If you have a specific use case where you use it, maybe it's better to expose another API.

danielroe:
I raised the PR as I was turning on type checking at this place in our codebase:

https://github.com/nuxt/nuxt/blob/a793e7ad48c56a034fd547081ed32b47a810c23e/packages/vite/src/runtime/vite-node.mjs#L41

We currently use this hook to spy on ids being resolved so we can give better errors in the event that there is an error resolving the id.

sheremet-va:
> We currently use this hook to spy on ids being resolved so we can give better errors in the event that there is an error resolving the id.

I know you have that - this is why I am pointing out that it might not be the best solution, because it will fail in the next version.

sheremet-va:
> I know you have that - this is why I am pointing out that it might not be the best solution, because it will fail in the next version.

Even now it doesn't work like you expect. `resolveId` is not called most of the time since ids are already resolved.

danielroe:
Well, this PR was mostly aimed at either fixing the type to match how it is currently used - or surfacing whether our usage is wrong - so I'm very happy to be told that we shouldn't use it as we do (if that's what I'm hearing from you).

sheremet-va:
> Well, this PR was mostly aimed at either fixing the type to match how it is currently used - or surfacing whether our usage is wrong - so I'm very happy to be told that we shouldn't use it as we do (if that's what I'm hearing from you).

As I said, the fix is fine if we are talking about syntax. It is allowed to return a value straight away, but it's usage like this is questionable considering Vite's resolution is async (which is what this function is used for). This is why I was concerned at first.

I think we can merge PR, but you should be aware that your usage of this method is incorrect.

PR #3314, opened 2023-05-07
fix: display error message correctly

### LINKED ISSUE #3309: The expected and received values ​​seem to be reversed
### Describe the bug

```expect(xx).toBe(xx)``` 
When an error occurs, the expected value and received value prompt information seem to be reversed.

vitest

image

jest

image

### Reproduction

https://stackblitz.com/edit/node-w1fgff?file=package.json

### System Info

```shell
System:
    OS: Windows 10 10.0.19044
    CPU: (8) x64 Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz
    Memory: 2.54 GB / 15.86 GB
  Binaries:
    Yarn: 1.22.10 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 7.22.0 - ~\AppData\Roaming\npm\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (112.0.1722.68)
  npmPackages:
    @vitest/coverage-c8: ^0.29.8 => 0.29.8
    vite: ^4.2.1 => 4.2.1
    vitest: ^0.31.0 => 0.31.0
```

### Used Package Manager

pnpm

### Validations

btea:
fix #3309

### DIFF
--- a/packages/utils/src/descriptors.ts
+++ b/packages/utils/src/descriptors.ts
@@ -23,8 +23,8 @@ export function getConcordanceTheme() {
       value: c.blue,
     },
     diffGutters: {
-      actual: `  ${c.red('-')} `,
-      expected: `  ${c.green('+')} `,
+      actual: `  ${c.green('-')} `,
+      expected: `  ${c.red('+')} `,
       padding: '    ',
     },
     error: {
@@ -94,5 +94,5 @@ export function getConcordanceTheme() {
 }
 
 export function diffDescriptors(actual: unknown, expected: unknown, options: DisplayOptions): string {
-  return concordanceModule.diff(actual, expected, options)
+  return concordanceModule.diff(expected, actual, options)
 }
--- a/packages/utils/src/descriptors.ts
+++ b/packages/utils/src/descriptors.ts
@@ -94,5 +94,5 @@ export function getConcordanceTheme() {
 }
 
 export function diffDescriptors(actual: unknown, expected: unknown, options: DisplayOptions): string {
-  return concordanceModule.diff(expected, actual, options)
+  return concordanceModule.diff(actual, expected, options)
 }
--- a/test/core/test/__snapshots__/mocked.test.ts.snap
+++ b/test/core/test/__snapshots__/mocked.test.ts.snap
@@ -6,8 +6,8 @@ exports[`mocked function which fails on toReturnWith > just one call 1`] = `
 Received: 
     1st spy call return:
 
-  - 2
-  + 1
+  - 1
+  + 2
 
 
 Number of calls: 1
@@ -20,18 +20,18 @@ exports[`mocked function which fails on toReturnWith > multi calls 1`] = `
 Received: 
     1st spy call return:
 
-  - 2
-  + 1
+  - 1
+  + 2
 
     2nd spy call return:
 
-  - 2
-  + 1
+  - 1
+  + 2
 
     3rd spy call return:
 
-  - 2
-  + 1
+  - 1
+  + 2
 
 
 Number of calls: 3
@@ -45,22 +45,22 @@ Received:
     1st spy call return:
 
     {
-  -   a: '4',
-  +   a: '1',
+  -   a: '1',
+  +   a: '4',
     }
 
     2nd spy call return:
 
     {
-  -   a: '4',
-  +   a: '1',
+  -   a: '1',
+  +   a: '4',
     }
 
     3rd spy call return:
 
     {
-  -   a: '4',
-  +   a: '1',
+  -   a: '1',
+  +   a: '4',
     }
 
 
--- a/test/core/test/__snapshots__/mocked.test.ts.snap
+++ b/test/core/test/__snapshots__/mocked.test.ts.snap
@@ -6,8 +6,8 @@ exports[`mocked function which fails on toReturnWith > just one call 1`] = `
 Received: 
     1st spy call return:
 
-  - 1
-  + 2
+  - 2
+  + 1
 
 
 Number of calls: 1
@@ -20,18 +20,18 @@ exports[`mocked function which fails on toReturnWith > multi calls 1`] = `
 Received: 
     1st spy call return:
 
-  - 1
-  + 2
+  - 2
+  + 1
 
     2nd spy call return:
 
-  - 1
-  + 2
+  - 2
+  + 1
 
     3rd spy call return:
 
-  - 1
-  + 2
+  - 2
+  + 1
 
 
 Number of calls: 3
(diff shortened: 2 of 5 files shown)


packages/utils/src/descriptors.ts
```diff
@@ -94,5 +94,5 @@ export function getConcordanceTheme() {
 }
 
 export function diffDescriptors(actual: unknown, expected: unknown, options: DisplayOptions): string {
-  return concordanceModule.diff(expected, actual, options)
+  return concordanceModule.diff(actual, expected, options)
```
btea, line 97: Refer to the following code, the first parameter should be actual, the second parameter should be expected

https://github.com/concordancejs/concordance/blob/791d2a89b40eb13f2c889ac270dd8be190cf8073/lib/diff.js#L388-L389

However, the test failed after the modification. According to the error message, it seems to be related to the content of `html.meta.json.gz`. I don't know how to deal with this?

sheremet-va: Run vitest with `-u` flag and update snapshots then

btea: Some files seem to be deleted after running, I'm not sure if this is correct.

sheremet-va: You should've run it only inside the test package that was failing, not all test packages,

btea: According to the error message, it seems that the two tests `mocked.test.ts` and `diff.test.ts` failed, but I ran these two tests locally and passed. I don’t know if there is a problem with my local environment. 😕

sheremet-va: Seems like you have some problems with your local env, check your branch, and rebuild the project.

btea: It seems to be ok after reinstalling dependencies and rebuilding.

test/core/test/__snapshots__/mocked.test.ts.snap
```diff
@@ -45,22 +45,22 @@ Received:
     1st spy call return:
 
     {
-  -   a: '4',
-  +   a: '1',
+  -   a: '1',
```
sheremet-va, line 48: This is now incorrect. "1" should be received

btea: `expect(a).toBe(b)` a is expected, b is received?

sheremet-va: It's not expected to be, it's what was called (expected) compared to what was received (actual)

The function was called with 1, but diff shows it was called with 4

btea: After this PR change it seems to become called(received) compared to what was expected, so this is correct.

sheremet-va: I don't know what you mean, but the diff here is incorrect. The function returned `{ a: '1' }`, but test expected `{ a: '4' }`. So the correct diff is:

```
- a: '4'
+ a: '1'
```

btea: I'm a little dizzy now 😵‍💫

According to your description, the tests seem to be reversed.

sheremet-va: With your change it is reversed, yes, but it shouldn't be.

btea: Sorry, I'm not sure what the correct modification is now. Maybe just reverse the colors here?
https://github.com/vitest-dev/vitest/blob/9836ccb48c5a39c0999484f55731f53ba494be17/packages/utils/src/descriptors.ts#L26-L27

sheremet-va: > Sorry, I'm not sure what the correct modification is now. Maybe just reverse the colors here?

Ok, I see now. Give me a bit more time to think

sheremet-va:
You are correct, colors should be changed here:

https://github.com/vitest-dev/vitest/blob/9836ccb48c5a39c0999484f55731f53ba494be17/packages/utils/src/descriptors.ts#L26-L27

btea:
> You are correct, colors should be changed here:
> 
> https://github.com/vitest-dev/vitest/blob/9836ccb48c5a39c0999484f55731f53ba494be17/packages/utils/src/descriptors.ts#L26-L27
> 

Keep the current modified state, and then update the color?

> The spy diff should also be updated with a prefix (but it's probably out of scope):
> 
> ```
> expect(red{vi.fn()}).toName(green{expected})
> ```

Do all tests in the `mocked.test.ts`  file have to add this prefix?

sheremet-va:
> Keep the current modified state, and then update the color?

Yes

> Do all tests in the mocked.test.ts file have to add this prefix?

Just forget about this for now

ghiscoding:
Posting this just an info, this PR seems to be the inverse of recent PR #3289, is that the intent of this PR?

Also looking at possible improvement on the colors, if we run the same test in Jest and Vitest. Jest is coloring in green all the expectations (not just the value but also the test title in green), but in Vitest that entire portion (not just the title) is yellow. I prefer Jest UI a bit more, I find it easier to read and spot the expectation error. Also this PR doesn't correct the red/green missing coloring on the expected value found, does it?

image

in Vitest, both result values are yellow, I thought this PR would fix this?
image

btea:
> Posting this just an info, this PR seems to be the inverse of recent PR #3289, is that the intent of this PR?

I seem to have made a wrong modification, maybe I should just update the colors. This is the result after changing

image

> in Vitest, both result values are yellow, I thought this PR would fix this? Actually this is more complicate than I taught, the text coloring in the expectation comes from theme `getConcordanceTheme()` and is different per variable type. It seems a bit complicate to change

Yes, this PR does not deal with the color problem of the value, it can be improved later.

sheremet-va:
> This is the result after changing

The string diff is confusing now (the red bg is in green "minus" sign) 😞 

We should probably improve the diff overall. The concordance tools don't seem to meet our expectations, but jest-diff doesn't print good string diff (line by line, not character by character), which also doesn't meet our expectations

ghiscoding:
@btea I think he's talking about string diffs, a removed character is currently showing with red background and white text, while an addition is showing as green background and white text. So I assume that his comment meant that you somehow changed these colors? Take this for example with current code, how does it show with your new code?

image

btea:
@ghiscoding Oh, thanks for the detailed explanation. I think I now understand what he means, and now there is indeed a problem as he said. Here is how the new code behaves in this example.



Maybe the correct string diff should look like this?



I just tried to modify the color of `delete` and `insert`, not sure if there are other problems. 
https://github.com/vitest-dev/vitest/blob/0f0d75631c3b9d6f335d822a24e25b46b2219eda/packages/utils/src/descriptors.ts#L69-L76

ghiscoding:
@btea which chars are additions and which ones are removal? I don't think the descriptor colors should be touched, because if you scroll up you'll see this comment (no adjustment were made)

https://github.com/vitest-dev/vitest/blob/0f0d75631c3b9d6f335d822a24e25b46b2219eda/packages/utils/src/descriptors.ts#L13-L18

btea:
> @btea which chars are additions and which ones are removal? 

To be honest, I'm not sure what the correct modification is now. Maybe I was wrong in the first place. 😵
>I don't think the descriptor colors should be touched, because if you scroll up you'll see this comment (no adjustment were made)

You are right, I just noticed this comment.

PR #3337, opened 2023-05-10
feat(dev): add modulePatterns option to the vitest config

### LINKED ISSUE #3265: Mock package in monorepo
### Describe the bug

I am facing an issue mocking a package in a monorepo.

When I use `vi.mock('whatever')` without specifying an inline factory, I would expect vitest to look for the mock in the app's `__mocks__` folder in the monorepo, but if I mock a package of the monorepo, that does not happen.
I have added some logs [here](https://github.com/vitest-dev/vitest/blob/7c8f0ba989aaa3152f939d6d6adbf55a14e68473/packages/vitest/src/runtime/mocker.ts#L82) and [here](https://github.com/vitest-dev/vitest/blob/7c8f0ba989aaa3152f939d6d6adbf55a14e68473/packages/vitest/src/runtime/mocker.ts#L178) to gather more information.

When I mock a node module (e.g. undici), the output is as follows:

```
resolvePath {
  rawId: 'undici',
  id: '[...]/monorepo-vi-mock-resolver/node_modules/.pnpm/undici@5.22.0/node_modules/undici/index.js',
  fsPath: '[...]/monorepo-vi-mock-resolver/node_modules/.pnpm/undici@5.22.0/node_modules/undici/index.js'
}
const external = !isAbsolute(fsPath) || fsPath.includes("/node_modules/") ? rawId : null
{
  'isAbsolute(fsPath)': true,
  'fsPath.includes("/node_modules/")': true,
  external: 'undici'
}

resolveMockPath {
  mockPath: '[...]/monorepo-vi-mock-resolver/node_modules/.pnpm/undici@5.22.0/node_modules/undici/index.js',
  external: 'undici',
  path: 'undici',
  'isNodeBuiltin(mockPath)': false,
  'existsSync(mockPath)': true
}
=> enter the if (external || isNodeBuiltin(mockPath) || !existsSync(mockPath))
{
  mockDirname: '.',
  mockFolder: '[...]/monorepo-vi-mock-resolver/apps/foo/__mocks__',
  'existsSync(mockFolder)': true
}
__mocks__ content { files: [ 'sample.ts', 'undici.ts' ], baseOriginal: 'undici' }
return [...]/monorepo-vi-mock-resolver/apps/foo/__mocks__/undici.ts
```

If I mock a monorepo package instead, the output looks like this:

```
resolvePath {
  rawId: 'sample',
  id: '[...]/monorepo-vi-mock-resolver/packages/sample/dist/index.js',
  fsPath: '[...]/monorepo-vi-mock-resolver/packages/sample/dist/index.js'
}
const external = !isAbsolute(fsPath) || fsPath.includes("/node_modules/") ? rawId : null
{
  'isAbsolute(fsPath)': true,
  'fsPath.includes("/node_modules/")': false,
  external: null
}

resolveMockPath {
  mockPath: '[...]/monorepo-vi-mock-resolver/packages/sample/dist/index.js',
  external: null,
  path: '[...]/monorepo-vi-mock-resolver/packages/sample/dist/index.js',
  'isNodeBuiltin(mockPath)': false,
  'existsSync(mockPath)': true
}
=> skip the if (external || isNodeBuiltin(mockPath) || !existsSync(mockPath))
{
  dir: '[...]/monorepo-vi-mock-resolver/packages/sample/dist',
  baseId: 'index.js',
  fullPath: '[...]/monorepo-vi-mock-resolver/packages/sample/dist/__mocks__/index.js'
}
return null // because the mock file (fullPath) does not exist
```

And the final map is:

```
mockMap {
  '[...]/monorepo-vi-mock-resolver/apps/foo/src/index.test.ts' => {
    '[...]/monorepo-vi-mock-resolver/node_modules/.pnpm/undici@5.22.0/node_modules/undici/index.js': '[...]/monorepo-vi-mock-resolver/apps/foo/__mocks__/undici.ts',
    '[...]/monorepo-vi-mock-resolver/packages/sample/dist/index.js': null
  }
}
```

In the second case, since the `fsPath` is an absolute path and does not contain `/node_modules/`, the dependency is not marked as "external" and the mock file is searched for in the `dist/__mocks__` folder of the package itself.

Is this an intended behaviour?

### Reproduction

https://github.com/fooddilsn/monorepo-vi-mock-resolver

### System Info

```shell
System:
  OS: macOS 13.2.1
  CPU: (10) arm64 Apple M1 Max
  Memory: 72.38 MB / 32.00 GB
  Shell: 5.8.1 - /bin/zsh
Binaries:
  Node: 16.17.1 - ~/.nvm/versions/node/v16.17.1/bin/node
  Yarn: 1.22.19 - ~/.nvm/versions/node/v16.17.1/bin/yarn
  npm: 8.15.0 - ~/.nvm/versions/node/v16.17.1/bin/npm
Browsers:
  Chrome: 112.0.5615.137
  Safari: 16.3
```

### Used Package Manager

pnpm

### Validations

fooddilsn:
Close #3265


sheremet-va (changes requested):
I don't really see a reason to use glob patterns here. I don't like bringing another dependency inside the runtime (we use MM only on the server part now), and I think we should follow `deps.inline` pattern.

fooddilsn:
> I don't really see a reason to use glob patterns here. I don't like bringing another dependency inside the runtime (we use MM only on the server part now), and I think we should follow `deps.inline` pattern.

It makes sense, I was very undecided about glob patterns and I preferred to get your opinion, thanks for the quick feedback. I will change it!

So you mean `moduleDirectories: (string | RegExp)[]`, right? Or am I missing the scenario for the `true` value?

sheremet-va:
> So you mean `moduleDirectories: (string | RegExp)[]`, right?

Yes, I mean this one. Without `true` 😄

sheremet-va:
A few other points:

- Would be nice to have a test case for this
- Would be nice to explain in the docs how this is used. Currently, it is only for mocking, I think it should be mentioned. Maybe even in the `__mocks__` [algorithm](https://vitest.dev/api/vi.html#vi-mock)

sheremet-va:
Maybe we should also reuse this in `vite-node/externalize`? It checks `/node_modules/${name}` right now, and I think it might be a good idea to also respect that config option

fooddilsn:
> Maybe we should also reuse this in `vite-node/externalize`? It checks `/node_modules/${name}` right now, and I think it might be a good idea to also respect that config option

So you are proposing to change the implementation of `matchExternalizePattern` function to do
```
id.includes(`/${dir.replace(/(^\/?|\/?$)/g, '')}/`)
```
like in the `VitestMocker`, instead of
```
id.includes(`/node_modules/${ex}/`)
```
right?

fooddilsn:
Actually, to solve this [issue](https://github.com/vitest-dev/vitest/issues/3265), I first tried using the `deps.external` option to tell Vitest to mock my package as an external module.
We could also consider using this option instead of adding another one, but I am not sure if I want to bypass the package to native Node or lose the support for HMR on reload, as described [here](https://vitest.dev/config/#deps-external).
What do you think @sheremet-va?

fooddilsn:
> `modulePatterns` option is better in this case. The fact that it works now doesn't guarantee that it will work like that in the future. `deps.external` only tells Vitest to not process your module with Vite. It is not used by mocker.

I know, I was just thinking about the possibility of starting to use it in the mocker as well. But I'm fine with continuing with `moduleDirectories` 👍

sheremet-va:
> Today, only `node_modules/foo` would be externalized, but if we change the behavior of `matchExternalizePattern` folders such as `whatever/foo` could also be involved.

But the same can be said about mocking? I don't see how this is a breaking change. `moduleDirectories` doesn't exist yet.

fooddilsn:
> But the same can be said about mocking? I don't see how this is a breaking change. `moduleDirectories` doesn't exist yet.

Precisely because `moduleDirectories` doesn't exist yet, nobody uses this option and therefore the current behavior will not be changed in any existing project. We only provide the ability to do something that wasn't possible before.

But of course, if you think it's the right thing, I can definitely apply the change to `deps.external` in this PR.

sheremet-va:
> But of course, if you think it's the right thing, I can definitely apply the change to `deps.external` in this PR.

I do think it would be nice to implement these features together.

fooddilsn:
@sheremet-va I'm sorry, but I haven't had time to work on it lately. Thanks for taking care of this.

Issue #3251, opened 2023-04-25
Workspace coverage wrong when a file transforms differently in projects


AriPerkkio:
### Describe the bug

Writing down the findings from https://github.com/vitest-dev/vitest/pull/3226#discussion_r1173453639 with reproduction case. This affects both `@vitest/coverage-c8` and `@vitest/coverage-istanbul`.

If a single file can transform differently in different workspace projects, the final coverage reported by `@vitest/coverage-<x>` packages will be incorrect. For example:

```
├── vitest.workspace.ts // Declares workspaces ['client', 'server']
├── vitest.config.ts // Enables coverage and sets the options
|
├── client
|  └── test
|     └── user-list.test.js // Imports user-list.js with web/client-side specific transforms
|
├── server
|  └── test
|     └── user-list.test.js // Imports user-list.js with SSR transforms
|
└── shared
   └── user-list.js // Can output both client and server side code depending on environment
```

Here both client and server import `shared/user-list.js` with different code transforms. These produce two different code outputs and source maps.

When generating coverage report the `@vitest/coverage-c8` picks the first source map it finds from project's `vite-node` cache and uses it for both workspace projects. Instead it should be aware of which workspace project the current V8 coverage report belongs to, and use that project's `vite-node` cache.

I haven't yet debugged the `@vitest/coverage-istanbul` case but it's likely picking the first source map it encounters in `istanbul-lib-source-maps` and re-uses that one for both file coverages. Instead it should separate each workspace projects' `coverageMap` completely, and merge them at the end into one.

### Reproduction

- https://github.com/AriPerkkio/vitest/tree/repro/workspace-coverage-with-transforms/test/workspaces
- https://github.com/AriPerkkio/vitest/commit/f9d095cddac68bff6ce9b0233aa0fb7f88c128f6

<details>
  <summary>Source</summary>

```js
export default function run(a = 1, b = 2) {
  //! WORKSPACE_1_REMOVE_START

  // In workspace 1 we remove two functions
  function removedFunctionOne() {
    return 1
  }
  function removedFunctionThree() {
    return 1
  }

  //! WORKSPACE_1_REMOVE_END

  const unusedFunction1 = () => 1
  const unusedFunction2 = () => a ? b : null

  // Coverage of this branch breaks when combined coverage of workspaces is not handled correctly
  if (a === 1000 && b === 1) {
    // This should be uncovered
    return 1001
  }

  //! WORKSPACE_3_REMOVE_START

[...]

PR #3339, opened 2023-05-10
feat!: add `@vitest/coverage-v8` package

AriPerkkio:
- Adds new coverage provider package `@vitest/coverage-v8`. For end-users it should provide identical coverage results as `@vitest/coverage-c8`.
  - This package does not implement the coverage options `excludeNodeModules`, `allowExternal` and `src`. I think we can implement these if needed, but at the moment even the `@vitest/coverage-istanbul` doesn't have these.
- Adds deprecation warning to `@vitest/coverage-c8`. This package is still around but will be removed in future breaking-change releases.
- Changes default `coverage.provider` from `'c8'` to `'v8'`

### Migration

```sh
npm uninstall @vitest/coverage-c8
npm install --save-dev @vitest/coverage-v8
```

```diff
// vitest.config.ts
export default defineConfig({
  test: {
    coverage: {
-      provider: 'c8'
+      provider: 'v8'
```

#### Troubleshooting

- If your coverage report is completely empty, please check that your `test.root` does not exclude the source files. The `root` is now resolved correctly, identical to `@vitest/coverage-istanbul`: https://github.com/vitest-dev/vitest/issues/3555

### Motivation for new package

The `@vitest/coverage-c8` has always had to utilize some private APIs of `c8`. This has been required since `c8` isn't really intended for libraries as it's rather for standalone usage. Non-breaking changes in `c8` like https://github.com/bcoe/c8/pull/469 can break `@vitest/coverage-c8` without major version bumps.

The `@vitest/coverage-c8` is using `c8` only for merging, processing and converting the V8 coverage reports, and for creating the final reports. Code instrumentation moved from expensive `process.env.NODE_V8_COVERAGE` to `node:inspector` usage already in #2786.

Now with `@vitest/coverage-v8` we don't use `c8` package at all. New dependencies compared to `@vitest/coverage-c8` are added to preserve the functionality:
- bunch of `istanbul` packages to process coverage results and create reports. Same packages are used by `@vitest/coverage-istanbul` already.
- `@bcoe/v8-coverage` for merging V8 coverage reports - this package is also used by `c8` and `jest`
- `v8-to-istanbul` for converting the coverage reports from V8 to istanbul format

Enables (or at least makes it easier to implement) https://github.com/vitest-dev/vitest/issues/3251 as now we have more control over `c8`.

### BREAKING CHANGES:
 - `@vitest/coverage-istanbul` requires latest `vitest` version. Some utilities were moved to `vitest` package for sharing them with the new `v8` provider.
 - Default `coverage.provider` is now `'v8'` instead of `'c8'`

### Next steps

Once this PR is released we can:

- Remove `@vitest/coverage-c8` package completely. The old releases will still remain in NPM registry. Latest release has deprecation warning in place.
- Remove all mentions of `c8` from documentation

### Testing:

Compare results between existing `@vitest/coverage-c8` and the new `@vitest/coverage-v8`.

 - [x] `test/coverage-test`
 - [x] `test/workspaces`
 - Manual testing with third party projects:
   - [x] [`zustand`](https://github.com/pmndrs/zustand): 100% coverage match
   - [x] [`vuejs/core`](https://github.com/vuejs/core): 99% coverage match. [Some conditional `T extends U`](https://github.com/vuejs/core/blob/a374d7e6ed973cde7fae36ee82618cf46a8ba68a/packages/shared/src/typeUtils.ts#L4) types are not considered as branches anymore. Earlier we didn't pass source maps for files containing only typings (not sure why), and let `c8` resolve the source map. Now we correctly pass the source map from `vite-node`.
   - [x] [`netlify/cli`](https://github.com/netlify/cli): 100% coverage match

### DIFF
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -687,7 +687,7 @@ If you are using coverage options with dot notation, don't forget to specify `--
 #### coverage.provider
 
 - **Type:** `'c8' | 'v8' | 'istanbul' | 'custom'`
-- **Default:** `'c8'`
+- **Default:** `'v8'`
 - **CLI:** `--coverage.provider=<provider>`
 
 Use `provider` to select the tool for coverage collection.
--- a/docs/guide/coverage.md
+++ b/docs/guide/coverage.md
@@ -16,7 +16,7 @@ The `c8` provider is being replaced by the [`v8`](https://v8.dev/blog/javascript
 Since Vitest v0.22.0
 :::
 
-Both `v8` and `istanbul` support are optional. By default, `c8` will be used.
+Both `v8` and `istanbul` support are optional. By default, `v8` will be used.
 
 You can select the coverage tool by setting `test.coverage.provider` to `v8` or `istanbul`:
 
--- a/packages/vitest/src/defaults.ts
+++ b/packages/vitest/src/defaults.ts
@@ -27,7 +27,7 @@ const defaultCoverageExcludes = [
 
 // These are the generic defaults for coverage. Providers may also set some provider specific defaults.
 export const coverageConfigDefaults: ResolvedCoverageOptions = {
-  provider: 'c8',
+  provider: 'v8',
   enabled: false,
   clean: true,
   cleanOnRerun: true,
--- a/packages/vitest/src/node/cli-api.ts
+++ b/packages/vitest/src/node/cli-api.ts
@@ -59,11 +59,21 @@ export async function startVitest(
   const ctx = await createVitest(mode, options, viteOverrides)
 
   if (mode === 'test' && ctx.config.coverage.enabled) {
-    const provider = ctx.config.coverage.provider || 'c8'
+    const provider = ctx.config.coverage.provider || 'v8'
     const requiredPackages = CoverageProviderMap[provider]
 
     if (requiredPackages) {
-      if (!await ensurePackageInstalled(requiredPackages, root)) {
+      // Remove this message once support for @vitest/coverage-c8 has been removed completely
+      const defaultProviderInfo = 'Default coverage provider has changed from "c8" to "v8". '
+        + 'New package is required to be installed. '
+        + 'To use the old deprecated coverage provider use "--coverage.provider c8" option.\n'
+        + 'See https://github.com/vitest-dev/vitest/pull/3339 for more information.\n\n'
+
+      const isUsingDefaultProvider
+        = ctx.server.config.test?.coverage?.provider === undefined
+        && options.coverage?.provider === undefined
+
+      if (!await ensurePackageInstalled(requiredPackages, root, isUsingDefaultProvider ? defaultProviderInfo : undefined)) {
         process.exitCode = 1
         return ctx
       }
--- a/packages/vitest/src/node/pkg.ts
+++ b/packages/vitest/src/node/pkg.ts
@@ -9,11 +9,15 @@ const __dirname = url.fileURLToPath(new URL('.', import.meta.url))
 export async function ensurePackageInstalled(
   dependency: string,
   root: string,
+  errorMessage?: string,
 ) {
   if (isPackageExists(dependency, { paths: [root, __dirname] }))
     return true
 
   const promptInstall = !isCI && process.stdout.isTTY
+
+  if (errorMessage)
+    process.stderr.write(c.red(errorMessage))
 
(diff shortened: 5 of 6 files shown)


dammy001:
> Does the package name `@vitest/coverage-c8` make any sense anymore? Maybe `@vitest/coverage-v8` instead?

@AriPerkkio if changed, @vitest/coverage-c8 will be depreciated at first right?

AriPerkkio:
Moved all changes of `@vitest/coverage-c8` package to new package `@vitest/coverage-v8` instead. This PR does not remove anything `c8` related - it only adds a deprecation warning message to `@vitest/coverage-c8` initializer. 

Small breaking change in `@vitest/coverage-istanbul` which now requires latest `vitest` version.

docs/config/index.md
```diff
@@ -645,7 +645,11 @@ Isolate environment for each test file. Does not work if you disable [`--threads
 
 ### coverage<NonProjectOption />
 
-You can use [`c8`](https://github.com/bcoe/c8), [`istanbul`](https://istanbul.js.org/)  or [a custom coverage solution](/guide/coverage#custom-coverage-provider) for coverage collection.
+You can use [`c8`](https://github.com/bcoe/c8), [`v8`](https://v8.dev/blog/javascript-code-coverage), [`istanbul`](https://istanbul.js.org/)  or [a custom coverage solution](/guide/coverage#custom-coverage-provider) for coverage collection.
```
antfu, line 648:
```suggestion
You can use [`v8`](https://v8.dev/blog/javascript-code-coverage), [`istanbul`](https://istanbul.js.org/)  or [a custom coverage solution](/guide/coverage#custom-coverage-provider) for coverage collection.
```

We could start not listing it.

docs/config/index.md
```diff
@@ -645,7 +645,11 @@ Isolate environment for each test file. Does not work if you disable [`--threads
...
+
+::: info
+The `c8` provider is being replaced by `v8` and will eventually be removed completely.
+:::
```
antfu, line 652:
```suggestion
::: info
The `c8` provider is being replaced by the `v8` provider. It will be deprecated in the next major version.
:::
```

docs/config/index.md
```diff
@@ -659,7 +663,7 @@ If you are using coverage options with dot notation, don't forget to specify `--
 
 #### coverage.provider
 
-- **Type:** `'c8' | 'istanbul' | 'custom'`
+- **Type:** `'c8' | 'v8' | 'istanbul' | 'custom'`
 - **Default:** `'c8'`
```
antfu, line 667:
```suggestion
- **Type:** `'v8' | 'istanbul' | 'custom'`
- **Default:** `'v8'`
```

Do we see any blocker we could make `v8` the new default without breaking existing usage?

antfu (commented):
The direction SGTM 👍

docs/config/index.md
antfu: Sounds good to me!

AriPerkkio: There is now a message printed before prompting the installation of new package:



This is only shown when `test.coverage.provider` or `--coverage.provider` option is not explicitly set.

docs/guide/coverage.md
```diff
@@ -4,17 +4,21 @@ title: Coverage | Guide
...
+:::
 
 ## Coverage Providers
 
 :::tip
 Since Vitest v0.22.0
 :::
 
-Both `c8` and `istanbul` support are optional. By default, `c8` will be used.
+Both `v8` and `istanbul` support are optional. By default, `c8` will be used.
```
sheremet-va, line 19:
```suggestion
Both `v8` and `istanbul` support are optional. By default, `v8` will be used.
```

sheremet-va (approved):

C0ZEN:
Thanks for the description, it's neat 💯

aleksi-magner:
Package @vitest/coverage-v8 does not work correctly in terms of coverage for sure. Package @vitest/coverage-c8 is more accurate - sees all uncovered and skiped areas in Vue templates, takes into account missing test files and overall coverage.
Other things being equal settings, package @vitest/coverage-v8 simply always passes the tests and never outputs missing lines or coverage problems.
So for now, these are not interchangeable packages. I rolled back to package @vitest/coverage-c8.

AriPerkkio:
@aleksi-magner these two packages should have 100% identical output. We've had some issues with Vue's coverage before so there might be something that we are missing in the new package as well. Please file a bug report with reproduction case and let's fix that.

Issue #3266, opened 2023-04-27
CLI: Show full test suite when testing 1 spec file at a time (like Jest does)


ghiscoding:
### Clear and concise description of the problem

Coming from Jest, I find that this feature is missing in Vitest and I quite like to see the full list of tests when I'm currently working on a test suite. The way it works in Jest is that if you are testing all test suites, then it shows only the title of each test suite (Vitest does that too), but if you then add/update some test(s) on a test suite (or if you are using "p" for pattern that finds only 1 spec file), then it will show the entire list of tests written under that test suite (Vitest does not that, it only shows the test suite title all the time).

For example in Jest when testing 1 spec file test suite, it shows all tests under it (in this case 12 tests), it's quite handy to see the full list, also sometime there are skipped tests that I can see in this list but I can't see them in Vitest

![image](https://user-images.githubusercontent.com/643976/234990436-a18532d4-8364-45da-9fd6-4291e1677e9a.png)

On the other hand, Vitest is only ever showing **only** the title of the describe test suite title. There are 4 tests under that spec file but it never shows them and I would really like to get the same view that Jest provides me which is helpful when troubleshooting a specific unit test file

![image](https://user-images.githubusercontent.com/643976/234990595-5d06eeb4-b209-407f-800c-dfb879dc42e0.png)

### Suggested solution

Provide the same view as Jest does (which is to show the full list of tests under the spec file test suite)

### Alternative

it's simply missing in Vitest, there are no alternative that I know of

### Additional context

There are more things to note about the print screens that I provided on top, I mention that we do not see the full list of tests under the test suite in Vitest... however that is not entirely true when we have an test failing, then we do see a summary similar to Jest. But even then, I still prefer Jest's UI in this case, the color they chose and the alignment of the text is simply easier to read in Jest. Also in Jest the spec filename is not shown on the same line as the describe test suite title, but in Vitest it's all on the same line (and it most often overflows) and I find that also harder to read and find which spec file and tests really failed

when failing in Jest, it's quite easy to read and find which test failed
![image](https://user-images.githubusercontent.com/643976/234991806-a091ee04-4a3b-4fec-863b-2a6db3b708c3.png)

when failing in Vitest, it's not as easy to read and trying to find the test that failed also seems much harder to see in the result window, I think it could be improved (for example separating the spec filename and the test describe under separate lines would help, and also use better or less colors)
[...]

PR #3349, opened 2023-05-10
fix(cli): improve colors used when erroring

ghiscoding:
- partially fixes issue #3266 (following this maintainer [comment](https://github.com/vitest-dev/vitest/issues/3266#issuecomment-1539326843))
- previously spec filename link were shown in yellow, I changed the color to cyan making easier to find (same as Jest)
- also remove yellow color from test code highlight, simply reverting to default text color

I find this change to be easier to read, especially in dark mode, the link color that is much quicker to find compare to before since everything used to be yellow.

See before/after below (left/right), the biggest difference is obviously in dark mode because of the contrast

image

image

### DIFF
--- a/packages/vitest/src/node/error.ts
+++ b/packages/vitest/src/node/error.ts
@@ -57,6 +57,10 @@ export async function printError(error: unknown, ctx: Vitest, options: PrintErro
     printErrorType(type, ctx)
   printErrorMessage(e, ctx.logger)
 
+  // E.g. AssertionError from assert does not set showDiff but has both actual and expected properties
+  if (e.diff)
+    displayDiff(e.diff, ctx.logger.console)
+
   // if the error provide the frame
   if (e.frame) {
     ctx.logger.error(c.yellow(e.frame))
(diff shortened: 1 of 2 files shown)


ghiscoding:
I'm not sure why the Lint fails, I ran `pnpm run lint:fix` and see no git changes, feel free to recreate the PR if need be, it's only 2 line of code change

sheremet-va:
Do you want to change the placing of the diff to make it easier to spot?

ghiscoding:
@sheremet-va do you mean to move the Expected/Received count before the test code block but after the spec filename (in cyan)? However my demo is a really small unit test, in some cases the user might need to scroll up to see that info, is that still how you want to order these results?

sheremet-va:
> do you mean to move the Expected/Received count before the test code block but after the spec filename (in cyan)?

No, cyan is part of the callstack. The code block is there because it point to a file in source code (the error can be outside of the test file)

I was thinking about moving the diff before the cyan part

ghiscoding:
@sheremet-va I'm not really sure where to change that in the code, I couldn't find in the code where this output is being built.

sheremet-va:
> @sheremet-va I'm not really sure where to change that in the code, I couldn't find in the code where this output is being built.

https://github.com/vitest-dev/vitest/blob/4a807df37bc5da735048040a94b820f06841bfac/packages/vitest/src/node/error.ts#L99

ghiscoding:
@sheremet-va the code you mention just points to a console error, I have a hard time understanding how to do the change you asked, would it be possible to merge the PR without this change? I was more interested in the color change myself, the location of the expected/received doesn't bother me

https://github.com/vitest-dev/vitest/blob/4a807df37bc5da735048040a94b820f06841bfac/packages/vitest/src/node/error.ts#L187-L189

Issue #3329, opened 2023-05-09
Inline snapshot doesn't work with `test.each()`


fenghan34:
### Describe the bug

`Vitest` returns strange inline snapshots when using `test.each()`, which are invalid javascript strings, ie:

```ts
import { expect, test } from 'vitest';

test.each([1, 2])('', (input) => {
  expect(Math.pow(input, 2)).toMatchInlineSnapshot('1''4')

  expect({ num: input }).toMatchInlineSnapshot(`
    {
      "num": 1,
    }
  ``
    {
      "num": 2,
    }
  `)
});
```

<img width="446" alt="image" src="https://user-images.githubusercontent.com/59411875/236989807-59ba3116-a9eb-4a26-a2d9-25004a5d7a3f.png">

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-1ejwed?file=test/basic.test.ts

### System Info

```shell
System:
    OS: macOS 13.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 129.00 MB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.15.1 - ~/.nvm/versions/node/v16.15.1/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v16.15.1/bin/yarn
    npm: 8.11.0 - ~/.nvm/versions/node/v16.15.1/bin/npm
  Browsers:
    Chrome: 112.0.5615.137
    Chrome Canary: 108.0.5339.0
    Safari: 16.2
  npmPackages:
    @vitejs/plugin-react: ^3.1.0 => 3.1.0 
    vite: ^4.2.1 => 4.2.1 
    vitest: ^0.29.7 => 0.29.7 
```
```


### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
[...]

PR #3360, opened 2023-05-12
feat: throw error if using inline snapshot inside of `test.each` or `describe.each`

fenghan34:
#3329

### DIFF
--- a/packages/runner/src/suite.ts
+++ b/packages/runner/src/suite.ts
@@ -53,7 +53,7 @@ export function createSuiteHooks() {
 }
 
 // implementations
-function createSuiteCollector(name: string, factory: SuiteFactory = () => { }, mode: RunMode, concurrent?: boolean, shuffle?: boolean, suiteOptions?: number | TestOptions) {
+function createSuiteCollector(name: string, factory: SuiteFactory = () => { }, mode: RunMode, concurrent?: boolean, shuffle?: boolean, each?: boolean, suiteOptions?: number | TestOptions) {
   const tasks: (Test | TaskCustom | Suite | SuiteCollector)[] = []
   const factoryQueue: (Test | Suite | SuiteCollector)[] = []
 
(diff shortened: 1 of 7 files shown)


fenghan34:
Hi @sheremet-va 

Could you please help to review this PR, thanks!

packages/snapshot/src/client.ts
```diff
@@ -32,6 +32,7 @@ interface AssertOptions {
   name?: string
   message?: string
   isInline?: boolean
+  isInsideEach?: boolean
```
sheremet-va, line 35: I don't think it should be part of the `packages/snapshot`. It couples too tightly with the runner.

sheremet-va (changes requested):

packages/snapshot/src/client.ts
fenghan34: Make sense! I've removed it.

fenghan34:
> Tests seem to be failing.

I'm looking into it, haven't found the reason.

packages/runner/src/suite.ts
```diff
@@ -215,6 +217,8 @@ function createSuite() {
           ? suite(formatTitle(name, items, idx), () => fn(...items), options)
           : suite(formatTitle(name, items, idx), () => fn(i), options)
       })
+
+      this.withContext({ each: undefined })
```
sheremet-va, line 221: I don't like this usage of the function. Maybe just `delete this.each` or `this.each = undefined`?

sheremet-va (changes requested):

packages/runner/src/suite.ts
fenghan34: we can't do that as `this` reference to a bound function(`suite` with context), so here `this.each` is the function(`suite.each`), not the flag `each`.

sheremet-va (approved):

Issue #3334, opened 2023-05-09
`repeats` does not work at all


Shinigami92:
### Describe the bug

There is a v0.31.0 new feature `repeats` #2652 

At "real world" it does not work at all.

There are tests in the code base, but somehow the bundled dist seems not to work

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-kzdmn9?file=test/suite.test.ts

### System Info

```shell
System:
    OS: Linux 6.1 Arch Linux
    CPU: (28) x64 Intel(R) Core(TM) i9-10940X CPU @ 3.30GHz
    Memory: 26.91 GB / 62.48 GB
    Container: Yes
    Shell: 5.9 - /usr/bin/zsh
  Binaries:
    Node: 18.16.0 - ~/.nvm/versions/node/v18.16.0/bin/node
    Yarn: 1.22.19 - /usr/bin/yarn
    npm: 9.5.1 - ~/.nvm/versions/node/v18.16.0/bin/npm
  Browsers:
    Firefox: 112.0.2
  npmPackages:
    @vitest/coverage-c8: ~0.31.0 => 0.31.0 
    @vitest/ui: ~0.31.0 => 0.31.0 
    vite: ~4.3.5 => 4.3.5 
    vitest: ~0.31.0 => 0.31.0
```

### Used Package Manager

pnpm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

sheremet-va:
It doesn't repeat the code in "describe" block, it repeats _tests_. "describe" just collects and groups tests

Shinigami92:
> It doesn't repeat the code in "describe" block, it repeats _tests_. "describe" just collects and groups tests

[...]

PR #3369, opened 2023-05-13
fix: test repeats

fenghan34:
Fixes: #3334

### DIFF
--- a/packages/runner/src/suite.ts
+++ b/packages/runner/src/suite.ts
@@ -53,7 +53,7 @@ export function createSuiteHooks() {
 }
 
 // implementations
-function createSuiteCollector(name: string, factory: SuiteFactory = () => { }, mode: RunMode, concurrent?: boolean, shuffle?: boolean, suiteOptions?: number | TestOptions) {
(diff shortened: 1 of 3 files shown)


sheremet-va:
I don't think this is a solution to the problem as a whole. Does this work?

```ts
describe('1', () => {
  describe('2', () => {
    describe('3', () => {
      describe('4', () => {
        test('5', () => {
          expect(1).toBe(1)
        })
      })
    })
  })
}, { repeats: 10 })
```

I also don't like calling `getCurrentSuite` inside the suite function itself, I think we have enough information in the `createSuiteCollector`

fenghan34:
> I don't think this is a solution to the problem as a whole. Does this work?
> 
> ```ts
> describe('1', () => {
>   describe('2', () => {
>     describe('3', () => {
>       describe('4', () => {
>         test('5', () => {
>           expect(1).toBe(1)
>         })
>       })
>     })
>   })
> }, { repeats: 10 })
> ```
>
 
Yes, the test runs 10 times.

> I also don't like calling `getCurrentSuite` inside the suite function itself, I think we have enough information in the `createSuiteCollector`

Me neither. But I didn't find a feasible way to get outer `describe` options when init inner `describes`. Do you have any suggestions?

Anyway, feel free to close the PR if this is not acceptable.

sheremet-va (approved):

Issue #2918, opened 2023-02-25
[vite-node] error transforming template literals with source map magic string


Tanimodori:
### Describe the bug

`vite-node` will transform the following code

```ts
export function foo(src: string): string {
  return `//# sourceMappingURL=data:application/json;base64,${src}`;
}
```

into

```js
// /home/projects/vitest-dev-vitest-dq6zze/foo.ts
function foo(src) {
  return `
}
Object.defineProperty(__vite_ssr_exports__, "foo", { enumerable: true, configurable: true, get(){ return foo }});

//# sourceMappingSource=vite-node
//# sourceMappingURL=data:application/json;charset=utf-8;base64,eyJ2ZXJzaW9uIjozLCJtYXBwaW5ncyI6IkFBQU8sU0FBUyxJQUFJLEtBQXFCO0FBQ3ZDLFNBQU8scURBQXFEO0FBQzlEO2lIQUFBIiwibmFtZXMiOltdLCJzb3VyY2VzIjpbIi9ob21lL3Byb2plY3RzL3ZpdGVzdC1kZXYtdml0ZXN0LWRxNnp6ZS9mb28udHMiXSwic291cmNlc0NvbnRlbnQiOlsiZXhwb3J0IGZ1bmN0aW9uIGZvbyhzcmM6IHN0cmluZyk6IHN0cmluZyB7XHJcbiAgcmV0dXJuIGAvLyMgc291cmNlTWFwcGluZ1VSTD1kYXRhOmFwcGxpY2F0aW9uL2pzb247YmFzZTY0LCR7c3JjfWA7XHJcbn1cclxuIl0sImZpbGUiOiIvaG9tZS9wcm9qZWN0cy92aXRlc3QtZGV2LXZpdGVzdC1kcTZ6emUvZm9vLnRzIn0=
```

The backtick is missing so it will resulting in `SyntaxError: Unexpected end of input` in run-time.

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-dq6zze?file=foo.ts

Reproduction steps:

1. `VITE_NODE_DEBUG_DUMP=true && vite-node ./foo.ts`

```bash
[vite-node] [debug] dump modules to /home/projects/vitest-dev-vitest-dq6zze/.vite-node/dump
SyntaxError: Unexpected end of input
[...]

PR #3379, opened 2023-05-16
fix(vite-node): don't remove sourcemap string in source code (#2918)

rxliuli:
fix https://github.com/vitest-dev/vitest/issues/2918


sheremet-va (changes requested):
We need to clear extra source maps to reduce the payload.

rxliuli:
> We need to clear extra source maps to reduce the payload.

This is the sourcemap string in the source code, which should not be removed from the original code. Also, I just updated the matching rule to not match the sourcemap in the string, but only remove the generated sourcemap.

For example, in the following code, the sourcemap string in foo function should not be deleted

```js
// /home/projects/vitest-dev-vitest-dq6zze/foo.ts
function foo(src) {
  return `//# sourceMappingURL=data:application/json;base64,${src}`;
}
Object.defineProperty(__vite_ssr_exports__, "foo", { enumerable: true, configurable: true, get(){ return foo }});

//# sourceMappingURL=data:application/json;charset=utf-8;base64,eyJ2ZXJzaW9uIjozLCJtYXBwaW5ncyI6IkFBQU8sU0FBUyxJQUFJLEtBQXFCO0FBQ3ZDLFNBQU8scURBQXFEO0FBQzlEO2lIQUFBIiwibmFtZXMiOltdLCJzb3VyY2VzIjpbIi9ob21lL3Byb2plY3RzL3ZpdGVzdC1kZXYtdml0ZXN0LWRxNnp6ZS9mb28udHMiXSwic291cmNlc0NvbnRlbnQiOlsiZXhwb3J0IGZ1bmN0aW9uIGZvbyhzcmM6IHN0cmluZyk6IHN0cmluZyB7XHJcbiAgcmV0dXJuIGAvLyMgc291cmNlTWFwcGluZ1VSTD1kYXRhOmFwcGxpY2F0aW9uL2pzb247YmFzZTY0LCR7c3JjfWA7XHJcbn1cclxuIl0sImZpbGUiOiIvaG9tZS9wcm9qZWN0cy92aXRlc3QtZGV2LXZpdGVzdC1kcTZ6emUvZm9vLnRzIn0=
```

sheremet-va:
> This is the sourcemap string in the source code, which should not be removed from the original code. Also, I just updated the matching rule to not match the sourcemap in the string, but only remove the generated sourcemap.

We should have a test that checks source map is removed correctly then.

rxliuli:
> > This is the sourcemap string in the source code, which should not be removed from the original code. Also, I just updated the matching rule to not match the sourcemap in the string, but only remove the generated sourcemap.
> 
> We should have a test that checks source map is removed correctly then.

Later I try to add a unit test to test this match

rxliuli:
> > This is the sourcemap string in the source code, which should not be removed from the original code. Also, I just updated the matching rule to not match the sourcemap in the string, but only remove the generated sourcemap.
> 
> We should have a test that checks source map is removed correctly then.

Relevant unit tests have been added

rxliuli:
It seems that the main branch has some unit test bugs that prevent CI from passing

image

sheremet-va (approved):

PR #3386, opened 2023-05-17
fix: stop spying on a method, when it's restored

### LINKED ISSUE #3127: mockRestore doesn't restore spied on properties to their original values
### Describe the bug

When using `vi.spyOn()` to create function mocks, the `spy.mockRestore()`, `vi.mockRestoreAllMocks()` doesn't actually restore the original descriptor of the spied-on object. This can lead to tests being written without ensuring that a method was already mocked, and deviates from [Jest's equivalent behavior](https://jestjs.io/docs/mock-function-api#mockfnmockrestore).

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-csdybj?file=test%2Fbasic.test.ts

```
test('should restore a spy', () => {
  const originalImpl = instance.fn1;
  const spy = vi.spyOn(instance, 'fn1');
  instance.fn1();
  spy.mockRestore();
  expect(instance.fn1).toEqual(originalImpl); // fails
});

test('should fail because it doesn\'t set up a spy', () => {
  instance.fn1();
  expect(instance.fn1).toHaveBeenCalledOnce(); // doesn't throw
});
```

### System Info

```shell
System:
    OS: macOS 12.6.3
    CPU: (10) arm64 Apple M1 Pro
    Memory: 88.91 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.19.1 - ~/.local/share/homebrew/bin/node
    Yarn: 3.4.1 - ~/.local/share/homebrew/bin/yarn
    npm: 8.19.3 - ~/.local/share/homebrew/bin/npm
  Browsers:
    Firefox: 102.2.0
    Safari: 16.3
```

envinfo doesn't seem to work with my project's setup (Yarn 3 monorepo), here are my vitest versions:

```
    "vite": "^4.2.1",
    "vitest": "^0.29.8"
```

### Used Package Manager

yarn

### Validations

sheremet-va:
Fixes #3127

### DIFF
--- a/packages/spy/src/index.ts
+++ b/packages/spy/src/index.ts
@@ -213,6 +213,13 @@ function enhanceSpy<TArgs extends any[], TReturns>(
   let onceImplementations: ((...args: TArgs) => TReturns)[] = []
   let implementationChangedTemporarily = false
 
+  function mockCall(this: unknown, ...args: any) {
+    instances.push(this)
+    invocations.push(++callOrder)
+    const impl = implementationChangedTemporarily ? implementation! : (onceImplementations.shift() || implementation || state.getOriginal() || (() => {}))
+    return impl.apply(this, args)
+  }
+
   let name: string = (stub as any).name
 
   stub.getMockName = () => name || 'vi.fn()'
@@ -237,6 +244,7 @@ function enhanceSpy<TArgs extends any[], TReturns>(
 
   stub.mockRestore = () => {
     stub.mockReset()
+    state.restore()
     implementation = undefined
     return stub
   }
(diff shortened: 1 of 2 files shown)


Lordfirespeed:
So what are vitest consumers meant to do when they want to restore original implementation and continue spying ?

sheremet-va:
> So what are vitest consumers meant to do when they want to restore original implementation and continue spying ?
> 
> `mockClear` doesn't affect implementation and `mockReset` restores an empty implementation (not the original)

Please, don’t comment on old and unrelated pull requests. If you have a question, formulate it and ask in discussions page or in discord.

PR #3392, opened 2023-05-17
fix!: improve globs

nickmccurdy:
This includes various improvements to globs focused on readability, maintainability, consistency, and Jest compatibility.

## Fixes
- Add `mts` and `cts` extensions to `coverage.exclude`
- Make test include patterns more consistent with [Jest's](https://jestjs.io/docs/configuration#testmatch-arraystring)*, specifically allowing these file paths and names:
  - `test`
  - `test-d`
  - `spec`
  - `spec-d`
  - `__tests__/*`

*This is a potentially back incompatible change, since tests that were not running before may now be run, especially if you were migrating from Jest. However, since it's reasonable to expect Vitest to be compatible with Jest, I'd consider this a bug fix rather than a breaking change. The release notes should highlight the fixes and back incompatibility related to Jest compatibility, as the other changes are basically just refactors.

## Future work
It might be a good idea to extract out some of the more complex and frequently used parts of our globs, such as the supported extensions for JS files, at some point.


packages/vitest/src/defaults.ts
```diff
@@ -1,10 +1,10 @@
 import type { BenchmarkUserOptions, ResolvedCoverageOptions, UserConfig } from './types'
 import { isCI } from './utils/env'
 
-export const defaultInclude = ['**/*.{test,spec}.{js,mjs,cjs,ts,mts,cts,jsx,tsx}']
+export const defaultInclude = ['**/*.{test,spec}.{,m,c}{j,t}sx?']
```
AriPerkkio, line 4: Is `x?` valid on glob or is it just for regexp?

This works:
```suggestion
export const defaultInclude = ['**/*.{test,spec}.{,m,c}{j,t}s{,x}']
```

nickmccurdy: Yea, I got the syntax wrong, though I think `?(x)` looks better

nickmccurdy:
I've pivoted this into making general improvements to globs, including fixing the original issue with `coverage.exclude`.

docs/config/index.md
```diff
@@ -81,14 +81,14 @@ All configuration options that are not supported inside a [workspace](/guide/wor
 ### include
 
 - **Type:** `string[]`
-- **Default:** `['**/*.{test,spec}.{js,mjs,cjs,ts,mts,cts,jsx,tsx}']`
+- **Default:** `['**/__tests__/**/*.?([cm])[jt]s?(x)', '**/?(*.){test,spec}.?([cm])[jt]s?(x)']`
 
 Files to include in the test run, using glob pattern.
```
AriPerkkio, line 86: I'm not sure if the readability of these patterns is clear enough anymore. Maybe we should have some kind of description here exactly as Jest has? https://jestjs.io/docs/configuration#testmatch-arraystring

Just to describe what exactly is considered as test file. Especially the new special handling for `__tests__` directory where even `__tests__/test-utils.ts` would be considered as test.

docs/config/index.md
```diff
@@ -81,14 +81,14 @@ All configuration options that are not supported inside a [workspace](/guide/wor
 ### include
 
 - **Type:** `string[]`
-- **Default:** `['**/*.{test,spec}.{js,mjs,cjs,ts,mts,cts,jsx,tsx}']`
+- **Default:** `['**/__tests__/**/*.?([cm])[jt]s?(x)', '**/?(*.){test,spec}.?([cm])[jt]s?(x)']`
```
AriPerkkio, line 84: The `?([cm])` parts could be `?(c|m)`. Same applied to other patterns as well.

```suggestion
- **Default:** `['**/__tests__/**/*.?(c|m)[jt]s?(x)', '**/?(*.){test,spec}.?(c|m)[jt]s?(x)']`
```

nickmccurdy: I was picking visual consistency over brevity, though I'm not against this change

docs/config/index.md
```diff
@@ -81,14 +81,14 @@ All configuration options that are not supported inside a [workspace](/guide/wor
...
 
 Files to include in the test run, using glob pattern.
 
 ### exclude
 
 - **Type:** `string[]`
-- **Default:** `['**/node_modules/**', '**/dist/**', '**/cypress/**', '**/.{idea,git,cache,output,temp}/**', '**/{karma,rollup,webpack,vite,vitest,jest,ava,babel,nyc,cypress,tsup,build}.config.*']`
+- **Default:** `['**/node_modules/**', '**/dist/**', '**/cypress/**', '**/.{idea,git,cache,output,temp}/**', '**/{karma,rollup,webpack,vite?(st),jest,ava,babel,nyc,cypress,tsup,build}.config.*']`
```
sheremet-va, line 91: I don't see any reason to change vite,vitest and bench,benchmark

It's fine with extensions, but here readability suffers greatly

docs/config/index.md
nickmccurdy: @sheremet-va I noticed you resolved this discussion. Do you think it's unnecessary? I was going to document it, just haven't had time yet.

sheremet-va: I thought it was implemented, sorry. We should definitely document it.

nickmccurdy:
Is anyone working on docs for this? If not I can.

PR #3415, opened 2023-05-20
fix: dot reporter scrollback buffer spam

### LINKED ISSUE #3384: Dot reporter spams terminal if all the dots don't fit in the given height
### Describe the bug

If the number of tests is high enough that the dot reporter can't fit all the dots on the terminal at once, it'll keep pushing lines onto stdout instead of clearing it. This also introduces some stuttering if the terminal fails to keep up with the continuous logging.

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-1pwqpk?file=vite.config.ts,test%2Fbasic.test.ts&initialPath=__vitest__/

### System Info

```shell
(Stackblitz)

System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    @vitest/ui: latest => 0.31.0
```

### Used Package Manager

npm

### Validations

gtm-nayan:
fixes #3384

### DIFF
--- a/packages/vitest/src/node/reporters/renderers/dotRenderer.ts
+++ b/packages/vitest/src/node/reporters/renderers/dotRenderer.ts
@@ -33,6 +33,11 @@ function render(tasks: Task[], width: number): string {
   let currentTasks = 0
   let previousLineWidth = 0
   let output = ''
+
+  // The log-update uses various ANSI helper utilities, e.g. ansi-warp, ansi-slice,
+  // when printing. Passing it hundreds of single characters containing ANSI codes reduces
+  // performances. We can optimize it by reducing amount of ANSI codes, e.g. by coloring
+  // multiple tasks at once instead of each task separately.
   const addOutput = () => {
     const { char, color } = currentIcon
     const availableWidth = width - previousLineWidth
@@ -41,6 +46,8 @@ function render(tasks: Task[], width: number): string {
       previousLineWidth += currentTasks
     }
     else {
+      // We need to split the line otherwise it will mess up log-update's height calculation
+      // and spam the scrollback buffer with dots.
       let buf = `${char.repeat(availableWidth)}\n`
       let remaining = currentTasks - availableWidth
       buf += `${char.repeat(width)}\n`.repeat(Math.floor(remaining / width))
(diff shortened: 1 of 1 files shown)


packages/vitest/src/node/reporters/renderers/dotRenderer.ts
```diff
@@ -27,54 +27,68 @@ function getIcon(task: Task) {
...
+      buf += `${char.repeat(width)}\n`.repeat(Math.floor(remaining / width))
+      remaining %= width
+
+      if (remaining > 0) {
+        buf += char.repeat(remaining)
+        previousLineWidth = remaining
+      }
+      else {
+        previousLineWidth = 0
+      }
```
AriPerkkio, line 62: While this works fine I found it a bit difficult to read. I had to do some refactoring to it in order to understand what exactly is happening. I came up with following refactoring in the end. Would it be more readable?

```suggestion
      let buf = `${char.repeat(availableWidth)}\n`
      const remaining = currentTasks - availableWidth

      // Add full rows first
      const fullRows = Math.floor(remaining / width)
      if (fullRows > 0)
        buf += `${char.repeat(width)}\n`.repeat(fullRows)

      // Add remaining dots which don't make a full row
      const partialRow = remaining % width
      if (partialRow > 0) {
        buf += `${char.repeat(partialRow)}`
        previousLineWidth = partialRow
      }
      else {
        previousLineWidth = 0
      }
```

AriPerkkio (changes requested):
These changes also improve performance as less work is expected from `log-update`'s dependencies. I ran the examples used in https://github.com/vitest-dev/vitest/pull/3006 to verify this. About 9s to 5s. I also verified manually that the mentioned bug is fixed by these changes. Great work overall @gtm-nayan!

packages/vitest/src/node/reporters/renderers/dotRenderer.ts
gtm-nayan: Sounds good. I've pushed the suggested changes with a few tweaks.

AriPerkkio (approved):

PR #3418, opened 2023-05-21
fix(vite-node): circular import stuck

Dunqing:
Both `main.js` and `mock(./src/B)` will import `A`, resulting in a pending state all the time.

fix: https://github.com/vitest-dev/vitest/issues/2008#issuecomment-1542554048

### DIFF
--- a/examples/mocks/src/main.js
+++ b/examples/mocks/src/main.js
@@ -1,5 +1,10 @@
 import { funcA } from './A'
+import { funcB } from './B'
 
 export function main() {
   return funcA()
+}
+
+export function mainB() {
+  return funcB()
 }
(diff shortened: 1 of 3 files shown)


AriPerkkio:
This seems to fix the circular import bug mentioned in https://github.com/vitest-dev/vitest/issues/2008#issuecomment-1542554048.

Dunqing:
> This seems to fix the circular import bug mentioned in [#2008 (comment)](https://github.com/vitest-dev/vitest/issues/2008#issuecomment-1542554048).

Yes, same issue

sheremet-va:
This breaks deep circular imports because `mod.importers.has(importee)` always returns `true`. We are reverting this change in #3479

PR #3430, opened 2023-05-22
fix(vite-node): coerce to string in import(dep)

### LINKED ISSUE #3429: __vite_ssr_dynamic_import__ doesn't coerce import(specifier) to a string
### Describe the bug

If you pass in a non-string into an import statement when using Vitest, you'll get a cryptic "id.startsWith is not a function" message. I think this is due to the custom `__vite_ssr_dynamic_import__()` function that seems to replace all `import()` stuff not properly doing ``myArg = `${myArg}`;`` like browsers and Node.js do natively to make sure the arg is a true string.

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-mamgrv?file=test%2Fbasic.test.ts

image

### System Info

```shell
~/projects/vitest-dev-vitest-mamgrv 1m 15s
❯ npx envinfo --system --npmPackages '{vitest,@vitest/*,vite,@vitejs/*}' --binaries --browsers
success Install finished in 0.928s


  System:
    OS: Linux 5.0 undefined
    CPU: (4) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    @vitest/ui: latest => 0.31.1 
    vite: latest => 4.3.8 
    vitest: latest => 0.31.1 
```
```


### Used Package Manager

npm

### Validations

### PR DESCRIPTION
**Pull request by jcbhmr**
This PR would...
- Coerce `dep` to string using the string coercion algorithm in `import(dep)`

Need to use `` `${s}` `` instead of `"" + s` in order to get the right hint in `Symbol.toPrimitive`

> ```js
> const obj2 = {
>   [Symbol.toPrimitive](hint) {
>     if (hint === "number") {
>       return 10;
>     }
>     if (hint === "string") {
>       return "hello";
>     }
>     return true;
>   },
> };
> console.log(+obj2); // 10        — hint is "number"
> console.log(`${obj2}`); // "hello"   — hint is "string"
> console.log(obj2 + ""); // "true"    — hint is "default"
> ```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toPrimitive#description

closes  #3429 

-----
<a href="https://stackblitz.com/~/github.com/jcbhmr/vitest/tree/jcbhmr%2Fissue3429">Review PR in StackBlitz Codeflow</a> _Submitted with [StackBlitz Codeflow](https://stackblitz.com/~/github.com/jcbhmr/vitest/tree/jcbhmr%2Fissue3429)._

### DIFF
--- a/packages/vite-node/src/client.ts
+++ b/packages/vite-node/src/client.ts
@@ -267,6 +267,7 @@ export class ViteNodeRunner {
     const mod = this.moduleCache.getByModuleId(moduleId)
 
     const request = async (dep: string) => {
+      dep = `${dep}`
       const [id, depFsPath] = await this.resolveUrl(dep, fsPath)
       return this.dependencyRequest(id, depFsPath, callstack)
     }
--- a/packages/vite-node/src/client.ts
+++ b/packages/vite-node/src/client.ts
@@ -267,8 +267,7 @@ export class ViteNodeRunner {
     const mod = this.moduleCache.getByModuleId(moduleId)
 
     const request = async (dep: string) => {
-      dep = `${dep}`
-      const [id, depFsPath] = await this.resolveUrl(dep, fsPath)
+      const [id, depFsPath] = await this.resolveUrl(`${dep}`, fsPath)
       return this.dependencyRequest(id, depFsPath, callstack)
     }
 
--- a/test/core/test/imports.test.ts
+++ b/test/core/test/imports.test.ts
@@ -52,6 +52,12 @@ test('data with dynamic import works', async () => {
   expect(hi).toBe('hi')
 })
 
+test('dynamic import coerces to string', async () => {
+  const dataUri = 'data:text/javascript;charset=utf-8,export default "hi"'
+  const { default: hi } = await import({ toString: () => dataUri } as string)
+  expect(hi).toBe('hi')
+})
+
 test('dynamic import has Module symbol', async () => {
   const stringTimeoutMod = await import('./../src/timeout')
 

### CODE REVIEW

File: packages/vite-node/src/client.ts [hunk 1]
```diff
@@ -267,6 +267,7 @@ export class ViteNodeRunner {
     const mod = this.moduleCache.getByModuleId(moduleId)
 
     const request = async (dep: string) => {
+      dep = `${dep}`
```
sheremet-va, line 270:
```suggestion

```

packages/vite-node/src/client.ts
```diff
@@ -267,6 +267,7 @@ export class ViteNodeRunner {
...
       const [id, depFsPath] = await this.resolveUrl(dep, fsPath)
```
sheremet-va, line 271:
```suggestion
      const [id, depFsPath] = await this.resolveUrl(`${dep}`, fsPath)
```

sheremet-va (changes requested):

sheremet-va (changes requested):
Can you add a test case for this, please?

jcbhmr:
Added a test case:

```ts
test('dynamic import coerces to string', async () => {
  const dataUri = 'data:text/javascript;charset=utf-8,export default "hi"'
  const { default: hi } = await import({ toString: () => dataUri } as string)
  expect(hi).toBe('hi')
})
```

PR #3436, opened 2023-05-23
test: reduce usage of `execa`, add `runVitest` test utils

AriPerkkio:
Using `execa` on CI is slow and flaky. This PR removes usage of `execa` from every test where it's possible and uses `startVitest` Node API instead. This is done by adding `test/test-utils` package which provides centralized abstractions for running `startVitest`. As some of the tests still require running Vitest through CLI, there is also utility method `runVitestCli` which is a simple abstraction over `execa` and supports interactive watch-mode simulation. 

Goal is to make CI stable and faster, and to not have multiple duplicated implementations for running `npx vitest` in test files/projects. When adding new tests the `runVitest` should be used by default. If CLI usage is explicitly required, use `runVitestCli` instead.

There's quite much changes so I've rebased the commits clearly for possible commit-by-commit review.


sheremet-va:
Not sure about speed improvements. CI from #3430 is faster 🤔

PR #3453, opened 2023-05-26
feat(coverage): add `reportOnFailure` option

### LINKED ISSUE #2908: Don't generate coverage reports when tests fail.

A test coverage report (either text or generated html/json) does not have a lot of value if the test run itself has failed. The text output (if large enough) can cover the entire screen and hide the fact that the test has failed.

### Suggested solution

If the test failed, don't show or generate any coverage report.

### Alternative

A configuration flag could be useful as well, for those that want the older behavior (or the other way around).

```
test: {
  coverage: {
    generateOnFailure: true
  }
}
```

### Additional context

_No response_

### Validations

### LINKED ISSUE #3388: skip coverage when spec runner bails

When there is a failure and --bail is set to any value the coverage gets generated and in our project always fails to meet thresholds.

### Suggested solution

do not bother with coverage when spec runner bailed

### Alternative

_No response_

### Additional context

_No response_

### Validations

AriPerkkio:
- Closes #2908
- Closes #3388

Adds new option `coverage.reportOnFailure` which can be used to disabled coverage report generation when tests fail. Currently this is disabled by default but we should enable this by default some day.

TODO: Tests


docs/config/index.md
```diff
@@ -778,6 +778,15 @@ The reporter has three different types:
 
 Since Vitest 0.31.0, you can check your coverage report in Vitest UI: check [Vitest UI Coverage](/guide/coverage#vitest-ui) for more details.
 
+#### coverage.reportOnFailure
+
+- **Type:** `boolean`
+- **Default:** `true`
+- **Available for providers:** `'c8' | 'istanbul'`
+- **CLI:** `--coverage.reportOnFailure`, `--coverage.reportOnFailure=false`
```
sheremet-va, line 786:
```suggestion
- **CLI:** `--coverage.reportOnFailure`, `--coverage.reportOnFailure=false`
- **Version:** Since Vitest 0.31.2
```

sheremet-va (approved):

PR #3455, opened 2023-05-27
fix(runner): suite timeout does not take effect

### LINKED ISSUE #3454: Setting a `timeout` to a `describe` does not have an effect in `0.31`
### Describe the bug

It seems as if something broke when upgrading from `0.30.1` to `0.31.1` with the test timeouts. Passing `{ timeout: XXX }` to `describe` does not have an effect any longer, the default `testTimeout` is always used.

### Reproduction

[Run the foll](https://stackblitz.com/edit/vitest-dev-vitest-mmxrtd?file=test%2Fsuite.test.ts&initialPath=__vitest__/)

Short description:
```ts
describe(
  "the submitted issue",
  () => {
    test("true is true after 10s", async () => {
      await new Promise((resolve) => setTimeout(resolve, 10000));

      expect(true).toBe(true);
    });
  },
  {
    timeout: 30000,
  }
);
```

I'm expecting the test to pass after approx 10s instead it fails after 5s with the error:
```sh
Error: Test timed out in 5000ms.

```

### System Info

```shell
System:
    OS: macOS 13.4
    CPU: (12) arm64 Apple M2 Max
    Memory: 52.25 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.16.0 - ~/.nvm/versions/node/v18.16.0/bin/node
    npm: 9.5.1 - ~/.nvm/versions/node/v18.16.0/bin/npm
  Browsers:
    Chrome: 113.0.5672.126
    Safari: 16.5
  npmPackages:
    @vitest/coverage-istanbul: ^0.31.1 => 0.31.1
    vite: ^4.3.9 => 4.3.9
    vitest: ^0.31.1 => 0.31.1
```

### Used Package Manager

pnpm

### Validations

btea:
fix #3454

### DIFF
--- a/test/core/test/timeout.spec.ts
+++ b/test/core/test/timeout.spec.ts
@@ -8,3 +8,10 @@ describe('suite timeout', () => {
 }, {
   timeout: 6000,
 })
+
+describe('suite timeout simple input', () => {
+  test('true is true after 5100ms', async () => {
+    await new Promise(resolve => setTimeout(resolve, 5100))
+    expect(true).toBe(true)
+  })
+}, 6000)
--- a/test/core/test/timeout.spec.ts
+++ b/test/core/test/timeout.spec.ts
@@ -1,0 +1,10 @@
+import { describe, expect, test } from 'vitest'
+
+describe('suite timeout', () => {
+  test('true is true after 5100ms', async () => {
+    await new Promise(resolve => setTimeout(resolve, 5100))
+    expect(true).toBe(true)
+  })
+}, {
+  timeout: 6000,
+})
--- a/test/fails/test/__snapshots__/runner.test.ts.snap
+++ b/test/fails/test/__snapshots__/runner.test.ts.snap
@@ -36,8 +36,9 @@ TypeError: failure
 TypeError: failure"
 `;
 
-exports[`should fails > test-timeout.test.ts > test-timeout.test.ts 1`] = `
-"Error: Test timed out in 100ms.
+exports[`should fail test-timeout.test.ts > test-timeout.test.ts 1`] = `
+"Error: Test timed out in 200ms.
+Error: Test timed out in 100ms.
 Error: Test timed out in 10ms."
 `;
 
--- a/test/fails/test/__snapshots__/runner.test.ts.snap
+++ b/test/fails/test/__snapshots__/runner.test.ts.snap
@@ -36,7 +36,10 @@ TypeError: failure
 TypeError: failure"
 `;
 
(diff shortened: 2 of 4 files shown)


sheremet-va (commented):
Please, add a test case for this

test/core/test/timeout.spec.ts
```diff
@@ -0,0 +1,10 @@
+import { describe, expect, test } from 'vitest'
+
+describe('suite timeout', () => {
+  test('true is true after 5100ms', async () => {
+    await new Promise(resolve => setTimeout(resolve, 5100))
+    expect(true).toBe(true)
+  })
+}, {
+  timeout: 6000,
+})
```
marcus13371337, line 10: This is also a valid syntax, which might be worth adding a test for (can't see that this case is handled in your code changes). However, it might be parsed in a different, but worth adding a test for it IMO!
```suggestion
describe('suite timeout simple input', () => {
  test('true is true after 5100ms', async () => {
    await new Promise(resolve => setTimeout(resolve, 5100))
    expect(true).toBe(true)
  })
}, 6000)
```

sheremet-va: Yes, we should also add a failing test case for this in `test/failing`

sheremet-va (changes requested):

test/core/test/timeout.spec.ts
btea: Done.

test/fails/test/__snapshots__/runner.test.ts.snap
```diff
@@ -36,7 +36,10 @@ TypeError: failure
 TypeError: failure"
 `;
 
-exports[`should fails > test-timeout.test.ts > test-timeout.test.ts 1`] = `"Error: Test timed out in 10ms."`;
+exports[`should fails > test-timeout.test.ts > test-timeout.test.ts 1`] = `
+"Error: Test timed out in 100ms.
+Error: Test timed out in 10ms."
```
sheremet-va, line 41: Shouldn't this have 3 errors?

sheremet-va (changes requested):

test/fails/test/__snapshots__/runner.test.ts.snap
sheremet-va: Maybe we should use another timer than. :)

btea: Ok, let me modify it.

sheremet-va (approved):

marcus13371337:
Thanks for fixing this issue so quickly @btea 🎉

PR #3469, opened 2023-05-30
feat: exclude e2e tests folder by default

### LINKED ISSUE #3102: exclude default setting should exclude e2e tests folder

Since a generally accepted good practice* is to put unit tests next to the code they test, and e2e tests in a separate folder, excluding such a folder would be a better default setting.

In many projects I see that all the configuration file does is modify `include` or `exclude` to exclude e2e tests. This change would allow an additional configuration file to be avoided.

I can do the PR.

### Suggested solution

Add any of the following paths to the default exclude configuration:

**Option 1**: `'**/tests/**'`
This is the most opinionated option. It looks pretty clean, but I can see how some people may have put unit tests inside folders with that name. With this option, they would have to modify the config or rename the folder to something like `/unit-test/`.

**Option 2**: `'**/e2e-tests/**'`
Non-opinionated, and not potentially conflicting. 

**Option 3**:  `'**/tests/**', '**/e2e-tests/**'`
Despite the potential conflict explained in option 1... the goal of default settings is to respond to the most popular settings, so possibly this is also a good option.

### Alternative

_No response_

### Additional context

*Validated in all the discussions I found on the subject. For example [here](https://www.reddit.com/r/node/comments/kesbv9/comment/gg4tpn7/?utm_source=reddit&utm_medium=web2x&context=3), or [here](https://kit.svelte.dev/docs/project-structure).

### Validations

GermanJablo:
close #3102

### DIFF
--- a/packages/vitest/src/defaults.ts
+++ b/packages/vitest/src/defaults.ts
@@ -2,7 +2,7 @@ import type { BenchmarkUserOptions, ResolvedCoverageOptions, UserConfig } from '
 import { isCI } from './utils/env'
 
 export const defaultInclude = ['**/*.{test,spec}.{js,mjs,cjs,ts,mts,cts,jsx,tsx}']
-export const defaultExclude = ['**/{node_modules,dist,cypress,tests,e2e-tests}/**', '**/.{idea,git,cache,output,temp}/**', '**/{karma,rollup,webpack,vite,vitest,jest,ava,babel,nyc,cypress,tsup,build}.config.*']
+export const defaultExclude = ['**/{node_modules,dist,cypress,playwright,e2e-tests}/**', '**/.{idea,git,cache,output,temp}/**', '**/{karma,rollup,webpack,vite,vitest,jest,ava,babel,nyc,cypress,tsup,build}.config.*']
 export const benchmarkConfigDefaults: Required<Omit<BenchmarkUserOptions, 'outputFile'>> = {
   include: ['**/*.{bench,benchmark}.{js,mjs,cjs,ts,mts,cts,jsx,tsx}'],
   exclude: defaultExclude,
--- a/packages/vitest/src/defaults.ts
+++ b/packages/vitest/src/defaults.ts
@@ -2,7 +2,7 @@ import type { BenchmarkUserOptions, ResolvedCoverageOptions, UserConfig } from '
 import { isCI } from './utils/env'
 
 export const defaultInclude = ['**/*.{test,spec}.{js,mjs,cjs,ts,mts,cts,jsx,tsx}']
-export const defaultExclude = ['**/node_modules/**', '**/dist/**', '**/cypress/**', '**/.{idea,git,cache,output,temp}/**', '**/{karma,rollup,webpack,vite,vitest,jest,ava,babel,nyc,cypress,tsup,build}.config.*']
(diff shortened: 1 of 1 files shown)


packages/vitest/src/defaults.ts
```diff
@@ -2,7 +2,7 @@ import type { BenchmarkUserOptions, ResolvedCoverageOptions, UserConfig } from '
 import { isCI } from './utils/env'
 
 export const defaultInclude = ['**/*.{test,spec}.{js,mjs,cjs,ts,mts,cts,jsx,tsx}']
-export const defaultExclude = ['**/node_modules/**', '**/dist/**', '**/cypress/**', '**/.{idea,git,cache,output,temp}/**', '**/{karma,rollup,webpack,vite,vitest,jest,ava,babel,nyc,cypress,tsup,build}.config.*']
+export const defaultExclude = ['**/{node_modules,dist,cypress,tests,e2e-tests}/**', '**/.{idea,git,cache,output,temp}/**', '**/{karma,rollup,webpack,vite,vitest,jest,ava,babel,nyc,cypress,tsup,build}.config.*']
```
sheremet-va, line 5: Why does it exclude test folder now?

GermanJablo: The explanation is in https://github.com/vitest-dev/vitest/issues/3102

sheremet-va: How exactly is it a good choice to exclude every test?

GermanJablo: It's not excluding every test, it's excluding the tests folder, because unit tests are not usually grouped in a folder, but placed next to the file they test, unlike e2e tests. As I explained in the issue:

> Since a generally accepted good practice* is to put unit tests next to the code they test, and e2e tests in a separate folder, excluding such a folder would be a better default setting.
> 
> In many projects I see that all the configuration file does is modify include or exclude to exclude e2e tests.

sheremet-va: I don't think this is true. Even in Vitest tests are grouped inside a tests folder

packages/vitest/src/defaults.ts
```diff
@@ -2,7 +2,7 @@ import type { BenchmarkUserOptions, ResolvedCoverageOptions, UserConfig } from '
 import { isCI } from './utils/env'
 
 export const defaultInclude = ['**/*.{test,spec}.{js,mjs,cjs,ts,mts,cts,jsx,tsx}']
-export const defaultExclude = ['**/node_modules/**', '**/dist/**', '**/cypress/**', '**/.{idea,git,cache,output,temp}/**', '**/{karma,rollup,webpack,vite,vitest,jest,ava,babel,nyc,cypress,tsup,build}.config.*']
+export const defaultExclude = ['**/{node_modules,dist,cypress,playwright,e2e-tests}/**', '**/.{idea,git,cache,output,temp}/**', '**/{karma,rollup,webpack,vite,vitest,jest,ava,babel,nyc,cypress,tsup,build}.config.*']
```
GermanJablo, line 5: It's okay. What do you think now?  @sheremet-va

packages/vitest/src/defaults.ts
GermanJablo: @sheremet-va Alright. I have removed that path. Please see my comment below. Thank you!

PR #3479, opened 2023-05-31
revert: "fix(vite-node): circular import stuck" #3418

sheremet-va:
Reverts #3418

`mod.importers.has(importee)` is always true because we add "importee" to importers a few lines prior

### DIFF
--- a/examples/mocks/src/main.js
+++ b/examples/mocks/src/main.js
@@ -1,10 +1,5 @@
 import { funcA } from './A'
-import { funcB } from './B'
 
 export function main() {
   return funcA()
-}
-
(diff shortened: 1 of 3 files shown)


antfu (approved):

antfu:
We can do a quick hotfix if you think it's necessary.

sheremet-va:
> We can do a quick hotfix release if you think it's necessary.

Yes, I think we should do a quick hotfix today when CI passes.

AriPerkkio:
The tests in `examples/mocks/test/circular.spec.ts` should get stuck after this change. Need to revert / skip those ones too.

PR #3480, opened 2023-05-31
fix(vite-node): circular import stuck

Dunqing:
related PR: #3418 #3479.

I'm sorry about the last PR.

The PR has been tested at https://github.com/ant-design/ant-design/pull/42506 and works. Also, the deep cycle import works well.

### DIFF
--- a/packages/vite-node/src/client.ts
+++ b/packages/vite-node/src/client.ts
@@ -194,7 +194,7 @@ export class ViteNodeRunner {
     const getStack = () => `stack:\n${[...callstack, fsPath].reverse().map(p => `  - ${p}`).join('\n')}`
 
     // check circular dependency
-    if (callstack.includes(fsPath) || callstack.some(c => this.moduleCache.get(c).importers?.has(fsPath))) {
+    if (callstack.includes(fsPath) || callstack.some(c => this.moduleCache.get(c).importers?.has(fsPath)) || mod.importers.has(importee)) {
       if (mod.exports)
         return mod.exports
     }
--- a/packages/vite-node/src/client.ts
+++ b/packages/vite-node/src/client.ts
@@ -99,6 +99,7 @@ export class ModuleCacheMap extends Map<string, ModuleCache> {
     delete mod.promise
     delete mod.exports
     mod.importers.clear()
+    mod.imports.clear()
     return true
   }
 
--- a/packages/vite-node/src/client.ts
+++ b/packages/vite-node/src/client.ts
@@ -61,23 +61,23 @@ export class ModuleCacheMap extends Map<string, ModuleCache> {
   update(fsPath: string, mod: Partial<ModuleCache>) {
     fsPath = this.normalizePath(fsPath)
     if (!super.has(fsPath))
-      super.set(fsPath, mod)
+      this.setByModuleId(fsPath, mod)
     else
       Object.assign(super.get(fsPath) as ModuleCache, mod)
     return this
   }
 
-  setByModuleId(modulePath: string, mod: ModuleCache) {
-    return super.set(modulePath, mod)
+  setByModuleId(modulePath: string, mod: Partial<ModuleCache>) {
+    return super.set(modulePath, { imports: new Set(), importers: new Set(), ...mod })
   }
 
-  set(fsPath: string, mod: ModuleCache) {
+  set(fsPath: string, mod: Partial<ModuleCache>) {
     return this.setByModuleId(this.normalizePath(fsPath), mod)
   }
 
   getByModuleId(modulePath: string): ModuleCache {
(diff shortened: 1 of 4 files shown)


packages/vite-node/src/client.ts
```diff
@@ -205,8 +204,11 @@ export class ViteNodeRunner {
 
     try {
       // cached module
-      if (mod.promise)
+      if (mod.promise) {
+        if (Array.from(mod.imports?.values() || []).some(i => mod.importers!.has(i)))
```
sheremet-va, line 208: This looks like a weird place to check this. Can you add comments explaining your changes?

Also, can you simplify this code to look cleaner? For example,
```ts
const importers = mod.importers // TS knows they are defined, so there is no need to use `!`
const imports = Array.from(mod.imports?.values() || [])
const circular = imports.some(i => importers.has(i))
```
The naming is what I understood from this. It can be different if needed

packages/vite-node/src/client.ts
```diff
@@ -269,6 +271,10 @@ export class ViteNodeRunner {
 
     const request = async (dep: string) => {
       const [id, depFsPath] = await this.resolveUrl(`${dep}`, fsPath)
+      const depMod = this.moduleCache.getByModuleId(depFsPath)
+      ;(depMod.importers ??= new Set()).add(moduleId)
+      ;(mod.imports ??= new Set()).add(depFsPath)
```
sheremet-va, line 276: Can you simplify this to looks cleaner?

```ts       
depMod.importers ??= new Set()
depMod.importers.add(moduleId)

mod.imports ??= new Set()
mod.imports.add(depFsPath)
```

We can also move this logic to `ModuleCache` with a nice name:

```ts
addImporter(mod, id) {
  mod ??= new Set()
  mod.importers.add(id)
}
addImport(mod, id) {
	mod.imports ??= new Set()
	mod.imports.add(id)
}
```

sheremet-va (changes requested):

packages/vite-node/src/client.ts
```diff
@@ -61,23 +61,23 @@ export class ModuleCacheMap extends Map<string, ModuleCache> {
...
+      this.setByModuleId(fsPath, mod)
     else
       Object.assign(super.get(fsPath) as ModuleCache, mod)
     return this
   }
 
-  setByModuleId(modulePath: string, mod: ModuleCache) {
-    return super.set(modulePath, mod)
+  setByModuleId(modulePath: string, mod: Partial<ModuleCache>) {
+    return super.set(modulePath, { imports: new Set(), importers: new Set(), ...mod })
```
sheremet-va, line 71: Can we not create a new object here?

sheremet-va (changes requested):

packages/vite-node/src/client.ts
sheremet-va: Maybe instead of `.set({})` inside of `.get()` we can just provide `{ imports: new Set(), importers: new Set() }`?

packages/vite-node/src/client.ts
```diff
@@ -98,7 +98,7 @@ export class ModuleCacheMap extends Map<string, ModuleCache> {
     delete mod.resolving
     delete mod.promise
     delete mod.exports
-    mod.importers?.clear()
+    mod.importers.clear()
```
sheremet-va, line 101: imports should also be cleared now

sheremet-va (changes requested):

sheremet-va (approved):

packages/vite-node/src/client.ts
```diff
@@ -269,6 +277,10 @@ export class ViteNodeRunner {
 
     const request = async (dep: string) => {
       const [id, depFsPath] = await this.resolveUrl(`${dep}`, fsPath)
+      const depMod = this.moduleCache.getByModuleId(depFsPath)
+      depMod.importers.add(moduleId)
```
sheremet-va, line 281: I think this is already done on line 200?

sheremet-va (changes requested):

packages/vite-node/src/client.ts
Dunqing: This is not always the case. The `dependencyRequest` is an async function that will not be executed immediately.

sheremet-va: Why does it matter? There is nothing async between this calls

Dunqing: The current test case will be stuck if I remove this line

Dunqing: image

The execution sequence looks like this

sheremet-va: Interesting 🤔 Ok, then, should be fine

PR #3489, opened 2023-06-01
fix(runner): ensure Vitest is deduped

### LINKED ISSUE #3426: Snapshots do not work with Yarn PnP
### Describe the bug

```ts
expect(...).toMatchInlineSnapshot('...');
```

```
> TypeError: Cannot read properties of undefined (reading 'match')
```

### Reproduction

- Create a Yarn-based project with a coupe of workspaces (can be the root workspace + a child workspace)
- Ensure that Yarn PnP is enabled (`yarn config set pnp`)
- Add a couple of Vitest project configurations + the root Vitest (workspace) config, a few tests with inline snapshots
- Run `yarn vitest`

It only works if all the peer dependencies in all Vitest workspaces match exactly, otherwise Yarn creates a separate virtual folder for each `vitest` instance which breaks the way Vitest works with inline snapshots. It looks like snapshot initialization happens in one "vitest" instance, while the test result matching uses a different "vitest" instance.

You can check virtual packages using `yarn info --virtuals vitest` command, or by looking up for `vitest-virtual` occurrences in `.pnp.cjs`.

Reproduction: https://github.com/koistya/vitest-workspaces
Referene: https://yarnpkg.com/advanced/pnp-spec#virtual-folders

### System Info

```shell
System:
    OS: macOS 11.6
    CPU: (8) arm64 Apple M1
    Memory: 141.69 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 18.16.0 - /opt/homebrew/bin/node
    Yarn: 4.0.0-rc.44 - /opt/homebrew/bin/yarn
    npm: 9.5.1 - /opt/homebrew/bin/npm
    Watchman: 2023.04.24.00 - /opt/homebrew/bin/watchman
```

### Used Package Manager

yarn

### Validations

sheremet-va:
Fixes #3426


sheremet-va:
> With these changes I'm seeing the same error message as before:

Hm, workspace project `root` by default is still cwd 🤔

sheremet-va:
> Using the reproduction setup from [#3301 (comment)](https://github.com/vitest-dev/vitest/pull/3301#issuecomment-1534420250).

Works correctly if I update `vitest.workspace.js` to the correct glob pattern: `packages/*/{vite,vitest}.config.ts`, so the idea is correct, but it doesn't work if users don't use workspaces :(

Another solution would be to always resolve Vitest to the root one ignoring package.json.

sheremet-va:
@AriPerkkio I found an elegant solution 🙈

sheremet-va:
> @AriPerkkio I found an elegant solution 🙈

Turns out it doesn't work with workspaces 🙈
Because we have a different server, we cannot use `resolve.dedupe` there

PR #3490, opened 2023-06-01
feat: support ssr optimizer

sheremet-va:


sheremet-va:
"experimental" optimizer is not under semver, so we can merge it in a minor.

PR #3491, opened 2023-06-01
feat!: transformMode affects only test files, not regular files

### LINKED ISSUE #1918: vue-router's useRoute is not correctly mocked
### Describe the bug

When using vitest-suggested testing libraries, I can't successfully mock `useRoute` and `useRouter` from the `vue-router` package.

I'm using:
- jsdom
- @testing-library/vue

Any time I run a component test with a mocked `useRoute`, the mock returns undefined, and Vue warns about missing injections.

```
...
[Vue warn]: injection "Symbol(route location)" not found. 
  at <HelloWorld ref="VTU_COMPONENT" > 
  at <VTUROOT>
...
TypeError: Cannot read properties of undefined (reading 'path')
     19| 
     20|   <div class="card">
     21|     <button type="button" @click="count++">count is {{ count }}</button>
       |                     ^
     22|     <p>
     23|       Edit
```

### Reproduction

See a minimal reproduction, using a fresh `yarn create vite` project here: https://github.com/hidde-jan/vitest-use-route-example/blob/main/src/components/__test__/HelloWorld.test.ts

### System Info

```shell
System:
    OS: macOS 12.3.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 76.22 MB / 32.00 GB
    Shell: 3.3.1 - /opt/homebrew/bin/fish
  Binaries:
    Node: 16.14.0 - ~/.nodenv/versions/16.14.0/bin/node
    Yarn: 1.22.15 - ~/.nodenv/versions/16.14.0/bin/yarn
    npm: 8.3.1 - ~/.nodenv/versions/16.14.0/bin/npm
  Browsers:
    Chrome: 104.0.5112.101
    Firefox: 102.0.1
    Safari: 15.4
  npmPackages:
    @vitejs/plugin-vue: ^3.0.3 => 3.0.3 
    vite: ^3.0.7 => 3.0.9 
    vitest: ^0.22.1 => 0.22.1 
```
```


### Used Package Manager

yarn

### Validations

### PR DESCRIPTION
**Pull request by sheremet-va**
Supporting different transform modes inside the same test process causes a mismatch with resolved IDs which causes all sorts of side effects (namely, different global instances in Solid/Preact or Vue Router).

After this PR `transformMode` is removed and replaced with `testTransformMode` and now it affects only test files, not just any imported file. By default, Vitest relies on the environment to get the transform mode.

If you are using vite-node directly and rely on this property, you can use `test.server.transformMode` directly.

Fixes #1918

### CODE REVIEW

**Comment by antfu** [comment]
I don't have a strong opinion on this, but I am not very sure the approach of "relies on environment to define it instead" is a better solution or not. As we still need to apply for different mode for some modules, the root problem seems not to be going away. Removing them seems to make users harder to control the behavior, as "based on env" can be implicit sometimes. I am also a bit worried about how would this affect existing projects, we might need to test more carefully before moving on (we probably really need a ecosystem CI then).

**Comment by sheremet-va** [comment]
> I don't have a strong opinion on this, but I am not very sure the approach of "relies on environment to define it instead" is a better solution or not.

It already works like this with enabled optimizer.

> As we still need to apply for different mode for some modules, the root problem seems not to be going away. 

I don't think this should ever be done. If your test file uses SSR features, every imported file should also rely on this. Otherwise, we can even create a separate module cache where the code is different, but the module is the same.

So, transform mode should be applied per _test file_, not per any file.

**Comment by antfu** [comment]
> per test file, not per any file.

Ok that makes sense. As long as it works and won't create large breakage I'd say go ahead :)

**Comment by sheremet-va** [comment]
Blocked by @vitejs/plugin-react (doesn't support `hmr: false` yet - fixed, but not released).

**Comment by sheremet-va** [comment]
@AriPerkkio so looks like we need to make Tinypool create new workers for new environments in 0.33.0. This PR expects the environment to be defined only once :(

**Comment by AriPerkkio** [comment]
Let's track Tinypool's feature at https://github.com/tinylibs/tinypool/issues/62.

I'm not yet sure about the ideal API but there are some ideas to consider.

File: docs/config/index.md [hunk 1]
```diff
@@ -1120,41 +1120,28 @@ Will call [`vi.unstubAllEnvs`](/api/vi#vi-unstuballenvs) before each test.
 
 Will call [`vi.unstubAllGlobals`](/api/vi#vi-unstuballglobals) before each test.
 
-### transformMode
+### testTransformMode
 
-- **Type:** `{ web?, ssr? }`
+ - **Type:** `{ web?, ssr? }`
+ - **Version:** Since Vitest 0.32.0
```
sheremet-va, line 1126: 0.33.0

Issue #3493, opened 2023-06-02
Allow to use function and class names for describe and it names


alfaproject:
### Clear and concise description of the problem

I'm trying to migrate our codebase from Jest to Vitest but now I have to update almost all of our test suites because we use function or class names for the describe names like so:

```ts
describe(VitestService, () => {
  describe(VitestService.prototype.isBroken, () => {
    // ...
  });
});
```

But we get type errors with Vitest ):

### Suggested solution

Allow for function and class names to be used in describe/it/test/etc names

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

JoshuaKGoldberg:
Looks like this was fixed by #3497 (hooray! 🥳) - can the issue be closed?

PR #3497, opened 2023-06-02
feat: support using function/class as `describe`/`test` name

fenghan34:
#3493


test/describe-test-name/test/runner.test.ts
```diff
@@ -0,0 +1,14 @@
+import { expect, test } from 'vitest'
+import { resolve } from 'pathe'
+import { runVitest } from '../../test-utils'
+
+test('should print function name', async () => {
+  const filename = resolve('./fixtures/test/example.test.ts')
+  const { stderr } = await runVitest({ root: './fixtures' }, [filename])
```
sheremet-va, line 7: I am not sure why do tests need to fail? I think we can just add these `describe`/`test` calls to test/reporters 🤔

sheremet-va (changes requested):

test/describe-test-name/test/runner.test.ts
fenghan34: We could also let them pass. Thanks for the suggestion, finding the perfect spot is challenging. 😅

sheremet-va (approved):

Issue #3109, opened 2023-03-30
Allow filtering UI down to individual blocks/tests


simon-abbott:
### Clear and concise description of the problem

Currently clicking on a test suite focuses the output on that suite, but clicking on a `describe` section or an individual test doesn't do anything.

### Suggested solution

Clicking on a `describe` section or an individual test should focus the right panel to only the output from that scope.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

g4rry420:
Hi, can I work on this issue ?

sheremet-va:
> Hi, can I work on this issue ?

You don't need to ask, you can work on it

PR #3505, opened 2023-06-04
feat: Filtering UI down to Individual test blocks

g4rry420:
Resolves: #3109 

I have considered two ways to add this functionality:
- In first method, we will get index and nestedIndex of items when user clicks on test name and if the that indexes are present then we will show that code block. Potential Issue: If the test file is not formatted properly, i.e, starting and ending of test block on the same line, then it might not work
- Second Method (fallback) - Here, based on the test name, we will search for that test and filter the string block for that test. Potential issue with this fallback method is that if there are two tests with the same name, then it will show the test block whose name comes first in the string.

 There is also Codemirror add-ons for code folding which can be useful for this feature, but I haven't taken a deep look into it. https://codemirror.net/docs/ref/#language.foldCode

Video:
https://www.loom.com/share/9fdbf88ef5114b2fa76e45b119b31e61

Disclaimer: I am coming from React background, so I am learning Vue on the go, please let me know if I am missing best practices specific to Vue

Also, when I was trying to make commit, I am facing this eslint error, not sure why that occurred, so I have commited the files with `-n` flag.
Screenshot 2023-06-04 003953


userquin:
run `nr lint` in your local, there are a few lint errors, check `Files changed` tab.

packages/ui/client/components/views/ViewEditor.vue
```diff
@@ -120,13 +244,22 @@ async function onSave(content: string) {
...
-    v-model="code"
-    h-full
-    v-bind="{ lineNumbers: true }"
-    :mode="ext"
-    data-testid="code-mirror"
-    @save="onSave"
-  />
+  <div>
+    <IconButton
+      v-if="testIndex !== null || isDescribeBlock !== null || selectedTest !== null"
```
dammy001, line 249: Can we create a computed property that holds this to avoid duplicate.

g4rry420: Sure

dammy001:
@userquin take a look at this again.

PR #3506, opened 2023-06-04
fix: the cli option is passed to coverage.exclude

### LINKED ISSUE #3450: CLI --coverage.exclude filter throws TypeError
### Describe the bug

I'm using Vitest in a monorepository (with `vitest.workspace.ts`), and it looks like the only way to provide coverage options is through CLI.

The command I use includes this filter: `--coverage.exclude='**/node_modules/**'`, which causes Vitest to throw this error: 
```sh
TypeError: resolved.coverage.exclude.push is not a function
```

### Reproduction

In a monorepository that leverages [Vitest Workspaces](https://vitest.dev/guide/workspace.html), run this command:

```sh
npx vitest run --coverage.enabled --coverage.lines=50 --coverage.functions=50 --coverage.branches=50 --coverage.statements=50 --coverage.all --coverage.include='packages/**' --coverage.include='apps/**' --coverage.exclude='**/node_modules/**'
```

### System Info

```shell
System:
    OS: macOS 12.6.3
    CPU: (10) arm64 Apple M1 Max
    Memory: 2.26 GB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Yarn: 3.2.3 - ~/.nvm/versions/node/v14.20.0/bin/yarn
  Browsers:
    Brave Browser: 109.1.47.171
    Chrome: 113.0.5672.126
    Firefox: 113.0.2
    Safari: 16.3
  npmPackages:
    @vitest/coverage-c8: ^0.31.1 => 0.31.1 
    @vitest/ui: ^0.31.1 => 0.31.1 
    vitest: ^0.31.1 => 0.31.1
```

### Used Package Manager

yarn

### Validations

btea:
fix #3450

### DIFF
--- a/packages/vitest/src/node/config.ts
+++ b/packages/vitest/src/node/config.ts
@@ -212,7 +212,9 @@ export function resolveConfig(
         ?? resolve(resolved.root, file),
     ),
   )
-  resolved.coverage.exclude.push(...resolved.setupFiles.map(file => relative(resolved.root, file)))
+  const exclude = toArray(resolved.coverage.exclude || [])
+  exclude.push(...resolved.setupFiles.map(file => relative(resolved.root, file)))
+  resolved.coverage.exclude = exclude
 
   resolved.forceRerunTriggers = [
     ...resolved.forceRerunTriggers,
--- a/packages/vitest/src/node/config.ts
+++ b/packages/vitest/src/node/config.ts
@@ -3,7 +3,7 @@ import { normalize, relative, resolve } from 'pathe'
 import c from 'picocolors'
 import type { ResolvedConfig as ResolvedViteConfig } from 'vite'
 
-import type { ApiConfig, ResolvedConfig, UserConfig, VitestRunMode } from '../types'
+import type { ApiConfig, BaseCoverageOptions, CoverageC8Options, CoverageIstanbulOptions, ResolvedConfig, UserConfig, VitestRunMode } from '../types'
 import { defaultBrowserPort, defaultPort } from '../constants'
 import { benchmarkConfigDefaults, configDefaults } from '../defaults'
 import { isCI, toArray } from '../utils'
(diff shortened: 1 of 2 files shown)


packages/vitest/src/node/config.ts
```diff
@@ -212,7 +212,9 @@ export function resolveConfig(
         ?? resolve(resolved.root, file),
     ),
   )
-  resolved.coverage.exclude.push(...resolved.setupFiles.map(file => relative(resolved.root, file)))
+  const exclude = toArray(resolved.coverage.exclude || [])
```
sheremet-va, line 215: As I said in https://github.com/vitest-dev/vitest/issues/3450#issuecomment-1567149596, we should guard all CLI option that can be passed down as array with such a guard

sheremet-va (changes requested):

packages/vitest/src/node/config.ts
btea: Refer to the [UserConfig](https://github.com/vitest-dev/vitest/blob/2deb70abcac02d96b1d38c1300e5f70a32e9dc1c/packages/vitest/src/types/config.ts#L625) type for option guard, I don’t know if there is any omission.

sheremet-va: Only options in CLI are respected:

https://github.com/vitest-dev/vitest/blob/952b5be6ee1cda5016b33d233a6abf7620745e2c/packages/vitest/src/node/cli.ts#L12

sheremet-va (changes requested):

packages/vitest/src/node/config.ts
```diff
@@ -55,6 +55,76 @@ export function resolveApiServerConfig<Options extends ApiConfig & UserConfig>(
   return api
 }
 
+function guardOptions(options: UserConfig) {
+  if (options.benchmark) {
```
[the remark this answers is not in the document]
btea: Maybe I just need to move the coverage guard here?
https://github.com/vitest-dev/vitest/blob/15253890e6b1876b3de330d64c9b3f51c4642fc9/packages/vitest/src/node/cli.ts#L134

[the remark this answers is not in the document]
sheremet-va: Sure 👍🏻

Issue #3435, opened 2023-05-23
[Feature]: Expect.soft and soft assertions


unickq:
### Soft assertions. 
The idea of soft assertions - do not terminate test execution, but mark the test as failed.

Description can be taken from https://github.com/jestjs/jest/issues/13205 and https://github.com/microsoft/playwright/pull/11800.
Playwright currently supports it, which makes its test runner amazing. I think it would be great to have this in vitest

```js
// Make a few checks that will not stop the test when failed...
await expect.soft(page.getByTestId('status')).toHaveText('Success');
await expect.soft(page.getByTestId('eta')).toHaveText('1 day');

// ... and continue the test to check more things.
await page.getByRole('link', { name: 'next page' }).click();
await expect.soft(page.getByRole('heading', { name: 'Make another order' })).toBeVisible();
```

### Suggested solution

https://playwright.dev/docs/test-assertions#soft-assertions
https://github.com/microsoft/playwright/pull/11800

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

sheremet-va:
There was an idea to implement it as an optional field on a test in #2139, but I agree that `expect.soft` is a better API.

PR #3507, opened 2023-06-04
feat(expect): support `expect.soft`

Dunqing:
close: #3435


packages/expect/src/jest-expect.ts
```diff
@@ -8,16 +8,16 @@ import { arrayBufferEquality, generateToBeMessage, iterableEquality, equals as j
...
 
 // Jest Expect Compact
 export const JestChaiExpect: ChaiPlugin = (chai, utils) => {
   const c = () => getColors()
 
   function def(name: keyof Assertion | (keyof Assertion)[], fn: ((this: Chai.AssertionStatic & Assertion, ...args: any[]) => any)) {
     const addMethod = (n: keyof Assertion) => {
-      utils.addMethod(chai.Assertion.prototype, n, fn)
-      utils.addMethod((globalThis as any)[JEST_MATCHERS_OBJECT].matchers, n, fn)
+      utils.addMethod(chai.Assertion.prototype, n, wrapSoft(utils, fn))
```
sheremet-va, line 19: `chai.Assertion` and `globalThis[JEST_MATCHERS_OBJECT]` should reference the same function, so fn needs to be wrapped only once

packages/expect/src/utils.ts
```diff
@@ -16,3 +20,25 @@ export function recordAsyncExpect(test: any, promise: Promise<any> | PromiseLike
...
+      : getState((globalThis as any)[GLOBAL_EXPECT])
+
+    if (!state.soft)
+      return fn.apply(this, args)
+
+    try {
+      return fn.apply(this, args)
+    }
+    catch (err) {
+      test.result.state = 'softFail'
```
sheremet-va, line 39: I'm not sure if we should introduce new states for this. I think `fail` is enough - otherwise it might break reporters

packages/runner/src/run.ts
```diff
@@ -156,21 +156,25 @@ export async function runTest(test: Test, runner: VitestRunner) {
...
         }
-
-        // some async expect will be added to this array, in case user forget to await theme
-        if (test.promises) {
-          const result = await Promise.allSettled(test.promises)
-          const errors = result.map(r => r.status === 'rejected' ? r.reason : undefined).filter(Boolean)
-          if (errors.length)
-            throw errors
+        if (test.result.state === 'softFail') {
+          failTask(test.result, [])
```
sheremet-va, line 160: Why do we need that?

sheremet-va (changes requested):

packages/runner/src/run.ts
```diff
@@ -218,17 +228,22 @@ export async function runTest(test: Test, runner: VitestRunner) {
...
-    ? err
-    : [err]
-  for (const e of errors) {
-    const error = processError(e)
-    result.error ??= error
-    result.errors ??= []
-    result.errors.push(error)
-  }
+  const errors = Array.isArray(err) ? err : [err]
+  // errors maybe push in expect.soft, so we need to process them
```
sheremet-va, line 235: Can't you just process them in `expect.soft`? I would suggest moving `serializeError` to `vitest/utils` and import it from there

test/core/test/fixtures/expects/soft.test.ts
```diff
@@ -0,0 +1,77 @@
...
+
+test('with expect', () => {
+  expect.soft(1).toEqual(2)
+  expect(10).toEqual(20)
+  expect.soft(2).toEqual(3)
+})
+
+test('with expect.extend', () => {
+  expect.soft(1).toEqual(2);
+  (expect.soft(3) as any).toBeDividedBy(4)
```
sheremet-va, line 53: Why do you need to write `as any`?

sheremet-va (changes requested):
Thank you for the work, looks very nice! Have some comments for now.

Also, can you add a few tests that are actually passing? Including type tests

packages/snapshot/package.json
```diff
@@ -43,13 +43,13 @@
     "prepublishOnly": "pnpm build"
   },
   "dependencies": {
+    "@vitest/utils": "workspace:*",
```
sheremet-va, line 46: in @vitest/snapshot utils should be in dev dependencies, it is bundled with the package

sheremet-va (changes requested):

packages/snapshot/package.json
sheremet-va: No, it should be bundled

sheremet-va: `error` should be a different entry point `@vitest/utils/error` because it depends on diff which is a separate entry point

sheremet-va: in this PR there should be zero changes to packages/snapshot

move runner error to `@vitest/utils/error` and move `createSimpleStackTrace` into `base.ts`

sheremet-va (approved):

test/core/test/expect.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    expect(stderr).toContain('AssertionError: expected 4 to be 5')
+    expect(stderr).toContain('4/4')
+  })
+
+  test('using expect.soft for test', async () => {
+    const { stderr } = await run({
+      testNamePattern: 'retry will failed',
+    })
+    expect.soft(stderr).toContain('AssertionError: expected 1 to be 4')
+    expect.soft(stderr).toContain('AssertionError: expected 2 to be 5')
```
sheremet-va, line 67: I am not sure if this is expected behavior. If you use regular expect it only shows errors that happened in the last run. Should this follow the same rule?

sheremet-va (changes requested):

test/core/test/expect.test.ts
sheremet-va: I don't understand this test then. It says that vitest will print "expected 1 to be 4", but it's not possible because the error happens before the last retry

test/core/test/expect.test.ts
```diff
@@ -0,0 +1,72 @@
+import { resolve } from 'node:path'
+import type { UserConfig } from 'vitest'
+import { describe, expect, expectTypeOf, test } from 'vitest'
+import { getCurrentTest } from '@vitest/runner'
+import { runVitest } from '../../test-utils'
+
+describe('expect.soft', () => {
+  const run = (config?: UserConfig) => runVitest({ root: resolve(__dirname, './fixtures/expects'), exclude: [], setupFiles: [], testNamePattern: getCurrentTest()?.name, testTimeout: 4000, ...config }, ['soft'])
```
sheremet-va, line 8: We don't run Vitest separetly in core tests. Core tests are regular tests that are running in Vitest.

Failing tests should be defined in `test/failing`

sheremet-va (changes requested):

sheremet-va:
So, from a testing perspective, all tests that check the CLI output should be checked in `test/failing`.

Tests that do not check the output and just run something should be defined in `tets/core` (so, actual tests that people run).

test/core/test/expect.test.ts
sheremet-va: > Rety do not clear the errors, so all errors are included

Ok, I checked and you are right. 👍🏻

sheremet-va: > Just to use expect.soft I can remove it if we don't need it

This test doesn't actually test anything new, so I don't see a reason to have it like that

sheremet-va (changes requested):

packages/expect/src/utils.ts
```diff
@@ -16,3 +21,25 @@ export function recordAsyncExpect(test: any, promise: Promise<any> | PromiseLike
 
   return promise
 }
+
+export function wrapSoft(utils: Chai.ChaiUtils, fn: (this: Chai.AssertionStatic & Assertion, ...args: any[]) => void) {
+  return function (this: Chai.AssertionStatic & Assertion, ...args: any[]) {
+    const test = utils.flag(this, 'vitest-test')
```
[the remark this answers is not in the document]
Dunqing: The type of test is defined in the `packages/runner`, the `packages/expect` doesn't have a `runner` dependency. Do we need to move types to `packages/util`? Or add `runner` to `devDependency` ?

packages/expect/src/utils.ts
```diff
@@ -16,3 +21,25 @@ export function recordAsyncExpect(test: any, promise: Promise<any> | PromiseLike
...
+      : getState((globalThis as any)[GLOBAL_EXPECT])
+
+    if (!state.soft)
+      return fn.apply(this, args)
+
+    try {
+      return fn.apply(this, args)
+    }
+    catch (err) {
+      test.result.state = 'fail'
```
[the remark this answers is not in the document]
sheremet-va: I am getting this error because test is undefined:

Screenshot 2023-06-06 at 12 17 25

packages/expect/src/utils.ts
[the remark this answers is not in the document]
sheremet-va: I think we can bring runner into dev dependencies. It should not be included in the bundle because we don't reexport this publically

packages/expect/src/utils.ts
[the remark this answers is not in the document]
Dunqing: Oh, I got it, need to do something to avoid this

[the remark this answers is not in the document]
sheremet-va: Yeah, some kind of a guard with a nice error that `expect.soft` should be called only inside a test

packages/expect/package.json
```diff
@@ -39,6 +39,7 @@
     "chai": "^4.3.7"
   },
   "devDependencies": {
+    "@vitest/expect": "workspace:*",
```
sheremet-va, line 42: Why would `expect` have itself in dependencies?

packages/expect/src/utils.ts
```diff
@@ -16,3 +22,26 @@ export function recordAsyncExpect(test: any, promise: Promise<any> | PromiseLike
...
+export function wrapSoft(utils: Chai.ChaiUtils, fn: (this: Chai.AssertionStatic & Assertion, ...args: any[]) => void) {
+  return function (this: Chai.AssertionStatic & Assertion, ...args: any[]) {
+    const test: Test = utils.flag(this, 'vitest-test')
+
+    // @ts-expect-error local is untyped
+    const state: MatcherState = test?.context._local
+      ? test.context.expect.getState()
+      : getState((globalThis as any)[GLOBAL_EXPECT])
+
+    if (!test || !state.soft)
```
sheremet-va, line 35: If soft is true and there is no test, we should throw an error

packages/expect/src/utils.ts
```diff
@@ -16,3 +22,26 @@ export function recordAsyncExpect(test: any, promise: Promise<any> | PromiseLike
...
+      return fn.apply(this, args)
+
+    try {
+      return fn.apply(this, args)
+    }
+    catch (err) {
+      test.result!.state = 'fail'
```
sheremet-va, line 42: We should probably add a guard here `test.result ??= {}` instead of suppressing errors

sheremet-va (changes requested):

packages/expect/package.json
Dunqing: typo 😅

packages/expect/src/utils.ts
```diff
@@ -27,12 +27,15 @@ export function wrapSoft(utils: Chai.ChaiUtils, fn: (this: Chai.AssertionStatic
   return function (this: Chai.AssertionStatic & Assertion, ...args: any[]) {
     const test: Test = utils.flag(this, 'vitest-test')
 
+    if (!test)
+      throw new Error('expect.soft() can only be used inside a test')
```
sheremet-va, line 31: It should only throw if `soft` is true

sheremet-va (changes requested):

packages/expect/src/utils.ts
sheremet-va: I think it's nice to have for custom implementations

sheremet-va:
> Thank you @sheremet-va so much for your great effort on `vitest`, and I am happy to be involved !!!

Thank _you_ for your wonderful work 👍🏻

sheremet-va (changes requested):
Last thing - we are missing a test case for `expect.soft` failing. I think adding a file with just `expect.soft(1).toBe(2)` to `test/failing/fixtures` should be enough - there is a test that dynamically loads every file and creates a snapshot automatically

Dunqing:
> Last thing - we are missing a test case for `expect.soft` failing. I think adding a file with just `expect.soft(1).toBe(2)` to `test/failing/fixtures` should be enough - there is a test that dynamically loads every file and creates a snapshot automatically

should be in `test/fails/fixture` ？

sheremet-va:
> should be in `test/fails/fixture` ？

Yes, but I meant the error we added recently - when `expect.soft` is called outside of the test. Sorry for the miscommunication.

Dunqing:
> Yes, but I meant the error we added recently - when `expect.soft` is called outside of the test. Sorry for the miscommunication.

Ok, I added it.

sheremet-va (approved):

PR #3509, opened 2023-06-04
feat: support soft expect

fenghan34:
#3435 

Support using soft expect both in `test` and hooks.


sheremet-va:
Hello! It's interesting that today two different people opened PR for the same feature 😀 But I feel like your runner implementation is too coupled with expect assertion. #3507 direction looks mote like what we need

fenghan34:
> Hello! It's interesting that today two different people opened PR for the same feature 😀 But I feel like your runner implementation is too coupled with expect assertion. #3507 direction looks mote like what we need

Aha, just realized that, will close this PR 😂

PR #3532, opened 2023-06-07
fix(ui): Module Graph view is blocking the Console view

g4rry420:
Once module graph view is opened, after we transitioned to Console View, it's content gets shifted to bottom of screen
Bug:
https://www.loom.com/share/8c3174cdab20473fa6f41fe817080c14


sheremet-va (commented):
I'm not sure, but I think there are performance implications? @userquin do you remember anything about this?

g4rry420:
Previous logic was indeed added because of performance implications - https://github.com/vitest-dev/vitest/pull/1719
I have made changes, please take a look into it

sheremet-va (approved):

PR #3536, opened 2023-06-07
perf: do not resolve static import since it's already resolved

sheremet-va:
All imports in `__vite_ssr_import__` should already be resolved, so we don't need to resolve it again.

It's possible that import in `__vite_ssr_dynamic_import__` is not resolved, so we still need to resolve it. @patak-dev do you think it's possible to provide information in `__vite_ssr_dynamic_import__` if import is static or dynamic? Something like this:

```ts
__vite_ssr_dynamic_import__('/path/to/file.ts', { resolved: true })
__vite_ssr_dynamic_import__(someVariable, { resolved: false })
```

### DIFF
--- a/packages/vite-node/src/client.ts
+++ b/packages/vite-node/src/client.ts
@@ -6,7 +6,7 @@ import { fileURLToPath, pathToFileURL } from 'node:url'
 import vm from 'node:vm'
 import { resolve } from 'pathe'
 import createDebug from 'debug'
-import { VALID_ID_PREFIX, cleanUrl, isInternalRequest, isNodeBuiltin, isPrimitive, normalizeModuleId, normalizeRequestId, slash, toFilePath } from './utils'
+import { cleanUrl, isInternalRequest, isNodeBuiltin, isPrimitive, normalizeModuleId, normalizeRequestId, slash, toFilePath } from './utils'
 import type { HotContext, ModuleCache, ViteNodeRunnerOptions } from './types'
 import { extractSourceMap } from './source-map'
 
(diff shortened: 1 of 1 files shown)


patak-dev:
In Vite, it seems we only need to resolve them when the path is relative. See https://github.com/vitejs/vite/blob/d814d6c56765696f766b32b5fce604acf715bf3c/packages/vite/src/node/ssr/ssrModuleLoader.ts#L181
Maybe doing the same here would work?

sheremet-va:
> In Vite, it seems we only need to resolve them when the path is relative. See [vitejs/vite@`d814d6c`/packages/vite/src/node/ssr/ssrModuleLoader.ts#L181](https://github.com/vitejs/vite/blob/d814d6c56765696f766b32b5fce604acf715bf3c/packages/vite/src/node/ssr/ssrModuleLoader.ts#L181)
> Maybe doing the same here would work?

It would mean that people cannot do this:

```ts
const filepath = condition ? '@/src/a.ts' : '@/src/b.ts'
await import(filepath)
```

patak-dev:
Is `@` an alias there? Does this work in Vite? I think it may be ok to say that this kind of completely dynamic file path needs to be resolved already by the user 🤔

sheremet-va:
> Is `@` an alias there? Does this work in Vite? I think it may be ok to say that this kind of completely dynamic file path needs to be resolved already by the user 🤔

It doesn't work in Vite, but it does work in vite-node which is a feature.

patak-dev:
I see. Two ideas. Is the id already added to the module graph at this point? Maybe a way to see if it has been resolved is to check if `moduleGraph.getModuleById(id)` returns a module? 
@antfu maybe you'll know better here if we should move this as a feature request to Vite core.

sheremet-va:
> I see. Is the id already added to the module graph at this point? Maybe a way to see if it has been resolved is to check if `moduleGraph.getModuleById(id)` returns a module?
> @antfu maybe you'll know better here if we should move this as a feature request to Vite core.

module graph stores modules only by their filepath in vite-node. And it's possible that it's not there yet.

sheremet-va:
Turns out we still need to resolve url for external modules (Vite does this in `nodeImport` function)

PR #3537, opened 2023-06-07
fix(test): vitest module directories and projects in vitest repo path

userquin:
This PR only fixes Vitest tests configuration files when Vitest is cloned in local under paths containing `packages` and/or `projects`.

### DIFF
--- a/examples/mocks/vite.config.ts
+++ b/examples/mocks/vite.config.ts
@@ -32,7 +32,7 @@ export default defineConfig({
     deps: {
       external: [/src\/external/],
       interopDefault: true,
-      moduleDirectories: ['node_modules', 'projects'],
+      moduleDirectories: ['node_modules', 'mocks/projects'],
     },
   },
 })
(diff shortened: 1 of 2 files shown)


sheremet-va:
This is not how `moduleDirectories` are intended to work.

If you just want to clone Vitest for yourself, don't clone it into packages folder.

Or better, fix the underlying problem.

userquin:
@sheremet-va so I cannot run test on my local... this was a temporary workaround, can you just try to run `pnpm vitest` inside your local `test/core` folder?

sheremet-va:
> @sheremet-va so I cannot run test on my local... this was a temporary workaround, can you just try to run `pnpm vitest` inside your local `test/core` folder?

Maybe instead of removing root we can remove the directory where the root config file is located? And `process.cwd` if there is no root config file?

```ts
// inside Vitest (node/core.ts)
dirname(config.configFile || process.cwd())
// inside Workspace (node/workspace.ts)
dirname(ctx.config.configFile || process.cwd())
```

sheremet-va:
What we can also do is treat `node_modules` differently (check `includes`), but ask people to provide absolute path? Or resolve path to be absolute. That way we can use `.includes` for `node_modules` and `.startsWith` for absolute paths 🤔

PR #3543, opened 2023-06-07
feat(reporters): show full test suite when testing 1 spec file at a time

Dunqing:
close: #3266


test/reporters/tests/default.test.ts
```diff
@@ -0,0 +1,24 @@
+import { resolve } from 'pathe'
+import { describe, expect, test } from 'vitest'
+import { runVitest } from '../../test-utils'
+
+const root = resolve(__dirname, '../fixtures')
+const run = async (fileFilter: string[]) => await runVitest({ root, reporters: 'default', config: false }, fileFilter)
+
+describe.skipIf(process.env.CI)('default reporter', async () => {
```
sheremet-va, line 8: Do I even need to explain why you shouldn't skip tests in CI?

Dunqing: This function only works with TTY, but CI will cause isTTY to be false

Dunqing: Aha, yes, I forgot that I could use this very useful method.😅

sheremet-va (approved):

packages/vitest/src/node/reporters/default.ts
```diff
@@ -8,6 +8,11 @@ export class DefaultReporter extends BaseReporter {
   renderer?: ReturnType<typeof createListRenderer>
   rendererOptions: ListRendererOptions = {} as any
 
+  onPathsCollected(paths: string[] = []) {
+    if (paths.length === 1)
+      this.rendererOptions.renderSucceed = true
+  }
```
sheremet-va, line 14: We should probably undo the change if watch mode triggers rerun again but with >1 tests

sheremet-va (changes requested):

packages/vitest/src/node/reporters/default.ts
Dunqing: Done, Thank you for your review!

sheremet-va:
Looks like tests are failing now

Dunqing:
The tests looks good now.

test/reporters/tests/default.test.ts
```diff
@@ -0,0 +1,58 @@
...
+    await vitest.waitForStdout('Input filename pattern')
+    vitest.write('b\n')
+    await vitest.waitForStdout('Waiting for file changes')
+    expect(vitest.stdout).toContain('RERUN')
+    expect(vitest.stdout).toContain('b1.test.ts')
+    expect(vitest.stdout).toContain('b2.test.ts')
+    expect(vitest.stdout).not.toContain('nested b1 test')
+    expect(vitest.stdout).not.toContain('b1 test')
+    expect(vitest.stdout).not.toContain('b2 test')
+  }, {
```
sheremet-va, line 55: This only proves the test is unstable. Maybe it's better to add some kind of waiting if it's the problem? It usually fails on line 53 - why?

sheremet-va (changes requested):

test/reporters/tests/default.test.ts
Dunqing: Yes, it does seem to be unstable in CI, but I'm not sure what's causing it yet and I need to spend some time checking it out.

sheremet-va (approved):

Dunqing:
> Looks like the test is still very unstable. Please, have a look when you have more time @Dunqing

Ok, I will do it later.

sheremet-va:
> > Looks like the test is still very unstable. Please, have a look when you have more time @Dunqing
> 
> Ok, I will do it later.

Another issue I see now is that `console.log` is hidden at the top (before the screen) if there are a lot of tests inside a file which leads to degraded user experience in my opinion.

Dunqing:
> Another issue I see now is that `console.log` is hidden at the top (before the screen) if there are a lot of tests inside a file which leads to degraded user experience in my opinion.

This behaviour should be the same as `jest`, If I cared about the contents of the `console.log`, I think I would use `test.only`. 

Or do we need to make sure that the number of tests in the file < xxxx before enabling this behaviour?

sheremet-va:
> This behaviour should be the same as `jest`,

Is it the same as jest? This test file runs quite differently:

```ts
it.each(new Array(100).fill(0).map((_, i) => i))('test', async (num) => {
  expect(true).toBe(true)
  await new Promise(resolve => setTimeout(resolve, num))
})
```

Vitest shows all 100 tests, but jest only shows the file name

Dunqing:
> > This behaviour should be the same as `jest`,
> 
> Is it the same as jest? This test file runs quite differently:
> 
> ```ts
> it.each(new Array(100).fill(0).map((_, i) => i))('test', async (num) => {
>   expect(true).toBe(true)
>   await new Promise(resolve => setTimeout(resolve, num))
> })
> ```
> 
> Vitest shows all 100 tests, but jest only shows the file name

image

I tested the code you provided with jest and at the end of the run, all the tests are printed.

Dunqing:
> > I tested the code you provided with jest and at the end of the run, all the tests are printed.
> 
> Yes, in the end, but not _during_

Oh, yes, It looks like this is the wrong implementation and needs to be changed to print only at the end of the test.

Dunqing:
@sheremet-va 

When I revert this pr and tested again using `vitest` with the code you provided, the results were the same, `vitest` will show all 100 tests. I guess this is because `it.each` is handled differently in `jest` and `vitest`?

sheremet-va:
> @sheremet-va
> 
> When I revert this pr and tested again using `vitest` with the code you provided, the results were the same, `vitest` will show all 100 tests. I guess this is because `it.each` is handled differently in `jest` and `vitest`?

I don't know how it's handled in jest. In vitest `test.eash` is the same as calling `test` function 100 times. I guess if behavior didn't change then it should be ok for the time being?..

sheremet-va:
Tets fails too often tho, we might have to revert this PR before release. Do you have plans to fix the issue with a flaky test?

sheremet-va:
> I'm working on this, I'm still not sure what's causing the different results, can't we really add retries for it?

Retry just halves the time it will fail, not fix the issue

ghiscoding:
@Dunqing just curious was this PR issue fixed or is it still falling because that caused some unit tests to fail right? I can't wait to see this in action though 😉

Dunqing:
> @Dunqing just curious was this PR issue fixed or is it still falling because that caused some unit tests to fail right? I can't wait to see this in action though 😉

It has been fixed, and we should see it in the next release.

ghiscoding:
@Dunqing  sweet thanks, however just to let you know that the issue #3266 was actually reopened by @sheremet-va, someone might need to recheck if it can be closed again :)

PR #3552, opened 2023-06-09
fix: automatically remove define related configuration

### LINKED ISSUE #3550: "vitest" command getting `TypeError: Cannot use 'in' operator to search for 'NO_COLOR' in undefined at chunk-colors
### Describe the bug

When I run the command vitest, I get the error;

TypeError: Cannot use 'in' operator to search for 'NO_COLOR' in undefined
 ❯ createColors node_modules/@vitest/utils/dist/chunk-colors.js:47:66
 ❯ setupGlobalEnv node_modules/vitest/dist/entry.js:53:15
 ❯ run node_modules/vitest/dist/entry.js:242:3
 ❯ run node_modules/vitest/dist/worker.js:60:5
 ❯ async file:/Users/aysegulkavakli/WebstormProjects/metamax-console/node_modules/tinypool/dist/esm/worker.js:109:20

and no test files found. Including vitest version 0.28.5 and before is working fine.

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-mafqkz?file=package.json,vite.config.ts&initialPath=__vitest__/

### System Info

```shell
System:
    OS: macOS Big Sur 11.5.2
  Binaries:
    Node: 16.14.0 - /usr/local/opt/node@16/bin/node
  Browsers:
    Chrome: 113.0.5672.126
  yarnPackages:
    testing-library/react: 14.0.0,
    vitejs/plugin-react: 4.0.0,
```

### Used Package Manager

yarn

### Validations

btea:
fix #3550

### DIFF
--- a/packages/vitest/src/node/plugins/utils.ts
+++ b/packages/vitest/src/node/plugins/utils.ts
@@ -1,6 +1,6 @@
 import { builtinModules } from 'node:module'
 import { version as viteVersion } from 'vite'
-import type { DepOptimizationOptions } from 'vite'
+import type { DepOptimizationOptions, UserConfig as ViteConfig } from 'vite'
 import type { DepsOptimizationOptions, InlineConfig } from '../../types'
 
 export function resolveOptimizerConfig(testOptionc: DepsOptimizationOptions | undefined, viteOptions: DepOptimizationOptions | undefined, testConfig: InlineConfig) {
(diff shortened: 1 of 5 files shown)


packages/vitest/src/node/plugins/utils.ts
```diff
@@ -32,3 +32,39 @@ export function resolveOptimizerConfig(testOptionc: DepsOptimizationOptions | un
   }
   return newConfig
 }
+
+export function deleteDefineConfig(viteConfig: ViteConfig) {
+  const defines: Record<string, any> = {}
+  if (viteConfig.define) {
+    delete viteConfig.define['import.meta.vitest']
+    delete viteConfig.define['process.env']
+    delete viteConfig.define.process
```
sheremet-va, line 41: We should probably also remove `global`.

And having a test would be nice 😄 I think adding `define` in `test/core` and checking the value in a test should be fine.

sheremet-va (changes requested):

packages/vitest/src/node/plugins/utils.ts
btea: ok, I have updated the code. 😄

sheremet-va (approved):

PR #3554, opened 2023-06-10
feat(runner): support `test.extend` 

### LINKED ISSUE #1660: `test` / `it`, and `describe` should follow the same pattern as `@playwright/test` for test context extensions

As a developer I'd love if the API for test context  (fixtures) mirrored that of [playwright test](https://playwright.dev/docs/test-fixtures)

This would be better for the following reasons:

- It makes context modifications easier to share
- Its not reliant on setting up `beforeEach` before a test to modify its context. You could easily share customized `test` / `describe` and/or `it` blocks. Using `beforeEach` is clunky and not easily sharable.
- Its (perhaps subjectively) more ergonomic and discoverable. Certainly works better for typing with TypeScript, which for large teams (especially enterprises) this is a huge win.

### Suggested solution

Allow for test context to be set on the test function level, via `test` /`it`, and/or `describe`. e.g.

```typescript
// my-test.ts
import { describe as base } from 'vitest;
import type { Todo } from './todo';
import * as faker from '@faker-js/faker';

// Declare the types of your fixtures.
interface TodoFixtures {
  todos: Todo[];
};

// Extend base test by providing "todo"
// This new "describe" can be used in multiple test files, and each of them will get the fixtures.
export const describe = base.extend<TodoFixtures>({
  todo: async (use) => {
    let i = 0;
    const todos = [];
    for (i; i < 25; i++) {
           todos.push({
              due: faker.date.between('2020-01-01T00:00:00.000Z', '2030-01-01T00:00:00.000Z'),
              assigned: faker.name.findName(),
              text: faker.lorem.lines()
          })
     }
     use(todos)
  },
});
```

### Alternative

_No response_

### Additional context

This is largely brought about because on my team of 24+ engineers we would like to be able to have standard patterns for faking data. One of the requirements when it comes to sharing API data is having fake datasets that can be re-used for testing and tested against schema so that can test that our validators work correctly. Its also the same fake data factory datasets used for our API test responses, and for data transformers, among other things. We really started to scale the `@playwright/test` approach well and our team is finding this API to be rather clunky by comparison, and having to declare a module augmentation adds more overhead and is prone to error.

### Validations

fenghan34:
Close #1660 

TODO:

- [x] Smartly initialize fixtures based on usage like Playwright does.


sheremet-va:
Very nice, thank you!

Can you add more documentation to fixtures guide and API?

fenghan34:
> Very nice, thank you!
> 
> Can you add more documentation to fixtures guide and API?

Sure, will do it.😀

sheremet-va:
Also would be nice to have documentation here 😄

fenghan34:
> Also would be nice to have documentation here for the next release 😄

Aha I'll try my best 😊

docs/api/index.md
```diff
@@ -51,6 +51,36 @@ In Jest, `TestFunction` can also be of type `(done: DoneCallback) => void`. If t
   })
   ```
 
+### test.extend
+
+- **Type:** `<T extends Record<string, any>>(fixtures: Fixtures<T>): TestAPI<ExtraContext & T>`
+- **Alias:** `it.extend`
+
+  Use `test.extend` to extend the test context with custom fixtures. This will return a new `test` and it's also extendable, so you can compose more fixtures or override existing ones by extending it as you need.
```
sheremet-va, line 59: Would be nice to have a link to the guide for more information here

sheremet-va (changes requested):

docs/api/index.md
fenghan34: Updated.

sheremet-va (approved):

outerlook:
Hey, thanks a lot for this one, that's one of the things I love the most on playwright! I'm sure will be loved on vitest too!

 One thing I'd like to ask, does the current implementation supports using a fixture from another fixture? I didn't actually see an example at the new docs, so just asking :sweat_smile: 

i.e. at playwright we may want resources to be dependent on one or many:

```ts
export const newTest = test.extend({
  server: async ({}, use) => {
    const { apiUrl, directApi } = await startServer();
    const server = { apiUrl, directApi };

    await use(server);

    await server.destroy();
  },
  // client depends on server
  client: async ({ server }, use) => {
    const client = await new Client({ apiUrl: server.apiUrl });
    await use(client);

    // Clean up the fixture.
    await client.destroy();
  },
  // user depends on client and server
  user: async ({ client, server }, use) => {
    const user = await client.createUser();
    await use(user);

    // Clean up the fixture.
    await server.directApi.destroyUser({ id: user.id });
  },
});

// although didn't ask for server, it will be instantiated because of the dependency of client
newTest("publish message", async ({ client, user }) => {
  const message = await client.publish(
    {
      id: user.id,
      partition: 0,
    },
    {
      messageNo: 1,
    }
  );
});
```
```

fenghan34:
> One thing I'd like to ask, does the current implementation supports using a fixture from another fixture?

No, currently it doesn't. Thank you for asking. It would be nice to have this feature, I'd love to add it.

PR #3556, opened 2023-06-11
feat(expect): support `expect.unreachable`

### LINKED ISSUE #3321: Add `assert.unreachable`

I'm trying to migrate from `uvu` to `vitest` and noticed that there's no `assert.unreachable`

### Suggested solution

https://github.com/lukeed/uvu/blob/56e1cb25ecd76cbf5cd3fd28c7a4c07db405c5c3/src/assert.js#L48

### Alternative

Just do `throw new Error();`. Feel free to close this issue if that's preferred

### Additional context

_No response_

### Validations

fenghan34:
Close #3321


sheremet-va:
LGTM 👍
Can you also add documentation for this feature?

fenghan34:
> LGTM 👍
> Can you also add documentation for this feature?

Will do it too.😄

packages/vitest/src/integrations/chai/index.ts
```diff
@@ -53,6 +53,10 @@ export function createExpect(test?: Test) {
     return assert
   }
 
+  expect.unreachable = (message?: string) => {
+    chai.assert.fail(`expected ${message ? `"${message}"` : ''} not to be reached`)
```
sheremet-va, line 57: I think if you don't pass a name, it will have double space.

sheremet-va (changes requested):

packages/vitest/src/integrations/chai/index.ts
fenghan34: 😅You are right!

sheremet-va:
Any news on the progress? Would like to have this in the next version.

fenghan34:
> Any news on the progress? Would like to have this in the next version.

I will update the documentation today. Quite busy in recent days.🥲

docs/api/expect.md
```diff
@@ -1136,6 +1136,32 @@ If the value in the error message is too truncated, you can increase [chaiConfig
...
+    throw new Error('Oops')
+  }
+
+  test('throwError throws an error with a message', () => {
+    try {
+      throwError()
+      expect.unreachable('I will not run')
+    }
+    catch (err: any) {
+      expect(err.message).toBe('Oops')
```
sheremet-va, line 1160: I feel like this example is too simplistic, because it's better to rewrite it into `expect(() => throwError()).toThrowError('Oops')`

Maybe we can add more checks to justify this pattern?

Like:

```ts
expect(err).toBeInstanceOf(Error)
expect(err.stack).toContain('throwError')
```

sheremet-va (changes requested):

docs/api/expect.md
fenghan34: I agree with you. I'll add them for now, but we better find another example or just leave it without examples.

fenghan34: 👀 Just found some `assert.unreachable` calls in SvelteKit codebase, I will create a new example.

docs/api/expect.md
```diff
@@ -1136,6 +1136,49 @@ If the value in the error message is too truncated, you can increase [chaiConfig
...
+- **Type:** `(message?: string) => never`
+
+  This method is used to asserting that a line should never be reached.
+
+  For example, if we want to test that `build()` throws due to receiving directories having no `src` folder, and also handle each error separately, we could do this:
+
+  ```ts
+  import { expect, test } from 'vitest'
+
+  async function build(dir: string) {
```
sheremet-va, line 1150: We try not to use ts syntax in examples
```suggestion
  async function build(dir) {
```

sheremet-va (changes requested):

docs/api/expect.md
fenghan34: OK, got it.

sheremet-va (approved):

PR #3557, opened 2023-06-12
docs: update extending-matchers.md

sagargurtu:
Interfaces `Assertion` and `AsymmetricMatchersContaining` must be imported from `'vitest'` for correct module augmentation. Otherwise, it will declare whole new interfaces.


docs/guide/extending-matchers.md
```diff
@@ -26,6 +26,8 @@ expect.extend({
 If you are using TypeScript, since Vitest 0.31.0 you can extend default `Assertion` interface in an ambient declaration file (e.g: `vitest.d.ts`) with the code below:
 
 ```ts
+import type { Assertion, AsymmetricMatchersContaining } from 'vitest';
```
sheremet-va, line 29:
```suggestion
import type { Assertion, AsymmetricMatchersContaining } from 'vitest'
```

PR #3562, opened 2023-06-13
feat(runner): `describe`/`test` name support anonymous function

btea:
Supplement #3497


packages/runner/src/suite.ts
```diff
@@ -275,7 +275,7 @@ function createTest(fn: (
 }
 
 function formatName(name: string | Function) {
-  return typeof name === 'string' ? name : name instanceof Function ? name.name : String(name)
+  return typeof name === 'string' ? name : name instanceof Function ? (name.name || 'Anonymous') : String(name)
```
sheremet-va, line 278: Shouldn't it be `<anonymous>`? This is what we do for serialized error:

https://github.com/vitest-dev/vitest/blob/158c4bb056417828833088508b43e6917d65d23f/packages/utils/src/error.ts#L29

sheremet-va (changes requested):

packages/runner/src/suite.ts
btea: Ok.

PR #3563, opened 2023-06-13
fix: `createVitest` need also register console shortcuts (#3561)

### LINKED ISSUE #3561: `createVitest` should also use `registerConsoleShortcuts`
### Describe the bug

When I use `createVitest` to apply the test case, I can't respond to the terminal input, but `startVitest` can.

- `createVitest`
```ts
import { createVitest } from 'vitest/node'

const vitest = await createVitest('test', {
  watch: true
})
vitest.start()
```

- `startVitest`
```ts
import { startVitest } from 'vitest/node'

await startVitest('test', [], {
  watch: true
})
```

I compared the two methods and found that only `startVitest` called `registerConsoleShortcuts`([link](https://github.com/vitest-dev/vitest/blob/main/packages/vitest/src/node/cli-api.ts#L91)). I think `createVitest` should also use  `registerConsoleShortcuts` when it executed first time.

### Reproduction

[link](https://stackblitz.com/edit/vitest-dev-vitest-y1kojr?file=app.ts&initialPath=__vitest__/)

- `createVitest`

CleanShot 2023-06-13 at 17 57 24

- `startVitest`

CleanShot 2023-06-13 at 18 00 45

### System Info

```shell
System:
    OS: macOS 12.6.3
    CPU: (8) x64 Intel(R) Core(TM) i7-6820HQ CPU @ 2.70GHz
    Memory: 673.48 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 19.8.1 - ~/.nvm/versions/node/v19.8.1/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v19.8.1/bin/yarn
    npm: 9.5.1 - ~/.nvm/versions/node/v19.8.1/bin/npm
  Browsers:
    Chrome: 113.0.5672.92
    Firefox: 98.0
    Safari: 16.3
```

### Used Package Manager

npm

### Validations

deot:
Closes #3561

### DIFF
--- a/packages/vitest/src/node/cli-api.ts
+++ b/packages/vitest/src/node/cli-api.ts
@@ -6,6 +6,7 @@ import { getEnvPackageName } from '../integrations/env'
 import type { UserConfig, Vitest, VitestRunMode } from '../types'
 import { ensurePackageInstalled } from './pkg'
 import { createVitest } from './create'
+import { registerConsoleShortcuts } from './stdin'
 
 export interface CliOptions extends UserConfig {
   /**
@@ -85,6 +86,11 @@ export async function startVitest(
     process.exitCode = 1
     return ctx
   }
+
+  if (process.stdin.isTTY && ctx.config.watch)
+    registerConsoleShortcuts(ctx)
+  else
+    process.on('SIGINT', () => ctx.cancelCurrentRun('keyboard-input'))
 
   ctx.onServerRestart((reason) => {
     ctx.report('onServerRestart', reason)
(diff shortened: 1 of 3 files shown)


antfu:
I think it's intentional. `createVitest` can be used in a more controlled way that might not be directly exposed to the terminal. The shortcuts should live in the CLI integration

sheremet-va:
> I think it's intentional. `createVitest` can be used in a more controlled way that might not be directly exposed to the terminal. The shortcuts should live in the CLI integration

Agree. `createVitest` shouldn't have this kind of side effects

deot:
In this case, should `registerConsoleShortcuts` be exposed?

sheremet-va (approved):

antfu (approved):

PR #3570, opened 2023-06-14
fix: Fix errors caused by commonjs export circular references

rxliuli:
Close https://github.com/vitest-dev/vitest/issues/3519


test/vite-node/test/self-export.test.ts
```diff
@@ -0,0 +1,24 @@
...
+import { resolve } from 'pathe'
+import ansiEscapes, { HelloWorld } from '../src/self-export'
+
+it('should export self', () => {
+  expect(ansiEscapes.HelloWorld).eq(HelloWorld)
+  expect(Reflect.get(ansiEscapes, 'default').HelloWorld).eq(HelloWorld)
+  expect(HelloWorld).eq(1)
+})
+
+const cliPath = resolve(__dirname, '../../../packages/vite-node/src/cli.ts')
```
sheremet-va, line 12: You should be able to reference `vite-node/dist/cli.js` to call already built package with `node` and not `esno`

sheremet-va (changes requested):

test/vite-node/test/self-export.test.ts
sheremet-va: Then it's probably fine. I will look into vite-node tests separately then.

sheremet-va (approved):

Issue #3571, opened 2023-06-14
When vite-node runs the script, the parameters should be passed transparently


rxliuli:
### Describe the bug

For example, the following is the behavior of `node/tsx/ts-node/vite-node` four clis running a script. I think the behavior of vite-node should be consistent with others

```sh
> vite-node-execute-demo@1.0.0 start /Users/rxliuli/code/web/demo/vite-node-execute-demo
> pnpm start-node && pnpm start-tsx && pnpm start-ts-node && pnpm start-vite-node


> vite-node-execute-demo@1.0.0 start-node /Users/rxliuli/code/web/demo/vite-node-execute-demo
> node hello.js -h

[
  '/usr/local/bin/node',
  '/Users/rxliuli/code/web/demo/vite-node-execute-demo/hello.js',
  '-h'
]

> vite-node-execute-demo@1.0.0 start-tsx /Users/rxliuli/code/web/demo/vite-node-execute-demo
> tsx hello.js -h

[
  '/usr/local/bin/node',
  '/Users/rxliuli/code/web/demo/vite-node-execute-demo/hello.js',
  '-h'
]

> vite-node-execute-demo@1.0.0 start-ts-node /Users/rxliuli/code/web/demo/vite-node-execute-demo
> ts-node hello.js -h

[
  '/Users/rxliuli/code/web/demo/vite-node-execute-demo/node_modules/ts-node/dist/bin.js',
  '/Users/rxliuli/code/web/demo/vite-node-execute-demo/hello.js',
  '-h'
]

> vite-node-execute-demo@1.0.0 start-vite-node /Users/rxliuli/code/web/demo/vite-node-execute-demo
> vite-node hello.js -h

vite-node/0.32.0

Usage:
  $ vite-node [...files]

Commands:
  [...files]  

For more info, run any command with the `--help` flag:
  $ vite-node --help

Options:
  -v, --version        Display version number 
  -r, --root <path>    Use specified root directory 
  -c, --config <path>  Use specified config file 
  -m, --mode <mode>    Set env mode 
  -w, --watch          Restart on file changes, similar to "nodemon" 
  --script             Use vite-node as a script runner 
  --options <options>  Use specified Vite server options 
  -h, --help           Display this message 
```

hello.js

```js
console.log(process.argv)
```

### Reproduction

1. git clone https://github.com/rxliuli/vite-node-execute-demo
2. cd vite-node-execute-demo
3. pnpm i
4. pnpm start

### System Info

```shell
[...]

PR #3574, opened 2023-06-14
feat: Make vite-node parsing behavior consistent with node/tsx/ts-node

rxliuli:
Closes https://github.com/vitest-dev/vitest/issues/3571


packages/vite-node/src/cli.ts
```diff
@@ -35,10 +45,13 @@ export interface CliOptions {
   mode?: string
   watch?: boolean
   options?: ViteNodeServerOptionsCLI
+  version?: boolean
+  help?: boolean
   '--'?: string[]
 }
 
 async function run(files: string[], options: CliOptions = {}) {
+  // console.log('run', files, options)
```
sheremet-va, line 54: Looks like you left a comment here

sheremet-va (changes requested):

packages/vite-node/src/cli.ts
rxliuli: Sorry, I forgot to remove the debug code, I'll fix it later

rxliuli:
@sheremet-va What prevents the merge?

sheremet-va:
> @sheremet-va What prevents the merge?

No functional review done yet

marviobezerra:
I’m also waiting for it. Any news on merging/releasing?

packages/vite-node/src/cli.ts
```diff
@@ -48,6 +60,15 @@ async function run(files: string[], options: CliOptions = {}) {
     process.argv = [...process.argv.slice(0, 2), ...(options['--'] || [])]
   }
 
+  if (options.version) {
+    cli.version(version)
+    cli.outputVersion()
+    process.exit(0)
+  }
+  if (options.help) {
+    cli.outputHelp()
```
sheremet-va, line 69: `--help` doesn't print version anymore

sheremet-va (changes requested):

sheremet-va:
Looks good to me. Only one small problem was found. We can merge it after it is fixed.

rxliuli:
An error occurred, but it doesn't seem to be because of my modification. . .

sheremet-va:
@rxliuli looks like this broke the usage. `vite-node t.js --watch` doesn't work anymore - `--watch` is inside `--`:

```ts
console.log(options)
// { '--': [ '--watch' ] }
```

sheremet-va:
Tests are testing `esno ./path/` instead of `vite-node --h` which would be preferable here.

rxliuli:
> @rxliuli looks like this broke the usage. `vite-node t.js --watch` doesn't work anymore - `--watch` is inside `--`:
> 
> ```ts
> console.log(options)
> // { '--': [ '--watch' ] }
> ```

Yes, this is consistent with the behavior of node/ts-node/tsx

image

> Tests are testing esno ./dist/cli.js instead of vite-node --h which would be preferable here.

I'm not sure how to run the vite-node command in unit tests, maybe I need to use zx to run `pnpm vite-node`?

sheremet-va:
> Yes, this is consistent with the behavior of node/ts-node/tsx

This breaks watch mode - we don't need consistency when our own features are broken. Doesn't `--script` satisfy your needs?

> I'm not sure how to run the vite-node command in unit tests, maybe I need to use zx to run `pnpm vite-node`?

I will push new utilities in a few minutes.

sheremet-va:
I pushed utilities here: https://github.com/vitest-dev/vitest/pull/3842

Please, open a new PR that specifies the changes in the README. "consistent with node" is not a change - please, reference all breaking changes that you introduced and how to migrate.

Issue #3154, opened 2023-04-09
Add an example of vue component testing via browser mode with playwright


jd-solanki:
### Clear and concise description of the problem

Hi 👋🏻 
I'm building Vue UI framework - [Anu](https://anu-vue.netlify.app/)

I really liked browser mode and I'm eagerly waiting for the stability of this feature. I must say we can't completely test our 
 vue components with jsdom when we build the UI framework and use vitest for testing. Browser mode is a huge encouragement for using vitest but in the early stage I did experience the #3124 and found that adding an example of testing vue component with Playwright's chrome as a browser provider that demonstrates browser API access like `window.getComputedStyle` will be easy for us to adopt and use, Moreover, it will also allow vitest team to quickly identify if there's any bug.

### Suggested solution

n/a

### Alternative

Not writing tests at all 😓 

### Additional context

I'm waiting for browser mode stability to make robust tests for testing vue components. I hope using browser mode will provide the same environment as the real browsers for robustness.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

Aslemammad:
I'm happy you find this feature useful, sure, I think an example would help! Feel free to send a PR if you want. I'm not that experienced in vue.

jd-solanki:
Hey, Thanks for the response. I'm still learning tests so I created this issue for creating example for beginners like me.

Aslemammad:
Ok then, PRs are welcome!

lmiller1990:
Not a direct solution, but I've been working on testing Vue [at Cypress](https://docs.cypress.io/guides/component-testing/vue/overview) - we've got a Vite dev server that works great, this might an alternative (it should all just work out of the box - some other projects I work on use it pretty successfully, like [Vuetify](https://github.com/vuetifyjs/vuetify/tree/master)).

Aslemammad:
@lmiller1990 Why not? If it works, then it'd be good to have those examples. Anyone can send a PR if they want, so feel free to do so. Thank you so much.

kingyue737:
[...]

PR #3584, opened 2023-06-15
feat(browser): isolate browser tests

userquin:
This PR isolates tests creating an iframe per test file.

On my local tests, I need to start the tests a few times to get it working, I have added vue and react examples and vue should also work (this is one we need to revert changes).

DRAFT: we'll need to revert some changes in existing tests and cleanup some modules.


sheremet-va:
> > creating an iframe per test file
> 
> There can be only 1 focused element per page, so I don't think iframes are good enough for isolation. Maybe 1 page per test? Not sure if it's possible to preserve/emulate focus like with the devtools' "Emulate a focused page" command.

This is a valid concern. We need to add a test with focusing elements.

userquin:
> > > creating an iframe per test file
> > 
> > 
> > There can be only 1 focused element per page, so I don't think iframes are good enough for isolation. Maybe 1 page per test? Not sure if it's possible to preserve/emulate focus like with the devtools' "Emulate a focused page" command.
> 
> This is a valid concern. We need to add a test with focusing elements.

I'll try to add a focus test...

userquin:
working

imagen

userquin:
To run browser test:
1) from root `ni && nr build` and then
2) open a new terminal, change to `examples/vue-browser`  and run `nr test`
3) open a new terminal, change to `examples/react-testing-lib-browser`  and run `nr test`

I'll try to add more frameworks examples: lit, svelte, preact and solid, and maybe also sveltekit

packages/browser/src/client/tester.ts
```diff
@@ -0,0 +1,51 @@
...
+    config.snapshotOptions.snapshotEnvironment = new BrowserSnapshotEnvironment()
+
+  try {
+    await setupCommonEnv(config)
+
+    await startTests([filename], runner)
+  }
+  finally {
+    // TODO: remove version once logs removed from main.ts module
+    channel.postMessage({ type: 'done', filename, version })
```
userquin, line 46: Remove version here

packages/vitest/src/node/pools/browser.ts
```diff
@@ -41,8 +44,9 @@ export function createBrowserPool(ctx: Vitest): ProcessPool {
     const origin = `http://${ctx.config.browser.api?.host || 'localhost'}:${project.browser.config.server.port}`
     const paths = files.map(file => relative(project.config.root, file))
 
-    const isolate = project.config.isolate
-    if (isolate) {
+    // TODO@browser: review this
+    // const isolate = project.config.isolate
+    /* if (isolate) {
```
userquin, line 49: Remove old code

packages/browser/src/client/main.ts
```diff
@@ -1,141 +1,204 @@
...
+      const idx = waitingPaths.indexOf(filename)
+      if (idx === -1)
+        return
+
+      waitingPaths.splice(idx, 1)
+
+      if (!waitingPaths.length) {
+        await rpcDone()
+        await rpc().onDone('no-isolate')
+        listenToRun = true
```
userquin, line 187: Use try/finally?

packages/vitest/src/types/browser.ts
```diff
@@ -53,6 +54,13 @@ export interface BrowserConfigOptions {
    */
   api?: ApiConfig | number
 
+  /**
+   * When running on browser, enable this flag to show the result.
+   *
+   * @default false
+   */
+  enableUI?: boolean
```
Aslemammad, line 62: @userquin  Do we actually need this? I don't have any strong preference over having or not having it as an option, just asking. Can we show it without an option?

userquin: You can run test in the ui and those test may not have ui, check test/browser

userquin: It is read in the ui when loading the configuration, it is used to show the tab in the details panel

userquin: @Aslemammad 

imagen

Aslemammad: I see, That's better, thanks.

sheremet-va:
@userquin can you resolve conflicts last time?

packages/vitest/src/types/browser.ts
sheremet-va: I think the use case when you run it in the browser without a UI is not so popular. `@vitest/browser` is primarily to allow component testing. examples/browser is just a test to make sure everything works seamlessly

sheremet-va (changes requested):

userquin:
> @userquin can you resolve conflicts last time?

Sure

jfgreffier:
Closes #3154

shairez:
Looks like an awesome feature!
Is there anything blocking it other than fixing the conflicts?

DanWebb:
Is the browser UI tab still a planned addition after #5036? 🤞🙏

PR #3603, opened 2023-06-17
feat: introduce global configuration for retry setting (fix #3598)

### LINKED ISSUE #3598: introduce global configuration for retry setting

imentu:
- Close #3598


imentu:
Hi @sheremet-va ,

I am a new contributor, I can't trigger the CI without a contributor's approval, could you please help me approve it and take a code review?

Thanks!

docs/config/index.md
```diff
@@ -1443,3 +1443,12 @@ This config option affects truncating values in `test.each` titles and inside th
 Stop test execution when given number of tests have failed.
 
 By default Vitest will run all of your test cases even if some of them fail. This may not be desired for CI builds where you are only interested in 100% successful builds and would like to stop test execution as early as possible when test failures occur. The `bail` option can be used to speed up CI runs by preventing it from running more tests when failures have occured.
+
+### retry
+
+- **Type:** `number`
+- **Default:** `1`
+- **CLI:** `--retry=<value>`
+- **Version:** Since Vitest 0.33.0
```
sheremet-va, line 1452: It will be released in 0.32.2

packages/vitest/src/node/cli.ts
```diff
@@ -48,6 +48,7 @@ cli
   .option('--inspect-brk', 'Enable Node.js inspector with break')
   .option('--test-timeout <time>', 'Default timeout of a test in milliseconds (default: 5000)')
   .option('--bail <number>', 'Stop test execution when given number of tests have failed', { default: 0 })
+  .option('--retry <number>', 'Retry the test specific number of times if it fails.', { default: 1 })
```
sheremet-va, line 51: This should also be mentioned in CLI documentation

sheremet-va (changes requested):

packages/runner/src/suite.ts
```diff
@@ -77,7 +77,7 @@ function createSuiteCollector(name: string, factory: SuiteFactory = () => { }, m
       mode,
       suite: undefined!,
       fails: this.fails,
-      retry: options?.retry,
+      retry: options?.retry ?? runner.config.retry,
```
Dunqing, line 80:
```suggestion
      retry: options?.retry || runner.config.retry,
```

I don't think it makes sense to be 0 here, so feel free to use `||`

Dunqing (changes requested):

packages/runner/src/suite.ts
sheremet-va: Why doesn't it makes sense to you? 0 is the default. If someone wants to override it, it should be possible to override

Dunqing: If the retry is 0, the test will not run. this may lead to some misunderstandings. In addition, the default should be 1.

sheremet-va: Why wouldn't it run? It should retry only if actual run failed. This is how it works in other frameworks. (`retry: 1` calls test _twice_ - one actual test, the second one is the retry)

test/core/vitest.config.ts
```diff
@@ -46,6 +46,7 @@ export default defineConfig({
     exclude: ['**/fixtures/**', ...defaultExclude],
     slowTestThreshold: 1000,
     testTimeout: 2000,
+    retry: 3,
```
Dunqing, line 49: It looks like this configuration will cause all core tests to be retried three times.

Dunqing (changes requested):

packages/runner/src/suite.ts
Dunqing: You can see here that the retry is ensured at least once, if it is 0, the for loop will not be executed and the test will be skipped.
https://github.com/vitest-dev/vitest/blob/0c498dab10cf6a40742fb1136bf6e2b3f51635f6/packages/runner/src/run.ts#L138-L140

Also, now that retry has a default value, we can delete line 38

sheremet-va: Then it is a bug in Vitest implementation, retry should be 0 if test didn't fail

Dunqing:
This #3602 PR test looks to be useful here, you can copy from that.

packages/runner/src/suite.ts
Dunqing: Yes, I aggre with you, this looks a bit confusing to me at the beginning

sheremet-va:
> This #3602 PR test looks to be useful here, you can copy from that.

What is the rationale to run two separate tests?

Dunqing:
> What is the rationale to run two separate tests?

Avoid retry configurations that interfere with unrelated tests.

packages/vitest/src/node/cli.ts
imentu: done, see change in `docs/guide/cli.md`

I also found some other CLI options which not appear in the documentation:
- single-thread
- hideSkippedTests
- segfaultRetry
- test-timeout

docs/config/index.md
imentu: done

imentu:
> > What is the rationale to run two separate tests?
> 
> Avoid retry configurations that interfere with unrelated tests.

I agree with @Dunqing, what's your opinion @sheremet-va ?

sheremet-va:
> > > What is the rationale to run two separate tests?
> > 
> > 
> > Avoid retry configurations that interfere with unrelated tests.
> 
> I agree with @Dunqing, what's your opinion @sheremet-va ?

I agree

packages/vitest/src/defaults.ts
```diff
@@ -94,6 +94,7 @@ const config = {
     exclude: defaultExclude,
   },
   slowTestThreshold: 300,
+  retry: 1,
```
Dunqing, line 97: This line should be removed

Dunqing (changes requested):

packages/runner/src/suite.ts
sheremet-va: We should fix it first then, before adding a global option

Dunqing: Can i do it？

Dunqing (approved):

imentu:
@sheremet-va PTAL

sheremet-va (approved):

PR #3615, opened 2023-06-19
feat!: disable `coverage.reportOnFailure` by default

AriPerkkio:
- Continues #3453
- It makes more sense to have reports not generated when test failures occur

### BREAKING CHANGES:

- `coverage.reportOnFailure` is now `false` by default

### DIFF
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -804,7 +804,7 @@ Since Vitest 0.31.0, you can check your coverage report in Vitest UI: check [Vit
 #### coverage.reportOnFailure
 
 - **Type:** `boolean`
-- **Default:** `true`
+- **Default:** `false` (since Vitest `0.33.0`)
 - **Available for providers:** `'c8' | 'v8' | 'istanbul'`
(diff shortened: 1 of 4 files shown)


docs/config/index.md
```diff
@@ -804,7 +804,7 @@ Since Vitest 0.31.0, you can check your coverage report in Vitest UI: check [Vit
 #### coverage.reportOnFailure
 
 - **Type:** `boolean`
-- **Default:** `true`
```
sheremet-va, line 807: I feel like we should mention that it was `true` before 0.33.0

sheremet-va (changes requested):

docs/config/index.md
```diff
@@ -804,7 +804,7 @@ Since Vitest 0.31.0, you can check your coverage report in Vitest UI: check [Vit
...
+- **Default:** `false` (since Vitest `0.33.0`)
```
AriPerkkio, line 807:
```suggestion
- **Default:** `false` (since Vitest `0.34.0`)
```

PR #3618, opened 2023-06-19
fix(watch): cancel using `h` key

### LINKED ISSUE #3612: Allow to interrupt test running in watch mode

Currently if I have thousands of tests to run and want to use the watch feature to filter only say one test, or test file, then I have to wait until the entire test suite runs to then press "h" and do that. This takes a lot of time and also is different to how Jest works where you can interrupt the test running and do filtering.

### Suggested solution

Pressing enter, or any of the keys like "h" or whatever should interrupt the current run. That way I could start the watch, press enter, then start filtering to the tests I care about. This is how Jest works and it would be great if Vite did similar.

### Alternative

_No response_

### Additional context

_No response_

### Validations

AriPerkkio:
- Watch mode should be cancellable using any key that Vitest tracks
- Track `h` key for cancelling on-going test run
- Fixes #3612 completely

Tested manually. We have test case for testing cancelling with `c` key but testing all of these would be slow and flaky.

### DIFF
--- a/packages/vitest/src/node/stdin.ts
+++ b/packages/vitest/src/node/stdin.ts
@@ -13,7 +13,7 @@ const keys = [
   ['t', 'filter by a test name regex pattern'],
   ['q', 'quit'],
 ]
-const cancelKeys = ['space', 'c', ...keys.map(key => key[0]).flat()]
+const cancelKeys = ['space', 'c', 'h', ...keys.map(key => key[0]).flat()]
 
 export function printShortcutsHelp() {
   stdout().write(


sheremet-va (approved):

sheremet-va (changes requested):
Wait, should we really reuse `h` for help _and_ cancel? What is the reasoning to use the same key for drastically different actions?

sheremet-va:
> This change means that when tests are running and `h` is pressed once, the test run cancels. Nothing more. When `h` is pressed again now that tests are not running, the regular handler kicks in and renders the help section.
> 
> Basically any key that Vitest watch mode uses can be used to cancel the on-going run.

Ah, I see now.

sheremet-va (approved):

Issue #3470, opened 2023-05-30
Module graph not working in workspace


TECH7Fox:
### Describe the bug

I have a workspace with multiple projects. When I run `vitest --ui --coverage` all the module graphs are empty. They do work if I run vitest in the project directly.

### Reproduction

https://github.com/TECH7Fox/vitest-monorepo

### System Info

```shell
System:
    OS: Linux 5.15 Ubuntu 20.04.6 LTS (Focal Fossa)
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i5-1145G7 @ 2.60GHz
    Memory: 2.60 GB / 15.36 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 18.14.0 - ~/.nvm/versions/node/v18.14.0/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 9.6.7 - ~/repositories/my_project/node_modules/.bin/npm
  Browsers:
    Chrome: 113.0.5672.92
    Firefox: 113.0.2
  npmPackages:
    @vitest/coverage-c8: ^0.31.1 => 0.31.1 
    vitest: ^0.31.1 => 0.31.1
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

TECH7Fox:
Here is a reproduction: https://github.com/TECH7Fox/vitest-monorepo

TECH7Fox:
Here is the module graph when I run vitest in the project itself.
![image](https://github.com/vitest-dev/vitest/assets/32220029/e3b25794-ba87-442e-82cb-d69d10682a14)

[...]

PR #3636, opened 2023-06-21
fix: Module graph in workspace of UI package

g4rry420:
fix: #3470

Screenshot 2023-06-30 225652


sheremet-va (changes requested):
I don't think this is a good solution. Module Graph should show module from the same workspace as the test file, not a random one

g4rry420:
@sheremet-va Please, correct me if I am wrong, but isn't this code will show module from the same workspace as the test file, as I am showing the module based on the id of the file.
Or is there a specific scenario that I am missing related to workspaces ? Please, let me know :)

```js
const workspaceModules = workspaceProject?.ctx.getModuleProjects(id)
let workspaceMod: ModuleNode | undefined
if (workspaceModules && workspaceModules.length === 1) {
    const { server, browser } = workspaceModules[0]
    const mod = server.moduleGraph.getModuleById(id) || browser?.moduleGraph.getModuleById(id)
    workspaceMod = mod
}
```

sheremet-va:
> @sheremet-va Please, correct me if I am wrong, but isn't this code will show module from the same workspace as the test file, as I am showing the module based on the id of the file. Or is there a specific scenario that I am missing related to workspaces ? Please, let me know :)
> 
> ```js
> const workspaceModules = workspaceProject?.ctx.getModuleProjects(id)
> let workspaceMod: ModuleNode | undefined
> if (workspaceModules && workspaceModules.length === 1) {
>     const { server, browser } = workspaceModules[0]
>     const mod = server.moduleGraph.getModuleById(id) || browser?.moduleGraph.getModuleById(id)
>     workspaceMod = mod
> }
> ```

It's possible to have the same test file in different workspace projects and process them differently. There is an example in test/workspaces with `space_shared`

g4rry420:
Ah okay, Thanks, Do you have suggestions about how this should be resolved then ?

packages/ui/client/components/TaskItem.vue
```diff
@@ -23,7 +23,7 @@ const duration = computed(() => {
   >
     <StatusIcon :task="task" mr-2 />
     <div flex items-end gap-2 :text="task?.result?.state === 'fail' ? 'red-500' : ''">
-      <span text-sm truncate font-light>{{ task.name }}</span>
+      <span text-sm truncate font-light>{{ `${(task.type === "suite" && task.projectName) ? `${task.projectName}/` : ''}${task.name}` }}</span>
```
sheremet-va, line 26: I don't like how it's shown. In CLI it's marked as `[${name}]` and has a color

sheremet-va (changes requested):

packages/ui/client/components/TaskItem.vue
g4rry420: @sheremet-va Like this ?
Screenshot 2023-07-04 204747

sheremet-va: yes

sheremet-va: Yellow color is unreadable in light theme:

Screenshot 2023-07-11 at 09 40 45

g4rry420: @sheremet-va How about we if use a bit of blackish-shade with yellow - `#7c7c05`, green - `#025402` and blue - `rgb(2 2 131)` color ?
Screenshot 2023-07-24 214457

sheremet-va: happy-dom still looks unreadable to me

g4rry420: @sheremet-va  How's this ?
Screenshot 2023-07-25 215100

sheremet-va: Looks readable. Is it only for light mode? Won't be readable in dark mode

userquin: Change colors based on theme, use `isDark` from here https://github.com/vitest-dev/vitest/blob/main/packages/ui/client/composables/dark.ts

packages/vitest/src/utils/graph.ts
```diff
@@ -29,7 +29,18 @@ export async function getModuleGraph(ctx: Vitest, id: string): Promise<ModuleGra
...
+  const workspaceModules = workspaceProject?.ctx.getModuleProjects(id)
+  const selectedModule = workspaceModules?.find(module => module.getName() === workspaceName)
+  let workspaceMod: ModuleNode | undefined
+  if (selectedModule) {
+    const { server, browser } = selectedModule
+    const mod = server.moduleGraph.getModuleById(id) || browser?.moduleGraph.getModuleById(id)
+    workspaceMod = mod
+  }
+
+  await get(ctx.server.moduleGraph.getModuleById(id) || workspaceMod)
```
sheremet-va, line 43: Shouldn't you fallback to `ctx.server.moduleGraph` only if you didn't find the `workspaceMod`? This line looks suspicious to me

packages/vitest/src/utils/graph.ts
```diff
@@ -29,7 +29,18 @@ export async function getModuleGraph(ctx: Vitest, id: string): Promise<ModuleGra
     graph[id] = (await Promise.all(mods.map(m => get(m, seen)))).filter(Boolean) as string[]
     return id
   }
-  await get(ctx.server.moduleGraph.getModuleById(id))
+
+  const workspaceProject = ctx.projects.find(project => project.getName() === workspaceName)
+  const workspaceModules = workspaceProject?.ctx.getModuleProjects(id)
+  const selectedModule = workspaceModules?.find(module => module.getName() === workspaceName)
```
sheremet-va, line 35: You do the same thing you did on #L33. Why?

You need to find the project by name and fallback to `ctx.getCoreWorkspaceProject().server || ctx.server`

sheremet-va (changes requested):

packages/vitest/src/utils/graph.ts
g4rry420: @sheremet-va  `getCoreWorkspaceProject()` is coming as `null` for `test/workspaces` folder.
Screenshot 2023-08-08 213513

Upon investigating, I have found that the conditons that executes the `createCoreWorkspace` function in `packages\vitest\src\node\core.ts` aren't satisfying

Screenshot 2023-08-08 213414

> You do the same thing you did on #L33. Why?

This might be hacky, but I have found only this way where we get the `workspaceMod`

sheremet-va: > `getCoreWorkspaceProject()` is coming as `null` for `test/workspaces` folder.

Yes, it will be null if there is no core project. This is why I added `|| ctx.server`. Core project is not created if main config file is not used by a workspace (so, when there is a global config and project configs and workspace specifies only project configs without manually using core project)

> This might be hacky, but I have found only this way where we get the `workspaceMod`

Then you must be doing something wrong?

packages/vitest/src/utils/graph.ts
```diff
@@ -29,7 +29,18 @@ export async function getModuleGraph(ctx: Vitest, id: string): Promise<ModuleGra
     graph[id] = (await Promise.all(mods.map(m => get(m, seen)))).filter(Boolean) as string[]
     return id
   }
-  await get(ctx.server.moduleGraph.getModuleById(id))
+
+  const workspaceProject = ctx.projects.find(project => project.getName() === workspaceName)
+  const workspaceModules = workspaceProject?.ctx.getModuleProjects(id)
```
sheremet-va, line 34: This is useless, you already have access to `ctx`

packages/vitest/src/utils/graph.ts
sheremet-va: Workspace project is just a project with the same name. If there is no such projects, just fallback to context which is always defined

TECH7Fox:
Any update on this? I would be happy to create another PR to get this fix merged.

sheremet-va:
> Any update on this? I would be happy to create another PR to get this fix merged.

Feel free to open a separate PR.

sheremet-va:
This is fixed in `main`.

Issue #3631, opened 2023-06-20
`retryCount` and `repeatCount` is 1 even if test run only once


sheremet-va:
### Describe the bug

`test.result.retryCount` and `test.result.repeatCount` are equal to 1 even if the test didn't fail or run only once:

https://github.com/vitest-dev/vitest/blob/1b0e6ad5d8cdcf404d8daba01524f1b26111b3ee/packages/runner/src/run.ts#L145

Also, if a test was repeated multiple times, the `retryCount` will not include previous retries, if there were any.

`retryCount` and `repeatCount` should be 0, if the test didn't fail and was running only once. 

**retryCount**

- If `retry` is `0`, then after the test fails the first time, we don't retry it. Set its state to "failed"
- If `retry` is `1`, then the test should retry one time after it fails.
- If `retry` is `2`, then the test should retry two times after it fails. If it fails after two times it should have state "failed" and `retryCount` as 2

**repeatCount**

- If `repeats` is `0`, then the test runs only _once_ (the original run, but no repeats)
- If `repeats` is `1`, then the test runs _two_ times (the original run, and 1 repeat)
- If `repeats` is `2`, then the test runs _three_ times (the original run, and 2 repeats)

**retryCount** and **repeatCount**

- If `repeats` and `retry` is `0`, then the test runs only _one_ time
- If `repeats` and `retry` both are `1`, then each repeat inherits retry, and it _can_ run 2-4 times:
  - original run passed, repeat passed
  - original run failed, retry passes, retry passes
  - original run failed, retry passes, repeat run failed, repeat run retried
  - `retryCount` in this case should be a total of all retries

### Reproduction

https://stackblitz.com/fork/github/vitest-dev/vitest/tree/main/examples/basic?initialPath=__vitest__/

### System Info

```shell
Any
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
[...]

PR #3638, opened 2023-06-21
refactor(core)!: make the default value of `retry` and `repeats` 0

Dunqing:
close: #3631

### DIFF
--- a/packages/runner/src/run.ts
+++ b/packages/runner/src/run.ts
@@ -140,7 +140,6 @@ export async function runTest(test: Test, runner: VitestRunner) {
       try {
         await runner.onBeforeTryTest?.(test, { retry: retryCount, repeats: repeatCount })
 
-        test.result.retryCount = retryCount
         test.result.repeatCount = repeatCount
 
         beforeEachCleanups = await callSuiteHook(test.suite, test, 'beforeEach', runner, [test.context, test.suite])
@@ -189,6 +188,7 @@ export async function runTest(test: Test, runner: VitestRunner) {
       if (retryCount < retry) {
         // reset state when retry test
         test.result.state = 'run'
+        test.result.retryCount = (test.result.retryCount ?? 0) + 1
       }
 
       // update retry info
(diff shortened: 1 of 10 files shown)


packages/runner/src/run.ts
```diff
@@ -132,12 +132,10 @@ export async function runTest(test: Test, runner: VitestRunner) {
...
 
-  const repeats = typeof test.repeats === 'number' ? test.repeats : 1
+  const { repeats, retry } = test
 
-  for (let repeatCount = 0; repeatCount < repeats; repeatCount++) {
-    const retry = test.retry || 1
-
-    for (let retryCount = 0; retryCount < retry; retryCount++) {
+  for (let repeatCount = 0; repeatCount <= repeats; repeatCount++) {
+    for (let retryCount = 0; retryCount <= retry; retryCount++) {
```
sheremet-va, line 138: There is still an issue that retry count resets with each repeat

sheremet-va (changes requested):
I think we can also add a test that calls startVitest and checks retry count on vitest.state

packages/runner/src/run.ts
Dunqing: In my opinion, I think repeat should reset the retry count, repeat should have a separate retry count.

Dunqing: Repeat does not make the test stable, when a test is bound to fail once or twice, then if he uses repeat the test will not succeed

sheremet-va: Repeat is not for stability, retry is. We lose all information about previous retries which is bad. We should either store it differently or increment it each time

Dunqing: "increment" sounds good

Dunqing: image

Now retryCount will increase with each repeat.

sheremet-va (approved):

sheremet-va:
I don't think this is a breaking change, it's publically available only with `createVitest` and it's not subjected to semver

Issue #3569, opened 2023-06-13
CTRL+C does not immediately stop running tests


nstepien:
### Describe the bug

<kbd>CTRL</kbd>+<kbd>C</kbd> does not immediately stop Vitest when tests are slow or stuck.

I've seen Vitest being impossible to terminate when tests are stuck because of an issue with fake timers for example, I had to completely close the terminal.
When test are just particularly slow it's frustrating that Vitest doesn't immediately stops when pressing <kbd>CTRL</kbd>+<kbd>C</kbd>.
I think this affects `bail: 1` too, not 100% sure.

Jest does not have this issue.

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-c5wvvm?file=test%2Fsuite1.test.ts,test%2Fslow.test.ts,test%2Fsuite2.test.ts,vite.config.ts,test%2Fsuite0.test.ts&initialPath=__vitest__/

1. run `npm run test:run slow`
2. press <kbd>CTRL</kbd>+<kbd>C</kbd> when the test runs
3. Vitest hangs and does not immediately terminate
   ![image](https://github.com/vitest-dev/vitest/assets/567105/93a70bb1-2303-416d-9f65-f4ed012d1761)

### System Info

```shell
System:
    OS: Windows 10 10.0.22621
    CPU: (64) x64 AMD Ryzen Threadripper 3970X 32-Core Processor
    Memory: 17.54 GB / 31.86 GB
  Binaries:
    Node: 20.2.0 - C:\Program Files\nodejs\node.EXE
    npm: 9.6.7 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22621.1778.0), Chromium (114.0.1823.43)
    Internet Explorer: 11.0.22621.1
  npmPackages:
    @vitejs/plugin-react: ^4.0.0 => 4.0.0
    @vitest/coverage-v8: ^0.32.0 => 0.32.0
    vite: ^4.3.5 => 4.3.9
    vitest: ^0.32.0 => 0.32.0
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
[...]

PR #3642, opened 2023-06-21
fix: `CTRL+C` to terminate run

AriPerkkio:
- Fixes https://github.com/vitest-dev/vitest/issues/3569

Instead of listening for `SIGINT`, let's listen for keypress and decide how to handle `CTRL` + `c`. The `process.on("SIGINT")` seems to be too difficult to handle: https://github.com/vitest-dev/vitest/issues/3569#issuecomment-1600330892.

https://github.com/vitest-dev/vitest/assets/14806298/51099026-4727-44e7-96e1-6620da7a6a09

Windows + Powershell:

https://github.com/vitest-dev/vitest/assets/14806298/8c884681-2b03-40d1-9e08-9f286c02359d

### DIFF
--- a/packages/vitest/src/node/cli-api.ts
+++ b/packages/vitest/src/node/cli-api.ts
@@ -87,10 +87,8 @@ export async function startVitest(
     return ctx
   }
 
-  if (process.stdin.isTTY && ctx.config.watch)
+  if (process.stdin.isTTY)
     registerConsoleShortcuts(ctx)
-  else
-    process.on('SIGINT', () => ctx.cancelCurrentRun('keyboard-input'))
 
   ctx.onServerRestart((reason) => {
     ctx.report('onServerRestart', reason)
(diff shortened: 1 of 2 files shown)


AriPerkkio:
Looks like this makes exiting run-mode stuck now... It's not visible on CI since it's not TTY. No idea how I missed that. @sheremet-va I'll need to fix this before next release. Looking into this now.

```
~/x/vitest/test/watch (main) $ pnpm run test --root fixtures --run

> @vitest/test-watch@ test /x/y/vitest/test/watch
> vitest "--root" "fixtures" "--run"


 RUN  v0.32.2 /x/y/vitest/test/watch/fixtures

 ✓ example.test.ts (1)
   ✓ getHello
 ✓ math.test.ts (1)
   ✓ sum

 Test Files  2 passed (2)
      Tests  2 passed (2)
   Start at  10:49:44
   Duration  162ms (transform 19ms, setup 0ms, collect 27ms, tests 2ms, environment 0ms, prepare 86ms)

close timed out after 10000ms
Tests closed successfully but something prevents Vite server from exiting
You can try to identify the cause by enabling "hanging-process" reporter. See https://vitest.dev/config/#reporters
```

PR #3643, opened 2023-06-21
test: update to the correct test

Dunqing:
I think I know what causes Ci to always fail. default reporter also prints all the test logs when it runs. However, as the test task is updated, some tests will no longer be printed out.

So in `runVitestCli`, the run logs of all tests will be stored in stdout, so `b1 test` may exist due to different machines and different run times, etc. Now I've added `✓` to them to make sure they are not included in the final test log.

And this change is consistent with the intent of the original test.


AriPerkkio (changes requested):
Maybe also increase the `chaiConfig.truncateThreshold` of `test/reporters/vitest.config.ts` as the current error messages do not show what the actual output was:

```
AssertionError: expected '? Input filename pattern › b✔ Input f…' to not include 'b1 test'
```

Example:

https://github.com/vitest-dev/vitest/blob/fbb56ad8ae5159f3c4fe2eaaf51f07eef1a2ca0a/test/fails/vite.config.ts#L6-L8

Dunqing:
> Maybe also increase the `chaiConfig.truncateThreshold` of `test/reporters/vitest.config.ts` as the current error messages do not show what the actual output was:
> 
> 
> 
> ```
> 
> AssertionError: expected '? Input filename pattern › b✔ Input f…' to not include 'b1 test'
> 
> ```
> 
> 
> 
> Example:
> 
> 
> 
> https://github.com/vitest-dev/vitest/blob/fbb56ad8ae5159f3c4fe2eaaf51f07eef1a2ca0a/test/fails/vite.config.ts#L6-L8

Wow, Thank you, I regret learning about it too late.

sheremet-va (approved):

Issue #3299, opened 2023-05-04
Remove polyfills in browser mode


sheremet-va:
### Clear and concise description of the problem

By providing polyfills we create a separate environment that is different from the usual browser one. This is dangerous because tests might run successfully, but the code will actually fail in production.

### Suggested solution

Remove polyfills.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #3648, opened 2023-06-23
deps(browser): remove modern-node-polyfills package

Marcel-G:
resolves: https://github.com/vitest-dev/vitest/issues/3299

The browser runner indiscriminately applies pollyfills for node builtin modules. This PR removes completely removes the dependence on `modern-node-polyfills`.

### DIFF
--- a/packages/browser/src/node/index.ts
+++ b/packages/browser/src/node/index.ts
@@ -2,14 +2,9 @@ import { fileURLToPath } from 'node:url'
 
 import { resolve } from 'node:path'
 import { builtinModules } from 'node:module'
-import { polyfillPath } from 'modern-node-polyfills'
 import sirv from 'sirv'
 import type { Plugin } from 'vite'
 import { injectVitestModule } from './esmInjector'
-
-const polyfills = [
-  'util',
-]
 
 // don't expose type to not bundle it here
 export default (project: any, base = '/'): Plugin[] => {
@@ -36,13 +31,12 @@ export default (project: any, base = '/'): Plugin[] => {
       },
     },
     {
-      name: 'modern-node-polyfills',
+      name: 'vitest:browser:tests',
       enforce: 'pre',
       config() {
         return {
           optimizeDeps: {
             exclude: [
-              ...polyfills,
               ...builtinModules,
               'vitest',
               'vitest/utils',
@@ -63,22 +57,17 @@ export default (project: any, base = '/'): Plugin[] => {
         }
       },
       async resolveId(id) {
-        if (!builtinModules.includes(id) && !polyfills.includes(id) && !id.startsWith('node:')) {
-          if (!/\?browserv=\w+$/.test(id))
-            return
+        if (!/\?browserv=\w+$/.test(id))
+          return
 
-          let useId = id.slice(0, id.lastIndexOf('?'))
-          if (useId.startsWith('/@fs/'))
-            useId = useId.slice(5)
+        let useId = id.slice(0, id.lastIndexOf('?'))
+        if (useId.startsWith('/@fs/'))
+          useId = useId.slice(5)
 
-          if (/^\w:/.test(useId))
-            useId = useId.replace(/\\/g, '/')
+        if (/^\w:/.test(useId))
+          useId = useId.replace(/\\/g, '/')
 
-          return useId
-        }
-
-        id = normalizeId(id)
-        return { id: await polyfillPath(id), moduleSideEffects: false }
+        return useId
       },
     },
     {
@@ -92,17 +81,4 @@ export default (project: any, base = '/'): Plugin[] => {
       },
     },
   ]
-}
-
-function normalizeId(id: string, base?: string): string {
-  if (base && id.startsWith(base))
-    id = `/${id.slice(base.length)}`
-
-  return id
-    .replace(/^\/@id\/__x00__/, '\0') // virtual modules start with `\0`
-    .replace(/^\/@id\//, '')
-    .replace(/^__vite-browser-external:/, '')
-    .replace(/^node:/, '')
-    .replace(/[?&]v=\w+/, '?') // remove ?v= query
-    .replace(/\?$/, '') // remove end query mark
 }
--- a/packages/browser/package.json
+++ b/packages/browser/package.json
@@ -44,7 +44,6 @@
   "dependencies": {
     "estree-walker": "^3.0.3",
     "magic-string": "^0.30.1",
-    "modern-node-polyfills": "^1.0.0",
     "sirv": "^2.0.3"
   },
   "devDependencies": {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -934,9 +934,6 @@ importers:
       magic-string:
         specifier: ^0.30.1
         version: 0.30.1
-      modern-node-polyfills:
-        specifier: ^1.0.0
-        version: 1.0.0(esbuild@0.18.11)(rollup@3.26.0)
       sirv:
         specifier: ^2.0.3
         version: 2.0.3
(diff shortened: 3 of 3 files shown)


sheremet-va:
We can also remove polyfills now with modern-node-polyfills package

packages/browser/src/node/index.ts
```diff
@@ -35,51 +29,6 @@ export default (project: any, base = '/'): Plugin[] => {
...
-            return
-
-          let useId = id.slice(0, id.lastIndexOf('?'))
-          if (useId.startsWith('/@fs/'))
-            useId = useId.slice(5)
-
-          if (/^\w:/.test(useId))
-            useId = useId.replace(/\\/g, '/')
-
-          return useId
```
userquin, line 76: also needed

Marcel-G:
The name of this plugin now doesn't seem appropriate, any naming ideas?

https://github.com/vitest-dev/vitest/blob/064f56d448bec374ea44042934c5eddaf2150b2a/packages/browser/src/node/index.ts#L34

packages/browser/src/node/index.ts
```diff
@@ -29,6 +30,47 @@ export default (project: any, base = '/'): Plugin[] => {
...
+              'vitest > pretty-format',
+              'vitest > pretty-format > ansi-styles',
+              'vitest > pretty-format > ansi-regex',
+              'vitest > chai',
+            ],
+          },
+        }
+      },
+      async resolveId(id) {
+        if (!builtinModules.includes(id) && !id.startsWith('node:')) {
```
userquin, line 59: can you remove this redundant if statement?

userquin: it seems we need if statement back, sorry...

Marcel-G:
As far as I can tell `util` is still included by `@sinonjs/fake-timers`:

https://github.com/vitest-dev/vitest/blob/c8c1b4a6a32177934adda0d5705a1f7112c2d76b/packages/vitest/src/integrations/mock/timers.ts#L38

sheremet-va:
> As far as I can tell `util` is still included by `@sinonjs/fake-timers`:
> 
> https://github.com/vitest-dev/vitest/blob/c8c1b4a6a32177934adda0d5705a1f7112c2d76b/packages/vitest/src/integrations/mock/timers.ts#L38

1. It is included dynamically if running inside Node.js
2. We bundle it

Marcel-G:
Heres the error in the browser if we don't include `util` polyfill

```
__vite-browser-external:util:3 Uncaught (in promise) Error: Module "util" has been externalized for browser compatibility. Cannot access "util.default" in client code.  See http://vitejs.dev/guide/troubleshooting.html#module-externalized-for-browser-compatibility for more details.
    at Object.get (__vite-browser-external:util:3:11)
    at withGlobal (vendor-vi.f4f5ef45.js:1370:47)
    at vendor-vi.f4f5ef45.js:3017:32
    at vendor-vi.f4f5ef45.js:3023:3
```

Marcel-G:
Heres the error in the browser if we don't include `util` pollyfil

```
__vite-browser-external:util:3 Uncaught (in promise) Error: Module "util" has been externalized for browser compatibility. Cannot access "util.default" in client code.  See http://vitejs.dev/guide/troubleshooting.html#module-externalized-for-browser-compatibility for more details.
    at Object.get (__vite-browser-external:util:3:11)
    at withGlobal (vendor-vi.f4f5ef45.js:1370:47)
    at vendor-vi.f4f5ef45.js:3017:32
    at vendor-vi.f4f5ef45.js:3023:3
```

sheremet-va:
> Heres the error in the browser if we don't include `util` pollyfil
> 
> ```
> __vite-browser-external:util:3 Uncaught (in promise) Error: Module "util" has been externalized for browser compatibility. Cannot access "util.default" in client code.  See http://vitejs.dev/guide/troubleshooting.html#module-externalized-for-browser-compatibility for more details.
>     at Object.get (__vite-browser-external:util:3:11)
>     at withGlobal (vendor-vi.f4f5ef45.js:1370:47)
>     at vendor-vi.f4f5ef45.js:3017:32
>     at vendor-vi.f4f5ef45.js:3023:3
> ```

Probably because we have `process` mocked, which we also shouldn't do, but it should be addressed in a separate issue.

Marcel-G:
> Probably because we have `process` mocked, which we also shouldn't do, but it should be addressed in a separate issue.

Something like this https://github.com/vitest-dev/vitest/pull/3701?

sheremet-va:
> > Probably because we have `process` mocked, which we also shouldn't do, but it should be addressed in a separate issue.
> 
> Something like this #3701?

Yes

sheremet-va:
> Heres the error in the browser if we don't include `util` pollyfil

Why are tests passing then 🤔

Marcel-G:
> > Heres the error in the browser if we don't include `util` pollyfil
> 
> Why are tests passing then 🤔

Sorry, In the current state `modern-node-polyfills` is still applied to `util` but I'll fix it up

Marcel-G:
OK, the tests will fail until https://github.com/vitest-dev/vitest/pull/3701 is in

sheremet-va (changes requested):
`normalizeId` is not used anymore

Marcel-G:
> `normalizeId` is not used anymore

Very true! Removed ✅

Marcel-G:
@sheremet-va are the tests generally a bit flakey? It seems to be a different spec failing each time

sheremet-va:
> @sheremet-va are the tests generally a bit flakey? It seems to be a different spec failing each time

Some tests are flaky. Your PR is open until we have a browser test in ecosystem-ci

Marcel-G:
> Some tests are flaky. Your PR is open until we have a browser test in ecosystem-ci

@sheremet-va what is the status of that? is it something I can help with?

sheremet-va:
> > Some tests are flaky. Your PR is open until we have a browser test in ecosystem-ci
> 
> @sheremet-va what is the status of that? is it something I can help with?

Sorry for keeping you waiting! I am working on a small test suite for ecosystem-ci now, we can probably merge this PR today or next week.

AriPerkkio:
How should the `util` be polyfilled now? It's still used by `@sinon-js/fake-timers`. v1-beta fails with error `SyntaxError: The requested module '/node_modules/.pnpm/util@0.12.5/node_modules/util/util.js?v=c8f99d41' does not provide an export named '__vi_inject__'`. And when adding `"vitest > util"` to `optimizeDeps.include`, it fails with:

```
ReferenceError: process is not defined
 ❯ node_modules/.pnpm/util@0.12.5/node_modules/util/util.js ../../../../node_modules/.vite/deps/vitest___util.js:1308:5
 ❯ __require ../../../../node_modules/.vite/deps/chunk-UXIASGQL.js:9:50
 ❯ ../../../../node_modules/.vite/deps/vitest___util.js:1799:25
 ```

sheremet-va:
> How should the `util` be polyfilled now?

Vitest doesn't use it and as far as I know `@sinonjs/fake-timers` checks it at runtime, so this _shouldn't_ be a problem? If it is, maybe we can patch it and inline the package?

> And when adding `"vitest > util"` to `optimizeDeps.include`, it fails with:

The error looks correct to me since we don't define a global `process` variable since it doesn't exist in the browser.

AriPerkkio:
> Vitest doesn't use it 

Looks like `@vitest/browser` imports `vitest/runners`, which uses `vi`. Importing `vi` brings in inlined fake-timers which has `process`.

sheremet-va:
> > Vitest doesn't use it
> 
> Looks like `@vitest/browser` imports `vitest/runners`, which uses `vi`. Importing `vi` brings in inlined fake-timers which has `process`. The `vendor-vi` has `import require$$2 from 'util';`.

fake-timers have a check for `process`. What I assume happened in your case is that polyfill is using `process` for some reason.

> The `vendor-vi` has `import require$$2 from 'util';`.

Yeah, this is why I am saying we should just patch fake timers to remove this check.

PR #3651, opened 2023-06-23
feat: support accessing other fixtures in fixture function

### LINKED ISSUE #3626: Support accessing other fixtures in the fixture function when using `test.extend`

This idea comes from the [question](https://github.com/vitest-dev/vitest/pull/3554#issuecomment-1597344051) by @outerlook. 

When setting up fixtures in `test.extend`,  we may dependent on other fixtures. Would be very useful to have this feature.

```ts
export const newTest = test.extend({
  server: async (use) => {
    const server = await startServer();
    await use(server);

    await server.destroy();
  },
  // client depends on server
  client: async (use, { server }) => {
    const client = await new Client({ apiUrl: server.apiUrl });
    await use(client);

    await client.destroy();
  }
});
```

### Suggested solution

Make other fixtures accessible as the second argument of each fixture function.

### Alternative

_No response_

### Additional context

_No response_

### Validations

fenghan34:
Closes #3626


packages/runner/src/fixture.ts
```diff
@@ -1,66 +1,162 @@
...
 
-      const colonIndex = _prop.indexOf(':')
-      if (colonIndex > 0)
-        _prop = _prop.slice(0, colonIndex).trim()
+  return pendingFixtures
+}
+
+function getFixtureFnDepProps(fn: Function, allProps: string[]) {
+  if (fn.length !== 2)
+    throw new Error('fixture function should have two arguments, the fist one is the use function that should be called with fixture value, and the second is other fixtures that should be used with destructured expression. For example, `async ({ a, b }, use) => { await use(a + b) }`')
```
sheremet-va, line 107: This looks misleading 😄

sheremet-va (changes requested):

packages/runner/src/fixture.ts
fenghan34: Aha, I made a mistake. 😅

sheremet-va:
I wonder if it would be a good idea to support the same arguments order, so we also support third-party playwright plugins?

sheremet-va:
Also, would be nice to release this in the next version alongside the support for `test.extend`. Since we didn't release it yet, changing arguments order is not a breaking change.

fenghan34:
> I wonder if it would be a good idea to support the same arguments order, so we also support third-party playwright plugins?

We can do that, but I think it's kind of strict that users must follow this pattern `async ({}, use) => {}` to declare a fixture function. Btw, what are third-party playwright plugins?

fenghan34:
In Playwright, users must use the object destructuring pattern `{a, b, c}` to access fixtures through the first argument both in the test and fixture function, the default function parameters `({ a = 1 }) => {}` and rest parameters `({ a, b, ...rest }) => {}` are not allowed. 

I prefer to make it more flexible with mentioned features above, so the following test/fixture function declarations would work with the current implementation but not Playwright. 

```ts
// Be aware the second argument is the `use` function in Playwright 
const myTest = test.extend({
  a: async (use) => {
    await use()
  },
  // b depends on a and c
  b: async (use, context) => {
    await use()
  },
  // c depends on a and b
  c: async (...args) => {
  	await args[0]()
  },
})

const myTest2 = myTest.extend({
  // d depends on a
  d: async (use, { a = 1 }) => {
  	await use()
  },
  // e depends on a, b, c and d
  e: async (use, { a, ...rest }) => {
    await use()
  }
})

myTest2('', (context) => {}) // all fixtures will be initialized
myTest2('', (...args) => {}) // all fixtures will be initialized
myTest2('', ({ a, ...rest }) => {}) // all fixtures will be initialized
myTest2('', ({ d = 1 }) => {}) // only a and d will be initialized
```

This gives users more convenience but also risks of circular dependency when declaring fixture functions with `(...rest) => {}`, `(use, { a, ...rest}) => {}` or `(use, context) => {}`. Would like to hear your thoughts 😄.

sheremet-va:
> Would like to hear your thoughts 😄.

I think it's better to be more strict. We don't know the order of fixtures that we need to call to support `context` without destructuring, and I don't see big benefits for this pattern - it's better to use only things you need without initializing _everything_ (even things you don't rely on).

> the default function parameters

It doesn't make sense to support a default value here.

> what are third-party playwright plugins?

Any package that uses `test.extend` under the hood.

fenghan34:
> I think it's better to be more strict. We don't know the order of fixtures that we need to call to support `context` without destructuring, and I don't see big benefits for this pattern - it's better to use only things you need without initializing _everything_ (even things you don't rely on).

I see. Make sense to me. So do we need to restrict the test function(destructuring context) as well?

sheremet-va:
> I see. Make sense to me. So do we need to restrict the test function(destructuring context) as well?

I would think so

fenghan34:
> I would think so

Got it, thank you for replying, will push more commits.

sheremet-va:
@fenghan34 any news? We are planning to make a release today 👀

fenghan34:
> @fenghan34 any news? We are planning to make a release today 👀 

I'm writing tests，will update it today😄

docs/guide/test-context.md
```diff
@@ -115,15 +115,17 @@ const myTest = test.extend({
...
 
 // todosFn will run
-myTest('', ({ todos }) => {}) // only todos is available
-myTest('', (context) => {}) // both are available
-myTest('', ({ archive, ...rest }) => {}) // both are available
+myTest('', ({ todos }) => {})
 ```
 
+::: warning
+When using `test.extend()`, you should always use the object destructuring pattern `{ todos }` to access context both in fixture function and test function.
```
sheremet-va, line 126: I think we should allow normal access (without destructuring) if there are no fixtures added with `test.extend` for backwards compatibility

sheremet-va (changes requested):

docs/guide/test-context.md
fenghan34: Make sense

sheremet-va (approved):

PR #3674, opened 2023-06-26
feat(browser): Add `none` browser provider

### LINKED ISSUE #3084: `none` browser provider

While working on landing #3079, I found out that the initial feature of the browser mode which simply is just starting a server and waiting for the user to navigate to that URL in their own browser is removed! I believe we need to have it again since it's a huge DX win and can play really well with vitest ui for the users.  

### Suggested solution

Having a new provider as @sheremet-va said which handles this case, I assume not so much should be gone into this browser provider and only the provider methods should be filled (for the sake of the provider)! I'd name this provider the `none` provider which does not open any browser for you as wd/playwright does.

### Alternative

It was there before #2999.

### Additional context

#3079 

### Validations

userquin:
This PR should be merged once https://github.com/vitest-dev/vitest/pull/3584 merged.

I have left `webdriverio` as the default provider, ready to be used in `CI`.

We need to add some entries in the docs for local development and `CI`.

/cc @Aslemammad 

closes #3084
supersedes #3273

### DIFF
--- a/packages/vitest/src/integrations/browser.ts
+++ b/packages/vitest/src/integrations/browser.ts
@@ -1,24 +1,22 @@
+import { provider } from 'std-env'
 import { NoneBrowserProvider } from '../node/browser/none'
 import { PlaywrightBrowserProvider } from '../node/browser/playwright'
 import { WebdriverBrowserProvider } from '../node/browser/webdriver'
 import type { BrowserProviderModule, ResolvedBrowserOptions } from '../types/browser'
-import { StackBlitzBrowserProvider } from '../node/browser/stackblitz'
 
 interface Loader {
   executeId: (id: string) => Promise<{ default: BrowserProviderModule }>
 }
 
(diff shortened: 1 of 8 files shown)


sheremet-va:
Why do we need a separate stakblitz provider? The main reason `none` exists is because of the stackblitz usecase.

userquin:
Because we cannot open browser, server.open cannot be set to true, I need to check it (maybe tmr).

If server.open works, we can remove it.

sheremet-va:
> Because we cannot open browser, server.open cannot be set to true, I need to check it (maybe tmr).
> 
> If server.open works, we can remove it.

You can check stackblitz with `std-env`. We already do this for coverage:

https://github.com/vitest-dev/vitest/blob/445a7b6d4ab906c05c211e7b1a8102779fa35187/packages/coverage-c8/src/provider.ts#L84

userquin:
Update provider logic to fallback to none provider when running in StackBlitz and using `webdriverio` or `playwright` providers via `std-env`.

PR #3698, opened 2023-06-28
fix!(browser): print correct stacktrace

### LINKED ISSUE #3628: Improve stack traces in browser mode
### Describe the bug

Currently, stack traces lead to URLs instead of filepaths:

```
❯ http:/localhost:63315/Users/user/Projects/vitest/test/browser/test/failing.test.ts?browserv=1687246540867:4:26
❯ runTests http:/localhost:63315/assets/index-0eaf4198.js:933:7
❯ WebSocket.<anonymous> http:/localhost:63315/assets/index-0eaf4198.js:901:3
```

Expected stack trace (URLs are converted into file paths relative to root, internal functions are removed):

```
❯ test/browser/test/failing.test.ts:4:26
```

### Reproduction

Download and run `npm run test`

https://stackblitz.com/edit/vitest-dev-vitest-kuisqc?file=vite.config.ts,test%2Fbasic.test.ts&initialPath=__vitest__/

### System Info

```shell
Any
```

### Used Package Manager

npm

### Validations

sheremet-va:
Fixes #3628


robertjpayne:
@sheremet-va I'm having an interesting issue which led me onto this PR. I'm trying to ascertain why with the parser options in printError of `vitest/src/node/error.ts` only uses a source map when running in the browser?

```
const parserOptions: StackTraceParserOptions = {
    // only browser stack traces require remapping <------------
    getSourceMap: file => project.getBrowserSourceMapModuleById(file),
  }
```

I'm trying to create a reproduction but I have a project in which the stack traces are wrong because the source maps are not being loaded. If I manually make this file use `project.getSourceMapModuleById(file)` instead it works fine.

sheremet-va:
> I'm trying to create a reproduction but I have a project (node runners only) in which the stack traces are wrong because the source maps are not being used.

Because they are patched before that, and should already be resolved.

PR #3702, opened 2023-06-29
Add missing JSDom living keys

DerZade:
This PR adds `SubmitEvent`, `DOMRectReadOnly` and `DOMRect` to the JSDom living keys so that they are available in tests when running in the JSDom environment.

- `SubmitEvent` was added to the living interfaces in version 21.1.0 (see [commit](https://github.com/jsdom/jsdom/commit/79c351bc5a#diff-192ca7362e7fb7c4aeae2fa4d091fc7fc66238b69bc2da2ca4e45e73b7511c03)) 
- `DOMRectReadOnly` and `DOMRect` were added to the living interfaces in version 22.1.0 (see [commit](https://github.com/jsdom/jsdom/commit/c9d6b72500#diff-192ca7362e7fb7c4aeae2fa4d091fc7fc66238b69bc2da2ca4e45e73b7511c03))

### DIFF
--- a/packages/vitest/src/integrations/env/jsdom-keys.ts
+++ b/packages/vitest/src/integrations/env/jsdom-keys.ts
@@ -115,6 +115,7 @@ const LIVING_KEYS = [
   'StorageEvent',
   'ProgressEvent',
   'PageTransitionEvent',
+  'SubmitEvent',
   'UIEvent',
   'FocusEvent',
   'InputEvent',
--- a/packages/vitest/src/integrations/env/jsdom-keys.ts
+++ b/packages/vitest/src/integrations/env/jsdom-keys.ts
@@ -164,6 +164,8 @@ const LIVING_KEYS = [
   'AbortController',
   'AbortSignal',
   'ArrayBuffer',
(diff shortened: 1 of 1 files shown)


sheremet-va (approved):

sheremet-va:
Would be nice to have a test that confirms these globals are available. You can test it in `dom.test.ts`, we have some tests for Image/Audio there

DerZade:
> Would be nice to have a test that confirms these globals are available. You can test it in `dom.test.ts`, we have some tests for Image/Audio there

The globals may or may not be available, depending on the jsdom version that is used. But to my understanding, it should not break anything if the properties are not available. The `LIVING_KEYS` array is (together with the `OTHER_KEYS`) more like a whitelist. It contains all keys that will be copied over from the JSDOM instance, doesn't matter if they are defined or undefined. If any globals are missing, `undefined` is simply copied over and they are not available in the tests.

I tried running my vitest version with jsdom version 21 and it worked fine, except that `DOMRect` is just `undefined` inside tests, which is expected since it has only been added in version 22.

Issue #3523, opened 2023-06-05
Add more examples with enabled `deps.optimizer`


sheremet-va:
### Clear and concise description of the problem

Since `deps.inline` is deprecated (#3521), we should change our existing examples that rely on it like Solid.js to use optimizer instead.

### Suggested solution

Enable `deps.optimizer` in JSDOM/happy-dom examples. Since Node.js tests don't usually benefit from it, we can ignore them if they already don't rely on `deps.inline`. 

We should also add examples where optimizer helps with alias incompatibility like preact (#3502).

This PR should also remove experimental status from `deps.experimentalOptimizer`.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

eryue0220:
Can I pick it up？

sheremet-va:
> Can I pick it up？

Sure

Issue #1652, opened 2022-07-16
Can not run tests when using preact compat alias


wight554:
### Describe the bug

When package tries to import react by alias, it picks CJS version of preact specific module, which leads to following error:
```
 FAIL  src/app.test.tsx > should render app
TypeError: Cannot read properties of undefined (reading '__H')
 ❯ l node_modules/preact/hooks/dist/hooks.js:1:201
 ❯ d node_modules/preact/hooks/dist/hooks.js:1:600
 ❯ Object.exports.useRef node_modules/preact/hooks/dist/hooks.js:1:2323
 ❯ d.BrowserRouter [as constructor] node_modules/react-router-dom/umd/react-router-dom.development.js:80:28
 ❯ d.O [as render] ../../../../../../../../../C:/Users/wight/Documents/preact-vitest-repro/node_modules/preact/dist/preact.mjs:1:8158
 ❯ j ../../../../../../../../../C:/Users/wight/Documents/preact-vitest-repro/node_modules/preact/dist/preact.mjs:1:5870
 ❯ w ../../../../../../../../../C:/Users/wight/Documents/preact-vitest-repro/node_modules/preact/dist/preact.mjs:1:2137
 ❯ j ../../../../../../../../../C:/Users/wight/Documents/preact-vitest-repro/node_modules/preact/dist/preact.mjs:1:6138
 ❯ P ../../../../../../../../../C:/Users/wight/Documents/preact-vitest-repro/node_modules/preact/dist/preact.mjs:1:8272
 ❯ ../../../../../../../../../C:/Users/wight/Documents/preact-vitest-repro/node_modules/@testing-library/preact/dist/esm/pure.mjs:52:7
```

### Reproduction

https://github.com/wight554/preact-vitest-repro/

### System Info

```shell
System:
    OS: Windows 10 10.0.22621
    CPU: (16) x64 AMD Ryzen 7 3700X 8-Core Processor
    Memory: 16.89 GB / 31.93 GB
  Binaries:
    Node: 16.14.0 - c:\program files\nodejs\node.EXE
    Yarn: 1.22.4 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 8.3.1 - c:\program files\nodejs\npm.CMD
  Browsers:
    Chrome: 103.0.5060.114
    Edge: Spartan (44.22621.290.0), Chromium (103.0.1264.49)
    Internet Explorer: 11.0.22621.1
  npmPackages:
[...]

PR #3712, opened 2023-07-01
feat: remove optimizer experimental and add preact examples

eryue0220:
Fix [issue #3523](https://github.com/vitest-dev/vitest/issues/3523).


docs/config/index.md
```diff
@@ -107,10 +107,10 @@ When defined, Vitest will run all matched files with `import.meta.vitest` inside
 
 Handling for dependencies resolution.
 
-#### deps.experimentalOptimizer
+#### deps.optimizer
 
 - **Type:** `{ ssr?, web? }`
-- **Version:** Since Vitest 0.29.0
+- **Version:** Since Vitest 0.32.3
```
sheremet-va, line 113: 0.34.0

sheremet-va: Should also mention that it was named `deps.experimentalOptimizer` before

examples/preact/vitest.config.ts
```diff
@@ -0,0 +1,23 @@
...
+
+import { defineConfig } from 'vite'
+
+export default defineConfig({
+  test: {
+    deps: {
+      optimizer: {
+        web: {
+          enabled: true,
+          include: [
```
sheremet-va, line 11: The idea with preact example is to show how to use react dependencies that alias to preact: https://github.com/vitest-dev/vitest/issues/1652

You don't need to include preact deps in optimizer. You need to include dependencies that rely on react so aliasing starts working:

```ts
{
  include: ['react-router-dom']
}
```

examples/preact-testing-lib/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "@vitest/example-preact-testing-lib",
```
sheremet-va, line 2: We don't need 2 examples. Let's have "preact" that uses testing lib

sheremet-va (changes requested):

sheremet-va (changes requested):

sheremet-va:
I pushed changes, but it closed the PR, hm 🤔

Issue #3659, opened 2023-06-25
Tests using different environments should always be isolated


AriPerkkio:
### Describe the bug

- Ref: https://github.com/vitest-dev/vitest/pull/3203#discussion_r1238299921

When threads are enabled and there are multiple test environments used by the test cases, there should be no multiple test environments used in a single thread. Instead the test cases should be grouped by environment and run in separate threads. 

It's already intended in code but it's not working:

https://github.com/vitest-dev/vitest/blob/fbb56ad8ae5159f3c4fe2eaaf51f07eef1a2ca0a/packages/vitest/src/node/pools/threads.ts#L165-L166

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-ihs847?file=vite.config.ts

```js
import { defineConfig } from 'vitest/config';

export default defineConfig({
  test: {
    singleThread: true,
    isolate: false,
  },
});
```

```js
/**
 * @vitest-environment jsdom
 */

import { test, expect } from 'vitest';

test('Node globals not found', async () => {
  globalThis.__leaking_from_jsdom = 'leaking';
  expect(globalThis.__leaking_from_node).toBe(undefined);
});
```

```js
/**
 * @vitest-environment node
 */

import { test, expect } from 'vitest';

test('jsdom globals not found', async () => {
  globalThis.__leaking_from_node = 'leaking';
  expect(globalThis.__leaking_from_jsdom).toBe(undefined);
});
```

### System Info

```shell
Stackblitz
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #3715, opened 2023-07-02
fix: isolate workers between envs and workspaces

AriPerkkio:
- Fixes https://github.com/vitest-dev/vitest/issues/3659
- Requires: https://github.com/tinylibs/tinypool/pull/63
- Enables: https://github.com/vitest-dev/vitest/pull/3491

When running files with threads enabled, make sure workers are recycled when ever environment options or workspace changes. 

Also noticed that when running tests of workspace through `startVitest` Node API, the logger does not include project names in log entries.

### DIFF
--- a/packages/vitest/src/node/pools/threads.ts
+++ b/packages/vitest/src/node/pools/threads.ts
@@ -147,9 +147,29 @@ export function createThreadsPool(ctx: Vitest, { execArgv, env }: PoolProcessOpt
 
       if (multipleThreads.length) {
         const filesByEnv = await groupFilesByEnv(multipleThreads)
-        const promises = Object.values(filesByEnv).flat()
-        const results = await Promise.allSettled(promises
-          .map(({ file, environment, project }) => runFiles(project, getConfig(project), [file], environment, invalidates)))
+        const files = Object.values(filesByEnv).flat()
+        const results: PromiseSettledResult<void>[] = []
+
+        if (ctx.config.isolate) {
+          results.push(...await Promise.allSettled(files.map(({ file, environment, project }) =>
+            runFiles(project, getConfig(project), [file], environment, invalidates))))
+        }
+        else {
+          // When isolation is disabled, we still need to isolate environments and workspace projects from each other.
+          // Tasks are still running parallel but environments are isolated between tasks.
+          const grouped = groupBy(files, ({ project, environment }) => project.getName() + environment.name + JSON.stringify(environment.options))
+
+          for (const group of Object.values(grouped)) {
+            // Push all files to pool's queue
+            results.push(...await Promise.allSettled(group.map(({ file, environment, project }) =>
+              runFiles(project, getConfig(project), [file], environment, invalidates))))
+
+            // Once all tasks are running or finished, recycle worker for isolation.
+            // On-going workers will run in the previous environment.
+            await new Promise<void>(resolve => pool.queueSize === 0 ? resolve() : pool.once('drain', resolve))
+            await pool.recycleWorkers()
+          }
+        }
 
         const errors = results.filter((r): r is PromiseRejectedResult => r.status === 'rejected').map(r => r.reason)
         if (errors.length > 0)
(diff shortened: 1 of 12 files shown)


packages/vitest/src/node/pools/threads.ts
```diff
@@ -147,13 +147,32 @@ export function createThreadsPool(ctx: Vitest, { execArgv, env }: PoolProcessOpt
 
       if (multipleThreads.length) {
```
AriPerkkio, line 148: This and the `if (singleThreads.length) {` block below are now quite similar. I guess these could be merged into a single one..?

sheremet-va: They shouldn't be similar though, every worker should be a new one

AriPerkkio: > every worker should be a new one

I think my earlier assumptions were incorrect. Are the statements below right?

- `--single-thread`
  - Workers should be recycled when environment options change
  - Workers should be recycled when workspace project changes
  - Multiple files should be run in a single worker
  - `--isolate true/false` should have no effect, since only one worker is used..? Should we actually raise an error when such condition is used, like we do with other conflicting options in `resolveConfig`?

- `--threads`
  - Workers should be recycled when environment options change
  - Workers should be recycled when workspace project changes
  - When `--isolate`, workers should be recycled between files
  - When `--no-isolate`, workers are not recycled between files

sheremet-va: > `--isolate true/false` should have no effect

It actually has an effect, because it doesn't reset the module graph and mock registry between tests

> `--threads`

I think the current implementation expects `--isolate` to be enabled. From what I see it's possible to have 4 workers created, but only one would be busy (but all 4 workers are already created), so it's kind of a waste. It would be good to have more things running in parallel (simplest example: 4 workspace projects with 1 test inside will run one test after another instead of 4 workers in parallel). I might be wrong here, feel free to correct me.

But yes, recycling options should be correct I think.

AriPerkkio: > It would be good to have more things running in parallel (simplest example: 4 workspace projects with 1 test inside will run one test after another instead of 4 workers in parallel)

Right, we actually don't need any changes to the `--threads` + `--isolate` part. We can fill the pool with tasks as before and rely on `isolateWorkers`. Tests using different workspace projects and environments are run parallel as before.

But the case with `--threads` + `--no-isolate` part might be tricky, as we still need to isolate between workspace projects and environments. Easiest way to achieve this is to group all tests by environments and workspaces, and then run these test sets sequentially and call `pool.recycleWorkers` in between. But there may be idle workers waiting for next set's tests in this case.  

I'll need to check if we could fill the pool with all tasks and call `pool.recycleWorkers()` between the tasks when needed. Some on-going workers can then be running in different environments, but all new ones are started with their environments. But in this case the order of tasks has to be stable.

The `--single-thread` case seems simple and should be correct already in this PR.

AriPerkkio: This is now implemented. When `--threads` + `--no-isolate`, the tasks are run parallel as efficient as possible. This is done by checking whether pool has any tasks in queue, and once the queue is empty, we can recycle workers and push more tasks to queue. In practice the pool is running different environment + workspace projects in parallel.

packages/vitest/src/node/pools/threads.ts
```diff
@@ -147,9 +147,29 @@ export function createThreadsPool(ctx: Vitest, { execArgv, env }: PoolProcessOpt
...
+
+          for (const group of Object.values(grouped)) {
+            // Push all files to pool's queue
+            results.push(...await Promise.allSettled(group.map(({ file, environment, project }) =>
+              runFiles(project, getConfig(project), [file], environment, invalidates))))
+
+            // Once all tasks are running or finished, recycle worker for isolation.
+            // On-going workers will run in the previous environment.
+            await new Promise<void>(resolve => pool.queueSize === 0 ? resolve() : pool.once('drain', resolve))
+            await pool.recycleWorkers()
```
AriPerkkio, line 170: To summarize this:
- Pool is filled with tasks that have same options (env, workspace)
- When pool takes the last task from queue, we tell pool to recycle all workers and fill queue with tasks that have new options
- The earlier tasks will run and recycle once they finish
- New tasks in the queue will use fresh workers

This guarantees that tasks with different options are run parallel.

PR #3729, opened 2023-07-03
fix!: default glob

### LINKED ISSUE #3530: `src/test.ts` is picked up as a test
### Describe the bug

This may not be a bug but I would consider it non-intuitive.

A file named `src/test.ts` is picked up as a test. I would suggest that `test.ts` should be required as a suffix rather than directly as a file name.

**Context**:
Our CI started failing because we have a `test` command in our source code: https://github.com/nuxt/nuxt/actions/runs/5193588330/jobs/9364317315?pr=21410

Introduced in glob changes in https://github.com/vitest-dev/vitest/pull/3392.

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-e9txbr?file=test.ts,test/basic.test.ts&initialPath=__vitest__/

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    @vitest/ui: latest => 0.32.0 
    vite: latest => 4.3.9 
    vitest: latest => 0.32.0
```

### Used Package Manager

npm

### Validations

so1ve:
close #3530


antfu (approved):
Looks good to me. I think we should have this as soon before the original breaking change gets spread wide.

As mentioned in https://github.com/vitest-dev/vitest/issues/3530#issuecomment-1620440424, I do think we should also improve our documentation on the Jest migration, it can be another PR tho.

antfu:
@so1ve any thing missing from making it ready to review?

so1ve:
@antfu I think it is ready for review now!

sheremet-va:
> Looks good to me. I think we should have this as soon before the original breaking change gets spread wide.

I'm fine with releasing 0.33.0 with just this change.

sheremet-va (approved):

so1ve:
...Forgot to use squash merge?

PR #3737, opened 2023-07-04
docs: improve globs

nickmccurdy:
Adds documentation that I originally intended on adding to #3392

### DIFF
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -83,7 +83,7 @@ All configuration options that are not supported inside a [workspace](/guide/wor
 - **Type:** `string[]`
 - **Default:** `['**/__tests__/**/*.?(c|m)[jt]s?(x)', '**/?(*.){test,spec}.?(c|m)[jt]s?(x)']`
 
-Files to include in the test run, using glob pattern.
+Files to include in the test run, using glob patterns. By default it looks for files in any folders named `__tests__` with the extensions `.js`,`.mjs`, `.cjs`, `.ts`, `.mts`, `.cts`, `.jsx`, and `.jsx` . It also includes files ending with `.test` or `.spec` along with files just named `test` or `spec` (before a valid extension). These patterns are compatible with Jest's default patterns, but with additional extensions.
 
 ### exclude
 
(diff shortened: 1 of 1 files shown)


sheremet-va (approved):

docs/config/index.md
```diff
@@ -83,7 +83,7 @@ All configuration options that are not supported inside a [workspace](/guide/wor
 - **Type:** `string[]`
 - **Default:** `['**/__tests__/**/*.?(c|m)[jt]s?(x)', '**/?(*.){test,spec}.?(c|m)[jt]s?(x)']`
 
-Files to include in the test run, using glob pattern.
+Files to include in the test run, using glob patterns. By default it looks for files in any folders named `__tests__` with the extensions `.js`,`.mjs`, `.cjs`, `.ts`, `.mts`, `.cts`, `.jsx`, and `.jsx` . It also includes files ending with `.test` or `.spec` along with files just named `test` or `spec` (before a valid extension). These patterns are compatible with Jest's default patterns, but with additional extensions.
```
dammy001, line 86:
```suggestion
Files to include in the test run, using glob patterns. By default it looks for files in any folders named `__tests__` with the extensions `.js`,`.mjs`, `.cjs`, `.ts`, `.mts`, `.cts`, `.jsx`, and `.tsx`. It also includes files ending with `.test` or `.spec` along with files just named `test` or `spec` (before a valid extension). These patterns are compatible with Vitest default patterns but with additional extensions.
```

docs/config/index.md
```diff
@@ -1372,14 +1372,15 @@ You can also pass down a path to custom binary or command name that produces the
 - **Type**: `string[]`
 - **Default**: `['**/?(*.){test,spec}-d.?(c|m)[jt]s?(x)']`
 
-Glob pattern for files that should be treated as test files
+Files to include in typechecking test runs, using glob patterns. By default all tests inside files ending with `test-d` along with files just named `test-d` are considered type tests.
```
dammy001, line 1375:
```suggestion
Files to include in typechecking test runs, using glob patterns. By default, all tests inside files ending with `test-d` along with files just named `test-d` are considered type tests. 
```

antfu (changes requested):
I don't think we should move forward on this. As mentioned in https://github.com/vitest-dev/vitest/issues/3530#issuecomment-1620440424, I think we should revert this behavior and consider it as a regression. And if we are going to do it, I rather not document it as it will confuse users when are we will going to change it back soon.

antfu:
Close this as we reverted this behaviour in the recent minor: https://github.com/vitest-dev/vitest/pull/3729

Thanks for bringing this up

PR #3749, opened 2023-07-09
fix(node): guard error info

### LINKED ISSUE #3746: Error handler crashes when not Error is thrown and caught
### Describe the bug

While it’s the best practice to only throw instances of `Error`, this does not happen always.

The error logger does not anticipate that to the full extent and crashes when catches a not-instance-of-Error value and tries to log it: 

```
TypeError: Cannot read properties of undefined (reading 'length')
 ❯ printErrorMessage node_modules/vitest/dist/vendor-cli-api.cb31e1db.js:12433:21
 ❯ printError node_modules/vitest/dist/vendor-cli-api.cb31e1db.js:12320:3
 ❯ Logger.printError node_modules/vitest/dist/vendor-cli-api.cb31e1db.js:12542:12
 ❯ DefaultReporter.printTaskErrors node_modules/vitest/dist/vendor-cli-api.cb31e1db.js:7971:29
 ❯ DefaultReporter.printErrorsSummary node_modules/vitest/dist/vendor-cli-api.cb31e1db.js:7907:18
 ❯ DefaultReporter.reportSummary node_modules/vitest/dist/vendor-cli-api.cb31e1db.js:7834:16
 ❯ DefaultReporter.onFinished node_modules/vitest/dist/vendor-cli-api.cb31e1db.js:7689:16
 ❯ DefaultReporter.onFinished node_modules/vitest/dist/vendor-cli-api.cb31e1db.js:9052:17
 ❯ Vitest.report node_modules/vitest/dist/vendor-cli-api.cb31e1db.js:14102:5
```

[The line in question](https://github.com/vitest-dev/vitest/blob/8dd5ea5ba10c2a99568ff18f6efff356824cb2cf/packages/vitest/src/node/error.ts#L194) tries to do `if (error.message.length > 5000)` but I believe a proper fix should be applied higher in the stack when catching an unknown type value from the runner.

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-fanzfc?file=test%2Fsuite.test.ts

### System Info

```shell
System:
    OS: Linux 6.2 Ubuntu 23.04 23.04 (Lunar Lobster)
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
    Memory: 16.07 GB / 31.13 GB
    Container: Yes
    Shell: 3.6.0 - /usr/bin/fish
  Binaries:
    Node: 20.4.0 - /usr/bin/node
    Yarn: 1.22.19 - /usr/bin/yarn
    npm: 9.7.2 - /usr/bin/npm
    pnpm: 8.6.6 - /usr/bin/pnpm
  Browsers:
    Chrome: 114.0.5735.198
  npmPackages:
    vitest: ^0.32.4 => 0.32.4
```

### Used Package Manager

yarn

### Validations

btea:
fix #3746

### DIFF
--- a/packages/vitest/src/node/error.ts
+++ b/packages/vitest/src/node/error.ts
@@ -191,7 +191,7 @@ export function displayDiff(diff: string | null, console: Console) {
 
 function printErrorMessage(error: ErrorWithDiff, logger: Logger) {
   const errorName = error.name || error.nameStr || 'Unknown Error'
-  if (!(error instanceof Error)) {
+  if (!error.message) {
     logger.error(error)
     return
   }
--- a/packages/vitest/src/node/error.ts
+++ b/packages/vitest/src/node/error.ts
@@ -191,6 +191,10 @@ export function displayDiff(diff: string | null, console: Console) {
 
 function printErrorMessage(error: ErrorWithDiff, logger: Logger) {
   const errorName = error.name || error.nameStr || 'Unknown Error'
+  if (!(error instanceof Error)) {
+    logger.error(error)
+    return
+  }
   if (error.message.length > 5000) {
     // Protect against infinite stack trace in picocolors
     logger.error(`${c.red(c.bold(errorName))}: ${error.message}`)


sheremet-va:
error is not instance of Error most of the time, it is transferred as a simple object

sheremet-va (approved):

joseferben:
Until this lands in a release, if you are using Remix you can catch the Remix responses and simply return the response value in tests:

```typescript
async function catchRemixResponse(
  handler: () => Promise<Response>
): Promise<Response> {
  try {
    const resp = await handler();
    return resp;
  } catch (e: any) {
    if (e instanceof Response) {
      return e;
    } else {
      throw e;
    }
  }
}

    const response = await catchRemixResponse(() =>
      action({ request, context: {}, params: {} })
    );
```

PR #3758, opened 2023-07-11
feat(browser): add support for custom browser capabilities

haywirez:
The aim of this PR is to add support for passing [custom browser capability arguments for WebdriverIO](https://webdriver.io/docs/capabilities#custom-capabilities). This was done rather quickly so improvements and suggestions are welcome, I only saw tests for playwiright.

### DIFF
--- a/docs/guide/browser.md
+++ b/docs/guide/browser.md
@@ -108,7 +108,50 @@ npx vitest --browser.name=chrome --browser.headless
 
 In this case, Vitest will run in headless mode using the Chrome browser.
 
+## Custom Provider Options
+
+You may pass provider-specific options, such as [WebdriverIO custom capabilities](https://webdriver.io/docs/capabilities#custom-capabilities) like:
+
+```ts
+export default defineConfig({
+  test: {
+    browser: {
+      enabled: true,
+      headless: true,
+      name: 'webdriverio',
+      options: {
+        webdriverio: {
+          'goog:chromeOptions': {
+            args: ['disable-gpu'],
+          },
+        },
+      },
+    },
+  },
+})
+```
+
+or in case of [Playwright-specific options](https://playwright.dev/docs/api/class-browsertype#browser-type-launch):
+
+```ts
+export default defineConfig({
+  test: {
+    browser: {
+      enabled: true,
+      headless: true,
+      name: 'playwright',
+      options: {
+        playwright: {
+          args: ['disable-gpu'], // chromium
+        },
+      },
+    },
+  },
+})
+```
+
 ## Limitations
+
 ### Thread Blocking Dialogs
 
 When using Vitest Browser, it's important to note that thread blocking dialogs like `alert` or `confirm` cannot be used natively. This is because they block the web page, which means Vitest cannot continue communicating with the page, causing the execution to hang.
--- a/packages/vitest/src/node/browser/playwright.ts
+++ b/packages/vitest/src/node/browser/playwright.ts
@@ -1,5 +1,5 @@
 import type { Page } from 'playwright'
-import type { BrowserProvider, BrowserProviderOptions } from '../../types/browser'
+import type { BrowserProvider, BrowserProviderOptions, ProviderSpecificOptions } from '../../types/browser'
 import { ensurePackageInstalled } from '../pkg'
 import type { WorkspaceProject } from '../workspace'
 import type { Awaitable } from '../../types'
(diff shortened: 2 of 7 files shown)


docs/guide/browser.md
```diff
@@ -108,6 +108,26 @@ npx vitest --browser.name=chrome --browser.headless
...
+If using [WebdriverIO](https://webdriver.io/) (default), you may pass [custom capabilities](https://webdriver.io/docs/capabilities#custom-capabilities) as:
+
+```ts
+export default defineConfig({
+  test: {
+    browser: {
+      enabled: true,
+      headless: true,
+    },
+    capabilities: {
```
sheremet-va, line 122: I think it's better to add a general option for providers:

```ts
{
  providerOptions: {
    webdriverio: WebdriverIOOptions
    playwright: PlaywrightOptions
    custom: unknown
  }
}
```

sheremet-va (changes requested):

haywirez:
Pushed changed in the spirit of the requested modifications 👍

docs/guide/browser.md
```diff
@@ -108,7 +108,50 @@ npx vitest --browser.name=chrome --browser.headless
...
+
+You may pass provider-specific options, such as [WebdriverIO custom capabilities](https://webdriver.io/docs/capabilities#custom-capabilities) like:
+
+```ts
+export default defineConfig({
+  test: {
+    browser: {
+      enabled: true,
+      headless: true,
+      name: 'webrdiverio'
```
dammy001, line 121:
```suggestion
      name: 'webdiverio'
```

docs/guide/browser.md
```diff
@@ -108,7 +108,50 @@ npx vitest --browser.name=chrome --browser.headless
...
+    },
+    providerOptions: {
```
sheremet-va, line 123: Why is this top level option? It should be in `browser`

docs/guide/browser.md
```diff
@@ -108,7 +108,50 @@ npx vitest --browser.name=chrome --browser.headless
...
+      webdriverio: {
+        'goog:chromeOptions': {
```
sheremet-va, line 125: Webdriver should provide all option to `remote`, not just capabilities

packages/vitest/src/node/browser/playwright.ts
```diff
@@ -40,7 +43,7 @@ export class PlaywrightBrowserProvider implements BrowserProvider {
 
     const playwright = await import('playwright')
 
-    const playwrightInstance = await playwright[this.browser].launch({ headless: options.headless })
+    const playwrightInstance = await playwright[this.browser].launch({ headless: options.headless, ...this.providerOptions })
```
sheremet-va, line 46: We should override user `headless` option

packages/vitest/src/node/browser/webdriver.ts
```diff
@@ -59,6 +63,7 @@ export class WebdriverBrowserProvider implements BrowserProvider {
       capabilities: {
         'browserName': this.browser,
         'wdio:devtoolsOptions': { headless: options.headless },
+        ...this.providerOptions,
```
sheremet-va, line 66: We should override browserName and `wdio:devtoolsOptions.headless`

sheremet-va (changes requested):
I don't see any changes to config types? And how do you pass down these new provider options?

packages/vitest/src/node/browser/webdriver.ts
haywirez: Do you mean should or shouldn't? Right now it will overwrite all former if present in `this.providerOptions`

packages/vitest/src/types/browser.ts
```diff
@@ -2,14 +2,23 @@ import type { Awaitable } from '@vitest/utils'
 import type { WorkspaceProject } from '../node/workspace'
 import type { ApiConfig } from './config'
 
+export interface ProviderSpecificOptions {
+  webdriverio?: WebDriver.Capabilities | WebDriver.DesiredCapabilities
+  playwright?: any
```
haywirez, line 7: Couldn't find a good way here to import playwright types only

haywirez: What would be your preferred way? Would removing added types from package.json and using `unknown` for both be acceptable?

sheremet-va: Should be acceptable for now

sheremet-va (approved):

packages/vitest/src/node/browser/webdriver.ts
```diff
@@ -57,8 +60,9 @@ export class WebdriverBrowserProvider implements BrowserProvider {
     this.cachedBrowser = await remote({
       logLevel: 'error',
       capabilities: {
```
sheremet-va, line 62: Why are passing down only capabilities? Shouldn't it be `remote` options?

sheremet-va (changes requested):

haywirez:
Will take another look in the coming two weeks

PR #3769, opened 2023-07-13
docs: Clarify external option

danez:
I was using globs in the external option, which I noticed had no effect. After switching to regexes it started working.

I looked at [this code ](https://github.com/vitest-dev/vitest/blob/main/packages/vite-node/src/externalize.ts#L141-L150) and tried to explain what the option is doing and make it more clear that strings are not simply checked against the id, but rather the `moduleDirectories`.


dammy001:
@danez This has been mentioned here #3203

sheremet-va (approved):

PR #3771, opened 2023-07-14
feat: support `describe.sequential`

### LINKED ISSUE #3731: Add API `describe.sequential` to be able to override `--sequence.concurrent`

Related to #3573 (thank you for the feature).
There's one case users would probably miss:
- when setting `--sequence.concurrent`, there's no way to override this behavior for an individual `describe` suite
  - for example, we have `describe.concurrent` now to set the sequence.concurrent behavior for an individual describe, but if a user would set `.concurrent` by default, there's no way to add a sequential override for an individual `describe`

### Suggested solution

Please consider to add API: `describe.sequential(...)`, so a user could override `--sequence.concurrent` for a specific describe.

### Alternative

_No response_

### Additional context

_No response_

### Validations

fenghan34:
Closes #3731

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -639,6 +639,24 @@ When running concurrent tests, Snapshots and Assertions must use `expect` from t
 You cannot use this syntax, when using Vitest as [type checker](/guide/testing-types).
 :::
 
+### describe.sequential
+
+- **Type:** `(name: string | Function, fn: TestFunction, options?: number | TestOptions) => void`
+
+  `describe.sequential` in a suite marks every tests as sequential. This is useful if you want to run tests in sequential within `describe.concurrent` or with the `--sequence.concurrent` command option.
+
+  ```ts
+  describe.concurrent('suite', () => {
+    test('concurrent test 1', async () => { /* ... */ })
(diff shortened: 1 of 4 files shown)


sheremet-va (changes requested):
Documentation is missing

docs/api/index.md
```diff
@@ -639,6 +639,24 @@ When running concurrent tests, Snapshots and Assertions must use `expect` from t
 You cannot use this syntax, when using Vitest as [type checker](/guide/testing-types).
 :::
 
+### describe.sequential
+
+- **Type:** `(name: string | Function, fn: TestFunction, options?: number | TestOptions) => void`
+
+  `describe.sequential` in a suite marks every tests as sequential. This is useful if you want to run tests in sequential within `describe.concurrent` or with the `--sequence.concurrent` command option.
```
dammy001, line 646:
```suggestion
  `describe.sequential` in a suite marks every test as sequential. This is useful if you want to run tests in sequential within `describe.concurrent` or with the `--sequence.concurrent` command option.
```

fenghan34: Thank you for reviewing!

dammy001 (approved):

PR #3773, opened 2023-07-15
fix(core): test change should trigger failed test rerun

### LINKED ISSUE #3687: Watch mode "f" command does not rerun failed tests
### Describe the bug

When in watch mode and pressing "f" to rerun failed tests, "rerun failed" is printed and all tests are rerun. Not sure how it worked before, but I would also expect that a file change keeps reruning only the failed tests as long as I have the "f" command selected like in Jest.

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-jpe9pa?file=test%2Fsuite.test.ts&initialPath=__vitest__/

1. wait for the tests to run
2. in watch console press "f"
3. notice "rerun failed" is quickly flashed before test results are printed
image
4. notice all tests have ran
5. notice triggering a file change reruns all the tests, disregarding the "f" command

### System Info

```shell
The stackblitz env is this 

System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 16.20.0 - /usr/local/bin/node
    pnpm: 8.6.3 - /usr/local/bin/pnpm
  npmPackages:
    @vitest/ui: latest => 0.32.2 
    vite: latest => 4.3.9 
    vitest: latest => 0.32.2 
```
But I have the same problem locally on

```
  System:
    OS: macOS 13.3.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 3.76 GB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 16.10.0 - ~/.nvm/versions/node/v16.10.0/bin/node
    npm: 8.18.0 - ~/.nvm/versions/node/v16.10.0/bin/npm
  Browsers:
    Brave Browser: 113.1.51.110
    Chrome: 109.0.5414.87
    Edge: 114.0.1823.51
    Firefox: 112.0.1
    Safari: 16.4
  npmPackages:
    @vitejs/plugin-basic-ssl: ^1.0.1 => 1.0.1
    @vitest/coverage-c8: ^0.32.2 => 0.32.2
    @vitest/ui: ^0.32.2 => 0.32.2
    vite: ^4.1.1 => 4.1.1
    vitest: ^0.32.2 => 0.32.2
```
```


### Used Package Manager

npm

### Validations

### PR DESCRIPTION
**Pull request by btea**
fix #3687

### CODE REVIEW

**Review by AriPerkkio — CHANGES_REQUESTED** [review]
- Add test case to `test/watch/test/stdin.test.ts`
- The implementation doesn't seem to consider the filter enabled by pressing `f`? This PR should only affect that use case.

**Comment by btea** [comment]
> * The implementation doesn't seem to consider the filter enabled by pressing `f`? This PR should only affect that use case.

Maybe I misunderstood you guys? I thought that as long as the test content is modified, if there is a failed test case, it will automatically rerun.

File: test/watch/test/stdin.test.ts [hunk 1]
```diff
@@ -27,6 +27,40 @@ test('rerun current pattern tests', async () => {
   await vitest.waitForStdout('1 passed')
 })
 
+test('rerun failed tests', async () => {
```
AriPerkkio, line 30: We'll need a test case that fails in `main`, and is fixed by these changes. This test case does not fail on `main` branch. 

Maybe something like below?

```ts
test('rerun failed tests', async () => {
  const testPath = 'fixtures/error.test.ts'
  const failAssertion = 'expect(1).toBe(2)'
  const testCase = `// Dynamic test case
  import { test, expect } from 'vitest'

  // Failure re-run should run this
  test('failing test', () => {
    ${failAssertion}
  })

  // Failure re-run should NOT run this
  test('passing test', () => {
    expect(1).toBe(1)
  })
  `

  writeFileSync(testPath, testCase, 'utf8')
  cleanups.push(() => rmSync(testPath))

  const vitest = await runVitestCli(...cliArgs)
  await vitest.waitForStdout('Test Files  1 failed | 2 passed (3)')
  await vitest.waitForStdout('Tests  1 failed | 3 passed (4)')

  // Re-run failed tests only
  vitest.write('f')

  // Only a single test should have run
  await vitest.waitForStdout('Test Files  1 failed (1)')

  // Only a single test case should have run
  await vitest.waitForStdout('Tests  1 failed (1)')

  // Fix failing test case
  writeFileSync(testPath, testCase.replace(failAssertion, ''), 'utf8')

  // Only the previously failed test file should run
  await vitest.waitForStdout('Test Files  1 pass (1)')

  // Only the previously failed test case should run
  await vitest.waitForStdout('Tests  1 pass (1)')
})
```

packages/vitest/src/node/stdin.ts
```diff
@@ -76,14 +76,20 @@ export function registerConsoleShortcuts(ctx: Vitest) {
     if (name === 'u')
       return ctx.updateSnapshot()
     // rerun all tests
-    if (name === 'a' || name === 'return')
+    if (name === 'a' || name === 'return') {
+      ctx.isFailedModel = false
```
AriPerkkio, line 80: I don't think we should set properties of `ctx` outside it like this. Could this filter be toggled on/off inside the `core.ts` instead?

But your approach is right in the way that we indeed need to save the failure filter in state.

AriPerkkio (changes requested):
> Maybe I misunderstood you guys? I thought that as long as the test content is modified, if there is a failed test case, it will automatically rerun.
> 
> I seem to understand that only after pressing `f`, if the content of the test file is modified, the failed test will be automatically run, is it?

The issue contains two bugs:

- When user has activated "Run only failed test"-filter in watch mode by pressing `f`, this filter does not preserve if files are modified
- The "Run only failed test"-filter runs all **test cases** inside **test files**. It should run only failed **test cases**. Now it's only filtering the files.

btea:
> The issue contains two bugs:
> 
> * When user has activated "Run only failed test"-filter in watch mode by pressing `f`, this filter does not preserve if files are modified
> * The "Run only failed test"-filter runs all **test cases** inside **test files**. It should run only failed **test cases**. Now it's only filtering the files.

If I didn't understand the error, it means that if a test file has test cases that fail, after pressing `f`, only the failed suite in the test file is re-run instead of the entire test file and all cases are re-run? (If so, filter test cases in a test file, I don't know how to do it, can you give me a hint? Thanks.)

And if the content of the file is modified, the failed test will not be filtered, and the original operation logic can be followed.

AriPerkkio:
> (If so, filter test cases in a test file, I don't know how to do it, can you give me a hint? Thanks.)

I'm not 100% sure, but the changes should be made somewhere around `packages/runner` I think.

btea:
Sorry, I can't seem to find any relevant instructions in the official documentation. Maybe I'm missing something. 🤔

@sheremet-va  Can you give me a little hint? Thanks!

sheremet-va:
> @sheremet-va Can you give me a little hint? Thanks!

I'm not sure, but I think you can set `configOverride.testNamePattern` to filter tests (this is what vscode does) - maybe we should improve this somehow internally 🤷🏻

sheremet-va:
Seems like this was fixed at one point

PR #3788, opened 2023-07-20
fix: correctly render ansi diff

### LINKED ISSUE #3675: It looks like this change may have caused the output in `@vitest/ui` to print ANSI escape codes.
I commented on #3582 but realized I should probably open a new issue:

It looks like something in #3582 may have caused the output in `@vitest/ui` to print ANSI escape codes.

image

I'm not sure what the root cause is, but I'm sure it started after I upgraded vitest this week.

_Originally posted by @wycats in https://github.com/vitest-dev/vitest/issues/3582#issuecomment-1605079741_

so1ve:
close #3675

### DIFF
--- a/packages/ui/client/components/views/ViewReportError.vue
+++ b/packages/ui/client/components/views/ViewReportError.vue
@@ -14,13 +14,13 @@ function relative(p: string) {
   return p
 }
 
+const filter = computed(() => createAnsiToHtmlFilter(isDark.value))
+
 const isDiffShowable = computed(() => {
   return props.error?.expected && props.error?.actual
 })
 
-function diff() {
-  return props.error.diff
-}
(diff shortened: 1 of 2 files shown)


so1ve:
Force pushed for a wrong refactor 😅

antfu:
The lock changes do not seem to needed?

so1ve:
> The lock changes do not seem to needed?

I'm not sure about it. If I don't push it, the ci will fail.

sheremet-va:
CI still fails without it it seems. Please, revert all package-lock changes you made

so1ve:
@sheremet-va Thanks! I've just reverted this change.

so1ve:
I'll create a new branch to avoid unnecessary changes

PR #3804, opened 2023-07-24
fix: port the latest `defineConfig` type from vite

sodatea:
Ported from https://github.com/vitejs/vite/pull/13792

Fully fixes issues like https://github.com/vuejs/create-vue/issues/317 and https://github.com/vuejs/create-vue/issues/313

For easy reproduction, please refer to https://github.com/sodatea/viteste-define-config-bug-repro

I didn't include a test in this PR because:

1. The reproduction is a bit complicated, it only fails on the very latest Vite versions, but updating Vite would mess up the lockfile.
2. By not updating the dependencies, we can make sure that this fix is compatible with older versions of Vite, too.


antfu (approved):

vinassefranche:
Hi! I actually did another small update of the types in vite (see [this P.R.](https://github.com/vitejs/vite/pull/13958/files))
Is it possible to integrate this change in this P.R.?

PR #3806, opened 2023-07-24
fix: multi-suite browser coverage for istanbul (fix #3514)

### LINKED ISSUE #3514: Browser mode: Istanbul coverage report incomplete
### Describe the bug

The following test file causes trouble with vitest in browser mode (chrome). Notice the nested `describe` calls:

```js
import { describe, it, expect } from 'vitest';
import { foo, bar, baz } from '../src/stuff';

describe('stuff 1', () => {
  it('should baz', () => {
    expect(baz()).toBe('doo');
  });

  describe('stuff 2', () => {
    describe('stuff 3', () => {
      it('should foo', () => {
        expect(foo()).toBe('bar');
      });
    });
    it('should bar', () => {
      expect(bar()).toBe('foo');
    });
  });
});

```

What seems to happen is that all three tests are run (and pass), but only the `should baz` and `should foo` count towards the coverage report. In the repro repo, this means line coverage is reported as 66%.

The issue only occurs when the tests are browser based. With the jsdom environment, things work as expected and the coverage is reported as 100%.

### Reproduction

Reproduction repo is [here](https://github.com/lsim/vitest-istanbul-issue-repro/tree/main)

### System Info

```shell
System:
    OS: macOS 13.4
    CPU: (10) arm64 Apple M1 Max
    Memory: 1.84 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.16.0 - ~/.nvm/versions/node/v18.16.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v18.16.0/bin/yarn
    npm: 9.5.1 - ~/.nvm/versions/node/v18.16.0/bin/npm
  Browsers:
    Chrome: 113.0.5672.126
    Firefox: 107.0.1
    Safari: 16.5
  npmPackages:
    @vitest/browser: ^0.31.1 => 0.31.4
    @vitest/coverage-istanbul: ^0.31.4 => 0.31.4
    @vitest/ui: ^0.31.1 => 0.31.4
    vite: ^4.3.2 => 4.3.9
    vitest: ^0.31.1 => 0.31.4
```

### Used Package Manager

yarn

### Validations

elby22:
## The issue
Browser coverage breaks with istanbul for multiple suites in some situations. Comparing the example tests side-by-side with the default runner (no fix applied):

Screenshot from 2023-07-24 15-28-00

However, when tests were run without any describe blocks, browser test coverage would be properly represented.

## The Fix  
`packages/browser/src/client/runner.ts` now takes coverage `onAfterRun` instead of `onAfterRunSuite` in a manner similar to `packages/vitest/src/runtime/entry.ts`.

## Notes
I have a demo test package in `test/istanbul-browser-coverage`. I could use some guidance on the proper way to test this, if necessary. The code change did not fail the existing tests.


elby22:
@AriPerkkio I've made the suggested changes. Thanks for the feedback!

pnpm-lock.yaml
AriPerkkio, line 1: Revert all changes of `pnpm-lock.yaml`. It should not be included in this PR.

elby22: @AriPerkkio done

AriPerkkio (approved):
Looks good, thanks!

sheremet-va:
Can you resolve conflicts?

elby22:
@sheremet-va done

Issue #3691, opened 2023-06-28
Support customized junit report format 


InfiniteXyy:
### Clear and concise description of the problem

Currently, when using `reporters: ["junit"]`, only `testsuites` is configurable, could we also make `classNameTemplate` and`titleTemplate` comfigurable, so that I can use custom seperator between classname and title, or remove classname totally.

https://github.com/vitest-dev/vitest/blob/f91da4849ffafd52b0969c21927873a87c47222a/packages/vitest/src/node/reporters/junit.ts#L167-L169

### Suggested solution

In `jest-junit` there are some configurable env values, like `JEST_JUNIT_CLASSNAME`

and we also have `VITEST_JUNIT_SUITE_NAME` in vitest, we could add more env variables to help config the junit output

### Additional context

The configratuion for `jest-junit` documentation:
https://github.com/jest-community/jest-junit#configuration

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

AriPerkkio:
Instead of supporting environment variables we should open an API for providing reporter options. The coverage reporters already support this kind of API:

https://github.com/vitest-dev/vitest/blob/726a57e71c94b8b1a281e2079e3f2ce7d712d65a/docs/config/index.md?plain=1#L786-L800

This is something that 3rd party custom reporters could also utilize.

InfiniteXyy:
I created a simple PR to solve my own usecase (to remove the filepath in className), would it be OK to merge that PR first?
https://github.com/vitest-dev/vitest/pull/3808

PR #3808, opened 2023-07-25
feat: support config junit reporter className with env

InfiniteXyy:
Add some partial config for https://github.com/vitest-dev/vitest/issues/3691

It would be helpful if some want don't want the `filePath` to be in the `className ` in Junit reporter

### DIFF
--- a/packages/vitest/src/node/reporters/junit.ts
+++ b/packages/vitest/src/node/reporters/junit.ts
@@ -168,7 +168,7 @@ export class JUnitReporter implements Reporter {
   async writeTasks(tasks: Task[], filename: string): Promise<void> {
     for (const task of tasks) {
       await this.writeElement('testcase', {
-        classname: filename,
+        classname: process.env.VITEST_JUNIT_CLASSNAME ?? filename,
         name: task.name,
         time: getDuration(task),
       }, async () => {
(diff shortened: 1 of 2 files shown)


packages/vitest/src/node/reporters/junit.ts
```diff
@@ -168,7 +168,7 @@ export class JUnitReporter implements Reporter {
   async writeTasks(tasks: Task[], filename: string): Promise<void> {
     for (const task of tasks) {
       await this.writeElement('testcase', {
-        classname: filename,
+        classname: process.env.VITEST_JUNIT_CLASSNAME ?? filename,
```
AriPerkkio, line 171: Update documentation too:

https://github.com/vitest-dev/vitest/blob/b7ce1959bd469b83c3b7035579b7583386503489/docs/config/index.md?plain=1#L486

AriPerkkio (changes requested):
I guess this change is fine for now. As long term solution we should allow configuring reporter specific options: https://github.com/vitest-dev/vitest/issues/3691#issuecomment-1611162791

packages/vitest/src/node/reporters/junit.ts
InfiniteXyy: Added
```md
JUnit XML reporter (you can configure `testsuites` tag name with `VITEST_JUNIT_SUITE_NAME` environmental variable, and `classname` tag property with `VITEST_JUNIT_CLASSNAME`)
```

AriPerkkio (approved):

sheremet-va (approved):

PR #3810, opened 2023-07-25
fix: correctly render ansi diff

so1ve:
close https://github.com/vitest-dev/vitest/issues/3675


sheremet-va (changes requested):
Can you add a test for this? We have cypress tests in UI package.

sheremet-va:
> I'd like to add a test for this but... I'm not familiar with cypress, and I'm not sure where to add these tests. Any suggestions?

You can look at other tests. To see the script to run it, you can look into package.json

nasso:
is anything blocking this?

so1ve:
@nasso Yes, missing tests. I'm unfamiliar to e2e tests, so…

PR #3832, opened 2023-07-28
docs: clarify how config file works

sheremet-va:
Closes #3799
Closes #2989

### DIFF
--- a/docs/guide/index.md
+++ b/docs/guide/index.md
@@ -46,7 +46,7 @@ One of the main advantages of Vitest is its unified configuration with Vite. If 
 - Pass `--config` option to CLI, e.g. `vitest --config ./path/to/vitest.config.ts`
 - Use `process.env.VITEST` or `mode` property on `defineConfig` (will be set to `test` if not overridden) to conditionally apply different configuration in `vite.config.ts`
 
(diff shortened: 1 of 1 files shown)


docs/guide/index.md
```diff
@@ -46,7 +46,7 @@ One of the main advantages of Vitest is its unified configuration with Vite. If
 - Pass `--config` option to CLI, e.g. `vitest --config ./path/to/vitest.config.ts`
 - Use `process.env.VITEST` or `mode` property on `defineConfig` (will be set to `test` if not overridden) to conditionally apply different configuration in `vite.config.ts`
 
-To configure `vitest` itself, add `test` property in your Vite config. You'll also need to add a reference to Vitest types using a [triple slash command](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html#-reference-types-) at the top of your config file, if you are importing `defineConfig` from `vite` itself.
+If you are not using Vite as your build tool, you can configure Vitest using `test` property in your config file:
```
dammy001, line 49:
```suggestion
If you are not using Vite as your build tool, you can configure Vitest using the `test` property in your config file:
```

dammy001: @sheremet-va I think we should also specify different Vitest extension and not just `vitest.config.ts` only in line 45. Maybe `vitest.config.ts|js|mjs`

Issue #3866, opened 2023-08-02
Failed to load url error when using sveltekit and jsdom environment (0.34.1)


EugeneDraitsev:
### Describe the bug

I have "Failed to load url" error when using sveltekit and jsdom environment for `setupFiles` file. It works fine with vitest 0.33, this error appear only in vitest 0.34.1

### Reproduction

1. Create empty sveltekit project with vitest `npm create svelte@latest my-app` or clone my repo with all dependencies and files: https://github.com/EugeneDraitsev/vitest-setup-files-bug
2. Add jsdom to dependencies, add empty `setupTest.ts` and use following `vite.config.ts`
```ts
export default defineConfig({
	plugins: [sveltekit()],
	test: {
		include: ['src/**/*.{test,spec}.{js,ts}'],
		environment: 'jsdom',
		setupFiles: './setupTest.ts'
	}
});
```
3. Run `yarn test` and see "Failed to load url" error

If you comment `environment: 'jsdom',` or downgrade vitest to `0.33.0`, then it will work fine

### System Info

```shell
System:
    OS: Windows 10 10.0.22621
    CPU: (24) x64 13th Gen Intel(R) Core(TM) i7-13700KF
    Memory: 13.58 GB / 31.82 GB
  Binaries:
    Node: 18.15.0 - ~\scoop\apps\nvm\current\nodejs\nodejs\node.EXE
    Yarn: 1.22.19 - ~\scoop\apps\yarn\current\bin\yarn.CMD
    npm: 9.5.0 - ~\scoop\apps\nvm\current\nodejs\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22621.1992.0), Chromium (115.0.1901.188)
    Internet Explorer: 11.0.22621.1
```

### Used Package Manager

yarn

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
[...]

PR #3868, opened 2023-08-02
fix: disable strict access when running inside Node.js

sheremet-va:
### Description

Related #3866

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/plugins/index.ts
+++ b/packages/vitest/src/node/plugins/index.ts
@@ -87,6 +87,11 @@ export async function VitestPlugin(options: UserConfig = {}, ctx = new Vitest('t
             open,
             hmr: false,
             preTransformRequests: false,
+            fs: {
+              // disable strict access since we are working inside Node.js,
+              // and outside of root imports are allowed here
+              strict: false,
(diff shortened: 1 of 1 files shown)


sheremet-va:
@patak-dev can you help me understand if this is a good idea for Vitest? Or should we try to add related files to `fs.allow`?

patak-dev:
I think it depends on whether the Vite server can be accessed from out of the Vitest project. We need this restriction mainly when the server is exposed to the network, if that isn't possible in Vitest then disabling strict sounds ok to me. Probably a good idea to double-check getting @antfu's opinion too.

sheremet-va:
> I think it depends on whether the Vite server can be accessed from out of the Vitest project. We need this restriction mainly when the server is exposed to the network, if that isn't possible in Vitest then disabling strict sounds ok to me. Probably a good idea to double-check getting @antfu's opinion too.

The thing is that when we use `ssr` flag, then it bypasses this check. So I'm not sure if either way is secure enough for Vitest 🤔

sheremet-va:
This was fixed in #3956

Issue #3867, opened 2023-08-02
allowExternal option for v8 coverage provider


vojvodics:
### Clear and concise description of the problem

Since the @vitest/coverage-c8 is removed in favor of @vitest/coverage-v8 (https://github.com/vitest-dev/vitest/pull/3339) and coverage-v8 doesn't provide the allowExternal option this will break coverage reports when upgrading to the latest vitest version (and switching providers).

### Suggested solution

Implement the allowExternal option (as suggested in the above-mentioned PR). I'd be happy to provide the implementation (LMK if you have any inputs/suggestions on how to do this)

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

AriPerkkio:
This should also be implemented to `@vitest/coverage-istanbul`. They both use the same `test-exclude` dependency which I think provides this in `c8`: https://github.com/bcoe/c8/blob/a13584d5be5259ebb6a00455d352c3e8b16006de/lib/report.js#L48-L52.

>  I'd be happy to provide the implementation

PR is welcome! Here is one PR for reference to show which files, tests and documentation needs changes when adding new coverage options: https://github.com/vitest-dev/vitest/pull/3453.
Let me know if you need any help.

vojvodics:
@AriPerkkio I drafted a PR for this, can you check it out? https://github.com/vitest-dev/vitest/pull/3894 

I actually would need help with writing good tests for this feature. I added some tests but I don't think they do what they should do. Please check out the comments in the PR 🙏

PR #3894, opened 2023-08-05
feat(coverage): add `allowExternal` option

vojvodics:
### Description

Implements allowExternal option removed in v0.34.0

Resolves https://github.com/vitest-dev/vitest/issues/3867


test/coverage-test/vitest.config.ts
```diff
@@ -43,7 +43,8 @@ export default defineConfig({
     coverage: {
       provider: provider as any,
       customProviderModule: provider === 'custom' ? 'custom-provider' : undefined,
-      include: ['src/**'],
+      include: ['**/src/**', '**/external/**.external.ts'],
+      allowExternal: true,
```
vojvodics, line 47: Added this for simplicity of adding new tests, but I don't think this is actually testing the allowExternal good enough. I need help with how I would construct these tests.

I have verified that the allowExternal option works by setting include to `['**/.ts', '**/.vue']` and moving the external folder outside the `test/coverage-test` folder:
- if callExternal is calling the function from the external folder:
Screenshot 2023-08-05 at 15 58 21

- if callExternal is not calling the function from the external folder:
Screenshot 2023-08-05 at 15 59 04


Any ideas how I could write a test like this?

packages/vitest/src/node/config.ts
```diff
@@ -273,7 +273,7 @@ export function resolveConfig(
         ?? resolve(resolved.root, file),
     ),
   )
-  resolved.coverage.exclude.push(...resolved.setupFiles.map(file => relative(resolved.root, file)))
+  resolved.coverage.exclude.push(...resolved.setupFiles.map(file => `${resolved.coverage.allowExternal ? '**/' : ''}${relative(resolved.root, file)}`))
```
vojvodics, line 276: this was added because the test for `src/another-setup.ts` was failing. Because test-exclude will try to resolve absolute path if allowExternal is set to true I added `**/` prefix. 
I don't think this is a good solution as it may exclude some files we didn't wish to exclude (e.g. if the setup file is called `src/index.ts`)

test/coverage-test/vitest.config.ts
AriPerkkio: Currently the project `root` is `/Users/x/vitest/test/coverage-test`. The `external/` directory is not outside of this directory. Let's remove following files:

- `vitest/test/coverage-test/src/callExternal.ts`
- `vitest/test/coverage-test/external/index.ts`
- `vitest/test/coverage-test/external/file.external.ts`

And add a generic fixture file `vitest/test/test-utils/fixtures/math.ts`:

```ts
export function sum(a: number, b: number) {
  return a + b
}

export function multiply(a: number, b: number) {
  return a * b
}
```

Then in `vitest/test/coverage-test/test/coverage.test.ts` test it as:

```ts
import * as ExternalMath from '../../test-utils/fixtures/math'

test('calling files outside project root', () => {
  expect(ExternalMath.sum(2, 3)).toBe(5)
})
```

And configure `vitest/test/coverage-test/vitest.config.ts` as:

```ts
  include: ['**/src/**', '**/test-utils/fixtures/**'],
  allowExternal: true,
```

This setup will test that `allowExternal: true` works. Our current testing setup doesn't really allow testing different values for different configuration options. We'll need to manually verify that 
`fixtures/math.ts` is not in report when changing value of `allowExternal`.

packages/vitest/src/node/config.ts
AriPerkkio: This makes sense but is indeed not ideal. Let's leave this open for now and proceed with other changes.

test/coverage-test/vitest.config.ts
AriPerkkio: You'll also need to change multiple paths in `coverage-report-tests/*` like:

```diff
- const coveragePath = resolve('./coverage/src')
+ const coveragePath = resolve('./coverage/coverage-test/src')
```

I'm not really sure if we should use `allowExternal: true` for all tests as it changes the setup so much.

AriPerkkio (changes requested):
I'm not sure if setting `allowExternal: true` for all coverage tests is good approach. But let's do that for now and I'll see if better way to test this is possible.

test/coverage-test/coverage-report-tests/generic.report.test.ts
```diff
@@ -58,10 +58,10 @@ test('files should not contain a setup file', () => {
   const coveragePath = resolve('./coverage')
```
AriPerkkio, line 58:
```suggestion
  const coveragePath = resolve('./coverage/coverage-test')
```

test/coverage-test/coverage-report-tests/generic.report.test.ts
```diff
@@ -58,10 +58,10 @@ test('files should not contain a setup file', () => {
...
   const files = fs.readdirSync(coveragePath)
 
-  expect(files).not.toContain('coverage-test')
+  // expect(files).not.toContain('coverage-test')
```
AriPerkkio, line 61:
```suggestion
```

test/coverage-test/coverage-report-tests/generic.report.test.ts
```diff
@@ -104,7 +104,7 @@ test('coverage provider does not conflict with built-in reporter\'s outputFile',
 
 test('virtual files should be excluded', () => {
   const files = fs.readdirSync(resolve('./coverage'))
```
AriPerkkio, line 106:
```suggestion
  const files = fs.readdirSync(resolve('./coverage/coverage-test'))
```

vojvodics:
@AriPerkkio I had the idea to run the tests via cli argument (e.g. `"test:v8": "node ./testing.mjs --provider v8 && node ./testing.mjs --provider v8  --allowExternal"`) But.. I didn't manage to make it work - not sure if it's possible to have conditional snapshots (that could automatically be updated, not by matching inline snapshot)?
What I basically want is if allowExternal is passed to check if it matches one snapshot, and if not if it matches the original snapshot (everything currently points me to having 2 separate folders is the solution to having separate tests for both options).

AriPerkkio:
Maybe we should have a separate `testing.mjs` script for testing different coverage options like `allowExternal`. It could be a single script in `package.json`. 

Something like:

```js
// vitest/test/coverage-test/testing-options.mjs

import { startVitest } from 'vitest/node'

const testCases = [
  {
    testConfig: {
      include: ['test/allow-external.test.ts'],
      coverage: { allowExternal: true, include: ['**/src/**', '**/test-utils/fixtures/**'] },
    },
    assertionConfig: {
      include: ['coverage-report-tests/allow-external.test.ts'],
    },
  },
]

for (const provider of ['v8', 'istanbul']) {
  for (const { testConfig, assertionConfig } of testCases) {
    // Run test case
    await startVitest('test', ['tests/'], {
      name: `With settings: ${JSON.stringify(testConfig)}`,
      ...testConfig,
      coverage: {
        enabled: true,
        provider,
        ...testConfig.coverage,
      },
    })

    // Check generated coverage report
    await startVitest('test', ['coverage-report-tests'], {
      ...assertionConfig,
    })
  }
}

```

And then we could leave the regular `testing.mjs` without `allowExternal`.

packages/vitest/src/node/config.ts
AriPerkkio: I think this change is fine. Users should defined `setupFiles` with absolute path if this causes issues.

AriPerkkio (approved):
I've refactored the testing setup of `test/coverage-test`. Now both `true|false` values of `allowExternal` are tested. 

Thanks @vojvodics!

packages/vitest/src/node/config.ts
vojvodics: Wouldn't actually something like this make more sense?
```suggestion
  resolved.coverage.exclude.push(...resolved.setupFiles.map(file => resolved.coverage.allowExternal ? file : relative(resolved.root, file)))
```

PR #3902, opened 2023-08-07
fix: pass environmentOptions to happy-dom integration

### LINKED ISSUE #3903: happy-dom integration does not uses `environmentOptions`
### Describe the bug

I notice that while using `environment: 'happy-dom'`, the values passed via `environmentOptions` are not being forward, forcing the users to define settings after it's being already initialized (e.g., via setupFiles)

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-v3zosm?file=vite.config.ts&initialPath=__vitest__/

### System Info

```shell
System:
    OS: macOS 13.5
    CPU: (8) arm64 Apple M1 Pro
    Memory: 48.19 MB / 16.00 GB
    Shell: 5.9 - /opt/homebrew/bin/zsh
  Binaries:
    Node: 16.20.1 - ~/.volta/tools/image/node/16.20.1/bin/node
    npm: 8.19.4 - ~/.volta/tools/image/node/16.20.1/bin/npm
    pnpm: 8.6.6 - ~/.volta/tools/image/pnpm/8.6.6/bin/pnpm
  Browsers:
    Brave Browser: 115.1.56.14
    Chrome: 115.0.5790.114
    Edge: 115.0.1901.188
    Safari: 16.6
  npmPackages:
    @vitejs/plugin-react-swc: 3.3.2 => 3.3.2 
    @vitest/coverage-c8: 0.31.4 => 0.31.4 
    vite: 4.4.8 => 4.4.8 
    vitest: 0.31.4 => 0.31.4
```

### Used Package Manager

pnpm

### Validations

### LINKED ISSUE #3905: Default URL when using Happy DOM as environment

Happy DOM and JSDOM both default the URL to "about:blank". For the JSDOM environment the URL is set to "http://localhost:3000", while when using Happy DOM it has not been set to any default value.

This is causing issues when constructing relative URLs (see https://github.com/capricorn86/happy-dom/issues/868).

### Suggested solution

My suggestion is to default the Happy DOM URL to the same URL as JSDOM (which is "http://localhost:3000").

### Alternative

_No response_

### Additional context

_No response_

### Validations

raulfdm:
### Description

This PR forwards the `environmentOptions` to `happy-dom` integration.

It follows the same principle defined in the `jsdom` integration

https://github.com/vitest-dev/vitest/blob/f4e6e99fa3a81fd7a97fb9b435da367dddba7fa4/packages/vitest/src/integrations/env/jsdom.ts#L32

> Reference: https://github.com/capricorn86/happy-dom/wiki/Getting-started#usage

Closes #3903 
Closes #3905

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/integrations/env/happy-dom.ts
+++ b/packages/vitest/src/integrations/env/happy-dom.ts
@@ -5,9 +5,12 @@ import { populateGlobal } from './utils'
 export default <Environment>({
   name: 'happy-dom',
   transformMode: 'web',
-  async setupVM() {
+  async setupVM({ 'happy-dom': happyDom = {} }) {
     const { Window } = await importModule('happy-dom') as typeof import('happy-dom')
-    const win = new Window() as any
+    const win = new Window({
+      ...happyDom,
+      url: happyDom.url || 'http://localhost:3000',
+    }) as any
 
     // TODO: browser doesn't expose Buffer, but a lot of dependencies use it
     win.Buffer = Buffer
@@ -26,11 +29,14 @@ export default <Environment>({
       },
     }
   },
-  async setup(global) {
+  async setup(global, { 'happy-dom': happyDom = {} }) {
     // happy-dom v3 introduced a breaking change to Window, but
     // provides GlobalWindow as a way to use previous behaviour
     const { Window, GlobalWindow } = await importModule('happy-dom') as typeof import('happy-dom')
-    const win = new (GlobalWindow || Window)()
+    const win = new (GlobalWindow || Window)({
+      ...happyDom,
+      url: happyDom.url || 'http://localhost:3000',
+    })
 
     const { keys, originals } = populateGlobal(global, win, { bindFunctions: true })
 
--- a/test/core/test/happy-dom.test.ts
+++ b/test/core/test/happy-dom.test.ts
@@ -2,6 +2,7 @@
 
 /**
  * @vitest-environment happy-dom
+ * @vitest-environment-options { "settings": { "disableCSSFileLoading": true }, "width": 1920 }
  */
 
 /* eslint-disable vars-on-top */
@@ -11,7 +12,18 @@ import { expect, it, vi } from 'vitest'
 declare global {
   // eslint-disable-next-line no-var
   var __property_dom: unknown
+  // eslint-disable-next-line no-var
+  var happyDOM: unknown
 }
+
+it('defaults URL to localhost:3000', () => {
+  expect(location.href).toBe('http://localhost:3000/')
+})
+
+it('accepts custom environment options', () => {
+  // default is false
+  expect((window.happyDOM as any).settings.disableCSSFileLoading).toBe(true)
+})
 
 it('defined on self/window are defined on global', () => {
   expect(self).toBeDefined()


packages/vitest/src/integrations/env/happy-dom.ts
```diff
@@ -5,9 +5,9 @@ import { populateGlobal } from './utils'
 export default <Environment>({
   name: 'happy-dom',
   transformMode: 'web',
-  async setupVM() {
+  async setupVM({ 'happy-dom': happyDom = {} }) {
     const { Window } = await importModule('happy-dom') as typeof import('happy-dom')
-    const win = new Window() as any
+    const win = new Window(happyDom) as any
```
userquin, line 10: can you include default url?:
```ts
    const {
      innerWidth,
      innerHeight,
      url = 'http://localhost:3000',
      settings,
    } = happyDom
    const win = new (GlobalWindow || Window)({
      innerWidth,
      innerHeight,
      url,
      settings,
    })
```

packages/vitest/src/integrations/env/happy-dom.ts
```diff
@@ -26,11 +26,11 @@ export default <Environment>({
       },
     }
   },
-  async setup(global) {
+  async setup(global, { 'happy-dom': happyDom = {} }) {
     // happy-dom v3 introduced a breaking change to Window, but
     // provides GlobalWindow as a way to use previous behaviour
     const { Window, GlobalWindow } = await importModule('happy-dom') as typeof import('happy-dom')
-    const win = new (GlobalWindow || Window)()
+    const win = new (GlobalWindow || Window)(happyDom)
```
userquin, line 33: same here

packages/vitest/src/integrations/env/happy-dom.ts
capricorn86: The issue for the default URL is #3905 (there is also a PR up that most likely will conflict when either is merged).

userquin: uhmm, check the title, it is about provide happy-dom env, maybe it would be better to include my suggestion here and remove default rul form your PR.

capricorn86: If this PR is approved, it is of course better to include both. Another option would be to merge #3907 first, as it is smaller and have a low risk.

raulfdm: Wouldn't that be a breaking change? I mean, if by default everyone is using the `about:blank` and suddenly it flips to `http://localhost:3000`, wouldn't break things? 🤔

capricorn86: I think that the scenarios where it could be breaking are very uncommon and most likely it is not breaking for anyone, but it is a good point.

The only scenario I can think of for now is if someone directly test against the current page URL (location.*) somehow.

capricorn86: I also want to add that I think it is expected that JSDOM and Happy DOM are using the same URL when switching between environments.

packages/vitest/src/integrations/env/happy-dom.ts
raulfdm: @userquin done

packages/vitest/src/integrations/env/happy-dom.ts
raulfdm: I've added in this PR the fallback url

capricorn86:
I have closed #3907 now as this PR has added the default URL to it.

packages/vitest/src/integrations/env/happy-dom.ts
```diff
@@ -5,9 +5,12 @@ import { populateGlobal } from './utils'
...
   name: 'happy-dom',
   transformMode: 'web',
-  async setupVM() {
+  async setupVM({ 'happy-dom': happyDom = {} }) {
     const { Window } = await importModule('happy-dom') as typeof import('happy-dom')
-    const win = new Window() as any
+    const win = new Window({
+      ...happyDom,
+      url: happyDom.url || 'http://localhost:3000',
+    }) as any
```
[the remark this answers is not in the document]
capricorn86: Shouldn't be the other way around in case somebody wants to override the URL?

capricorn86 (approved):

userquin (approved):

userquin:
We should add a test

raulfdm:
> We should add a test

@userquin I've added them. Let me know if it's enough

test/core/test/happy-dom.test.ts
```diff
@@ -11,8 +12,19 @@ import { expect, it, vi } from 'vitest'
...
 }
 
+it('defaults URL to localhost:3000', () => {
+  expect(location.href).toBe('http://localhost:3000/')
+})
+
+it('accepts custom environment options', () => {
+  // default is false
+  expect((window.happyDOM as any).settings.disableCSSFileLoading).toBe(true)
+})
```
raulfdm, line 26: Because we leverage on passing envrionment options via JSDoc code block, I didn't find a way where I could test passing the option vs not passing.

test/core/test/happy-dom.test.ts
```diff
@@ -11,8 +12,19 @@ import { expect, it, vi } from 'vitest'
 declare global {
   // eslint-disable-next-line no-var
   var __property_dom: unknown
+  // eslint-disable-next-line no-var
+  var happyDOM: unknown
 }
 
+it('defaults URL to localhost:3000', () => {
```
raulfdm, line 19: same applies here.

I'd like to try passing and not passing URL, but no idea how to do that using JSDoc comment.

packages/vitest/src/integrations/env/happy-dom.ts
```diff
@@ -5,9 +5,12 @@ import { populateGlobal } from './utils'
 export default <Environment>({
   name: 'happy-dom',
   transformMode: 'web',
-  async setupVM() {
+  async setupVM({ 'happy-dom': happyDom = {} }) {
```
sheremet-va, line 8: I think naming it `happyDom` in the config is fine

test/core/test/happy-dom.test.ts
sheremet-va: You can create a separate test file

sheremet-va (changes requested):
Please, can you add types to `environmentOptions` in the config?

packages/vitest/src/integrations/env/happy-dom.ts
capricorn86: The types can be found here:
```javascript
     import IHappyDOMOptions from 'happy-dom/lib/window/IHappyDOMOptions.js';
```

sheremet-va:
I'm going to do a release in a few minutes, but you didn't allow maintainers to push to your PR, so I created a separate one: #3972

Thank you for your contribution!

Issue #3816, opened 2023-07-26
Getting Node low level errors running vitest


lemoustachiste:
### Describe the bug

So I am testing migration of my project from jest to vitest.

But when I run `vitest run` through `npm run test`, I keep getting Node errors such as follows:

```
node(39994,0x7000083d9000) malloc: *** error for object 0x600002c44cd0: pointer being freed was not allocated
node(39994,0x700006fca000) malloc: *** error for object 0x600002c44cd0: pointer being freed was not allocated
node(39994,0x7000083d9000) malloc: *** set a breakpoint in malloc_error_break to debug
node(39994,0x700006fca000) malloc: *** set a breakpoint in malloc_error_break to debug
```

This error does not even get to a point of running tests. 

Or if some tests start running, then they stop unexpectedly with the following error:

```
[1]    31930 segmentation fault  npm run test
```

Errors occur with both node 16.20.0 and 18.16.1.

My tests are all TS, so is my project.

Any clue what could be causing this?

### Reproduction

Not sure if I can provide a repro at this point.

### System Info

```shell
System:
    OS: macOS 13.5
    CPU: (8) x64 Intel(R) Core(TM) i5-1038NG7 CPU @ 2.00GHz
    Memory: 21.56 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.16.1 - ~/.nvm/versions/node/v18.16.1/bin/node
    Yarn: 1.22.18 - /usr/local/bin/yarn
    npm: 8.12.1 - ~/work/cert-verifier-js/node_modules/.bin/npm
  Browsers:
    Chrome: 115.0.5790.102
    Safari: 16.6
  npmPackages:
    vitest: ^0.33.0 => 0.33.0
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
[...]

Issue #3143, opened 2023-04-06
Hitting Segmentation fault when running Vitest .


Christopher-Stevers:
### Describe the bug

When I run vitest on without limiting the threads, the process will crash half way through with the following output:
```
Segmentation fault (core dumped)
error Command failed with exit code 139.
```

### Reproduction

https://github.com/OpenQDev/OpenQ-Frontend/tree/vite-staging
I can't really provide a minimal reproduction because it seems to depend on the number of tests you run. To reproduce it with this repo just run `yarn` and then `yarn vitest`

### System Info

```shell
System:
    OS: Linux 5.4 Ubuntu 20.04.5 LTS (Focal Fossa)
    CPU: (16) x64 Intel(R) Xeon(R) Platinum 8272CL CPU @ 2.60GHz
    Memory: 26.75 GB / 31.36 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 19.7.0 - ~/nvm/current/bin/node
    Yarn: 1.22.19 - /usr/bin/yarn
    npm: 9.5.0 - ~/nvm/current/bin/npm
  npmPackages:
    @vitejs/plugin-react: ^3.1.0 => 3.1.0 
    vitest: ^0.29.8 => 0.29.8
```

### Used Package Manager

yarn

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

rushil23:
Facing the same here. Really blocking our build flows. 

In our case we're facing this issue when running the tests on a github action

Christopher-Stevers:
My work around at the moment has been to run it with the --no-threads flag when in circleci, but that slows down the workflow a lot.
[...]

Issue #3106, opened 2023-03-30
Vitest crashes when running tests with Prisma


nakleiderer:
### Describe the bug

NOTE: I'm cross-posting this issue from https://github.com/prisma/prisma/issues/18577 because it's unclear which tool is causing the issue, the bug might require collaboration between vitest and prisma, and for other members of the vitest or prisma communities to be able to search this issue in their respective repositories.

When running tests with Vitest and Prisma, I receive a variety of crashes: segmentation fault, bus error, and abort. Sometimes, Rust is able to print a stack trace before crashing:

<details><summary>Super long debug output and stack trace</summary>

```
❯ npm run test -- --run

> vitest-crash-reproduction@0.0.0 test
> vitest --run

  vite:config bundled config file loaded in 24.26ms +0ms
  vite:esbuild init tsconfck (root: /vitest-crash-reproduction) +0ms
  vite:esbuild init tsconfck (root: /vitest-crash-reproduction) +0ms
  vite:esbuild init tsconfck end +1ms
  vite:esbuild init tsconfck end +0ms
  vite:config using resolved config: {
  vite:config   test: {},
  vite:config   logLevel: 'error',
  vite:config   configFile: '/vitest-crash-reproduction/vite.config.ts',
  vite:config   mode: 'test',
  vite:config   plugins: [
  vite:config     'vite:pre-alias',
  vite:config     'alias',
  vite:config     'vitest',
  vite:config     'vitest:env-replacer',
  vite:config     'vitest:global-setup-plugin',
  vite:config     'vitest:css-disable',
  vite:config     'vite:modulepreload-polyfill',
  vite:config     'vite:resolve',
  vite:config     'vite:html-inline-proxy',
  vite:config     'vite:css',
  vite:config     'vite:esbuild',
  vite:config     'vite:json',
  vite:config     'vite:wasm-helper',
  vite:config     'vite:worker',
  vite:config     'vite:asset',
  vite:config     'vitest:coverage-transform',
  vite:config     'vite:wasm-fallback',
  vite:config     'vite:define',
  vite:config     'vite:css-post',
  vite:config     'vite:worker-import-meta-url',
  vite:config     'vite:asset-import-meta-url',
  vite:config     'vite:dynamic-import-vars',
  vite:config     'vite:import-glob',
  vite:config     'vitest:css-empty-post',
  vite:config     'vite:client-inject',
  vite:config     'vite:import-analysis'
  vite:config   ],
[...]

PR #3925, opened 2023-08-10
feat!: support multiple parallel `child_process`

AriPerkkio:
### Description

- Many projects are unable to use `node:worker_threads` and are forced to use a single `node:child_process` via `--no-threads`. That's stable, but slow. This PR allows using multiple `node:child_process` for running tests similarly as `--threads` does.
  - https://github.com/vitest-dev/vitest/issues/3143
  - https://github.com/vitest-dev/vitest/issues/3106
  - https://github.com/vitest-dev/vitest/issues/3816
  - https://github.com/tinylibs/tinypool/issues/48
- Ref. https://github.com/tinylibs/tinypool/issues/64

#### BREAKING CHANGES

`useFakeTimers()` no longer automatically mocks `process.nextTick`. Users can still mock it by explicitly specifying `useFakeTimers({ toFake: ['nextTick'] })`.

Mocking `process.nextTick` by using `useFakeTimers({ toFake: ['nextTick'] })` is no longer supported when running in `--no-threads`. Node's `child_process` uses `process.nextTick` internally and it is really difficult to support this use case. Users can still mock `process.nextTick` when running with `--threads`.

`--no-threads` no longer runs tests in the same test context. Tests are now isolated in separate processes similar to how `--threads` does with `worker_threads`.

#### Testing

 - [x] `pnpm test:ci:no-threads`
 - Manual testing with third party projects:
   - [x] [`zustand`](https://github.com/pmndrs/zustand): Tests require isolation so comparing speed change doesn't make sense here. All tests pass with new isolated `--no-threads`.
   - [x] [`vuejs/core`](https://github.com/vuejs/core). Uses `--no-threads` on CI. Speed improvement 48.57s -> 26.96s.
   - [x] [`netlify/cli`](https://github.com/netlify/cli). Tests pass with `--no-threads`.
   - [x] [`lerna-lite`](https://github.com/lerna-lite/lerna-lite). Tests require `child_process` and have been using `--no-threads`. Speed improvement 67s -> 24s.

___

### Tests

### Documentation

### Changesets

### DIFF
--- a/test/core/test/fixtures/timers.suite.ts
+++ b/test/core/test/fixtures/timers.suite.ts
@@ -15,6 +15,8 @@ import { FakeTimers } from '../../../../packages/vitest/src/integrations/mock/ti
 
 class FakeDate extends Date {}
 
+const isChildProcess = globalThis.__vitest_worker__.isChildProcess
+
 describe('FakeTimers', () => {
   afterEach(() => {
     vi.useRealTimers()
@@ -49,7 +51,7 @@ describe('FakeTimers', () => {
       expect(global.clearInterval).not.toBe(undefined)
     })
 
-    it('mocks process.nextTick if it exists on global', () => {
+    it.skipIf(isChildProcess)('mocks process.nextTick if it exists on global', () => {
       const origNextTick = () => {}
       const global = {
         Date: FakeDate,
@@ -62,6 +64,30 @@ describe('FakeTimers', () => {
       const timers = new FakeTimers({ global, config: { toFake: ['nextTick'] } })
       timers.useFakeTimers()
       expect(global.process.nextTick).not.toBe(origNextTick)
+    })
+
+    it.runIf(isChildProcess)('does not mock process.nextTick if it exists on global and is child_process', () => {
+      const origNextTick = () => {}
+      const global = {
+        Date: FakeDate,
+        clearTimeout,
+        process: {
+          nextTick: origNextTick,
+        },
+        setTimeout,
+      }
+      const timers = new FakeTimers({ global })
+      timers.useFakeTimers()
+      expect(global.process.nextTick).toBe(origNextTick)
+    })
+
+    it.runIf(isChildProcess)('throws when is child_process and tries to mock nextTick', () => {
+      const global = { process, setTimeout, clearTimeout }
+      const timers = new FakeTimers({ global, config: { toFake: ['nextTick'] } })
+
+      expect(() => timers.useFakeTimers()).toThrow(
+        'process.nextTick cannot be mocked inside child_process',
+      )
     })
 
     it('mocks setImmediate if it exists on global', () => {
@@ -96,7 +122,7 @@ describe('FakeTimers', () => {
   })
 
   describe('runAllTicks', () => {
-    it('runs all ticks, in order', () => {
+    it.skipIf(isChildProcess)('runs all ticks, in order', () => {
       const global = {
         Date: FakeDate,
         clearTimeout,
@@ -144,7 +170,7 @@ describe('FakeTimers', () => {
       expect(nextTick).toHaveBeenCalledTimes(0)
     })
 
-    it('only runs a scheduled callback once', () => {
+    it.skipIf(isChildProcess)('only runs a scheduled callback once', () => {
       const global = {
         Date: FakeDate,
         clearTimeout,
@@ -168,7 +194,7 @@ describe('FakeTimers', () => {
       expect(mock1).toHaveBeenCalledTimes(1)
     })
 
-    it('throws before allowing infinite recursion', () => {
+    it.skipIf(isChildProcess)('throws before allowing infinite recursion', () => {
       const global = {
         Date: FakeDate,
         clearTimeout,
@@ -322,7 +348,7 @@ describe('FakeTimers', () => {
       )
     })
 
-    it('also clears ticks', () => {
+    it.skipIf(isChildProcess)('also clears ticks', () => {
       const global = { Date: FakeDate, clearTimeout, process, setTimeout }
       const timers = new FakeTimers({ global, config: { toFake: ['nextTick', 'setTimeout'] } })
       timers.useFakeTimers()
@@ -428,7 +454,7 @@ describe('FakeTimers', () => {
       )
     })
 
-    it('also clears ticks', async () => {
+    it.skipIf(isChildProcess)('also clears ticks', async () => {
       const global = { Date: FakeDate, clearTimeout, process, setTimeout, Promise }
       const timers = new FakeTimers({ global, config: { toFake: ['setTimeout', 'nextTick'] } })
       timers.useFakeTimers()
@@ -1062,7 +1088,7 @@ describe('FakeTimers', () => {
       expect(global.clearInterval).toBe(nativeClearInterval)
     })
 
-    it('resets native process.nextTick when present', () => {
+    it.skipIf(isChildProcess)('resets native process.nextTick when present', () => {
       const nativeProcessNextTick = vi.fn()
 
       const global = {
@@ -1143,7 +1169,7 @@ describe('FakeTimers', () => {
       expect(global.clearInterval).not.toBe(nativeClearInterval)
     })
 
-    it('resets mock process.nextTick when present', () => {
+    it.skipIf(isChildProcess)('resets mock process.nextTick when present', () => {
       const nativeProcessNextTick = vi.fn()
 
       const global = {
@@ -1216,7 +1242,7 @@ describe('FakeTimers', () => {
       timers.useRealTimers()
     })
 
-    it('includes immediates and ticks', () => {
+    it.skipIf(isChildProcess)('includes immediates and ticks', () => {
       const timers = new FakeTimers({ global, config: { toFake: ['setTimeout', 'setImmediate', 'nextTick'] } })
 
       timers.useFakeTimers()
--- a/test/core/test/fixtures/timers.suite.ts
+++ b/test/core/test/fixtures/timers.suite.ts
@@ -59,7 +59,7 @@ describe('FakeTimers', () => {
         },
         setTimeout,
       }
-      const timers = new FakeTimers({ global })
+      const timers = new FakeTimers({ global, config: { toFake: ['nextTick'] } })
       timers.useFakeTimers()
       expect(global.process.nextTick).not.toBe(origNextTick)
     })
@@ -106,7 +106,7 @@ describe('FakeTimers', () => {
         setTimeout,
       }
 
-      const timers = new FakeTimers({ global })
+      const timers = new FakeTimers({ global, config: { toFake: ['nextTick'] } })
       timers.useFakeTimers()
 
       const runOrder = []
@@ -154,7 +154,7 @@ describe('FakeTimers', () => {
         setTimeout,
       }
 
-      const timers = new FakeTimers({ global })
+      const timers = new FakeTimers({ global, config: { toFake: ['nextTick'] } })
       timers.useFakeTimers()
 
       const mock1 = vi.fn()
@@ -178,7 +178,7 @@ describe('FakeTimers', () => {
         setTimeout,
       }
 
-      const timers = new FakeTimers({ global, config: { loopLimit: 100 } })
+      const timers = new FakeTimers({ global, config: { loopLimit: 100, toFake: ['nextTick'] } })
 
       timers.useFakeTimers()
 
@@ -324,7 +324,7 @@ describe('FakeTimers', () => {
 
     it('also clears ticks', () => {
       const global = { Date: FakeDate, clearTimeout, process, setTimeout }
-      const timers = new FakeTimers({ global })
+      const timers = new FakeTimers({ global, config: { toFake: ['nextTick', 'setTimeout'] } })
       timers.useFakeTimers()
 
       const fn = vi.fn()
@@ -430,7 +430,7 @@ describe('FakeTimers', () => {
 
     it('also clears ticks', async () => {
       const global = { Date: FakeDate, clearTimeout, process, setTimeout, Promise }
-      const timers = new FakeTimers({ global })
+      const timers = new FakeTimers({ global, config: { toFake: ['setTimeout', 'nextTick'] } })
       timers.useFakeTimers()
 
       const fn = vi.fn()
@@ -1071,7 +1071,7 @@ describe('FakeTimers', () => {
         process: { nextTick: nativeProcessNextTick },
         setTimeout,
       }
-      const timers = new FakeTimers({ global })
+      const timers = new FakeTimers({ global, config: { toFake: ['nextTick'] } })
       timers.useFakeTimers()
 
       // Ensure that timers has overridden the native timer APIs
@@ -1152,7 +1152,7 @@ describe('FakeTimers', () => {
         process: { nextTick: nativeProcessNextTick },
         setTimeout,
       }
-      const timers = new FakeTimers({ global })
+      const timers = new FakeTimers({ global, config: { toFake: ['nextTick'] } })
       timers.useRealTimers()
 
       // Ensure that the real timers are installed at this point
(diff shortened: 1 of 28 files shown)


AriPerkkio:
Seeing some process hangs when running Vitest's test cases with `--no-threads`. Debugging these is really time consuming and I've spent quite some time looking into this. I'm starting to think we should implement graceful `worker_thread` and `child_process` termination in Tinypool. That would allow us to implement Jest's `--detect-open-handles` feature too.

These errors come up locally and on CI:

```
⎯⎯⎯⎯⎯⎯ Unhandled Error ⎯⎯⎯⎯⎯⎯⎯
Error: [birpc] timeout on calling "onTaskUpdate"
 ❯ Timeout._onTimeout ../../packages/vitest/dist/vendor-index.9acf798a.js:38:22
 ❯ listOnTimeout node:internal/timers:569:17
 ❯ processTimers node:internal/timers:512:7
```

sheremet-va:
> Debugging these is really time consuming and I've spent quite some time looking into this. 

Welcome to my world! For me it's almost always something related to timer mocks (nextTick included).

> I'm starting to think we should implement graceful worker_thread and child_process termination in Tinypool.

Will it help? Or just to debug?

AriPerkkio:
> Will it help? Or just to debug?

Mostly for debugging. But still, I'm not sure if it's good practice to terminate workers and processess with `.kill()` and `.terminate()` when there can still be all kind of stuff running, e.g. chokidar with fsevents. I'm starting to think that it's causing some of these process hangs but don't really have proof for it.

The way how `jest-worker` does this is that main thread instructs thread/process (worker) to [remove all its message handlers](https://github.com/jestjs/jest/blob/25a8785584c9d54a05887001ee7f498d489a5441/packages/jest-worker/src/workers/processChild.ts#L155C52-L156) so that there is nothing keeping the worker running, and then it waits for the worker to gracefully exit. If it doesn't exit in specific timeout, main threads sends it SIGTERM. If it still doesn't exit, [it sends SIGKILL](https://github.com/jestjs/jest/blob/25a8785584c9d54a05887001ee7f498d489a5441/packages/jest-worker/src/workers/ChildProcessWorker.ts#L463C8-L477). What Tinypool currently does is that it immediately sends SIGTERM and then hangs if worker doesn't exit - or throws [timeout error](https://github.com/tinylibs/tinypool/pull/50).

But if Vitest used graceful termination for workers, it could impact in performance. We would not be killing workers as fast as possible - instead we would be waiting for them to shut down themselves. I'm not sure if this performance issue is real or just speculation. Jest or Vitest's `vm-threads` don't really suffer from this as isolation doesn't require recycling workers.

Maybe I'll implement `--detect-open-handles` into Tinypool so that it's optional flag that can be turned on when needed. It would definitely help debugging this PR's issues.

test/core/test/fixtures/timers.suite.ts
```diff
@@ -338,7 +338,7 @@ describe('FakeTimers', () => {
     })
   })
 
-  describe('runAllTimersAsync', () => {
+  describe.skipIf('TODO Enable this')('runAllTimersAsync', () => {
```
AriPerkkio, line 341: > For me it's almost always something related to timer mocks

To talk about timers... 😅 
Every other test is now passing (even on CI) but this one. This is throwing the `Error: [birpc] timeout on calling "onTaskUpdate"` error. I think that error used to come up on `main` branch's CI maybe a month ago but now it's gone from there.

I can reproduce this error locally on MacOS when using [`stress(1)` CLI tool](https://github.com/resurrecting-open-source-projects/stress) to simulate CPU load. That tool is available on `brew` for easy installation.

```
# In one terminal
$ stress --cpu 8 --io 4 --vm 2 --vm-bytes 128M --timeout 200s

# In second terminal. Run couple of times if it doesn't come up immediately.
$ cd vitest/test/core 
$ pnpm run test --run --no-threads --allow-only
```

I've been looking at Vitest's `onTaskUpdate` usage for long now and cannot understand why that error is coming up. Runner should be `await`'ing that call to finish. But I guess when timers are mocked, all kinds of unexpected behavior can come up. The test case is even passing `global.Promise` to `fake-timers`. 🤔 

I think I'll take a look into `@sinonjs/fake-timers` sources next.

AriPerkkio: Looks like this is actually a bug on `main` and doesn't exactly relate to this PR's changes.

When a test case mocks `process.nextTick` during test case, any execution that relies on `nextTick` will not be run. When `useRealTimers()` is called, those pending tasks are still not executed. In `birpc` we are storing the original `setTimeout`, `clearTimeout` and `Math.random` to prevent similar cases. I'm not sure if we even can keep internal dependencies running properly when test execution shares the global context with runner. VM isolation might help here?

Here is test case that will fail on latest Vitest release or `main` branch when run as `vitest --no-threads`. It might need to be ran multiple times, or slightly increase the timeout value. 

```ts
import { test, vi } from "vitest";

test("leaky process mocking", async () => {
  const ogSetTimeout = setTimeout;

  // Fake `process.nextTick` and execute longer than 10ms. Use heavy looping or simple `setTimeout`
  vi.useFakeTimers({ toFake: ["nextTick"] });

  process.stdout.write("\nThis is executed\n\n");
  await new Promise((resolve) => ogSetTimeout(resolve, 100)); // onTaskUpdate uses 10ms
  process.stdout.write("\nThis is also executed\n\n");

  vi.useRealTimers();

  process.stdout.write("\nTest finished\n\n");
});
```

The test case runs for 60s and `birpc` timeouts. Also it fails to report failed test case from child process now. 

image



Here is minimal-ish reproduction using just `@sinonjs/fake-timers` and `node:child_process`. When main process uses `process.send`, and subprocess has `process.nextTick` mocked, the `process.on("message", callback)` is not executed. And after `useRealTimers()` call those pending tasks are not executed at all.

<details>

```js
/*
 * Sends multiple messages from main process to subprocess.
 * Subprocess will fail if it does not receive all the messages.
 * Between the message sending, subprocess will install fake-timers.
 * After a while, it will uninstall fake-timers.
 *
 * Requires:
 * pnpm i @sinonjs/fake-timers@11.0.0
 */

import { fork } from "node:child_process";

const filename = fileURLToPath(import.meta.url);
const isMainThread = !Boolean(process.env.CHILD_PROCESS);

const INTERVAL = 1000; // Time in ms between messages
const LOOPS = Array(10).fill(0).map((_, i) => i); // Amount of messages to be sent

if (isMainThread) {
  console.log("[main] running");
  const subprocess = fork(filename, { env: { CHILD_PROCESS: "true" } });
  const onExit = new Promise((r) => subprocess.on("exit", r));
  const onReady = new Promise((r) => subprocess.on("message", (m) => m === "ready" && r()));

  await onReady;
  console.log("[main] subprocess ready\n");

  // Send amount of LOOPS to subprocess with INTERVAL delay in between
  for (const index of LOOPS) {
    await new Promise((r) => setTimeout(r, INTERVAL));
    console.log("[main] sending message", index);
    subprocess.send({ loop: index });
  }

  // Let process exit when it's done
  subprocess.send({ exit: "graceful" });

  await onExit;
  console.log("[main] subprocess exited");
} else {
  console.log("[subprocess] running");

  const ogSetTimeout = setTimeout;
  const ogClearTimeout = clearTimeout;

  // These are used to indicate whether all messages were not received
  const timeouts = LOOPS.map((i) =>
    ogSetTimeout(() => { throw new Error(`Timeout ${i}`) }, INTERVAL * LOOPS.length * 1.5)
  );

  async function onMessage(message) {
    // Remove timeout from given LOOP message. If any messages are missed, a timeout error is expected
    if (message?.loop !== undefined) {
      console.log("[subprocess] received loop", message.loop);
      ogClearTimeout(timeouts[message.loop]);
      return;
    }

    if (message?.exit === "graceful") {
      process.off("message", onMessage);
      return console.log("[subprocess] exitting gracefully");
    }

    console.log("[subprocess] received unknown message", message);
  }

  process.on("message", onMessage);

  // Setup fake-timers
  const { withGlobal } = await import("@sinonjs/fake-timers");

  class FakeDate extends Date {}
  const global = { Date: FakeDate, clearTimeout, process, setTimeout, Promise };
  const fakeTimers = withGlobal(global);

  // Install fake-timers after 3s. After another 3s uninstall them.
  ogSetTimeout(() => {
    console.log("[subprocess] installing fake-timers");
    const clock = fakeTimers.install({
      now: Date.now(),
      toFake: Object.keys(fakeTimers.timers),
    });

    setTimeout(() => {
      console.log("[subprocess] uninstalling fake-timers");
      clock.uninstall();
    }, 3000);
  }, 3000);

  process.send("ready");
}
```

</details>

Does this analysis sound correct? The amount of debugging across `vitest`, `birpc`, `tinypool` and `@sinonjs/fake-timers` was a lot. 😫

sheremet-va: Sounds correct, I also had a lot of problems with `nextTick` the first time child process wad implemented. I fixed it by restoring original `nextTick` before calling rpc method, does it not work anymore?

AriPerkkio: > fixed it by restoring original `nextTick` before calling rpc method

Good point, I was unaware of this. This is done by `withSafeTimers`, right?

That one is not `await`'ing the given callback. Calls of `birpc` are Promises when they are not specified in its `eventNames` config. By adding `await` here the failing minimal test case above is constantly passing:

https://github.com/vitest-dev/vitest/blob/91853bb204ed8dfc3750d54b1b79409168ea440b/packages/vitest/src/runtime/rpc.ts#L30-L31

However that one still doesn't fix all issues, as now `rpc().snapshotSaved` seems to timeout on Windows. I'll continue debugging. 🙃

sheremet-va: The main reason why it wasn’t awaiting is because the function itself was running synchronously and returned a promise, so all globals should’ve been called by that time. Maybe implementation changed in the meantime? Or maybe it was never working as intended 👀

AriPerkkio: > function itself was running synchronously and returned a promise, so all globals should’ve been called by that time.

When the rpc method is called synchronously, it means that only the subprocess's `process.send` is ensured to be run using original `nextTick`. The response coming from main process is not received in `process.on('message', callback)` if `nextTick` is mocked. We'll need to `await` the whole rpc call to make sure response is received from main process. 

But this also means that any code in user's test case that runs between this cycle is running without mocked `nextTick`.

sheremet-va: I would just disallow mocking `nextTick` to be honest

AriPerkkio: I'm starting to think that would be the best option too. This is driving me crazy. Debugging NodeJS when such low level internals are mocked is not fun. 
We could raise an error if `useFakeTimers({toFake: ['nextTick']})` is called when `node:child_process` is used. I don't think that many users are even interested in `nextTick` mocking.

ghiscoding:
> * [`lerna-lite`](https://github.com/lerna-lite/lerna-lite). Tests require `child_process` and have been using `--no-threads`. Speed improvement 67s -> 24s. Some snapshots are wrapped in double quotes now. 🤔

@AriPerkkio I'm really excited about this PR, especially that I'm on Windows so this will be a big boost for me... and I just saw that you started testing Lerna-Lite, awesome, so about the comment you wrote, I might be able to shed some light, I had a few issues when I first moved to Vitest (like this issue #3129 that I logged in Vitest). Basically Lerna (Lerna-Lite is a fork of Lerna) had many custom serializers and running them with `--no-threads` had the consequence of being run in the same global scope causing some problems as described by @sheremet-va in the referenced issue 

> All tests with the same environment in `--no-threads` runs in a single child_process and they all share the same global scope. Is it possible that one test just affects the other?

I fixed this problem in this Lerna-Lite [PR](https://github.com/lerna-lite/lerna-lite/pull/650) by duplicating some of these serializer in certain spec files that had quote escaping issues

So in short, it might be related to the double quotes problem you are encountering when testing Lerna-Lite. I don't mind reverting this Lerna-Lite [PR](https://github.com/lerna-lite/lerna-lite/pull/650) or changing any code that could help with the extra quotes you're seeing, whatever makes it works, I'm all good with it 😉

AriPerkkio:
Thanks for the insight @ghiscoding - you just save me an hour or two of debugging. In the current testing setup of `lerna-lite` the snapshot serializers are indeed leaking from one test file to another. I fixed the snapshot issues by adding the serializer to following files: `version-build-metadata.spec.ts`, `version-conventional-commits.spec.ts`, `version-ignore-changes.spec.ts`. Now all tests are passing and running quickly.

AriPerkkio:
What kind of API should `child_process` pool expose to users, @sheremet-va?

- `--single-process`, `minProcess`, `maxProcess` (equivalents of `--single-thread`, `minThreads`, `maxThreads`
- Or simply call threads and processes as "workers" as Jest does, and have `--single-worker`, `minWorkers`, `maxWorkers`? Pool selection would be done using `--threads`, `--no-threads`? 

Currenly this PR does not implement a `child_process` equivalent of `--single-threads` - do we need one?

sheremet-va:
The problem here is naming 😄

Maybe implement current names for bow, but we already discussed the new naming convention for pools - maybe we can implement it in the same breaking change release, but in a separate PR.

ghiscoding:
> Thanks for the insight @ghiscoding - you just saved me an hour or two of debugging. In the current testing setup of `lerna-lite` the snapshot serializers are indeed leaking from one test file to another. I fixed the snapshot issues by adding the serializer to following files: `version-build-metadata.spec.ts`, `version-conventional-commits.spec.ts`, `version-ignore-changes.spec.ts`. Now all tests are passing and running quickly.

I would be glad to receive a PR with all your changes for this upcoming feature, even if it's in draft until this is released, that would save me some time when it becomes available. Thanks a ton for all your hard work :)

AriPerkkio:
> Maybe implement current names for now, but we already discussed the new naming convention for pools - maybe we can implement it in the same breaking change release, but in a separate PR.

Yes, now I remember it - again. 😄 Configuring via pools make this much easier. 

I'll leave this PR to use `--no-threads` and implement `--single-thread` equivalent for `child_process`. I don't think there's a way to use that without introducing new configuration option but let's leave that to another PR.

AriPerkkio:
> [...] and implement `--single-thread` equivalent for `child_process`. I don't think there's a way to use that without introducing new configuration option but let's leave that to another PR.

Looks like I already implemented this. It's available with `vitest --no-threads --single-thread`. The `--no-thread` makes Vitest pick `pools/child.ts` and `--single-thread` makes it run in a single process. 

> #### TODO
>
> - [ ] Reporters work weird in test/public-api and test/stacktraces when run with --no-threads. Enabling CI=true fixes this.

Looks like this bug is present on `main` too. When running Vitest through JS API (`startVitest()`) with `--no-threads`, the reporters are not resolved correctly. This can be tested with:

```sh
cd vitest/test/public-api

# Pass
pnpm run test --threads --run
CI=true pnpm run test --no-threads --run

# Fail
pnpm run test --no-threads --run
```

ghiscoding:
so is this complete? on which milestone should we expect this to land, in v0.35 or v1.0? As you can see I'm very exited about it 😄

sheremet-va:
> so is this complete? on which milestone should we expect this to land, in v0.35 or v1.0? As you can see I'm very exited about it 😄

The next major is 1.0.0-beta.0

ghiscoding:
> The next major is 1.0.0-beta.0

@sheremet-va ok but just so you know, you still have a v0.35 milestone that exist with 1 PR assign to it 😋 
https://github.com/vitest-dev/vitest/milestone/11

awesome work guys

ghiscoding:
> Many projects are unable to use `node:worker_threads` and are forced to use a single `node:child_process` via `--no-threads`. That's stable, but slow. This PR allows using multiple `node:child_process` for running tests similarly as `--threads` does.
> * [Hitting Segmentation fault when running Vitest . #3143](https://github.com/vitest-dev/vitest/issues/3143)
> * [Vitest crashes when running tests with Prisma #3106](https://github.com/vitest-dev/vitest/issues/3106)
> * [Getting Node low level errors running vitest #3816](https://github.com/vitest-dev/vitest/issues/3816)

@AriPerkkio I'm just curious, were the issues you referenced supposed to be fixed (aka closed) since this PR got merged? Thanks a ton for the PR on my side 😃

AriPerkkio:
Yes, I think we should close those now. These changes are the best solution we can offer for cases where `node:worker_threads` simply won't work with users code. 

I'll go through those later today and leave a comment describing the changes while closing them.

PR #3944, opened 2023-08-14
fix: define process variable and using import.meta.env together did not work

### LINKED ISSUE #3735: Functions with parameters cant find import.meta.env
### Describe the bug

When testing a function that has parameters and also utilises import.meta.env, the test will always fail as `import.meta.env` is undefined. This problem does not exist for functions that take no arguments and they resolve `import.meta.env` to an object which can be edited via `vi.stubEnv`. 

### Reproduction

This problem is very easy to reproduce. All that is needed is two functions, one take takes an argument and returns an object with the argument and some `import.meta.env` value, and one that simply has the `import.meta.env` value. This is the simple example I use to demonstrate the issue: 

```javascript
export const getDefaults = () => ({ process: '', name: 'some name', mode: import.meta.env.MODE });

export const getDefaultsWithParam = (process: string) => ({
  process,
  name: 'some name',
  mode: import.meta.env.MODE,
});
```

This needs to be tested. Create a file that covers both these functions and for each of them test that when the function is called, it returns an object that contains the correct values. For each of them have 1 attempt that does NOT stub the environment variable and one that does. 

A full working example of this can be found on [StackBlitz](https://stackblitz.com/edit/vitest-dev-vitest-fccwec?file=test%2Fexample.test.ts)

### System Info

```shell
System:
    OS: macOS 13.4
    Memory: 462.12 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.11.0 - ~/.nvm/versions/node/v18.11.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v18.11.0/bin/yarn
    npm: 8.19.2 - ~/.nvm/versions/node/v18.11.0/bin/npm
  Browsers:
    Chrome: 114.0.5735.198
    Edge: 114.0.1823.51
    Firefox: 114.0.2
    Firefox Developer Edition: 100.0
    Safari: 16.5
  npmPackages:
    @vitest/coverage-c8: ^0.24.5 => 0.24.5 
    @vitest/coverage-v8: ^0.32.0 => 0.32.0 
    @vitest/ui: latest => 0.32.0 
    vite: ^4.2.1 => 4.2.2 
    vitest: ^0.32.0 => 0.32.0
```

### Used Package Manager

yarn

### Validations

Dunqing:
### Description

close #3735

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vite-node/src/utils.ts
+++ b/packages/vite-node/src/utils.ts
@@ -199,3 +199,21 @@ function traverseBetweenDirs(
     longerDir = dirname(longerDir)
   }
 }
+
+export function createImportMetaEnvProxy() {
+  // packages/vitest/src/node/plugins/index.ts:146
+  const booleanKeys = [
+    'DEV',
+    'PROD',
+    'SSR',
+  ]
+  return new Proxy(process.env, {
+    get(_, key) {
+      if (typeof key !== 'string')
+        return undefined
+      if (booleanKeys.includes(key))
+        return !!process.env[key]
+      return process.env[key]
+    },
+  })
+}
--- a/packages/vite-node/src/utils.ts
+++ b/packages/vite-node/src/utils.ts
@@ -215,5 +215,16 @@ export function createImportMetaEnvProxy() {
         return !!process.env[key]
       return process.env[key]
     },
+    set(_, key, value) {
+      if (typeof key !== 'string')
+        return true
+
+      if (booleanKeys.includes(key))
+        process.env[key] = value ? '1' : ''
+      else
+        process.env[key] = value
+
+      return true
+    },
   })
 }
--- a/packages/vite-node/src/client.ts
+++ b/packages/vite-node/src/client.ts
@@ -299,7 +299,7 @@ export class ViteNodeRunner {
     const modulePath = cleanUrl(moduleId)
     // disambiguate the `<UNIT>:/` on windows: see nodejs/node#31710
     const href = pathToFileURL(modulePath).href
-    const meta = { url: href }
+    const meta = { url: href, env: process.env }
     const exports = Object.create(null)
     Object.defineProperty(exports, Symbol.toStringTag, {
       value: 'Module',
--- a/packages/vite-node/src/client.ts
+++ b/packages/vite-node/src/client.ts
@@ -6,7 +6,7 @@ import { fileURLToPath, pathToFileURL } from 'node:url'
 import vm from 'node:vm'
 import { resolve } from 'pathe'
 import createDebug from 'debug'
-import { VALID_ID_PREFIX, cleanUrl, isInternalRequest, isNodeBuiltin, isPrimitive, normalizeModuleId, normalizeRequestId, slash, toFilePath } from './utils'
+import { VALID_ID_PREFIX, cleanUrl, createImportMetaEnvProxy, isInternalRequest, isNodeBuiltin, isPrimitive, normalizeModuleId, normalizeRequestId, slash, toFilePath } from './utils'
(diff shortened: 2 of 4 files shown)


sheremet-va (approved):

sheremet-va:
You changed your avatar, and now I don't recognize you 😄 

What do you think about also proxying the env to allow booleans while we are at it? I wrote a proxy, but didn't public it yet:

```ts
export function createImportMetaEnvProxy() {
  const booleanKeys = [
    'DEV',
    'PROD',
    'SSR',
  ]
  return new Proxy({}, {
    get(_, key) {
      if (typeof key !== 'string')
        return undefined
      if (booleanKeys.includes(key))
        return !!process.env[key]
      return process.env[key]
    },
    has(_, key) {
      if (typeof key !== 'string')
        return false
      return !!process.env[key]
    },
    set(_, key, newValue) {
      if (typeof key !== 'string')
        return true
      process.env[key] = newValue
      return true
    },
    deleteProperty(_, key) {
      if (typeof key !== 'string')
        return true
      delete process.env[key]
      return true
    },
  })
}
```

Dunqing:
> You changed your avatar, and now I don't recognize you 😄

Haha, I'll be participating in `vitest` contributions when I have time, you'll be impressed with my new avatar!

> What do you think about also proxying the env to allow booleans while we are at it? I wrote a proxy, but didn't public it yet:
> 
> ```ts
> export function createImportMetaEnvProxy() {
>   const booleanKeys = [
>     'DEV',
>     'PROD',
>     'SSR',
>   ]
>   return new Proxy({}, {
>     get(_, key) {
>       if (typeof key !== 'string')
>         return undefined
>       if (booleanKeys.includes(key))
>         return !!process.env[key]
>       return process.env[key]
>     },
>     has(_, key) {
>       if (typeof key !== 'string')
>         return false
>       return !!process.env[key]
>     },
>     set(_, key, newValue) {
>       if (typeof key !== 'string')
>         return true
>       process.env[key] = newValue
>       return true
>     },
>     deleteProperty(_, key) {
>       if (typeof key !== 'string')
>         return true
>       delete process.env[key]
>       return true
>     },
>   })
> }
> ```

I think we should allow booleans, but this implementation is a little complicated. Look like we  just only passing `get` function

```ts
export function createImportMetaEnvProxy() {
  const booleanKeys = [
    'DEV',
    'PROD',
    'SSR',
  ]
  return new Proxy(process.env, {
    get(_, key) {
      if (typeof key !== 'string')
        return undefined
      if (booleanKeys.includes(key))
        return !!process.env[key]
      return process.env[key]
    }
  })
}
```

sheremet-va:
> I think we should allow booleans, but this implementation is a little complicated. Look like we just only passing `get` function

I'm fine with any implementation that does the job 😄

packages/vite-node/src/utils.ts
```diff
@@ -199,3 +199,21 @@ function traverseBetweenDirs(
...
+  ]
+  return new Proxy(process.env, {
+    get(_, key) {
+      if (typeof key !== 'string')
+        return undefined
+      if (booleanKeys.includes(key))
+        return !!process.env[key]
+      return process.env[key]
+    },
+  })
```
sheremet-va, line 218: With this implementation I have a question: if you set `import.meta.env.SSR = false` will it be `false` or `true`? Since `process.env` casts a value to a string, I would assume that `process.env.SSR === 'false'` which will be `true` when casted back to boolean

sheremet-va (changes requested):

packages/vite-node/src/utils.ts
Dunqing: I think it's correct now.

PR #3949, opened 2023-08-14
fix: only consider cwd when using glob (fix #3802)

### LINKED ISSUE #3802: `exclude` filter should not consider absolute path
### Describe the bug

In our private CI system, all projects will be put into `$HOME/.cache/projects` when cloning repositories. However, Vitest's default `exclude` filter will ignore all the files whose paths contains `.cache` . This sounds OK when there're some directories called `.cache` in current working directory. But for us, `$HOME/.cache` isn't not a directory under current working directory.

When researching the logic of Vitest, I found that the `absolute` option of `fast-glob` was set to `true`, so all the test files (e.g. `$HOME/.cache/projects/example.test.ts`) will match that `exclude` filter, which causes Vitest exits with `No test files found` error.

Here is the `absolute` option: https://github.com/vitest-dev/vitest/blob/8693449b412743f20a63fd9bfa1a9054aa74613f/packages/vitest/src/node/workspace.ts#L140

I know we can change our Vitest config to let Vitest not ignore the paths contain `.cache`, but I'm still curious why the `absolute` option was enabled. In my opinion, both `include` and `exclude` options should only be considered under current working directory with relative path, not absolute path.

### Reproduction

This issue can't be simulated with online sandbox or a reproduction repo, but this issue can be reproduced by moving a project to a directory called `.cache`. You can simply create a directory called `.cache` and move into it.

### System Info

```shell
System: Linux
Vitest version: v0.33.0
```

### Used Package Manager

pnpm

### Validations

g-plane:

### Tests

### Documentation
No docs changes.

### Changesets

Fixes #3802 .

When using glob, only current working directory should be considered; otherwise, parent directories named with `.cache` or something matched the default exclude patterns will be ignored, which causes the whole project gets ignored.


antfu (approved):

sheremet-va (approved):

g-plane:
Can we re-run CI on Windows? It failed but didn't provide logs.

g-plane:
Two checks failed but they aren't related to this PR. What should I do?

sheremet-va (approved):

PR #3952, opened 2023-08-14
fix: don't mock css-module if inline is specified

thebanjomatic:
### Description

As per the vite docs here: https://vitejs.dev/guide/features.html#disabling-css-injection-into-the-page

Using the 'inline' query parameter for CSS has different semantics than regular css imports and the expected return value should be a string as the default export.

### Tests

### Documentation

### Changesets

Fixes: #3954

### DIFF
--- a/packages/vitest/src/node/plugins/cssEnabler.ts
+++ b/packages/vitest/src/node/plugins/cssEnabler.ts
@@ -16,6 +16,17 @@ function isCSSModule(id: string) {
   return cssModuleRE.test(id)
 }
 
+// inline css requests are expected to just return the
+// string content directly and not the proxy module
+function isInline(id: string) {
+  const queryStart = id.indexOf('?');
+  if (queryStart === -1) {
+    return false;
+  }
+  const queryParts = id.substring(queryStart + 1).split('&');
+  return queryParts.some(part => part === 'inline');
+}
+
 function getCSSModuleProxyReturn(strategy: CSSModuleScopeStrategy, filename: string) {
   if (strategy === 'non-scoped')
     return 'style'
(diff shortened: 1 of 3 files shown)


test/css/testing.mjs
```diff
@@ -4,6 +4,7 @@ const configs = [
   ['test/default-css', {}],
   ['test/process-css', { include: [/App\.css/] }],
   ['test/process-module', { include: [/App\.module\.css/] }],
+  ['test/process-inline', { include: [/App\.module\.css/] }],
```
sheremet-va, line 7: These have the same config so it's better to allow passing an array as the first argument I think

packages/vitest/src/node/plugins/cssEnabler.ts
```diff
@@ -55,7 +66,7 @@ export function CSSEnablerPlugin(ctx: { config: ResolvedConfig }): VitePlugin[]
         if (!isCSS(id) || shouldProcessCSS(id))
           return
 
-        if (isCSSModule(id)) {
+        if (isCSSModule(id) && !isInline(id)) {
```
sheremet-va, line 69: Just `/[?&]inline/` should be enough I think

sheremet-va (changes requested):
Looks good. Thank you for PR 👍🏻

packages/vitest/src/node/plugins/cssEnabler.ts
sheremet-va: Sure

sheremet-va (changes requested):
Looks good to me 👍🏻 Just a minor issue.

PR #3961, opened 2023-08-15
feat!(snapshot): make snapshots more visually pleasing

sheremet-va:
### Description

This changes the output for snapshots:

1. Quotes are no longer escaped:

```diff
expect({ foo: 'bar' }).toMatchInlineSnapshot(`
  Object {
-    \\"foo\\": \\"bar\\",
+    "foo": "bar",
  }
`)
```

2. One-line snapshots now use "`" quotes instead of ':

```diff
- expect('some string').toMatchInlineSnapshot('"some string"')
+ expect('some string').toMatchInlineSnapshot(`"some string"`)
```

This is a breaking change, but it's long overdue. If you prefer the old behavior or want to have more time to migrate, you can keep escaping with this config:

```ts
export default defineConfig({
  test: {
    snapshotFormat: {
      escapeString: true,
    },
  },
})
```

### Tests

### Documentation

### Changesets

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -5,6 +5,9 @@
     // prefer global Buffer to not initialize the whole module
     "n/prefer-global/buffer": "off",
     "@typescript-eslint/no-invalid-this": "off",
+    "@typescript-eslint/quotes": ["error", "single", {
+      "allowTemplateLiterals": true
+    }],
     "no-restricted-imports": [
       "error",
       {
--- a/examples/basic/test/__snapshots__/suite.test.ts.snap
+++ b/examples/basic/test/__snapshots__/suite.test.ts.snap
@@ -1,4 +1,4 @@
-// Vitest Snapshot v1
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
 exports[`suite name > snapshot 1`] = `
 {
--- a/examples/mocks/test/error-mock.spec.ts
+++ b/examples/mocks/test/error-mock.spec.ts
@@ -4,5 +4,5 @@ vi.mock('../src/default', () => {
 
 test('when using top level variable, gives helpful message', async () => {
   await expect(() => import('../src/default').then(m => m.default)).rejects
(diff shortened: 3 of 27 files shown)


silverwind:
Shouldn't the snapshot version have been bumped to v2 with this? I mean this header:

```js
// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
```

sheremet-va:
> Shouldn't the snapshot version have been bumped to v2 with this? I mean this header:
> 
> 
> 
> ```js
> 
> // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
> 
> ```
> 
> 

Snapshot API wasn't changed, so no. Only default values are changed.

silverwind:
So the snapshot format has no real version? Why even write it to the files?

sheremet-va:
> So the snapshot format has no real version? Why even write it to the files?

Yes, it doesn't have a version. Any change is a breaking change in Vitest. The version here is in case the file format changes (exports -> _exports) where the previous parser would fail.

agriffis:
Hi! Is it possible that this change broken my snapshots with nested JSON?

I have a snapshot file that looked like this before updating vitest. Note that the snapshotted object has a JSON string which itself is an object:

```js
// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html

exports[`getPalettes full 1`] = `
[
  {
    "blocks": [
      {
        "defaultConfigJson": "{
  \\"FREEMARKER\\": {
    \\"background.color.color\\": \\"text\\",
    \\"background.color.custom\\": \\"text\\",
  }
}",
      },
    ],
  },
]`
```

I just upgraded vitest and now my nested JSON no longer has escapes:

```js
// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html

exports[`getPalettes full 1`] = `
[
  {
    "blocks": [
      {
        "defaultConfigJson": "{
  "FREEMARKER": {
    "background.color.color": "text",
    "background.color.custom": "text",
  }
}",
      },
    ],
  },
]`
```

Is this part of the change intentional?

PR #3971, opened 2023-08-17
fix: should remove mockPath from callstack whether success or failed

miserylee:
### Description

When mock a module import exception, the mock function throws but the mockPath not removed from the callstack  
this makes the retry import  not perform correctly

### Tests

### Documentation

### Changesets


sheremet-va:
Looks good, but can you add a test case where this fails without your change? You can add it to examples/mocks.

examples/mocks/src/retry-dynamic-import.ts
```diff
@@ -0,0 +1,17 @@
+export async function retryDynamicImport() {
+  let retryTimes = 0
+  const load = async () => {
+    try { // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-expect-error
```
sheremet-va, line 5: Why do you need this here?

sheremet-va (changes requested):

examples/mocks/src/retry-dynamic-import.ts
miserylee: This case is for dynamic import in browser, bundled by webpack.  
The modules dynamically imported are from remote over network, so it maybe failed cause network exception or any unknown reason.  
We want to retry import several times for reducing the final failed possibility.

miserylee: Emmm , did you mean the comments upon?  
Without the comments, the ts language service will emit an error diagnostic message as below 
image

miserylee: I have appended a commit, changed the import path to have extension, to avoid this ts check issue.

sheremet-va (approved):

Issue #3941, opened 2023-08-13
toThrow does not show a diff with actual result


FeldrinH:
### Describe the bug

Using `expect(..).toThrow(Error)` the error shown on failure is simply `expected error to be instance of Error`. This contains very little useful info because it does not say what the actual error was or if an error was thrown at all.

I would have expected `toThrow` to have an output similar to `to.throw`, which provides a very detailed diff of the expected and actual errors.

### Reproduction

```javascript
test('simplerror', async () => {
    expect(async () => {}).toThrow(Error)
});
```
outputs
```
AssertionError: expected error to be instance of Error
...
```
instead of a detailed diff like `to.error` does.

### System Info

```shell
  System:
    OS: Windows 10 10.0.19045
    CPU: (20) x64 13th Gen Intel(R) Core(TM) i5-13600KF
    Memory: 19.28 GB / 31.83 GB
  Binaries:
    Node: 18.15.0 - C:\Program Files\nodejs\node.EXE
    npm: 9.5.0 - C:\Program Files\nodejs\npm.CMD
    pnpm: 8.2.0 - ~\AppData\Local\pnpm\pnpm.EXE
  npmPackages:
    vitest: ^0.34.1 => 0.34.1
  Browsers:
    N/A
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #3979, opened 2023-08-18
fix(expect): improve the error message when nothing is thrown when ing `toThrow`

Dunqing:
### Description

fix: #3941 
image

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/expect/src/jest-expect.ts
+++ b/packages/expect/src/jest-expect.ts
@@ -529,7 +529,7 @@ export const JestChaiExpect: ChaiPlugin = (chai, utils) => {
       }
 
       if (!isThrow && !isNot) {
-        const message = utils.flag(this, 'message') || 'expected function to throw an error, but it didn`t'
+        const message = utils.flag(this, 'message') || 'expected function to throw an error, but it didn\'t'
         const error = {
           showDiff: false,
         }
--- a/packages/expect/src/jest-expect.ts
+++ b/packages/expect/src/jest-expect.ts
@@ -499,6 +499,7 @@ export const JestChaiExpect: ChaiPlugin = (chai, utils) => {
     const promise = utils.flag(this, 'promise')
     const isNot = utils.flag(this, 'negate') as boolean
     let thrown: any = null
+
     if (promise === 'rejects') {
       thrown = obj
     }
--- a/packages/expect/src/jest-expect.ts
+++ b/packages/expect/src/jest-expect.ts
@@ -499,7 +499,6 @@ export const JestChaiExpect: ChaiPlugin = (chai, utils) => {
     const promise = utils.flag(this, 'promise')
     const isNot = utils.flag(this, 'negate') as boolean
     let thrown: any = null
-
     if (promise === 'rejects') {
       thrown = obj
     }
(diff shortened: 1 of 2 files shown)


sheremet-va (changes requested):
You checked that you added a test, but I don't see any tests.

Dunqing:
> You checked a checkbox that you added a test, but I don't see any tests.

Sorry, test case added

packages/expect/src/jest-expect.ts
```diff
@@ -519,12 +519,22 @@ export const JestChaiExpect: ChaiPlugin = (chai, utils) => {
...
       try {
         obj()
       }
       catch (err) {
+        isThrow = true
         thrown = err
       }
+
+      if (!isThrow && !isNot) {
+        const message = utils.flag(this, 'message') || 'expected function to throw an error, but it didn`t'
```
sheremet-va, line 532:
```suggestion
        const message = utils.flag(this, 'message') || 'expected function to throw an error, but it didn\'t'
```

test/core/test/jest-expect.test.ts
```diff
@@ -332,6 +332,13 @@ describe('jest-expect', () => {
     expect(null).toBe(true)
   })
 
+  it.fails('toThrow should failure when does not throw ', () => {
+    expect(() => {
+    }).toThrow(Error)
+    expect(async () => {
```
sheremet-va, line 338: This line will never run. Either make 2 separate tests with `.fails` or wrap it inside another `expect(() => {}).toThrow()` and remove `.fails`

sheremet-va (changes requested):

test/core/test/jest-expect.test.ts
```diff
@@ -375,6 +368,18 @@ describe('jest-expect', () => {
       },
     ])
   })
+
+  it('toThrow didn\'t throw', () => {
+    expect(() => {
+      expect(async () => {
+      }).toThrow(Error)
```
sheremet-va, line 375: Async error should be with `.rejects`:

```ts
await expect(async () => {}).rejects.toThrow(Error)
```

sheremet-va (changes requested):

test/core/test/jest-expect.test.ts
Dunqing: The results are consistent with `jest`
image

sheremet-va: Yes, but it doesn't mean that code is correct. Function didn't finish running yet

sheremet-va: `toThrow` receives a Promise, it doesn't wait it to resolve to _ACTUALLY_ check that error didn't happen

If you throw an error there, you will get unhandled promise error, not `toThrow` error

Dunqing: This test case actually comes from [here](https://github.com/vitest-dev/vitest/issues/3941), which is exactly what I wanted to test

Dunqing: > Async error should be with `.rejects`:
> 
> ```ts
> await expect(async () => {}).rejects.toThrow(Error)
> ```

I got an error
image

sheremet-va: > This test case actually comes from https://github.com/vitest-dev/vitest/issues/3941, which is exactly what I wanted to test

Well, then your error description is not enough "but it didn't threw" doesn't really tell much. In the current example there are two errors: 1) async wasn't awaited, 2) error wasn't thrown

> I got an error

Please, read how to test asynchronous code: https://vitest.dev/api/expect.html#resolves

PR #3982, opened 2023-08-18
fix(vite-node): mjs files watch not work

Dunqing:

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vite-node/src/cli.ts
+++ b/packages/vite-node/src/cli.ts
@@ -1,3 +1,4 @@
+import { isAbsolute, join } from 'node:path'
 import cac from 'cac'
 import c from 'picocolors'
 import { createServer } from 'vite'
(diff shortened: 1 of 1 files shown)


packages/vite-node/src/cli.ts
```diff
@@ -74,11 +75,21 @@ async function run(files: string[], options: CliOptions = {}) {
     cli.version(version).outputHelp()
     process.exit(1)
   }
+  const root = options.root || process.cwd()
 
   const serverOptions = options.options
     ? parseServerOptions(options.options)
     : {}
 
+  serverOptions.deps ??= {}
```
sheremet-va, line 84: This will only work for an entry file. Here is how it's done in Vitest when using vite-node:

https://github.com/vitest-dev/vitest/blob/41e828f59c63d39e7b40ace0d226b37b0a1e3d52/packages/vitest/src/node/config.ts#L17

sheremet-va (changes requested):

packages/vite-node/src/cli.ts
Dunqing: So I should handle this in the vite node server? I'm not sure if my solution is correct

sheremet-va: Yes, you should handle it in vite node server

sheremet-va (approved):

Dunqing:
Maybe this will improve the performance of testing esm

sheremet-va (approved):

PR #3983, opened 2023-08-18
fix(vitest): Define nested CAC options

### LINKED ISSUE #3962: command line flags don't override config file
### Describe the bug

As a tool user, I usually have a set of expectations respect to the precedence order of certain ways to set configuration values for the tools I use. One of these typical rules is that the order of application is similar to:

1. config files
2. environment variables
3. command line options/flags

where the later overrides the former (in case it is present).

In the case of Vitest, I'm finding that this is not the case when it comes to coverage related settings. For example, when I have the following config file:
```typescript

export default defineConfig({
	test: {
		coverage: {
			include: ['src/*.{ts,tsx}'],
			exclude: ['src/**/*.test.{ts,tsx}', 'src/stories/**/*'],
			thresholdAutoUpdate: true,
			branches: 50,
			lines: 50,
			functions: 50,
			statements: 50,
		},
	},
})
```

but I call `vitest` with `vitest run --coverage.enabled=true --coverage.thresholdAutoUpdate=false`, the later flag is not being applied, giving more importance to the one inside the config file (and as a consequence, if my coverage is higher, the config file is auto-updated to reflect the new coverage values).

### Reproduction

The previous description is enough. We just need a test with high enough coverage, and set the thresholds in our config to be below that if we want to see how the command line flag is not taking precedence over the config file value.

### System Info

```shell
System:
    OS: macOS 13.5
    Memory: 14.52 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.1.0 - ~/Library/Caches/fnm_multishells/14573_1692095309449/bin/node
    Yarn: 3.6.1 - ~/Library/Caches/fnm_multishells/14573_1692095309449/bin/yarn
    npm: 9.6.4 - ~/Library/Caches/fnm_multishells/14573_1692095309449/bin/npm
    pnpm: 8.6.12 - ~/Library/Caches/fnm_multishells/14573_1692095309449/bin/pnpm
  Browsers:
    Chrome: 115.0.5790.170
    Edge: 114.0.1823.82
    Safari: 16.6
  npmPackages:
    @vitest/coverage-v8: ^0.34.1 => 0.34.1 
    vitest: ^0.34.1 => 0.34.1
```

### Used Package Manager

pnpm

### Validations

LorenzoBloedow:
I thought of different ways of addressing this but after realizing `cac` doesn't provide a native solution to transform option values with dot notation I thought the best way was to convert it to a boolean only if the string is either `"true"` or `"false"` and just ignore it if it's something different since Vitest doesn't seem to do runtime type checking.

I also thought about implementing this at the provider level but it'd mean the code would have to be duplicated for both v8 and istanbul.

It also doesn't seem like something the provider should be concerned with.

Regarding the tests, I wasn't sure how to proceed since currently there aren't any tests to differentiate between CLI and config file arguments so I'm not sure it's something the Vitest team wants.

I also think this is a bigger problem that includes other dot-notation based options like `coverage.enabled` and I'd like to hear your guys' opinions on this because it doesn't seem to be something `cac` can currently solve natively.

StackBlitz demonstrating that this fixes #3962 by using my own fork of this pull request: https://stackblitz.com/edit/vitest-dev-vitest-ecfukt


packages/vitest/src/node/cli.ts
```diff
@@ -161,6 +161,12 @@ function normalizeCliOptions(argv: CliOptions): CliOptions {
 
     if ((coverage as CoverageIstanbulOptions).ignoreClassMethods)
       (coverage as CoverageIstanbulOptions).ignoreClassMethods = toArray((coverage as CoverageIstanbulOptions).ignoreClassMethods)
+
+    const thresholdAutoUpdate = (coverage as CoverageIstanbulOptions).thresholdAutoUpdate
+    if ((thresholdAutoUpdate as unknown as string) === 'true')
+      (coverage as CoverageIstanbulOptions).thresholdAutoUpdate = true
+    else if ((thresholdAutoUpdate as unknown as string) === 'false')
+      (coverage as CoverageIstanbulOptions).thresholdAutoUpdate = false
```
AriPerkkio, line 169: Would this be easier to read and have less type assertions?

```suggestion
    if ('thresholdAutoUpdate' in coverage) {
      // Convert string values ("true"|"false") to boolean
      if (typeof coverage.thresholdAutoUpdate === 'string') {
        const value = coverage.thresholdAutoUpdate === 'false' ? false : coverage.thresholdAutoUpdate

        coverage.thresholdAutoUpdate = value
      }
    }
```

AriPerkkio (changes requested):
I wonder if there's a better way to do these using `cac` directly. Could it be configured to understand the object structures?

> Regarding the tests, I wasn't sure how to proceed since currently there aren't any tests to differentiate between CLI and config file arguments so I'm not sure it's something the Vitest team wants.

We have some CLI flag tests at `test/config` but I'm not sure how this could be implemented there. We would need a way to print the resolved values of the test case and compare those depending on what CLI/config values were passed.

packages/vitest/src/node/cli.ts
LorenzoBloedow: I really like the idea but the TypeScript interface suggests Vitest expects a boolean, in the code you suggested, if `thresholdAutoUpdate` is `"true"` it will pass as a string, this could lead to bugs in the future if for some reason we check for exactly `true` instead of just a truthy value.

I have a suggestion that builds on yours:
```suggestion
    if ('thresholdAutoUpdate' in coverage) {
      // Convert string values ('true'|'false') to boolean
      if (typeof coverage.thresholdAutoUpdate === 'string') {
        const value = coverage.thresholdAutoUpdate === 'false'
        ? false
        : coverage.thresholdAutoUpdate === 'true'
        ? true
        : coverage.thresholdAutoUpdate
        
        coverage.thresholdAutoUpdate = value
      }
    }
```
This way the only reason it would pass as anything other than a boolean is due to a user error like using a number.
I also changed the double quotes to single quotes in the comment just because that seems to be what the codebase uses.

What do you think?

LorenzoBloedow:
> I wonder if there's a better way to do these using `cac` directly. Could it be configured to understand the object structures?
I'm not sure but I'll dig through the source later to find out, the docs seem to be lacking on that regard.

If I'm not mistaken, I think just doing `coverage.thresholdAutoUpdate` instead of `coverage.thresholdAutoUpdate=true` makes `thresholdAutoUpdate` a boolean, but I'm not sure and also [the docs currently suggest using true in dot notation].(https://vitest.dev/guide/coverage.html#other-options:~:text=enable-,coverage.enabled%3Dtrue,-in%20your%20configuration)
And even if it does, I'm also not sure if there's a way to pass `false` or even if it's a good idea to ignore the fact that people may not use the correct notation.

If `cac` can't do this natively I was thinking of forking it and adding that functionality so we can use it, what do you think?

packages/vitest/src/node/cli.ts
AriPerkkio: Looks good 👍

LorenzoBloedow:
@AriPerkkio Ok so I found out what the problem actually was.
`mri` (and by extension, `cac`) already does this conversion for us.

I haven't actually confirmed this but after digging through the source it seems `cac` [will only parse options with dot notation properly if they're explicitly defined](https://github.com/cacjs/cac/blob/8342919821fbffa733c6ab9558f4d60fc43f9ff0/src/utils.ts#L113), then `cac` will use `mri` to parse the option [which seems to do the string to boolean conversion for us](https://github.com/lukeed/mri/blob/e73e9f9d5b02124d14ac17dac2c4801687d3e99a/src/index.js#L9C3-L9C134).

Because of this, it seems this also affects any other property that uses dot notation and is not explicitly defined.
Should we turn this into a wider PR and define dot-notation based properties for all options?

AriPerkkio:
It's displayed correctly.

```
> vitest "--help"

Options:
  ...
  --coverage                             Enable coverage report 
  --coverage.thresholdAutoUpdate         Update threshold values: "lines", "functions", "branches" and "statements" to configuration file when current coverage is above the configured thresholds 
  ...
```

If listing all possible options makes `--help` too verbose, I think `cli.help(callback)` could be used to filter the message. https://github.com/cacjs/cac#clihelpcallback

LorenzoBloedow:
> If listing all possible options makes `--help` too verbose, I think `cli.help(callback)` could be used to filter the message. https://github.com/cacjs/cac#clihelpcallback

So should I filter dot-notation options?

LorenzoBloedow:
> Here's test case that you can use to validate these.

Shouldn't this be a concern for `cac`?
I think the most we should do is write snapshot tests for the shape of the CLI (is that already being done?), what do you think?

AriPerkkio:
> So should I filter dot-notation options?

Sure, we can then check if enabling them doesn't cause too verbose `--help` message.

> Shouldn't this be a concern for `cac`?
> I think the most we should do is write snapshot tests for the shape of the CLI 

Feel free to implement test cases as you see best. The example shown above is printing Vitest's resolved configuration values, which means it will read CLI and configuration files and resolve final configuration object.

LorenzoBloedow:
> Sure, we can then check if enabling them doesn't cause too verbose `--help` message.

I think I did all the nested options, I changed the description on some of them where it makes sense to have a more compact description or a description with a link since CLIs don't have anchor tags.

I also thought about what you said regarding the verbose output so I went ahead and omitted the nested options by default but enabled them when the `VERBOSE` env variable is set.
I used an environment variable because I couldn't find a native way to add a `verbose` option to `cac`.
Maybe I could've added another command like `verboseHelp` but I thought this way made more sense because it adds less complexity, is more DRY and less prone to bugs, since I still use a callback from `cac`.

Let me know once you guys think this looks good so I can start writing the tests and updating the docs.

AriPerkkio:
Filtering options from `--help` seems to work nice. When `VERBOSE` env is enabled, the output gets really verbose and reduces readability. I think it's OK to hide these new options completely from `--help`.

For the implementation I think we should structure the `cac` configuring so that options shown in `--help` are listed first, and the options excluded from `--help` are below.

```ts
cli
  .version(version)
  
  // Options shown in --help
  .option('-r, --root <path>', 'Root path')
  .option('-c, --config <path>', 'Path to config file')
  .option('--coverage', 'Enable coverage report')
  ...

  // Options hidden from --help. These are defined for type resolving.
  .option('--coverage.provider [name]', 'hide-from-help')
  .option('--coverage.enabled', 'hide-from-help')
  ...
```

Then in the `.help(callback)` it's easy to filter out by looking for `'hide-from-help'`.

LorenzoBloedow:
> Filtering options from `--help` seems to work nice. When `VERBOSE` env is enabled, the output gets really verbose and reduces readability. I think it's OK to hide these new options completely from `--help`.

I disagree, when using the verbose option in any program, the user expects the output to be exactly that; verbose.
Since a "verbose level" system would be too complex, I placed a link to the docs where printing the whole explanation would otherwise become too verbose.

I think the combination of the links and disabling the verbose output by default, such that it won't ever get in the way of the user unless they explicitly choose to enable it, makes for very sensible defaults.

> For the implementation I think we should structure the cac configuring so that options shown in --help are listed first, and the options excluded from --help are below.

I agree since it's probably easier to visually scan for the option you're looking for.

> Then in the .help(callback) it's easy to filter out by looking for 'hide-from-help'.

This is a good idea if we do decide to go with the route of completely disabling the nested options from `--help`.

What do you think?

AriPerkkio:
Is `env.VERBOSE` a convention across CLI programs? I haven't run into it before. 

Instead of adding a verbose mode for `--help` it would be better to have option specific `--help` for options that support dot notation, e.g. `--help --coverage` to see all sub options of `--coverage`. I'm not sure if `cac` supports this.

LorenzoBloedow:
> Is `env.VERBOSE` a convention across CLI programs? I haven't run into it before.

Yeah I've never run into a CLI where `VERBOSE` would apply only to the `--help` command, using an environment variable might not be the best way.

> Instead of adding a verbose mode for `--help` it would be better to have option specific `--help` for options that support dot notation, e.g. `--help --coverage` to see all sub options of `--coverage`. I'm not sure if `cac` supports this.

This is why I used an environment variable, from what I could gather on the docs and my own testing, `cac` searches for `--help` and prints the help output, it doesn't care what other options you put before and after it.

I think your idea of using `--help --coverage` would be the best way of dealing with this.
Unfortunately, `cac` doesn't seem to be actively maintained anymore, would a fork be appropriate? If so, I'm happy to start working on it!

AriPerkkio:
Here's minimal example for filtering `--help --<command>`. I think this would be better than forking whole project.

<details>

```ts
import { cac } from "cac";

const cli = cac("repro");

cli
  .version("1.2.3")
  .option("-r, --root <path>", "Root path")
  .option("-u, --update", "Update snapshot")
  .option(
    "--coverage",
    "Enable coverage report. Use --help --coverage to see more options"
  )
  .option("--coverage.reporter [name]", "Choose coverage reporter")
  .help((info) => {
    const options = info.find((current) => current.title === "Options");
    if (typeof options !== "object") return info;

    const helpIndex = process.argv.findIndex((arg) => arg === "--help");
    const subCommand = process.argv[helpIndex + 1];

    // Filter out options that are not related to the sub command
    if (subCommand && subCommand.startsWith("--")) {
      options.body = options.body
        .split("\n")
        .filter((line) => line.trim().startsWith(subCommand))
        .join("\n");
    }
    // No sub-command, filter out options with dot-notation
    else {
      options.body = options.body
        .split("\n")
        .filter((line) => /--\S+\./.test(line) === false)
        .join("\n");
    }

    return info;
  });

cli.parse();
```

</details>

```
ari ~/repro  $ node cac.mjs --help 
repro/1.2.3

Usage:
  $ repro <command> [options]

Options:
  -v, --version               Display version number 
  -r, --root <path>           Root path 
  -u, --update                Update snapshot 
  --coverage                  Enable coverage report. Use --help --coverage to see more options 
  -h, --help                  Display this message 

ari ~/repro  $ node cac.mjs --help --coverage
repro/1.2.3

Usage:
  $ repro <command> [options]

Options:
  --coverage                  Enable coverage report. Use --help --coverage to see more options 
  --coverage.reporter [name]  Choose coverage reporter 
```

LorenzoBloedow:
> Here's minimal example for filtering --help --<command>. I think this would be better than forking whole project.

Thanks for the great example! And while it's unfortunate we have to directly use `process.argv` (because it feels like something `cac` should be handling) I agree it's much better than forking the entire project.

I built on your example and added a few things (I already pushed them so you can check it out):

- Added support for providing multiple subcommands instead of just one
- Original help order is preserved even when using multiple subcommands in a random order
- You can use the `--verbose` option to display everything at once but it must be the only subcommand used
- Behavior that may be confusing to the user such as printing the whole help output when only invalid subcommands are found (imagine a typo) instead of printing nothing is shown with a warning explaining what happened

I may have gone a little overboard haha. Anyway, let me know your thoughts!

sheremet-va:
> * You can use the `--verbose` option to display everything at once but it must be the only subcommand used

I don't like using `--verbose` for this purpose. It's too ambiguous - vitest has a reporter with the same name for example. And jest also has this flag that shows more information about tests.

LorenzoBloedow:
> > * You can use the `--verbose` option to display everything at once but it must be the only subcommand used
> 
> I don't like using `--verbose` for this purpose. It's too ambiguous - vitest has a reporter with the same name for example. And jest also has this flag that shows more information about tests.

What do you suggest? Maybe `--show-all`? Or are you saying to remove this option completely?

sheremet-va:
> Maybe `--show-all`?

Maybe `--expand-help`?

LorenzoBloedow:
> Maybe `--expand-help`?

Done!


Sorry for the delay and the force push, I messed up the git history and had to force push otherwise I would commit like 700 different changes!

Do we just need the tests now?

sheremet-va:
> Do we just need the tests now?

Ideally, yes. And mention this in docs.

I also see that browser tests are failing with this change - are you sure all CLI arguments are correct?

LorenzoBloedow:
> Ideally, yes. And mention this in docs.

Done.

> I also see that browser tests are failing with this change - are you sure all CLI arguments are correct?

I forgot to put `<name>` in the `--browser.name` option, so that's fixed. But now one test failed and it seems like it might be a one-off issue, can we rerun the CI?

packages/vitest/src/node/cli.ts
```diff
@@ -20,20 +21,54 @@ cli
...
+  .option('--coverage.thresholdAutoUpdate', 'Update threshold values: "lines", "functions", "branches" and "statements" to configuration file when current coverage is above the configured thresholds. This option is not available for custom providers (default: false)')
+  .option('--coverage.lines', 'Threshold for lines. Visit https://github.com/istanbuljs/nyc#coverage-thresholds for more information. This option is not available for custom providers')
+  .option('--coverage.functions', 'Threshold for functions. Visit https://github.com/istanbuljs/nyc#coverage-thresholds for more information. This option is not available for custom providers')
+  .option('--coverage.branches', 'Threshold for branches. Visit https://github.com/istanbuljs/nyc#coverage-thresholds for more information. This option is not available for custom providers')
+  .option('--coverage.statements', 'Threshold for statements. Visit https://github.com/istanbuljs/nyc#coverage-thresholds for more information. This option is not available for custom providers')
+  .option('--coverage.100', 'Shortcut to set all coverage thresholds to 100. This option is only available for the v8 provider (default: false)')
```
AriPerkkio, line 53: There options were changed in https://github.com/vitest-dev/vitest/pull/4442. It also affects the title of this PR.

LorenzoBloedow: Are they good now?

PR #3996, opened 2023-08-21
fix(vitest): headless mode not working when using webdriverio (fix #3930)

### LINKED ISSUE #3930: Headless mode is not working for WebDriverIO
### Describe the bug

The current [code for setting the headless option for WebDriverIO is not working](https://github.com/vitest-dev/vitest/blob/f4e6e99fa3a81fd7a97fb9b435da367dddba7fa4/packages/vitest/src/node/browser/webdriver.ts#L61), i.e. the option is ignored and browser still opens:

```
      capabilities: {
        'browserName': this.browser,
        'wdio:devtoolsOptions': { headless: options.headless },
      },
```

Based on [current WebDriverIO documentation](https://webdriver.io/docs/capabilities/#run-browser-headless), it is now dependent on the browser selected. For example, for Chrome, it should be 

```
'goog:chromeOptions': {
    args: ['headless', 'disable-gpu']
}
```

I verified by modifying the code in the node_module/vitest/dist/vendor-node.caa511fc.js of my project and it works after the change.

### Reproduction

None

### System Info

```shell
System:
    OS: macOS 13.4.1
    Memory: 45.28 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.16.1 - /usr/local/opt/node@18/bin/node
    npm: 9.5.1 - /usr/local/opt/node@18/bin/npm
  Browsers:
    Chrome: 115.0.5790.170
    Edge: 115.0.1901.203
    Safari: 16.5.2
  npmPackages:
    @vitejs/plugin-react: ^4.0.1 => 4.0.3
    @vitest/browser: ^0.34.1 => 0.34.1
    @vitest/coverage-istanbul: ^0.34.1 => 0.34.1
    @vitest/ui: ^0.34.1 => 0.34.1
    vite: ^4.4.0 => 4.4.7
    vitest: ^0.34.1 => 0.34.1
```

### Used Package Manager

npm

### Validations

LorenzoBloedow:
I skimmed through the v6, v7, and v8 docs for `webdriverio` but I didn't really find any documentation for the `wdio:devtoolsOptions` key on the current code causing this issue:
```
capabilities: {
  'browserName': this.browser,
  'wdio:devtoolsOptions': { headless: options.headless },
},
```

**IMPORTANT:**
I couldn't reproduce this issue, I realize this was my mistake and I should've tried that before starting to write the code.
I have a suggestion regarding reproductions (let me know if I should open a discussion for this): they should be made mandatory and the issue should be automatically (and/or manually) closed until a repro that actually works is provided, otherwise false-positives made by user error may become very common, but that's just my take on it, let me know what you think!

This fixes #3930 (if it's confirmed to be an issue)


sheremet-va:
Please, don't remove PR checklist when creating PR. One of the points there is to not push pnpm-lockfile changes.

LorenzoBloedow:
> Please, don't remove PR checklist when creating PR. One of the points there is to not push pnpm-lockfile changes.

Sorry, I should've read it.
I've removed `pnpm-lockfile.yaml` now

LorenzoBloedow:
> > Please, don't remove PR checklist when creating PR. One of the points there is to not push pnpm-lockfile changes.
> 
> Sorry, I should've read it.
> I've removed `pnpm-lockfile.yaml` now

Just noticed I mistakenly ran `git rm` instead of checking out to the unchanged file 🤦

I'll revert this soon

sheremet-va:
I just noticed that `remote` accepts "headless" option. Can't we just pass it down?

LorenzoBloedow:
> I just noticed that `remote` accepts "headless" option. Can't we just pass it down?

webdriverio doesn't seem to have complete documentation for the options for `remote` on their website.
But I inspected the declaration file with the `headless` option and this is what I found:
```
/**
* Sauce Labs provides a headless offering that allows you to run Chrome and Firefox tests headless.
*/
headless?: boolean;
```
I thought it was weird so I decided to test it myself, and yeah I can confirm it doesn't work (at least without Sauce Labs)

sheremet-va:
> I thought it was weird so I decided to test it myself, and yeah I can confirm it doesn't work (at least without Sauce Labs)

Ok, no problem 👍🏻

LorenzoBloedow:
@sheremet-va I'm not sure how to proceed since we're getting the `ERR_PNPM_OUTDATED_LOCKFILE` error on CI and I'm not supposed to upload the lock file.

sheremet-va:
Looks like it's still broken

LorenzoBloedow:
> Looks like it's still broken

I'm having some trouble trying to update only `@wdio/types`.
I tried using many different command and version combinations including `pnpm add --filter ./vitest --prefer-offline --lockfile-only @wdio/types` (cwd: `vitest/packages`) but I always get a lot of unrelated packages updated as well (like `/@babel/helper-create-class-features-plugin@7.22.11(@babel/core@7.22.11)`.

I also [found a recent issue](https://github.com/pnpm/pnpm/issues/6985) from someone with the same problem as me.

sheremet-va:
> > Looks like it's still broken
> 
> 
> 
> I'm having some trouble trying to update only `@wdio/types`.
> 
> I tried using many different command and version combinations including `pnpm add --filter ./vitest --prefer-offline --lockfile-only @wdio/types` (cwd: `vitest/packages`) but I always get a lot of unrelated packages updated as well (like `/@babel/helper-create-class-features-plugin@7.22.11(@babel/core@7.22.11)`.
> 
> 
> 
> I also [found a recent issue](https://github.com/pnpm/pnpm/issues/6985) from someone with the same problem as me.

The jus updating it should be fine I think

LorenzoBloedow:
> The jus updating it should be fine I think

Not sure what you meant but I'm assuming you asked me to just run the previous install command and commit so that's what I did (it does look like there are some unrelated packages being updated in the lockfile though)

pnpm-lock.yaml
```diff
@@ -1314,6 +1314,9 @@ importers:
       '@vitest/utils':
         specifier: workspace:*
         version: link:../utils
+      '@wdio/types':
+        specifier: ^8.15.10
+        version: 8.15.10
```
AriPerkkio, line 1319: > it does look like there are some unrelated packages being updated in the lockfile though

As you are updating only the `@wdio/types` in `package.json`, you should not commit unrelated package updates in `package-lock.json`. Pick only the ones that are related to `@wdio/types` and revert the other lines. 
Once all intentional changes of `package-lock.json` are staged and unrelated ones are reverted, you can use `pnpm i --frozen-lockfile` to see if `package-lock.json` is valid. It will throw error if something is missing.

LorenzoBloedow: Sorry, I've never had to manually update a lockfile before. I think I got it now though

LorenzoBloedow:
Fixed the types and ran all CI commands on my machine, not sure why `test:ci` threw in the last run though, works fine here.

sheremet-va (approved):

sheremet-va:
I don't understand why your CI fails when main doesn't 🤔

LorenzoBloedow:
> I don't understand why your CI fails when main doesn't 🤔

The errors are weird because they don't seem to have much to do with what I think caused it, but basically I didn't need to add some mapping keys in the lockfile because they were already there.

I thought `pnpm install` would fail when that happens but apparently not.
It should be fixed now since I ran the github actions locally with [act](https://github.com/nektos/act) before and after and the problem seems to be fixed.

sheremet-va:
Browser tests fail now 😄 (Because of the changes here looks like)

LorenzoBloedow:
> Browser tests fail now 😄 (Because of the changes here looks like)

This PR is making me feel silly lol.
I tried to fix it but unfortunately I don't have a Mac and it's failing because of Safari so I can't confirm my changes locally.
It seems like a pretty simple fix though, let's just hope the next CI run is finally the last one! :)

LorenzoBloedow:
This is weird... I ran the failed CI commands on Windows and they all passed, the lockfile looks fine, and everything else hasn't
changed. Not sure how to proceed here, could this possibly be a problem with the CI cache?

packages/vitest/src/node/browser/webdriver.ts
```diff
@@ -53,13 +65,57 @@ export class WebdriverBrowserProvider implements BrowserProvider {
...
+      if (options.headless) {
+        switch (this.browser) {
+          case 'chrome':
+            webdriverBrowserExtensionMap[this.browser].value.args = webdriverBrowserExtensionMap[this.browser].value.args.concat(['headless', 'disable-gpu'])
+            break
+          case 'firefox':
+            webdriverBrowserExtensionMap[this.browser].value.args.push('-headless')
+            break
+          case 'edge':
+            webdriverBrowserExtensionMap[this.browser].value.args.push('--headless')
```
userquin, line 109: my test works with the same configuration for chrome

PR #4009, opened 2023-08-23
Fix: Removing obsolete warning from undesired snapshots

samkevin1:
### Description

This PR removes the `obsolete snapshot` warning from skipped suite tests.

Related to #2442.

### Tests

### Documentation

### Changesets


packages/vitest/src/runtime/runners/test.ts
```diff
@@ -36,6 +36,16 @@ export class VitestTestRunner implements VitestRunner {
   onAfterRunSuite(suite: Suite) {
     if (this.config.logHeapUsage && typeof process !== 'undefined')
       suite.result!.heap = process.memoryUsage().heapUsed
+
+    getTasks(suite).forEach((suiteTask) => {
+      if (suiteTask.mode === 'skip') {
+        getTasks(suiteTask).forEach((task) => {
```
sheremet-va, line 42: I don't think there are any tasks here when suite is skipped. We skip collection phase altogether

sheremet-va (changes requested):

packages/vitest/src/runtime/runners/test.ts
samkevin1: Weird, I was logging the tasks and they were there. The warning is also not happening anymore as reported in the issue.

samkevin1: Yeah, the tasks are there but they have `mode: 'skip'`

sheremet-va: You can see here that we don't collect tests if suite was marked as skipped, so I'm not sure how you tested:

https://github.com/vitest-dev/vitest/blob/00c432fad6c40b4ff21cc45b18e64f29816062db/packages/runner/src/run.ts#L284

sheremet-va (changes requested):
Can you add more tests? Create a fixture with snapshots and a suite with `.skip`. You can use  `test/test-utils/index.ts` to run Vitest on the fixture.

Then check that the output doesn't contain an error that there are obsolete snapshots

samkevin1:
@sheremet-va Added the tests, not sure why the CI is failing though.

test/core/test/snapshot.test.ts
```diff
@@ -112,3 +115,18 @@ test('renders inline mock snapshot', () => {
...
+
+const root = path.resolve(__dirname, './fixtures')
+function run(testNamePattern?: string) {
+  return runVitest({
+    root,
+    testNamePattern,
+  })
+}
+
+describe('snapshot should not be obsolete', () => {
```
sheremet-va, line 127: When I run this test without your changes, it gives the same result. Your tests should validate that the change that you made is working.

If you just run this command in main, you can see that there are already no obsolete warnings. Please, create a test that actually tests things.

sheremet-va (changes requested):

PR #4036, opened 2023-08-28
fix: resolving dep optimizer issues with workspace

### LINKED ISSUE #4020: deps.optimizer appears broken in 0.34.3
### Describe the bug

This is new behavior for me starting in 0.34.3. 0.34.2 works correctly for me.

If I pass dependencies to deps.optimizer.web.includes, it appears to usually work the first time, but then it fails to save the dependencies with the following error:
```
[vite] error while updating dependencies:
Error: ENOENT: no such file or directory, rename 'C:\source\test-project\node_modules\.vitest\deps_temp_c61dd53a' -> 'C:\source\test-project\node_modules\.vitest\deps'
```

And then on subsequent runs, there is a long delay before timing out and producing the following error:

```
file:///C:/source/test-project/node_modules/vite-node/dist/server.mjs:310
        throw new Error(`ViteNodeServer: ${id} not found. This is a bug, please report it.`);
        ^

Error: ViteNodeServer: C:/source/test-project/node_modules/.vitest/deps/vue.js not found. This is a bug, please report it.
    at Timeout._onTimeout (file:///C:/source/test-project/node_modules/vite-node/dist/server.mjs:310:15)
    at listOnTimeout (node:internal/timers:573:17)
    at process.processTimers (node:internal/timers:514:7)
```

### Reproduction

Doesn't actually appear to reproduce in stackblitz, but does reproduce for me on windows:
https://stackblitz.com/edit/vitest-dev-vitest-ycqdpe?file=vite.config.ts

### System Info

```shell
System:
    OS: Windows 10 10.0.22621
    CPU: (20) x64 12th Gen Intel(R) Core(TM) i7-12800H
    Memory: 13.28 GB / 31.64 GB
  Binaries:
    Node: 20.5.1 - ~\AppData\Local\Volta\tools\image\node\20.5.1\node.EXE
    Yarn: 3.6.1 - ~\AppData\Local\Volta\tools\image\yarn\1.22.19\bin\yarn.CMD
    npm: 9.8.0 - ~\AppData\Local\Volta\tools\image\node\20.5.1\npm.CMD
    pnpm: 8.6.12 - ~\AppData\Local\Volta\tools\image\pnpm\8.6.12\bin\pnpm.CMD
  Browsers:
    Edge: Spartan (44.22621.2134.0), Chromium (115.0.1901.200)
  npmPackages:
    @vitejs/plugin-vue2: ^2.2.0 => 2.2.0
    vite: ^4.0.4 => 4.4.4
    vitest: 0.34.3 => 0.34.3
```

### Used Package Manager

yarn

### Validations

thebanjomatic:
### Description

We now use a different cache directory per project in the workspace this way we don't wind up accidentally deleting the cache directory from one project as the next one starts running.

I'm using a hash of the project name as the basis for constructing the cache sub-directory. Using a hash here is primarily done as a way to avoid sanitizing the project name to remove characters that aren't valid directory names and dealing with collisions for similarly named projects that contain invalid characters ("Foo<" and "Foo>" might both sanitize to "Foo_") for example.

A potential downside to using project name (hashed or not) is that workspaces which don't define project names for their projects will fall back to using an index as the project name. As a result, changing the order of the projects in a workspace, can result in the wrong cache directory being used.

In some ways, it might be better to do something like hash the resolved config that way order wouldn't matter, and changing the config would result in a different hash, but I don't necessarily love that approach either since it would result in more directories being created and we don't currently ever clean-up the `node_modules/.vitest` directory to remove stale project dirs.

@sheremet-va please let me know if you have a preference for the above

Fixes #4020 

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/cache/index.ts
+++ b/packages/vitest/src/node/cache/index.ts
@@ -1,4 +1,5 @@
 import fs from 'node:fs'
+import crypto from 'node:crypto'
 import { findUp } from 'find-up'
 import { resolve } from 'pathe'
 import { loadConfigFromFile } from 'vite'
@@ -20,8 +21,11 @@ export class VitestCache {
     return this.stats.getStats(key)
   }
 
-  static resolveCacheDir(root: string, dir: string | undefined) {
-    return resolve(root, slash(dir || 'node_modules/.vitest'))
+  static resolveCacheDir(root: string, dir: string | undefined, projectName: string | undefined) {
+    const baseDir = slash(dir || 'node_modules/.vitest')
+    return projectName
+      ? resolve(root, baseDir, crypto.createHash('md5').update(projectName, 'utf-8').digest('hex'))
+      : resolve(root, baseDir)
   }
 
   static async clearCache(options: CliOptions) {
@@ -38,11 +42,12 @@ export class VitestCache {
       : undefined
 
     const cache = config?.test?.cache
+    const projectName = config?.test?.name
 
     if (cache === false)
       throw new Error('Cache is disabled')
 
-    const cachePath = VitestCache.resolveCacheDir(root, cache?.dir)
+    const cachePath = VitestCache.resolveCacheDir(root, cache?.dir, projectName)
 
     let cleared = false
 
--- a/packages/vitest/src/node/config.ts
+++ b/packages/vitest/src/node/config.ts
@@ -319,7 +319,7 @@ export function resolveConfig(
 
   resolved.cache ??= { dir: '' }
   if (resolved.cache)
-    resolved.cache.dir = VitestCache.resolveCacheDir(resolved.root, resolved.cache.dir)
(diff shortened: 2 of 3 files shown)


packages/vitest/src/node/cache/index.ts
```diff
@@ -2,6 +2,7 @@ import fs from 'node:fs'
 import { findUp } from 'find-up'
 import { resolve } from 'pathe'
 import { loadConfigFromFile } from 'vite'
+import { generateHash } from '@vitest/runner/utils'
```
sheremet-va, line 5: Runner package should not be used in node

sheremet-va (changes requested):
Looks good to me 👍🏻

packages/vitest/src/node/cache/index.ts
thebanjomatic: Ok, providing you are still ok with hashing the project name as a solution, I have switched this to use `node:crypto` to generate an md5 hash

sheremet-va (approved):

PR #4049, opened 2023-08-30
fix(vite-node): check more precisely for root/base paths

danielroe:
### Description

Discovered when debugging nuxt/nuxt#20446, if the root starts with the same path as something being resolved - for example if the root is `/app` and we are resolve the id `/app.vue` then we were not correctly prefixing the root directory to the incoming id.

In actual fact this PR is not required to fix the linked issue, which can be resolved entirely in vite (for which I have opened https://github.com/vitejs/vite/pull/14241). But I do think it is safer to test with trailing slash to ensure that we are indeed talking about files _within_ the directory and not within a sibling directory with a similar name.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vite-node/src/server.ts
+++ b/packages/vite-node/src/server.ts
@@ -251,7 +251,7 @@ export class ViteNodeServer {
         result = await this.server.ssrTransform(result.code, result.map, id)
     }
     else {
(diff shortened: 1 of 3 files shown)


antfu (approved):

sheremet-va:
Thank you! 👍🏻 🙏🏻

PR #4063, opened 2023-09-03
feat(config): add `diff` option

### LINKED ISSUE #2401: Custom snapshot console colors (A11Y)

* Colors work differently in different editors/consoles/terminals/shells/CI's/etc.
* Some people have different visual, or cognitive impairments that make certain color combinations frustrating or difficult to discern.

### Suggested solution

The default red/green approach is fine by me, and probably good for the majority of users. But it would be nice if there was a `vite.config.js` option for customizing the colors used by Vitest, particularly in snapshot failures.

```js
import chalk from 'chalk';

export default defineConfig({
  test: {
    diff: {
      // I'm not sure what the API would look like here, So I'll just use what
      // jest-diff documents for this example. The following is copied directly from
      // https://github.com/facebook/jest/tree/main/packages/jest-diff#options
      aAnnotation: 'Expected',
      aColor: chalk.green,
      aIndicator: '-',
      bAnnotation: 'Received',
      bColor: chalk.red,
      bIndicator: '+',
      changeColor: chalk.inverse,
      changeLineTrailingSpaceColor: string => string,
      commonColor: chalk.dim,
      commonIndicator: ' ',
      commonLineTrailingSpaceColor: string => string,
      compareKeys: undefined,
      contextLines: 5,
      emptyFirstOrLastLinePlaceholder: '',
      expand: true,
      includeChangeCounts: false,
      omitAnnotationLines: false,
      patchColor: chalk.yellow
    }
  }
});
```

By exposing all options, it empowers communities to create their own commonly recommended settings, like so:

```js
import diff from 'vitest-deuteranopia-diffs';

export default defineConfig({ test: { diff } });
```
```js
import diff from 'vitest-cognitive-diffs';

export default defineConfig({ test: { diff } });
```
```js
import diff from 'vitest-win-cmd-diffs';

export default defineConfig({ test: { diff } });
```
etc.
There could even be an `awesome-vitest-diffs` repo with screenshots of each library, so users can more easily choose what works best for them or their teams. This could all be community-lead, but we need Vitest to allow for this.

### Alternative

_No response_

### Additional context

This same feature request was submitted to Jest, and approved for PR.

* https://github.com/facebook/jest/issues/12576

### Validations

fenghan34:
### Description

Closes #2401 

### Tests

### Documentation

### Changesets


packages/runner/src/collect.ts
```diff
@@ -33,6 +33,7 @@ export async function collectTests(paths: string[], runner: VitestRunner): Promi
     try {
       const setupStart = now()
       await runSetupFiles(config, runner)
+      await loadDiffOptionFile(config, runner)
```
fenghan34, line 36: Not sure if here is the right place to load diff config. Could we do this earlier? 🤔

packages/runner/src/types/tasks.ts
```diff
@@ -259,6 +260,11 @@ export interface TestContext {
    */
   task: Readonly<Test>
 
+  /**
+   * Diff options
+   */
+  diff?: DiffOptions
```
sheremet-va, line 266: Why is it on a task?

packages/runner/src/setup.ts
```diff
@@ -15,3 +15,13 @@ export async function runSetupFiles(config: VitestRunnerConfig, runner: VitestRu
...
 }
+
+export async function loadDiffOptionFile(config: VitestRunnerConfig, runner: VitestRunner) {
+  if (typeof config.diff !== 'string')
+    return
+
+  const diffModule = await runner.importFile(config.diff, 'diff') as any
+
+  if (diffModule && diffModule.default)
+    runner.config.diff = diffModule.default
```
sheremet-va, line 26: Should have more checks and throw an error if it's not on default for example

packages/runner/src/context.ts
```diff
@@ -49,6 +50,7 @@ export function createTestContext(test: Test, runner: VitestRunner): TestContext
 
   context.meta = test
   context.task = test
+  context.diff = runner.config.diff as DiffOptions
```
sheremet-va, line 53: Why do we pollute context here?

packages/runner/src/collect.ts
sheremet-va: Is there a reason why you load it for each test file separately when there is only one config?

sheremet-va (changes requested):

packages/runner/src/collect.ts
fenghan34: Looks like I totally misunderstood "It should be loaded inside a test context" 😅

packages/runner/src/types/tasks.ts
fenghan34: My mistake, just realized we can access it by `runner.config`

packages/runner/src/collect.ts
fenghan34: Could you tell me where should I load this config file? Thanks!

sheremet-va: > inside a test context

It just means that we need to load it in the test runtime, not on the server side. I can recommend doing it in setup.common and then just pass this config down to the runner

fenghan34: Now I got it, thanks for your patience and guidance!

packages/vitest/src/runtime/entry-vm.ts
```diff
@@ -41,7 +41,9 @@ export async function run(files: string[], config: ResolvedConfig, executor: Vit
   if (config.chaiConfig)
     setupChaiConfig(config.chaiConfig)
 
+  const diffConfig = await loadDiffConfig(config, executor)
```
sheremet-va, line 44: Can't you do this inside `resolveTestRunner`?

sheremet-va (changes requested):
I don't see this option working in the browser?

packages/vitest/src/runtime/entry-vm.ts
fenghan34: Thank you for reviewing, already updated.

fenghan34:
> I don't see this option working in the browser?

I wasn't aware of that, just added browser mode support.

docs/config/index.md
```diff
@@ -1619,3 +1619,11 @@ By default Vitest will run all of your test cases even if some of them fail. Thi
...
 
 Retry the test specific number of times if it fails.
+
+### diff
+
+- **Type:** `string`
+- **CLI:** `--diff=<value>`
+- **Version:** Since Vitest 0.34.4
+
+Path to a diff config that will be used to generate diff interface. Useful if you want to customize diff display.
```
sheremet-va, line 1629: Can you add a code example here? To showcase how to define it using typescript. A little bit of information how colors are defined can also be useful (need start/end property I think).

PR #4069, opened 2023-09-04
chore: update sources type for vite 5

bluwy:
### Description

In Vite 5, the `map` type is `SourceMap | { mappings: '' }` and the new `{ mappings: '' }` type trips up TypeScript. This PR handles the typecheck fail from[ vite-ecosystem-ci](https://github.com/vitejs/vite-ecosystem-ci/actions/runs/6069325893/job/16463576124#step:8:795)

Disable whitespace when reviewing the git diff, as I'm only wrapping `if ('sources' in map) { ... }` between the affected code. 

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vite-node/src/source-map.ts
+++ b/packages/vite-node/src/source-map.ts
@@ -25,21 +25,23 @@ export function withInlineSourcemap(result: TransformResult, options: {
   if (!map || code.includes(VITE_NODE_SOURCEMAPPING_SOURCE))
     return result
 
-  map.sources = map.sources?.map((source) => {
-    if (!source)
+  if ('sources' in map) {
+    map.sources = map.sources?.map((source) => {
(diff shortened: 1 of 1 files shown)


sheremet-va:
Is there a reason why it returns such a weird type? Looks unintentional to me to be honest.

bluwy:
That's the type Rollup documents [here](https://rollupjs.org/plugin-development/#source-code-transformations) and some plugins during the transform hook can return that, which in turn affects `transformRequest`'s return type.

An empty map like `{ mappings: '' }` could happen in this [Vite `loadAndTransform` flow](https://github.com/vitejs/vite/blob/34826aae015ed16dc9b9096c0f778154ca6981a6/packages/vite/src/node/server/transformRequest.ts#L270-L287).

Issue #4045, opened 2023-08-29
`getTaskFullName()` includes leading space, breaking `testNamePattern`


jaswrks:
### Describe the bug

The `getTaskFullName()` function injects a leading space when it detects a suite. However, a suite itself includes a `suite` property with an empty name, and therefore it runs twice on a suite. The end result is that every test inside a suite ends up with a full task name that has a leading single-space indentation, which breaks regular expression patterns like this:

```
--testNamePattern '^my-suite-prefix:'
```

I did finally get it working, by using this to account for the leading space.

```
--testNamePattern '^\s*my-suite-prefix:'
```

### Reproduction

```
--testNamePattern '^my-suite-prefix:'
```

Will not match a suite like this:

```
describe('my-suite-prefix:', () => { ... });
```

### System Info

```shell
Vitest 0.34.3
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

jaswrks:
Problematic line:

https://github.com/vitest-dev/vitest/blob/0d0f35f6930c98e034bcf61a53ce21d3b63fdec2/packages/runner/src/utils/collect.ts#L49

sheremet-va:
@segrey would you be ok with reintroducing #4071? Is Webstorrm ready for this change?

PR #4071, opened 2023-09-05
fix(runner): incorrect test name pattern matching

Dunqing:
### Description

close: #4045

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/runner/src/utils/collect.ts
+++ b/packages/runner/src/utils/collect.ts
@@ -46,7 +46,7 @@ export function interpretTaskModes(suite: Suite, namePattern?: string | RegExp, 
 }
 
 function getTaskFullName(task: TaskBase): string {
-  return `${task.suite?.name ? `${getTaskFullName(task.suite)} ` : ''}${task.name}`
+  return `${task.suite ? `${getTaskFullName(task.suite)}${task.suite.name ? ' ' : ''}` : ''}${task.name}`
 }
 
 export function someTasksAreOnly(suite: Suite): boolean {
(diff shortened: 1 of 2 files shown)


packages/runner/src/utils/collect.ts
```diff
@@ -46,7 +46,7 @@ export function interpretTaskModes(suite: Suite, namePattern?: string | RegExp,
 }
 
 function getTaskFullName(task: TaskBase): string {
-  return `${task.suite ? `${getTaskFullName(task.suite)} ` : ''}${task.name}`
+  return `${task.suite?.name ? `${getTaskFullName(task.suite)} ` : ''}${task.name}`
```
sheremet-va, line 49: What if suites suite name exist, but this one doesn't have it? It should check the result of the function here

sheremet-va (changes requested):

packages/runner/src/utils/collect.ts
Dunqing: Yes, Thanks for your review!

sheremet-va: I meant you should do `getTaskFullName` first, and check against it

Dunqing: Yes, I know what you mean. The end result is the same. Because this function ends up returning `test.suite.name`

sheremet-va: But it leaks implementation details of the function

Dunqing: Go it

sheremet-va (approved):

sheremet-va:
This PR was reverted, but I think we are ready to reimplement it again for 1.0

Issue #4052, opened 2023-08-30
cannot run tests without opening broken dev server


vollmerr:
### Describe the bug

After upgrading from vitest `^0.33.0` to `^0.34.3` when running any tests the dev server now opens, and in that dev server that is opened it fails to load the files.

Have tried setting the vite config to have test.open to false but still opens (which shows as default of true in the typedef file but false in https://vitest.dev/config/#open)

### Reproduction

https://codesandbox.io/p/sandbox/gallant-ptolemy-fjnjwr

### System Info

```shell
System:
    OS: Windows 10 10.0.22621
    CPU: (20) x64 12th Gen Intel(R) Core(TM) i9-12900HK
    Memory: 14.30 GB / 31.68 GB
  Binaries:
    Node: 18.16.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - C:\Program Files\nodejs\yarn.CMD
    npm: 9.5.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22621.2134.0), Chromium (116.0.1938.62)
    Internet Explorer: 11.0.22621.1
  npmPackages:
    @vitejs/plugin-react: ^4.0.4 => 4.0.4
    @vitest/coverage-c8: ^0.33.0 => 0.33.0
    @vitest/ui: ^0.34.3 => 0.34.3
    vite: ^4.4.9 => 4.4.9
    vitest: ^0.34.3 => 0.34.3
```

### Used Package Manager

yarn

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

donkeykong91:
I'm having the same problem.

leonheess:
[...]

PR #4094, opened 2023-09-08
fix(vitest): shouldn't open the browser by default

Dunqing:
### Description

fix: #4052

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/plugins/index.ts
+++ b/packages/vitest/src/node/plugins/index.ts
@@ -57,7 +57,7 @@ export async function VitestPlugin(options: UserConfig = {}, ctx = new Vitest('t
 
         ;(options as ResolvedConfig).defines = defines
 
-        let open: string | boolean | undefined
+        let open: string | boolean | undefined = false
 
         if (testConfig.ui && testConfig.open)
           open = testConfig.uiBase ?? '/__vitest__/'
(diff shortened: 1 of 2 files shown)


packages/vitest/src/node/plugins/index.ts
```diff
@@ -57,7 +57,7 @@ export async function VitestPlugin(options: UserConfig = {}, ctx = new Vitest('t
 
         ;(options as ResolvedConfig).defines = defines
 
-        let open: string | boolean | undefined
+        let open: string | boolean | undefined = false
```
sheremet-va, line 60: This doesn't actually fix the issue. Server still starts on the same port. What is needed here is to update `resolveApiConfig` method - it should treat `optimizer.web.enabled` and `optimizer.ssr.enabled` like `--api` flag. They are enabled by default, by the way, so they need to be explicitly set to false for us to not start a server.

This change hovewer should also be included in the fix. We only want to open a server when `ui` is enabled

sheremet-va (changes requested):

packages/vitest/src/node/plugins/index.ts
Dunqing: Okay, I've changed it, I'm not sure I got your point entirely

sheremet-va: Mostly yes, but I got a step further: #4095

What do you think?

PR #4111, opened 2023-09-11
feat: support Bun by using ShadowRealm

sheremet-va:
### Description

Currently, all jsdom tests are failing because bun doesn't seem to correctly support VM module.

The reason why other tests are failing is still unknown to me,

### Tests

### Documentation

### Changesets


AriPerkkio:
> all jsdom tests are failing because bun doesn't seem to correctly support VM module.

The [compatibility page](https://bun.sh/docs/runtime/nodejs-apis#node-vm) says that "VM modules" are not supported but I interpreted that as "`importModuleDynamically` does not work". Maybe I'm just reading this wrong. 🙃 

It does say that core `node:vm` functionalities are implemented. I think JSDOM was using `createContext`+`runInContext` which seem to be exported: https://github.com/oven-sh/bun/blob/51d3d4382281f789f8175079ed426a63529eb3e7/src/js/node/vm.ts#L8. I haven't yet had time to try this out but maybe there's something still missing.

PR #4113, opened 2023-09-12
feat(vitest): support `vi.waitFor` method

### LINKED ISSUE #4061: Built-in waitFor and waitUntil

There are situations when it's not obvious when assertion will succeed because there is no way to track it (async operation performed inside a sync one). For this situation, community has come up with different solutions:

- https://www.npmjs.com/package/wait-for-expect
- https://www.npmjs.com/package/async-wait-until

### Suggested solution

I think it would be nice to provide these kinds of helpers in the core. 

### Alternative

_No response_

### Additional context

_No response_

### Validations

Dunqing:
### Description

Partial closes - #4061

The `waitFor` function was inspired by https://github.com/testing-library/web-testing-library/pull/2.

Currently only supports `timeout` and `interval` options. I think this is enough, Let me know if you need to add any other options! 

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/api/vi.md
+++ b/docs/api/vi.md
@@ -715,6 +715,7 @@ unmockedIncrement(30) === 31
 ## vi.isFakeTimers
 
 - **Type:** `() => boolean`
+- **Version:** Since Vitest 0.34.5
 
   Returns `true` if fake timers are enabled.
 
--- a/docs/api/vi.md
+++ b/docs/api/vi.md
@@ -254,10 +254,10 @@ import { vi } from 'vitest'
   ```ts
   // increment.test.js
   import { vi } from 'vitest'
-  
+
   // axios is a default export from `__mocks__/axios.js`
   import axios from 'axios'
-  
+
   // increment is a named export from `src/__mocks__/increment.js`
   import { increment } from '../increment.js'
 
@@ -371,7 +371,7 @@ test('importing the next module imports mocked one', async () => {
 
   ```ts
   import { vi } from 'vitest'
-  
+
   import { data } from './data.js' // Will not get reevaluated beforeEach test
 
   beforeEach(() => {
@@ -717,3 +717,60 @@ unmockedIncrement(30) === 31
 - **Type:** `() => Vitest`
 
   When timers are run out, you may call this method to return mocked timers to its original implementations. All timers that were run before will not be restored.
+
+### vi.waitFor
+
+- **Type:** `function waitFor<T>(callback: WaitForCallback<T>, options?: number | WaitForOptions): Promise<T>`
+- **Version**: Since Vitest 0.35.0
+
+Wait for the callback to execute successfully. If the callback throws an error or returns a rejected promise it will continue to wait until it succeeds or times out.
+
+This is very useful for testing, see the following example.
+
+```ts
+import { test, vi } from 'vitest'
+
+let server = false
+setTimeout(() => {
+  server = true
+}, 100)
+
+function checkServerStart() {
+  if (!server)
+    throw new Error('Server not started')
+
+  console.log('Server started')
+}
+
+test('Server started successfully', async () => {
+  const res = await vi.waitFor(checkServerStart, {
+    timeout: 500, // default is 1000
+    interval: 20, // default is 50
+  })
+  expect(server).toBe(true)
+})
+```
+
+It also works for asynchronous callbacks
+
+```ts
+import { test, vi } from 'vitest'
+
+async function startServer() {
+  return new Promise((resolve) => {
+    setTimeout(() => {
+      server = true
+      resolve('Server started')
+    }, 100)
+  })
+}
+
+test('Server started successfully', async () => {
+  const server = await vi.waitFor(startServer, {
+    timeout: 500, // default is 1000
+    interval: 20, // default is 50
+  })
+  expect(server).toBe('Server started')
+})
+```
+
--- a/docs/api/vi.md
+++ b/docs/api/vi.md
@@ -775,4 +775,4 @@ test('Server started successfully', async () => {
 })
 ```
 
-If `vi.useFakeTimers` is used, `vi.waitFor` automatically calls `vi.AdvanceTimersByTime(interval)` in every check callback.
+If `vi.useFakeTimers` is used, `vi.waitFor` automatically calls `vi.advanceTimersByTime(interval)` in every check callback.
--- a/docs/api/vi.md
+++ b/docs/api/vi.md
@@ -725,24 +725,25 @@ unmockedIncrement(30) === 31
 
 Wait for the callback to execute successfully. If the callback throws an error or returns a rejected promise it will continue to wait until it succeeds or times out.
 
-This is very useful for testing, see the following example.
+This is very useful when you need to wait for some asynchronous action to complete, for example, when you start a server and need to wait for it to start.
 
 ```ts
 import { test, vi } from 'vitest'
 
-let server = false
-setTimeout(() => {
-  server = true
-}, 100)
-
-function checkServerStart() {
-  if (!server)
-    throw new Error('Server not started')
-
-  console.log('Server started')
-}
-
 test('Server started successfully', async () => {
+  let server = false
+
+  setTimeout(() => {
+    server = true
+  }, 100)
+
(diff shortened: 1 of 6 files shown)


packages/vitest/src/integrations/wait.ts
```diff
@@ -0,0 +1,93 @@
...
+      if (intervalId)
+        clearInterval(intervalId)
+
+      resolve(result)
+    }
+
+    const handleTimeout = () => {
+      let error = lastError
+      if (!error)
+        error = copyStackTrace(new Error('Timed out in waitFor!'), STACK_TRACE_ERROR)
```
sheremet-va, line 49: Should we just add a `cause` to an error instead of patching it?

test/core/test/wait.test.ts
```diff
@@ -0,0 +1,107 @@
...
+      const callback = vi.fn(() => {
+        throw new Error('interval error')
+      })
+
+      await expect(
+        vi.waitFor(callback, {
+          timeout: 499,
+          interval: 100,
+        }),
+      ).rejects.toMatchInlineSnapshot('[Error: interval error]')
```
sheremet-va, line 28: Can you use `toThrowMatchingInlineSnaphsot`?

test/core/test/wait.test.ts
```diff
@@ -0,0 +1,107 @@
...
+      }
+    })
+    expect(throwError).toBe(true)
+  })
+
+  test('async function', async () => {
+    let finished = false
+    setTimeout(() => {
+      finished = true
+    }, 500)
```
sheremet-va, line 49: Can you lower the numbers so tests are finished faster?

sheremet-va (changes requested):

packages/vitest/src/integrations/wait.ts
Dunqing: This error needs stack because can help us find the call stack, If use `cause` instead of patching it the the error displays like it.

image

The expected should be this

image

AriPerkkio, line 49: One thing I've seen some developers running into is that `waitFor` is failing for them when `useFakeTimers()` is used and they are unaware of this. It could be that they simply forgot to call `useRealTimers()` or that timer mocking was done by some utility function which hides it. 

Could we add some notification here if `waitFor` fails and timers are mocked? E.g. `Timed out in waitFor. ${timersMocked ? 'Note that timers were mocked with useFakeTimers(). This may affect your waitFor usage.' : ''}` or similar. Goal of this would be to provide additional information for debugging failures.

sheremet-va: Function is using safe timers, so I'm not sure if we need this?

sheremet-va: The cause should be the other error, not timed-out one. If it can't show the preview then it's a bun in Vitest that we should fix instead of hide it like this.

AriPerkkio: `waitFor` itself is running safely and correctly, but if a developer is waiting for a change while timers are mocked, that change won't happen.

Dunqing: > The cause should be the user error, not timed-out one. 

What does mean "user error"?  The patching stack trace behaviour only executes when the callback doesn't throw any error. So we haven't `cause`, just timed out error.

> If it can't show the preview then it's a bug in Vitest that we should fix instead of hide it like this.

Seems like the `cause` doesn't show code frame is intentional.  Do you know why?

https://github.com/vitest-dev/vitest/blob/65b200f0dd44357d26eac740d7d097d364078bd8/packages/vitest/src/node/error.ts#L105-L107

sheremet-va: Hm, now I see what this code does. Ok, since we support different browsers, let's keep the patch.

Dunqing: I prefer to add some notifications here. mocked time may complicate

sheremet-va: And what should users do then? I don't understand how this is helpful instead of doing the expected? (advance timers)

Dunqing: The way it's handled now is what's expected, we need to do nothing here. Add notification just a friendly reminder and help users how to debug the test.

sheremet-va: How do you fix it tho? Users can't control the timer while `waitFor` is running.

Dunqing: Ahh, yes, waitFor will be pause until timed out

packages/vitest/src/integrations/wait.ts
```diff
@@ -52,6 +53,10 @@ export function waitFor<T>(callback: WaitForCallback<T>, options: number | WaitF
     }
 
     const checkCallback = () => {
+      // use fake timers
+      if (globalThis.setTimeout !== setTimeout)
```
Dunqing, line 57: I haven't found a better way to tell if `useFakeTimer` is being used or not

sheremet-va: I think fake timer has `_clock` property or something like that

sheremet-va: Sure, we can add it to vi object

packages/vitest/src/integrations/wait.ts
```diff
@@ -0,0 +1,101 @@
...
+      if (!error)
+        error = copyStackTrace(new Error('Timed out in waitFor!'), STACK_TRACE_ERROR)
+
+      reject(error)
+    }
+
+    const checkCallback = () => {
+      try {
+        // use fake timers
+        vi.getTimerCount()
```
sheremet-va, line 58: Huge performance loss if we do it like this, better to expose a method to check if it's enabled

packages/vitest/src/integrations/vi.ts
```diff
@@ -12,6 +12,7 @@ import { fn, isMockFunction, spies, spyOn } from './spy'
 import { waitFor } from './wait'
 
 interface VitestUtils {
+  isFakeTimers: () => boolean
```
sheremet-va, line 15: Let's keep the syntax similar to other methods:

```suggestion
  isFakeTimers(): boolean
```

docs/api/vi.md
```diff
@@ -712,6 +712,12 @@ unmockedIncrement(30) === 31
 
   The implementation is based internally on [`@sinonjs/fake-timers`](https://github.com/sinonjs/fake-timers).
 
+## vi.isFakeTimers
+
+- **Type:** `() => boolean`
```
sheremet-va, line 717:
```suggestion
- **Type:** `() => boolean`
- **Version:** Since Vitest 0.34.5
```

sheremet-va (changes requested):

docs/api/vi.md
```diff
@@ -712,8 +712,73 @@ unmockedIncrement(30) === 31
...
 ## vi.useRealTimers
 
 - **Type:** `() => Vitest`
 
   When timers are run out, you may call this method to return mocked timers to its original implementations. All timers that were run before will not be restored.
+
+### vi.waitFor
+
+- **Type:** `function waitFor<T>(callback: WaitForCallback<T>, options?: number | WaitForOptions): Promise<T>`
+- **Version**: Since Vitest 0.35.0
```
sheremet-va, line 730:
```suggestion
- **Version**: Since Vitest 0.34.5
```

sheremet-va (changes requested):

sheremet-va (approved):

sheremet-va:
I would assume `waitUntil` would be the same function but instead of try/catch it will just check the return value :P

Dunqing:
> I would assume `waitUntil` would be the same function but instead of try/catch it will just check the return value :P

Yes. Do we still need it? Looks like `waitFor` is enough.

sheremet-va:
> Yes. Do we still need it? Looks like `waitFor` is enough.

Yes, we still need both. `waitUntil` should be more performant because throwing an error is expensive, but `waitFor` can have assertions inside which is a better DX. So, it's about trade-offs.

PR #4118, opened 2023-09-12
fix(inline-snapshot): support mix of normal/with placeholders snapshots

RobertPechaCZ:
### Description

- Undesired behavior happens when you use multiple `expect(...).toMatchInlineSnapshot()` methods within same `it` or `test` block and you mix method with shape object with placeholders as 1st argument and method without that.
- Vitest either doesn't update any inline snapshot or it may update even wrong inline snapshot under rare curcumstances.

### Tests

### Documentation

### Changesets


sheremet-va (approved):

sheremet-va:
LGTM!

PR #4119, opened 2023-09-12
Another way to allow leading space in `testNamePattern`

### LINKED ISSUE #4103: Changed `testNamePattern` behavior breaks IntelliJ / WebStorm (vitest 0.34.4)
### Describe the bug

`testNamePattern` format  has been changed in vitest 0.34.4 (#4071). Unfortunately, it broke running a single test/suite in IntelliJ / WebStorm ([WEB-62946](https://youtrack.jetbrains.com/issue/WEB-62946/Unable-to-run-debug-single-test-suite-using-vitest0.34.4)).
The difference is that, previously, `testNamePattern` had a leading space always. Now, `testNamePattern` doesn't have a leading space, but previously released versions of IntelliJ / WebStorm still pass `testNamePattern` starting with a space.

Unfortunarely, it will be hard to release a hotfix of IntelliJ/WebStorm quickly, because of the release schedule. Even if we would release it, our users would be required to upgrade their IDEs (which would be inconvenient for the users).

Sorry, may I ask you to restore the leading space in `testNamePattern` for now? I'll fix IntelliJ / WebStorm in the next version (2023.3) to support both formats of `testNamePattern`. I guess it would be OK to drop the leading space completely in 2024 Q1 - once WebStorm 2024.1 is released, to have two major WebStorm versions supporting both `testNamePattern` formats.

What do you think? @sheremet-va @Dunqing @jaswrks

### Reproduction

For example, to run [basic](https://github.com/segrey/vitest-basic/blob/54c97b505138c8482ee02c78f187a199d206cfe8/test/basic.test.ts#L3) test only in vitest 0.34.3, one should use the following command line:
```
./node_modules/.bin/vitest --run '--testNamePattern=^ basic$' test/basic.test.ts
```

```
 RUN  v0.34.3 /Users/sergey.simonchik/sandbox/vitest-basic

 ✓ test/basic.test.ts (3)
   ✓ basic
   ↓ my suite (2) [skipped]
     ↓ foo [skipped]
     ↓ bar [skipped]

 Test Files  1 passed (1)
      Tests  1 passed | 2 skipped (3)
   Start at  12:14:49
   Duration  142ms (transform 23ms, setup 0ms, collect 15ms, tests 1ms, environment 0ms, prepare 41ms)
```

Using the same command line doesn't work with vitest 0.34.4:
```
➜  vitest-basic git:(master) ✗ ./node_modules/.bin/vitest --run '--testNamePattern=^ basic$' test/basic.test.ts

 RUN  v0.34.4 /Users/sergey.simonchik/sandbox/vitest-basic

 ↓ test/basic.test.ts (3) [skipped]
   ↓ basic [skipped]
   ↓ my suite (2) [skipped]
     ↓ foo [skipped]
     ↓ bar [skipped]

 Test Files  1 skipped (1)
      Tests  3 skipped (3)
   Start at  12:43:12
   Duration  136ms (transform 24ms, setup 0ms, collect 13ms, tests 0ms, environment 0ms, prepare 40ms)
```

### System Info

```shell
System:
    OS: macOS 13.5.2
    CPU: (10) arm64 Apple M1 Max
    Memory: 3.46 GB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.6.1 - ~/.nvm/versions/node/v20.6.1/bin/node
    npm: 9.8.1 - ~/.nvm/versions/node/v20.6.1/bin/npm
    pnpm: 7.14.0 - ~/.nvm/versions/node/v20.6.1/bin/pnpm
  Browsers:
    Chrome: 116.0.5845.179
    Safari: 16.6
  npmPackages:
    vitest: 0.34.4 => 0.34.4
```

### Used Package Manager

npm

### Validations

segrey:
### Description

This is another attempt to fix #4103.
Pro: the fixed version of `getTaskFullName` is restored: no leading space makes #4045 fixed again.
Con: `taskFullName.match(namePattern)` is called twice.

@sheremet-va @Dunqing

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/runner/src/utils/collect.ts
+++ b/packages/runner/src/utils/collect.ts
@@ -27,8 +27,14 @@ export function interpretTaskModes(suite: Suite, namePattern?: string | RegExp, 
       }
     }
     if (t.type === 'test') {
-      if (namePattern && !getTaskFullName(t).match(namePattern))
-        t.mode = 'skip'
+      if (namePattern) {
+        const taskFullName = getTaskFullName(t)
+        // Match also the task full name with a leading space to be backward-compatible with tools like
+        // IntelliJ/WebStorm. Previous Vitest versions (<= 0.34.3) had the task full name starting
+        // with a space, and the tools passed `testNamePattern` matching it.
+        if (!(taskFullName.match(namePattern) || ` ${taskFullName}`.match(namePattern)))
+          t.mode = 'skip'
+      }
     }
     else if (t.type === 'suite') {
       if (t.mode === 'skip')
--- a/packages/runner/src/utils/collect.ts
+++ b/packages/runner/src/utils/collect.ts
@@ -46,7 +46,8 @@ export function interpretTaskModes(suite: Suite, namePattern?: string | RegExp, 
 }
 
 function getTaskFullName(task: TaskBase): string {
-  return `${task.suite ? `${getTaskFullName(task.suite)} ` : ''}${task.name}`
+  const fullName = task.suite ? getTaskFullName(task.suite) : null
+  return fullName ? `${fullName} ${task.name}` : task.name
 }
 
 export function someTasksAreOnly(suite: Suite): boolean {
--- a/test/filters/test/testname-pattern.test.ts
+++ b/test/filters/test/testname-pattern.test.ts
@@ -30,3 +30,14 @@ test('match by pattern that also matches current working directory', async () =>
   expect(stdout).toMatch('Test Files  1 passed (1)')
   expect(stdout).not.toMatch('test/example.test.ts')
 })
+
+test('match by test name pattern with ^', async () => {
+  const { stdout } = await runVitest({
+    root: './fixtures',
+    testNamePattern: '^this',
+  }, ['filters'])
+
+  expect(stdout).toMatch('✓ test/filters.test.ts > this will pass')
+  expect(stdout).toMatch('Test Files  1 passed (1)')
+  expect(stdout).not.toMatch('test/example.test.ts')
+})


segrey:
I'm not entirely sure if it's better than the current solution (#4119). Just wanted to share with you to discuss.

sheremet-va:
I think this change is ok. If you can fix the failing CI, we can merge it.

Dunqing:
It looks like it just needs a lint fix, do you have time to look here and finally merge it before releasing the next version

segrey:
@sheremet-va @Dunqing Done, please take a look.

Dunqing (approved):

segrey:
@sheremet-va Could you please take a look? Thanks!

sheremet-va:
I don't see the need for this PR personally. WebStorm should be fixed already, and the tool should change the behavior in the next version to not rely on the leading space.

Vitest eventually will remove the space (probably in v1?)

segrey:
Yes, WebStorm is fully covered: (a) now it passes `testNamePattern` with optional leading space and (b) thanks to releasing Vitest 0.34.5 with the restored previous behavior. I can only think of this PR as a smooth transition for other clients also passing `testNamePattern` with a leading space now. It might be a surprise for them when Vitest v1 is released. Otherwise, this PR adds no value indeed.

sheremet-va:
> I can only think of this PR as a smooth transition for other clients also passing `testNamePattern` with a leading space now. It might be a surprise for them when Vitest v1 is released. Otherwise, this PR adds no value indeed.

I would expect other clients to migrate as it is a breaking change like any other.

PR #4126, opened 2023-09-14
chore: update threadsCount logic to use availableParallelism

diogotorres97:
### Description
We found an issue running vitest in self-hosted github action runners where they use as MAX_THREADS the number of CPUs available in the node instead of respecting the resources.limits assigned to the pod.

Introduced in v19.4.0, v18.14.0, [`os.availableParallelism()`](https://nodejs.org/api/os.html#osavailableparallelism) allow us to estimate of the default amount of parallelism a program should use instead of using [`os.cpus`](https://nodejs.org/api/os.html#oscpus) .

> os.cpus().length should not be used to calculate the amount of parallelism available to an application. Use [os.availableParallelism()](https://nodejs.org/api/os.html#osavailableparallelism) for this purpose.

A similar work was already done in another libs like https://github.com/jestjs/jest/pull/13738.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/pools/threads.ts
+++ b/packages/vitest/src/node/pools/threads.ts
@@ -1,5 +1,5 @@
 import { MessageChannel } from 'node:worker_threads'
-import { cpus } from 'node:os'
+import * as nodeos from 'node:os'
 import { pathToFileURL } from 'node:url'
 import { createBirpc } from 'birpc'
 import { resolve } from 'pathe'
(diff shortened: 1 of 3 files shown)


packages/vitest/src/node/pools/threads.ts
```diff
@@ -1,5 +1,5 @@
 import { MessageChannel } from 'node:worker_threads'
-import { cpus } from 'node:os'
+import { availableParallelism, cpus } from 'node:os'
```
sheremet-va, line 2: This will just fail because `node:os` doesn't export this method on lower Node.js versions. And you cannot access undeclared named exports.

Using default export might be better

diogotorres97: done, or do you have another suggestion?

sheremet-va (approved):

PR #4129, opened 2023-09-15
feat(vitest): support `vi.waitUntil` method

Dunqing:
### Description

close: https://github.com/vitest-dev/vitest/issues/4061

A question for we need to strict the returns value for callback?

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/api/vi.md
+++ b/docs/api/vi.md
@@ -804,5 +804,3 @@ test('Element render correctly', async () => {
   expect(element.querySelector('.element-child')).toBeTruthy()
 })
 ```
-
-```
--- a/docs/api/vi.md
+++ b/docs/api/vi.md
@@ -784,3 +784,25 @@ If `vi.useFakeTimers` is used, `vi.waitFor` automatically calls `vi.advanceTimer
 - **Version**: Since Vitest 0.34.5
 
 This is similar to `vi.waitFor`, but if the callback throws any errors, execution is immediately interrupted and an error message is received. If the callback returns falsy value, the next check will continue until truthy value is returned. This is useful when you need to wait for something to exist before taking the next step.
+
+Look at the example below. We can use `vi.waitUntil` to wait for the element to appear on the page, and then we can do something with the element.
+
+```ts
+import { test, vi } from 'vitest'
+
+test('Element render correctly', async () => {
+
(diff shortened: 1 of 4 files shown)


packages/vitest/src/integrations/wait.ts
```diff
@@ -95,3 +95,82 @@ export function waitFor<T>(callback: WaitForCallback<T>, options: number | WaitF
     intervalId = setInterval(checkCallback, interval)
   })
 }
+
+export type WaitUntilCallback = () => boolean | Promise<boolean>
```
sheremet-va, line 99: I don't think we need to ask for a boolean. Truthy values are sufficient. It would also be nice to actually return the value:

```ts
const element = await vi.waitUntil(() => document.querySelector('.element'))
```

sheremet-va: Does waitFor also return the value? Would be nice to support it in both methods

Dunqing: > Does waitFor also return the value? Would be nice to support it in both methods

The `waitFor` already supports the return callback's value

docs/api/vi.md
```diff
@@ -777,3 +777,10 @@ test('Server started successfully', async () => {
 ```
 
 If `vi.useFakeTimers` is used, `vi.waitFor` automatically calls `vi.advanceTimersByTime(interval)` in every check callback.
+
+### vi.waitUntil
+
+- **Type:** `function waitUntil(callback: WaitUntilCallback, options?: number | WaitUntilOptions): Promise`
+- **Version**: Since Vitest 0.34.5
+
+This is similar to `vi.waitFor`, but if the callback throws any errors, execution is immediately interrupted and an error message is received. If the callback returns falsy value, the next check will continue until truthy value is returned. This is useful when you need to wait for something to exist before taking the next step.
```
sheremet-va, line 786: Can we add some code examples for both methods? I think `document.querySelector` is a nice one.

```ts
const element = await vi.waitUntil(() => document.querySelector('.element'))
vi.waitFor(() => {
  expect(document.querySelector('.element')).toBeTruthy()
})
```

Dunqing: Added.

Issue #4142, opened 2023-09-18
Should vite be a peerDependency instead of a full dependency?


thebanjomatic:
### Clear and concise description of the problem

Currently Vitest lists vite as a full dependency using:

```json
"vite": "^3.1.0 || ^4.0.0 || ^5.0.0-0",
```

In my project which uses Vite 4.4.9:
```json
"vite": "^4.4.9"
```

If I then run `yarn add vitest`, yarn will somewhat unhelpfully install both:
vite v5 and vite v4:
```yaml
"vite@npm:^3.0.0 || ^4.0.0, vite@npm:^4.4.9":
  version: 4.4.9

"vite@npm:^3.1.0 || ^4.0.0 || ^5.0.0-0":
  version: 5.0.0-beta.2
```

The fact that it doesn't see that `4.4.9` satisfies the new condition and just re-use the existing version of vite is annoying, but it winds up causing issues type-checking the vite configuration because there is a type mismatch and two versions of vite are used. The plugins (which use vite as a peer dependency) are using vite 4 types, and `"vitest/config"` is using the vite 5 types:

<img width="1035" alt="image" src="https://github.com/vitest-dev/vitest/assets/17907922/72e18047-5cf8-4d30-901b-e6c08f96a779">

### Suggested solution

If vitest's vite dependency was listed as a `peerDependency`, then it would always use the version of vite that is referenced in the project's `package.json` file and resolutions would not be required to align the two versions.

In terms of potential disadvantages, using vitest would now also require explicitly installing vite. For many users, vite is already installed as a devDependency and no extra work is required. There may be some usage of vitest in projects that don't use vite to build, and for those consumers, this would be a breaking change as it would require adding both vitest and vite under devDependencies. This could be mitigated by shipping vite as both a `dependency` and `peerDependency`, but then users would occasionally have two versions of vite installed in `node_modules` if the version mismatched (currently the case today as well), but only one consistent version would be used (better than today's case).

I'm also not sure what the expected behavior is when doing something like using vitest at the root of a monorepo and specifying a vitest.workspace.ts to find and run tests in different packages. Each of those packages could have their own vite version. With the current behavior, a single version of vite is used for all the different packages, but that version is somewhat opaque to the user and hidden in the lock-file. If vite was a peerDependency then the behavior would be essentially identical, but the version of vite used would be enforced and controlled by the vite dependency at the root level, and would be more transparent to the user.

[...]

Issue #4112, opened 2023-09-11
Since v0.34.4 Yarn installs two different versions of Vite.


jgerigmeyer:
### Describe the bug

[This PR](https://github.com/vitest-dev/vitest/pull/4015) in v0.34.4 introduced a union operator (`||`) for the Vite dependency to include support for Vite v5.0. Because of [how Yarn handles union operators](https://github.com/yarnpkg/yarn/issues/8243), this means that any Yarn project specifying a dependency on Vite v4 and Vitest v0.34.4 will end up with two different versions of Vite installed.

In other words, this `package.json`:

```json
{
  "dependencies": {
    "vite": "^4.4.9"
  },
  "devDependencies": {
    "vitest": "^0.34.4"
  }
}
```

Results in this output from `yarn why vite`:

```
➜ yarn why vite
├─ root-workspace-0b6124@workspace:.
│  └─ vite@npm:4.4.9 [dc3fc] (via npm:^4.4.9 [dc3fc])
│
├─ vite-node@npm:0.34.4
│  └─ vite@npm:4.4.9 [dc3fc] (via npm:^4.4.9 [dc3fc])
│
├─ vitest@npm:0.34.4
│  └─ vite@npm:5.0.0-beta.1 (via npm:^3.1.0 || ^4.0.0 || ^5.0.0-0)
│
└─ vitest@npm:0.34.4 [dc3fc]
   └─ vite@npm:5.0.0-beta.1 [1bd26] (via npm:^3.1.0 || ^4.0.0 || ^5.0.0-0 [1bd26])
```

### Reproduction

https://github.com/jgerigmeyer/vitest-vite-version

### System Info

```shell
System:
    OS: macOS 13.5.2
    CPU: (10) arm64 Apple M1 Pro
    Memory: 112.77 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.0.0 - ~/.nvm/versions/node/v20.0.0/bin/node
    Yarn: 3.6.3 - ~/.yarn/bin/yarn
    npm: 9.6.4 - ~/.nvm/versions/node/v20.0.0/bin/npm
  Browsers:
    Chrome: 116.0.5845.179
    Chrome Canary: 119.0.6002.0
    Firefox: 115.0.2
    Safari: 16.6
  npmPackages:
    vite: ^4.4.9 => 4.4.9
    vitest: ^0.34.4 => 0.34.4
```

### Used Package Manager

yarn

[...]

PR #4144, opened 2023-09-18
chore: making vite a peer depencency

thebanjomatic:
### Description

Breaking Change:
vitest and vite-node now both specify vite as a peerDependency and thus require that you add vite as an explicit dependency wherever you are using vitest or vite-node.

Closes: #4142 
Closes: #4112

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vite-node/package.json
+++ b/packages/vite-node/package.json
@@ -82,16 +82,19 @@
     "prepublishOnly": "pnpm build",
     "typecheck": "tsc --noEmit"
   },
+  "peerDependencies": {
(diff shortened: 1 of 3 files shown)


thebanjomatic:
Note I wasn't able to run `pnpm install` even in the `main` branch without a ton of unexpected diffs in the `pnpm-lock.yaml`. I tried to prune the lock file changes down to just what I was expecting, but the maintainers will probably want to verify that and maybe run `pnpm install` themselves.

sheremet-va:
Vite is a dependency of Vitest, it will not work without it. The same is true about vite-node. We cannot force people who don't use vite to install it.

PR #4193, opened 2023-09-27
Fix a code example

NoamAnisfeld:
When testing an expression that's supposed to error, it must be wrapped in a function


sheremet-va:
Rejected promises don't have to be wrapped

PR #4209, opened 2023-10-02
fix(vitest): inline chai types instead of using @types/chai

### LINKED ISSUE #2118: Typescript types adding unwanted / buggy 'Chai.Assertion' type requirement on `should` function interface return
### Describe the bug

When this package is installed in a fresh vite project, a function that has a type annotation to return an interface (of anything, as long as it is not an empty interface) mysteriously requires that a returned object's `should` key be of type `Assertion` if it returns an object literal

This is a problem, even if it is intended behaviour because a larger project that contains a function anywhere (in typescript) generally suffers from this mysterious behaviour.
It was not immediately obvious that this typing behaviour was from this package, however it is clearly the source or is causing it, an issue with this repository.

This bug was initially found by refactoring large sections of typing after installing this library.
This library extends the Object interface to add `should: Chai.Assertion`, which was *very* annoying to identify and results in this package being uninstalled (as it was not being used at the time)

### Reproduction

[Exact file from repo](https://github.com/ActuallyHappening/messing-about/blob/master/vite-project/src/bug.ts),
`git clone https://github.com/ActuallyHappening/messing-about.git`

Code (from repo):
```ts
interface test {
  anything: "here";
  asLongAs: "there is at least one property";
  inThe: "interface";
}
function func(): test {
  return {
    should: "what is this `Assertion` type?", // Error here, copy-pasted below
    /*
    Type 'string' is not assignable to type 'Assertion'.ts(2322)
    index.d.ts(1940, 5): The expected type comes from property 'should' which is declared here on type 'test'
    */
  };
};
```

### System Info

```shell
System:
    OS: macOS 12.6
    CPU: (8) arm64 Apple M2
    Memory: 108.75 MB / 8.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.17.0 - ~/.nvm/versions/node/v16.17.0/bin/node
    npm: 8.15.0 - ~/.nvm/versions/node/v16.17.0/bin/npm
  Browsers:
    Brave Browser: 104.1.42.97
    Chrome: 105.0.5195.125
    Safari: 16.0
```

### Used Package Manager

npm

### Validations

sheremet-va:
### Description

Fixes #2118

By the way, this is how Cypress fixes it 😄 

### Tests

### Documentation

### Changesets

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,6 +82,9 @@
       "ignoreMissing": [
         "@algolia/client-search"
       ]
+    },
+    "patchedDependencies": {
+      "@types/chai@4.3.6": "patches/@types__chai@4.3.6.patch"
     }
   },
   "simple-git-hooks": {
--- a/packages/expect/index.d.ts
+++ b/packages/expect/index.d.ts
@@ -1,0 +1,3 @@
+import './dist/chai'
+
+export * from './dist/index'
--- a/packages/expect/package.json
+++ b/packages/expect/package.json
@@ -17,14 +17,14 @@
   "sideEffects": false,
   "exports": {
     ".": {
-      "types": "./dist/index.d.ts",
+      "types": "./index.d.ts",
       "import": "./dist/index.js"
     },
     "./*": "./*"
   },
   "main": "./dist/index.js",
   "module": "./dist/index.js",
-  "types": "./dist/index.d.ts",
+  "types": "./index.d.ts",
   "files": [
     "dist"
   ],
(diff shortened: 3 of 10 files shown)


antfu (approved):
Interestingly they are so short :P

sheremet-va:
> Interestingly they are so short :P

The link in the issue points to chai v2, but we use the latest version which is 2k lines 😄 (mostly documentation)

kosgdk:
Now `not` is broken:

image

sheremet-va:
> Now `not` is broken:
> 
> 
> 
> image
> 
> 

If you can open a reproduction, that would be lovely

PR #4216, opened 2023-10-02
chore: use `ofetch` and latest `node-fetch-native`

pi0:
### Description

Hi! I just discovered vitest is using lots of unjs packages (❤️). We had renamed from `ohmyfetch` to `ofetch` almost year ago (there are unrelavant lockfile updates seems pnpm forces to update them as well in parallel).


sheremet-va:
Should we also update the usage? 🤔

pi0:
(pushed usage changes too however haven't tested, API is same and should be as before)

antfu (approved):

Issue #3888, opened 2023-08-04
Coverage report is incorrect since 0.34.0


nstepien:
### Describe the bug

When multiple environments are used, it seems like this affects how coverage is computed.
I've also noticed that the coverage report is different on my machine vs CI.

Please check this branch: https://github.com/adazzle/react-data-grid/pull/3298
The coverage dropped significantly
- [CI](https://github.com/adazzle/react-data-grid/actions/runs/5755182074/job/15602627691?pr=3298):
  - ![image](https://github.com/vitest-dev/vitest/assets/567105/32924d89-69d5-4e14-a28e-a619c54b39a2)
- On my machine:
  - ![image](https://github.com/vitest-dev/vitest/assets/567105/ddedc74f-8414-4fe8-a4fb-2655d9fd8c43)
- Codecov:
  - ![image](https://github.com/vitest-dev/vitest/assets/567105/c4b95094-b11c-4507-b3d0-66a2f3da8415)

In `test/ssr.test.tsx`, if I skip the test, the coverage INCREASES:
![image](https://github.com/vitest-dev/vitest/assets/567105/bdec9eb3-1af9-4cf1-a2f1-e2cf471b0577)

Again in `test/ssr.test.tsx`, if I skip the test AND remove `// @vitest-environment node`, the coverage INCREASES further, back to normal levels:
![image](https://github.com/vitest-dev/vitest/assets/567105/315e3ec7-4feb-4888-b09a-767e6142b660)

It seems like Vitest 0.34 does not like multiple environments. 🤔 

### Reproduction

- Clone and checkout this branch https://github.com/adazzle/react-data-grid/pull/3298
- `npm i`
- `npm t -- --coverage`

### System Info

```shell
System:
    OS: Windows 10 10.0.22621
    CPU: (64) x64 AMD Ryzen Threadripper 3970X 32-Core Processor
    Memory: 19.10 GB / 31.88 GB
  Binaries:
    Node: 20.5.0 - C:\Program Files\nodejs\node.EXE
    npm: 9.8.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22621.2070.0), Chromium (115.0.1901.188)
    Internet Explorer: 11.0.22621.1
  npmPackages:
    @vitejs/plugin-react: ^4.0.4 => 4.0.4
    @vitest/coverage-v8: ^0.34.1 => 0.34.1
    vite: ^4.4.8 => 4.4.8
[...]

PR #4221, opened 2023-10-03
fix(vite-node): have a separate cache for web/ssr transforms

sheremet-va:

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vite-node/src/server.ts
+++ b/packages/vite-node/src/server.ts
@@ -11,19 +11,33 @@ import { withInlineSourcemap } from './source-map'
 
 export * from './externalize'
 
+interface FetchCache {
+  duration?: number
+  timestamp: number
+  result: FetchResult
+}
+
 const debugRequest = createDebug('vite-node:server:request')
 
 export class ViteNodeServer {
-  private fetchPromiseMap = new Map<string, Promise<FetchResult>>()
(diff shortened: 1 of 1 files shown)


sheremet-va:
@AriPerkkio this should help you with the coverage, right?

packages/vite-node/src/server.ts
```diff
@@ -11,19 +11,33 @@ import { withInlineSourcemap } from './source-map'
...
-    duration?: number
-    timestamp: number
-    result: FetchResult
-  }>()
+  fetchCaches = {
+    ssr: new Map<string, FetchCache>(),
+    web: new Map<string, FetchCache>(),
+  }
+
+  fetchCache = new Map<string, FetchCache>()
```
AriPerkkio, line 40: Is this one left here for backwards compatibility or why is it needed?

AriPerkkio (commented):
Yes this is needed for https://github.com/vitest-dev/vitest/issues/3888, thank you!

We need to remember to make runners pass transformMode to coverage providers before releasing v1. Just so that it would not cause a breaking change. The fix for #3888 can then be implemented later.

packages/vite-node/src/server.ts
sheremet-va: Yes

PR #4239, opened 2023-10-04
feat(expect): toContain can handle classList and Node.contains

### LINKED ISSUE #2395: Support toContain with element.classList

We are building a lot of components and want to test as many features as possible. One of the things we often have to test is if classes are attached correctly to an HTMLElement. Currently, we use this approach:

```ts
expect(element.classList.contains('something')).toBeTruthy();
```

This works but does not give us a good error message if the expectation fails. In this case, we just get the message that it is not `true` but `false` (except we add an error message manually).

### Suggested solution

It would be great if we could use `.toContain` for this purpose. It would have the benefit of meaningful error messages if the expectation fails (e.g. a list of the actual classes of the element).

```ts
expect(element.classList).toContain('something');
```

### Alternative

_No response_

### Additional context

_No response_

### Validations

sheremet-va:
### Description

Fixes #2395

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/expect/src/jest-expect.ts
+++ b/packages/expect/src/jest-expect.ts
@@ -164,7 +164,7 @@ export const JestChaiExpect: ChaiPlugin = (chai, utils) => {
       return this.match(expected)
   })
   def('toContain', function (item) {
-    const actual = this._obj
+    const actual = this._obj as Iterable<unknown> | string | Node | DOMTokenList
 
     if (typeof Node !== 'undefined' && actual instanceof Node) {
       if (!(item instanceof Node))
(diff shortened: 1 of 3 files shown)


ferdynator:
Any updates on this? `toContains` was working with classLists just fine in jest and this is breaking tests on migrating away to vitest

charkour:
This update also broke my tests in https://github.com/charkour/zundo. I'm not sure why that's the case yet.

packages/expect/src/jest-expect.ts
```diff
@@ -164,6 +173,36 @@ export const JestChaiExpect: ChaiPlugin = (chai, utils) => {
...
+        actual.contains(item),
+        `expected "${actual.value}" to contain "${item}"`,
+        `expected "${actual.value}" not to contain "${item}"`,
+        expectedClassList,
+        actual.value,
+      )
+    }
+    // make "actual" indexable to have compatibility with jest
+    if (actual != null && typeof actual !== 'string')
+      utils.flag(this, 'object', Array.from(actual as Iterable<unknown>))
```
charkour, line 205: I think this `Array.from` is causing bugs. If `actual` is an object, then it will always return an empty array which is an issue.

Issue #4150, opened 2023-09-19
Custom coverage reporters support


akcyp:
### Clear and concise description of the problem

Context: I'm planning to create custom istanbul reporter based on official html reporter.

Problem 1: Custom reporters config error:
Currently vitest allow only official istanbul reporters
![image](https://github.com/vitest-dev/vitest/assets/67923777/bcc8c31b-33ca-4f65-895d-83ab8e76aada)
However we can override this by disabling typescript check for next line
![image](https://github.com/vitest-dev/vitest/assets/67923777/aa31b340-a400-47d4-8987-cd94a37ed36a)
... and this works pretty good because istanbul is using `require()` to handle this scenario https://github.com/istanbuljs/istanbuljs/blob/master/packages/istanbul-reports/index.js#L19

Problem 2: No possibility to replace coverage page in `@vitest/ui` 
![image](https://github.com/vitest-dev/vitest/assets/67923777/3dc30029-0c6c-497a-af7c-f695ee6ebb41)
I would like to use report generated from custom reporter under `@vitest/ui` coverage page, but seems like currently it's not possible to do.

### Suggested solution

1. Replace
```ts
type CoverageReporter = keyof ReportOptions;
// with
type CoverageReporter = keyof ReportOptions | string;
```
2. Add necessary logic with following option in vitest config:
```diff
export default defineConfig({
+ ui: {
+  enabled: true,
+  coverageReportProvider: 'custom-coverage-reporter' // default "html"
+ },
});
```

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

AriPerkkio:
I think this would be great feature. I'm not yet sure what would be the best way to handle the `@vitest/ui` part though.  Currently it's simply checking whether `reporters` includes `html` reporter.

akcyp:
I created a draft PR: https://github.com/vitest-dev/vitest/pull/4270 could someone take a look and see if this is the right direction?
Custom coverage reporter I'm planning to use: https://github.com/akcyp/coverage-pretty-html-reporter

cenfun:
[...]

PR #4270, opened 2023-10-08
Enable custom coverage reporters & ui coverage provider

akcyp:
### Description

This PR solves https://github.com/vitest-dev/vitest/issues/4150
What changed:
- (deprecated) `open` option - use `ui.open` instead
- (depreacted) `uiBase` option - use `ui.baseURL` instead
- added new `ui.enabled` option as alternative to *boolean* `ui` option
- [!] added new `ui.coverageProvider` option to enable using preview of custom coverage reporters (instead of harcoded html reporter)
- [!] added `string` type as correct reporter name

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/types/coverage.ts
+++ b/packages/vitest/src/types/coverage.ts
@@ -52,14 +52,11 @@ export interface CoverageProviderModule {
   stopCoverage?(): unknown | Promise<unknown>
 }
 
-export type CoverageReporter = keyof ReportOptions
+type CoverageReporterWithOptions = {
+  [key in keyof ReportOptions]: [key, ReportOptions[key] extends never ? {} : Partial<ReportOptions[key]>]
+}[keyof ReportOptions] | [string, Record<string, any>]
 
-type CoverageReporterWithOptions<ReporterName extends CoverageReporter = CoverageReporter> =
-   ReporterName extends CoverageReporter
-     ? ReportOptions[ReporterName] extends never
-       ? [ReporterName, {}] // E.g. the "none" reporter
-       : [ReporterName, Partial<ReportOptions[ReporterName]>]
-     : never
+export type CoverageReporter = CoverageReporterWithOptions[0]
 
 type Provider = 'v8' | 'istanbul' | 'custom' | undefined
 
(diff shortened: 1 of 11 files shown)


packages/vitest/src/types/coverage.ts
```diff
@@ -52,14 +52,11 @@ export interface CoverageProviderModule {
...
-   ReporterName extends CoverageReporter
-     ? ReportOptions[ReporterName] extends never
-       ? [ReporterName, {}] // E.g. the "none" reporter
-       : [ReporterName, Partial<ReportOptions[ReporterName]>]
-     : never
+type CoverageReporterWithOptions = {
+  [key in keyof ReportOptions]: [key, ReportOptions[key] extends never ? {} : Partial<ReportOptions[key]>]
+}[keyof ReportOptions] | [string, Record<string, any>]
+
+export type CoverageReporter = CoverageReporterWithOptions[0]
```
AriPerkkio, line 59: Do these changes still pass the type checking tests in `test/coverage-test`? This part was quite tricky to get right.

Run `pnpm test:types` in `vitest/test/coverage-test`.

AriPerkkio (commented):
I think it would be better to split this PR into two: one for the `coverage.repoter` change and another one for UI parts.

packages/vitest/src/types/coverage.ts
akcyp: This one is passing:
```ts
type CoverageReporterWithOptions = {
  [key in keyof ReportOptions]: [key, ReportOptions[key] extends never ? {} : Partial<ReportOptions[key]>]
}[keyof ReportOptions]
```
But when we add ` | [string, Record<string, any>]` it's not.
Unfortunately  I'm not sure if it's possible to solve because typescript doesn't have "string except ..." type.
image
This is happening because `['html', { blockName: 'string' }]` is assignable to `[string, Record<string, any>]` which should be a type **only** for custom reporter.

However this one looks promising:
```ts
type CoverageReporterWithOptions = {
  [key in keyof ReportOptions]: [key, ReportOptions[key] extends never ? {} : Partial<ReportOptions[key]>]
}[keyof ReportOptions] | [string, Record<string, never>]
```
With this one we cannot pass custom properties to custom reporters but still, it's the best solution.
image
Tests only complains about unknown reporters - but it's expected behavior, I mean tests should be adjusted.

What you think?

sheremet-va:
The linked issue is resolved so I assume we can close this PR. Feel free to open a new one if it's not the case.

PR #4279, opened 2023-10-09
feat: add `--exclude` CLI flag

### LINKED ISSUE #4148: Allow to exclude test files with cli

I have the following use case for an exclude flag in the vitest cli:

We have a codebase with some unit tests using vitest in `./src/`
Alongside this code base there are some integration tests sharing the dto interfaces from `./src/`. The integration tests are in `./tests/api/` and use their own vitest config in `./tests/api/vitest.config.ts`.

We use `lint-staged` to execute relevant tests with `vite related` like the [docs suggest](https://vitest.dev/guide/cli.html#vitest-related).

However - those integration tests should not run in our commit hook due to their fragile nature. So I'd like to pass an exclude flag like so `vitest related --exclude './tests/api/**'`.

I couldn't find a solution with the already available CLI options. `-t` will only work on test names, but I want to exlude *files*.

### Suggested solution

Introduce an `--exclude` that'll exclude all tests matching a regex pattern in this CLI run. It'll override any exclude/include pattern from the vitest config files.

### Alternative

_No response_

### Additional context

_No response_

### Validations

Namchee:
### Description

This pull request adds a new flag `--exclude` as a CLI option that allows users to specify extra glob patterns to be ignored from tests. By default, it will merge the globs with existing default excludes.

Closes #4148 

### Tests

### Documentation

### Changesets


packages/vitest/src/node/cli.ts
```diff
@@ -151,6 +153,9 @@ function normalizeCliOptions(argv: CliOptions): CliOptions {
   else
     delete argv.dir
 
+  if (argv.exclude)
+    argv.exclude = [...defaultExclude, ...toArray(argv.exclude)]
```
sheremet-va, line 157: I don't think we should do that. We probably want to extend user's `exclude` pattern, not the default one. Maybe we can just move it into another field (`extraExclude`/`cliExclude`) and spread it in the `vitest/src/config.ts` file?

sheremet-va (changes requested):
It would also be nice if you can provide a test case for this feature. I don't think we have tests specifically for CLI commands, but you can create one.

Namchee:
Good idea, will create a test for CLIs later.

packages/vitest/src/node/cli.ts
Namchee: Solved by ae05f63

Namchee:
@sheremet-va , I think I have difficulties creating the CLI test since I don't know how to access the config from the running `vitest` instance. Do you have a pointer regarding this?

sheremet-va:
> I think I have difficulties creating the CLI test since I don't know how to access the config from the running vitest instance. Do you have a pointer regarding this?

You can see an example of a CLI test in #4399

packages/vitest/src/node/config.ts
```diff
@@ -175,6 +175,9 @@ export function resolveConfig(
       resolved.server.deps![option] = resolved.deps[option] as any
   })
 
+  if (options.exclude)
+    resolved.exclude.push(...options.exclude)
```
sheremet-va, line 179: I don't think this works as you expect. On line 84 `resolved` is created - you can see that options from CLI will just override user config.

This is why I proposed storing the value in something like `cliExcludes` and remove `excludes` before merging into resolved

sheremet-va (changes requested):

Namchee:
It seems like the test utils cannot navigate to the test fixtures correctly. Will debug this issue later.

packages/vitest/src/node/config.ts
Namchee: Solved by 9d382a3

sheremet-va (approved):

PR #4282, opened 2023-10-10
fix(config): type issue of `pool` and `poolMatchGlobs` in defineConfig

InfiniteXyy:
### Description

Fixed a type issue in the config file, where pool and poolMatchGlob should use `Exclude<T>` to exclude type `"browser"` instead of `Omit<T>`.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/defaults.ts
+++ b/packages/vitest/src/defaults.ts
@@ -62,7 +62,7 @@ const config = {
   watch: !isCI,
   globals: false,
   environment: 'node' as const,
-  pool: 'threads',
+  pool: 'threads' as const,
   clearMocks: false,
   restoreMocks: false,
   mockReset: false,
--- a/packages/vitest/src/node/config.ts
+++ b/packages/vitest/src/node/config.ts
@@ -398,9 +398,6 @@ export function resolveConfig(
   return resolved
 }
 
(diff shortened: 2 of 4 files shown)


AriPerkkio (commented):
Great finding @InfiniteXyy - I think I've mixed the `Omit` and `Exclude` couple of times before. 

The CI error on `packages/vitest/src/node/config.ts` seems valid. `poolMatchGlobs` does not allow `browser` so I wonder why `isBrowserEnabled` even checks for it. 🤔

sheremet-va:
> I wonder why `isBrowserEnabled` even checks for it. 🤔

It was allowed at the time the function was written. Now the check can be removed.

AriPerkkio (approved):

InfiniteXyy:
> Great finding @InfiniteXyy - I think I've mixed the `Omit` and `Exclude` couple of times before.
> 
> The CI error on `packages/vitest/src/node/config.ts` seems valid. `poolMatchGlobs` does not allow `browser` so I wonder why `isBrowserEnabled` even checks for it. 🤔

Thank you! I removed this check, but I'm not sure if it will affect some exsiting codebase.

sheremet-va (approved):

PR #4296, opened 2023-10-12
feat!: bump minimum node version to 18 and match Vite 5 requirement

ghiscoding:
### Description

- this PR bumps Node (and `@types/node`) to 18+ or 20+ which is the same as Vite 5 new minimum requirement
- this change was announced in the Vitest 1.0 roadmap but was not yet bumped since 1.0 beta came out, this PR fixes that

  - the reference is Vitest 1.0 roadmap 

**Note** 
~I ran `pnpm install` but there were too many changes, so I looked at my change in particular and applied it manually to the `pnpm-lock.yaml` to avoid having too many unrelated changes in the lock file

### Tests

### Documentation

### Changesets

I used the exact same PR title that Vite used in this Vite 5 [PR](https://github.com/vitejs/vite/commit/2c1a45c86cab6ecf02abb6e50385f773d5ed568e)

### Why create this PR?

I created this PR because I noticed that after upgrading to 1.0 Beta, my pnpm lock file changed in my project but was still showing a minimum requirement of Node 14 even though I did see that it was mentioned in the 1.0 roadmap that this will be bumped to 18 to match Vite 5 as shown below

> One thing that should be mentioned is that Vitest 1.0 release will drop support for Vite 4 and Vite 3 since it relies on `optimizeDeps` having `noDiscovery` option. This also means Vitest will not support Node.js 16 anymore.

So, I went with creating this PR and used exactly what Vite 5 has set in their Node engine requirement [here](https://github.com/vitejs/vite/blob/40fd2d9bf4073420e6c334f48dc3b63558b688ce/packages/vite/package.json#L49-L51)

> `"node": "^18.0.0 || >=20.0.0"`
and `@types/node` peerDependencies [here](https://github.com/vitejs/vite/blob/40fd2d9bf4073420e6c334f48dc3b63558b688ce/packages/vite/package.json#L145-L146)

> `"@types/node": "^18.0.0 || >=20.0.0"`

### Extra Notes

Please note that the CI workflow(s) are still being tested against Node 16 and 18, this should probably be changed to Node 18, 20 but that should probably be done in a separate PR (for example this PR #3492 was opened a while ago but never merged)

https://github.com/vitest-dev/vitest/blob/edddc8ff76325d7a85b98df74a935dbdfc58d86b/.github/workflows/ci.yml#L57-L60


ghiscoding:
@sheremet-va I'm not sure how you usually update the pnpm lock file but the case of this PR after running `pnpm install`, the lock file changed a lot and all CI tasks fails. I could remove the lock file change from the PR or you could take over the PR and update it yourself, feel free to go ahead with changes or let me know if a special pnpm command must be run to fix this problem (I mean I could go by hand to update the lock file but I would still prefer to do it through pnpm lol)

Dunqing:
> EDIT 2
> 
> I think some tests are failing because `1.0.0-beta.2` was not published to npm (I've opened a separate issue #4297 for that), as we can see in below print screen
> 
> ```shell
> test/setup test:  ❯ tests/console-setup.test.ts  (1 test | 1 failed) 1395ms
> test/setup test:    ❯ tests/console-setup.test.ts > setup files console > print stdout and stderr correctly
> test/setup test:      → expected '\n RUN  v1.0.0-beta.2 D:/a/vitest/vit…' to include 'stdout | setupFiles/console-setup.ts'
> test/core test:  ✓ |core| test/lot-of-tests.test.ts  (1 test) 4ms
> test/setup test:  ✓ tests/empty-setup.test.ts  (1 test) 4ms
> test/core test:  ✓ |core| test/sequential.test.ts  (2 tests) 147ms
> test/setup test:  ✓ tests/fixtures/console.test.ts  (1 test) 4ms
> test/setup test: ⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯
> test/setup test:  FAIL  tests/console-setup.test.ts > setup files console > print stdout and stderr correctly
> test/setup test: AssertionError: expected '\n RUN  v1.0.0-beta.2 D:/a/vitest/vit…' to include 'stdout | setupFiles/console-setup.ts'
> ```

I created #4245 to fix this test case failing

packages/vite-node/package.json
```diff
@@ -74,7 +74,7 @@
     "*.mjs"
   ],
   "engines": {
-    "node": ">=v14.18.0"
+    "node": "^18.0.0 || >=20.0.0"
```
sheremet-va, line 77: 1. Why can't we use `>=18.0.0`? 
2. We should probably use 18.12.0 as it was the first LTS release (https://nodejs.org/ru/blog/release/v18.12.0)

sheremet-va (changes requested):

packages/vite-node/package.json
ghiscoding: @sheremet-va  I copy+paste from Vite (see [here](https://github.com/vitejs/vite/blob/40fd2d9bf4073420e6c334f48dc3b63558b688ce/packages/vite/package.json#L49-L51)), I don't mind changing it though, I'm not sure why they decided to go with `^18.0` instead of `>=18`, I would also prefer the latter. Just tell me what you want to use and I could update it

sapphi-red: Vite uses `^18.0.0 || >=20.0.0` to explicitly say Vite doesn't support Node 19 (EOLed in this june).

ghiscoding: thanks @sapphi-red I keep forgetting about these odd versions, @sheremet-va I have reapplied the same as Vite. Ready for final review. Thanks

ghiscoding:
@sheremet-va is there anything left to do in this PR for it to be merged or are you just waiting for the next beta release to merge it? I'm just wondering if there's anything I need to change. Thanks

Issue #4298, opened 2023-10-13
Object.freeze breaks toEqual


JonathanMEdwards:
### Describe the bug

Object.freeze causes toEqual to trap because it is trying to mutate my data. This only happens when it actually finds a difference. 

### Reproduction

import { expect, test } from 'vitest'
test('vitest trap', () => {
  const x = {a: { b: { c: 1 }}}
  const y = {a: { b: { c: 2 }}}
  Object.freeze(x.a);
  expect(x).toEqual(y);
})

Throws the error:
TypeError: Cannot assign to read only property 'b' of object '#<Object>'

### System Info

```shell
System:
    OS: macOS 14.0
    CPU: (10) arm64 Apple M1 Pro
    Memory: 2.24 GB / 16.00 GB
    Shell: 3.3.1 - /usr/local/bin/fish
  Binaries:
    Node: 20.8.0 - /usr/local/bin/node
    npm: 10.1.0 - /usr/local/bin/npm
  Browsers:
    Brave Browser: 87.1.18.78
    Chrome: 117.0.5938.149
    Safari: 17.0
  npmPackages:
    @vitejs/plugin-react-swc: ^3.4.0 => 3.4.0
    vite: ^4.4.9 => 4.4.9
    vitest: ^0.34.6 => 0.34.6
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #4303, opened 2023-10-13
fix(jest-expect): `Object.freeze` breaks `toEqual`

Dunqing:
### Description

close: #4298

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/utils/src/error.ts
+++ b/packages/utils/src/error.ts
@@ -163,8 +163,11 @@ export function replaceAsymmetricMatcher(actual: any, expected: any, actualRepla
         actualReplaced,
         expectedReplaced,
       )
-      actual[key] = replaced.replacedActual
-      expected[key] = replaced.replacedExpected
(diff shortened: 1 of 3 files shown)


packages/utils/src/error.ts
```diff
@@ -163,8 +163,11 @@ export function replaceAsymmetricMatcher(actual: any, expected: any, actualRepla
         actualReplaced,
         expectedReplaced,
       )
-      actual[key] = replaced.replacedActual
-      expected[key] = replaced.replacedExpected
+      if (actualValue !== replaced.replacedActual)
```
sheremet-va, line 166: But what if it's frozen but inner value is different?

sheremet-va (changes requested):

packages/utils/src/error.ts
Dunqing: I found out the root cause of the problem and won't have that concern now

sheremet-va (approved):

PR #4306, opened 2023-10-14
fix: address conflicts in port usage for browser-related test suites.

marpme:
### Description

When running browser tests in parallel, there is a potential issue where multiple instances may attempt to use the same port for communication. This port conflict can lead to scenarios where a test instance communicates with the wrong port, causing inaccurate results or failures in test execution.

### Solution

To mitigate the issue, this pull request introduces a solution that ensures each test instance receives a unique port for communication. This is achieved by assigning a constant port value to each instance, preventing conflicts and enhancing the reliability of parallel test execution.

- [x] ~It's really useful if your PR references an issue where it is discussed ahead of time. If the feature is substantial or introduces breaking changes without a discussion, PR might be closed.~ Although it lacks an issue reference, it addresses an outstanding TODO in the code.

### Tests

### Documentation

### Changesets

### DIFF
--- a/examples/playwright/test/basic.test.ts
+++ b/examples/playwright/test/basic.test.ts
@@ -5,6 +5,8 @@ import { chromium } from 'playwright'
 import type { Browser, Page } from 'playwright'
 import { expect } from '@playwright/test'
 
+const PORT = 3001
+
 // unstable in Windows, TODO: investigate
 describe.runIf(process.platform !== 'win32')('basic', async () => {
   let server: PreviewServer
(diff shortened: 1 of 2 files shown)


examples/playwright/test/basic.test.ts
```diff
@@ -5,15 +5,17 @@ import { chromium } from 'playwright'
 import type { Browser, Page } from 'playwright'
 import { expect } from '@playwright/test'
 
+const PORT = 3001
```
antfu, line 8: In addition, we could use https://www.npmjs.com/package/get-port-please to get the available port based on the env. 3000 and 3001 can still be too common sometimes.

antfu (approved):
In general LGTM

examples/playwright/test/basic.test.ts
marpme: Thank you for your input; I agree that it's a valid point. However, for the sake of keeping the examples as simple as possible, I would suggest maintaining their simplicity. 

The issue of port overlap arises primarily when running multiple test suites simultaneously, which is barely the case for projects utilizing vitest outside this repository. _Or maybe it isn't, then please feel free to correct me if I'm judging incorrectly_

dammy001 (approved):

PR #4309, opened 2023-10-15
fix!(coverage): use `transformMode` and workspace project based source maps

AriPerkkio:
### Description

- Fixes https://github.com/vitest-dev/vitest/issues/3888
- Fixes https://github.com/vitest-dev/vitest/issues/3251

This introduces breaking change as coverage providers now expect Vitest to pass `transformMode` in `AfterSuiteRunMeta`. But if this is included in v1 releases/beta releases, users won't notice anything as long as they don't mix `vitest@v1` and `@vitest/coverage-<provider>@0.x.x` packages.

I also ran into https://github.com/bcoe/c8/issues/172 in `test/workspaces`. V8 reports two different counts of blocks (branches) when run with different inputs.

https://github.com/vitest-dev/vitest/blob/9112cc96b4fe7f9db096ca83bee81ec7165e0710/test/workspaces/src/math.ts#L2-L9

This can't be fixed on our side - at least not without AST analysis or with hacky solutions like https://github.com/istanbuljs/istanbuljs/issues/531#issuecomment-656826893. As solution for now I simply set `test/workspaces` to use `coverage.provider: 'istanbul'`. Below is minimal reproduction of this issue showing how raw V8 reports show different count of branches:

<details>

image

</details>

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/coverage-istanbul/src/provider.ts
+++ b/packages/coverage-istanbul/src/provider.ts
@@ -6,7 +6,7 @@ import { BaseCoverageProvider } from 'vitest/coverage'
 import c from 'picocolors'
 import libReport from 'istanbul-lib-report'
 import reports from 'istanbul-reports'
-import type { CoverageMap } from 'istanbul-lib-coverage'
+import type { CoverageMap, CoverageMapData } from 'istanbul-lib-coverage'
 import libCoverage from 'istanbul-lib-coverage'
 import libSourceMaps from 'istanbul-lib-source-maps'
 import { type Instrumenter, createInstrumenter } from 'istanbul-lib-instrument'
@@ -45,7 +45,7 @@ export class IstanbulCoverageProvider extends BaseCoverageProvider implements Co
    * If storing in memory causes issues, we can simply write these into fs in `onAfterSuiteRun`
    * and read them back when merging coverage objects in `onAfterAllFilesRun`.
    */
-  coverages: any[] = []
+  coverages: Record<AfterSuiteRunMeta['transformMode'], CoverageMapData[]> = { ssr: [], web: [] }
 
   initialize(ctx: Vitest) {
     const config: CoverageIstanbulOptions = ctx.config.coverage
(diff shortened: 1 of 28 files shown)


AriPerkkio:
I marked this PR to fix https://github.com/vitest-dev/vitest/issues/3251 as well. Just to be sure - @sheremet-va can two workspace projects use different Vite plugins? Or are Vite plugins the same for each project? Only the test environment can cause different transforms between projects?

packages/coverage-istanbul/src/provider.ts
```diff
@@ -106,36 +106,47 @@ export class IstanbulCoverageProvider extends BaseCoverageProvider implements Co
...
-      return map
-    }, libCoverage.createCoverageMap({}))
+    const coverageMaps = await Promise.all([
+      mergeAndTransformCoverage(this.coverages.ssr),
+      mergeAndTransformCoverage(this.coverages.web),
+    ])
+
+    if (this.options.all && allTestsRun) {
+      // Try to guess which transform mode should be used to uncovered files
+      const transformMode = this.coverages.ssr.length > this.coverages.web.length ? 'ssr' : 'web'
```
AriPerkkio, line 136: Picking correct transform mode for uncovered files should not be that important. But is there a better way to do this? At this point we know the filenames, but they have not been run through Vite.

sheremet-va: We can import a matching environment and take it there I guess.

sheremet-va:
> can two workspace projects use different Vite plugins?

Yes, they can and they do.

AriPerkkio:
> > can two workspace projects use different Vite plugins?
> 
> Yes, they can and they do.

I see, `vitest.workspace.ts` can even contain inline plugins like:

```ts
  { test: { name: 'Project #1' }, plugins: [/* Project specific plugins */] },
  { test: { name: 'Project #2' }, plugins: [/* Project specific plugins */] },
])
```

I'll remove https://github.com/vitest-dev/vitest/issues/3251 from linked issues. For that we'll need to pass the project name or some kind of ID in the `AfterSuiteRunMeta`.

packages/coverage-istanbul/src/provider.ts
AriPerkkio: How would that work in practice?

Looks like there's also a method on Vite-node for this: `project.vitenode.getTransformMode`. But even here we would first need to figure out which project would use this uncovered file. And if multiple ones do, there's no clear way to decide which one to use. 

I think it's OK to leave this as is.

AriPerkkio: `groupFilesByEnv` looks promising. But still, we would need to figure out how to pick `project` for a file that has not been run. Anything that should be considered there? Simply pick the first `project` from `ctx.projects`?

sheremet-va: I think for those cases we should just use `Vites.runner`. We already document that untested files are loaded with a default config file: https://vitest.dev/guide/workspace.html#coverage

AriPerkkio: `ctx.runner` seems to use `ctx.vitenode` underneath. 

https://github.com/vitest-dev/vitest/blob/23968580cecde0f8a4cd3caa54c894ca721fd033/packages/vitest/src/node/core.ts#L110-L116

Switched providers to use `ctx.vitenode.transformRequest` and `vitenode.fetchCache` for uncovered files now.

AriPerkkio:
Added fix for #3251 as well. `test/workspaces` contains good test case for this where two projects run identically but transform the source file differently.

sheremet-va (approved):

Issue #4304, opened 2023-10-13
Support old browsers in @vitest/browser


edoardocavazza:
### Clear and concise description of the problem

Right now, the `@vitest/browser` is built using default vite targets. Since the client library is served through a sirv middleware, it does not get transpiled by the vite server during tests. This lead old browsers without optional chaining support to a syntax error.

### Suggested solution

Would you consider to build `@vitest/browser` using the lowest esbuild target available?

### Alternative

Is it possibile to make vite to serve the client library without sirv?

### Additional context

I am building a provider for Saucelabs browsers.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

sheremet-va:
Should be fine as long as they support import syntax.

sheremet-va:
Actually, `@vitest/browser` assets are now served by the Vite server, not `sirv` so they follow the Vite target.

PR #4320, opened 2023-10-17
fix(browser): support more browsers

edoardocavazza:
### Description

This PR solves #4304 adding the `es2017` target to builds and polyfilling `globalThis` in html entrypoints.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/browser/src/client/index.html
+++ b/packages/browser/src/client/index.html
@@ -25,9 +25,6 @@
   <body>
     <iframe id="vitest-ui" src=""></iframe>
     <script>
-      if (typeof globalThis === 'undefined')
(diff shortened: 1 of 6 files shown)


sheremet-va:
We should not polyfill anything by default.

edoardocavazza:
I see. I've removed the polyfill for globalThis, which will narrow down the list of supported browsers a bit, particularly excluding Safari versions up to 12.2, but I think it should still work.

sheremet-va (changes requested):
Should we also add spy package here?

sheremet-va:
#4304 is closed

PR #4338, opened 2023-10-19
feat: filter stacktraces (fix #1999)

### LINKED ISSUE #1999: Ability to filter the stack trace

In some cases, stack traces contain a bunch of lines from third party packages, and I would prefer to omit those lines from the stack trace, as it's extremely unlikely that I need to be aware of them.

### Suggested solution

Add an option like `filterStack` that takes a function that receives a file path and returns a boolean to indicate whether the stack frame should be removed from the trace. This is what I do in Saus (my Vite-based framework), and [it works great](https://github.com/alloc/saus/blob/next/src/core/vm/formatAsyncStack.ts#L63).

### Alternative

_No response_

### Additional context

_No response_

### Validations

clarkf:
### Description
This change introduces a new optional configuration parameter, `onStackTrace`. If included, each frame of each error stacktrace encountered during the test run will be tested by the provided function. If the test fails, the frame will be omitted from the displayed trace.

The only real note is that this might not give enough information for more complex situations (e.g. filtering that depends on the error, or the other frames).

Fixes #1999.

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -1785,6 +1785,34 @@ export default defineConfig({
 })
 ```
 
+### onStackTrace
+
+- **Type**: `(error: Error, frame: ParsedStack) => boolean | void`
+- **Version**: Since Vitest 1.0.0-beta.3
+
+Apply a filtering function to each frame of each stacktrace when handling errors. The first argument, `error`, is an object with the same properties as a standard `Error`, but it is not an actual instance.
+
+Can be useful for filtering out stacktrace frames from third-party libraries.
+
+```ts
+import type { ParsedStack } from 'vitest'
+import { defineConfig } from 'vitest/config'
+
+export default defineConfig({
+  test: {
+    onStackTrace(error: Error, { file }: ParsedStack): boolean | void {
+      // If we've encountered a ReferenceError, show the whole stack.
+      if (error.name === 'ReferenceError')
+        return
+
+      // Reject all frames from third party libraries.
+      if (file.includes('node_modules'))
+        return false
+    },
+  },
+})
+```
+
 ### diff
 
 - **Type:** `string`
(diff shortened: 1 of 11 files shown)


sheremet-va:
> The only real note is that this might not give enough information for more complex situations (e.g. filtering that depends on the error, or the other frames).

I think it would be nice to provide an error object as the first argument (or second?)

clarkf:
That would be my instinct too. `(error: Error, frame: ParsedStack) => boolean` makes the most sense to me. I'll make the change.

docs/config/index.md
```diff
@@ -1785,6 +1785,33 @@ export default defineConfig({
...
+Can be useful for filtering out stacktrace frames from third-party libraries.
+
+```ts
+import type { ParsedStack } from 'vitest'
+import { defineConfig } from 'vitest/config'
+
+export default defineConfig({
+  test: {
+    onStackTrace(error: Error, { file }: ParsedStack): boolean | void {
+      // If we've encountered a ReferenceError, show the whole stack.
```
sheremet-va, line 1803: Error object is always a plain object because we cannot transfer instance of error, so this is misleading. This should also be probably mentioned in docs

docs/config/index.md
```diff
@@ -1785,6 +1785,33 @@ export default defineConfig({
 })
 ```
 
+### onStackTrace
+
+- **Type**: `(error: Error, frame: ParsedStack) => boolean | void`
```
sheremet-va, line 1790:
```suggestion
- **Type**: `(error: Error, frame: ParsedStack) => boolean | void`
- **Version**: Since Vitest 1.0.0-beta.3
```

sheremet-va (changes requested):

sheremet-va (approved):

PR #4346, opened 2023-10-20
fix: coverage.100 crash when using as an cli argument

### LINKED ISSUE #4307: --coverage.100 crashes with TypeError
### Describe the bug

`vitest run --coverage.100` appears to be broken – it immediately crashes with `TypeError: Cannot read properties of undefined (reading 'push')`.

### Reproduction

```
❯ npx -y vitest run --coverage.100
TypeError: Cannot read properties of undefined (reading 'push')
```

### System Info

```shell
System:
    OS: macOS 13.6
    CPU: (10) arm64 Apple M1 Pro
    Memory: 2.06 GB / 32.00 GB
    Shell: 3.6.1 - /opt/homebrew/bin/fish
  Binaries:
    Node: 20.5.0 - ~/.local/share/nvm/v20.5.0/bin/node
    npm: 9.8.0 - ~/.local/share/nvm/v20.5.0/bin/npm
    pnpm: 8.8.0 - /opt/homebrew/bin/pnpm
  Browsers:
    Chrome: 118.0.5993.70
    Safari: 16.6
```

### Used Package Manager

npm

### Validations

marcelobotega:
### Description

Set a default value for the option --coverage to fix a crash that happen when using the option --coverage.100
CAC is not handling a key with only numbers alone.
Also add an coverage option check to add the value `enabled: true` when using a dotted coverage option.

Fixes #4307 

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/cli-api.ts
+++ b/packages/vitest/src/node/cli-api.ts
@@ -44,6 +44,12 @@ export async function startVitest(
   if (typeof options.coverage === 'boolean')
     options.coverage = { enabled: options.coverage }
 
+  // running "vitest --coverage.all" or "vitest --coverage.100"
+  // @ts-expect-error Element implicitly has an any type for options.coverage["100"]
+  if (typeof options.coverage === 'object' && !('enabled' in options.coverage) && (options.coverage['100'] === true
+    || Object.keys(options.coverage).length > 1))
+    options.coverage.enabled = true
+
   // running "vitest --browser", assumes browser name is set in the config
   if (typeof options.browser === 'boolean')
     options.browser = { enabled: options.browser } as any
--- a/packages/vitest/src/node/cli.ts
+++ b/packages/vitest/src/node/cli.ts
@@ -25,7 +25,7 @@ cli
   .option('--hideSkippedTests', 'Hide logs for skipped tests')
   .option('--reporter <name>', 'Specify reporters')
   .option('--outputFile <filename/-s>', 'Write test results to a file when supporter reporter is also specified, use cac\'s dot notation for individual outputs of multiple reporters')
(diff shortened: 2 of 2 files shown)


packages/vitest/src/node/cli-api.ts
```diff
@@ -44,6 +44,12 @@ export async function startVitest(
   if (typeof options.coverage === 'boolean')
     options.coverage = { enabled: options.coverage }
 
+  // running "vitest --coverage.all" or "vitest --coverage.100"
+  // @ts-expect-error Element implicitly has an any type for options.coverage["100"]
+  if (typeof options.coverage === 'object' && !('enabled' in options.coverage) && (options.coverage['100'] === true
+    || Object.keys(options.coverage).length > 1))
+    options.coverage.enabled = true
+
```
AriPerkkio, line 52: So this sets `coverage.enabled` when any coverage options are passed as CLI arguments? We don't do that for any other options either so it should not be done here either.

marcelobotega: It's done on the browser check [here](https://github.com/vitest-dev/vitest/blob/aa6b6f501889f0255ed9b00375605c257b00b830/packages/vitest/src/node/cli-api.ts#L62) that's why I added it =P
But I'll update it. 
Should I update the docs to let the user know that he needs to use the --coverage.enabled when using other coverage options ?

AriPerkkio: Let's leave this change out of this PR and include only the fix for the crash. 
This change might be good to do in another PR though.

AriPerkkio (approved):

sheremet-va (approved):

PR #4368, opened 2023-10-25
chore!: update vite@5 and rollup@4

Dunqing:

### Tests

### Documentation

### Changesets


packages/vitest/LICENSE.md
```diff
@@ -197,6 +197,35 @@ Repository: https://github.com/sinonjs/fake-timers.git
 
 ---------------------------------------
 
+## acorn
```
sheremet-va, line 200: Why do we bundle acorn now?

sheremet-va (changes requested):

packages/vitest/LICENSE.md
Dunqing: image

Since we removed `acorn` from the dependencies and now `external` no longer contains `acorn`, we now need to explicitly write `acorn` to `external`

sheremet-va:
I want to merge this, but PR has merge conflicts. @Dunqing can you resolve them?

Dunqing:
> I want to merge this, but PR has merge conflicts. @Dunqing can you resolve them?

Sure!

packages/vitest/rollup.config.js
```diff
@@ -57,6 +57,8 @@ const external = [
   'node:worker_threads',
   'node:fs',
   'rollup',
+  'rollup/parseAst',
+  'acorn',
```
sheremet-va, line 61: This should not be needed since we shouldn't import it directly anywhere. If it's still inlined, then we need to investigate it.

packages/vitest/rollup.config.js
```diff
@@ -57,6 +57,8 @@ const external = [
   'node:worker_threads',
   'node:fs',
   'rollup',
```
sheremet-va, line 59: I think we should remove both of these from external - they should never be bundled. Maybe there is some kind of a mechanism to throw an error if import from rollup is detected? 🤔

packages/vitest/rollup.config.js
```diff
@@ -57,6 +57,8 @@ const external = [
...
+  'rollup/parseAst',
```
sheremet-va, line 60: We shouldn't import from rollup because it's not in our dependencies. We shouldn't add it to our dependencies because we might have a version mismatch with Vite.

We might be able to trick Vite into giving us `parse` method from `transform` context somehow?

```ts
let _parse
export function ParseStorePlugin() {
  return { name: 'vitest:parse-store', transform() { _parse ??= this.parse } }
}
export function getParser() {
  return _parse
}
```

packages/vitest/package.json
```diff
@@ -139,7 +139,7 @@
     "@vitest/snapshot": "workspace:*",
     "@vitest/spy": "workspace:*",
     "@vitest/utils": "workspace:*",
-    "acorn": "^8.9.0",
+
```
sheremet-va, line 142: There should be no empty lines

sheremet-va (changes requested):

packages/vitest/rollup.config.js
Dunqing: This way seems a little strange. But I'll give it a try.

packages/vitest/rollup.config.js
Dunqing: Yeah, I'll look into that.

packages/vitest/rollup.config.js
Dunqing: I have seen some discussion related to `parseAst` in https://github.com/vitejs/vite/pull/14508#discussion_r1342340771. Looks like we should use `parseAst`.

packages/vitest/rollup.config.js
Dunqing: I figured out why this. Because the `mlly` package is place in the devDependencies, so external missing `mlly`

packages/vitest/rollup.config.js
sheremet-va: Yeah, `parseAst` is exposed on `server.pluginContext.parse` - we can use it from there without importing `rollup/parseAst`

sheremet-va: Oh, the context is not exposed there, let me look into it

sheremet-va: > I have seen some discussion related to parseAst 

This discussion doesn't matter for us because we cannot have `rollup` in dependencies (see the first comment).

Dunqing: Yes, I agree

packages/vitest/rollup.config.js
Dunqing: Good catch! The `shouldExternalize` and `isValidNodeImport` have very much the same logic, and it looks like we can just remove `isValidNodeImport`, which should be a big performance improvement!

sheremet-va: Hm, well, no, we can't really remove it, we still need to check if it's valid. But we can probably just inline this one function

Dunqing: Okay, Yes, I tried to remove `isValidNodeImport`, but very test cases failed. Now I inline this one function and remove the logic we don't need

packages/vitest/rollup.config.js
Dunqing: We just need to remove the rollup from `external`. If the rollup is bundled, the build and test will fail.

packages/vite-node/src/externalize.ts
```diff
@@ -1,10 +1,14 @@
-import { existsSync } from 'node:fs'
-import { isValidNodeImport } from 'mlly'
-import { join } from 'pathe'
+import { existsSync, promises as fsp } from 'node:fs'
+import { readPackageJSON } from 'pkg-types'
```
sheremet-va, line 2: pkg-types uses mlly, so there is really no reason to remove mlly now

we already have an optimized function to ready package.json's: I suggest you port it to vite-node from: packages/vitest/src/runtime/external-executor.ts

packages/vite-node/src/server.ts
```diff
@@ -98,7 +98,7 @@ export class ViteNodeServer {
   }
 
   shouldExternalize(id: string) {
-    return shouldExternalize(id, this.options.deps, this.externalizeCache)
```
sheremet-va, line 101: 1. This code is processed by Vite, we need to check against actual code
2. If it's processed, then it was inlined

Dunqing: Assuming that fetchCache already has a record, this means that this ·id· should be externalized, am I understanding this correctly? You can see that  https://github.com/vitest-dev/vitest/pull/4368/commits/20ed6ce9894515b65c07a8853ed56120f0e26f91

packages/vite-node/src/externalize.ts
Dunqing: I didn't realize that `pkg-types` and `mlly` depend on each other. 😅

packages/vite-node/src/server.ts
sheremet-va: If it's processed then it was already inlined yes, and this function will never even hit your new check because it's already cached. We don't need this new check at all I think.

Dunqing: Yes, I thought wrong ...

Dunqing:
I roughly compared the ci test time between this branch and the main branch, and the testing time seems to have decreased 😀

sheremet-va:
We should wait for https://github.com/vitejs/vite/pull/14833 to be released, so we can import `parseAstAsync` from vite directly

Dunqing:
> We should wait for https://github.com/vitejs/vite/pull/14833 to be released, so we can import `parseAstAsync` from vite directly

Should we specify version 5 for vite, otherwise we don't have parseAst

sheremet-va:
> > We should wait for [vitejs/vite#14833](https://github.com/vitejs/vite/pull/14833) to be released, so we can import `parseAstAsync` from vite directly
> 
> Should we specify version 5 for vite, otherwise we don't have parseAst

Yep

gajus:
FYI Vite released the changes.

Thank you for this PR @Dunqing

packages/vitest/package.json
```diff
@@ -152,7 +151,7 @@
     "strip-literal": "^1.0.1",
     "tinybench": "^2.5.0",
     "tinypool": "^0.8.1",
-    "vite": "^3.1.0 || ^4.0.0 || ^5.0.0-0",
+    "vite": "^5.0.0-0",
```
sheremet-va, line 154: This will break with beta<15. Let's have it as `^5.0.0-beta.15 || ^5.0.0` until 5.0.0 is released

Dunqing: Yeah, I think you're right

sheremet-va:
Interesting errors in your tests 🤔

Dunqing:
> Interesting errors in your tests 🤔

Looking good now. 

Recent tests seem to have some unusual errors, and not just in this branch.

gajus:
@sheremet-va Is there anything else that needs to be done here?

ghiscoding:
Vite is now at `5.0.0-beta.16`, does this PR need another update? It looks like Vite is getting closer to a release since a PR for [Vite 5 announcement blog post](https://github.com/vitejs/vite/pull/14860) was created. I guess that Vitest is not quite ready for the official 1.0 yet, right? From the Vitest 1.0 roadmap, it looks like migrating `vite-node` into Vite is the biggest task left but I assume this might not happen right away since Vite 5 is very close to be shipped?

Thanks for the great work 🎉

sheremet-va:
vite-node will not be in Vite 5, yes.

not sure what update this PR needs tho, can you elaborate?

ghiscoding:
> vite-node will not be in Vite 5, yes.
> 
> not sure what update this PR needs tho, can you elaborate?

just Vite Beta 15 to 16, nothing much

PR #4372, opened 2023-10-25
feat: add tsconfig.json to basic example

### LINKED ISSUE #3402: vitest example should have explicit `tsconfig.json`

Currently, there is only one example for pure js/ts testing, and it doesn't have a TS config.

This is an issue for new developers (to typescript or vite) as they don't have much resources to look for basic vite configuration

### Suggested solution

It would be nice if the basic example had explicit `tsconfig.json` and an `src` directory, or at least create another 'advanced' example including the aforementioned concepts and even TS references

### Alternative

_No response_

### Additional context

qol feature

### Validations

allisons11:
### Description

Resolves #3402. Added `tconfig.json` and `src` directory to the basic example to show possible configuration while keeping the example simple.

Included a few basic compiler options in the `tsconfig` (module, target, moduleResolution, strict)  -as well as the `include` and `exclude` fields - to reflect configuration that one might commonly see. Added an example of a function export to `src` to provide an example of testing a module.

### Tests

### Documentation

### Changesets


sheremet-va (changes requested):

examples/basic/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "compilerOptions": {
+    "module": "esnext",
+    "target": "esnext",
+    "moduleResolution": "node",
```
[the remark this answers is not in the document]
allisons11: Got it, thanks! 

Updated to follow the strict config, and removed the `moduleResolution` option as it's redundant with `node16` as module.

examples/basic/tsconfig.json
```diff
@@ -0,0 +1,13 @@
...
+  "compilerOptions": {
+    "module": "node16",
+    "target": "es2020",
+    "strict": true,
+    "verbatimModuleSyntax": true,
+    "declaration": true,
+    "sourceMap": true,
+    "declarationMap": true
+  },
+  "include": ["src"],
```
sheremet-va, line 11: Shouldn't it also cover tests? 🤔

sheremet-va (changes requested):

examples/basic/tsconfig.json
allisons11: good catch, definitely should - updated!

nopeless:
thanks!

Issue #4402, opened 2023-10-31
Asynchronous fixtures used concurrently cause Vitest to hang


dsyddall:
### Describe the bug

Given a test with an asynchronously resolved fixture, Vitest hangs when the fixture is used in multiple tests concurrently.

Example test:

```
import { expect, test } from 'vitest';

const myTest = test.extend<{ a: number }>({
  a: async ({ task }, use) => {
    await new Promise<void>((resolve) => setTimeout(resolve, 200));

    console.log('running test', task.id);
    await use(0);
    console.log('cleaning up test', task.id);
  },
});

myTest.concurrent('fixture - test 1', ({ a }) => {
  expect(a).toBe(0);
});

myTest.concurrent('fixture - test 2', ({ a }) => {
  expect(a).toBe(0);
});
```

Results of the test:

```
 DEV  v1.0.0-beta.3 /home/projects/vitest-dev-vitest-9d9pqh
      UI started at http://localhost:51204/__vitest__/

stdout | test/async-fixture.test.ts > fixture - test 2
running test 1084931944_0
running test 1084931944_1
cleaning up test 1084931944_1

 ❯ test/async-fixture.test.ts (2)
   ✓ fixture - test 1
   ✓ fixture - test 2
```

The first test fixture is never cleaned up, and since the promise returned from the call to use() never resolves, Vitest hangs.

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-9d9pqh?file=test%2Fasync-fixture.test.ts

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 18.18.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 9.4.2 - /usr/local/bin/npm
    pnpm: 8.9.2 - /usr/local/bin/pnpm
  npmPackages:
    @vitest/ui: 1.0.0-beta.3 => 1.0.0-beta.3 
    vite: latest => 4.5.0 
    vitest: 1.0.0-beta.3 => 1.0.0-beta.3
```

### Used Package Manager

pnpm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
[...]

PR #4403, opened 2023-10-31
fix: make asynchronous fixtures work concurrently

dsyddall:
### Description

Fixes https://github.com/vitest-dev/vitest/issues/4402

When multiple tests concurrently use an asynchronous fixture, the fixture value is not necessarily isolated between tests. This leads to a race condition and unpredictable behaviour where a fixture value created in one test may or may not be attached to the context of a separate test, sometimes causing Vitest to hang as the promise returned in the call to use() is awaited but never resolved. This PR addresses the race condition by isolating fixture values between tests.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/runner/src/fixture.ts
+++ b/packages/runner/src/fixture.ts
@@ -44,7 +44,7 @@ export function mergeContextFixtures(fixtures: Record<string, any>, context: { f
   return context
 }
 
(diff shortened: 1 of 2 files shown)


Dunqing (approved):
LGTM

PR #4417, opened 2023-10-31
feat: allow custom pools

sheremet-va:
### Description

This PR adds a possibility to specify a filepath or package name in `pool` option. Vitest expects that specified module will have a default export which is a function that accepts Vitest as its parameter. This function should return on object that satisfies this interface:

```ts
import { ProcessPool, WorkspaceProject } from 'vitest/node'

export interface ProcessPool {
  name: string
  runTests: (files: [project: WorkspaceProject, testFile: string][], invalidates?: string[]) => Promise<void>
  close?: () => Promise<void>
}
```

`runTests` is called when user runs `vitest` command and then each time watch mode is triggered. This API is primarily designed to be used by library authors because it requires deep knowledge of how Vitest works internally. More about this option can be found in documentation. 

### Tests

### Documentation

### Changesets


docs/advanced/pool.md
```diff
@@ -0,0 +1,88 @@
...
+
+::: warning
+This is advanced API. If you are just running tests, you probably don't need this. It is primarily used by library authors.
+:::
+
+Vitest runs tests in pools. By default, there are several pools:
+
+- `threads` to run tests using `node:worker_threads` (isolation is provided with a new worker context)
+- `forks` to run tests using `node:child_process` (isolation is provided with a new `child_process.fork` process)
+- `vmThreafs` to run tests using `node:worker_threads` (but isolation is provided with `vm` module instead of a new worker context)
```
AriPerkkio, line 11:
```suggestion
- `vmThreads` to run tests using `node:worker_threads` (but isolation is provided with `vm` module instead of a new worker context)
```

docs/advanced/pool.md
```diff
@@ -0,0 +1,88 @@
...
+- `browser` to run tests using browser providers
+- `typescript` to run typechecking on tests
+
+You can provide your own pool by specifying a file path:
+
+```ts
+export default defineConfig({
+  test: {
+    // will run every file with a custom pool by default
+    pool: './my-custom-pool.ts',
```
AriPerkkio, line 21: Would this also allow specifying names of published npm packages? If so, when ever we introduce new built-in pool, it would be considered as breaking change. Imagine a situation where someone publishes `vmForks` custom pool into npm before we implement that as built-in.

sheremet-va: Yes, it would be a breaking change then.

AriPerkkio: We could require the `pool` option to be prefixed with `vitest-pool-` if it's not path. Then all non-fixed would be preserved for built-in ones.

AriPerkkio: Looks like the same problem applies to custom reporters so I guess it's not an issue really.

sheremet-va:
Note: this PR just opens up a possibility to do this, we still need to improve how to run tests in the new runtime - like exposing the logic of `runtime/{worker,child,vmThreads}.ts`. But I don't think we can do this without exposing this option first.

docs/advanced/pool.md
```diff
@@ -0,0 +1,90 @@
...
+    poolMatchGlobs: [
+      ['**/*.custom.test.ts', './my-custom-pool.ts'],
+    ],
+  },
+})
+```
+
+## API
+
+The file specified in `pool` option should export a function (can be async) that accepts `Vitest` interface as its first option. This functions needs to return an object matching `ProcessPool` interface:
```
AriPerkkio, line 38:
```suggestion
The file specified in `pool` option should export a function (can be async) that accepts `Vitest` interface as its first option. This function needs to return an object matching `ProcessPool` interface:
```

docs/advanced/pool.md
```diff
@@ -0,0 +1,90 @@
...
+
+Vitest calls `runTest` when new tests are scheduled to run. It will not call it if `files` is empty. The first argument is an array of tuples: the first element is a reference to a workspace project and the second one is an absolute path to a test file. Files are sorted using [`sequencer`](/config/#sequence.sequencer) before `runTests` is called. It's possible (but unlikely) to have the same file twice, but it will always have a different project - this is implemented via [`vitest.workspace.ts`](/guide/workspace) configuration.
+
+Vitest will wait until `runTests` is executed before finishing a run (i.e., it will emit [`onFinished`](/guide/reporters) only after `runTests` is resolved).
+
+If you are using a custom pool, you will have to provide test files and their results yourself - you can reference [`vitest.state`](https://github.com/vitest-dev/vitest/blob/feat/custom-pool/packages/vitest/src/node/state.ts) for that (mose important are `collectFiles` and `updateTasks`). Vitest uses `startTests` function from `@vitest/runner` package to do that.
```
AriPerkkio, line 56:
```suggestion
If you are using a custom pool, you will have to provide test files and their results yourself - you can reference [`vitest.state`](https://github.com/vitest-dev/vitest/blob/feat/custom-pool/packages/vitest/src/node/state.ts) for that (most important are `collectFiles` and `updateTasks`). Vitest uses `startTests` function from `@vitest/runner` package to do that.
```

And the URL should eventually point at `main` instead of work branch.

packages/vitest/src/node/pool.ts
```diff
@@ -92,46 +112,63 @@ export function createPool(ctx: Vitest): ProcessPool {
...
+    const Sequencer = ctx.config.sequence.sequencer
+    const sequencer = new Sequencer(ctx)
+
+    async function sortSpecs(specs: WorkspaceSpec[]) {
+      if (ctx.config.shard)
+        specs = await sequencer.shard(specs)
+      return sequencer.sort(specs)
+    }
+
+    await Promise.all(Object.entries(filesByPool).map(async (entry) => {
```
AriPerkkio, line 128: This issue seems to be in `main` branch too, but all pools are executed parallel here, right?

If user has `threads` pool and `vmThreads` pools set, and each are using default max-threads, the execution will spawn too many workers at the same time.

docs/advanced/pool.md
```diff
@@ -0,0 +1,90 @@
...
+- `typescript` to run typechecking on tests
+
+You can provide your own pool by specifying a file path:
+
+```ts
+export default defineConfig({
+  test: {
+    // will run every file with a custom pool by default
+    pool: './my-custom-pool.ts',
+    // you can provide options using `poolOptions` object
```
AriPerkkio, line 22: While testing this I tried to use the `test.pool` here in `test.poolOptions.<key>` as key. Might be good to mention that the key has to be pool's `name` property:

```suggestion
    // you can provide options using `poolOptions` object and name of the custom pool
```

AriPerkkio (approved):
Tested a custom pool in locally linked project and seems to work nice.

packages/vitest/src/node/pool.ts
sheremet-va: Yes, this is how it works. That's why I wanted agnostic implementation of tinypool.

docs/advanced/pool.md
sheremet-va: You can use any key, actually.

AriPerkkio: Oh right, it just depends on how the pool itself is accessing the options.

packages/vitest/src/node/pool.ts
AriPerkkio: Tinypool can run `worker_threads` and `child_process` parallel. 

https://github.com/tinylibs/tinypool/blob/999778ed8acccf18303760558f807f0e208c8b3b/test/runtime.test.ts#L151-L170

It's possible to push all `threads` tasks in the pool, then wait for queue to be empty (remaining tasks are running, nothing in queue), and then change the `runtime` and push more tasks in the queue to wait for previous ones to finish. Like here is done:

https://github.com/vitest-dev/vitest/blob/ab55637653c40c47d9f6ffe694f6c2335dc73260/packages/vitest/src/node/pools/threads.ts#L171-L174

sheremet-va: Ideally, I would like an API that we can expose to reuse the same Tinypool instance in different pools:

```ts
function createThreadsPool(ctx, { pool }) {
  function runTests(testFile) {
    const context = { config: ctx.config, files: [testFile] }
    await pool.run(context, { runtime: 'worker_threads', name: 'run' })
  }
}
```

Maybe instead of having separate `worker.js/child.js/vmWorker.js` files we have a single file where inside of a context we pass down information on how we want to run it? So, we don't need to change `filename` option in Tinypool.

AriPerkkio: I like the idea. From Tinypool's side I think that kind of API should already be possible. 
There's also some need for refactoring in threads and forks pools to reduce code duplication. While doing that it might be good to look into this kind of API. But all this can be left out of this PR as it's not relevant to these changes.

PR #4422, opened 2023-11-01
feat: implement provide/inject API to transfer data from the main thread

### LINKED ISSUE #4025: Passing serializable data from globalSetup to tests
## The problem

I would like to use a single resource (say, a server with a DB) for all the tests, but I need to pass that server (in a form of hostname / port) to the tests.
I'm aware that passing actual instances to a worker is somewhere between smelly and impossible, so I want to pass a small bit of serializeable data, hostname and port in my case.

## globalSetup and the data

Unfortunately, `globalSetup` files don't get any data from the outside world (Jest counterparty [does actually get](https://jestjs.io/docs/configuration#globalsetup-string) `globalConfig` and `projectConfig` as arguments), so I made a tiny patch to fix that:
```diff
# packages/vitest/src/node/plugins/globalSetup.ts#L59-65

      try {
        for (const globalSetupFile of globalSetupFiles) {
-         const teardown = await globalSetupFile.setup?.()
+         const teardown = await globalSetupFile.setup?.(project)
          if (teardown == null || !!globalSetupFile.teardown)
            continue
          if (typeof teardown !== 'function')
# packages/vitest/src/node/plugins/globalSetup.ts#L78-84
      if (globalSetupFiles?.length) {
        for (const globalSetupFile of globalSetupFiles.reverse()) {
          try {
-           await globalSetupFile.teardown?.()
+           await globalSetupFile.teardown?.(project)
          }
          catch (error) {
            logger.error(`error during global teardown of ${globalSetupFile.file}`, error)
```

## Passing data to the test

I found that I can pass `environmentOptions` to the environment which later can inject whatever needs to be injected directly in the global scope of the test (as well as setup file, I believe it is the same global scope).
I didn't find a way to modify those `environmentOptions` other way than adding those to `configOverrides` prop of the project config, so I ended up doing that.

My setup to achieve data passing around looked like this:
`vitest.config.ts`
```ts

export default defineConfig({
  test: {
    globalSetup: "./tests/global.setup.ts",
    environment: "./tests/environment.ts",
    setupFiles: "./tests/tests.setup.ts",
    experimentalVmThreads: true,
  },
});
```

`global.setup.ts`
```ts
import type { Vitest } from "vitest";

declare module "vitest" {
  interface EnvironmentOptions {
    serverConfig: {
      port: number;
      hostname: string;
    };
  }
}

export default async (vitest: Vitest) => {
  const serverConfig = {
    // implementation is not important
    port: await getFreePort(),
    hostname: "localhost",
  };
  // Passing the data to the environment, it would be nice to have a less hacky way to do that
  vitest.configOverride.environmentOptions = { serverConfig };
  // implementation is not important
  await startServer(serverConfig);
};
```

`environment.ts`
```ts
import vm from "node:vm";
import type { Environment, EnvironmentOptions } from "vitest";
import { builtinEnvironments } from "vitest/environments";

declare global {
  // That smells as it will be visible in every global context being used in only one
  var serverConfig: EnvironmentOptions["serverConfig"];
}

export default <Environment>{
  name: "custom",
  transformMode: "ssr",
  setupVM: async (options: EnvironmentOptions) => {
    const { getVmContext, teardown } = await builtinEnvironments.node.setupVM!(
      options,
    );
    const context = getVmContext();
    const globalInContext: typeof global = vm.runInContext("this", context);
    globalInContext.serverConfig = options.serverConfig;
    return {
      getVmContext: () => context,
      teardown,
    };
  },
  // not used with experimentalVmModules
  setup: () => ({ teardown: () => {} }),
};
```

`tests.setup.ts`
```ts
import { beforeAll, beforeEach } from "vitest";

const { port, hostname } = globalThis.serverConfig;
// createClient implementation is not important
const client = createClient(`http://${hostname}:${port}`);

// .. or whichever way you need to use that passed data
beforeAll(async () => {
  const ref = await client.setupSuite();
  return async () => {
    await client.destroySuite(ref);
  }
});

beforeEach(async () => {
  const ref = await client.setupTest();
  return async () => {
    await client.destroyTest(ref);
  }
});
```

## Cutting the (environment) middleman

I didn't like that I used environment as a middleman to just pass the data to the test, so I started digging.
I found that `ResolvedConfig` could actually be found in `globalThis.__vitest_worker__.config` in the global scope of a worker, so I removed the environment and a final version of `tests.setup.ts` looks like this:
```ts
import { beforeAll, beforeEach, ResolvedConfig } from "vitest";

const config = globalThis.__vitest_worker__.config as ResolvedConfig;
const { port, hostname } = config.environmentOptions.routerConfig;
// createClient implementation is not important
const client = createClient(`http://${hostname}:${port}`);

// The same as before
...
```

### Suggested solution

I believe the main idea behind not letting data pass between a Vitest global scope and test worker global scope was to prevent leaking.
The data flow between those two scopes is not hard to do (considering Vitest not passing any data to `globalSetup` is a quirk, not a reasonable decision), so I suggest we can implement a more user-friendly way to pass it around (maybe a serializeability check can be added to add more precautionary measures).

### Alternative

Keep my hacky solution as it is.

### Additional context

There was an issue [similar](https://github.com/vitest-dev/vitest/issues/2238) to this one and the solution provided was using filesystem to pass serializable data around (`jest-puppeteer` made this suggestion popular).
I believe that's not the way (it adds one more point of potential errors), managing resources should be easier.

I don't have much experience using workers and VM contexts, but it even could be a way to pass handles around so we can have a strictly typed messaging system between Vitest global scope and test worker global scope or pass actual object handles around.

### Validations

sheremet-va:
### Description

Closes #4025

This PR adds `project.provide` API that allows transfering data from the main thread like global setup to tests. This API cannot be used to transfer data back to the main thread - users need to implement that themselves. If they want to keep communication inside Vitest, it's recommended to use [`task.meta`](https://vitest.dev/advanced/metadata.html) to transfer data back to the main thread alongside tests.

To get the data, you need to write `import('vitest').inject(key)` inside the test:

```ts
// globalSetup.ts
declare module 'vitest' {
  export interface ProvidedContext {
    dbUrl: string
  }
}

export default async ({ provide }) => {
  const db = new DB()
  const url = db.getUrl()
  provide('dbUrl', url) // has typechecking
}

// db.test.ts
import { inject } from 'vitest'
const dbUrl = inject('dbUrl') // typed as a string
const connection = new Connection(dbUrl)
```

`provide` is scoped to the project since global setup is also scoped the same way. `provide` API uses strings as keys because symbols are not transferable between threads/websocket/forks. It is possible to override values by using the same key (internally it's just an object with string keys)

You can only transfer serializable data! The rules for serialization are the same used for `task.meta`: https://vitest.dev/advanced/metadata.html

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/workspace.ts
+++ b/packages/vitest/src/node/workspace.ts
@@ -348,6 +348,7 @@ export class WorkspaceProject {
         this.typechecker?.stop(),
         this.browser?.close(),
         this.teardownGlobalSetup(),
+        () => this._provided = {},
       ].filter(Boolean))
     }
     return this.closingPromise
--- a/packages/vitest/src/node/workspace.ts
+++ b/packages/vitest/src/node/workspace.ts
@@ -8,7 +8,7 @@ import { ViteNodeServer } from 'vite-node/server'
 import c from 'picocolors'
 import type { RawSourceMap } from 'vite-node'
 import { createBrowserServer } from '../integrations/browser/server'
-import type { ArgumentsType, Reporter, ResolvedConfig, UserConfig, UserWorkspaceConfig, Vitest } from '../types'
+import type { ArgumentsType, ProvidedContext, Reporter, ResolvedConfig, UserConfig, UserWorkspaceConfig, Vitest } from '../types'
 import { deepMerge } from '../utils'
 import type { Typechecker } from '../typecheck/typechecker'
 import type { BrowserProvider } from '../types/browser'
@@ -71,6 +71,7 @@ export class WorkspaceProject {
 
   private _globalSetupInit = false
   private _globalSetups: GlobalSetupFile[] = []
+  private _provided: ProvidedContext = {}
 
   constructor(
     public path: string | number,
@@ -85,6 +86,14 @@ export class WorkspaceProject {
     return this.ctx.getCoreWorkspaceProject() === this
   }
 
+  provide = (key: string, value: unknown) => {
+    (this._provided as any)[key] = value
+  }
+
+  getProvidedContext() {
+    return this._provided
+  }
+
   async initializeGlobalSetup() {
     if (this._globalSetupInit)
       return
@@ -95,7 +104,7 @@ export class WorkspaceProject {
 
     try {
       for (const globalSetupFile of this._globalSetups) {
-        const teardown = await globalSetupFile.setup?.()
+        const teardown = await globalSetupFile.setup?.({ provide: this.provide, config: this.config })
         if (teardown == null || !!globalSetupFile.teardown)
           continue
         if (typeof teardown !== 'function')
--- a/packages/vitest/src/node/workspace.ts
+++ b/packages/vitest/src/node/workspace.ts
@@ -90,7 +90,7 @@ export class WorkspaceProject {
       structuredClone(value)
     }
     catch (err) {
-      throw new Error(`Cannot provide value because it's not serializable.`, {
+      throw new Error(`Cannot provide "${key}" because it's not serializable.`, {
         cause: err,
       })
     }
--- a/packages/vitest/src/node/workspace.ts
+++ b/packages/vitest/src/node/workspace.ts
@@ -69,8 +69,7 @@ export class WorkspaceProject {
 
   testFilesList: string[] = []
 
-  private _globalSetupInit = false
-  private _globalSetups: GlobalSetupFile[] = []
+  private _globalSetups: GlobalSetupFile[] | undefined
   private _provided: ProvidedContext = {}
 
   constructor(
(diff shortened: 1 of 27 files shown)


packages/vitest/src/node/workspace.ts
```diff
@@ -85,6 +86,14 @@ export class WorkspaceProject {
     return this.ctx.getCoreWorkspaceProject() === this
   }
 
+  provide = (key: string, value: unknown) => {
```
luixo, line 89: Should a warning be added if `value` is not serializable?
A simple one might be `strictEqual(JSON.parse(JSON.stringify(value)), value)`.

Also, it could help adding `SerializableValue` type to help user see complex structure will not work (though the calculation might be slow).

sheremet-va: No, serialization depends on the pool used. It's not json.stringify.

luixo: Should user be warned in case their data is not serializable?

sheremet-va: Vitest will throw an error during serialization

sheremet-va: I guess we can technically check with `structureEqual`

luixo: > Vitest will throw an error during serialization

Oh, so the user will be able to get a hint on what's gone wrong?
Can the test for that be added?

packages/vitest/src/runtime/execute.ts
```diff
@@ -64,7 +63,8 @@ export interface ContextExecutorOptions {
 }
 
 export async function startVitestExecutor(options: ContextExecutorOptions) {
-  const state = () => getWorkerState() || options.state
+  // @ts-expect-error injected untyped global
+  const state = () => globalThis.__vitest_worker__ || options.state
```
AriPerkkio, line 67: What's the reason for these changes?

AriPerkkio (approved):

packages/vitest/src/runtime/execute.ts
sheremet-va: getState now throws an error. here it's possible that it's not initialized (in vm for example)

antfu (approved):

PR #4431, opened 2023-11-02
docs: fix incorrect instructions for avoiding in-source tests in production builds

kare-rentelligent:
### Description

The following configuration needs to be in `vite.config.ts` instead of `vitest.config.ts`, otherwise it has no effect and the in-source unit tests are included in the production build.

```ts
define: {
  'import.meta.vitest': 'undefined', // prevent unit tests from being included in other builds
},
```

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/guide/in-source.md
+++ b/docs/guide/in-source.md
@@ -52,19 +52,18 @@ $ npx vitest
 
 ## Production build
 
-For the production build, you will need to set the `define` options in your config file, letting the bundler do the dead code elimination. For example, in Vite
+For the production build, you will need to set the `define` options in your _main_ config file, letting the bundler do the dead code elimination. For example, in Vite
 
 ```ts
-// vitest.config.ts
-import { defineConfig } from 'vitest/config'
+// vite.config.ts
+import { defineConfig } from 'vite';
 
-export default defineConfig({
-  define: { // [!code ++]
-    'import.meta.vitest': 'undefined', // [!code ++]
-  }, // [!code ++]
-  test: {
-    includeSource: ['src/**/*.{js,ts}']
-  },
+export default defineConfig(({ command, mode }) => {
+  return {
+    define: { // [!code ++]
+      'import.meta.vitest': 'undefined', // [!code ++]
+    }, // [!code ++]
+  }
 })
 ```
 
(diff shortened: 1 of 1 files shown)


sheremet-va:
I am honestly so tired that people mix configs. I think we should just update every config example in docs to use vite.config by default because it's actually _always_ Vite config (even inside vitest.config). We don't recommend creating two separate files because one overrides the other which is confusing for some people.

docs/guide/in-source.md
```diff
@@ -52,19 +52,18 @@ $ npx vitest
...
+import { defineConfig } from 'vite';
 
-export default defineConfig({
-  define: { // [!code ++]
-    'import.meta.vitest': 'undefined', // [!code ++]
-  }, // [!code ++]
-  test: {
-    includeSource: ['src/**/*.{js,ts}']
-  },
+export default defineConfig(({ command, mode }) => {
```
sheremet-va, line 61: Why did it became a function?

docs/guide/in-source.md
```diff
@@ -52,19 +52,18 @@ $ npx vitest
 
 ## Production build
 
-For the production build, you will need to set the `define` options in your config file, letting the bundler do the dead code elimination. For example, in Vite
+For the production build, you will need to set the `define` options in your _main_ config file, letting the bundler do the dead code elimination. For example, in Vite
 
 ```ts
-// vitest.config.ts
-import { defineConfig } from 'vitest/config'
+// vite.config.ts
```
sheremet-va, line 58: Lets update example higher to use vite.config

docs/guide/in-source.md
```diff
@@ -52,19 +52,18 @@ $ npx vitest
...
+import { defineConfig } from 'vite';
 
-export default defineConfig({
-  define: { // [!code ++]
-    'import.meta.vitest': 'undefined', // [!code ++]
-  }, // [!code ++]
-  test: {
```
sheremet-va, line 65: When we update the example higher, this should be returned

sheremet-va (changes requested):

sheremet-va (approved):

PR #4441, opened 2023-11-04
refactor: make a standard API to  run tests inside a worker

sheremet-va:
Instead of having several entry points for each pool, Vitest now exposes a single entry point that accepts an object with `worker` property that should export an object with `getRpcOptions` and `runTests` methods.

- `getRpcOptions` should return an object that is passed down to `birpc` - this is primarily to establish communication between the threads (Vitest exposes `createForksRpcOptions` and `createThreadsRpcOptions` from `vitest/workers` entry point)
- `runTests` receives `WorkerGlobalState` object which should be used to decide what tests to run and what the current config is (Vitest exposes `runVmTests` and `runBaseTests` methods from `vitest/workers`)

All pools workers are still bundled, but now they are located inside `workers` folder:

- forks pool: `vitest/dist/workers/forks.js` 
- threads pool: `vitest/dist/workers/threads.js` 
- vmForks pool: `vitest/dist/workers/vmForks.js` 
- vmThreads pool: `vitest/dist/workers/vmThreads.js`

`projectFiles` on Vitest instance doesn't exist anymore, instead, there is `distPath` that can be used to resolve pool paths.
 
We don't consider this to be a breaking change because Vitest API is in an experimental state and doesn't follow SemVer. `pool` option is a public API but it doesn't have to rely on experimental Vitest API to work, - it's functionality was not changed.


packages/vitest/src/types/rpc.ts
```diff
@@ -46,10 +45,13 @@ export interface ResolvedTestEnvironment {
 }
 
 export interface ContextRPC {
+  pool: Pool
+  runner: string
```
AriPerkkio, line 49: At first I thought `runner` was related to `@vitest/runner`. Maybe `worker` instead?

packages/vitest/src/runtime/workers/forks.ts
```diff
@@ -0,0 +1,15 @@
+import v8 from 'node:v8'
+import type { ContextRPC } from '../../types/rpc'
+import type { WorkerRpcOptions } from './types'
+import { BaseVitestWorker } from './base'
+import { createForksRpcOptions } from './utils'
+
+export default class ForksVitestWorker extends BaseVitestWorker {
```
AriPerkkio, line 7: This one also needs the `unwrapConfig` for `testNamePattern`.

AriPerkkio (commented):
The `runtime/workers` look good. I would imagine similar changes on main thread's side would simplify things a lot.

packages/vitest/src/integrations/vi.ts
```diff
@@ -370,7 +370,7 @@ function createVitest(): VitestUtils {
 
   const utils: VitestUtils = {
     useFakeTimers(config?: FakeTimerInstallOpts) {
-      if (workerState.isChildProcess) {
+      if (process.send) {
```
AriPerkkio, line 373: Maybe create an utility method `isChildProcess()` which describes this check better.

packages/vitest/src/runtime/worker.ts
```diff
@@ -1,94 +1,55 @@
...
 import { setupInspect } from './inspector'
-import { createSafeRpc, rpcDone } from './rpc'
+import { createRuntimeRpc, rpcDone } from './rpc'
+import type { VitestWorkerConstructor } from './workers/types'
 
-async function init(ctx: WorkerContext) {
-  // @ts-expect-error untyped global
-  const isInitialized = typeof __vitest_worker__ !== 'undefined'
-  const isIsolatedThreads = ctx.config.pool === 'threads' && (ctx.config.poolOptions?.threads?.isolate ?? true)
+export async function run(ctx: ContextRPC) {
```
AriPerkkio, line 10: Would this file become part of public API for custom pool implementers? 
They would be providing their pool which loads this file and spins up the runner?

AriPerkkio:
Something like this should work on main thread's side for single instance of Tinypool. I didn't test it but it might work as is.

```ts
import { Tinypool } from 'tinypool'

const pool = new Tinypool({
  filename: 'dist/worker.js',
  // min/max options here. Shared by all pools.
})

async function waitForEmptyQueue() {
  return new Promise<void>((resolve) => {
    if (pool.queueSize === 0) {
      return resolve()
    }

    pool.once('drain', resolve)
  })
}

const runningTasks: Promise<void>[] = [];

// TODO handle vm ones too
async function runTests(files: string[], vitestPool: 'threads' | 'forks') {
  const runtime = vitestPool === 'threads' ? 'worker_threads' : 'child_process'

  await pool.recycleWorkers({ runtime })

  const data = {
    pool,
    files,
    // ... add more options here
  }

  // At this point there may be previous tasks running on different runtime.
  // All tasks that are now pushed to queue with `run` will start with the new runtime.
  // Note that `run` is not await'ed. Empty queue will indicate finished tasks.
  runningTasks.push(pool.run(data))
  await waitForEmptyQueue()
}

type TestFile = string
const testsByPool: ['threads' | 'forks', TestFile[]][] = [
  ['threads', ['... files ...']],
  ['forks', ['... files ...']],
]

// Run tests
for (const [pool, files] of testsByPool) {
  await runTests(files, pool)
}

// Wait for last round's tasks to finish too
await Promise.all(runningTasks);

```

packages/vitest/src/integrations/vi.ts
sheremet-va: I think this check is enough. Or do you want to expose it for public use? Or just to make this assertion clearer?

sheremet-va:
> await pool.recycleWorkers({ runtime })

Is it fast though? What does this mean for a non-isolated environment? Does it mean we would need to wait to start another process/worker and then restart it again? Can we optimize it somehow? Maybe create half workers and half forks, and close/restart other runtimes (so, 2/2 -> 3/1) when they finish running?

packages/vitest/src/types/rpc.ts
sheremet-va: This is an interesting question. jest has runner and testRunner. What we call `runner` in jest is called `testRunner` 👀 

In this PR I also tried to standardize naming (so, everything is a worker, just not a threads worker). I think here it would be fine to name in a `worker`

packages/vitest/src/runtime/worker.ts
sheremet-va: Yes. Ideally, I just wanted to have a single tinypool instance and other pools would just provide a different `.worker` value. But right now I guess we can just expose this

packages/vitest/src/integrations/vi.ts
AriPerkkio: Just to make the check more clear. It's a bit magical to check for `process.send` in code blocks where we don't even use `process.send`. Some may not know that this is how to identify child processes. Makes code more readable.

AriPerkkio:
> Is it fast though?

Calling `recycleWorkers` will create new worker with given runtime as soon as it's possible. If there are idle workers with previous runtime, those ones will be destroyed and new ones will be spawned. Recycling this many threads/processes is fast. 

> What does this mean for a non-isolated environment? Does it mean we would need to wait to start another process/worker and then restart it again?

From Tinypool's API's perspective it's identical usage. When we want to change the `runtime`, `recycleWorkers` is needed and new process/worker is spawned.
But note that `isolateWorkers` option cannot change, just like `maxThreads` and `minThreads`. If they need changing, we need multiple instances of Tinypool.

> Maybe create half workers and half forks, and close/restart other runtimes (so, 2/2 -> 3/1) when they finish running? (So, when there is a workers queue, but forks are idling)

Currently it's working like this:
- Imagine we have 5 tasks where 3 are for `worker_threads` and 2 for `child_process`. First 3 tasks take 2s, rest 1s. We allocate 2 threads for Tinypool.
- First create `worker_threads` pool with 2 threads. Push 3 tasks there. 2 tasks start running.
- 2 tasks finish. 1 task from queue is started.
- Pool emits `pool.drain` event and indicates queue is empty.
- We call `recycleWorkers` and change runtime to `child_process`. Now Tinypool will destroy the one idle worker and spawns a new one.
- 1 task is started in `child_process`, 1 task is still running in `worker_threads`.
- Both tasks are finished. `worker_thread` worker is recycled and new `child_process` one is spawned. Both take tasks from queue.

Explaining this would be easier with whiteboard :smile:. Hopefully this explains the situation a bit.

sheremet-va:
> Calling recycleWorkers will create new worker with given runtime as soon as it's possible. If there are idle workers with previous runtime, those ones will be destroyed and new ones will be spawned. Recycling this many threads/processes is fast.

But if we call it 100 times at the same time, and isolation is disabled (so, we only need 4 workers for example) - won't it just create new ones and destroy previous instead of reusing them?

AriPerkkio:
> won't it just create new ones and destroy previous instead of reusing them

It will destroy workers and create new ones, yes. And that's expected. We are also using this mechanism when looping over environments:

https://github.com/vitest-dev/vitest/blob/7f5022994ec645838d9fa1867602b2f30d51673a/packages/vitest/src/node/pools/threads.ts#L198-L200

But we definitely should not call it 100 times. Instead, we should sort files by pools and their options like done currently. 

1. Push tasks to pool with initial options
2. Wait for queue to be empty
3. Call `recycleWorkers` to change `runtime` and other options.
4. Push tasks to queue and repeat 2-4 until no tasks left.

sheremet-va:
@AriPerkkio looks like vm now takes even more memory :(

AriPerkkio:
I haven't checked the latest changes yet but by comparing `--heap-prof` outputs between `main` and PR it might be visible what's consuming the memory now.

```ts
export default defineConfig({
  test: {
    pool: "vmThreads",
    poolOptions: {
      vmThreads: {
        execArgv: ["--heap-prof", "--heap-prof-dir=./profiling"],
      },
    },
  },
});
```

Then open the files in Chrome dev-tools

sheremet-va:
> I haven't checked the latest changes yet but by comparing `--heap-prof` outputs between `main` and PR it might be visible what's consuming the memory now.
> 
> 
> 
> ```ts
> 
> 
>   test: {
> 
>     pool: "vmThreads",
> 
>     poolOptions: {
> 
>       vmThreads: {
> 
>         execArgv: ["--heap-prof", "--heap-prof-dir=./profiling"],
> 
>       },
> 
>     },
> 
>   },
> 
> });
> 
> ```
> 
> 
> 
> Then open the files in Chrome dev-tools  

It looks like vite-node test calls mocked `process.exit`, can't exit and the memory piles up. I will check it with your example, thank you!

AriPerkkio:
> It looks like vite-node test calls mocked `process.exit`, can't exit

I've also run into this during development. Back then it caused some of those 100% CPU spikes and left zombie processes. Maybe we should add some handling for these cases:

```diff
process.exit(1)

+ // This is run if exit was overwritten
+ throw new Error('Unable to exit since process.exit is overwritten')
```

AriPerkkio:
> looks like vm now takes even more memory

What kind of setup does reproducing this require? I'm seeing some memory increment on `vm.ts`'s `resolveModule`.

sheremet-va:
> What kind of setup does reproducing this require? I'm seeing some memory increment on `vm.ts`'s `resolveModule`.

I don't know actually, but CI always fails here. I didn't start working on it locally yet.

sheremet-va:
@AriPerkkio this is reproducible 100% on my local machine:

```
cd test/vite-node
pnpm test --pool=forks --run
```

It's called because `SIGTERM` is triggered by someone (tinypool?)

AriPerkkio:
> It's called because `SIGTERM` is triggered by someone (tinypool?).

Tinypool uses first `SIGTERM`, which is the default signal when calling `process.kill()`: https://nodejs.org/api/child_process.html#child_processforkmodulepath-args-options. If process doesn't terminate in time (Tinypool default 1000ms), it calls `process.kill('SIGKILL')`. This is similar to Jest.

If overwriting `process.exit` becomes issue I think we could store original copy in Tinypool's entrypoints before any user-provided code is even loaded. And when we start to terminate the processes (or threads), we could restore it back to original one. Doing the same thing in Vitest might be a good thing too.

Is this issue still reproducible easily? I'm interested to see where Node hangs when proces/thread is trying to exit in these cases. Might be able to capture that with debugger. 🤔 

I'll take a good look at this PR later this week and run some manual tests myself.

sheremet-va:
> Is this issue still reproducible easily? I'm interested to see where Node hangs when proces/thread is trying to exit in these cases. Might be able to capture that with debugger. 🤔

Yes, if you remove the `process.exit` assignment in the forks worker.

AriPerkkio:
Need to add `await vnServer.server.close()` in the `test/vite-node/test/server.test.ts` tests. Then tests will pass without hacking around `process.exit`. I think following is happening now:

1. Vite server remains open once tests exit
2. Tinypool kills process with `process.kill()` -> SIGTERM is sent to test process
3. Vite server running inside that `child_process` has SIGTERM handler which captures this and calls `process.exit`: https://github.com/vitejs/vite/blob/f43945c1c35de34d64b2b1772b1629aba83cb476/packages/vite/src/node/server/index.ts#L625-L632
4. Test process's RPC catches `process.exit` and sends it to main process via `onWorkerExit`
5. Main thread exits process: https://github.com/vitest-dev/vitest/blob/c946b1c841c04f27f026c2b10b7b8f16ac12ceab/packages/vitest/src/node/pools/rpc.ts#L10
6. Other tests that were running become zombie processes, something like https://github.com/nodejs/node/issues/46569. But I think in this case it's `child_process` running in `child_process`

I wonder why we exit whole process here instead of `ctx.close()`. Maybe we should also call `ctx.cancelCurrentRun()` before that. 

https://github.com/vitest-dev/vitest/blob/c946b1c841c04f27f026c2b10b7b8f16ac12ceab/packages/vitest/src/node/pools/rpc.ts#L8-L11

packages/vitest/src/runtime/workers/base.ts
```diff
@@ -0,0 +1,47 @@
...
+    ctx.invalidates.forEach((fsPath) => {
+      moduleCache.delete(fsPath)
+      moduleCache.delete(`mock:${fsPath}`)
+    })
+  }
+  ctx.files.forEach(i => state.moduleCache.delete(i))
+
+  const [executor, { run }] = await Promise.all([
+    startViteNode({ state, requestStubs: getDefaultRequestStubs() }),
+    import('../runBaseTests'),
```
AriPerkkio, line 39: Why dynamic import? `runBaseTests` is not defined as separate Rollup entry, like `runVmTests` is. Should it?
I don't see any side effects on `runBaseTests`, so if it's not meant to be in a separate bundle I guess it could be imported on top level?

packages/vitest/src/runtime/workers/utils.ts
```diff
@@ -0,0 +1,42 @@
...
+  return {
+    post: (v) => { port.postMessage(v) },
+    on: (fn) => { port.addListener('message', fn) },
+  }
+}
+
+export function createForksRpcOptions(v8: typeof import('v8')): WorkerRpcOptions {
+  return {
+    serialize: v8.serialize,
+    deserialize: v => v8.deserialize(Buffer.from(v)),
```
AriPerkkio, line 17: This argument has to be renamed. Right now Rollup (or maybe rollup dts-plugin) produces invalid type declarations for this. Check `packages/vitest/dist/workers.d.ts`:

image

Renaming the variable to something else than its module name fixes this.

```suggestion
export function createForksRpcOptions(nodeV8: typeof import('v8')): WorkerRpcOptions {
  return {
    serialize: nodeV8.serialize,
    deserialize: v => nodeV8.deserialize(Buffer.from(v)),
```

packages/vitest/src/runtime/workers/base.ts
sheremet-va: It imports from Vitest which relies on __vitest_worker__ which is not defined until this function is called

packages/vitest/src/runtime/worker.ts
```diff
@@ -1,94 +1,74 @@
...
 import { setupInspect } from './inspector'
-import { createSafeRpc, rpcDone } from './rpc'
+import { createRuntimeRpc, rpcDone } from './rpc'
+import type { VitestWorker } from './workers/types'
 
-async function init(ctx: WorkerContext) {
-  // @ts-expect-error untyped global
-  const isInitialized = typeof __vitest_worker__ !== 'undefined'
-  const isIsolatedThreads = ctx.config.pool === 'threads' && (ctx.config.poolOptions?.threads?.isolate ?? true)
+// this is what every pool executes when running tests
```
AriPerkkio, line 11: `process.title` is no longer set to forked processes. Maybe include it here?

```ts
if (isChildProcess()) {
  try {
    process.title = `node (vitest ${poolId})`
  }
  catch {}
}
```

packages/vitest/src/runtime/workers/forks.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  async runTests(state: WorkerGlobalState) {
+    const exit = process.exit
+    state.ctx.config = unwrapForksConfig(state.ctx.config)
+
+    try {
+      await runBaseTests(state)
+    }
+    finally {
+      process.exit = exit
```
AriPerkkio, line 20: Another option for this is to set following in `pools/rpc.ts`:

```ts
async onWorkerExit(error, code) {
  await ctx.cancelCurrentRun('test-failure')
  await ctx.runningPromise

  await ctx.logger.printError(error, { type: 'Unexpected Exit', fullStack: true })

  process.exitCode = code || 1
  await ctx.exit(true)
},
```

Right now this can hide `process.exit()` calls that user's code executes after tests have finised, like that unclosed Vite server in vite-node's tests. 

It would also be nice to print out the test file that caused this exit.

sheremet-va: We should do this in a separate PR before releasing the next version. I am not sure of the impact of such a change yet.

sheremet-va: #4903

AriPerkkio (approved):

Issue #3709, opened 2023-06-30
Per-directory coverage thresholds


lcswillems:
### Clear and concise description of the problem

I cannot set coverage thresholds globally and also for a specific directory.

### Suggested solution

Something similar to coverageThreshold of Jest: https://jestjs.io/docs/configuration#coveragethreshold-object

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

AriPerkkio:
Now that `@vitest/coverage-c8` is being deprecated, we can implement this for both coverage providers. I think it would be breaking change though, as option structure would change.

adriengibrat:
> I think it would be breaking change though, as option structure would change.

does not have to introduce Breaking change, "coverageThreshold" could be an alternative property like it is implemented in jest.

AriPerkkio:
The current option format would be replaced with new one.

These options:

```js
{
  test: {
    coverage: {
      branches: 100,
      lines: 100,
      statements: 100,
      functions: 100,
    }
  }
}
```

would be replaced with these ones:

```js
{
  test: {
    coverage: {
      threshold: {
        global {
          branches: 100,
          lines: 100,
          statements: 100,
          functions: 100,
        },
        "./src/utils/**": {
          lines: 80,
        }
      }
    }
  }
}
```

But sure, it doesn't have to be a breaking change at first. The `coverage['lines'|'branches'|'statements'|'functions']` could be used as fallback for `coverage.threshold.global` before they are completely removed from configuration API.

PR #4442, opened 2023-11-05
feat!(coverage): glob based coverage thresholds

AriPerkkio:
### Description

Closes https://github.com/vitest-dev/vitest/issues/3709.

Refactors coverage thresholds API into new shape and adds support for specifying thresholds for specific files using glob patterns. Introducing breaking changes for the configuration options made sense here and final results looks clean and simple. 

```ts

export default defineConfig({
  test: {
    coverage: {
      thresholds: {
        100: false,
        perFile: true,
        autoUpdate: true,

        // Thresholds for all files
        lines: 100,
        functions: 100,
        branches: 80,
        statements: 90,

        // Thresholds for utilities
        "**/utils/**": {
          lines: 80,
          statements: 50,
        },
      },
    },
  },
});
```

### BREAKING CHANGES

- `coverage.perFile` is now `coverage.thresholds.perFile`
- `coverage.thresholdAutoUpdate` is now `coverage.thresholds.autoUpdate`
- `coverage.100` is now `coverage.thresholds.100`
- `coverage.lines` is now `coverage.thresholds.lines`
- `coverage.functions` is now `coverage.thresholds.functions`
- `coverage.branches` is now `coverage.thresholds.branches`
- `coverage.statements` is now `coverage.thresholds.statements`

### Tests

### Documentation

### Changesets


packages/vitest/src/types/coverage.ts
```diff
@@ -157,40 +157,62 @@ export interface BaseCoverageOptions {
...
+     * {
+     *  'src/utils/**.ts': {
+     *    statements: 80,
+     *    functions: 95,
+     *    branches: 100,
+     *    lines: 50,
+     * }
+     * ```
+     */
+    [glob: string]: unknown
```
AriPerkkio, line 214: These can also be typed: https://www.typescriptlang.org/play?#code/C4TwDgpgBAKgFgJwgZzgewDYBNkCECGy0AvFAN4BQUUyw+wEAthAHbDID8AXFCwK6MARhARUoAMz4sAxsACWaFpx78hIsYIT4ZcFN14Dho6hjks9Kw+uoBGAAx39gtJgjaxkBADE5GCE5c-dwBfClBIWEQUdGxkKFJ4JFRMHAIiKAAfKAAKMigAbQBzDDRBHloEM0KAXR5E6JS8QmhgqAAySKSY1OaASgoB6UVaKCGANxF8QogePOAo5Ni6he641tJKanmuxtmxagB6A6gAcRLBfAwaCGB5FkLkfZo6BmY2ZB57OwAaJ8kZeTDT4OX7UaiabTSXQfKBfUFg0zmGFwp5fHjABB8CDw6ieHx+dGY7EDMFQI6nc5QC5ELDXW5VR6kgBEACoWQc+PIMMgDmymXtSdRaPQmKx2MCfk9qP9ZAolBKcWCIToUAqpVBEarYSD1WioBisYrcSJ8TMJJciIrgr8nqz2WApiheSz+eR1ZqYQBWSWCqB43xm8QW4mk61iMOhIA

sheremet-va: What is stopping us from using these types?

AriPerkkio: I'm planning to rewrite the current implementation to use the typings shown in the linked TS playground. Nothing is stopping from using those. 

Earlier I thought using such typed dynamic keys in `interface` would not be possible in Typescript, but someone in Vitest Discrod provided solution that works really nice. Intellisense and type checking works even for those `[glob: string]` types there. I received these types just this morning and haven't had the time to implement it yet. Droppped the link here for later as notification.

AriPerkkio:
@arianshirepublik this feature is not yet finished but should make it to v1 release. I'm introducing some breaking changes to coverage option structure at the same time.

AriPerkkio:
Adding support to `thresholdAutoUpdate` now that there are multiple different coverage thresholds in configuration file is a bit tricky. Earlier we've used just search-and-replace with regexp but now we may need to parse the configuration file's AST.

sheremet-va:
> Adding support to `thresholdAutoUpdate` now that there are multiple different coverage thresholds in configuration file is a bit tricky. Earlier we've used just search-and-replace with regexp but now we may need to parse the configuration file's AST.

Maybe you can try https://github.com/unjs/magicast ?

packages/vitest/package.json
```diff
@@ -146,6 +146,7 @@
     "execa": "^8.0.1",
     "local-pkg": "^0.5.0",
     "magic-string": "^0.30.5",
+    "magicast": "^0.3.2",
```
sheremet-va, line 149: I don't like having magicast in Vitest dependencies. I think it should be defined only in coverage for better size management

AriPerkkio: Ok, we can do that. There's already similar approach for `istanbul-lib-coverage`. `vitest` package has it's types as `devDependencies` and actual package is in coverage providers' `dependencies`. It's then passed from providers to `vitest` side as function arguments when needed.

packages/vitest/src/utils/coverage.ts
```diff
@@ -1,120 +1,194 @@
...
+
+      resolvedThresholds.push({
+        name: glob,
+        coverageMap: globCoverageMap,
+        thresholds: globThresholds,
+      })
+    }
+
+    // TODO: Global threshold should include all files, even the ones defined for specific globs.
+    // Remove filter.
```
AriPerkkio, line 172: Note for future references:

This is actually working correctly now. Files included in their own glob thresholds should not be counted into global coverage when checking thresholds. When using `text` reporter it might cause some confusion as coverage percentages do not match thresholds configuration. 

Below is example from Jest where `text` reporter reports 33% coverage for branches, but custom threshold configuration moves one file into its own set and lowers global branch coverage:

image

arianshirepublik:
> @arianshirepublik this feature is not yet finished but should make it to v1 release. I'm introducing some breaking changes to coverage option structure at the same time.

Hi @AriPerkkio 
Can we use this feature in v1.0.0-beta.4?

AriPerkkio:
No, this feature is not yet finished. Please wait patiently 😄 
I'm planning to finalize this during this week.

packages/vitest/src/utils/coverage.ts
```diff
@@ -1,120 +1,195 @@
 import { readFileSync, writeFileSync } from 'node:fs'
 import { relative } from 'pathe'
+import mm from 'micromatch'
+import type { parseModule as MagicastParseModule } from 'magicast'
```
sheremet-va, line 4: Is this correctly bundled? I don't want to have a big jump in size here since there is no tree shaking for dts files

AriPerkkio: Good catch. The `dist/coverage.d.ts` seems to increase from 4000 characters to 67000 characters. 😄

AriPerkkio: Fixed by moving all `magicast` logic to providers and removed `magicast` from `vitest` package's `devDependencies`. There's now a generic `{ read, write }` interface for this.

sheremet-va (approved):

arianshirepublik:
Hi @AriPerkkio 
I tested [v1.0.0-beta.5](https://github.com/vitest-dev/vitest/releases/tag/v1.0.0-beta.5)
The report will not be generated. Can you check it?

image

arianshirepublik:
Hi @AriPerkkio 

When I run the command: yarn run test --coverage, 

The desired test case has been executed but no test report or coverage table has been generated, and then the program stops here.

image

AriPerkkio:
@arianshirepublik please open new issue with reproducible setup.

PR #4462, opened 2023-11-08
chore: add defaults to exports

goosewobbler:
### Description

Most of the packages use conditional exports without a `default` or `require` value so are not able to be used in CJS projects.

### Tests

### Documentation
None required

### Changesets


sheremet-va:
This is intentional. They cannot be used in CJS project even with this because we don't bundle CJS.

goosewobbler:
Well, we're trying to use `@vitest/spy` and our CJS E2Es aren't working, rollup can't resolve that dependency.  Simply adding a default export fixes them.  Why are there default exports in some of the packages but not others?

sheremet-va:
> rollup can't resolve that dependency

It cannot resolve it because it follows Node.js resolution, and fails correctly. Vitest and its dependencies cannot be imported using `require` which rollup assumes will be the output.

sheremet-va:
On the other hand, it gives different errors:

Screenshot 2023-11-10 at 13 43 53
Screenshot 2023-11-10 at 13 43 59

The last one (that you currently get) is kind of cryptic, so it might be better to add this `default` condition.

sheremet-va (changes requested):
I think we can merge this PR if `"import"` is just replaced with `"default"` (not added as an additional one) - even though `import` is the correct export, `default` gives a more meaningful error message

sheremet-va (approved):

PR #4492, opened 2023-11-13
fix: disable ESBuild when user config disables it

### LINKED ISSUE #4479: Vitest uses ESBuild even when esbuild is set to `false`
### Describe the bug

I'm attempting to set up Typia, which relies on `ts-patch` to allow transform plugins. I followed [these instructions](https://typia.io/docs/setup/#vite), which failed in my tests. The reason it failed is because TypeScript files were getting transformed _before_ they got to the TypeScript Rollup plugin, meaning `esbuild` was still active, even when set to `false`.  At first, I thought this was a Vite issue, but when I went to reproduce, it worked in Vite.

So then I went to repro for Vitest, and I discovered that Vitest ignores the `esbuild` setting from the Vite config, and always transforms.

In the reproduction, the TypeScript annotations would/should fail with `esbuild: false`, as TypeScript annotations wouldn't be runnable in a Node environment. Instead, tests continue to pass.

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-de2cnr?file=vite.config.ts,tsconfig.json&initialPath=__vitest__/

### System Info

```shell
System:
    OS: macOS 13.4.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 45.73 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.13.0 - ~/Library/Caches/fnm_multishells/47748_1699638513910/bin/node
    Yarn: 1.22.19 - ~/Library/Caches/fnm_multishells/47748_1699638513910/bin/yarn
    npm: 8.19.3 - ~/Library/Caches/fnm_multishells/47748_1699638513910/bin/npm
  Browsers:
    Chrome: 115.0.5790.170
    Safari: 16.5.2
```

### Used Package Manager

pnpm

### Validations

Namchee:
### Description

This pull request disables ESBuild when it has been disabled in the user config. When it is enabled or unspecified by the config, it will 'merge' the esbuild config instead.

Closes #4479 

### Tests

### Documentation

### Changesets


packages/vitest/src/node/plugins/index.ts
```diff
@@ -64,12 +64,15 @@ export async function VitestPlugin(options: UserConfig = {}, ctx = new Vitest('t
...
-          esbuild: {
-            sourcemap: 'external',
-
-            // Enables using ignore hint for coverage providers with @preserve keyword
-            legalComments: 'inline',
-          },
+          esbuild: viteConfig.esbuild === false
+            ? false
+            : {
+                ...viteConfig.esbuild,
```
sheremet-va, line 70: You don't need to merge it, the result of this function is merged with the custom config

sheremet-va (changes requested):

sheremet-va (approved):

PR #4538, opened 2023-11-19
fix: testFileDir always matches

### LINKED ISSUE #4425: `testFileDir` always matches
### Describe the bug

Given the tree:
```
.
├── tests1
│   └── test.spec.ts
├── tests11
│   └── test.spec.ts
└── tests2
    └── test.spec.ts
```

Vitest will match `tests1` & `tests11` when supplied with filter `tests1/`. 

Whilst i recognise that you can use the `include` option or even specifying direct filenames, that is a very heavyhanded and hardcoded approach which doesn't fit large test repositories (in my case i have 400+ test files in nested directories). 

Instead it would be less-surprising for it to match based on inputs such as `tests1/` to only use that directory. A regex `\b` would also be acceptable. 

### Reproduction

Using this [StackBlitz](https://stackblitz.com/edit/vitest-dev-vitest-grbmeo?file=package.json) you can execute the following scripts that illustrates a plausible scenario:
- `pnpm test:all` : ✅
- `pnpm test:1` : ❌ - Runs tests for both `folder1` and `folder11`
- `pnpm test:2` : ✅
- `pnpm test:11` : ✅

> [!IMPORTANT]  
> In reality, most scenarios wouldn't be matching based on numbers, but i've got into weird scenarios where expressions collide with each other: `test_blocks` and `test_blockscope`

### System Info

```shell
all
```

### Used Package Manager

pnpm

### Validations

ibuibu:
### Description
fix #4425 

### Tests

### Documentation

### Changesets


packages/vitest/src/node/workspace.ts
```diff
@@ -254,7 +254,8 @@ export class WorkspaceProject {
       return testFiles.filter((t) => {
         const testFile = relative(dir, t)
         return filters.some((f) => {
-          return testFile.includes(f) || testFile.includes(relative(dir, f))
+          const relativePath = f.endsWith('/') ? `${relative(dir, f)}/` : relative(dir, f)
```
timbrinded, line 257: seems safer to use the `node:path` functions then hard coding it like this.

sheremet-va (approved):

PR #4588, opened 2023-11-25
browser mode - puppeteer basic support

oriSomething:
### Description

Support for `puppeteer` provider to **browser mode**.

**Ideally this PR shouldn't be needed and I still investigating if adding a new provider, is a good idea**. 

The issue is that `@playwight/test` doesn't play well with `playwright` package installed, so it forces me to use `webdriverio`. The issues is that in workspaces with lots of packages that use **browser mode** with `webdriverio`, I got to very unstable CI. All e2e tests with Playwright are stable, so I thought using `puppeteer` which is essentially the same as `playwright` without the limitation of using it with `@playwight/test` in the same module would improve the stability. **I still need more testing in the CI with it to make sure it adds stability (I hope to update result in recent days)**.

**Anyway, I'm adding this PR as a reference until I finish to see if it worth it**

### Tests

**I also tested it manually with Chrome and Firefox**

### Documentation

### Changesets

### DIFF
--- a/packages/browser/package.json
+++ b/packages/browser/package.json
@@ -30,6 +30,9 @@
     "./providers/playwright": {
       "types": "./dist/providers/playwright.d.ts"
     },
+    "./providers/puppeteer": {
+      "types": "./dist/providers/puppeteer.d.ts"
+    },
     "./*": "./*"
   },
   "main": "./dist/index.js",
@@ -50,6 +53,7 @@
   },
   "peerDependencies": {
     "playwright": "*",
+    "puppeteer": "*",
     "safaridriver": "*",
     "vitest": "^1.0.0-0",
     "webdriverio": "*"
@@ -62,6 +66,9 @@
       "optional": true
     },
     "playwright": {
+      "optional": true
+    },
+    "puppeteer": {
       "optional": true
     }
   },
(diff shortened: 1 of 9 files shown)


nicojs:
I'm just passing by here, but if `@playwright/test` doesn't play nice with a custom `@playwright` installed, shouldn't that be fixed in `@playwright/test`? 🤷

oriSomething:
@nicojs Yes this is why I wrote

> Ideally this PR shouldn't be needed and I still investigating if adding a new provider, is a good idea.

For quite a time, Playwright itself blocked the ability to use Playwright stand alone if `@playwright/test` installed. Since it wasn't clear it was ever changed I opened it, as there are browser bugs open for years and you need a work around. 

Anyway, since I haven't find the issue, it seems the issue form Playwight resolved, so I'm closing this PR

nicojs:
> Yes this is why I wrote

Sorry about that; I read this PR late at night and overlooked your comment.

Glad to know the issue got resolved on Playwright's end. 

Don't let this discourage you, a feature to support puppeteer might be valuable all on its own 👐

oriSomething:
@nicojs I'm **not** discouraged at all! I truly prefer projects which are less bloated, And the main idea is to resolve issues I had, not get code merged to Vitest repo.

If there is something might be better to change is to make a provider API, to allow custom providers usage and not hard coded one, but I personally don't need one currently

PR #4615, opened 2023-11-28
feat(expect): compare URL objects by href

### LINKED ISSUE #4614: expect: compare URLs by href (so that different URLs aren't considered equal)

When comparing URLs in Vitest, tests easily pass for completely different URLs (in a sense, this can be considered a bug):

```js
expect(new URL('https://example.org')).toEqual(new URL('https://different-example.org'))
// This passes, but it shouldn't:                               ^^^^^^^^^^
```

The main equals comparison utility can be easily updated to handle this case better. I'd be happy to contribute this.

### Suggested solution

Add a special case for `URL` objects in https://github.com/vitest-dev/vitest/blob/main/packages/expect/src/jest-utils.ts#L96-L97:

```diff
    if (a instanceof Error && b instanceof Error)
      return a.message === b.message

+   if (a instanceof URL && b instanceof URL)
+     return a.href === b.href
```

And add test cases for this in https://github.com/vitest-dev/vitest/blob/main/test/core/test/jest-expect.test.ts#L51-L52:

```diff
      expect(new Date(0)).toEqual(new Date(0))
      expect(new Date('inValId')).toEqual(new Date('inValId'))

+     expect(new URL('https://example.org')).toEqual(new URL('https://example.org'))
+     expect(new URL('https://example.org')).not.toEqual(new URL('https://different-example.org'))
```

### Alternative

Fail URL comparisons unless they're the exact same object.

### Additional context

I initially noticed this behavior while working on a small wrapper around fetch where I have a lot of tests that very roughly look like this (though in place of `fetch` I'm calling my fetch wrapper, replaced here for brevity):

```ts
test('test', async () => {
	vi.spyOn(window, 'fetch').mockImplementation(() => Promise.resolve(new Response('OK')))

	await fetch(new URL('https://example.org'))

	expect(fetch).toHaveBeenCalledWith(new URL('https://different-example.org'))
})
```

I noticed that I couldn't get the assertion to fail at all. As a workaround, my tests actually look like this:

```ts
test('test', async () => {
	vi.spyOn(window, 'fetch').mockImplementation((input) => {
		expect(toUrl(input).href).toEqual(toUrl('https://different-example.org').href)

		return Promise.resolve(new Response('OK'))
	})

	await fetch(new URL('https://example.org'))

	expect(fetch).toHaveBeenCalled()
})

function toUrl(url: string | URL | Request): URL {
	return typeof url === 'string'
		? new URL(url)
		: url instanceof URL
			? url
			: new URL(url.url)
}
```

While this works great, it's really a little too cumbersome for my liking. Using `toHaveBeenCalledWith` would be much nicer for this use case.

### Validations

kleinfreund:
### Description

Change the "equal comparison" logic (as used by `toEqual`, `toStrictEqual`, etc.) to compare `URL` objects by their `href` property.

Closes #4614.

### Tests

### Documentation

### Changesets

### DIFF
--- a/test/core/test/jest-expect.test.ts
+++ b/test/core/test/jest-expect.test.ts
@@ -58,6 +58,10 @@ describe('jest-expect', () => {
     expect(new URL('https://example.org')).not.toEqual(new URL('https://different-example.org'))
     expect(new URL('https://example.org?query=value')).toEqual(new URL('https://example.org?query=value'))
     expect(new URL('https://example.org?query=one')).not.toEqual(new URL('https://example.org?query=two'))
+    expect(new URL('https://subdomain.example.org/path?query=value#fragment-identifier')).toEqual(new URL('https://subdomain.example.org/path?query=value#fragment-identifier'))
+    expect(new URL('https://subdomain.example.org/path?query=value#fragment-identifier')).not.toEqual(new URL('https://subdomain.example.org/path?query=value#different-fragment-identifier'))
+    expect(new URL('https://example.org/path')).toEqual(new URL('/path', 'https://example.org'))
+    expect(new URL('https://example.org/path')).not.toEqual(new URL('/path', 'https://example.com'))
 
     expect(BigInt(1)).toBeGreaterThan(BigInt(0))
     expect(1).toBeGreaterThan(BigInt(0))
--- a/test/core/test/jest-expect.test.ts
+++ b/test/core/test/jest-expect.test.ts
@@ -51,6 +51,12 @@ describe('jest-expect', () => {
     expect(new Date(0)).toEqual(new Date(0))
     expect(new Date('inValId')).toEqual(new Date('inValId'))
 
+    expect(new Error('message')).toEqual(new Error('message'))
+    expect(new Error('message')).not.toEqual(new Error('different message'))
+
+    expect(new URL('https://example.org')).toEqual(new URL('https://example.org'))
+    expect(new URL('https://example.org')).not.toEqual(new URL('https://different-example.org'))
+
     expect(BigInt(1)).toBeGreaterThan(BigInt(0))
     expect(1).toBeGreaterThan(BigInt(0))
     expect(BigInt(1)).toBeGreaterThan(0)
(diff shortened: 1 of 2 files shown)


test/core/test/jest-expect.test.ts
```diff
@@ -51,6 +51,12 @@ describe('jest-expect', () => {
     expect(new Date(0)).toEqual(new Date(0))
     expect(new Date('inValId')).toEqual(new Date('inValId'))
 
+    expect(new Error('message')).toEqual(new Error('message'))
+    expect(new Error('message')).not.toEqual(new Error('different message'))
```
kleinfreund, line 55: **Comment**: When adding tests, I also added the same kind of cases for `Error` objects. Let me know if you'd rather not have them in this PR.

kleinfreund:
One of the CI test runs failed, but it doesn't look like it's caused by changes in this PR: https://github.com/vitest-dev/vitest/actions/runs/7023800471/job/19111142242?pr=4615

test/core/test/jest-expect.test.ts
```diff
@@ -51,6 +51,12 @@ describe('jest-expect', () => {
...
+
+    expect(new URL('https://example.org')).toEqual(new URL('https://example.org'))
+    expect(new URL('https://example.org')).not.toEqual(new URL('https://different-example.org'))
+
```
AriPerkkio, line 59: Maybe also add a test case that includes search query parameters, just so that replacing `href` with other URL properties would fail these tests.

```suggestion
    expect(new URL('https://example.org?query=value')).toEqual(new URL('https://example.org?query=value'))
    expect(new URL('https://example.org?query=one')).not.toEqual(new URL('https://example.org?query=two'))

```

kleinfreund: Good idea. I’ve applied your suggestion.

sheremet-va: While we are at it, maybe we should also check hash and query+hash - different variations.

kleinfreund: I added two more sets of test cases:

- One set for URLs that contain a subdomain, path segment, query parameter, and fragment identifier
- One set for URLs using the `URL` constructor’s second argument for setting the base

sheremet-va (approved):

Issue #4619, opened 2023-11-29
vi.mock makes vi.spyOn  to fail


hadson172:
### Describe the bug

My code looks like this:
```
//vi.mock(("anything???!!!!"))
describe('tests', () => {

    it("this works but fails when i uncomment vi.mock, why", () => {
        vi.spyOn(Module, 'moduleFunction')
        expect(moduleFunction).not.toHaveBeenCalled()
    })

    it("this works as well", () => {
        const spy = vi.spyOn(Module, 'moduleFunction')
        expect(spy).not.toHaveBeenCalled()
    })
});
```
In short, in jest everything it is working, in vitest first case it is failing when i add any call to vi.mock.
It would be nice to have any documentation about this issue or explanation in migration guide.

### Reproduction
Edit: https://stackblitz.com/edit/vitest-dev-vitest-t5pdkp?file=test%2Fsuite.test.ts,src%2Fbasic.ts&initialPath=__vitest__/

```
//vi.mock(("anything???!!!!"))
describe('tests', () => {

    it("this works but fails when i uncomment vi.mock, why", () => {
        vi.spyOn(Module, 'moduleFunction')
        expect(moduleFunction).not.toHaveBeenCalled()
    })

    it("this works as well", () => {
        const spy = vi.spyOn(Module, 'moduleFunction')
        expect(spy).not.toHaveBeenCalled()
    })
});
```

### System Info

```shell
System:
    OS: macOS 14.1.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 293.77 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.18.2 - /usr/local/bin/node
    npm: 9.8.1 - /usr/local/bin/npm
    pnpm: 8.10.5 - /opt/homebrew/bin/pnpm
  Browsers:
    Chrome: 119.0.6045.199
    Safari: 17.1
  npmPackages:
    @vitejs/plugin-react: ^4.2.0 => 4.2.0 
    vite: ^5.0.0 => 5.0.2 
    vitest: ^0.34.6 => 0.34.6
```

### Used Package Manager

pnpm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
[...]

PR #4650, opened 2023-12-04
refactor(vitest): call hoistMocks in directRequest instead of mocks plugin

Dunqing:
### Description

close: #4619

### Tests

### Documentation

### Changesets

### DIFF
--- a/test/core/test/injector-mock.test.ts
+++ b/test/core/test/injector-mock.test.ts
@@ -15,9 +15,9 @@ test('hoists mock, unmock, hoisted', async () => {
     vi.unmock('path')
     vi.hoisted(() => {})
 
-
-
-
+      
+      
+      
       "
   `)
 })
@@ -35,9 +35,9 @@ test('always hoists import from vitest', async () => {
     vi.hoisted(() => {})
 
       import { vi } from 'vitest'
-
-
-
+      
+      
+      
       import { test } from 'vitest'
       "
   `)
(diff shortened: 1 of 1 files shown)


sheremet-va:
> To fix the bug, as I already said, we need to process property access the same way Vite does for SSR. That way we can keep it working both in browser and Node.js.

We actually already have code for this in the browser package. It does a similar thing if `hijackESM` is enabled.

Dunqing:
> To fix the bug, as I already said, we need to process property access the same way Vite does for SSR. That way we can keep it working both in browser and Node.js.

I see. But it can be complicated to deal with

sheremet-va:
> But it can be complicated to deal with

I mean, we already have the code. How hard can it be 😄

Dunqing:
> I mean, we already have the code. How hard can it be 😄

Ah, I just saw it. I have another question, can we transform only the import source that is mocked?

Dunqing:
> > I have another question, can we transform only the import source that is mocked?
> 
> Not sure what you mean. We only transform the file that calls `vi.mock`

What I mean is that imports without mock should be filtered here.

https://github.com/vitest-dev/vitest/blob/main/packages/vitest/src/node/hoistMocks.ts#L103-L115

It looks like this.
```ts
import a from './a'
import b from './b' // should be filtered out because here doesn't call vi.mock('./b');
vi.mock('./a')

```

Dunqing:
I will refer to `hijackESM` to solve this problem

PR #4651, opened 2023-12-04
feat(snapshot): add option to configure snapshot directory

d3lm:
### Description

This PR adds options to the `NodeSnapshotEnvironment` which allows to configure the snapshot directory. For a project of mine I extended `NodeSnapshotEnvironment` and had to overwrite `resolvePath` because we also use `jest` which already writes snapshots into `__snapshots__`. Therefore I needed a different directory. I figured it might be a nice addition to be able to configure this without extending `NodeSnapshotEnvironment`.

### Tests

### Documentation

### Changesets


sheremet-va:
I don't really see how this is useful. If you are using this class, you can probably just override this method. If you are using vitest, you don't even have access to it. Is it just for public usage of `@vitest/snapshot` package? Do you have an example project where it's used?

d3lm:
Yes exactly, it's when you use `@vitest/snapshot` directly which we are cause it's very lightweight compared to `jest-snapshot`. We use it for WebContainer which is closed source but it basically does the same as described in the README except that I need to overwrite the snapshot directory to be different from what jest uses.

sheremet-va (approved):

Issue #4604, opened 2023-11-27
vi.mock breaks live bindings


nathanmmiller:
### Describe the bug

If vi.mock is used in a test module, live bindings break. I've included an example reproduction below, using "latest" version of vitest.

It doesn't matter what format you do the vi.mock in or if it mocks another module or an external package.

I've tried (using file-saver as an example):
- vi.mock('file-saver');
- vi.mock('file-saver', () => ({}));
- vi.mock('file-saver', async (importOriginal) => {
  const mod = await importOriginal();
    return mod;
    });
- ...etc. All to the same effect

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-bq5brv?file=test%2Fbasic.test.ts

Simply comment out line 5 in basic.test.ts to see the tests pass, then uncomment it and one test (the live bindings one) will fail because the live binding no longer works.

### System Info

```shell
Stackblitz, but this is reproducible elsewhere as well.
```

### Used Package Manager

yarn

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

sheremet-va:
Yeah, it's a known issue. There is an open issue already, but I can't find for now.

nathanmmiller:
@sheremet-va yikes! That impacts me pretty significantly (I provided a tiny repro, but essentially I rely on live binding all over my tests for a number of reasons). Is there any kind of expectation of when this could be fixed, or should I stick with jest for the time being?

sheremet-va:
Workaround is described here: https://github.com/vitest-dev/vitest/issues/3300

Issue #3300, opened 2023-05-04
variable of module not shared if vi.mock is used in v0.31.0


freakzlike:
### Describe the bug

After update to `v0.31.0` some tests are failing which rely on exporting a variable which is set within `beforeEach` when `vi.mock` is used:

`basic.test.ts`
```typescript
import { it, expect, vi } from 'vitest';
import { foo } from './set-foo';

// Comment this to make the test work
vi.mock('./somefile');

it('should be 1', () => {
  expect(foo).toBe(1);
});
```

`set-foo.ts`
```typescript
import { beforeEach, afterEach } from 'vitest';

export let foo: number;

beforeEach(() => {
  foo = 1;
});

afterEach(() => {
  foo = 2;
});
```

Running the test will result in `AssertionError: expected undefined to be 1 // Object.is equality`

I have verified that `beforeEach` and `afterEach` are called correctly. Commenting the `vi.mock` will make the test work. Not sure if I've missed something in the change log.

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-hcvj1e?file=test/basic.test.ts

`v0.30.1` with `vi.mock` = working
`v0.31.0` without `vi.mock` = working
`v0.31.0` with `vi.mock` = not working

### System Info

```shell
System:
    OS: Linux 5.15 Ubuntu 20.04.6 LTS (Focal Fossa)
    CPU: (5) x64 AMD Ryzen 5 4500U with Radeon Graphics
    Memory: 4.67 GB / 6.78 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 18.15.0 - ~/.nvm/versions/node/v18.15.0/bin/node
    npm: 9.5.0 - ~/.nvm/versions/node/v18.15.0/bin/npm
  npmPackages:
    @vitejs/plugin-vue: 4.2.1 => 4.2.1
    @vitest/coverage-istanbul: 0.31.0 => 0.31.0
    vite: 4.3.4 => 4.3.4
    vitest: 0.31.0 => 0.31.0
```

### Used Package Manager

pnpm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
[...]

PR #4664, opened 2023-12-05
refactor(vitest): reimplement hoistMocks to fix incorrect dynamic imports

Dunqing:
### Description

fix: #4619 
fix: #4604
fix: #3300

### Tests

### Documentation

### Changesets


packages/browser/rollup.config.js
```diff
@@ -14,6 +14,7 @@ const external = [
   /^@?vitest(\/|$)/,
   'worker_threads',
   'node:worker_threads',
+  '@vitest/utils/ast',
```
sheremet-va, line 17: It should already be covered by a regexp on L14

packages/vitest/src/node/hoistMocks.ts
```diff
@@ -71,58 +72,114 @@ export function hoistMocks(code: string, id: string, parse: PluginContext['parse
...
+  const declaredConst = new Set<string>()
+
+  esmWalker(ast, {
+    onIdentifier(id, parent, parentStack) {
+      const grandparent = parentStack[1]
+      const binding = idToImportMap.get(id.name)
+      if (!binding)
+        return
+
+      if (isStaticProperty(parent) && parent.shorthand) {
```
sheremet-va, line 150: I feel like these if statements are duplicated here and in browser. Maybe insteadof `parentStack` we can pass down some kind of options object? `hasBindingShortcut: true, classDeclaration: false, classExpression: false` (don't forget to add comments with example why we are doing g the transform like in property shorthand)

packages/vitest/src/node/hoistMocks.ts
```diff
@@ -71,58 +72,114 @@ export function hoistMocks(code: string, id: string, parse: PluginContext['parse
...
+        !(parent.type === 'ClassExpression' && id === parent.id)
+      ) {
+        s.update(id.start, id.end, binding)
+      }
+    },
+    onDynamicImport() {},
+    onImportMeta() {},
+  })
+
   simpleWalk(ast, {
```
sheremet-va, line 183: Can we include that in a previous walk for performance reasons?

test/core/test/spy.test.ts
```diff
@@ -1,5 +1,7 @@
 import { describe, expect, test, vi } from 'vitest'
 import * as mock from './fixtures/hello-mock'
+import * as squaredModule from './fixtures/squared'
```
sheremet-va, line 3: We need more tests for this behaviour. I also added two more issues in your description - you can look it up there. The tests should be in examples/mocks

test/core/test/injector-mock.test.ts
```diff
@@ -67,7 +67,6 @@ test('correctly mocks namespaced', () => {
   `)).toMatchInlineSnapshot(`
     "const { vi } = await import('vitest')
     vi.mock('../src/add', () => {})
-    const AddModule = await import('../src/add')
-    const { default: add } = AddModule"
```
sheremet-va, line 71: There are no tests that check if properties are accessed correctly. Maybe we can take them from Vite? This function originated there

sheremet-va (changes requested):

packages/vitest/src/node/hoistMocks.ts
Dunqing: Sounds good

packages/vitest/src/node/hoistMocks.ts
Dunqing: Yes, Let me make some changes in `esmWalker`

test/core/test/injector-mock.test.ts
Dunqing: It doesn't look like we need to do this. The current behavior also works fine

sheremet-va: No, we _need_ to have tests.

Dunqing: I checked in vite. Looks like we don't need this line `const { default: add } = AddModule"
`
```ts
  expect(await ssrTransformSimpleCode(`
    import { vi } from 'vitest'
    import add, * as AddModule from '../src/add'
    add();
    vi.mock('../src/add', () => {})`))
.toMatchInlineSnapshot(`
    "const __vite_ssr_import_0__ = await __vite_ssr_import__("vitest", {"importedNames":["vi"]});
    const __vite_ssr_import_1__ = await __vite_ssr_import__("../src/add", {"importedNames":["default"]});

      
      
      __vite_ssr_import_1__.default();
      __vite_ssr_import_0__.vi.mock('../src/add', () => {})"
  `)
```

sheremet-va: Then add these tests in this pr

sheremet-va:
I cannot accept this PR without tests. Vitest is a testing framework and the lack of tests for our own features is not a good sign. Initially, this was an experimental feature to test how it works, but now it does more than before. Since it doesn't use Vite's SSR transformer directly, we need to have tests in our repo. You can just copy SSR transformer tests from Vite.

Dunqing:
> I cannot accept this PR without tests. Vitest is a testing framework and the lack of tests for our own features is not a good sign. Initially, this was an experimental feature to test how it works, but now it does more than before. Since it doesn't use Vite's SSR transformer directly, we need to have tests in our repo. You can just copy SSR transformer tests from Vite.

Sorry! I finally got what you meant. I will add these tests later

Dunqing:
I added tests and made some changes. Since the `hoistMocks` must the code has mocks code. so I added some mock code combined with the original code. Please feel free to point out if it's not right!

test/core/test/injector-mock.test.ts
```diff
@@ -67,7 +72,1130 @@ test('correctly mocks namespaced', () => {
...
+
+      
+      
+      __vi_import_0__.default();"
   `)
 })
+
+describe('transform', () => {
+  const hoistSimpleCodeWithoutMocks = (code: string) => {
+    return hoistMocks(`import {vi} from "vitest";\nvi.mock(faker);\n${code}`, '/test.js', parse)?.code.trim()
```
sheremet-va, line 98: using `faker` without any reference to it feels bad. Maybe you can just use a string identifier?

test/core/test/injector-mock.test.ts
```diff
@@ -67,7 +72,1130 @@ test('correctly mocks namespaced', () => {
...
+    expect(
+      await hoistSimpleCodeWithoutMocks(
+      `import { ref } from 'vue';function foo() { return ref(0) }`,
+      ),
+    ).toMatchInlineSnapshot(`
+      "const { vi } = await import('vitest')
+      vi.mock(faker)
+      const __vi_import_0__ = await import('vue')
+
+      ;
```
sheremet-va, line 123: I know that it was before your refactor, but this hanging `;` feels lonely, maybe we can remove it somehow?

test/core/test/injector-mock.test.ts
Dunqing: Sorry. I don't understand the second one. Do you mean I should put `vi.mock` after the original code?

Dunqing: I seem to understand what you mean, I've changed to string literal. this test code makes more sense

test/core/test/injector-mock.test.ts
```diff
@@ -67,7 +72,1149 @@ test('correctly mocks namespaced', () => {
...
+
+  test('test', async () => {
+    expect(
+      await hoistSimpleCodeWithoutMocks(`
+      import { vi } from 'vitest'
+      import add, * as AddModule from '../src/add'
+      vi.mock('../src/add', () => {})`),
+    ).toMatchInlineSnapshot(`
+      "const { vi } = await import('vitest')
+      const { vi } = await import('vitest')
```
sheremet-va, line 1208: This test is a bit weird. It was here before the refactor and it tested that the default export was correctly extracted

Right now, since `add` and `AddModule` are not used anywhere it looks redundant. Also the name "test" for a test is not telling much

test/core/test/injector-mock.test.ts
sheremet-va: Ok

test/core/test/injector-mock.test.ts
Dunqing: I'm sorry, I added this test case to test certain scenarios. I forgot to remove it.

test/core/test/injector-mock.test.ts
Dunqing: This is an easy fix, I've already committed the code 😅

sheremet-va (approved):

sheremet-va:
@Dunqing this broke tests using imported variables inside the factory: #4872
PR with a failing test at #4873

Issue #4264, opened 2023-10-06
Default import doesn't work with vitest browser mode


quolpr:
### Describe the bug

Here is the test:

```
import React from "react";
import { expect, it } from "vitest";

it("works", () => {
  expect(React).not.toBe(undefined);
});
```

It will fail in browser mode, but will run in CLI mode. I suspect the error happen cause react still using CJS packages and default export is not configured properly for vitest.

### Reproduction

https://github.com/quolpr/vitest-browser-default-import-bug

### System Info

```shell
System:
    OS: macOS 14.0
    CPU: (10) arm64 Apple M1 Max
    Memory: 1.77 GB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 16.16.0 - ~/.asdf/installs/nodejs/16.16.0/bin/node
    Yarn: 1.22.19 - ~/.asdf/installs/nodejs/16.16.0/.npm/bin/yarn
    npm: 8.11.0 - ~/.asdf/plugins/nodejs/shims/npm
    Watchman: 2023.09.04.00 - /opt/homebrew/bin/watchman
  npmPackages:
    @vitejs/plugin-react-swc: ^3.4.0 => 3.4.0
    @vitest/browser: ^0.34.6 => 0.34.6
    vite: ^4.4.5 => 4.4.11
    vitest: ^0.34.6 => 0.34.6
```
```


### Used Package Manager

yarn

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

**Comment by quolpr** [comment]
I can even confim that react examples in `example/react` stops working after I change config to this:

```
export default defineConfig({
  test: {
    globals: true,
    browser: {
      name: "chromium",
[...]

PR #4701, opened 2023-12-08
fix: add proxyHijackESM for better spyOn in browser

samthor:
### Description

Changes the way `vi.spyOn` works in the browser via a new experimental flag `proxyHijackESM` that's intended to replace `slowHijackESM`.
(I'm not making promises about the speed of this approach, just that it works via a `Proxy`)

Refixes #4264 (which wasn't "fixed", it was just disabled) - importing `React` now works, the React sample included now uses this 👍 

The approach is vaguely:
 - Always convert static imports to dynamic ones (using `HoistMocks`)
 - dedups these dynamic imports inside `__vitest_mocker__.wrap`, so matching imports can be modified in a central location
 - instead of modifying modules that might be spied on (`slowHijackESM`), wraps these imports in a `Proxy`, which lets `tinyspy` work on it
   - the functions/callables of a module are further wrapped in a `Proxy`, so code like `import { callable } from 'foo'` continues to work — `callable` is itself now a "bound" callable `Proxy`, even though it looks like an orphan var

For more context, this is the start of an approach to #3046. This makes every module's import be wrapped so it can be piggybacked by `vi.mock` in a follow-up PR. (A complete version [is here](https://github.com/vitest-dev/vitest/compare/main...samthor:vitest:sam-test-update?expand=1) but it's not good enough for review)

**Alternatives**: Using iframes-per-test might make some things simpler, but I think this PR still helps — because `spyOn` is something we allow on any module - the `HoistMocks`/`Proxy` combination needs to run on anything a user might spy.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/browser/src/client/fakeModule.ts
+++ b/packages/browser/src/client/fakeModule.ts
@@ -3,85 +3,56 @@ function neverFunctionForProxy() {
   throw new Error(`should not get here`)
 }
 
-const originalSymbol = Symbol('orig')
-
 // allows replacing 'function' types on this module: passes through to the 'current value'
-function prepareFakeModuleForSpy(module: any) {
-  for (const key in module) {
-    let value: Function = module[key]
-    if (typeof value !== 'function')
-      continue
-
-    try {
-      delete module[key]
-    }
-    catch {
+function prepareFakeModuleForSpy(fakeModule: any, contents: any) {
+  for (const key in contents) {
+    if (typeof contents[key] !== 'function') {
+      fakeModule[key] = contents[key]
       continue
     }
 
     // uses Reflect to implement Proxy, but use 'current value' and not the original target
+    // this is a Proxy pointing _to_ Reflect, which means it "supports all operations" but on the dynamic target
     const currentValueReflect = new Proxy(Reflect, {
       get(_reflect, reflectKey) {
-        // TODO: does this support ctor mocking (new.target)
-        return (...args: any[]) => (Reflect as any)[reflectKey](value, ...args.slice(1))
+        return (...args: any[]) => (Reflect as any)[reflectKey](contents[key], ...args.slice(1))
       },
     })
-    const p = new Proxy(neverFunctionForProxy, currentValueReflect)
-
-    const get = () => p
-    get[originalSymbol] = true
-
-    const set = (v: any) => value = v
-    set[originalSymbol] = true
-
-    Object.defineProperty(module, key, {
-      get,
-      set,
-      enumerable: true,
-      configurable: true,
-    })
+    fakeModule[key] = new Proxy(neverFunctionForProxy, currentValueReflect)
   }
 }
 
 // this enables bound-like named imports with the import rewriting scheme
-export function buildFakeModule<T extends Record<string | symbol, any>>(contents: T): T {
-  const out: Record<string | symbol, any> = { [Symbol.toStringTag]: 'Module', ...contents }
+export function buildFakeModule<T extends Record<string | symbol, any>>(module: T): T {
+  const fakeModule: Record<string | symbol, any> = { [Symbol.toStringTag]: 'Module' }
+  const contents: Record<string | symbol, any> = { ...module }
 
-  prepareFakeModuleForSpy(out)
+  prepareFakeModuleForSpy(fakeModule, contents)
 
   // this intercepts tinyspy which uses Object.defineProperty to rewrite an object (even a "module")
   return new Proxy({}, {
     ownKeys(_target) {
-      return Reflect.ownKeys(out)
+      return Reflect.ownKeys(fakeModule)
     },
 
-    get(_target, p) {
-      return out[p]
+    set(_target, _property) {
+      return false
+    },
+
+    get(_target, property) {
+      return fakeModule[property]
     },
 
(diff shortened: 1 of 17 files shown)


packages/browser/src/node/esmProxy.ts
```diff
@@ -0,0 +1,96 @@
...
+        // this is import(not-a-literal), ignore: can't mock this for now
+        return
+      }
+
+      const { value } = expression
+      const resolved = resolvedImports[value]
+      if (!resolved)
+        return
+
+      s.overwrite(node.start, node.end, `__vitest_mocker__.import(${JSON.stringify(resolved)}, ${JSON.stringify(value)}, import.meta.url)`)
```
samthor, line 85: fyi: This step basically stops further module resolution working normally, because it no longer looks like a dynamic/static import.

This is _probably_ why React was unhappy.

packages/browser/src/client/fakeModule.ts
```diff
@@ -0,0 +1,91 @@
...
+        out[property] = contents[property]
+        return true
+      }
+
+      if ('value' in attributes) {
+        out[property] = attributes.value
+        return true
+      }
+
+      if ('get' in attributes) {
```
samthor, line 78: fyi: I am trying to simplify this. It's more complex than it needs to be _because_ `tinyspy` sees it as a getter/setter combo (which I set up on line 37); if it's just a regular value, I think `tinyspy` will just try to reset it when it's done.

samthor:
FYI I simplified the approach a bit, using the natural comparison of imported modules to match their 'proxies'. This makes the `import()` code less weird and relies more on whatever Vite does.

samthor:
@sheremet-va Is this approach interesting to the team? Should I bother rebasing it?

I think being able to `.spyOn` module imports that are callables is like 50% of the way there for most people (I can `.spyOn` to replace a React component before it's rendered, or a function before it's called), and it's a relatively simple PR.

sheremet-va:
> @sheremet-va Is this approach interesting to the team? Should I bother rebasing it?
> 
> I think being able to `.spyOn` module imports that are callables is like 50% of the way there for most people (I can `.spyOn` to replace a React component before it's rendered, or a function before it's called), and it's a relatively simple PR.

I am not sure yet how this works from your PR. Can you explain by providing code examples? How does the transformed file look in the end?

For now the priority is to finish #5036

samthor:
There's actually very little transform happening. Think of it this way:

1. Every source file has its static imports converted to dynamic ones, `import { foo } from 'x'` => `const { foo } = await import ('x');`
   - This now happens _everywhere_, not just where the `vi.mock` regexp is found. The regexp helps with _hoisting_ mocks, but its behavior isn't really relevant to the this PR.

2. Each import is then wrapped in a helper which converts the output to `Proxy`, `const { foo } = await __vitest_mocker__.wrap(import('x'));

3. The wrap function creates faux-callables which are basically ways to dereference their internal contents, which can also be set.

Here's what the final transformed source might look like:

```ts
// src.ts: the source of foo function
export function foo() { return 1; }

// consumer.ts: someone random that uses foo
const { foo } = await __vitest_mocker__.wrap(import('./src.ts'));
// outputs 1
console.debug(foo());
const oldFoo = foo;

const wholeModule = await __vitest_mocker__.wrap(import('./src.ts'));
// this is what spyOn does internally
Object.defineProperty(wholeModule, 'foo', { get() { return () => 2; } });

// outputs 2
console.debug(foo());
const newFoo = foo;

// true, because 'foo' just dereferences the current value when called
console.info(newFoo === oldFoo);
```

The key is that the consumer of `foo` always has the same reference, but now it's basically to a helper which passes through to whatever _current_ value is stored, which can be changed by `defineProperty`. (It could be by literally saying `wholeModule.foo = ...`, but `spyOn` doesn't try that internally.)

This whole PR basically does the same as `slowHijackESM` for `.spyOn`, _but_ continues to work in the React case as it doesn't mess up the default import.

sheremet-va:
> Every source file has its static imports converted to dynamic ones

This would break live bindings though.

This is why the previous approach changes the imports and all references of the module use the new export. This also allows the browser to fetch static imports in a more performant way.

samthor:
> This would break live bindings though.

Good point. It's already a compromise made for mocking whole modules in vitest, though, and it's [not called out](https://vitest.dev/guide/mocking.html#modules).

> to fetch static imports in a more performant way.

I appreciate this isn't a perfect solution, but it does help people test inside a browser.

If this approach isn't good for you (even behind a flag, where we can iterate on it) I can go away and think about a better solution that still fixes #4264. I suspect it might be a combination of this and `slowHijackESM`.

sheremet-va:
> It's already a compromise made for mocking whole modules in vitest

It was a bug and it has been fixed for some time now. The bug was introduced when we migrated to more browser-friendly solution, but now it uses a code transform similar to Vite SSR.

> I appreciate this isn't a perfect solution, but it does help people test inside a browser.

I believe we can provide a solution that works well out of the box, I don't feel like having it now would greatly benefit us to be honest. I think we should build on the current solution with ESM hijacking and the new mocker transform.

samthor:
Fair enough, I misunderstood the project's goals re: live binding.

Issue #4731, opened 2023-12-11
Make it possible to override pool and poolOptions per workspace


ArtemZag:
### Clear and concise description of the problem

As a developer I would like to have possibility to customize the following configuration parameters per each workspace:
- singleFork
- singleThreads
- pool
- poolOptions

Actually pools implementation already assumes it can be configured differently, it's just needed to pass it down correctly.

### Suggested solution

Extend list of configuration parameters supported by workspace level.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

sheremet-va:
> * poolOptions (including "threads.isolate")

This cannot be inherited because for performance reasons the pool is initiated for Vitest process, not a workspace project (it's the whole reason why we have workspace support). This is how it worked since workspaces were introduced.

The only config options that _can_ be inherited:

- pool
- poolOptions.forks.singleFork
- poolOptions.threads.singleThread

sheremet-va:
I think it is a bug. To me, this sounds like we missed this in the review, I did not expect that the mentioned options were not inherited.

AriPerkkio:
Looks like `pool` is already supported on workspace project level:
- https://stackblitz.com/edit/vitest-dev-vitest-z1zwko?file=vitest.workspace.ts

But is `pool` ignored when it's defined on root level? https://stackblitz.com/edit/vitest-dev-vitest-jyrgtp?file=vite.config.ts

sheremet-va:
> But is `pool` ignored when it's defined on root level? https://stackblitz.com/edit/vitest-dev-vitest-jyrgtp?file=vite.config.ts

Yes, this is consistent with [the documentation](https://vitest.dev/guide/workspace.html#configuration): 

> None of the configuration options are inherited from the root-level config file

sheremet-va:
> I would have expected it to be used similarly as `reporters` or `coverage` as it was ignored on workspace project level.
> 
> 
> 
> Does this eventually mean that when using workspaces, it is not possible to configure `poolOptions.<pool>.isolate` at all? Not at project level, not at root level?

[...]

PR #4765, opened 2023-12-17
fix(workspace): support overring `pool` and `poolOptions` on project level

AriPerkkio:
### Description

- Fixes https://github.com/vitest-dev/vitest/issues/4731

Adds support for specifying following options in workspace projects configurations:

- `pool`
- `poolOptions.threads.isolate`
- `poolOptions.threads.singleThread`
- `poolOptions.vmThreads.isolate`
- `poolOptions.forks.isolate`
- `poolOptions.forks.singleFork`

Example:

```ts
import { defineWorkspace } from 'vitest/config'

  {
    // Tests with side effects - require isolation
    test: {
      name: 'DOM tests',
      include: ['./src/web/**'],
    },
  },
  {
    // Tests that don't have side effects - can be run without isolation
    test: {
      name: 'Side-effect free unit tests',
      include: ['./src/utils/**'],
      poolOptions: { 
        threads: {
          singleThread:  true,
          isolate: false
        }
      },
    },
  },
  {
    // Tests that require forks pool for compatibility reasons
    test: {
      name: 'Database tests',
      include: ['./src/db/**'],
      pool: 'forks',
    },
  },
])
```

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/workspace.ts
+++ b/packages/vitest/src/node/workspace.ts
@@ -319,12 +319,28 @@ export class WorkspaceProject {
 
   getSerializableConfig() {
     const optimizer = this.config.deps?.optimizer
+    const poolOptions = this.config.poolOptions
+
+    // Resolve from server.config to avoid comparing against default value
+    const isolate = this.server?.config?.test?.isolate
+
     return deepMerge({
       ...this.config,
       coverage: this.ctx.config.coverage,
 
-      pool: this.ctx.config.pool,
-      poolOptions: this.ctx.config.poolOptions,
+      poolOptions: {
+        forks: {
+          singleFork: poolOptions?.forks?.singleFork ?? this.ctx.config.poolOptions?.forks?.singleFork ?? false,
+          isolate: poolOptions?.forks?.isolate ?? isolate ?? this.ctx.config.poolOptions?.forks?.isolate ?? true,
+        },
+        threads: {
+          singleThread: poolOptions?.threads?.singleThread ?? this.ctx.config.poolOptions?.threads?.singleThread ?? false,
+          isolate: poolOptions?.threads?.isolate ?? isolate ?? this.ctx.config.poolOptions?.threads?.isolate ?? true,
+        },
+        vmThreads: {
+          singleThread: poolOptions?.vmThreads?.singleThread ?? this.ctx.config.poolOptions?.vmThreads?.singleThread ?? false,
+        },
+      },
 
       reporters: [],
       deps: {
--- a/packages/vitest/src/types/config.ts
+++ b/packages/vitest/src/types/config.ts
@@ -817,7 +817,6 @@ export type ProjectConfig = Omit<
   | 'update'
   | 'reporters'
   | 'outputFile'
-  | 'pool'
   | 'poolOptions'
   | 'teardownTimeout'
   | 'silent'
@@ -842,6 +841,11 @@ export type ProjectConfig = Omit<
 > & {
   sequencer?: Omit<SequenceOptions, 'sequencer' | 'seed'>
   deps?: Omit<DepsOptions, 'moduleDirectories'>
+  poolOptions?: {
+    threads?: Pick<NonNullable<PoolOptions['threads']>, 'singleThread' | 'isolate'>
+    vmThreads?: Pick<NonNullable<PoolOptions['vmThreads']>, 'singleThread'>
+    forks?: Pick<NonNullable<PoolOptions['forks']>, 'singleFork' | 'isolate'>
+  }
 }
 
 export type RuntimeConfig = Pick<
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -655,7 +655,7 @@ Please, be aware of these issues when using this option. Vitest team cannot fix 
 - **Type:** `Record<'threads' | 'forks' | 'vmThreads', {}>`
 - **Default:** `{}`
 
-#### poolOptions.threads<NonProjectOption />
+#### poolOptions.threads
 
 Options for `threads` pool.
 
@@ -687,7 +687,7 @@ Maximum number of threads. You can also use `VITEST_MAX_THREADS` environment var
 
 Minimum number of threads. You can also use `VITEST_MIN_THREADS` environment variable.
 
-##### poolOptions.threads.singleThread<NonProjectOption />
+##### poolOptions.threads.singleThread
 
 - **Type:** `boolean`
 - **Default:** `false`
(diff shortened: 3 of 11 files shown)


packages/vitest/src/node/workspace.ts
```diff
@@ -315,12 +315,26 @@ export class WorkspaceProject {
...
+
     return deepMerge({
       ...this.config,
       coverage: this.ctx.config.coverage,
 
-      pool: this.ctx.config.pool,
-      poolOptions: this.ctx.config.poolOptions,
+      poolOptions: {
+        forks: {
+          ...this.ctx.config.poolOptions?.forks,
```
AriPerkkio, line 326: Should we actually spread the root config here? As mentioned on https://github.com/vitest-dev/vitest/issues/4731#issuecomment-1852921150, projects do not inherit `pool` and `poolOptions` from root configuration. 

And as project level configuration does not allow setting `pool.<name>.isolate`, does this mean that when using workspaces, there is no way to configure `isolate` at all?

packages/vitest/src/types/config.ts
```diff
@@ -843,6 +842,11 @@ export type ProjectConfig = Omit<
 > & {
   sequencer?: Omit<SequenceOptions, 'sequencer' | 'seed'>
   deps?: Omit<DepsOptions, 'moduleDirectories'>
+  poolOptions?: {
+    threads?: Pick<NonNullable<PoolOptions['threads']>, 'singleThread'>
+    vmThreads?: Pick<NonNullable<PoolOptions['vmThreads']>, 'singleThread'>
+    forks?: Pick<NonNullable<PoolOptions['forks']>, 'singleFork'>
+  }
```
AriPerkkio, line 849: TODO: 

```suggestion
  poolOptions?: {
    threads?: Pick<NonNullable<PoolOptions['threads']>, 'singleThread' | 'isolate'>
    vmThreads?: Pick<NonNullable<PoolOptions['vmThreads']>, 'singleThread' | 'isolate'>
    forks?: Pick<NonNullable<PoolOptions['forks']>, 'singleFork' | 'isolate'>
  }
```

And make sure `poolOptions` from root `vitest.config.ts` is ignored. Pool is always initialized to use multiple workers and with isolation. Workspace projects can disable `isolation` in runner - not on pool level. 
Project level `singleThread`/`singleWorker` affects how many files are passed to Tinypool at a time.

packages/vitest/src/node/workspace.ts
```diff
@@ -319,12 +319,25 @@ export class WorkspaceProject {
...
+
     return deepMerge({
       ...this.config,
       coverage: this.ctx.config.coverage,
 
-      pool: this.ctx.config.pool,
-      poolOptions: this.ctx.config.poolOptions,
+      poolOptions: {
+        forks: {
+          singleFork: poolOptions?.forks?.singleFork ?? this.ctx.config.poolOptions?.forks?.singleFork ?? false,
```
sheremet-va, line 330: We also have root-level `isolate` now. I think it's correctly processed only for the core workspace project (so, only the main config file). If you set `isolate: false` in the workspace config, it won't be accounted for when running tests

sheremet-va (changes requested):

packages/vitest/src/node/workspace.ts
AriPerkkio: Root level `isolate` is used to set `poolOptions.<nam>.isolate` values early on Vite plugins. At this point we can rely on the resolved configuration values. Example below with current implementation:

```ts
// <root>/vitest.config.ts

export default defineConfig({ test: { isolate: false } })
```

```ts
// <root>/vitest.workspace.ts
import { defineWorkspace } from 'vitest/config'

  {
    // This will be run without isolation
    // Inside worker: globalThis.__vitest_worker__.config.poolOptions.threads.isolate === false
    test: { pool: 'threads' },
  },
  {
    // This will be run with isolation
    // Inside worker: globalThis.__vitest_worker__.config.poolOptions.threads.isolate === true
    test: {
      pool: 'threads',
      poolOptions: { threads: { isolate: true } },
    },
  },
])
```

So projects are using `test.isolate` from root level configuration, but allow overriding it via project level configuration. It's working correctly, right?

sheremet-va: By root level I mean in the same config, not the core project config.

sheremet-va:
```ts
  {
    // This will be run with isolation
    // Inside worker: globalThis.__vitest_worker__.config.poolOptions.threads.isolate === true
    test: {
      pool: 'threads',
      isolate: true,
    },
  },
])
```

^ This is what I am talking about. From what I can see here, it doesn't work like that. `test.isolate` is only resolved for `poolOptions` inside the plugin in the core config, not in workspace configs.

AriPerkkio: Added `test.isolate` to project config resolving. There's also test cases for this now.

AriPerkkio: Now `poolOptions` are ignored. I guess `test.isolate` is resolved earlier and is not `undefined`. 🤔

AriPerkkio: Root-level `isolate` is now only used if it's set to `false`. If it's the default `true` we'll continue checking other options. But this also means that explicitly set `true` will be ignored.

sheremet-va: We need to refactor how configs are resolved, this is a mess 😅

packages/vitest/src/node/workspace.ts
```diff
@@ -319,12 +319,28 @@ export class WorkspaceProject {
 
   getSerializableConfig() {
     const optimizer = this.config.deps?.optimizer
+    const poolOptions = this.config.poolOptions
+
+    // Use isolate from root only if it's explicitly disabled. Otherwise fallback to other options.
+    const isolate = this.config.isolate === false ? false : undefined
```
sheremet-va, line 325: Maybe we can resolve this in `setServer` before we resolve the config?

AriPerkkio: Default value is set already before `setServer`:

https://github.com/vitest-dev/vitest/blob/a73c1c2ca3a3ca93b317657b9328889434c36344/packages/vitest/src/node/plugins/workspace.ts#L101-L106

But what we could do is resolve `isolate` from `this.server` instead here:

```suggestion
    // Resolve from server.config to avoid comparing against default value
    const isolate = this.server?.config?.test?.isolate
```

sheremet-va: Sounds good to me

sheremet-va (approved):

PR #4781, opened 2023-12-19
fix(vm-threads): tests not cancelled on key press, cancelled tests shown twice

AriPerkkio:
### Description

- Fixes bug where test run is not cancelled when one of the registered keys is pressed. This is only present on `pool: 'vmThreads'`.
- Fixes bug where cancelled tests were shown twice by reporters

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/pools/browser.ts
+++ b/packages/vitest/src/node/pools/browser.ts
@@ -44,7 +44,7 @@ export function createBrowserPool(ctx: Vitest): ProcessPool {
     if (project.config.browser.isolate) {
       for (const path of paths) {
         if (isCancelled) {
-          ctx.state.cancelFiles(files.slice(paths.indexOf(path)), ctx.config.root, project.getName())
+          ctx.state.cancelFiles(files.slice(paths.indexOf(path)), ctx.config.root, project.config.name)
           break
         }
 
--- a/packages/vitest/src/node/pools/child.ts
+++ b/packages/vitest/src/node/pools/child.ts
@@ -118,7 +118,7 @@ export function createChildProcessPool(ctx: Vitest, { execArgv, env, forksPath }
 
         // Intentionally cancelled
         else if (ctx.isCancelling && error instanceof Error && /The task has been cancelled/.test(error.message))
-          ctx.state.cancelFiles(files, ctx.config.root, project.getName())
+          ctx.state.cancelFiles(files, ctx.config.root, project.config.name)
 
         else
           throw error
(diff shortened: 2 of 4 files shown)


AriPerkkio:
There's still one more bug to fix. Once tests have been interrupted and are re-run, the "skip status" of the tests is not reseted. Earlier this line has done it:

https://github.com/vitest-dev/vitest/blob/9497f484cf55392cac224fe0d049d8d5242f9564/packages/vitest/src/node/pools/vm-threads.ts#L91

Skip status is set here:

https://github.com/vitest-dev/vitest/blob/9497f484cf55392cac224fe0d049d8d5242f9564/packages/vitest/src/node/state.ts#L153-L168

This bug is present on all pools.

packages/vitest/src/node/pools/browser.ts
```diff
@@ -44,7 +44,7 @@ export function createBrowserPool(ctx: Vitest): ProcessPool {
     if (project.config.browser.isolate) {
       for (const path of paths) {
         if (isCancelled) {
-          ctx.state.cancelFiles(files.slice(paths.indexOf(path)), ctx.config.root, project.getName())
+          ctx.state.cancelFiles(files.slice(paths.indexOf(path)), ctx.config.root, project.config.name)
```
AriPerkkio, line 47: This needs to match with the collected files:

https://github.com/vitest-dev/vitest/blob/9497f484cf55392cac224fe0d049d8d5242f9564/packages/runner/src/collect.ts#L28

This feels a bit hacky. The `WorkspaceProject.getSerializableConfig()` always returns `undefined` for `.name` instead of `''` when there's no name set for the project (no workspaces used).

The `state.clearFiles` uses the `projectName` when comparing files. When they don't match, there can be duplicate entries where one is from collection and one from cancelling:

```js
'/x/vitest-example-project/vitest/test/sum.2.test.js' => [
  {
    filepath: '/x/vitest-example-project/vitest/test/sum.2.test.js',
    id: '/x/vitest-example-project/vitest/test/sum.2.test.js',
    mode: 'skip',   // <-- From test cancel
    name: 'test/sum.2.test.js',
    projectName: '',
    type: 'suite',
    ...
  },
  {
    filepath: '/x/vitest-example-project/vitest/test/sum.2.test.js',
    mode: 'run',     // <-- From test collect
    name: 'test/sum.2.test.js',
    type: 'suite',
    ...
  }
],
```

These show up in reports like below:

https://github.com/vitest-dev/vitest/assets/14806298/fe61cb8b-e47d-40c8-a9f0-e77a2cb1b519

sheremet-va (approved):

PR #4802, opened 2023-12-24
fix: don't crash when memoryLimit is assumed to be a percentage of system memory

inottn:
### Description

When I upgraded vitest to version 1.1.0 and replaced the original `experimentalVmWorkerMemoryLimit` config with `poolOptions.vmThreads.memoryLimit`, an error occurred when running `pnpm test`: "For a percentage based memory limit, a percentageReference must be provided."

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/pools/vm-threads.ts
+++ b/packages/vitest/src/node/pools/vm-threads.ts
@@ -167,6 +167,7 @@ function getMemoryLimit(config: ResolvedConfig) {
     )
   }
 
+// If totalmem is not supported we cannot resolve percentage based values like 0.5, "50%"
   if ((typeof limit === 'number' && limit > 1) || (typeof limit === 'string' && limit.at(-1) !== '%'))
     return stringToBytes(limit)
 
--- a/packages/vitest/src/node/pools/vm-threads.ts
+++ b/packages/vitest/src/node/pools/vm-threads.ts
@@ -167,9 +167,9 @@ function getMemoryLimit(config: ResolvedConfig) {
     )
   }
 
-  if (limit && limit > 1)
+  if ((typeof limit === 'number' && limit > 1) || (typeof limit === 'string' && limit.at(-1) !== '%'))
     return stringToBytes(limit)
 
-  // just ignore "experimentalVmWorkerMemoryLimit" value because we cannot detect memory limit
+  // just ignore "memoryLimit" value because we cannot detect memory limit
   return null
 }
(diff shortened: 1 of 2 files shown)


sheremet-va:
> When I upgraded vitest to version 1.1.0 and replaced the original `experimentalVmWorkerMemoryLimit` config with `poolOptions.vmThreads.memoryLimit`, an error occurred when running `pnpm test`: "For a percentage based memory limit, a percentageReference must be provided."

Can you give an example of a value that throws an error here?

inottn:
I'm sorry for not providing a specific example in the description earlier.

Here is an example that throws an error when `memoryLimit` is less than 1.

```mjs
import { cpus } from 'os';

const cpuNum = Math.max(cpus().length - 1, 1);

export default defineConfig({
  test: {
    // ...other config
    pool: 'vmThreads',
    poolOptions: {
      vmThreads: {
        memoryLimit: Math.min(1 / (cpuNum * 2), 0.1),
      },
    },
  },
});

```

AriPerkkio:
I can reproduce the crash. But now that I'm looking at `function getMemoryLimit` I can't remember all parts of its implementation. 😅

I'll need to do some more debugging and testing before we can merge this.

packages/vitest/src/node/pools/vm-threads.ts
```diff
@@ -167,9 +167,9 @@ function getMemoryLimit(config: ResolvedConfig) {
     )
   }
 
-  if (limit && limit > 1)
+  if ((typeof limit === 'number' && limit > 1) || (typeof limit === 'string' && limit.at(-1) !== '%'))
```
AriPerkkio, line 170:
```suggestion
// If totalmem is not supported we cannot resolve percentage based values like 0.5, "50%"
  if ((typeof limit === 'number' && limit > 1) || (typeof limit === 'string' && limit.at(-1) !== '%'))
```

packages/vitest/src/node/pools/vm-threads.ts
```diff
@@ -167,9 +167,10 @@ function getMemoryLimit(config: ResolvedConfig) {
     )
   }
 
-  if (limit && limit > 1)
+// If totalmem is not supported we cannot resolve percentage based values like 0.5, "50%"
```
AriPerkkio, line 170:
```suggestion
  // If totalmem is not supported we cannot resolve percentage based values like 0.5, "50%"
```

AriPerkkio (approved):
Looks good, thanks!

PR #4826, opened 2023-12-29
fix: match jest json output by reporting location of error in test file

bard:
### Description

This change improves compatibility of the JSON reporter with Jest's `--json` output by pointing to the last stack frame (where the error surfaced in the test) instead of the first. 

#### Repro

```sh
$ git clone https://github.com/bard/vitest-json-reporter-bugs-repros && cd vitest-json-reporter-bugs-repros && pnpm i
$ pnpm run repro-lineindex-bug

$ pnpm repro-location-bug

> repro@1.0.0 repro-location-bug /tmp/vitest-json-reporter-bugs-repros
> echo '### WITH JEST:'; pnpm run location-bug-jest; echo; echo '### WITH VITEST:'; pnpm run location-bug-vitest

### WITH JEST:

> repro@1.0.0 location-bug-jest /tmp/vitest-json-reporter-bugs-repros
> jest --json --testLocationInResults src/location-bug/jest.test.ts 2>/dev/null | jq .testResults[0].assertionResults[0].location

{
  "column": 5,
  "line": 4
}

### WITH VITEST:

> repro@1.0.0 location-bug-vitest /tmp/vitest-json-reporter-bugs-repros
> vitest run --reporter=json src/location-bug/vitest.test.ts | jq .testResults[0].assertionResults[0].location

{
  "line": 1,
  "column": 8
}
```

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/reporters/json.ts
+++ b/packages/vitest/src/node/reporters/json.ts
@@ -188,7 +188,7 @@ export class JsonReporter implements Reporter {
       getSourceMap: file => project.getBrowserSourceMapModuleById(file),
       frameFilter: this.ctx.config.onStackTrace,
     })
-    const frame = stack[0]
+    const frame = stack[stack.length - 1]
     if (!frame)
       return
 


sheremet-va:
I don't think we can make this change. It was already reverted in #3645

bard:
> I don't think we can make this change. It was already reverted in #3645

@sheremet-va apologies for not searching previous issues enough.

I'm a little perplexed. Consider the following output relating to a test run involving `index.ts` and `index.test.ts`, where an error was thrown inside `index.ts`:

```
  "testResults": [
    {
      "assertionResults": [
        {
          "ancestorTitles": [
            ""
          ],
          "fullName": " addition",
          "status": "failed",
          "title": "addition",
          "duration": 8,
          "failureMessages": [
            "error thrown in source file"
          ],
          "location": {
            "line": 4,
            "column": 9
          }
        }
      ],
      "startTime": 1704030241630,
      "endTime": 1704030241638,
      "status": "failed",
      "message": "",
      "name": "/tmp/repro/src/index.test.ts"
    }
  ]
```

There's a line and column number, but no way of telling in which file that line and column are, which makes the `location` field... not very useful.

Perhaps the actual bug here (if alignment with `jest` is not desired) is that there should be a `file` property in `location`?

sheremet-va:
> There's a line and column number, but no way of telling in which file that line and column are, which makes the `location` field... not very useful.

What even is the location in json result? Is it the `test` function location? Or error location? I can also see that jest prints the whole stack trace in the failure message - maybe we should also do that?

bard:
> > There's a line and column number, but no way of telling in which file that line and column are, which makes the `location` field... not very useful.
> 
> What even is the location in json result? Is it the `test` function location? Or error location? I can also see that jest prints the whole stack trace in the failure message - maybe we should also do that?

Jest seems to define location as "location of the `test()` call for the test that failed".

A stack trace would be great IMHO. Sure, it requires parsing, but it has all the information and none of the ambiguity. If it's an acceptable addition, I'd be happy to look into adding it to the JSON reporter.

sheremet-va:
> Jest seems to define location as "location of the `test()` call for the test that failed".

Maybe we should try to actually locate it then? Both the first and the last frames are not the `test` call. On top of that, I think we can add a stack trace and/or parsed stack trace (we always parse it, so it's always available).

bard:
> > Jest seems to define location as "location of the `test()` call for the test that failed".
> 
> Maybe we should try to actually locate it then? Both the first and the last frames are not the `test` call. On top of that, I think we can add a stack trace and/or parsed stack trace (we always parse it, so it's always available).

I've opened https://github.com/vitest-dev/vitest/pull/4843 for the stack trace part (it's minimal). Will close this one.

PR #4886, opened 2024-01-06
fix(vite-node): dedupe -v flag

GauBen:
### Description

Current `vite-node -h` output includes `-v, --version` twice, as `cac` [will add one for us](https://www.npmjs.com/package/cac#cliversionversion-customflags)

```
Options:
  -r, --root <path>    Use specified root directory 
  -c, --config <path>  Use specified config file 
  -m, --mode <mode>    Set env mode 
  -w, --watch          Restart on file changes, similar to "nodemon" 
  --script             Use vite-node as a script runner 
  --options <options>  Use specified Vite server options 
  -v, --version        Output the version number 
  -h, --help           Display help for command 
  -v, --version        Display version number 
```

I removed custom version handling and let `cac` do all the work

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vite-node/src/cli.ts
+++ b/packages/vite-node/src/cli.ts
@@ -19,8 +19,8 @@ cli
   .option('-w, --watch', 'Restart on file changes, similar to "nodemon"')
   .option('--script', 'Use vite-node as a script runner')
   .option('--options <options>', 'Use specified Vite server options')
-  .option('-v, --version', 'Output the version number')
   .option('-h, --help', 'Display help for command')
+  .version(version)
 
 cli
   .command('[...files]')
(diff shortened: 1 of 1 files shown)


sheremet-va:
There was a reason for this change: #3574

GauBen:
Ok sorry for the noise, should we open an issue upstream?

sheremet-va:
> Ok sorry for the noise, should we open an issue upstream?

No, CAC doesn't have to work like Node CLI, we just need to work around it

GauBen:
That makes sense, let's keep things as they are

Thanks for being so reactive @sheremet-va!

PR #4945, opened 2024-01-12
fix(vitest): add inlined deps to ssr.noExternal so they are added to the module graph

### LINKED ISSUE #4785: watchExclude option doesn't work for code in node_modules in some cases
### Describe the bug

From what I tried so far, the `watchExclude` for code in `node_modules` option doesn't work when:

- the environment is `node`.
- the environment is `jsdom` and
  - the first time you edit the main file of a node module, or
  - every time you edit a non-main file of a node module

More discussion info: https://github.com/vitest-dev/vitest/discussions/4772#discussioncomment-7896620

### Reproduction

https://github.com/Jinjiang/reproductions/tree/vitest-watch-node_modules-20231220

```
pnpm install

# you can switch environment between 'node' (default value) and 'jsdom'
pnpm test

# in 'node' environment:
#
# modify any code in node_modules
# - the test is not re-run

# in 'jsdom' environment:
#
# modify code in node_modules/foo-cjs/index.js or node_modules/foo-esm/index.js
# - when you first time modify it, the test is not re-run
# - when you further modify it, the test is re-run
#
# modify code in node_modules/foo-cjs/foo.js or node_modules/foo-esm/foo.js
# - the test is not re-run
```

The expected behavior is that the test is always re-run when any code in `node_modules/foo-cjs` and `node_modules/foo-esm` is modified whatever the environment is.

### System Info

```shell
System:
    OS: macOS 14.3
    CPU: (8) arm64 Apple M1
    Memory: 88.61 MB / 8.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.1.0 - ~/.asdf/installs/nodejs/20.1.0/bin/node
    Yarn: 1.22.19 - ~/.asdf/shims/yarn
    npm: 9.6.4 - ~/.asdf/plugins/nodejs/shims/npm
    pnpm: 8.12.0 - ~/Library/pnpm/pnpm
  Browsers:
    Safari: 17.3
  npmPackages:
    vite: ^5.0.10 => 5.0.10 
    vitest: ^1.0.4 => 1.0.4
```

### Used Package Manager

pnpm

### Validations

sheremet-va:
### Description

Fixes #4785

This fixes the first case when changing code in dependencies with `node` environment did not trigger an HMR update.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vite-node/src/server.ts
+++ b/packages/vite-node/src/server.ts
@@ -65,7 +65,8 @@ export class ViteNodeServer {
       }
       else if (options.deps.inline !== true) {
         options.deps.inline ??= []
-        options.deps.inline.push(...toArray(ssrOptions.noExternal))
+        const inline = options.deps.inline
+        options.deps.inline.push(...toArray(ssrOptions.noExternal).filter(dep => !inline.includes(dep)))
       }
     }
     if (process.env.VITE_NODE_DEBUG_DUMP) {
--- a/packages/vitest/src/config.ts
+++ b/packages/vitest/src/config.ts
@@ -8,6 +8,7 @@ export interface UserWorkspaceConfig extends ViteUserConfig {
 // will import vitest declare test in module 'vite'
 export { configDefaults, defaultInclude, defaultExclude, coverageConfigDefaults } from './defaults'
 export { mergeConfig } from 'vite'
+export { extraInlineDeps } from './constants'
 
 export type { ConfigEnv, ViteUserConfig as UserConfig }
 export type UserConfigFnObject = (env: ConfigEnv) => ViteUserConfig
--- a/packages/vitest/src/constants.ts
+++ b/packages/vitest/src/constants.ts
@@ -6,6 +6,15 @@ export const EXIT_CODE_RESTART = 43
 
 export const API_PATH = '/__vitest_api__'
 
+export const extraInlineDeps = [
+  /^(?!.*(?:node_modules)).*\.mjs$/,
+  /^(?!.*(?:node_modules)).*\.cjs\.js$/,
+  // Vite client
+  /vite\w*\/dist\/client\/env.mjs/,
+  // Nuxt
+  '@nuxt/test-utils',
+]
+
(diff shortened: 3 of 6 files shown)


Jinjiang:
@sheremet-va after testing in the same repo https://github.com/Jinjiang/reproductions/tree/vitest-watch-node_modules-20231220 with v1.2.0:

in 'jsdom' environment:

modify code in `node_modules/foo-cjs/index.js` or `node_modules/foo-esm/index.js` or `node_modules/foo-cjs/foo.js` or `node_modules/foo-esm/foo.js`
- when you first time modify it, the test is not re-run
- when you further modify it, the test is re-run, but the result is the same

in default ('node') environment:

modify code in `node_modules/foo-cjs/index.js` or `node_modules/foo-esm/index.js` or `node_modules/foo-cjs/foo.js` or `node_modules/foo-esm/foo.js`
- when you first time modify it, the test is not re-run
- when you further modify it, the test is re-run, and the result is the updated one

It seems still have issues.

Would you mind check again?

Thanks.

sheremet-va:
I mentioned it in another PR, this is upstream issue with a watcher, it doesn’t emit change event on the first edit - I don’t think there is anything we can do.

The error with jsdom with the same result I will check, it should update the result.

Jinjiang:
I see. For the upstream watcher, is there a GitHub issue that we can keep tracking on? Or do you think we can create one for them? (Sorry I'm not familiar with this behind.)

sheremet-va:
Vite uses [chokidar](https://github.com/paulmillr/chokidar) for its file watcher. I thought the reason why it doesn't work might be because `followSymlinks` is not enabled by default, but toggling it on and off didn't change anything for me unfortunately.

sheremet-va:
Ah, it looks like watcher uses another event for the first change in `node_modules`. I will look more into it.

Jinjiang:
Confirmed it works as expected with v1.2.1. Thanks so much!

Issue #4983, opened 2024-01-16
Method getUnhandledErrors called on RPC, but does not exist and throws error


enjikaka:
### Describe the bug

Running my tests in vitest browser mode with Playwright throws an error inside the vitest JS-bundle.

<img width="1048" alt="Skärmavbild 2024-01-16 kl  20 01 04" src="https://github.com/vitest-dev/vitest/assets/877455/3afd30c0-fa3c-4be6-b485-1386ad31e458">

<img width="1297" alt="Skärmavbild 2024-01-16 kl  20 01 16" src="https://github.com/vitest-dev/vitest/assets/877455/3e543505-5be8-4eec-b980-7422a504c1bc">

https://github.com/vitest-dev/vitest/blob/39814357024e96b81546ec22d6aaf777b8db6cef/packages/ui/client/composables/client/index.ts#L79

### Reproduction

- git clone https://github.com/enjikaka/rutabaga-failing-vitest
- run `npm test`

### System Info

```shell
System:
    OS: macOS 12.7.3
    CPU: (8) x64 Intel(R) Core(TM) i7-6820HQ CPU @ 2.70GHz
    Memory: 2.76 GB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.18.2 - ~/.nvm/versions/node/v18.18.2/bin/node
    npm: 9.8.1 - ~/.nvm/versions/node/v18.18.2/bin/npm
    bun: 1.0.22 - ~/.bun/bin/bun
  Browsers:
    Brave Browser: 120.1.61.109
    Chrome Canary: 122.0.6249.0
    Safari: 17.1.2
  npmPackages:
    @vitest/browser: ^1.2.0 => 1.2.0
    vite: ^5.0.11 => 5.0.11
    vitest: ^1.1.3 => 1.1.3
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
[...]

PR #4985, opened 2024-01-16
Require vitest 1.2.0 to run @vitest/browser >1.2.0

enjikaka:
### Description

Vitest browser is not compatible with Vitest versions below 1.2.0. See: https://github.com/vitest-dev/vitest/issues/4983
Change peerDependency version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,7 +77,7 @@ jobs:
         run: pnpm i
 
       - name: Install Playwright Dependencies
-        run: pnpm exec playwright install
+        run: pnpm playwright install --with-deps
 
       - name: Build
         run: pnpm run build
@@ -130,7 +130,7 @@ jobs:
         run: pnpm i
 
       - name: Install Playwright Dependencies
-        run: pnpx playwright install chromium
+        run: pnpm playwright install chromium
 
       - name: Build
         run: pnpm run build
@@ -167,7 +167,7 @@ jobs:
         run: pnpm i
 
       - name: Install Playwright Dependencies
-        run: pnpm exec playwright install
+        run: pnpm playwright install --with-deps
 
       - name: Build
         run: pnpm run build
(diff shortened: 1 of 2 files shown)


enjikaka:
For some reason tests get bonkers with the bump of vitest version on CI here. Not sure why playwright errors on not being installed when the GH action installs it?

AriPerkkio:
This is a breaking change without semver major bump. 😕

AriPerkkio:
Updating `@vitest/browser` package to new version without updating its peer dependency `vitest` will no longer work for users. Requiring a new minimum version of peer dependency is breaking change in semver. There's more discussion at https://github.com/semver/semver/issues/502.

If it was direct dependency listed in `dependencies`, or inlined during bundling from `devDependencies` it would not be an issue. Requiring actions from users without major version bump is an issue.

sheremet-va:
This sounds a bit restrictive, to be honest. Vite plugins update their peer dependencies without a major bump. I am also not convinced when ljharb is involved in the conversation knowing his stances on other topics 😅 

Making it a breaking change would mean that adding a new feature to Vitest is always a breaking change (for other packages, but in out case for all packages because it's a monorepo) because other packages might rely on it 🤔

AriPerkkio:
I don't like it either but it makes sense. I think we could state in all `@vitest/` sub-packages that they do follow semver, but they must be used with same version of `vitest`. Now when we are making changes to multiple packages in monorepo setup it is extremely difficult to spot these breaking API changes.

If that is the decided way to go, we should automate release process to update `peerDependency` versions to all subpackages to match the current release. cc. @antfu - you've been part of the earlier `peerDependency` related conversations.

antfu:
I don't consider this as breaking change at all. Breaking change could be inherited when you bump a dependency that contains breaking change, where here, vitesse 1.0 -> 1.2 shouldn't include breaking changes. 

Ideally vitest packages are supposed to be used with a same version - any use valid use case of `@vitest/browser@1.2` + `vitest@1.2`? If not, I don't think we should even allows that (could create bugs that hard to find). To lock the version, we could use `vitest: workspace:*` - during publishing it will be replaced by pnpm with the exact version - I don't know if it works for peerDeps tho, but I think it could also directly be in `dependencies`

My judge of `peerDeps` vs `deps` is "if it works without it" - if not, it should be `dependencies`

PR #5036, opened 2024-01-23
feat(browser): run test files in isolated iframes

sheremet-va:
### Description

This PR changes how Vitest runs browser tests internally. This is a successor to #3584, but without any visuals.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/browser/src/node/index.ts
+++ b/packages/browser/src/node/index.ts
@@ -32,6 +32,7 @@ export default (project: WorkspaceProject, base = '/'): Plugin[] => {
         const runnerHtml = readFile(resolve(distRoot, 'client/index.html'), 'utf8')
         const injectorJs = readFile(resolve(distRoot, 'client/esm-client-injector.js'), 'utf8')
         const favicon = `${base}favicon.svg`
+        const testerPrefix = `${base}__vitest_test__/__test__/`
         server.middlewares.use((_req, res, next) => {
           const headers = server.config.server.headers
           if (headers) {
@@ -44,21 +45,21 @@ export default (project: WorkspaceProject, base = '/'): Plugin[] => {
           if (!req.url)
             return next()
           const url = new URL(req.url, 'http://localhost')
-          if (!url.pathname.endsWith('__vitest_test__/tester.html') && url.pathname !== base)
+          if (!url.pathname.startsWith(testerPrefix) && url.pathname !== base)
             return next()
-          const id = url.searchParams.get('__vitest_id')
 
-          // TODO: more handling, id is required
-          if (!id) {
-            res.statusCode = 404
-            res.end()
-            return
-          }
-
+          res.setHeader('Cache-Control', 'no-cache, max-age=0, must-revalidate')
           res.setHeader('Content-Type', 'text/html; charset=utf-8')
 
+          const files = project.browserState?.files ?? []
+
+          const config = wrapConfig(project.getSerializableConfig())
+          config.env ??= {}
+          config.env.VITEST_BROWSER_DEBUG = process.env.VITEST_BROWSER_DEBUG || ''
+
           const injector = replacer(await injectorJs, {
-            __VITEST_CONFIG__: JSON.stringify(wrapConfig(project.getSerializableConfig())),
+            __VITEST_CONFIG__: JSON.stringify(config),
+            __VITEST_FILES__: JSON.stringify(files),
           })
 
           if (url.pathname === base) {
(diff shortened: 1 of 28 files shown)


packages/browser/src/node/index.ts
```diff
@@ -23,18 +29,81 @@ export default (project: WorkspaceProject, base = '/'): Plugin[] => {
         }
       },
       async configureServer(server) {
+        const [testerHtml, runnerHtml, injectorJs] = await Promise.all([
```
userquin, line 32: I guess you can defer resolving promises when required, you don't need to block vitest startup/configuration here, middlewares are async, declare them after distRoot declaration (I guess you can also split `Promise.all` into individual promises)

sheremet-va:
@userquin @Aslemammad I don't really like that we run all tests in parallel - this makes it impossible to use keyboard events for example. Should we default to running tests one after another?

userquin:
> @userquin @Aslemammad I don't really like that we run all tests in parallel - this makes it impossible to use keyboard events for example. Should we default to running tests one after another?

What do you mean about using keyboard events? related to this issue https://github.com/vitest-dev/vitest/issues/4918?

IIRC you cannot simulate`Tab` key since we cannot override `isTrusted` (set to true by the browser when the user press the tab key), we should add some helper.

sheremet-va:
> What do you mean about using keyboard events? related to this issue #4918?

Yes, look into the open PR. (I didn't test it yet, it's just a drafter POC). But the idea is that we need to have an element with a focus but if we run tests in parallel the focus can shift between iframes and elements at any time.

userquin:
> > What do you mean about using keyboard events? related to this issue #4918?
> 
> Yes, look into the open PR. (I didn't test it yet, it's just a drafter POC). But the idea is that we need to have an element with a focus but if we run tests in parallel the focus can shift between iframes and elements at any time.

I think there is no need, using `none` provider I have 2 tests files using focus and working: ofc if we have multiple focus tests in the same file the problem will be there.

Add `"test-ui": "PROVIDER=none vitest"` to `test/browser/package.json` and add the following test to `test/browser/test/another.test.ts` and `test/browser/test/basic.test.ts`:

```ts
it('test focus', async () => {
  const host2 = document.createElement('div')
  host2.setAttribute('id', 'host2')
  document.body.appendChild(host2)
  const btn1 = document.createElement('button')
  btn1.innerHTML = 'Click me 1'
  btn1.addEventListener('keydown', (e) => {
    console.log(e)
  })
  btn1.addEventListener('keyup', (e) => {
    console.log(e)
  })
  host2.appendChild(btn1)
  const btn2 = document.createElement('button')
  btn2.innerHTML = 'Click me 2'
  host2.appendChild(btn2)
  expect(document.activeElement).toBe(document.body)
  btn1.focus()
  expect(document.activeElement).toBe(btn1)
  btn2.focus()
  expect(document.activeElement).toBe(btn2)
  btn1.focus()
  expect(document.activeElement).toBe(btn1)
  await new Promise(resolve => setTimeout(resolve, 100))
  btn1.dispatchEvent(new KeyboardEvent('keydown', { key: 'Tab' }))
  btn1.dispatchEvent(new KeyboardEvent('keyup', { key: 'Tab' }))
  /* btn1.dispatchEvent(new KeyboardEvent('keypress', {
    key: 'Tab',
    view: window,
    bubbles: true,
    cancelable: true,
    /!* code: 'Tab',
    isComposing: true,
    charCode: 0,
    keyCode: 9,
    which: 9,
    composed: true,
    location: 0,
    view: window,
    bubbles: true,
    cancelable: true,
    shiftKey: false,
    ctrlKey: false,
    altKey: false,
    metaKey: false, *!/
  })) */
  await new Promise(resolve => setTimeout(resolve, 100))
  console.log(document.activeElement)
  // THIS WILL FAIL
  // expect(document.activeElement).toBe(btn2)
})
```

sheremet-va:
> I think there is no need, using `none` provider I have 2 tests files using focus and working: ofc if we have multiple focus tests in the same file the problem will be there.

There is a great need for that. Dispatching keyboard event doesn't work the same way as the user would do it. In the real world, the keyboard triggers multiple events and it is not possible to emulate the order of it down to the microtask in the browser (just like the click), so we need to do it through a webdriver API or google devtools API.

> Add "test-ui": "PROVIDER=none vitest" (or run it from your terminal, I'm using Windows...) to test/browser/package.json and add the following test to test/browser/test/another.test.ts and test/browser/test/basic.test.ts:

`none` provider is outside of this PR's responsibility and doesn't support the same set of APIs - if users want to "emulate" clicks and keyboard events they can just use `jsdom`.

PR #5056, opened 2024-01-28
fix(browser): support `coverage.reportsDirectory` with multiple directories

AriPerkkio:
### Description

- Fixes https://github.com/vitest-dev/vitest/issues/5013

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/browser/src/node/index.ts
+++ b/packages/browser/src/node/index.ts
@@ -1,9 +1,10 @@
 import { fileURLToPath } from 'node:url'
 
-import { resolve } from 'node:path'
+import { basename, resolve } from 'pathe'
 import sirv from 'sirv'
 import type { Plugin } from 'vite'
 import type { WorkspaceProject } from 'vitest/node'
+import { coverageConfigDefaults } from 'vitest/config'
 import { injectVitestModule } from './esmInjector'
 
 export default (project: WorkspaceProject, base = '/'): Plugin[] => {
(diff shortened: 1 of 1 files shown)


userquin:
@AriPerkkio both client and server should be in sync: https://github.com/vitest-dev/vitest/blob/main/packages/ui/node/index.ts

Check https://github.com/vitest-dev/vitest/pull/3917

AriPerkkio:
Are they currently in sync? Before this fix I see following with `coverage.reportsDirectory: './one/two/three/'`:

- `packages/ui/node/index.ts`: `coverageFolder: [ '/Users/xyz/vitest/test/repro/one/two/three', '/three/' ]`
- `packages/ui/client/composables/navigation.ts`: `coverageUrl: '/three/index.html'`

In browser mode this `'/three/index.html'` is not served. 

Would the correct fix be to add this logic into `packages/browser/src/node/index.ts` as @hi-ogawa suggested in #5013 - I think:

https://github.com/vitest-dev/vitest/blob/d53b85805d2a893276aaf19c9e746d13ad5174f2/packages/ui/node/index.ts#L20

AriPerkkio:
Are there any existing test cases for browser's watch mode? I guess only UI has those e2e-tests?

sheremet-va:
We have a browser test in `test/watch/file-watching.test.ts`

AriPerkkio:
There are no existing test cases where this could be added easily. I've tried to build a new one into `test/ui` but for some reason Playwright sees only a blank page there.

And now I'm again running into the case where browser mode picks non-existing test cases and fails:




I'll try to verify these changes manually with multiple different configuration options once the error above stops showing up. But I'm not sure if adding a regression test is doable for now.

userquin (approved):
The logic seems to be the same included in the `vitest:ui` plugin

AriPerkkio:
I've now tested multiple different variations of `reportsDirectory` and it seems to work OK.

About the error mentioned in https://github.com/vitest-dev/vitest/pull/5056#issuecomment-1921969349. This is something I've seen many times ever since browser mode was implemented. We've also seen this on CI. 

Browser mode seems to attempt to run test cases that do not belong to the current project that is being run. It includes test cases from previous browser mode runs. I'm starting to think there is some kind of 24h caching that causes this. When I run into this bug this time, I tried removing all `node_modules`, installing fresh dependencies and building project. I even rebooted my laptop and it still came up. I searched `node_modules` and `pnpm store path` location for matches of the erroneous test case but there were none. I was hoping some `.vite` or `.vitest` cache would be the root cause. The error came up with all `browser.provider` options - even the `none` one. The error disappeared now that I run tests again today - I had not done any changes in between. 🤷

PR #5076, opened 2024-01-30
feat(vitest): support hooks in bench runner

### LINKED ISSUE #5075: Support hooks in `bench` API

`beforeAll`, `afterAll`, `beforeEach` and `afterEach` are not run when using `vitest bench`

### Suggested solution

Support them in the benchmark runner

### Alternative

_No response_

### Additional context

_No response_

### Validations

adriencaccia:
Fix #5075

### DIFF
--- a/packages/vitest/src/runtime/runners/benchmark.ts
+++ b/packages/vitest/src/runtime/runners/benchmark.ts
@@ -1,5 +1,5 @@
-import type { Suite, Task, VitestRunner, VitestRunnerImportSource } from '@vitest/runner'
-import { updateTask as updateRunnerTask } from '@vitest/runner'
+import type { HookCleanupCallback, Suite, Task, VitestRunner, VitestRunnerImportSource } from '@vitest/runner'
+import { callCleanupHooks, callSuiteHook, updateTask as updateRunnerTask } from '@vitest/runner'
 import { createDefer, getSafeTimers } from '@vitest/utils'
 import { getBenchFn, getBenchOptions } from '../benchmark'
 import { getWorkerState } from '../../utils'
@@ -34,6 +34,10 @@ async function runBenchmarkSuite(suite: Suite, runner: NodeBenchmarkRunner) {
     else if (task.type === 'suite')
       benchmarkSuiteGroup.push(task)
   }
+
+  let beforeAllCleanups: HookCleanupCallback[] = []
+
+  beforeAllCleanups = await callSuiteHook(suite, suite, 'beforeAll', runner, [suite])
 
   if (benchmarkSuiteGroup.length)
     await Promise.all(benchmarkSuiteGroup.map(subSuite => runBenchmarkSuite(subSuite, runner)))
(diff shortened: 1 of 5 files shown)


packages/vitest/src/runtime/runners/benchmark.ts
```diff
@@ -90,7 +94,12 @@ async function runBenchmarkSuite(suite: Suite, runner: NodeBenchmarkRunner) {
       await task.warmup()
       tasks.push([
         await new Promise<BenchTask>(resolve => setTimeout(async () => {
-          resolve(await task.run())
+          let beforeEachCleanups: HookCleanupCallback[] = []
+          beforeEachCleanups = await callSuiteHook(benchmark.suite, benchmark, 'beforeEach', runner, [benchmark.context, benchmark.suite])
```
Aslemammad, line 98: Can we use tinybench's hooks? I don't have any issues with this, but just asking.

adriencaccia: Interesting idea!
I think we might need to refine what each hook does, since in benchmarking there is one more level of hooks.

At the moment, with my implementation, this is what would happen (I added some comment at the end to see the execution order):

```typescript
import {
  afterAll,
  afterEach,
  beforeAll,
  beforeEach,
  bench,
  describe,
} from 'vitest'

describe('hooks', () => {
  beforeAll(() => {
    console.log('beforeAll')
  })
  beforeEach(() => {
    console.log('beforeEach')
  })
  afterEach(() => {
    console.log('afterEach')
  })
  afterAll(() => {
    console.log('beforeAll')
  })

  bench('one', () => {
    console.log('bench one')
  }, {
    setup: () => {
      console.log('setup')
    },
    teardown: () => {
      console.log('tearDown')
    },
    iterations: 10,
    time: 1,
  })
})

// when the bench is executed, the following should be logged

// beforeAll
// beforeEach
// setup     x10
// bench one x10
// tearDown  x10
// afterEach
// beforeAll
```

Do you think it is what should be expected?

As far as using the hooks from tinybench, I can map the `*Each` hooks from vitest to the `Task` `*All` hooks from tinybench.

Aslemammad: Oh interesting edge case, nice, I think let's keep it as is and this should be documented somewhere. 

After all, @sheremet-va might have some interesting ideas and insights around this.

adriencaccia: Okay, I will go and try to add a test that verify the order of execution of the hooks

adriencaccia: I will switch the PR to draft for the moment, and mark it as ready when I will have added the tests

sheremet-va:
Benchmarking API has been rewritten form the ground up: main.vitest.dev/guide/benchmarking

arv:
I don't see the hooks being documented. I'm still worried that we are not exposing in between iteration setup/teardown runs?

sheremet-va:
> I don't see the hooks being documented. I'm still worried that we are not exposing in between iteration setup/teardown runs?

`bench` accepts the same options that tinybench accepts: https://main.vitest.dev/guide/benchmarking.html#options

Issue #5071, opened 2024-01-29
TypeError: Cannot read properties of null (reading 'transformMode')


MichaelBrandt4994:
### Describe the bug

When executing vitest we receive the following error
```
⎯⎯⎯⎯ Unhandled Rejection ⎯⎯⎯⎯⎯
TypeError: Cannot read properties of null (reading 'transformMode')
 ❯ getTransformMode node_modules/vitest/dist/vendor/execute.edwByI27.js:405:31
 ❯ Object.fetchModule node_modules/vitest/dist/vendor/execute.edwByI27.js:421:30
 ❯ VitestExecutor.directRequest node_modules/vite-node/dist/client.mjs:250:65
 ❯ VitestExecutor.cachedRequest node_modules/vite-node/dist/client.mjs:187:28
 ❯ VitestExecutor.dependencyRequest node_modules/vite-node/dist/client.mjs:233:23
 ❯ VitestExecutor.dependencyRequest node_modules/vitest/dist/vendor/execute.edwByI27.js:576:18
```

Sometimes the error doesn't show up and our tests are passing.

Here is our vitest.config.ts
```
import { fileURLToPath, URL } from "node:url"

import vue from "@vitejs/plugin-vue"
import vueJsx from "@vitejs/plugin-vue-jsx"
import VueMacros from "unplugin-vue-macros"
import { defineConfig } from "vitest/config"

export default defineConfig({
  appType: "spa",
  plugins: [VueMacros.vite({ plugins: { vue: vue(), vueJsx: vueJsx() } })],
  resolve: {
    alias: {
      "@": fileURLToPath(new URL("./src", import.meta.url)),
      "./runtimeConfig": "./runtimeConfig.browser",
    },
  },
  envPrefix: "VUE_",
  test: {
    environment: "jsdom",
    transformMode: {
      web: [/.[tj]sx$/],
    },
  },
})
```

### Reproduction

-

### System Info

```shell
System:
    OS: Linux 6.5 Ubuntu 22.04.3 LTS 22.04.3 LTS (Jammy Jellyfish)
    CPU: (8) x64 Intel(R) Core(TM) i7-8665U CPU @ 1.90GHz
    Memory: 16.05 GB / 30.97 GB
    Container: Yes
    Shell: 5.8.1 - /usr/bin/zsh
  Binaries:
    Node: 21.6.0 - ~/.nvm/versions/node/v21.6.0/bin/node
[...]

PR #5117, opened 2024-02-05
fix: execute.ts add nullsafe getTransformMode

ryinner:
### Description

Sometimes environment is null. Errors in vue async components and vue-router lib mocks. [issue](https://github.com/vitest-dev/vitest/issues/5071).

### Tests

### Documentation

### Changesets
-  [x] Changes in changelog are generated from PR name. Please, make sure that it explains your changes in an understandable manner. Please, prefix changeset messages with `feat:`, `fix:`, `perf:`, `docs:`, or `chore:`.

### DIFF
--- a/packages/vitest/src/runtime/execute.ts
+++ b/packages/vitest/src/runtime/execute.ts
@@ -69,7 +69,7 @@ export async function startVitestExecutor(options: ContextExecutorOptions) {
   process.on('unhandledRejection', e => catchError(e, 'Unhandled Rejection'))
 
   const getTransformMode = () => {
-    return state().environment.transformMode ?? 'ssr'
+    return state().environment?.transformMode ?? 'ssr'
   }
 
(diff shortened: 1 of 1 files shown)


sheremet-va:
Why "sometimes"? I don't think this is a fix, environment should always be setup when we call this.

The only time the environment can be removed is the teardown phase, but nothing should be called after that.

ryinner:
> Why "sometimes"? I don't think this is a fix, environment should always be setup when we call this.

Unfortunately, I cannot determine what exactly causes this behavior, the only dependency that I found is that it happens with a mock of the asynchronous functions and it happened with the transition to version 1.2.0.

The problem with the null environment does not happen in every test pass, but only in some, I think it is due to unresolved promises

> The only time the environment can be removed is the teardown phase, but nothing should be called after that.

I agree that this is not the best solution, but it will allow you to use 1.2.* now.

PR #5128, opened 2024-02-06
feat(vitest): add onTestFinished hook

sheremet-va:
### Description

This PR adds a hook that can be called during the test execution to cleanup state:

```ts
import { onTestFinished, test } from 'vitest'

test('performs a query', () => {
  const db = connectDb()
  onTestFinished(() => db.close())
  db.query('SELECT * FROM users')
})
```

We already had `onTestFailed` for some time now, but it was never documented so I added documentation for both.

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -921,7 +921,7 @@ These hooks will throw an error if they are called outside of the test body.
 
 ### onTestFinished <Badge type="info">1.3.0+</Badge>
 
-This hook is called after the test has finished running. It is called after `afterEach` hooks since they can infulence the test result. It receives a `TaskResult` object with the current test result.
+This hook is always called after the test has finished running. It is called after `afterEach` hooks since they can influence the test result. It receives a `TaskResult` object with the current test result.
 
 ```ts
 import { onTestFinished, test } from 'vitest'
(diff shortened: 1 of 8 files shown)


docs/api/index.md
```diff
@@ -906,3 +910,92 @@ afterAll(async () => {
...
+
+Vitest provides a few hooks that you can call _during_ the test execution to cleanup the state when the test has finished runnning.
+
+::: warning
+These hooks will throw an error if they are called outside of the test body.
+:::
+
+### onTestFinished <Badge type="info">1.3.0+</Badge>
+
+This hook is called after the test has finished running. It is called after `afterEach` hooks since they can infulence the test result. It receives a `TaskResult` object with the current test result.
```
AriPerkkio, line 924:
```suggestion
This hook is called after the test has finished running. It is called after `afterEach` hooks since they can influence the test result. It receives a `TaskResult` object with the current test result.
```

AriPerkkio, line 924: It might be worth to mention here that `onTestFinished` is called always - even when test fails.

AriPerkkio (approved):

docs/api/index.md
```diff
@@ -906,3 +910,92 @@ afterAll(async () => {
...
+
+test('performs an organization query', async () => {
+  const db = getTestDb()
+  expect(await db.query('SELECT * from organizations').perform()).toEqual([])
+})
+```
+
+### onTestFailed
+
+This hook is called after the test has failed. It is called after `afterEach` hooks since they can infulence the test result. It receives a `TaskResult` object with the current test result. This hook is useful for debugging.
```
AriPerkkio, line 973:
```suggestion
This hook is called after the test has failed. It is called after `afterEach` hooks since they can influence the test result. It receives a `TaskResult` object with the current test result. This hook is useful for debugging.
```

PR #5142, opened 2024-02-07
feat(vitest): "test" accepts options object as the second parameter

### LINKED ISSUE #5121: Move test and describe configuration options to be a second parameter

Currently, Vitest exposes `test` and `describe` options as the last parameter:

```ts
test('renders component', () => {
  //
}, { retries: 5 })
```

This makes it hard to see at first glance what options are used, especially if the body of a test is too long. Also, some code formatters like prettier can format this code in a weird way like so:

```js
test(
  'renders component',
  () => {}
  { retries: 5 }
)
```

This happens because prettier is hardcoded to treat `test` functions differently when the second parameter is an options object. In this case it is formatted like this:

```ts
test('renders component', { retries: 5 }, () => {
  //
})
```

### Suggested solution

I propose:

- Deprecate passing down an options object in the last argument (show a console warning)
- Support passing down an options object in the second argument
- Still support passing down a number for timeout as the last argument, but also expose `timeout` option in the second argument

In the next major we can remove support for options as the last parameter.

### Alternative

_No response_

### Additional context

I think this can also be a start of using `describe('name', { #flag: true }` instead of `describe.{flag}`. Specifically for situations like this: https://github.com/vitest-dev/vitest/issues/3922

### Validations

sheremet-va:
### Description

Fixes #5121

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -68,7 +68,6 @@ test('skipped test', { skip: true }, () => {
 
 ## test
 
-- **Type:** `(name: string | Function, fn: TestFunction, timeout?: number | TestOptions) => void`
 - **Alias:** `it`
 
 `test` defines a set of related expectations. It receives the test name and a function that holds the expectations to test.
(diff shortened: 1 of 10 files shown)


hi-ogawa (commented):
How about jest migration docs? It looks like jest only has `timeout: number` https://jestjs.io/docs/api, so it would be like `test("name", () => {}, 3000)  -->  test("name", { timeout: 3000 }, () => {}`. Maybe we can add it later though when removing the support in v2.

I was also wondering whether `beforeEach(() => {}, timeout)` can also switch argument positions, but I guess that's not done by anyone (for example https://nodejs.org/api/test.html#beforeeachfn-options), so it's not in the consideration?

sheremet-va:
> How about jest migration docs?

We still support this syntax without depreciation

PR #5156, opened 2024-02-08
fix(vitest): disable optimizer by default until it's stable

sheremet-va:
### Description

The optimizer is not stable enough at this point (some packages like `react` don't work with it and if all dependencies are not pre-bundled, there is a high chance users will get a "Cannot find module" error only in CI which is hard to debug). Since Vitest 1.3.0 focuses more on stability, we are disabling this option until we can figure out a good way to enable it without these hazards. If your tests are slower after this PR, you can always enable optimizer manually:

```ts
export default defineConfig({
  test: {
    deps: {
      optimizer: {
        web: {
           enabled: true,
        },
        ssr: {
          enabled: true,
        }
      }
    }
  }
})
```

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -230,7 +230,7 @@ You will not be able to edit your `node_modules` code for debugging, since the c
 #### deps.optimizer.{mode}.enabled
 
 - **Type:** `boolean`
-- **Default:** `true` if using >= Vite 4.3.2, `false` otherwise
+- **Default:** `false` since Vitest 1.3.0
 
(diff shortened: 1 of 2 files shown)


hi-ogawa (approved):
Sounds good to me. Considering the issue with `ssr.optimizeDeps` https://github.com/vitest-dev/vitest/issues/4910 is fixed since Vite 5.1 https://github.com/vitejs/vite/pull/15561, I think this affects jsdom/happy-dom use cases mostly.

PR #5168, opened 2024-02-09
ci: add publish workflow

sheremet-va:

### Tests

### Documentation

### Changesets


.github/workflows/publish.yml
```diff
@@ -0,0 +1,48 @@
+name: Public Package
```
patak-dev, line 1:
```suggestion
name: Publish Package
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,48 @@
...
+
+permissions:
+  contents: read
+  id-token: write
+
+jobs:
+  publish:
+    if: github.repository == 'vitest-dev/vitest'
+    runs-on: ubuntu-latest
+    environment: Release
```
sheremet-va, line 16: @patak-dev do you know if it will still have access to `GITHUB_TOKEN` if environment is custom? Does it have access only to the environment secrets or also to organization ones?

scripts/publish-ci.ts
```diff
@@ -0,0 +1,35 @@
...
+  : version.includes('alpha')
+    ? 'alpha'
+    : undefined
+
+console.log('Publishing version', version, 'with tag', releaseTag)
+
+if (releaseTag)
+  await $`pnpm -r publish --access public --tag ${releaseTag}`
+else
+  await $`pnpm -r publish --access public`
```
AriPerkkio, line 35: Does the `-r` now include all packages mentioned in `pnpm-workspace.yaml`?

https://github.com/vitest-dev/vitest/blob/e4e93dfd04dd6d680d5dc1431cd57f4c2c1f210e/pnpm-workspace.yaml#L1-L6

Should we instead add a `--filter` flag here to target only `vitest` and `@vitest/*` packages?

sheremet-va: This is the same command we use right now, why would it change?

scripts/publish-ci.ts
```diff
@@ -0,0 +1,35 @@
...
+const releaseTag = version.includes('beta')
+  ? 'beta'
+  : version.includes('alpha')
+    ? 'alpha'
+    : undefined
+
+console.log('Publishing version', version, 'with tag', releaseTag)
+
+if (releaseTag)
+  await $`pnpm -r publish --access public --tag ${releaseTag}`
```
AriPerkkio, line 33: How about the `--provenance` flag? Looks like it could be added to `.npmrc` directly:

https://github.com/pnpm/pnpm/issues/6435#issuecomment-1518397267

sheremet-va: Oh, yes, right! I forgot about it

.github/workflows/publish.yml
sheremet-va: GITHUB_ env should be available

AriPerkkio (approved):

PR #5171, opened 2024-02-09
feat: deprecate watchExclude

patak-dev:
### Description

We had to disable fs cached checks when custom `server.watch.ignored` patterns are present because Vitest has `watchExclude` with a default of `**/dist/**` (not using `outDir` from the project but harcoded `dist`):
- https://github.com/vitejs/vite/pull/15828

This gets in the way of some e2e SSR setups in Vite core and the main vite plugins (react, vue, svelte). My take is that these setups shouldn't have worked and we should enable fs cached checks all the time at one point. But it was a bit too breaking for Vite 5.1

We should align `watchExclude` with Vite core `server.watch.ignored`. We already started to ignore `outDir` for example, and `node_modules` should already be ignored before.

This PR deprecates the option for now, so later on it can be removed. Users are instructed to use `server.watch.ignored`.

We were wondering why this option was added in the first place and maybe the reason is DX of the options as `server.watch.ignored` may not be as clear as `watchExclude`. I think maybe it will depend on how much this feature is used. If it isn't extremely common, I think aligning with Vite is better here. If not, we would need to device a way to change the default to be `undefined` so the regular `ignored` values in Vite core kicks in.


sheremet-va:
I think it's not common enough to justify its existence. Since it's just an alias for `server.watch.ignored`, I think it's safe to remove it in the next major after the depreciation period.

sheremet-va (dismissed):

sheremet-va:
Lint is failing 👀

patak-dev:
Running `pnpm run lint` locally gives me nothing. No idea why it is complaining.

sheremet-va:
> Running `pnpm run lint` locally gives me nothing. No idea why it is complaining.

You probably have a space at the end of the line that you added

Issue #5161, opened 2024-02-09
Code coverage: cannot cover type of express request


wataruian:
### Describe the bug

I have a simple `nestjs` controller that I am testing that accepts `express` - `request` as its param.

I achieved almost 100% code coverage except:

<img width="627" alt="Screenshot 2024-02-09 at 5 07 10 PM" src="https://github.com/vitest-dev/vitest/assets/32057659/f4faa8a5-04e9-4e19-9b64-34d6df2ad6c2">

I cannot cover the type of `Request` from code coverage. I am using `vitest-mock-express` to mock the request object passed to the controller. I also tried `node-mocks-http` and had the same result.

`vitest-mock-express`:
```
import { getMockReq } from 'vitest-mock-express';
const request = getMockReq();
```

`node-mocks-http`:
```
import { createRequest } from 'node-mocks-http';
const request = createRequest();
```

The following test cases should cover all of the code:

```
describe('/', () => {
      it('should succeed', async () => {
        const response = await controller.index();
        expect(response).toStrictEqual({ status: 'ok' });
      });
});

describe('/csrf', () => {
      it('should succeed', async () => {
        const { res } = getMockRes();
        request.res = res;
        const response = await controller.csrf(request as Request);
        expect(response).toStrictEqual({ success: true });
      });

      it('should fail', async () => {
        const response = await controller.csrf(request as Request);
        expect(response).toStrictEqual({ success: false });
      });
});
```

Coverage result:
<img width="866" alt="Screenshot 2024-02-09 at 5 15 29 PM" src="https://github.com/vitest-dev/vitest/assets/32057659/2c846afe-7a87-44d3-83b7-de4c9569d176">

I have a repo for reproduction at: https://github.com/wataruian/vitest-nestjs-repro

### Reproduction

https://github.com/wataruian/vitest-nestjs-repro

### System Info

```shell
System:
    OS: macOS 13.5.2
    CPU: (10) arm64 Apple M2 Pro
    Memory: 84.00 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.15.0 - ~/.asdf/installs/nodejs/18.15.0/bin/node
    Yarn: 1.22.21 - ~/.asdf/installs/nodejs/18.15.0/bin/yarn
    npm: 9.5.0 - ~/.asdf/plugins/nodejs/shims/npm
[...]

PR #5206, opened 2024-02-14
fix(coverage): ignore generated TS decorators

AriPerkkio:
### Description

- Fixes https://github.com/vitest-dev/vitest/issues/5161

This is still quite experimental. `ts-jest` does similar hacks https://github.com/kulshekhar/ts-jest/pull/488.

Before on left, after on right:

image

### Tests

### Documentation

### Changesets


AriPerkkio:
I've also been thinking about exposing an option that users could use to apply their own exclusions. A callback that would be called with generated code, and users could overwrite sources they want. But using that would require deeper understanding of code transpiling and source maps so developer experience wouldn't be good. 🤔 

Maybe it's safer to have these exclusions built-in. And try to keep the patterns at minimum.

PR #5221, opened 2024-02-16
fix: Prevent merging of `poolOptions`

penalosa:
### Description
Arrays in poolOptions are merged with themselves, resulting in duplicating the entries. This PR fixes that, by ensuring `poolOptions` is not present in the config being merged. Bug repro at https://github.com/penalosa/vitest-custom-pool-repro (just clone and run ./test.sh to see the contents of the array option being duplicated), 

See https://github.com/solidjs/vite-plugin-solid/pull/101 for a similar fix

- [x] Discussed on Discord
- [x] Tests TBD
- [x] No lock file changes

### Tests

### Documentation
- [x] No new functionality, so no docs needed

### Changesets
- [x] TBD

### DIFF
--- a/packages/vitest/src/node/plugins/index.ts
+++ b/packages/vitest/src/node/plugins/index.ts
@@ -50,13 +50,6 @@ export async function VitestPlugin(options: UserConfig = {}, ctx = new Vitest('t
         )
         testConfig.api = resolveApiServerConfig(testConfig)
 
-        testConfig.poolOptions ??= {}
-        testConfig.poolOptions.threads ??= {}
-        testConfig.poolOptions.forks ??= {}
-        // prefer --poolOptions.{name}.isolate CLI arguments over --isolate, but still respect config value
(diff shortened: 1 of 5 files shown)


sheremet-va:
Can you add a test case in test/config?

penalosa:
> Can you add a test case in test/config?

I added one in `test/run`, but I can add one in `config` as well

sheremet-va:
> > Can you add a test case in test/config?
> 
> I added one in `test/run`, but I can add one in `config` as well

I wrote the comment when there were no tests. I suggest adding a test somewhere here: https://github.com/vitest-dev/vitest/blob/main/test/config/test/resolution.test.ts

sheremet-va:
Looks like this PR breaks other config options: https://github.com/vitest-dev/vitest/actions/runs/7932413357/job/21658846975?pr=5221

sheremet-va (approved):

PR #5250, opened 2024-02-20
fix(workspace): throw error when browser mode and `@vitest/coverage-v8` are used

AriPerkkio:
### Description

This error is not thrown when browser mode is used with `@vitest/coverage-v8`. Browser tests start and show errors related `node:inspector` usage outside Node.

https://github.com/vitest-dev/vitest/blob/4ff7159049e4dd63501cdf512aea3366b55691cf/packages/vitest/src/node/config.ts#L142-L143

The reason is that resolved configuration is for root configuration, not the workspace configuration. Coverage options are in root config, browser config in workspace config. At this scope we don't have access to those both I think. 

```ts
// vitest.config.ts

export default defineConfig({
  test: {
    coverage: {
      enabled: true,
    },
  },
});

// vitest.workspace.ts
import { defineWorkspace } from "vitest/config";

  {
    test: {
      name: "Browser project",
      browser: {
        headless: true,
        enabled: true,
        name: "chrome",
      },
    },
  },
]);
```

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/workspace.ts
+++ b/packages/vitest/src/node/workspace.ts
@@ -297,7 +297,15 @@ export class WorkspaceProject {
   }
 
   async setServer(options: UserConfig, server: ViteDevServer) {
-    this.config = resolveConfig(this.ctx.mode, options, server.config)
+    this.config = resolveConfig(
+      this.ctx.mode,
+      {
+        ...options,
+        coverage: this.ctx.config.coverage,
+      },
+      server.config,
+    )
+
     this.server = server
 
     this.vitenode = new ViteNodeServer(server, this.config.server)
@@ -329,7 +337,6 @@ export class WorkspaceProject {
 
     return deepMerge({
       ...this.config,
-      coverage: this.ctx.config.coverage,
 
       poolOptions: {
         forks: {
--- a/test/config/fixtures/workspace/browser/workspace-with-browser.ts
+++ b/test/config/fixtures/workspace/browser/workspace-with-browser.ts
@@ -1,0 +1,13 @@
+import { defineWorkspace } from "vitest/config";
+
+export default defineWorkspace([
+  {
+    test: {
+      name: "Browser project",
+      browser: {
+         enabled: true,
+         name: 'chrome'
+      },
+    }
+  }
+])
--- a/test/config/test/failures.test.ts
+++ b/test/config/test/failures.test.ts
@@ -61,6 +61,12 @@ test('v8 coverage provider cannot be used with browser', async () => {
   expect(stderr).toMatch('Error: @vitest/coverage-v8 does not work with --browser. Use @vitest/coverage-istanbul instead')
 })
 
+test('v8 coverage provider cannot be used with browser in workspace', async () => {
+  const { stderr } = await runVitest({ coverage: { enabled: true }, workspace: './fixtures/workspace/browser/workspace-with-browser.ts' })
+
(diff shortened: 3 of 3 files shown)


packages/vitest/src/node/workspace.ts
```diff
@@ -323,10 +323,14 @@ export class WorkspaceProject {
   getSerializableConfig() {
     const optimizer = this.config.deps?.optimizer
     const poolOptions = this.config.poolOptions
+    const coverage = this.ctx.config.coverage
 
     // Resolve from server.config to avoid comparing against default value
     const isolate = this.server?.config?.test?.isolate
 
+    if (coverage.provider === 'v8' && coverage.enabled)
+      throw new Error('@vitest/coverage-v8 does not work with --browser. Use @vitest/coverage-istanbul instead')
```
AriPerkkio, line 332: @sheremet-va can you think of any other place where this check could be done? Maybe a bit earlier than this one.

hi-ogawa (commented):
Thanks for investigation. I see why `resolveConfig` doesn't throw for workspace browser and your approach looks good to me.

As an alternative, I'm wondering if we can move up `getSerializableConfig`'s root config merging:

https://github.com/vitest-dev/vitest/blob/366a78b39aeb0173eee47c95187527fa5130b175/packages/vitest/src/node/workspace.ts#L330-L332

to before `resolveConfig`:

https://github.com/vitest-dev/vitest/blob/366a78b39aeb0173eee47c95187527fa5130b175/packages/vitest/src/node/workspace.ts#L299-L300

Something like this?

```tsx
this.config = resolveConfig(
  this.ctx.mode,
  {
    ...options,
    coverage: this.ctx.config.coverage,
  },
  server.config
);
```

I feel this is somewhat awkward, so I like doing checks during `resolveWorkspace` though.

AriPerkkio:
Merging coverage options as early as in `setServer`'s `resolveConfig` seems like the best option. Now we don't need to duplicate the logic of "browser with v8-coverage" checks.

The other configuration modifications done in `getSerializableConfig` are quite tricky as they may depend on configuration resolved in `resolveConfig`. Ideally the `poolOptions` should also be moved to `setServer()`'s scope but that doesn't seem possible now. I'll leave those as they were.

hi-ogawa (approved):
I see, throwing during `WorkspaceProject.setServer` propagates the error mostly same way, then that's good :+1: 

I didn't think about this approach when I was fixing workspace snapshot option https://github.com/vitest-dev/vitest/pull/5199, but probably I could've done something similar.

packages/vitest/src/node/workspace.ts
```diff
@@ -297,7 +297,15 @@ export class WorkspaceProject {
   }
 
   async setServer(options: UserConfig, server: ViteDevServer) {
-    this.config = resolveConfig(this.ctx.mode, options, server.config)
+    this.config = resolveConfig(
```
sheremet-va, line 300: Maybe we should add a warning here if `options.coverage` is set?

AriPerkkio: Would that come from workspace project's configuration? I don't think we have any runtime warnings for non-project options.

sheremet-va: Hm, yes, I remember. We don't have warnings because people can merge the main config with the project config.

AriPerkkio: Silently ignoring the unsupported options seems like best option for that.

PR #5260, opened 2024-02-21
chore: rename needs reproduction label

patak-dev:
### Description

We did this change in the Vite repo too. All other labels are "needs something".

After merging this PR, the label needs to be renamed.

We can also add `needs test` and `needs documentation` that are useful in the Vite repo at least.


hi-ogawa (approved):

patak-dev:
the label has been renamed to `needs reproduction` now

PR #5268, opened 2024-02-22
Suggest timeout in the second arg, not last

xuhdev:
Timeout specified in the last argument is deprecated.

The [doc](https://vitest.dev/api/#test-api-reference) says:

> Vitest 1.3.0 deprecates the use of options as the last parameter.


sheremet-va:
Timeout is the last argument, the options are the second argument.

xuhdev:
@sheremet-va Would you mind taking another look :)

sheremet-va:
> @sheremet-va Would you mind taking another look :)

What I am trying to say is that the use of a number as the last argument was not deprecated—only the use of an object.

Issue #5273, opened 2024-02-22
Undefined `define` constants error when `--pool=forks`


IanVS:
### Describe the bug

I attempted to set `--pool='forks'` to debug some hanging tests that I've been having intermittently, and started to get a few test failures about global constants not being defined.  I'm guessing it's a consequence of running the tests using `child_process`, but it surprised me a bit nonetheless.  

I'm not sure if anything can be done to address this from vitest, but it might be worth at the very least adding a note in the documentation to clarify that `define` values should not be `undefined`?

### Reproduction

https://github.com/IanVS/reproduction-vitest-define-pool-forks

### System Info

```shell
System:
    OS: macOS 14.0
    CPU: (12) arm64 Apple M2 Max
    Memory: 1.81 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.16.0 - ~/Library/Caches/fnm_multishells/35181_1708624240853/bin/node
    Yarn: 1.22.19 - ~/Library/Caches/fnm_multishells/35181_1708624240853/bin/yarn
    npm: 9.5.1 - ~/Library/Caches/fnm_multishells/35181_1708624240853/bin/npm
    pnpm: 8.8.0 - ~/Library/Caches/fnm_multishells/35181_1708624240853/bin/pnpm
  Browsers:
    Safari: 17.0
  npmPackages:
    vite: ^5.1.4 => 5.1.4
    vitest: ^1.3.1 => 1.3.1
```

### Used Package Manager

pnpm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

AriPerkkio:
[...]

PR #5284, opened 2024-02-23
fix(forks): wrap `defines` to support `undefined` values

AriPerkkio:
### Description

- Fixes https://github.com/vitest-dev/vitest/issues/5273

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/pools/forks.ts
+++ b/packages/vitest/src/node/pools/forks.ts
@@ -8,6 +8,7 @@ import type { ContextRPC, ContextTestEnvironment, ResolvedConfig, RunnerRPC, Run
 import type { PoolProcessOptions, ProcessPool, RunWithFiles } from '../pool'
 import type { WorkspaceProject } from '../workspace'
 import { envsOrder, groupFilesByEnv } from '../../utils/test-helpers'
+import { wrapSerializableConfig } from '../../utils/config-helpers'
 import { groupBy, resolve } from '../../utils'
 import { createMethodsRPC } from './rpc'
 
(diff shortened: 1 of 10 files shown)


packages/vitest/src/node/pools/forks.ts
```diff
@@ -9,6 +9,7 @@ import type { PoolProcessOptions, ProcessPool, RunWithFiles } from '../pool'
 import type { WorkspaceProject } from '../workspace'
 import { envsOrder, groupFilesByEnv } from '../../utils/test-helpers'
 import { groupBy, resolve } from '../../utils'
+import { wrapForksConfig } from '../../runtime/workers/utils'
```
AriPerkkio, line 12: It might be bad to import from `runtime/workers` on the main thread's side. Any better place for this?

AriPerkkio: `utils/config-helpers.ts` looks better.

packages/vitest/src/runtime/workers/utils.ts
```diff
@@ -27,16 +25,3 @@ export function createForksRpcOptions(nodeV8: typeof import('v8')): WorkerRpcOpt
     },
   }
 }
-
-function parsePossibleRegexp(str: string | RegExp) {
```
sheremet-va, line 31: I would be cautious to mix thread level utils and workers utils, rollup can bundle them together one time and we won't notice

AriPerkkio: Yep I agree. I'll refactor the d7c319e4af94cac7a1361523c989aa6f0b13476f once more.

PR #5304, opened 2024-02-27
chore(deps): update `istanbul-lib-source-maps` to v5

AriPerkkio:
### Description

- Ref. https://github.com/istanbuljs/istanbuljs/pull/743
- https://github.com/istanbuljs/istanbuljs/releases

Drops transitive dependency on `source-map` by replacing it with `@jridgewell/trace-mapping`:

> [trace-mapping](https://github.com/jridgewell/trace-mapping) is faster, smaller, and lighter than the `source-map` package, and doesn't require WASM, manual memory management, or async/await to use.
>
> https://github.com/istanbuljs/v8-to-istanbul/pull/186#issue-1209076763

### Tests

### Documentation

### Changesets

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -976,8 +976,8 @@ importers:
         specifier: ^3.0.1
         version: 3.0.1
       istanbul-lib-source-maps:
-        specifier: ^4.0.1
-        version: 4.0.1
+        specifier: ^5.0.4
+        version: 5.0.4
       istanbul-reports:
         specifier: ^3.1.6
         version: 3.1.6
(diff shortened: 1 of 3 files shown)


pnpm-lock.yaml
```diff
@@ -18829,6 +18841,18 @@ packages:
       source-map: 0.6.1
     transitivePeerDependencies:
       - supports-color
+    dev: true
+
+  /istanbul-lib-source-maps@5.0.4:
```
sheremet-va, line 18846: Did you replace it everywhere? The latest version is not removed

AriPerkkio: The remaining v4's are `dev: true` only. We have dependency on `jest` which brings in the older version.

Issue #3605, opened 2023-06-17
When using `coverage.all` typescript files with no executable code are colored red in the coverage report


HristoKolev:
### Describe the bug

I have a typescript file that only contains typescript interfaces. When I run vitest with coverage I get it reported as red (uncovered) even when the interfaces are referenced by other files that are covered by tests.

![image](https://github.com/vitest-dev/vitest/assets/13748164/95b0c27f-6c11-45c6-8712-8b76ad41daea)

The way jest handles this situation is to not count these file as neither covered or uncovered but just empty and colored in gray:

![image](https://github.com/vitest-dev/vitest/assets/13748164/317036ec-e4a8-40da-a14e-43f9cf9912d4)

### Reproduction

Here is the github repo from the images:

https://github.com/HristoKolev/vite-workshop/tree/6549514d7940d642a51aebcfd049466c3cc25e7b

to run it:

```sh
cd vite-app

npm i

npm run test
```

### System Info

```shell
System:
    OS: Windows 10 10.0.22621
    CPU: (32) x64 AMD Ryzen 9 7950X 16-Core Processor
    Memory: 14.66 GB / 31.16 GB
  Binaries:
    Node: 18.12.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - C:\Program Files\nodejs\yarn.CMD
    npm: 9.2.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22621.1778.0), Chromium (114.0.1823.51)
    Internet Explorer: 11.0.22621.1
  npmPackages:
    @vitejs/plugin-react: ^4.0.0 => 4.0.0
    @vitest/coverage-v8: ^0.32.2 => 0.32.2
    vite: ^4.3.2 => 4.3.9
    vitest: ^0.32.0 => 0.32.2
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
[...]

PR #5328, opened 2024-03-03
fix(coverage): v8 to ignore type-only files

AriPerkkio:
### Description

- Fixes https://github.com/vitest-dev/vitest/issues/3605
- The existing test cases that validate `coverage.json` cover this feature. If the fix is removed, the new `types.ts` shows up in snapshot and it fails.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/coverage-v8/src/provider.ts
+++ b/packages/coverage-v8/src/provider.ts
@@ -14,6 +14,7 @@ import remapping from '@ampproject/remapping'
 import { normalize, resolve } from 'pathe'
 import c from 'picocolors'
 import { provider } from 'std-env'
+import { stripLiteral } from 'strip-literal'
 import createDebug from 'debug'
 import { cleanUrl } from 'vite-node/utils'
 import type { EncodedSourceMap, FetchResult } from 'vite-node'
@@ -260,7 +261,13 @@ export class V8CoverageProvider extends BaseCoverageProvider implements Coverage
       }
 
       const coverages = await Promise.all(chunk.map(async (filename) => {
-        const { source } = await this.getSources(filename.href, transformResults)
+        const transformResult = await this.ctx.vitenode.transformRequest(filename.pathname).catch(() => {})
+
+        // Ignore empty files, e.g. files that contain only typescript types and no runtime code
+        if (transformResult && stripLiteral(transformResult.code).trim() === '')
+          return null
+
+        const { originalSource } = await this.getSources(filename.href, transformResults)
 
         const coverage = {
           url: filename.href,
@@ -269,7 +276,7 @@ export class V8CoverageProvider extends BaseCoverageProvider implements Coverage
           functions: [{
             ranges: [{
               startOffset: 0,
-              endOffset: source.length,
+              endOffset: originalSource.length,
               count: 0,
             }],
             isBlockCoverage: true,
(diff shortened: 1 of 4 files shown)


AriPerkkio:
Nuxt's failure looks like real issue. It's passing fine on `main`. 🤔

packages/coverage-v8/src/provider.ts
```diff
@@ -260,7 +261,13 @@ export class V8CoverageProvider extends BaseCoverageProvider implements Coverage
       }
 
       const coverages = await Promise.all(chunk.map(async (filename) => {
-        const { source } = await this.getSources(filename.href, transformResults)
+        const transformResult = await this.ctx.vitenode.transformRequest(filename.pathname)
+
+        // Ignore empty files, e.g. files that contain only typescript types and no runtime code
+        if (transformResult && await isFileEmpty(transformResult.code))
```
sheremet-va, line 267: Why do we really need to parse `transformResult` AST? It's always JavaScript, so it will never have type declarations 🤔

AriPerkkio: I guess we could make this decision based on `mappings` as well. Empty mappings would indicate that the file contains no runtime code.

```json
{
  "code": "\n\n//# sourceMappingSource=vite-node\n//# sourceMappingURL=data:application/json;charset=utf-8;base64,eyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpbInR5cGVzLnRzIl0sInNvdXJjZXNDb250ZW50IjpbImV4cG9ydCB0eXBlIEZpcnN0ID0gXCJpZ25vcmVcIiB8IFwidGhpc1wiIHwgXCJmaWxlXCI7XG5cbmV4cG9ydCB0eXBlIFNlY29uZCA9IFwiaWdub3JlXCIgfCBcInRoaXNcIiB8IFwiZmlsZVwiO1xuIl0sIm5hbWVzIjpbXSwibWFwcGluZ3MiOiIifQ==\n",
  "map": {
    "version": 3,
    "sources": [
      "types.ts"
    ],
    "sourcesContent": [
      "export type First = \"ignore\" | \"this\" | \"file\";\n\nexport type Second = \"ignore\" | \"this\" | \"file\";\n"
    ],
    "names": [],
    "mappings": ""
  },
  "deps": [],
  "dynamicDeps": []
}
```

sheremet-va: Can we trim the code an check that it doesn't start with `//# sourceMappingSource`?

AriPerkkio: It would work for most cases but there can be edge cases that won't fall into that check, e.g.:

```ts
// Some comment here @preserve
export type First = "ignore" | "this" | "file";
```

This will transform to 
```js
// Some comment here @preserve

...
```

At the moment I can't think of other such cases but I guess there could be. AST based check would always be accurate. I'll change this to `transformResult?.code.trim().startsWith('//# sourceMappingSource=')` - it should be accurate enough.

sheremet-va: Maybe we can use something like https://www.npmjs.com/package/strip-literal then?

AriPerkkio: That looks good. We could use `stripLiteral(transformResult.code).trim() === ''` check here.

sheremet-va (approved):

JoshuaKGoldberg:
Confirmed this fixed a whole bunch of `types.ts` files in my repositories, thank you very much @AriPerkkio & @sheremet-va! ❤️

PR #5342, opened 2024-03-05
feat: add a flag to include test location in tasks

sheremet-va:
### Description

This PR adds a config flag `includeTaskLocations` that adds `location` property to every collected task.

TODO:

- [ ] Tests
- [ ] Docs

### Tests

### Documentation

### Changesets


AriPerkkio (approved):
Looks ok. It's good that this is opt-in behind a flag as it could indeed cause performance issues in larger sets.

AriPerkkio:
One thing though: If this is something that only reporters need, we might not need `vitest.config.ts` option. It could be an option in core/`ctx` that custom reporters could toggle in their constructor. Also the built-in reporters could accept it as reporter option, e.g. `['json', { includeTaskLocations: true }]`.

sheremet-va:
> One thing though: If this is something that only reporters need, we might not need `vitest.config.ts` option. It could be a property in core/`ctx` that custom reporters could toggle in their constructor. Also the built-in reporters could accept it as reporter option, e.g. `['json', { includeTaskLocations: true }]`.

The reporter is just one example where it can be used. You can also get this data from `vitest.state.fileMap`

PR #5352, opened 2024-03-07
fix: .mockRestore() should restore original implementation when calling getMockImplementation()

kasperpeulen:
### Description

We do some shady stuff in storybook, trying to automatically showing vitest spy in the storybook action panel:

https://github.com/storybookjs/storybook/blob/1d765656e8d3657baaf1b025c2571be48078171f/code/addons/actions/src/loaders.ts#L20-L29

```ts
  const previous = value.getMockImplementation();
  if (previous?._actionAttached !== true && previous?.isAction !== true) {
    const implementation = (...params: unknown[]) => {
      action(key)(...params);
      return previous?.(...params);
    };
    implementation._actionAttached = true;
    value.mockImplementation(implementation);
  }
```

However, this does not work after a spy is restored. As value.getMockImplementation() won't the original implementation, which I believe it should after restoring (not resetting) a mock. I think this is a bug in @vitest/spy.

There is a workaround though, that I implemented in this PR:
https://github.com/storybookjs/storybook/pull/26364

But it seems weird that the `getMockImplementation()` value is out of sync with what is actually being "implemented" after restoring the spy. So feels like a bug to me, especially given the documentation:

```
  mockReset: () => this
  /**
   * Does what `mockReset` does and restores inner implementation to the original function.
   *
   * Note that restoring mock from `vi.fn()` will set implementation to an empty function that returns `undefined`. Restoring a `vi.fn(impl)` will restore implementation to `impl`.
   */
```

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/spy/src/index.ts
+++ b/packages/spy/src/index.ts
@@ -424,7 +424,7 @@ function enhanceSpy<TArgs extends any[], TReturns>(
   stub.mockRestore = () => {
     stub.mockReset()
     state.restore()
-    implementation = undefined
+    implementation = state.getOriginal()
     return stub
   }
 
--- a/test/core/test/jest-mock.test.ts
+++ b/test/core/test/jest-mock.test.ts
@@ -129,9 +129,10 @@ describe('jest mock compat layer', () => {
 
     spy.mockRestore()
 
-    expect(spy.getMockImplementation()).toBe(undefined)
-
     expect(spy.mock.results).toEqual([])
+
+    expect(spy.getMockImplementation()).toBe(originalFn)
+    expect(spy()).toEqual('original')
   })
 
   it('implementation async fn', async () => {
@@ -257,7 +258,7 @@ describe('jest mock compat layer', () => {
 
     obj.setter = 'last'
 
-    expect(spy.getMockImplementation()).toBe(undefined)
+    expect(spy.getMockImplementation()).toBe(Object.getOwnPropertyDescriptor(obj, 'setter')?.set)
 
     expect(setValue).toBe('last')
   })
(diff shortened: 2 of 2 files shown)


sheremet-va:
Current behavior is correct. Restoring the mock should remove `mockImplementation`. This doesn't contradict the documentation since `undefined` implementation will fallback to the original function when the function is called. With the proposed change, `getMockImlementation` will _always_ return a function which is not how it should work (according to the name).

kasperpeulen:
@sheremet-va I don't follow your reasoning completely.

When I create a mock function with `fn`:
```ts
const mock = vi.fn(() => true)
```

Then initially the values are like this:

```ts
expect(mock()).toBe(true)
expect(mock.getMockImplementation()).toBeDefined()
expect(mock.getMockImplementation()?.()).toBe(true)
```

When I restore the mock, I would expect, the mock implementation to be the same as it was initially, right? But instead it is:

```ts
mock.restore();
expect(mock()).toBe(true)
expect(testFn.getMockImplementation()).toBe(undefined)
```

This seems at least asymmetric? I think I would get your reasoning, if initially, the implementation and the mock implementation would also be different:

```
const mock = vi.fn(() => true)
expect(mock()).toBe(true)
expect(testFn.getMockImplementation()).toBe(undefined)
```

But maybe that is just me.

If I follow your reasoning, would you agree that it would make sense then if we can get a `mock.getImplementation` method in the Mock interface. Because I don't see a way to get the current implementation, if there is no "mock" implementation set.

kasperpeulen:
Anyway, will close this. And `use mock[Symbol.for('tinyspy:spy')].impl` for now to get the implementaiton.

sheremet-va:
> This seems at least asymmetric?

You have a point. But what we should do is the opposite - `restore` in this case should remove the `() => true` implementation. This will also make it jest-compatible. So it will behave like this:

```js
mock.mockRestore();
expect(mock()).toBe(undefined)
```

PR #5355, opened 2024-03-07
fix: `--inspect-brk` to pause before execution

AriPerkkio:
### Description

- Pauses code execution similarly as Node's `--inspect-brk` does
- Dev tools will stop on top of the first test file that is being run

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vite-node/src/source-map.ts
+++ b/packages/vite-node/src/source-map.ts
@@ -48,6 +48,11 @@ export function withInlineSourcemap(result: TransformResult, options: {
   while (OTHER_SOURCE_MAP_REGEXP.test(code))
     code = code.replace(OTHER_SOURCE_MAP_REGEXP, '')
 
+  // If the first line is not present on source maps, add simple 1:1 mapping ([0,0,0,0], [1,0,0,0])
+  // so that debuggers can be set to break on first line
+  if (map.mappings.startsWith(';'))
+    map.mappings = `AAAA,CAAA${map.mappings}`
+
   const sourceMap = Buffer.from(JSON.stringify(map), 'utf-8').toString('base64')
   result.code = `${code.trimEnd()}\n\n${VITE_NODE_SOURCEMAPPING_SOURCE}\n//# ${VITE_NODE_SOURCEMAPPING_URL};base64,${sourceMap}\n`
 
--- a/packages/vitest/src/runtime/inspector.ts
+++ b/packages/vitest/src/runtime/inspector.ts
@@ -1,15 +1,17 @@
 import { createRequire } from 'node:module'
 
-import type { ResolvedConfig } from '../types'
+import type { ContextRPC } from '../types'
 
 const __require = createRequire(import.meta.url)
 let inspector: typeof import('node:inspector')
+let session: InstanceType<typeof inspector.Session>
 
 /**
  * Enables debugging inside `worker_threads` and `child_process`.
  * Should be called as early as possible when worker/process has been set up.
  */
-export function setupInspect(config: ResolvedConfig) {
+export function setupInspect(ctx: ContextRPC) {
+  const config = ctx.config
   const isEnabled = config.inspect || config.inspectBrk
 
   if (isEnabled) {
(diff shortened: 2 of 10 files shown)


packages/vitest/src/runtime/inspector.ts
```diff
@@ -20,8 +20,14 @@ export function setupInspect(config: ResolvedConfig) {
...
       inspector.open()
 
-      if (config.inspectBrk)
+      if (config.inspectBrk) {
+        const session = new inspector.Session()
+        session.connect()
+        session.post('Debugger.enable')
+
         inspector.waitForDebugger()
+        session.post('Debugger.pause')
```
AriPerkkio, line 29: It would be great if we could actually call this on top of the first test file. I'll check later if that's doable easily.

AriPerkkio: `Debugger.setBreakpointByUrl` should work. Below minimal repro:

```js
// entry.mjs
import inspector from "node:inspector";

const session = new inspector.Session();
session.connect();
session.post("Debugger.enable");

inspector.open();
inspector.waitForDebugger();
session.post("Debugger.setBreakpointByUrl", {
  lineNumber: 0,
  url: new URL("./target.mjs", import.meta.url).href,
});

console.log("Importing");
await import("./target.mjs");

console.log("Exit");
```

```js
// target.mjs
console.log("Line #1"); // Dev tools break here automatically
console.log("Line #2");
console.log("Line #3");
```

packages/vitest/src/runtime/inspector.ts
```diff
@@ -20,8 +22,23 @@ export function setupInspect(config: ResolvedConfig) {
...
+
+        // Stop at first test file
+        if (firstTestFile) {
+          session = new inspector.Session()
+          session.connect()
+
+          session.post('Debugger.enable')
+          session.post('Debugger.setBreakpointByUrl', {
+            lineNumber: 1,
+            url: new URL(firstTestFile, import.meta.url).href,
```
AriPerkkio, line 36: This works but the debugger is not following source maps.
image

However when using `debugger` statement in test, dev tools show original source nicely:
image

I'll need to look more into the documentation: https://chromedevtools.github.io/devtools-protocol/v8/Debugger/#method-setBreakpointByUrl

AriPerkkio: Oh right the first line is not on source maps. If breakpoint is set on line 3, then source maps are followed correctly. 
image

This requires a bit similar knowledge of vite-node's wrapper code like V8 code coverage needs. First line contains `'use strict'`, second one the execution wrapper, and then there's the code.

AriPerkkio: Resolved in 945a23f by making sure vite-node's source maps always contain mappings for the first line.

packages/vite-node/src/source-map.ts
```diff
@@ -48,6 +48,11 @@ export function withInlineSourcemap(result: TransformResult, options: {
   while (OTHER_SOURCE_MAP_REGEXP.test(code))
     code = code.replace(OTHER_SOURCE_MAP_REGEXP, '')
 
+  // If the first line is not present on source maps, add simple 1:1 mapping ([0,0,0,0], [1,0,0,0])
+  // so that debuggers can be set to break on first line
+  if (map.mappings.startsWith(';'))
+    map.mappings = `AAAA,CAAA${map.mappings}`
```
sheremet-va, line 54: Don't we already modify source mapping for coverage support somewhere?

AriPerkkio: In V8 provider, yes:

https://github.com/vitest-dev/vitest/blob/6eda473f30c4670ec72a06bd3b083a724712ae2a/packages/coverage-v8/src/provider.ts#L394-L406

sheremet-va: No, I mean don't we fill the map somewhere because of the vite-node offset? How is this PR different? Should we not fill it anymore?

AriPerkkio: Vite-node's offset doesn't directly affect source maps in coverage. We just instruct `v8-to-istanbul` that the collected V8 coverage JSON reports include execution wrapper:

https://github.com/vitest-dev/vitest/blob/6eda473f30c4670ec72a06bd3b083a724712ae2a/packages/coverage-v8/src/provider.ts#L50-L51

https://github.com/vitest-dev/vitest/blob/6eda473f30c4670ec72a06bd3b083a724712ae2a/packages/coverage-v8/src/provider.ts#L362-L365

This PR has nothing to do with code coverage. It's about being able to set debugger statement on top of the file. You should be able to see this in VSCode's debugger as well.

This PR only affects `--inspect-brk` flag. It instructs debuggers to stop in the first test file before executing it. If the breakpoint is set on a line that is not included in source maps users will see transpiled code: https://github.com/vitest-dev/vitest/pull/5355#discussion_r1518540412. If the line is on source maps, users will see their original test file: https://github.com/vitest-dev/vitest/pull/5355#discussion_r1518541771. If Vite's SSR transformed import statements were included in source maps we could set breakpoint right after vite-node's wrapper - like line 3 or something. But as they are not, it's really difficult to try to guess how many import statements Vite added: https://github.com/vitest-dev/vitest/pull/5355#discussion_r1518560031. It's easiest to manually add breakpoint for source maps if there is none.

PR #5362, opened 2024-03-08
chore(runner): exporting a few interfaces from tasks.ts

jpb06:
### Description

Hello!

I would like to extend `it` function like in [here](https://github.com/jpb06/vitest-effect/blob/expect-extend/src/extend/it-effect.extend.ts), to be able to do something like this:

```typescript
import { describe, expect } from 'vitest';
import { it } from '../extend/it-effect.extend';

// just a dumb matcher for the example
const expectErrorWithTag = (tag: string) => Effect.catchAll((e: { _tag: string }) => {
  expect(e._tag).toBe(tag);

  return Effect.succeed(undefined);
});

describe('my cool test suite', () => {
  it.effect('should accept an effect', () => {
    const tag = 'Yolo';
    return pipe(
      Effect.gen(function* (_) {
        yield* _(Effect.fail({ _tag: tag, error: 'Oh no' }));
      }),
      expectErrorWithTag(tag),
    );
  });

  it('should just do classical vitest stuff', () => {
    expect(true).toBe(true);
  });
});

```

Three interfaces are not exported in `packages/runner/src/types/tasks.ts`:
- TestEachFunction
- TestCollectorCallable
- ExtendedAPI

I propose to export these types.

#### Why?

The idea is to simplify testing for [effect](https://effect.website/docs/introduction) lib. 

The change is trivial so I didn't create an issue. Please let me know if there is something I missed.

Thank you for you time 🙇🏻 

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/runner/src/types/tasks.ts
+++ b/packages/runner/src/types/tasks.ts
@@ -133,14 +133,14 @@ interface EachFunctionReturn<T extends any[]> {
   ): void
 }
 
-interface TestEachFunction {
+export interface TestEachFunction {
   <T extends any[] | [any]>(cases: ReadonlyArray<T>): EachFunctionReturn<T>
   <T extends ReadonlyArray<any>>(cases: ReadonlyArray<T>): EachFunctionReturn<ExtractEachCallbackArgs<T>>
   <T>(cases: ReadonlyArray<T>): EachFunctionReturn<T[]>
   (...args: [TemplateStringsArray, ...any]): EachFunctionReturn<any[]>
 }
 
-interface TestCollectorCallable<C = {}> {
+export interface TestCollectorCallable<C = {}> {
   /**
    * @deprecated Use options as the second argument instead
    */
@@ -204,7 +204,7 @@ export interface TestOptions {
   fails?: boolean
 }
 
-interface ExtendedAPI<ExtraContext> {
+export interface ExtendedAPI<ExtraContext> {
   skipIf: (condition: any) => ChainableTestAPI<ExtraContext>
   runIf: (condition: any) => ChainableTestAPI<ExtraContext>
 }


sheremet-va:
You know you can use `createTaskCollector` (https://vitest.dev/advanced/runner.html#your-task-function) for this?

jpb06:
Hello @sheremet-va, thank you.

Not sure creating a task with `createTaskCollector` would answer the use case. What I'd like to do is extending `it` function to only have a single operator for tests.

I also don't see what `createTaskCollector` would bring to the table. There is no easy way to coerce second parameter (`fn`) to be an effect, and doing something like the following seems pretty convoluted; we lose access to CustomAPI anyway :

```typescript
export const itEffect = <A, E>(
  name: string,
  effect: Effect.Effect<A, E, TestServices.TestServices>,
  timeout = 5_000,
) => {
  function inner(
    this: typeof inner,
    name: string,
    fn: (context: TaskContext<Custom<unknown>>) => Awaitable<void>,
    timeout: number,
  ) {
    getCurrentSuite().task(name, {
      ...this,
      meta: {
        customPropertyToDifferentiateTask: true,
      },
      handler: fn,
      timeout,
    });
  }

  const wrappedEffect = () => effect;

  return createTaskCollector(inner)(
    name,
    () =>
      pipe(
        Effect.suspend(wrappedEffect),
        Effect.provide(TestEnv),
        Effect.runPromise,
      ),
    timeout,
  );
};
```

image

Is there something I fail to understand? 

Thanks.

sheremet-va:
I don't think you understood how it works. The collector returns the `it` function, so you example should be something like this:

```ts
const effectTest = createTaskCollector(
  function effectTest(name: string, effect: Effect) {
    getTaskCollector().custom({
      ...this,
      name,
      handler() {
        return pipe(
          Effect.suspend(() => effect),
          Effect.provide(TestEnv),
          Effect.runPromise,
        )
      }
    })
  }
)

// supports all modifiers
effectTest.todo(/* support the same arguments that effectTest defines */)
```

jpb06:
What is `getTaskCollector` in your example? Where does it come from?

If I take back the example from the documentation:
image

`fn` can be anything. I Wan't it to only be an effect.

ie:
```typescript 
myCustomTask("cool", () => {
  // this works but it should not; it's not an effect
})
```

This is beside the point anyway. As I said, a custom task partially answers my need. I would like to extend `it` function. Is that against the philosophy of vitest library?

Perhaps I just have to accept the added boilerplate:
```typescript
it('should run a promise', () =>
  Effect.gen(function* (_) {
    // effect stuff
  }).pipe(
    Effect.runPromise, // transforming the effect to a promise
  ));
```

What do you think? Exporting these 3 interfaces is a bad idea?

Sorry to waste your time like this 😢

sheremet-va:
> What is `getTaskCollector` in your example?

It's just a typo. It should be `getCurrentSuite`

> `fn` can be anything. I Wan't it to only be an effect.

This is a bug with `createTaskCollector` utility types - it will actually call a function that you pass down with the same arguments.

> I would like to extend `it` function. Is that against the philosophy of vitest library?

It is not against the philosophy of Vitest, we expose `createTaskCollector` specifically for this usecase. What is against the philosophy is putting additional methods on `it`. Another things that is against Vitest philosophy is calling tests during the collection phase. `handler` is a function that will be called when a test needs to be called.

jpb06:
Alright thank you for your patience 🙇🏻

PR #5364, opened 2024-03-09
feat: remove unrelated noise from diff for toMatchObject()

### LINKED ISSUE #5353: Only report a subset of the actual and expected for toMatchObject

If you run this test.

```ts
describe('diff test', () => {
  it('should compare 2 objects', () => {
    const a = { a: 1, b: 2, c: { d: 4 } };
    const b = { b: 3 };

    expect(a).toMatchObject(b);
  });
});
```

You get the following output.

Screenshot 2024-03-07 at 12 43 23

It would be clearer if it would strip properties that are not relevant. For example, see `getObjectSubset` in `@jest/expect-utils`.

[getObjectSubset](https://github.com/jestjs/jest/blob/442c7f692e3a92f14a2fb56c1737b26fc663a0ef/packages/expect-utils/src/utils.ts#L116)

After patching the `processError` function (`@vitest/utils`) locally.

```ts
export function processError(err: any, diffOptions?: DiffOptions) {
...
  if (err.showDiff || (err.showDiff === undefined && err.expected !== undefined && err.actual !== undefined)) {
    ...
    const actualSubset = getObjectSubset(replacedActual, replacedExpected);
    err.actual = actualSubset;
    err.diff = diff(replacedExpected, actualSubset, { ...diffOptions, ...err.diffOptions });
    ...
  }
...
}
```

I get this output instead.

Screenshot 2024-03-07 at 12 56 42

Which makes it a lot easier to spot the difference. 

The output with the VSCode plugin looks like this now.

Screenshot 2024-03-07 at 12 22 44

Perhaps it would be useful to also add this into Vitest?

### Suggested solution

Extract a subset of properties from the actual and report that as output.

### Alternative

_No response_

### Additional context

_No response_

### Validations

geersch:
### Description

closes #5353 

Removes unrelated noise from diff for `toMatchObject()`.

### Tests

### Documentation

### Changesets


sheremet-va (approved):

test/core/test/jest-expect.test.ts
```diff
@@ -993,6 +977,92 @@ it('toHaveProperty error diff', () => {
...
+  expect(getError(() => expect({ a: 1, b: 2, c: { d: 4 } }).toMatchObject({ b: 3 }))).toMatchInlineSnapshot(`
+    [
+      "expected { a: 1, b: 2, c: { d: 4 } } to match object { b: 3 }",
+      "- Expected
+    + Received
+
+      Object {
+    -   "b": 3,
+    +   "b": 2,
+      }",
```
sheremet-va, line 993: Can we print something like `... 10 more properties`

geersch: I assume you mean reporting the number of properties that were stripped from the root of `actual`?

The assertion message printed by Chai already mentions that over a certain threshold. Does it have a lof of added value to mention it again?

For example (`...(2)`).

```ts
expect(getError(() => expect({ a: 1, b: 2, c: { d: 4 }, x: 5, y: 6, z: 7 }).toMatchObject({ a: 1, b: 2, c: { d: 5 } }))).toMatchInlineSnapshot(`
    [
      "expected { a: 1, b: 2, c: { d: 4 }, x: 5, …(2) } to match object { a: 1, b: 2, c: { d: 5 } }",
      "- Expected
    + Received

      Object {
        "a": 1,
        "b": 2,
        "c": Object {
    -     "d": 5,
    +     "d": 4,
        },
      }",
    ]
  `)
```

sheremet-va: Yes, but it won't show it in nested objects, would it?

geersch: I tweaked the message to list the number of properties that were stripped from the actual (incl. all the nested) ones. There are some examples in the `jest-expect.ts.ts`.

test/core/test/jest-expect.test.ts
```diff
@@ -903,24 +903,152 @@ it('correctly prints diff with asymmetric matchers', () => {
...
+}
+
+it('toMatchObject error diff', () => {
+  setupColors(getDefaultColors())
+
+  // single property on root (3 total properties, 1 expected)
+  expect(getError(() => expect({ a: 1, b: 2, c: 3 }).toMatchObject({ c: 4 }))).toMatchInlineSnapshot(`
+    [
+      "expected { a: 1, b: 2, c: 3 } to match object { c: 4 }
+    (2 more properties properties in actual)",
```
sheremet-va, line 929: It says "properties properties" 🤔 

Also not sure about a few things:

- Shouldn't it be under the object?
- The text itself is not very descriptive - can we put more information here? Like "2 more properties not shown since they are the same" (or maybe there is a better way to tell it more concisely?)

sheremet-va (changes requested):

test/core/test/jest-expect.test.ts
geersch: Thanks, missed the duplicate "properties" when I introduced the `pluralize` function. 

Changed the message to:

```
1 matching property omitted from actual
5 matching properties omitted from actual
...etc.
```

I think that expresses it more clearly that x matching properties from the actual were removed.

As for placing it under the object, that's a little bit more convoluted. The error is caught and processed (`processError` in `error.ts`) and then rendered later on. Is there a way to influence that behavior to render it beneath?

packages/expect/src/jest-utils.ts
```diff
@@ -537,6 +537,76 @@ export function generateToBeMessage(deepEqualityName: string, expected = '#{this
   return toBeMessage
 }
 
+const IRREGULAR_PLURALS: { [key: string]: string } = {
+  property: 'properties',
+}
+
 export function pluralize(word: string, count: number): string {
-  return `${count} ${word}${count === 1 ? '' : 's'}`
+  return count === 1 ? word : IRREGULAR_PLURALS[word] ?? `${word}s`
```
sheremet-va, line 545: Can we not change this function?

1. This is a breaking change
2. You can't even use it without adding "irregular" plurals for a single use case where you can just have a single `?` when you call it

sheremet-va (changes requested):

packages/expect/src/jest-utils.ts
geersch: Alright, agreed. Reverted the change to `pluralize` and just used a ternary.

sheremet-va (approved):

PR #5395, opened 2024-03-17
feat(vitest): support `--detectAsyncLeaks` cli flag

### LINKED ISSUE #3119: Detect hanging async operations

As a developer using Vitest I want to be able to detect if an async operation that was started during a test case hasn't completed at the end of that test so that I can catch hanging promise bugs. I am willing to contribute a PR if there is interest.

### Suggested solution

We could provide a `--detect-async-leaks` flag to turn on this type of detection. [Node.js's builtin `async_hooks` module](https://nodejs.org/api/async_hooks.html#printing-in-asynchook-callbacks) provides the necessary functionality

### Alternative

Alternatively, we could make this the default and provide an option to turn it off.

### Additional context

For example in Deno the following test fails with the message "Test case is leaking async ops":

```js
Deno.test("hanging interval", () => {
  setInterval(() => {}, 1000);
});
```

This helped me catch a few particularly hard to find bugs in a streaming form data parser I was working on. Web streams are particularly prone to this type of error but there are many other cases where such detection would be useful.

### Validations

fenghan34:
### Description

 - Close #3119 

This PR is still work in progress, not sure if it's the right solution, any feedback would be appreciated. 

### Tests

### Documentation

### Changesets


packages/vitest/src/node/logger.ts
```diff
@@ -212,4 +213,22 @@ export class Logger {
     }))
     this.log(c.red(divider()))
   }
+
+  async printAsyncLeaksWarning(hangingOps: HangingOps[]) {
+    if (hangingOps.length === 0)
+      return
+    const errorMessage = c.yellow(c.bold(
+      `\nVitest has detected the following ${hangingOps.length} hanging operation${hangingOps.length > 1 ? 's' : ''} potentially keeping Vitest from exiting: \n`,
```
AriPerkkio, line 221: > <ins>potentially</ins> keeping Vitest from exiting

This is well phrased. 👍 

None of these patterns (timers, promises) should prevent Vitest from exiting as we don't do graceful exits - or have you so far noticed any hangs? But this will still be useful for finding leaks from tests and source code.

fenghan34: Thanks for the feedback! Actually this phrase is ported from Jest and I thought it may can be used in Vitest as well. 

I haven't found explicit hangings in my daily use, but yeah I think people may benefit from this feature and am willing to be involved.

PR #5435, opened 2024-03-26
feat!: remove the empty suite from the runner

sheremet-va:
### Description

This PR removes the default empty suite that was reported with all tasks and led to a few confusing bugs:

- It always existed even if the test was not inside any suites
  - Because of this test names had an empty string at the start
  - Custom reporters had to check for the id/name of the internal suite to correctly report results
- All tasks now have a `file` property they can fallback to if they need the file information 
- `expect.getState().testPath` is always defined now
- `expect.getState().currentTestName` doesn't have the file anymore

Fixes [#4045](https://github.com/vitest-dev/vitest/issues/4045)

### Tests

### Documentation

### Changesets


packages/runner/src/collect.ts
```diff
@@ -41,24 +43,26 @@ export async function collectTests(paths: string[], runner: VitestRunner): Promi
...
         else if (c.type === 'collector') {
           const suite = await c.collect(file)
-          if (suite.name || suite.tasks.length)
+          if (suite.name || suite.tasks.length) {
+            mergeHooks(fileHooks, getHooks(suite))
             file.tasks.push(suite)
+          }
+        }
+        else {
+          c satisfies never
```
AriPerkkio, line 61: Is this intentional? What's it for?

sheremet-va: Make sure all options are exhausted - if we add another type, this will throw a type error

AriPerkkio: Oh right, that's nice trick. 👍 

It might be worth to add describing comment above this. Otherwise it looks like this line isn't really doing anything.

PR #5459, opened 2024-03-29
chore: improve PR template

patak-dev:
### Description

See https://github.com/vitejs/vite/pull/16266 for reference.

For now I let the text in the previous checklist without much changes. I think it is too long, and maybe it would be better to remove some points or summarize them to focus on the most important things to check before sending the PR. The first point for example could be eliminated IMO.


pnpm-lock.yaml
```diff
@@ -28099,4 +28099,4 @@ packages:
       '@vitest/test-dep1': '*'
     dependencies:
       '@vitest/test-dep1': file:test/vite-node/deps/dep1
-    dev: true
+    dev: true
```
patak-dev, line 28102:
```suggestion
    dev: true

```

sheremet-va:
I prefer our current one. It was added not so long ago, and I like that it forces new contributors to manually check checkboxes. Personally, I wouldn't even read the template if it was a comment.

PR #5503, opened 2024-04-07
chore: provide explicit type param to one of the calls

Andarist:
This is not needed today but you will have to workaround an improved inference for type predicated in the upcoming TS 5.5. See the failure reported [here](https://github.com/microsoft/TypeScript/issues/58104#issuecomment-2041579302) and the discussion~ [here](https://github.com/microsoft/TypeScript/pull/57952#issuecomment-2023641553)


sheremet-va:
Thank you!

sheremet-va (approved):

PR #5544, opened 2024-04-15
fix: close vite servers on all resolved projects

surc54:
### Description

I've been running into an issue where Vitest would hang after the testing is completed with the following output:

```
close timed out after 10000ms
Tests closed successfully but something prevents the main process from exiting
You can try to identify the cause by enabling "hanging-process" reporter. See https://vitest.dev/config/#reporters
```

Luckily, this repo has a reproducible setup at `test/cli/fixtures/project` -- in this directory, running `npx vitest run --project space_1` causes the error above (you may need to try multiple times, but I've been getting it consistently).

The only other report I've found is by this commenter https://github.com/vitest-dev/vitest/issues/2008#issuecomment-1933009750 (the issue itself seems to be about a different time out issue).

I've tracked this down to a Vite web-socket server not being cleaned up at the end of the Vitest session.

This PR edits the `Vitest#close()` method to close the Vite dev servers on ALL resolved projects instead of just the filtered projects.

I tried to write a new test under `test/cli` but it seems that the Vite servers for the fixture projects do not get launched due to a port conflict with the Vitest instance running the actual test. I could not find a way to use a different port.

- [x] **(N/A)** Please, don't make changes to `pnpm-lock.yaml` unless you introduce a new test example.

### Tests

### Documentation
- [x] **(N/A)** If you introduce new functionality, document it. You can run documentation with `pnpm run docs` command.

### Changesets

### DIFF
--- a/packages/vitest/src/node/core.ts
+++ b/packages/vitest/src/node/core.ts
@@ -868,10 +868,10 @@ export class Vitest {
         for await (const project of teardownProjects.reverse())
           await project.teardownGlobalSetup()
 
-        const closePromises: unknown[] = this.projects.map(w => w.close().then(() => w.server = undefined as any))
+        const closePromises: unknown[] = this.resolvedProjects.map(w => w.close().then(() => w.server = undefined as any))
         // close the core workspace server only once
         // it's possible that it's not initialized at all because it's not running any tests
-        if (!this.projects.includes(this.coreWorkspaceProject))
+        if (!this.resolvedProjects.includes(this.coreWorkspaceProject))
           closePromises.push(this.coreWorkspaceProject.close().then(() => this.server = undefined as any))
 
         if (this.pool) {
--- a/packages/vitest/src/node/core.ts
+++ b/packages/vitest/src/node/core.ts
@@ -905,7 +905,7 @@ export class Vitest {
         this.state.getProcessTimeoutCauses().forEach(cause => console.warn(cause))
 
         if (!this.pool) {
-          const runningServers = [this.server, ...this.projects.map(p => p.server)].filter(Boolean).length
+          const runningServers = [this.server, ...this.resolvedProjects.map(p => p.server)].filter(Boolean).length
 
           if (runningServers === 1)
             console.warn('Tests closed successfully but something prevents Vite server from exiting')


AriPerkkio:
Interesting finding! How did you narrow the root cause to the websocket connections? Looks like Vitest's `hanging-process` reporter is unable to show those connections - I wonder if we could improve it to include them as well.

```sh
$ npx vitest run --project space_1 --reporter hanging-process --reporter basic

 RUN  v1.5.0 /Users/x/vitest/test/cli/fixtures/project

 ✓ |space_1| base.test.ts  (1 test) 1ms

 Test Files  1 passed (1)
      Tests  1 passed (1)
   Start at  08:15:25
   Duration  147ms (transform 12ms, setup 0ms, collect 6ms, tests 1ms, environment 0ms, prepare 45ms)

There are 12 handle(s) keeping the process running

# Tinypool
node:internal/async_hooks:202                                                                                    
node:internal/async_hooks:504                                                                                    
file:///Users/x/vitest/node_modules/.pnpm/tinypool@0.8.3/node_modules/tinypool/dist/esm/index.js:37 
file:///Users/x/vitest/node_modules/.pnpm/tinypool@0.8.3/node_modules/tinypool/dist/esm/index.js:58 
file:///Users/x/vitest/node_modules/.pnpm/tinypool@0.8.3/node_modules/tinypool/dist/esm/index.js:945
file:///Users/x/vitest/packages/vitest/dist/vendor/cac.WCfoZ7tE.js:8776                             
file:///Users/x/vitest/packages/vitest/dist/vendor/cac.WCfoZ7tE.js:9425                             
file:///Users/x/vitest/packages/vitest/dist/vendor/cac.WCfoZ7tE.js:9450                             

# FILEHANDLE
node:internal/async_hooks:202

... FILEHANDLE x9 here

# FILEHANDLE
node:internal/async_hooks:202
close timed out after 10000ms
Tests closed successfully but something prevents the main process from exiting
You can try to identify the cause by enabling "hanging-process" reporter. See https://vitest.dev/config/#reporters
```

It's only reporting `FILEHANDLE`'s for this case. The `# Tinypool \n node:internal/async_hooks:202` part is coming from here https://github.com/tinylibs/tinypool/blob/f86e82927371b54ba26577d818b86393ccf902c2/src/EventEmitterAsyncResource.ts#L24 - we should probably add `emitDestroy` for those in Tinypool's `pool.destroy()` so that they wouldn't show up in the report.

AriPerkkio (changes requested):
Same logic should be used here:

https://github.com/vitest-dev/vitest/blob/df6a432856b1165d1e7c4129ee9f35cc4fa6a365/packages/vitest/src/node/core.ts#L908-L917

Then the reproduction case would print following:

```
close timed out after 10000ms
Tests closed successfully but something prevents 2 Vite servers from exiting
```

surc54:
@AriPerkkio Thanks for the review! I've updated that location to use `this.resolvedProjects`.

As for how I tracked it down, I used a debugger to run `process._getActiveHandles()` during that timeout period, which gave back something like this:

```
[ WriteStream, ReadStream, WriteStream, Server ]
```

The server caught my interest, and upon checking its address, I noticed that it was on port 24678. Searching the vite repository led me to the dev server code.

sheremet-va (approved):

PR #5598, opened 2024-04-23
fix: always run `onTestFinished` in reverse order

### LINKED ISSUE #5529: Execute `onTestFinished` handlers in the reverse order they are scheduled in

Currently, it looks like `onTestFinished` runs as a FIFO queue:

```ts
onTestFinished(() => console.log("foo"));
onTestFinished(() => console.log("bar"));
```

Output after test:
```
foo
bar
```

However, it seems to be more useful if those registrations are run in the **reverse order** they were scheduled in, so you can colocate setup and teardown code and never have to worry about the order of execution here:

```ts
// Suppose this code lives in a `getDb()` helper
const db = setupDb();
onTestFinished(() => db.teardown());

// And this code lives in a `getTable()` helper
const table = db.createTable();
onTestFinished(() => db.dropTable(table.name));
```

Right now, `db.teardown()` gets called before `db.dropTable()`, which causes an issue.

### Suggested solution

If the order the teardown functions are run in is reversed, this class of problems will disappear.

### Alternative

_No response_

### Additional context

_No response_

### Validations

sheremet-va:
### Description

Fixes #5529

TODO:
- [x] Test

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -1010,6 +1010,10 @@ test('performs an organization query', async () => {
 })
 ```
 
+::: tip
+This hook is always called in reverse order and is not affected by [`sequence.hooks`](/config/#sequence-hooks) option.
+:::
+
 ### onTestFailed
 
 This hook is called only after the test has failed. It is called after `afterEach` hooks since they can influence the test result. It receives a `TaskResult` object with the current test result. This hook is useful for debugging.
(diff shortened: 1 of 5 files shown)


nvie:
Lovely, thanks @sheremet-va 🙏 !

PR #5608, opened 2024-04-24
fix: use package.json name for a workspace project if not provided

### LINKED ISSUE #5607: Use unique package names for workspaces

I set up a workspace with roughly 50 workspace packages, but I could not run all tests of the workspace because I ran into the following issue:

```
Error: Project name "client" is not unique. All projects in a workspace should have unique names.
```

A big project benefits from having similar structures in different parts of a big monorepo. So it is not surprising that packages with similar purpose end up having the same directory names.

### Suggested solution

Why not following rather the same uniqueness restrictions that already exist for package managers and hence will guarantee unique names?

The vitest workspace should identify workspace packages by the "name" property of the `package.json`.

Additionally: This will also reduce e.g. the nesting level seen in vitest VSCode plugin in the "Testing" panel.

### Alternative

_No response_

### Additional context

_No response_

### Validations

sheremet-va:
### Description

Closes #5607

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/plugins/workspace.ts
+++ b/packages/vitest/src/node/plugins/workspace.ts
@@ -1,4 +1,5 @@
-import { basename, dirname, relative } from 'pathe'
+import { existsSync, readFileSync } from 'node:fs'
+import { basename, dirname, relative, resolve } from 'pathe'
 import type { UserConfig as ViteConfig, Plugin as VitePlugin } from 'vite'
 import { configDefaults } from '../../defaults'
 import { generateScopedClassName } from '../../integrations/css/css-modules'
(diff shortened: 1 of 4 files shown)


packages/vitest/src/node/plugins/workspace.ts
```diff
@@ -35,10 +36,20 @@ export function WorkspaceVitestPlugin(project: WorkspaceProject, options: Worksp
...
-            name = basename(options.workspacePath.endsWith('/') ? options.workspacePath.slice(0, -1) : dirname(options.workspacePath))
-          else
+          if (typeof options.workspacePath === 'string') {
+            // if there is a package.json, read the name from it
+            const dir = options.workspacePath.endsWith('/')
+              ? options.workspacePath.slice(0, -1)
+              : dirname(options.workspacePath)
+            const pkgJsonPath = resolve(dir, 'package.json')
+            if (existsSync(pkgJsonPath))
+              name = JSON.parse(readFileSync(pkgJsonPath, 'utf-8')).name
```
AriPerkkio, line 46: What happens when there's no `name` in `package.json`?

AriPerkkio (commented):
I like the idea 👍

packages/vitest/src/node/plugins/workspace.ts
sheremet-va: I did not consider that 🤔

AriPerkkio (approved):

PR #5610, opened 2024-04-24
fix(vm): support network imports

sheremet-va:

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/runtime/external-executor.ts
+++ b/packages/vitest/src/runtime/external-executor.ts
@@ -27,7 +27,7 @@ export interface ExternalModulesExecutorOptions {
 }
 
 interface ModuleInformation {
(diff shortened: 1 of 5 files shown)


hi-ogawa (commented):
Thanks for picking this up. 

I wasn't sure if we should enable code execution from network by default.
I'm not really familiar with what people consider secure or insecure, but I think NodeJs and Deno requires explicit flags to enable this, so I was a bit scared of this.
Also a few more limitations on NodeJS are written here https://nodejs.org/api/esm.html#https-and-http-imports, for example, they only support "https" and also they probably don't allow `import "file:..."` in network modules.

What's your thought?

Also just cross-linking my old PR https://github.com/vitest-dev/vitest/pull/4996/

sheremet-va:
> What's your thought?

I think we should follow the limitations of Node here. I will update the PR

Issue #5630, opened 2024-04-27
Exclude {test,spec}-d files from coverage by default


sagargurtu:
### Clear and concise description of the problem

https://stackblitz.com/edit/vitest-dev-vitest-rrhq6a

Currently, {test,spec}-d files need to be manually added to `coverage.exclude`.

### Suggested solution

Regular test files are excluded from coverage by default. Similarly, it will be helpful to add {test,spec}-d file pattern to `coverageConfigDefaults.exclude`

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

opiredev:
@nabby27 created a $50.00 using [Opire](https://opire.dev) 
 <details><summary>How to earn this reward?</summary>

Since this project does not have Opire bot installed yet 😞 you need to go to [Opire](https://app.opire.dev) and claim the rewards through your programmer's dashboard once you have your PR ready 💪</details>

MindfulPol:
👋🏿 Hi @sagargurtu,

I've modified an existing glob pattern to additionally take into account the ''-d'' flag. Let me know if that's fine. 👍🏿  
Btw, found this through Opire. Cool software 😄

opiredev:
🔥 The user @MindfulPol has claimed all rewards for this issue using [Opire](https://opire.dev).
🎯 You can see the PR [here](https://github.com/vitest-dev/vitest/pull/5634)
💰 @nabby27 you can pay the related rewards [here](https://app.opire.dev/dashboard/creator/rewards?platform=GitHub&platformId=2266867161)

opiredev:
🥳 @nabby27 completed the payment of $50.00 in rewards for this issue 💰
The rewarded user is @MindfulPol - congratulations and thanks for your effort! 🎉

PR #5634, opened 2024-04-28
fix(coverage): exclude `*.test-d.*` by default

MindfulPol:
### Description

Solves issue #5630 5630
Exclude {test,spec}-d files from coverage by default 

- [x ] It's really useful if your PR references an issue where it is discussed ahead of time. If the feature is substantial or introduces breaking changes without a discussion, PR might be closed.
- [ x] Please, don't make changes to `pnpm-lock.yaml` unless you introduce a new test example.

### Tests
- [ x] Run the tests with `pnpm test:ci`.

### Documentation

### Changesets
- [x ] Changes in changelog are generated from PR name. Please, make sure that it explains your changes in an understandable manner. Please, prefix changeset messages with `feat:`, `fix:`, `perf:`, `docs:`, or `chore:`.


opiredev:
The user @MindfulPol has claimed all the rewards in #5630 👀 We've notified the reward creators [here](https://github.com/vitest-dev/vitest/issues/5630#issuecomment-2081688655).
@MindfulPol make sure you have your [payment account ready](https://app.opire.dev/settings) to receive the payment for your hard work 💪

AriPerkkio (changes requested):
As test case let's add `should-be-excluded-from-report.test-d.ts` to `test/coverage-test/src/`. Simply adding that file should be enough as coverage tests are already validating all files inside that directory. If the new exclusion pattern is removed, some snapshot tests should fail.

MindfulPol:
Hey @AriPerkkio ,
File added, thank you for pointing out. 🙏🏿 
I've modified my last commit message name so it's called ''fix'' rather than ''feat'' 😄

PR #5651, opened 2024-05-02
chore: set `package-manager-strict`

btea:
### Description
After pnpm releases version [9.0.3](https://github.com/pnpm/pnpm/releases/tag/v9.0.3), when the local pnpm version and packageManger version do not match, an error message will appear and the process will exit when running `pnpm i`.

### Tests

### Documentation

### Changesets

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -4,3 +4,4 @@ provenance=true
 shell-emulator=true
 registry=https://registry.npmjs.org/
 VITE_NODE_DEPS_MODULE_DIRECTORIES=/node_modules/,/packages/
+package-manager-strict=false


AriPerkkio (approved):

dammy001 (approved):

hi-ogawa:
Is strict mode so bad?

dammy001:
I feel PNPM should make this false by default.

hi-ogawa:
I don't know enough about this to judge, so I was just wondering what's the downside.
For example, how is it different from not setting `packageManager` at all?

btea:
@hi-ogawa For example, if I run `pnpm build` in the root directory, the following error message will appear:

image

hi-ogawa:
The error message seems legitimate though. I see the downside could be that it might make it harder for new contributors to onboard, but wouldn't it be also a choice to let contributors opt-out by putting `COREPACK_ENABLE_STRICT=0` in their shell or something if we/they don't bother about the strictness?

Since pnpm has decided this behavior, I would imagine you'll essentially need to make `package-manager-strict=false` PRs for all the projects you contribute and that seems a little unrealistic.

As a Vitest member, I'm totally fine to `package-manager-strict=false` on this repo to increase the ease of contribution, but I'm just wondering what is your experience so far with pnpm 9 without corepack.

btea:
> The error message seems legitimate though. I see the downside could be that it might make it harder for new contributors to onboard, but wouldn't it be also a choice to let contributors opt-out by putting `COREPACK_ENABLE_STRICT=0` in their shell or something if we/they don't bother about the strictness?
> 

Yes, what you mentioned makes sense. However, it seems that existing contributors will also need to configure the environment individually. (After all, it is impossible for everyone’s locally installed pnpm version to definitely match the specified version.)
> Since pnpm has decided this behavior, I would imagine you'll essentially need to make `package-manager-strict=false` PRs for all the projects you contribute and that seems a little unrealistic.
> 

Yes, it's a bit of a hassle. So, I agree with @dammy001 . IMO, there won't be much difference in functionality between minor versions, and forcing an exact match seems unnecessary.

> but I'm just wondering what is your experience so far with pnpm 9 without corepack.

I have just used pnpm not long ago, and this is the point I am more concerned about right now.

btea:
Okay, let's close this PR.

Issue #5585, opened 2024-04-22
Sort file alphabetically with vitest --ui


zeid0ne:
### Clear and concise description of the problem

When i run my test with vitest --ui, they are all displayed in a random order.

![image](https://github.com/vitest-dev/vitest/assets/38461940/8aa0ad79-776e-4116-b459-c742e0adf2f5)

Sometime i want to re run the tests for one file, but if i have a large number of them, it's long to find the line corresponding to the file. If they were sorted alphabetically, i'll could do it faster. I thinks it's also more "beautifull".

### Suggested solution

Sort the files alphabetically.

### Alternative

Specify file order manually.

### Additional context

https://github.com/vitest-dev/vitest/discussions/5583

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

AriPerkkio:
Fixed by https://github.com/vitest-dev/vitest/pull/5652

PR #5652, opened 2024-05-02
feat(ui): sort items by file name

btea:
### Description
Rel #5585

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/ui/client/composables/client/index.ts
+++ b/packages/ui/client/composables/client/index.ts
@@ -40,9 +40,19 @@ export const client = (function createVitestClient() {
   }
 })()
 
+function sort(a: File, b: File) {
+  if (a.name < b.name)
+    return -1
+
+  if (a.name > b.name)
+    return 1
+
+  return 0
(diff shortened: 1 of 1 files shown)


packages/ui/client/composables/client/index.ts
```diff
@@ -40,9 +40,19 @@ export const client = (function createVitestClient() {
   }
 })()
 
+function sort(a: File, b: File) {
+  if (a.name < b.name)
```
sheremet-va, line 44: Not sure that this is correct. What if it has unicode characters? Shouldn't `.localeCompare` be used here?

sheremet-va (changes requested):

packages/ui/client/composables/client/index.ts
btea: Yeah, it looks better using `localeCompare`, I've updated it.

sheremet-va (approved):

PR #5677, opened 2024-05-06
ci: run tests sequentially 

sheremet-va:

### Tests

### Documentation

### Changesets

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,8 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        node_version: [18, 20, 22]
+        node_version: [18, 20]
+        # node_version: [18, 20, 22] 22 when LTS is close enough
         include:
           - os: macos-14
             node_version: 20
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "publish-ci": "tsx scripts/publish-ci.ts",
     "release": "bumpp package.json packages/*/package.json --commit --push --tag && git update-ref refs/heads/release refs/heads/main && git push origin release",
     "test": "pnpm --filter test-core test:threads",
-    "test:ci": "CI=true pnpm -r --reporter-hide-prefix --stream --filter '@vitest/test-*' --filter !test-browser run test --workspace-concurrency=1",
+    "test:ci": "CI=true pnpm -r --reporter-hide-prefix --stream --sequential --filter '@vitest/test-*' --filter !test-browser run test",
     "test:examples": "CI=true pnpm -r --reporter-hide-prefix --stream --filter '@vitest/example-*' run test",
     "test:ecosystem-ci": "ECOSYSTEM_CI=true pnpm test:ci",
     "typecheck": "tsc -p tsconfig.check.json --noEmit",
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "publish-ci": "tsx scripts/publish-ci.ts",
     "release": "bumpp package.json packages/*/package.json --commit --push --tag && git update-ref refs/heads/release refs/heads/main && git push origin release",
     "test": "pnpm --filter test-core test:threads",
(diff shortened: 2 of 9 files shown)


sheremet-va:
> Is #5678 on this branch?

Yes, these failings are why I opened a PR. No idea why it's failing yet.

sheremet-va:
> is failing randomly without any error logs

Do you think it might be because we have 3 different pools that have their own min/max worker count?

AriPerkkio:
Oh right, those pools are run in parallel. The failing linux CIs have 2 CPUs so each pool is using just one worker. But still there's main thread + 3 pool threads running.

AriPerkkio:
Actually I can reproduce this issue locally with Node `v22.1.0` - really weird:


https://github.com/vitest-dev/vitest/assets/14806298/6fe62d9f-1252-413a-b40a-a950bacc9408

AriPerkkio:
Isolated the crashing test to `pnpm run test --run --project=vmThreads injector-mock.test.ts`. Changing that file to contain just the content below is enough to crash:

```ts
import { test } from 'vitest'
import { hoistMocks } from '../../../packages/vitest/src/node/hoistMocks'

test.only('Helloo', () => {
  console.log(hoistMocks)
})
```

sheremet-va:
> Isolated the crashing test to `pnpm run test --run --project=vmThreads injector-mock.test.ts`. Changing that file to contain just the content below is enough to crash:
> 
> ```ts
> import { test } from 'vitest'
> import { hoistMocks } from '../../../packages/vitest/src/node/hoistMocks'
> 
> test.only('Helloo', () => {
>   console.log(hoistMocks)
> })
> ```

Interesting. The only thing I can think of is that it imports parseAst which relies on a native swc module 🤔

sheremet-va:
> The only thing I can think of is that it imports parseAst

Oh, it doesn't actually import it, the plugin provides it 🤔

sheremet-va:
Looks like importing `generateCodeFrame` can cause this. Digging deeper 🔍

sheremet-va:
Almost always fails when importing `vite`:

```
creating module : /vitest/node_modules/.pnpm/vite@5.2.6_@types+node@20.11.5/node_modules/vite/dist/node/index.js
```

🔍 

https://github.com/vitest-dev/vitest/blob/0227e8cbb4db2498d50cf0cc20dc12711e8374ae/packages/vitest/src/runtime/external-executor.ts#L232

sheremet-va:
Looks like it always fails here:

https://github.com/vitest-dev/vitest/blob/0227e8cbb4db2498d50cf0cc20dc12711e8374ae/packages/vitest/src/runtime/vm/file-map.ts#L11

No error, just exit with 129 code - interestingly, it kills even the main thread

AriPerkkio (commented):
Could you try if `v22.0.0` works on CI? Locally it doesn't crash like `v22.1.0` does. This could help to pinpoint the root cause. 👀 

https://github.com/vitest-dev/vitest/blob/bcccce6df63b7c0408c6d14978e7cde82f1c90ca/.github/workflows/ci.yml#L42

```diff
-        node_version: [18, 20, 22]
+        node_version: [18, 20, 22.0.0]
```

sheremet-va:
> Could you try if `v22.0.0` works on CI? Locally it doesn't crash like `v22.1.0` does. This could help to pinpoint the root cause. 👀
> 
> https://github.com/vitest-dev/vitest/blob/bcccce6df63b7c0408c6d14978e7cde82f1c90ca/.github/workflows/ci.yml#L42
> 
> ```diff
> -        node_version: [18, 20, 22]
> +        node_version: [18, 20, 22.0.0]
> ```

Actually, I am not even sure why we test against Node 22, it's not LTS yet. We usually wait for it to be LTS or close to it before we add it to CI because it's prone to bugs like this,

AriPerkkio:
Sure, we could drop it until its LTS plans have been announced and add it back some weeks/months before that.

PR #5689, opened 2024-05-09
docs: add docs for previous versions

### LINKED ISSUE #1110: Support to view the contents of old version documents

I can't find an entry to jump to the old version of the document on vite's official website https://vitest.dev/.
Although the header has a v0 8.4 buttons.
But I can't see the content of the old vite document.

image

I think this function is very important, because if the old project uses the old version and it is inconvenient to upgrade, in this case, I need to check the document content corresponding to the old version to enter business development.

### Suggested solution

We can refer to angular or jest. There is an entry to display the historical version number.

#### In angular js

https://docs.angularjs.org/tutorial

Documents like angular are consistent with the release version.

wlOE9FWBZU

#### In jest

https://jestjs.io/

Documents like jest are also consistent with the release version.

MoTsTvYXgy

### Alternative

_No response_

### Additional context

_No response_

### Validations

antfu:
fix #1110

Screenshot 2024-05-09 at 10 12 06


docs/.vitepress/config.ts
```diff
@@ -121,22 +121,39 @@ export default ({ mode }: { mode: string }) => {
...
+                {
+                  text: 'Releases Notes',
                   link: releases,
                 },
                 {
                   text: 'Contributing ',
                   link: contributing,
                 },
-              ],
+              ]
```
sheremet-va, line 145:
```suggestion
              ],
```

docs/.vitepress/config.ts
```diff
@@ -121,22 +121,39 @@ export default ({ mode }: { mode: string }) => {
...
             },
+            {
+              items: [
+                {
+                  text: 'v0.x',
+                  link: 'https://v0.vitest.dev/',
+                }
```
sheremet-va, line 152:
```suggestion
                },
```

docs/.vitepress/config.ts
```diff
@@ -121,22 +121,39 @@ export default ({ mode }: { mode: string }) => {
...
+              ]
```
sheremet-va, line 153:
```suggestion
              ],
```

docs/.vitepress/config.ts
```diff
@@ -121,22 +121,39 @@ export default ({ mode }: { mode: string }) => {
...
+            }
```
sheremet-va, line 154:
```suggestion
            },
```

PR #5708, opened 2024-05-11
feat: add expect.poll utility

sheremet-va:
### Description

When rewriting UI tests, I found that we don't have an assertion resistant to flaky values. This should make it easier to check if a flaky condition is working correctly. I also expect to have this functionality built-in in future browser-only assertions.

TODO:

- [x] Docs
- [x] Tests

### Tests

### Documentation

### Changesets


AriPerkkio (commented):
How is this different from

```ts
await vi.waitFor(() => {
  expect(..).toBe(...)
});

**Comment by sheremet-va** [comment]
> How is this different from
> 
> ```ts
> await vi.waitFor(() => {
>   expect(..).toBe(...)
> });
> ```

It's more compact. Having several `vi.waitFor` just doesn't look as clean.

Playwright also has `await expect(() => {}).toPass()` API which looks more like `vi.waitFor`

File: packages/vitest/src/integrations/chai/poll.ts [hunk 1]
```diff
@@ -0,0 +1,75 @@
...
+            let intervalId: any
+            let lastError: any
+            const { setTimeout } = getSafeTimers()
+            setTimeout(() => {
+              clearTimeout(intervalId)
+              reject(copyStackTrace(new Error(`Matcher did not succeed in ${timeout}ms`, { cause: lastError }), STACK_TRACE_ERROR))
+            }, timeout)
+            const check = async () => {
+              try {
+                chai.util.flag(this, 'object', await fn())
```
AriPerkkio, line 54: If the `fn` throws, the printed error message points to the `expect` call. Should we instead capture the error thrown by `fn` and show its source in the code frame?

<details>

```ts
import { expect, test } from "vitest";
import { sum } from "../src/example";

test("testing", async () => {
  await expect.poll(() => sum(1, 2)).toBe(3);
});
```

```ts
export function sum(a, b) {
  a.b.nonExistingField;
  return a + b + 1;
}
```

```
 FAIL  test/third.test.ts > third
Error: Matcher did not succeed in 1000ms
 ❯ test/third.test.ts:5:16
      3| 
      4| test("third", async () => {
      5|   await expect.poll(() => sum(1, 2)).toBe(3);
       |                ^
      6| });
      7| 

Caused by: TypeError: Cannot read properties of undefined (reading 'nonExistingField')
 ❯ Module.sum src/example.ts:6:7
 ❯ test/third.test.ts:5:27
```

</details>

Showing `expect` as erroneous lines is correct below when `result.call()` fails.

sheremet-va: Not sure how this would work. Do you want to replace the first error callstack? Do you want to rethrow the original error? How do you see the end result?

AriPerkkio: Something like this:

```ts
  let intervalId: any
  let fnError: any
  let lastError: any
  const { setTimeout } = getSafeTimers()
  setTimeout(() => {
    clearTimeout(intervalId)
    const error = fnError || copyStackTrace(new Error(`Matcher did not succeed in ${timeout}ms`, { cause: lastError }), STACK_TRACE_ERROR)
    reject(error)
  }, timeout)
  const check = async () => {
    let output

    try {
      output = await fn()
      fnError = null
    }
    catch (err) {
      fnError = err
      intervalId = setTimeout(check, interval)
      return
    }

    try {
      chai.util.flag(this, 'object', output)
      resolve(await result.call(this, ...args))
    }
    catch (err) {
      lastError = err
      intervalId = setTimeout(check, interval)
    }
  }
  check()
})
```

Now it will point to the errorneous line if `fn()` threw error:

```ts
TypeError: Cannot read properties of undefined (reading 'nonExistingField')
 ❯ Module.sum src/example.ts:6:7
      4| 
      5| export function sum(a, b) {
      6|   a.b.nonExistingField;
       |       ^
      7|   return a + b + 1;
      8| }
 ❯ test/third.test.ts:5:27
 ```
 
 And when the `fn()` successfully returns a value but `expect().<matcher>()` fails:
 
 ```ts
 Error: Matcher did not succeed in 1000ms
 ❯ test/third.test.ts:5:16
      3| 
      4| test("third", async () => {
      5|   await expect.poll(() => sum(1, 2)).toBe(3);
       |                ^
      6| });
      7| 
```

sheremet-va: But this is not entirely correct - the error happened because it didn't return a valid value before the interval timed out, so the `Matcher did not succeed` is the correct error here - it doesn't matter that it happened inside the callback

AriPerkkio: So it's expected that in `await expect.poll(() => sum())`, if there's an error inside `sum()`, it's expected to swallow the error and print just `'Matcher did not succeed in 1000ms'`. Not show that `sum()` actually threw an error? It's not the `.toBe()` that failed here.

sheremet-va: What do you mean by "swallow"? It is printed in `cause`

sheremet-va: I do agree that it's not that visible, but we also need to show the timeout error somewhere

sheremet-va: Interestingly enough, the playwright doesn't even allow errors in `poll`:

```js
  1) playwright.test.js:2:5 › basic test 

    Error: not yet

      3 |   await expect.poll(() => {
      4 |     throw new Error('not yet')
    > 5 |   }).toBe(3)
        |      ^
      6 | })
```

AriPerkkio: Right, sure it's visible in stack trace and even the error message of original error is printed there. I guess it's just the code frame that's bothering me.

I'm fine with the current implementation as is. 👍

sheremet-va: On the other hand, it would be more consistent to go with the `vi.waitFor` implementation where we always reject the last error

PR #5724, opened 2024-05-14
feat(utils): Produce valid snapshot names

### LINKED ISSUE #5681: Producing unstable snapshots
### Describe the bug

The name of the test printed in the snapshot file may include wrongly-formed characters. On windows, it leads Vitest to consider the snapshot as obsolete and recreate a new one for each run.

**Where do the bug comes from?** -> loupe

The bug is visible in Vitest but comes from loupe and has been reported on their side too at https://github.com/chaijs/loupe/issues/78.

In Vitest the call to loupe responsible to the problem is in `@vitest/utils/dist/chunk-display.js` within `objDisplay`. Once fixed on loupe's side I can probably help and come with some dedicated tests covering that issue to avoid it to come back.

**Temporary fix**

While waiting for a fix on loupe's side, I can propose a fix on your side that will drop/replace surrogates not being in pairs? What do you think?

### Reproduction

Consider the following example:

```js
import { expect, test } from "vitest";

test.each([{ value: "\u{1f431}".repeat(50) }])(
  "print the value properly for value=$value",
  ({ value }) => {
    expect(value).toMatchSnapshot();
  }
);
```

It leads to the snapshot-ed value:

```txt
exports[`print the value properly for value='🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱�…' 1`] = `"🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱🐱"`;
```

This snapshot name is invalid as it comes with a split-cat or more precisely a split-surrogate-pair which is not an existing character outside of JavaScript.

### System Info

```shell
System:
    OS: Windows 11 10.0.22621
    CPU: (4) x64 Intel(R) Core(TM) i5-3570K CPU @ 3.40GHz
    Memory: 1.50 GB / 7.89 GB
  Binaries:
    Node: 20.8.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.21 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 10.2.3 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Chromium (123.0.2420.97)
  npmPackages:
    vitest: ^1.5.3 => 1.5.3
```

### Used Package Manager

npm

### Validations

dubzzz:
### Description

The previous version of loupe pulled within Vitest was not properly splitting strings to produce names of snapshots. The proposed PR bumps it and adds the relevant tests.

Fixes #5681

_Note: I feel the lock has changed a bit more than what I would have expected. I updated it with pnpm@8.15.8 from a GitHub Codespace. I don't really get how to properly update it if it's not the right one._

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/utils/package.json
+++ b/packages/utils/package.json
@@ -63,7 +63,7 @@
   "dependencies": {
     "diff-sequences": "^29.6.3",
     "estree-walker": "^3.0.3",
-    "loupe": "^3.1.0",
+    "loupe": "^3.1.1",
     "pretty-format": "^29.7.0"
   },
   "devDependencies": {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -805,8 +805,8 @@ importers:
         specifier: ^3.0.3
         version: 3.0.3
       loupe:
-        specifier: ^3.1.0
-        version: 3.1.0
+        specifier: ^3.1.1
+        version: 3.1.1
       pretty-format:
         specifier: ^29.7.0
         version: 29.7.0
@@ -837,7 +837,7 @@ importers:
         version: 1.0.0
       vite:
         specifier: ^5.2.6
-        version: 5.2.6
+        version: 5.2.6(@types/node@20.12.11)
     devDependencies:
       '@jridgewell/trace-mapping':
         specifier: ^0.3.25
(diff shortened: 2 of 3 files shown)


sheremet-va (approved):

sheremet-va:
Looks good to me! Thank you! The lint error is also on main, two PRs merged too quickly 😄

PR #5727, opened 2024-05-14
docs: remove version badges

### LINKED ISSUE #5719: Remove version labels for Vitest 2 documentation

We now have different documentation websites for different Vitest versions. This gives us the opportunity to clear up the documentation. 

### Suggested solution

I think we should remove all 0+ and 1+ version labels for the Vitest 2 documentation. 

Labels will still be displayed in Vitest 1 and Vitest 0 documentations.

### Alternative

_No response_

### Additional context

_No response_

### Validations

sheremet-va:
### Description

This PR removes version badges (fixes #5719) and adds a migration guide from v1 to v2 (only mentions changes merged so far).

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/guide/migration.md
+++ b/docs/guide/migration.md
@@ -48,9 +48,9 @@ export default defineConfig({
 
 Previously, specifying `concurrent` on a suite would still group concurrent tests by suites and run them together one by one. Now, it follows jest's behaviour and runs all of them at once (still limited by [`maxConcurrency`](/config/#maxConcurrency))
 
-### enable `coverage.ignoreEmptyLines` by default
+### enable V8 coverage's `coverage.ignoreEmptyLines` by default
 
-Changes default value of `coverage.ignoreEmptyLines` to `true`. This change will have major impact on users' code coverage reports. It is likely that projects using coverage thresholds need to adjust those values after this.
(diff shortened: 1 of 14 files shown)


hi-ogawa (approved):
Looks good!

docs/guide/migration.md
```diff
@@ -7,11 +7,76 @@ outline: deep
...
+})
+```
+
+### `suite.concurrent` runs all its tests concurrently
+
+Previously, specifying `concurrent` on a suite would still group concurrent tests by suites and run them together one by one. Now, it follows jest's behaviour and runs all of them at once (still limited by [`maxConcurrency`](/config/#maxConcurrency))
+
+### enable `coverage.ignoreEmptyLines` by default
+
+Changes default value of `coverage.ignoreEmptyLines` to `true`. This change will have major impact on users' code coverage reports. It is likely that projects using coverage thresholds need to adjust those values after this.
```
AriPerkkio, line 53: Might be good to mention that it affects only V8-coverage users. Something like: 

```suggestion
### enable V8 coverage's `coverage.ignoreEmptyLines` by default

Changes default value of `coverage.ignoreEmptyLines` to `true`. This change will have major impact on users' code coverage reports. It is likely that projects using coverage thresholds need to adjust those values after this. This change affects only the default `coverage.provider: 'v8'`.
```

AriPerkkio (approved):
Looks good. Before the official v2 release let's take a look at the migration steps once more just in case.

PR #5745, opened 2024-05-17
feat: add median to `--output-json`

Joristdh:
### Description
In #5398 the JSON reporter got reworked for the benchmark results. In this rework, the samples array was dropped to prevent huge output files. However, without the samples it's not possible anymore to manually calculate some extra metrics after the bench run. The median is simple enough to calculate when having acces to the samples and can provide some interesting insights (e.g. significant difference between mean and median could indicate outliers).

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/reporters/benchmark/table/index.ts
+++ b/packages/vitest/src/node/reporters/benchmark/table/index.ts
@@ -129,6 +129,7 @@ interface FormattedBenchmarkGroup {
 export type FormattedBenchmarkResult = Omit<BenchmarkResult, 'samples'> & {
   id: string
   sampleCount: number
+  median: number
 }
 
 function createFormattedBenchamrkReport(files: File[]) {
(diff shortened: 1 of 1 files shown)


Joristdh:
If the Vitest framework isn't the right place for new metrics, the change could also be migrated to the underlying Tinybench results.

sheremet-va:
> If the Vitest framework isn't the right place for new metrics, the change could also be migrated to the underlying Tinybench results.

I guess it makes sense to have it in Vitest because you can extrapolate it in tinybench from `samples`, although it might also be nice for tinybench users to have a quick access to median on the object, so I am fine either way

sheremet-va (approved):

Joristdh:
I think indeed that calculating the median is trivial when having access to all the samples, so this mainly adds value for Vitest users. 

Maybe it'd be even better to just add a flag to preserve the original `samples`. Then Vitest users can deal with/use the results however they see fit. 

In any case, that would be something for another PR. Is there something I just do to get this merged after the approval @sheremet-va?

Issue #5744, opened 2024-05-17
test.each does not print -0 and -NaN correctly


arv:
### Describe the bug

When using `test.each` the descriptive text for `-0` and `-NaN` is printed as `0` and `NaN`.

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-gkmdyn?file=test%2Fbasic.test.ts

```ts
import { test, expect } from 'vitest';

test.each([[0], [-0], [NaN], [-NaN], [Infinity], [-Infinity]] as const)(
  '%f',
  (n) => {
    expect(n).toBe(n);
  }
);

```

The terminal output from this is:

```
 ✓ test/basic.test.ts (6)
   ✓ 0
   ✓ 0
   ✓ NaN
   ✓ NaN
   ✓ Infinity
   ✓ -Infinity
```

the expected output is:

```
 ✓ test/basic.test.ts (6)
   ✓ 0
   ✓ -0
   ✓ NaN
   ✓ -NaN
   ✓ Infinity
   ✓ -Infinity
```

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 18.18.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.2.3 - /usr/local/bin/npm
    pnpm: 8.15.6 - /usr/local/bin/pnpm
  npmPackages:
    @vitest/ui: latest => 1.6.0 
    vite: latest => 5.2.11 
    vitest: latest => 1.6.0
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
[...]

PR #5775, opened 2024-05-24
fix(utils): Ensure test.each print -0 and -NaN properly

pengooseDev:
### Description
Fixes [issue #5744](https://github.com/vitest-dev/vitest/issues/5744)

This bug fix ensures that when using test.each, the descriptive text for `-0` and `-NaN` is correctly printed as `-0` and `-NaN` instead of `0` and `NaN`.

For better readability, I have standardized the test conventions to `it should be formatted ...` for added tests. Additionally, I plan to create an issue soon to discuss improving test code readability and establishing conventions. I welcome any feedback on the code and the proposed direction. :)

- enhance format logic
- add test suite(formats special cases)
- add util function(isNegativeNaN)
- add test suite(isNegativeNaN)

image
+ I have ensured that the descriptive text for -0 and -NaN is printed correctly as -0 and -NaN instead of 0 and NaN when using test.each. The issue was not with the each method itself, and the tests have passed successfully.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -57,6 +57,7 @@
   },
   "devDependencies": {
     "@faker-js/faker": "^8.4.1",
+    "@iconify-json/carbon": "^1.1.32",
     "@iconify-json/logos": "^1.1.42",
     "@testing-library/vue": "^8.0.3",
     "@types/codemirror": "^5.60.15",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -717,6 +717,9 @@ importers:
       '@faker-js/faker':
         specifier: ^8.4.1
         version: 8.4.1
+      '@iconify-json/carbon':
+        specifier: ^1.1.32
+        version: 1.1.32
       '@iconify-json/logos':
         specifier: ^1.1.42
         version: 1.1.42


sheremet-va:
It doesn't look like this is how Node.js does it:

https://nodejs.org/api/util.html#utilformatformat-args

> %f: parseFloat(value) is used for all values expect Symbol.

pengooseDev:
@sheremet-va 
Thanks for yout feedback. :)

Issue #5744 requests that `-0` and `-NaN` should be rendered as `-0` and `-NaN` respectively when formatted with %f, instead of 0 and NaN. Considering the `pending triage` label has been removed and the `p2-edge-case` label has been added, I understood this as an indication of support for this feature.

1. I am curious if you would like to add the functionality to format -0 and -NaN as they are with the %f format. If it is determined to be difficult to implement, I will close this PR and work on other tasks. :)

2. If you would like to add this feature, I would appreciate your feedback on how the code should be changed to achieve this (e.g., changing the location of the logic).

### Why ArrayBuffer?
+ parseFloat returns 0 and NaN for -0 and -NaN respectively. Additionally, Object.is(NaN, -NaN) and Object.is(NaN, NaN) both return false, making it difficult to check the sign of NaN without the help of ArrayBuffer. This is why I added the isNegativeNaN function.

+ If you believe this approach does not align with your desired direction, please feel free to close this PR. I am always greatly helped by the amazing Vitest project. Thank you. :)

sheremet-va:
> Issue #5744 requests that `-0` and `-NaN` should be rendered as `-0` and `-NaN` respectively when formatted with %f, instead of 0 and NaN. Considering the `pending triage` label has been removed and the `p2-edge-case` label has been added, I understood this as an indication of support for this feature.

Yes, I am fine with changing this in `.each`, but it doesn't mean we can change the Node.js `format` since it would be a breaking change. I will be happy if we can bypass this somehow.

> Why ArrayBuffer?

I am fine with any implementation of `-NaN` detection.

pengooseDev:
I understand :) First of all, after revert the changes, I will also try to find a way with you.
If you find a good way, I would appreciate it if you could comfortably leave a comment on the PR.
Thank you for your direction!

pengooseDev:
I'll create new PR soon :)

PR #5791, opened 2024-05-29
fix: respect `isolate` in a workspace project

sheremet-va:
### Description

TODO

- [ ] Add a test

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/pools/threads.ts
+++ b/packages/vitest/src/node/pools/threads.ts
@@ -156,7 +156,8 @@ export function createThreadsPool(ctx: Vitest, { execArgv, env }: PoolProcessOpt
         const results: PromiseSettledResult<void>[] = []
 
         const { isolatedFiles, nonIsolatedFiles } = files.reduce((acc, spec) => {
-          const key = spec.project.config.isolate ? 'isolatedFiles' : 'nonIsolatedFiles'
+          const isolatedProject = getConfig(spec.project).poolOptions?.threads?.isolate ?? isolated
+          const key = isolatedProject ? 'isolatedFiles' : 'nonIsolatedFiles'
           acc[key].push(spec)
           return acc
         }, { isolatedFiles: [] as GroupedSpec[], nonIsolatedFiles: [] as GroupedSpec[] })
--- a/packages/vitest/src/node/pools/threads.ts
+++ b/packages/vitest/src/node/pools/threads.ts
@@ -6,6 +6,7 @@ import Tinypool from 'tinypool'
 import { resolve } from 'pathe'
 import type { ContextTestEnvironment, ResolvedConfig, RunnerRPC, RuntimeRPC, Vitest, WorkerContext } from '../../types'
 import type { PoolProcessOptions, ProcessPool, RunWithFiles } from '../pool'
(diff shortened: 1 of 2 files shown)


yannbf:
Here is a reproduction for the issue:
https://github.com/yannbf/reproduction-vitest-workspace-isolate

If you run `yarn vitest:single` which uses `isolate: false` in a config file you will get:
image

If you run `yarn vitest:workspace` which uses `isolate: false` in a workspace config file you will get:
image


If you open the `vitest-workspace.config.ts` file and uncomment the singleThread code, then run `yarn vitest:workspace` again, you will get:
image

packages/vitest/src/node/pools/threads.ts
```diff
@@ -154,14 +155,27 @@ export function createThreadsPool(ctx: Vitest, { execArgv, env }: PoolProcessOpt
...
-          results.push(...await Promise.allSettled(files.map(({ file, environment, project }) =>
+        const { isolatedFiles, nonIsolatedFiles } = files.reduce((acc, spec) => {
+          const isolatedProject = getConfig(spec.project).poolOptions?.threads?.isolate ?? isolated
+          const key = isolatedProject ? 'isolatedFiles' : 'nonIsolatedFiles'
+          acc[key].push(spec)
+          return acc
+        }, { isolatedFiles: [] as GroupedSpec[], nonIsolatedFiles: [] as GroupedSpec[] })
+
+        if (isolatedFiles.length) {
+          results.push(...await Promise.allSettled(isolatedFiles.map(({ file, environment, project }) =>
```
sheremet-va, line 166: It now looks the same as the one in `nonIsolatedFiles.length` 🤔 But it would actually break of the environment is different 🤔 

@AriPerkkio do we even need to check for isolation at pool level? 🤔

AriPerkkio: Resolving `poolOptions` in workspaces is complicated. 😢  

What exactly is not working here? Setting `{ poolOptions: { threads: { isolate: false } } }` in workspace project? It should have been fixed in https://github.com/vitest-dev/vitest/pull/4765 already. I think we decided that project specific `isolate` doesn't affect how pool is run from the main thread. It only affects how worker handles isolation.

sheremet-va: The issue is not with resolving options. It's if we do resolve them, then I can see that we don't even need the `if(isolate)` because both branches look identical

Why don't we already group isolated files by environment?

AriPerkkio: > Why don't we already group isolated files by environment?

I think that when isolation is enabled, there is no need to do any grouping when passing the files and options to pool. Each worker is always terminated after a test run so we can assume that environment is set up freshly. There should not be a need for (manual) `pool.recycleWorkers()` calls when isolation is enabled. 

When isolation is disabled, we run as many files in same worker as possible. If non-isolated files have different enviroments, we'll still need to recycle workers between those groups.

sheremet-va:
I don't think it's possible to make this work consistently on the project level

marcelarie:
This will be really helpful for big projects where half of the test are compatible with `--no-isolate` and the other don't.

PR #5810, opened 2024-06-01
feat(browser): do not reload the page during watch mode

sheremet-va:
### Description

Closes #3584

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/core.ts
+++ b/packages/vitest/src/node/core.ts
@@ -191,13 +191,13 @@ export class Vitest {
 
   public getProjectByTaskId(taskId: string): WorkspaceProject {
     const task = this.state.idMap.get(taskId)
-    const projectName = (task as File).projectName || task?.file?.projectName
+    const projectName = (task as File).projectName || task?.file?.projectName || ''
     return this.projects.find(p => p.getName() === projectName)
       || this.getCoreWorkspaceProject()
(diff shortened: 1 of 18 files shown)


packages/vitest/src/node/pools/browser.ts
```diff
@@ -46,7 +46,16 @@ export function createBrowserPool(ctx: Vitest): ProcessPool {
 
     const promise = waitForTests(project, files)
 
-    await provider.openPage(new URL('/', origin).toString())
+    const orchestrators = project.browserRpc.orchestrators
```
sheremet-va, line 49: Since we now have direct RPC to every orchestrator, we can just ask them to run tests

packages/vitest/src/node/core.ts
```diff
@@ -609,7 +609,7 @@ export class Vitest {
     await this.report('onPathsCollected', filepaths)
     await this.report('onSpecsCollected', specs.map(
       ([project, file]) =>
-        [{ name: project.getName(), root: project.config.root }, file] as SerializableSpec,
+        [{ name: project.config.name, root: project.config.root }, file] as SerializableSpec,
```
sheremet-va, line 612: This caused a bug duplicating tasks in the UI, it's noticeable in watch mode, but easy to miss if the page is refreshed

PR #5844, opened 2024-06-04
feat(ui): load module graph on tab selection

userquin:
### Description

This PR includes:
- load module graph only when the graph tab is selected and when the current module changes
- show file name relative to the root config in `FileDetails.vue` header (rigth now, `relativeToRoot` function using `config.root` where `config` is a vue ref, and so the full path is shown)
- added some icons to the tabs: since we've moved the details from the mid panel, we've some extra space (check screenshot below)
- normalize all composable imports using `~`  instead relative paths: I'll check next weekend to configure autoimport properly, we should remove all composable  imports from vue files
- ~~WIP: change module entry dialog header to include the project and the relative path: in the current version showing only the full path (check screenshot below)~~
- ~~added watcher to select report tab in details pane when selected file change~~
- ~~WIP: I'll try to remove previous watcher and reload the graph when the active file changes and the current tab is the graph~~
- ~~WIP: switch navigation behavior, right now selecting the task expands the nested tests/tasks and clicking on the details button opens the file details; the new behaviour will open the file details when clicking on the task and will expand the nested tests/tasks when clicking the details button~~

_New icon tabs_

imagen

_Module graph entry detail_

imagen

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/ui/client/components/FileDetails.vue
+++ b/packages/ui/client/components/FileDetails.vue
@@ -1,28 +1,27 @@
 <script setup lang="ts">
-import type { ModuleGraphData } from 'vitest'
 import { client, current, currentLogs, isReport, browserState, config } from '~/composables/client'
 import type { Params } from '~/composables/params'
 import { viewMode } from '~/composables/params'
 import type { ModuleGraph } from '~/composables/module-graph'
 import { getModuleGraph } from '~/composables/module-graph'
-import { getProjectNameColor } from '~/utils/task';
+import { getProjectNameColor } from '~/utils/task'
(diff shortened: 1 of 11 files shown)


sheremet-va:
> the new behaviour will open the file details when clicking on the task and will expand the nested tests/tasks when clicking the details button
 

This doesn't make any sense though? The current behaviour is correct and follows the same idea that we have in vscode extension

sheremet-va (approved):

sheremet-va:
LGTM 👍🏻

Issue #5859, opened 2024-06-07
docs: clarify need to await `expect.toMatchFileSnapshot`


benmccann:
### Clear and concise description of the problem

Most expects don't require that you await them. That means that it's frequently overlooked with `toMatchFileSnapshot`. Even in vitest's own code it's not always awaited: https://github.com/vitest-dev/vitest/blob/07876b7e0f4f69cd071f0c89107ebdd08927e182/test/core/test/snapshot-file.test.ts#L16

### Suggested solution

We should probably fix the test file linked to above. I wonder if there's also some docs update that should be added to explain what will happen if you forget to await it

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

sheremet-va:
We record the promise, so it's not required.

benmccann:
In that case the docs could be updated since they say "you need to use await": https://vitest.dev/api/expect.html#tomatchfilesnapshot

And if it's not required then perhaps it could just be synchronous to avoid the confusion?

sheremet-va:
It does return a Promise, so I don't know what else to tell you. It cannot be sync because to read a file in the browser we need to make an async request. In node.js it could technically be sync, but then the API would differ which is undesirable.

benmccann:
Could it just not return the promise?
```
function toMatchFileSnapshot() {
  _toMatchFileSnapshotInternal();
}
```

PR #5860, opened 2024-06-07
fix(test): add missing `await`

benmccann:
### Description

https://vitest.dev/api/expect.html#tomatchfilesnapshot says:

> Note that since file system operation is async, you need to use await with toMatchFileSnapshot().

However, that wasn't being done here. It might be nice for the docs to expand upon what could happen if you don't `await`. E.g. could it report that the test is successful when the assertion would otherwise fail or is it not really necessary?

### Tests

### Documentation

### Changesets


benmccann:
Maybe this doesn't need to be done though: https://github.com/vitest-dev/vitest/issues/5859#issuecomment-2155344155

PR #5875, opened 2024-06-11
fix: vi.waitFor/vi.waitUntil interval is now cleared after it times out

pedro00dk:
### Description

When using `vi.waitFor` in browser mode, an unsuccessful resolution of its callback will cause the wait for to run indefinitely.
A simple test like the following will cause the issue:

```ts
it.only('hangs', async () => {
    await vi.waitFor(
        () => {
            console.log('here', performance.now())
            throw new Error()
        },
        {
            interval: 100,
            timeout: 500,
        },
    )
})
```

I put some console logs in the waitFor code while debugging and it produced this:
Screenshot from 2024-06-11 15-41-46

After the handler timeout call, the check callback continues running because the interval is not cleared.
This PR just adds a `clearInterval` call to the `handleTimeout` function, just like in `onResolve`.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/integrations/wait.ts
+++ b/packages/vitest/src/integrations/wait.ts
@@ -115,6 +115,8 @@ export function waitUntil<T>(callback: WaitUntilCallback<T>, options: number | W
     let intervalId: ReturnType<typeof setInterval>
 
     const onReject = (error?: Error) => {
+      if (intervalId)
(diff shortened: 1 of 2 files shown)


sheremet-va (approved):

sheremet-va:
Thank you!

PR #5922, opened 2024-06-19
docs: change 'vitest' to '@vitest/expect' in the "Extending Matchers" example

Flixbox:
### Description

The ambient declaration must modify @vitest/expect instead of vitest. Otherwise the type override won't work. This change appears to be necessary since v1.6.0 or earlier.

### Tests

### Documentation

### Changesets


sheremet-va:
I don't think this is true. Did you check that your installed vitest packages have the same version?

Flixbox:
I'll set up a stackblitz to verify.

Flixbox:
You're right, the types compile perfectly.

https://stackblitz.com/~/github.com/Flixbox/vitest-demo?file=src/vitest.types.ts

Seems like I'll have some fixing to do. Thanks for the hint!

Issue #5913, opened 2024-06-18
The Node process exits via Node APIs when no test files found


Jinjiang:
### Describe the bug

When run Vitest via Node APIs, the Node process will exit when no test files found. What I expect is returning an error object or a special message instead (without the exit of the whole Node process).

### Reproduction

https://github.com/Jinjiang/reproductions/tree/vitest-no-test-found-20240618

### System Info

```shell
System:
    OS: macOS 15.0
    CPU: (8) arm64 Apple M1
    Memory: 75.66 MB / 8.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.11.1 - ~/.local/share/mise/installs/node/20/bin/node
    Yarn: 1.22.18 - ~/.yarn/bin/yarn
    npm: 10.2.4 - ~/.local/share/mise/installs/node/20/bin/npm
    pnpm: 9.0.6 - ~/Library/pnpm/pnpm
    bun: 1.0.29 - ~/.local/share/mise/installs/bun/latest/bin/bun
  Browsers:
    Safari: 18.0
  npmPackages:
    vite: ^5.3.1 => 5.3.1 
    vitest: ^1.6.0 => 1.6.0
```

### Used Package Manager

pnpm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

AriPerkkio:
The `process.exit` is coming from here:

https://github.com/vitest-dev/vitest/blob/f969fb0f9f0247a7daa2afee8f70de25ea5e123f/packages/vitest/src/node/core.ts#L485-L497

I think we should lift all `process.exit` calls above the `startVitest()` method.

PR #5926, opened 2024-06-19
fix(api): don't call process.exit manually

sheremet-va:
### Description

 fixes https://github.com/vitest-dev/vitest/issues/5913

Vitest API will now throw an error (`TestsNotFoundError`) when calling `start` manually. Vitest will also throw `GitNotFoundError` error if `changed` flag was used but there is no git repository available.

Both of these errors can be imported from `vitest/node`.

### Tests

### Documentation

### Changesets


sheremet-va:
Should also remove here:

https://github.com/vitest-dev/vitest/blob/67d6adde931b3704219a66e465e9823f55efe7a8/test/test-utils/index.ts#L19

packages/vitest/src/node/core.ts
```diff
@@ -492,7 +492,8 @@ export class Vitest {
 
       if (!this.config.watch || !(this.config.changed || this.config.related?.length)) {
         const exitCode = this.config.passWithNoTests ? 0 : 1
-        process.exit(exitCode)
+        process.exitCode = exitCode
+        throw new NoTestsFoundError()
```
AriPerkkio, line 496: This fixes https://github.com/vitest-dev/vitest/issues/5913

PR #5948, opened 2024-06-20
feat(ui): add action to explorer item to show the test/suite line in the source code tab

### LINKED ISSUE #5931: The html reporter allows viewing the current use case code

image

I hope that when I click on a use case, I can view the test code of the use case. It can bring the following benefits:

- It is equivalent to a usage example for users to refer to.
- It is equivalent to a code snippet, which helps developers quickly locate problems.
- Use cases are usage examples, which reduces the work of repeatedly writing sample code.

### Suggested solution

Suppose there is a use case `a`. When you click on the use case `a`, there is a parameter in the url, such as `?it=a`, which is used to mark the current use case. And you can view its corresponding code on the right.

I can paste the url link to other places, such as the usage document. When others click on this url, they can view the relevant information of the corresponding use case.

Even simply displaying the toString result of the test case function (ignoring other functions referenced within the function) is sufficient for participating in the example.

### Alternative

_No response_

### Additional context

Its functionality is similar to [mochawesome](https://www.npmjs.com/package/mochawesome).

Test case code:

image

UI test results:

image

### Validations

userquin:
### Description

Closes #5931

This PR includes:
- new action in the explorer item to go to the test/suite line in the source code tab: will navigate to the tab if the ViewEditor is not the code tab or the active file is not the selected
- enable `includeTaskLocation` if running in browser mode (not headless) or ui mode: will preserve user option - @sheremet-va review `config.ts` module 🙏 
- docs: update `config/index.md` to include previous entry
- when the user disables  `includeTaskLocation`, the explorer item will show a tooltip similar for coverage without html reporter (check image below): we can just disable the button...
- refactor codemirror logic: moved codemirrror shallow ref from the ViewEditor to a global shallow ref inside codemirror composable, removed exposed codemirror from the ViewEditor and updated all usages with the global one
- added `navigateTo` to `navigator.ts` composable to allow change the view mode: navigator sfc and explorer item using it, without view mode and with it respectively
- added disabled `run` test/suite for future usage: we'll need to change the logic to allow run individual tests/suites.

You can see this PR in action in [this video](https://streamable.com/2txqji).

imagen
_Disabled includeTaskLocation_

### Tests

### Documentation

### Changesets


userquin:
When running `test/core`  with Vitest UI I get this error that should be fixed (or maybe not):
imagen

packages/vitest/src/node/config.ts
```diff
@@ -699,6 +699,16 @@ export function resolveConfig(
...
 
+  // enable includeTaskLocation by default in UI mode
+  if (resolved.browser.enabled) {
+    if (resolved.browser.ui) {
+      resolved.includeTaskLocation ??= true
+    }
+  }
+  else if (resolved.ui) {
+    resolved.includeTaskLocation ??= true
+  }
```
sheremet-va, line 710: Does this work for `--reporter=html`?

userquin: Won't work since source tab is missing when running `pnpm vite preview --outDir=html`, maybe we should add the tab in `readonly` mode

userquin:
Added code tab and source files to the gz file: last commit also enables `includeTaskLocation` when using `html` reporter

packages/ui/client/components/views/ViewEditor.vue
```diff
@@ -24,7 +25,9 @@ watch(
       draft.value = false
       return
     }
-    code.value = (await client.rpc.readTestFile(props.file.filepath)) || ''
+    code.value = isReport
+      ? ('source' in props.file ? props.file.source : '') as string
+      : (await client.rpc.readTestFile(props.file.filepath)) || ''
```
sheremet-va, line 30: Let's not make our lives harder by having `isReport` everywhere. Just make `readTestFile` return the source

packages/ui/node/reporter.ts
```diff
@@ -60,7 +64,16 @@ export default class HTMLReporter implements Reporter {
   async onFinished() {
     const result: HTMLReportData = {
       paths: this.ctx.state.getPaths(),
-      files: this.ctx.state.getFiles(),
+      files: await Promise.all(this.ctx.state.getFiles().map(async (file) => {
```
sheremet-va, line 67: Let's just store a separate field with test files content, and return the content from `readTestFile` fake RPC

sheremet-va (changes requested):

packages/ui/node/reporter.ts
userquin: done

sheremet-va (approved):

userquin:
Dont merge yet, I'll change the explorer ítem ui again (missing tooltip when html reporter).

wll8:
@userquin Is codemirrorRef **guaranteed** to be initialized after 256 milliseconds?

>    // we need to await, CodeMirrow will take some time to initialize
>   await new Promise(r => setTimeout(r, 256))

[packages/ui/client/composables/codemirror.ts](https://github.com/vitest-dev/vitest/pull/5948/files/4a0257e2d89e2121885ba37f2f464d4cc673e973#diff-0b170bb5c42a5f26bf5af5f56613fd5f70ea4874409405d51864b2e5c184f3baR67-R69)

userquin:
It is about waiting for route change and 100ms in the onMounted, maybe I can try using another approach to load it.

userquin:
imagen

packages/ui/client/auto-imports.d.ts
```diff
@@ -108,6 +109,7 @@ declare global {
   const refDefault: typeof import('@vueuse/core')['refDefault']
   const refThrottled: typeof import('@vueuse/core')['refThrottled']
   const refWithControl: typeof import('@vueuse/core')['refWithControl']
+  const registerInitializeListener: typeof import('./composables/codemirror')['registerInitializeListener']
```
userquin, line 112: remove this, no idea why not being removed, every time I add some composable I need to remove it if not being used and removed

userquin:
> @userquin Is codemirrorRef **guaranteed** to be initialized after 256 milliseconds?
> 
> > // we need to await, CodeMirrow will take some time to initialize
> > await new Promise(r => setTimeout(r, 256))
> 
> [packages/ui/client/composables/codemirror.ts](https://github.com/vitest-dev/vitest/pull/5948/files/4a0257e2d89e2121885ba37f2f464d4cc673e973#diff-0b170bb5c42a5f26bf5af5f56613fd5f70ea4874409405d51864b2e5c184f3baR67-R69)

Now the logic will focus the line based on the new line router param, forcing the watcher to trigger in ViewEditor once the new code loaded

sheremet-va (approved):

packages/ui/client/components/explorer/ExplorerItem.vue
```diff
@@ -146,29 +176,46 @@ const highlighted = computed(() => {
...
+        </template>
+      </VueTooltip>
+      <IconButton
         v-if="!isReport"
         v-tooltip.bottom="'Run current test'"
         data-testid="btn-run-test"
         title="Run current test"
         icon="i-carbon:play-filled-alt"
         text-green5
+        :disabled="type !== 'file'"
```
maxpatiiuk, line 218: > added disabled run test/suite for future usage: we'll need to change the logic to allow run individual tests/suites.

Sorry if I am late, but could you please clarify why the ability to run individual test was disabled?

Is is something that is coming in the future? Is there a feature request for that or should I open one?

Thanks!

sheremet-va: This was never supported. It is possible to support - feel free to create an issue or a PR

PR #5972, opened 2024-06-25
fix: use TTY reporter when running in Deno

marvinhagemeister:
### Description

Over at Deno we've been constantly improving our Node compatibility layer and are trying to get `vitest` to work natively. We noticed that the test reporter looks different compared to Node.

Initially, we thought this would be a bug our compatibility layer, but it turns out that `vitest` has a check for Deno that prevents it from using the TTY mode in the reporter.

Note, that there is still more work on our side to be done to get `vitest` fully running. But one step at a time ;)

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/utils/env.ts
+++ b/packages/vitest/src/utils/env.ts
@@ -1,7 +1,6 @@
 export const isNode: boolean
   = typeof process < 'u'
   && typeof process.stdout < 'u'
-  && !process.versions?.deno
   && !globalThis.window
 export const isWindows = isNode && process.platform === 'win32'
 export const isBrowser: boolean = typeof window !== 'undefined'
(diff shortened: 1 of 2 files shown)


packages/vitest/src/utils/env.ts
```diff
@@ -1,7 +1,6 @@
 export const isNode: boolean
   = typeof process < 'u'
   && typeof process.stdout < 'u'
-  && !process.versions?.deno
```
sheremet-va, line 4: I am not sure this makes sense because the variable is named `isNode`. Maybe we should add `isDeno` variable instead? Or change the condition that relies on this variable to be less runtime-specific?

sheremet-va (approved):

packages/vitest/src/utils/env.ts
marvinhagemeister: updated the PR accordingly

Issue #5977, opened 2024-06-26
iTerm2 clear screen does not clear the screen properly


kxalex:
### Describe the bug

iTerm2 does not clear the screen properly when vitest is running in watch mode.

### Reproduction

Use iTerm2 under Mac.

Run vitest on any project and on re-run not whole screen was cleared, there are some leftovers from previous run.

vitest currently uses to clear the screen and erase scrollback.
```
this.console.log(`${CLEAR_SCREEN}${ERASE_SCROLLBACK}${message}`)
```

iTerm2 reacts properly on that sequence, although clear screen sequence leaves 1 page visible and erase scrollback removes the history leaving this not-cleared page.

By flipping the sequence iTerm2 behaves properly.
```
this.console.log(`${ERASE_SCROLLBACK}${CLEAR_SCREEN}${message}`)
```

I've added changes in this PR against 1.6.0 version. Although I'm still using 1.5.3 due to cloudflare limitations.

https://github.com/vitest-dev/vitest/pull/5978/files

The erase scrollback and clear screen can be tested in shell using
```
printf '\e[3J' && printf '\x1Bc' # does not clear all screen
printf '\x1Bc' && printf '\e[3J' # clears everything
```
This was also tested on default macos terminal and it works fine too.

### System Info

```shell
System:
    OS: macOS 14.5
    CPU: (10) arm64 Apple M2 Pro
    Memory: 65.22 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.2.0 - /opt/homebrew/bin/node
    npm: 10.7.0 - /opt/homebrew/bin/npm
    pnpm: 9.4.0 - /opt/homebrew/bin/pnpm
  Browsers:
    Brave Browser: 125.1.66.118
    Chrome: 126.0.6478.116
    Safari: 17.5
  npmPackages:
    @vitest/monorepo:  1.5.3
```

### Used Package Manager

pnpm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
[...]

PR #5978, opened 2024-06-26
fix: flipped erase scrollback and clear screen to fix the issue with iTerm2 refresh

kxalex:
### Description

iTerm2 does not fully clear the scrollback history and the screen. The problem that clearing scrollback history removes the history, and leaves current page in the scrollback history. The correct sequence is clear the screen, so it's empty and then the scrollback history leaving the terminal nice and clear. 

Issue: https://github.com/vitest-dev/vitest/issues/5977

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/logger.ts
+++ b/packages/vitest/src/node/logger.ts
@@ -63,7 +63,7 @@ export class Logger {
       return
     }
 
-    this.console.log(`${ERASE_SCROLLBACK}${CLEAR_SCREEN}${message}`)
+    this.console.log(`${CLEAR_SCREEN}${ERASE_SCROLLBACK}${message}`)
   }
 
   clearScreen(message: string, force = false) {


AriPerkkio (commented):
We'll need reproduction case to the issue first. https://github.com/vitest-dev/vitest/issues/5977#issuecomment-2192432713

Test case would be great but as this is terminal specific I think it's impossible here. So reproduction case is required before we can accept this.

AriPerkkio (approved):
Yup, I can verify this fixes the issue. 

Tested terminals:
- iTerm2, this is the only terminal that could reproduce the reported bug
- Mac's original terminal
- VSCode terminal on Mac
- Git Bash on virtualized windows
- Powershell on virtualized windows
- CMD on virtualized windows

PR #5982, opened 2024-06-27
feat(config): allow percentage value for workers option

syi0808:
### Description

I can't use maxWorkers option that jest supported after migrating to vitest.
So I modified it so that the percentage value can be used in maxworkers and minworkers.

https://github.com/jestjs/jest/issues/9012#issue-503013549
Although it is a feature request in jest, I would like to get the same feature in viteest.

### Tests

### Documentation

### Changesets


syi0808:
I want to add test for worker option but I found two problems.
1. Can't mock `node:os` inside bundled `vitest/node`
2. if get resolveConfig function from `package/vitest/node/config.ts` throw type error when running test

I don't think I can test the code properly if I can't mock `node:os`. Is it okay to test with cpus count without mocking?
Or is it okay to test only exception handling other than calculate logic?

sheremet-va:
> I want to add test for worker option but I found two problems.
> 
> 1. Can't mock `node:os` inside bundled `vitest/node`
> 2. If i import resolveConfig function from `package/vitest/node/config.ts` to test, throw type error from runtime.
> 
> I don't think I can test the code properly if I can't mock `node:os`. Is it okay to test with cpus count without mocking? Or is it okay to test only exception handling other than calculate logic?

You can create a separate utility that accepts the workers number and the amount of CPU and test that function.

syi0808:
> > I want to add test for worker option but I found two problems.
> > 
> > 1. Can't mock `node:os` inside bundled `vitest/node`
> > 2. If i import resolveConfig function from `package/vitest/node/config.ts` to test, throw type error from runtime.
> > 
> > I don't think I can test the code properly if I can't mock `node:os`. Is it okay to test with cpus count without mocking? Or is it okay to test only exception handling other than calculate logic?
> 
> You can create a separate utility that accepts the workers number and the amount of CPU and test that function.

Thanks for nice idea! I separate getWorkers function to utility and node:os mock is works. I added test now.

test/config/test/workers-option.test.ts
```diff
@@ -0,0 +1,33 @@
...
+}
+
+test('workers percent argument should not throw error', async () => {
+  const { stderr } = await runVitest({ maxWorkers: '100%', minWorkers: '10%' })
+
+  expect(stderr).toBe('')
+})
+
+describe('workers util', () => {
+  vi.mock('node:os', async importOriginal => ({
```
sheremet-va, line 17: `vi.mock `is always hoisted to the top of the file so it doesn't make a sense to have it inside `describe` function, let's put it at the top to make it less confusing

syi0808: Now that I look at it, I wonder how `vi.mock` worked. It's reflected.

sheremet-va (approved):

docs/guide/cli-table.md
```diff
@@ -80,8 +80,8 @@
 | `--poolOptions.vmForks.minForks <workers>` | Minimum number of processes to run tests in |
```
sheremet-va, line 80: We also have these options for every pool. I think they should all inherit the same behaviour then

sheremet-va (changes requested):

docs/guide/cli-table.md
syi0808: I implement percentage option every pool options and now all of workflows passed!

sheremet-va (changes requested):

packages/vitest/src/node/config.ts
```diff
@@ -436,6 +446,48 @@ export function resolveConfig(
...
+    ['vmThreads', 'minThreads'],
+    ['vmThreads', 'maxThreads'],
+  ] as const satisfies [keyof PoolOptions, keyof ThreadsOptions][]
+
+  for (const [poolOptionKey, workerOptionKey] of poolThreadsOptions) {
+    const workerInlineOption = resolved.poolOptions?.[poolOptionKey]?.[workerOptionKey]
+
+    if (workerInlineOption) {
+      resolved.poolOptions = {
+        ...resolved.poolOptions,
```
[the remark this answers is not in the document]
syi0808: I think much better. done!

packages/vitest/src/utils/workers.ts
```diff
@@ -0,0 +1,8 @@
+import { cpus } from 'node:os'
+
+export function getWorkersCountByPercentage(percent: string) {
+  const maxWorkersCount = cpus().length
```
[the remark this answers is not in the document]
syi0808: oh, I didn't know that function existed. done!

sheremet-va (approved):

syi0808:
test failed.. because maybe i changed node:os import method. I fix test in a minutes.

packages/vitest/src/utils/workers.ts
```diff
@@ -0,0 +1,8 @@
+import { availableParallelism, cpus } from 'node:os'
+
+export function getWorkersCountByPercentage(percent: string) {
+  const maxWorkersCount = availableParallelism?.() ?? cpus().length
```
sheremet-va, line 4: This needs to be on the default export otherwise Node.js will fail at runtime

syi0808: I see. I will change node:os mock code.

syi0808: I mock default export of node:os.

sheremet-va (approved):

sheremet-va:
LGTM! Thank you! 🎉

PR #5984, opened 2024-06-27
feat(browser): playwright provider doesn't allow resizing the browser viewport

userquin:
### Description

Playwright using custom viewport implementation, once the browser is opened PW will prevent resizing the Vitest UI.

We need to configure `viewport: null` when creating the context, for context check: https://github.com/microsoft/playwright/issues/20721

_Playwright prevents resizing the viewport_
imagen

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/browser/src/node/providers/playwright.ts
+++ b/packages/browser/src/node/providers/playwright.ts
@@ -1,6 +1,5 @@
 import type {
   Browser,
-
   BrowserContext,
   BrowserContextOptions,
   Frame,
@@ -67,10 +66,21 @@ export class PlaywrightBrowserProvider implements BrowserProvider {
 
       const playwright = await import('playwright')
 
-      const browser = await playwright[this.browserName].launch({
+      const launchOptions = {
         ...this.options?.launch,
         headless: options.headless,
-      })
+      } satisfies LaunchOptions
+
+      if (this.ctx.config.browser.launchViewport.maximized) {
+        if (!launchOptions.args) {
+          launchOptions.args = []
+        }
+        if (!launchOptions.args.includes('--start-maximized') && !launchOptions.args.includes('--start-fullscreen')) {
+          launchOptions.args.push('--start-maximized')
+        }
+      }
+
+      const browser = await playwright[this.browserName].launch(launchOptions)
       this.browser = browser
       this.browserPromise = null
       return this.browser
@@ -85,12 +95,15 @@ export class PlaywrightBrowserProvider implements BrowserProvider {
     }
 
     const browser = await this.openBrowser()
-    const context = await browser.newContext({
+    const options = {
       ...this.options?.context,
       ignoreHTTPSErrors: true,
       serviceWorkers: 'allow',
-      viewport: null,
-    })
+    } satisfies BrowserContextOptions
+    if (this.ctx.config.browser.launchViewport.maximized) {
+      options.viewport = null
+    }
+    const context = await browser.newContext(options)
     this.contexts.set(contextId, context)
     return context
   }
--- a/packages/browser/src/node/providers/playwright.ts
+++ b/packages/browser/src/node/providers/playwright.ts
@@ -71,7 +71,7 @@ export class PlaywrightBrowserProvider implements BrowserProvider {
         headless: options.headless,
       } satisfies LaunchOptions
 
-      if (this.ctx.config.browser.launchViewport.maximized) {
+      if (this.ctx.config.browser.ui) {
         if (!launchOptions.args) {
           launchOptions.args = []
         }
@@ -100,7 +100,7 @@ export class PlaywrightBrowserProvider implements BrowserProvider {
       ignoreHTTPSErrors: true,
       serviceWorkers: 'allow',
     } satisfies BrowserContextOptions
-    if (this.ctx.config.browser.launchViewport.maximized) {
+    if (this.ctx.config.browser.ui) {
       options.viewport = null
     }
     const context = await browser.newContext(options)
--- a/packages/browser/src/node/providers/playwright.ts
+++ b/packages/browser/src/node/providers/playwright.ts
@@ -71,7 +71,8 @@ export class PlaywrightBrowserProvider implements BrowserProvider {
         headless: options.headless,
       } satisfies LaunchOptions
 
-      if (this.ctx.config.browser.ui) {
+      // start Vitest UI maximized only on supported browsers
+      if (this.ctx.config.browser.ui && this.browserName === 'chromium') {
         if (!launchOptions.args) {
           launchOptions.args = []
         }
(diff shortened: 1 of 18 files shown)


userquin:
@sheremet-va we can also open it maximized via args if you want to add it here: https://discord.com/channels/807756831384403968/1255831070339891281/1255867837793439857

```ts
launchOptions: { args: ["--start-maximized"] }
```

kwojcik:
Is this the same problem as what causes my failure screenshots to get cutoff? This screenshot is taken with `viewport: { width: 800, height: 1500 }`
screenshot

sheremet-va:
> we can also open it maximized via args if you want to add it here

Maybe we should open it in max if `--browser.ui` is true, and pass down the same viewport  from the config otherwise

userquin:
> > we can also open it maximized via args if you want to add it here
> 
> Maybe we should open it in max if `--browser.ui` is true, and pass down the same viewport from the config otherwise

Added `launchViewport` to resolved browser options: Vitest UI maximized when ui enabled with PW and WDIO providers.

For WDIO I haven't found a way to maximize the window on `remove` call, using [browser.maximizeWindow](https://webdriver.io/docs/api/webdriver/#maximizewindow), on the browser opens , the window is maximized.

For PW, adding `--start-maximized` arg to `launch` (when missing `--start-maximized` and `--start-fullscreen` from args) just works: Vitest UI maximized when the browser is opened => https://peter.sh/experiments/chromium-command-line-switches/

sheremet-va:
Let’s not add any new options. Args can already be overriden by providerOptions

packages/browser/src/node/providers/playwright.ts
```diff
@@ -85,11 +95,15 @@ export class PlaywrightBrowserProvider implements BrowserProvider {
...
 
     const browser = await this.openBrowser()
-    const context = await browser.newContext({
+    const options = {
       ...this.options?.context,
       ignoreHTTPSErrors: true,
       serviceWorkers: 'allow',
-    })
+    } satisfies BrowserContextOptions
+    if (this.ctx.config.browser.launchViewport.maximized) {
```
sheremet-va, line 103: What I _was_ proposing is using the `browser.viewport` values for non-ui instances

userquin: we cannot use that entry, that viewport is about the iframe not the Vitest UI (top context)

userquin: how can we display Vitest UI in mobile viewport? no makes sense

sheremet-va: > for non-ui instances

When UI is disabled

userquin: when ui is disabled this PR doesn't modify the options, and so the options are passed to the provider (I'll check it again, PW for sure in the `newContext` call => `createContext` in playwright.ts module)

userquin: we need to review the logic to ensure the iframe is "visible", if we use the viewport size in headless mode, the iframe maybe not being "displayed" properly

userquin: sorry Vlad, forgot it...

userquin: adding viewport size when ui disabled...

userquin: too many things :sweat_smile:

userquin: done, included also a new `viewport.test.ts`

test/browser/test/viewport.test.ts
```diff
@@ -0,0 +1,26 @@
...
+    const { width: actualWidth, height: actualHeight } = window.document.documentElement.getBoundingClientRect()
+
+    // eslint-disable-next-line no-console
+    console.log({ width, height, actualWidth, actualHeight })
+
+    expect(actualWidth).toBe(width)
+    expect(actualHeight).toBe(height)
+  })
+
+  it.skipIf(page.config.browser.headless)('window has been maximized', async () => {
```
userquin, line 16: remove async

test/browser/test/viewport.test.ts
```diff
@@ -0,0 +1,26 @@
+import { page, server } from '@vitest/browser/context'
+import { describe, expect, it } from 'vitest'
+
+describe.skipIf(server.provider === 'preview')('viewport window has been properly initialized', () => {
+  it.skipIf(!page.config.browser.headless)('viewport has proper width and size', () => {
```
userquin, line 5: change this test: `'viewport has proper size' or `'viewport has proper width and height'

userquin:
We can use browser capabilities, but will differ between browsers (note from @christian-bromann ):
```ts
capabilities: {
  browserName: 'chrome',
  'goog:chromeOptions': {
    args: ['--start-maximized', '--start-fullscreen']
  }
}
```

packages/browser/src/node/providers/playwright.ts
userquin: tested with WDIO with chrome, edge and firefox with and without headless in my local, cannot test safari...

test/browser/test/viewport.test.ts
```diff
@@ -0,0 +1,36 @@
+import { page, server } from '@vitest/browser/context'
+import { describe, expect, it } from 'vitest'
+
+describe.skipIf(server.provider === 'preview')('viewport window has been properly initialized', () => {
+  it.skipIf(!page.config.browser.headless)('viewport has proper size', () => {
+    const { width, height } = page.config.browser.viewport
+    const { width: actualWidth, height: actualHeight } = window.document.documentElement.getBoundingClientRect()
+
+    // eslint-disable-next-line no-console
+    console.log({ width, height, actualWidth, actualHeight })
```
sheremet-va, line 10: why do you need logs here?

packages/browser/src/node/providers/webdriver.ts
```diff
@@ -80,6 +80,14 @@ export class WebdriverBrowserProvider implements BrowserProvider {
       capabilities: this.buildCapabilities(),
     })
 
+    if (this.ctx.config.browser.ui) {
+      await this.browser.maximizeWindow()
+    }
+    else {
+      const { width, height } = this.ctx.config.browser.viewport
+      await this.browser.setWindowSize(width, height)
```
sheremet-va, line 88: You do it here by not in playwright, is this intentional? Do you think we should keep viewport only for the iframe?

sheremet-va (changes requested):

packages/browser/src/node/providers/webdriver.ts
userquin: There is no need (intentional), PW will use the viewport size properly from the options, WDIO will ignore them.

test/browser/test/viewport.test.ts
userquin: removing the logs...

userquin: done

packages/browser/src/node/providers/webdriver.ts
userquin: > Do you think we should keep viewport only for the iframe?

Yes, why not?

userquin:
With latest change, the `viewport.test.ts` will not pass when running it with Vitest UI using Firefox or Safari (with both providers).

sheremet-va (approved):

Issue #5965, opened 2024-06-24
`coverage.exclude` should be fully configurable


AriPerkkio:
### Describe the bug

- Discussed in https://github.com/vitest-dev/vitest/discussions/5964

Following configuration should overwrite all default values of `coverage.exclude`. It should not exclude `dist` directory:

```ts
defineConfig({
  test: {
    // Only node_modules should be excluded, nothing else
    coverage: { exclude: ['**/node_modules/**'] },
  }
})
```

Actual: Hard coded default values are still applied. `dist` is excluded no matter what.

### Reproduction

https://github.com/donmccurdy/2024-06-24_vitest_coverage_issue

### System Info

```shell
Any
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #5997, opened 2024-06-28
fix(coverage): support overriding `exclude`

AriPerkkio:
### Description

- Fixes https://github.com/vitest-dev/vitest/issues/5965

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/defaults.ts
+++ b/packages/vitest/src/defaults.ts
@@ -29,6 +29,7 @@ export const benchmarkConfigDefaults: Required<
 const defaultCoverageExcludes = [
   'coverage/**',
   'dist/**',
+  '**/node_modules/**',
   '**/[.]**',
   'packages/*/test?(s)/**',
   '**/*.d.ts',
@@ -40,7 +41,7 @@ const defaultCoverageExcludes = [
   'test?(-*).?(c|m)[jt]s?(x)',
   '**/*{.,-}{test,spec,bench,benchmark}?(-d).?(c|m)[jt]s?(x)',
   '**/__tests__/**',
-  '**/{karma,rollup,webpack,vite,vitest,jest,ava,babel,nyc,cypress,tsup,build}.config.*',
+  '**/{karma,rollup,webpack,vite,vitest,jest,ava,babel,nyc,cypress,tsup,build,eslint,prettier}.config.*',
   '**/vitest.{workspace,projects}.[jt]s?(on)',
   '**/.{eslint,mocha,prettier}rc.{?(c|m)js,yml}',
 ]
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -32,6 +32,10 @@ All configuration options that are not supported inside a [workspace](/guide/wor
 
 A list of glob patterns that match your test files.
 
+::: tip NOTE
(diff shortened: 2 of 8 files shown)


packages/vitest/src/defaults.ts
```diff
@@ -43,6 +43,8 @@ const defaultCoverageExcludes = [
   '**/{karma,rollup,webpack,vite,vitest,jest,ava,babel,nyc,cypress,tsup,build}.config.*',
   '**/vitest.{workspace,projects}.[jt]s?(on)',
   '**/.{eslint,mocha,prettier}rc.{?(c|m)js,yml}',
+  ...defaultExclude,
+  ...defaultInclude,
```
sheremet-va, line 47: Shouldn't it include user defined include/exclude? Somewhere here: https://github.com/vitest-dev/vitest/blob/d7f23d08c4a5420bb0b6483c7e3dc87a1ae8eeaf/packages/vitest/src/node/config.ts#L510

AriPerkkio: I'm not sure about that. I would rather keep `test.include` and `test.coverage.include` runtime values separately. Less automation to make things simple. I don't think Jest that does that either. 

Though I don't have strong opinion about this.

sheremet-va: I mean what is the reason to merge them then?

AriPerkkio: Current coverage default excludes seem to include most of the patterns already. Let's just remove the merging here.

AriPerkkio: But overall: I would like to avoid situation where configuring `test.inlcude/exclude` magically affects `test.coverage.include/exclude` in cases where user doesn't configure `test.coverage.include/exclude` in the first place.

sheremet-va: Yeah I am fine with removing it or adding the configured include/exclude only to non-configured `coverage.exclude`, but before we make a decision we need to find the reason why we even had that at the first place. I think it makes sense to remove any test file from coverage by default, don't you?

AriPerkkio: Sure, we could do that. It's not really related to https://github.com/vitest-dev/vitest/issues/5965 though, as that one is only about being able to overwrite the defaults.

If we automatically added user's `test.include/exclude´ to `test.coverage.include/exclude´, would it then be impossible to fully overwrite `test.coverage.exclude´? Or would we only apply `test.include/exclude´ there if coverage ones were not set by user?

sheremet-va: The last one. Only apply if there is no exclude already

AriPerkkio: > Somewhere here:
> 
> [vitest/packages/vitest/src/node/config.ts](https://github.com/vitest-dev/vitest/blob/d7f23d08c4a5420bb0b6483c7e3dc87a1ae8eeaf/packages/vitest/src/node/config.ts#L510)

At this point the default values for coverage have already been applied, probably by `src/node/plugins/index.ts`.

sheremet-va: Can we do it in plugins/index then? 😄

AriPerkkio: Added it into plugins. There's already some configuration resolving so it's not that bad to add some more 😄

PR #6041, opened 2024-07-05
fix(vite-node): Generate sourcemap for inlined deps

### LINKED ISSUE #5605: Source maps in inlined deps are off by a few lines in VS Code
### Describe the bug

Source maps in **inlined** deps are off by a few lines in VS Code debugger

### Reproduction

https://github.com/bgoscinski/repro-vitest-sourcemap

1. `pnpm install`
1. `code test/the.test.ts`
1. Set breakpoint at line 3 of the test file
1. Start debugging session using 'Debug' configuration
1. Wait for debugger to attach
1. Press 'Step into' until you're in `add` function implementation
1. Press 'Step over' and see that debugger highlights invalid line

https://github.com/vitest-dev/vitest/assets/5730169/4d1b9170-5ca4-4bcd-be5e-f3c3166a1b8e

### System Info

```shell
System:
    OS: Windows 11 10.0.22000
    CPU: (16) x64 12th Gen Intel(R) Core(TM) i7-1260P
    Memory: 19.73 GB / 47.70 GB
  Binaries:
    Node: 18.16.0 - C:\Program Files\nodejs\node.EXE
    npm: 9.8.1 - C:\Program Files\nodejs\npm.CMD
    pnpm: 8.15.6 - ~\AppData\Roaming\npm\pnpm.CMD
    bun: 1.1.1 - ~\.bun\bin\bun.EXE
  Browsers:
    Edge: Spartan (44.22000.120.0), Chromium (123.0.2420.97)
  npmPackages:
    @vitest/ui: latest => 1.5.0
    vite: latest => 5.2.10
    vitest: latest => 1.5.0
```

### Used Package Manager

pnpm

### Validations

bgoscinski:
closes #5605

I implemented the change that allows to use the source maps but I could really use some guidance on if/how you want this tested

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vite-node/src/server.ts
+++ b/packages/vite-node/src/server.ts
@@ -412,7 +412,7 @@ export class ViteNodeServer {
     }
 
     const sourcemap = this.options.sourcemap ?? 'inline'
-    if (sourcemap === 'inline' && result && !id.includes('node_modules')) {
+    if (sourcemap === 'inline' && result && !(await this.shouldExternalize(id))) {
       result = await this.processTransformResult(filepath, result)
     }
 


packages/vite-node/src/server.ts
```diff
@@ -412,7 +412,7 @@ export class ViteNodeServer {
     }
 
     const sourcemap = this.options.sourcemap ?? 'inline'
-    if (sourcemap === 'inline' && result && !id.includes('node_modules')) {
+    if (sourcemap === 'inline' && result && !(await this.shouldExternalize(id))) {
```
sheremet-va, line 415: This doesn't make much sense because it will alway return `true` - this function is not called otherwise:

https://github.com/vitest-dev/vitest/blob/51319929e5e25069ce57f15f8a4cdccb314b65ad/packages/vite-node/src/server.ts#L347-L358

We don't inline source map because it's expensive to do for every file in node modules

sheremet-va (changes requested):

packages/vite-node/src/server.ts
bgoscinski: Is there a way for me as a vitest user force it to generate source maps for specific modules from node_modules? In my case I'd like to generate source maps for the deps that I inline (`server.deps.inline`) because of #5605

sheremet-va: No, there is no way to do that, all `node_modules` files ignore stack trace. Maybe we can add another option 🤷🏻

bgoscinski: A new option in the config would solve my usecase 💯!

If it makes sense to allow users to generate source maps for not necessairly inlined deps then `server.deps.sourcemap` accepting `(string | RegExp)[]` seem ok, but if it makes sense only for inline deps then how about extending `server.deps.inline` like this
```diff
-inline: (string | RegExp)[] | true
+inline: (string | RegExp)[] | true | { match: string | RegExp, sourcemap: boolean }[]
```

bgoscinski: Ok thanks!
I'll close this PR then and see if I'll be able to implement the `includeSourcemap` option

Issue #5856, opened 2024-06-07
coverage-istanbul does not work with workspaces and SvelteKit


Ernxst:
### Describe the bug

In a monorepo with a simple SvelteKit app, running coverage with `@vitest/coverage-istanbul` does not work. Tests pass but collecting coverage fails with the following error:

```shell
SyntaxError: /Users/ernest/Projects/Repros/vitest-workspaces-istanbul/packages/web/src/routes/+page.svelte: Support for the experimental syntax 'jsx' isn't currently enabled (1:1):

> 1 | <h1>Welcome to SvelteKit</h1>
    | ^
  2 | <p>Visit <a href="https://kit.svelte.dev">kit.svelte.dev</a> to read the documentation</p>
  3 |

Add @babel/preset-react (https://github.com/babel/babel/tree/main/packages/babel-preset-react) to the 'presets' section of your Babel config to enable transformation.
If you want to leave it as-is, add @babel/plugin-syntax-jsx (https://github.com/babel/babel/tree/main/packages/babel-plugin-syntax-jsx) to the 'plugins' section to enable parsing.

If you already added the plugin for this syntax to your config, it's possible that your config isn't being loaded.
You can re-run Babel with the BABEL_SHOW_CONFIG_FOR environment variable to show the loaded configuration:
        npx cross-env BABEL_SHOW_CONFIG_FOR=/Users/ernest/Projects/Repros/vitest-workspaces-istanbul/packages/web/src/routes/+page.svelte <your build command>
See https://babeljs.io/docs/configuration#print-effective-configs for more info.

1  |  <h1>Welcome to SvelteKit</h1>
   |  ^
2  |  <p>Visit <a href="https://kit.svelte.dev">kit.svelte.dev</a> to read the documentation</p>
3  |  
```

Works fine with `@vitest/coverage-v8` but I need istanbul (required by [`@cloudflare/vitest-pool/workers`](https://www.npmjs.com/package/@cloudflare/vitest-pool-workers))

### Reproduction

1. Clone [repo](https://github.com/Ernxst/vitest-workspaces-istanbul)
2. Install dependencies (`bun install`)
3. Run test script (`bun run test`) - everything passes
4. Run coverage script (`bun run coverage`) 

### System Info

```shell
System:
    OS: macOS 14.5
    CPU: (8) arm64 Apple M1
    Memory: 40.45 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
[...]

PR #6044, opened 2024-07-07
fix(coverage): use project specific `vitenode` for uncovered files

AriPerkkio:
### Description

- Fixes https://github.com/vitest-dev/vitest/issues/5856 

When picking up uncovered files (`coverage.all`) in workspace, try to find `vitenode` from a project whose `root` matches the uncovered file. If none are found, use core's `vitenode`. 

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/coverage-istanbul/src/provider.ts
+++ b/packages/coverage-istanbul/src/provider.ts
@@ -414,13 +414,15 @@ export class IstanbulCoverageProvider extends BaseCoverageProvider implements Co
     const cacheKey = new Date().getTime()
     const coverageMap = libCoverage.createCoverageMap({})
 
+    const transform = this.createUncoveredFileTransformer(this.ctx)
+
     // Note that these cannot be run parallel as synchronous instrumenter.lastFileCoverage
     // returns the coverage of the last transformed file
     for (const [index, filename] of uncoveredFiles.entries()) {
       debug('Uncovered file %s %d/%d', filename, index, uncoveredFiles.length)
 
       // Make sure file is not served from cache so that instrumenter loads up requested file coverage
-      await this.ctx.vitenode.transformRequest(`${filename}?v=${cacheKey}`)
+      await transform(`${filename}?v=${cacheKey}`)
       const lastCoverage = this.instrumenter.lastFileCoverage()
       coverageMap.addFileCoverage(lastCoverage)
     }
--- a/packages/vitest/src/utils/coverage.ts
+++ b/packages/vitest/src/utils/coverage.ts
@@ -1,6 +1,7 @@
 import { relative } from 'pathe'
 import mm from 'micromatch'
 import type { CoverageMap } from 'istanbul-lib-coverage'
+import type { Vitest } from '../node/core'
 import type { BaseCoverageOptions, ResolvedCoverageOptions } from '../node/types/coverage'
 import { resolveCoverageReporters } from '../node/config/resolveConfig'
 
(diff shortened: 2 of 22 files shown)


packages/coverage-istanbul/src/provider.ts
```diff
@@ -396,16 +401,29 @@ export class IstanbulCoverageProvider
...
     // returns the coverage of the last transformed file
     for (const [index, filename] of uncoveredFiles.entries()) {
       debug('Uncovered file %s %d/%d', filename, index, uncoveredFiles.length)
 
-      // Make sure file is not served from cache so that instrumenter loads up requested file coverage
-      await this.ctx.vitenode.transformRequest(`${filename}?v=${cacheKey}`)
+      for (const vitenode of viteNodeServers) {
+        try {
+          await vitenode.transformRequest(`${filename}?v=${cacheKey}`)
+          continue
```
sheremet-va, line 417: do you want to continue or do you want to break here? `continue` doesn't do anything here

sheremet-va (changes requested):

packages/coverage-istanbul/src/provider.ts
```diff
@@ -386,6 +386,11 @@ export class IstanbulCoverageProvider
       resolve(this.ctx.config.root, file),
     )
 
+    const viteNodeServers = [
+      this.ctx.vitenode,
+      ...this.ctx.projects.map(project => project.vitenode),
```
sheremet-va, line 391: Did you consider filtering servers by root first?

packages/coverage-istanbul/src/provider.ts
AriPerkkio: ohh it should be `break` 🤦 
Good catch!

packages/coverage-istanbul/src/provider.ts
AriPerkkio: I had something like that initially: https://github.com/AriPerkkio/vitest/commit/832d0b4a690376bc2d52b796c24b4d2f87409ae0#diff-a0a730c31ce78cf1c3cb20d55daa7167ff57fa8e062df213c6eeb918582d1cb5R282-R303

But sometimes projects can have same root - the current test case failed here. Workspace is used just to switch different configuration options for specific set of files. And there's also `coverage.allowExternal` 🤷

100-proof logic could be implemented if `coverage.include` was allowed in workspace projects. But I'd rather not expose any `coverage` options there.

AriPerkkio: Yep, I'll add this.

packages/vitest/src/utils/coverage.ts
```diff
@@ -1,6 +1,7 @@
 import { relative } from 'pathe'
 import mm from 'micromatch'
 import type { CoverageMap } from 'istanbul-lib-coverage'
+import type { Vitest } from '../node/core'
```
AriPerkkio, line 4: Importing from `vitest/node` here caused a build error 🤔

sheremet-va: you can't import from vitest/* inside the vitest package. they files are not built yet

AriPerkkio: That makes sense! I guess this relative import is fine here then?

packages/vitest/src/utils/coverage.ts
```diff
@@ -272,6 +273,37 @@ export class BaseCoverageProvider {
...
+    return async function transformFile(filename: string) {
+      let lastError
+
+      for (const { root, vitenode } of servers) {
+        if (!filename.startsWith(root)) {
+          continue
+        }
+
+        try {
+          return await vitenode.transformRequest(filename)
```
sheremet-va, line 296: What if the workspace uses browser?

sheremet-va (changes requested):

packages/vitest/src/utils/coverage.ts
AriPerkkio: I think it works the same with vite node here. For uncovered files we don't need to worry about the wrapper length as these are not executed.

packages/vitest/src/utils/coverage.ts
```diff
@@ -272,6 +273,37 @@ export class BaseCoverageProvider {
...
+        }
+        catch (error) {
+          lastError = error
+        }
+      }
+
+      // All vite-node servers failed to transform the file
+      throw lastError
```
sheremet-va, line 304: Hm, what if the file never started with root? Is it possible? Then this will be undefined

AriPerkkio: Oh good point! This might fail with allowExternal files. Ill need to add test case for those.

AriPerkkio: Opened https://github.com/vitest-dev/vitest/pull/6372 as follow-up. I didn't yet look into this a lot, created PR mostly so that I won't forget this issue.

PR #6056, opened 2024-07-08
feat(expect): add `toHaveBeenCalledAfter` and `toHaveBeenCalledBefore` utility

### LINKED ISSUE #6045: Integrate `toHaveBeenCalledBefore` and `toHaveBeenCalledAfter` to core

Hello,

Knowing when a mock is called relatively to another is very useful to ensure implementation order is correct. Actually, these is no expect function to know that. Do you thing it could be possible to add some or have you any recommanded way to check that a mock `vi.fn` is called before or after another one.

### Suggested solution

Integrate https://github.com/jest-community/jest-extended/blob/main/src/matchers/toHaveBeenCalledBefore.js to core of vitest.

### Alternative

_No response_

### Additional context

_No response_

### Validations

Barbapapazes:
### Description

fix #6045

Hello 👋,

This is my first contribution to Vitest so I really appreciate your feedback.

This PR adds two new test APIs:

- `toHaveBeenCalledBefore` - checks if a spy was called before another spy
- `toHaveBeenCalledAfter` - checks if a spy was called after another spy

The implementation will take a look at the invocation order (first call) of the spies and compare them to determine if the spies were called in the correct order.

I add them to `packages/expect/src/jest-expect.ts` but I'm not sure if this is the best place for them. I also update types but I'm not sure if I did it correctly so I would appreciate your feedback on that.

I add related tests to for these new APIs.

Once everything is ok for you, I will add related documentation!

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/expect/src/jest-expect.ts
+++ b/packages/expect/src/jest-expect.ts
@@ -656,17 +656,45 @@ export const JestChaiExpect: ChaiPlugin = (chai, utils) => {
       )
     },
   )
+
+  /**
+   * Used for `toHaveBeenCalledBefore` and `toHaveBeenCalledAfter` to determine if the expected spy was called before the result spy.
+   */
+  function predicate(expectInvocationCallOrder: number[], resultInvocationCallOrder: number[], failIfNoFirstInvocation: number): boolean {
+    if (expectInvocationCallOrder.length === 0) {
+      return !failIfNoFirstInvocation
+    }
+
+    if (resultInvocationCallOrder.length === 0) {
+      return false
+    }
+
+    return expectInvocationCallOrder[0] < resultInvocationCallOrder[0]
+  }
+
   def(
     ['toHaveBeenCalledBefore'],
-    function (resultSpy: MockInstance) {
+    function (resultSpy: MockInstance, failIfNoFirstInvocation = true) {
       const expectSpy = getSpy(this)
 
-      const [firstExpectSpyCall] = expectSpy.mock.invocationCallOrder
+      if (!isMockFunction(resultSpy)) {
+        throw new TypeError(
+          `${utils.inspect(resultSpy)} is not a spy or a call to a spy`,
+        )
+      }
 
-      const [firstResultSpyCall] = resultSpy.mock.invocationCallOrder
+      if (!isMockFunction(expectSpy)) {
+        throw new TypeError(
+          `${utils.inspect(expectSpy)} is not a spy or a call to a spy`,
+        )
+      }
 
       this.assert(
-        firstExpectSpyCall < firstResultSpyCall,
+        predicate(
+          expectSpy.mock.invocationCallOrder,
+          resultSpy.mock.invocationCallOrder,
+          failIfNoFirstInvocation,
+        ),
         `expected "${expectSpy.getMockName()}" to have been called before "${resultSpy.getMockName()}"`,
         `expected "${expectSpy.getMockName()}" to not have been called before "${resultSpy.getMockName()}"`,
         resultSpy,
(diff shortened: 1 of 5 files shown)


packages/expect/src/jest-expect.ts
```diff
@@ -699,6 +699,42 @@ export const JestChaiExpect: ChaiPlugin = (chai, utils) => {
       )
     },
   )
+  def(
+    ['toHaveBeenCalledBefore'],
+    function (resultSpy: MockInstance) {
+      const expectSpy = getSpy(this)
+
+      const [firstExpectSpyCall] = expectSpy.mock.invocationCallOrder
```
sheremet-va, line 707: Why first? Shouldn't it be the smallest?

https://github.com/jest-community/jest-extended/blob/main/src/matchers/toHaveBeenCalledAfter.js#L43

packages/expect/src/jest-expect.ts
```diff
@@ -699,6 +699,42 @@ export const JestChaiExpect: ChaiPlugin = (chai, utils) => {
       )
     },
   )
+  def(
+    ['toHaveBeenCalledBefore'],
+    function (resultSpy: MockInstance) {
+      const expectSpy = getSpy(this)
```
sheremet-va, line 705: We also need to validate the `resultSpy` to be a mock function

sheremet-va (changes requested):

packages/expect/src/jest-expect.ts
Barbapapazes: Indeed, I will update that part.

packages/expect/src/jest-expect.ts
Barbapapazes: As the calls are ordered in the table, I thought that taking the first was enough, but maybe I'm wrong?

packages/expect/src/jest-expect.ts
```diff
@@ -699,6 +699,42 @@ export const JestChaiExpect: ChaiPlugin = (chai, utils) => {
       )
     },
   )
+  def(
+    ['toHaveBeenCalledBefore'],
+    function (resultSpy: MockInstance) {
```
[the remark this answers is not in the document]
Barbapapazes: yes, this is definitely an interesting option.

packages/expect/src/jest-expect.ts
```diff
@@ -699,6 +699,82 @@ export const JestChaiExpect: ChaiPlugin = (chai, utils) => {
...
+    function (resultSpy: MockInstance, failIfNoFirstInvocation = true) {
+      const expectSpy = getSpy(this)
+
+      if (!isMockFunction(resultSpy)) {
+        throw new TypeError(
+          `${utils.inspect(resultSpy)} is not a spy or a call to a spy`,
+        )
+      }
+
+      if (!isMockFunction(expectSpy)) {
```
sheremet-va, line 729: `getSpy` already checks if it's a spy

packages/expect/src/jest-expect.ts
```diff
@@ -699,6 +699,82 @@ export const JestChaiExpect: ChaiPlugin = (chai, utils) => {
       )
     },
   )
+
+  /**
+   * Used for `toHaveBeenCalledBefore` and `toHaveBeenCalledAfter` to determine if the expected spy was called before the result spy.
+   */
+  function predicate(expectInvocationCallOrder: number[], resultInvocationCallOrder: number[], failIfNoFirstInvocation: number): boolean {
```
sheremet-va, line 706: I think the name is too generic for this giant file

sheremet-va (changes requested):

packages/expect/src/jest-expect.ts
Barbapapazes: Indeed but should I find a better name or is there a file where I could put this function?

sheremet-va: Let's find a better name. The refactoring should be done separetly

packages/expect/src/jest-expect.ts
Barbapapazes: Alright, I'll remove the check

sheremet-va (approved):
Sorry, this got lost in notifications. Looks good to me!

PR #6079, opened 2024-07-10
fix(browser): remove `crypto.randomUUID` from the client

### LINKED ISSUE #6076: Custom hosts no longer work in Browser mode
### Describe the bug

https://github.com/vitest-dev/vitest/blob/af7fec588d23755899d7ad5dc35d28559dcd0596/packages/browser/src/client/client.ts#L14

This line uses crypto.randomUUID() which is only available in a [secure context](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts). Localhost is considered secure, but when using a custom domain (we use one that is just an A record with 127.0.0.1) the page isn't secure without setting up SSL certs, so crypto.randomUUID() isn't available

### Reproduction

```ts
      api: {
        host: "aDomainThatPointsToLocalhost.com",
      },
```

then run browser tests as usual, they should fail

### System Info

```shell
System:
    OS: macOS 14.2.1
    CPU: (8) arm64 Apple M2
    Memory: 219.69 MB / 24.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 21.1.0 - /usr/local/bin/node
    Yarn: 1.22.21 - /opt/homebrew/bin/yarn
    pnpm: 8.6.1 - ~/Library/pnpm/pnpm
    bun: 1.0.26 - ~/.bun/bin/bun
  npmPackages:
    @vitest/browser: ^2.0.1 => 2.0.1 
    @vitest/ui: ^2.0.1 => 2.0.1 
    @vitest/utils: ^2.0.1 => 2.0.1 
    vite: ^5.3.3 => 5.3.3 
    vitest: ^2.0.1 => 2.0.1
```

### Used Package Manager

npm

### Validations

userquin:
### Description

closes #6076

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/browser/src/client/client.ts
+++ b/packages/browser/src/client/client.ts
@@ -11,7 +11,7 @@ export const HOST = [location.hostname, PORT].filter(Boolean).join(':')
 export const SESSION_ID
   = PAGE_TYPE === 'orchestrator'
     ? getBrowserState().contextId
-    : getBrowserState().testerContextId
+    : getBrowserState().testerId
 export const ENTRY_URL = `${
   location.protocol === 'https:' ? 'wss:' : 'ws:'
 }//${HOST}/__vitest_browser_api__?type=${PAGE_TYPE}&sessionId=${SESSION_ID}`
--- a/packages/browser/src/client/client.ts
+++ b/packages/browser/src/client/client.ts
@@ -11,7 +11,7 @@ export const HOST = [location.hostname, PORT].filter(Boolean).join(':')
 export const SESSION_ID
   = PAGE_TYPE === 'orchestrator'
     ? getBrowserState().contextId
-    : crypto.randomUUID()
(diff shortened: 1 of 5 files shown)


packages/browser/src/client/client.ts
```diff
@@ -11,7 +11,7 @@ export const HOST = [location.hostname, PORT].filter(Boolean).join(':')
 export const SESSION_ID
   = PAGE_TYPE === 'orchestrator'
     ? getBrowserState().contextId
-    : crypto.randomUUID()
+    : getBrowserState().testerContextId
```
sheremet-va, line 14: let's call it `testerId` instead - too much `context` bloat I feel like

userquin: done

sheremet-va (approved):

PR #6084, opened 2024-07-10
feat(browser): introduce built-in locators

sheremet-va:
### Description

This PR adds built-in locators.

```ts
import { page } from '@vitest/browser/context'

await page.getByRole('button').click()
await page.getByRole('list').getByRole('listitem', { name: 'Hello World' }).hover()
```

- all providers use `playwright` locators under the hood to support Shadow DOM. 
- when `webdriverio` needs to do an action (a click for example), the locator is transformed into a unique CSS
- when `preview` needs to do an action, we just use the HTML element found with the locator

Proposed API:

```ts
import { page } from '@vitest/browser/context'

interface Page {
  getByRole(role: string, options?: RoleOptions): Locator
  getByLabelText(text: string | RegExp): Locator
  getByTestId(testId: string | RegExp): Locator
  getByAltText(text: string | RegExp): Locator
  getByPlaceholder(text: string | RegExp): Locator
  getByText(text: string | RegExp): Locator
  getByTitle(title: string | RegExp): Locator

  elementLocator(element: Element): Locator
}

interface Locator {
  click(options?: UserEventClickOptions): Promise<void>
  dblClick(options?: UserEventClickOptions): Promise<void>
  tripleClick(options?: UserEventClickOptions): Promise<void>
  clear(): Promise<void>
  hover(): Promise<void>
  unhover(): Promise<void>
  fill(text: string, options?: UserEventFillOptions): Promise<void>
  screenshot(): Promise<ScreenshotResult>
  dropTo(element: Locator): Promise<void>

  // these are useful methods to combine them with third-party libraries that require Element or HTMLElement
  // return the element immediately or fails there are no elements or more than 2
  element(): Element
  elements(): Element[]
  query(): Element | null
  all(): Locator[]

  // chaining methods
  getByRole(role: string): Locator
  // ... same as on page
}
```

All `@testing-library/jest-dom` are support out of the box:

```ts
expect(page.getByRole('button').element()).toHaveTextContent('text')
// calls `locator.element()` automatically
await expect.element(page.getByRole('button')).toHaveTextContent('text')
```

### Tests

### Documentation

### Changesets


packages/browser/package.json
```diff
@@ -40,6 +40,10 @@
     "./providers/playwright": {
       "types": "./providers/playwright.d.ts"
     },
+    "./locator": {
+      "types": "./dist/locators/index.d.ts",
+      "default": "./dist/locators/index.js"
+    },
```
userquin, line 46: shouldn't be `locator`?

Can you add `typesVersions` for node10 below `"types"`?
```json
"typesVersions": {
    "*": {
      "client": [
        "./dist/client.d.ts"
      ],
      "locator": [
        "./dist/locators/index.d.ts"
      ]
    }
  },
```

sheremet-va: > Can you add `typesVersions` for node10 below `"types"`?

We don't support node 10

> shouldn't be `locator`?

What shouldn't be?

userquin: It seems client subpackage is missing...

sheremet-va: client subpackage doesn't need a dts, it's not public

userquin: `"moduleResolution": "node"` (areyourtypeswrong call it node10), add "locator" in "typesVersions"

sheremet-va: We don't support `moduleResolution: node` because Vite doesn't support it because Rollup doesn't support it (see `parseAst` exported from all)

packages/browser/providers/playwright.d.ts
```diff
@@ -41,7 +41,7 @@ declare module '@vitest/browser/context' {
   export interface UserEventDoubleClickOptions extends PWDoubleClickOptions {}
```
userquin, line 41: missing triple click options

PR #6088, opened 2024-07-11
fix(browser): userEvent.setup initiates a separate state for userEvent instance

### LINKED ISSUE #6082: userEvent.keyboard always releases pressed key with playwright provider
### Describe the bug

I am trying to use the new Interactive API to test the behavior of a button.
It seems that every keyboard command [is automatically released](https://github.com/vitest-dev/vitest/blob/a169d25a922936fabb603173337455ed6d59e86f/packages/browser/src/node/commands/keyboard.ts#L100) even when using the `{Enter>}` syntax.

### Reproduction

https://github.com/edoardocavazza/vitest-keyboard

### System Info

```shell
System:
    OS: macOS 14.2.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 261.33 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.19.0 - ~/.nvm/versions/node/v18.19.0/bin/node
    npm: 10.2.3 - ~/.nvm/versions/node/v18.19.0/bin/npm
    pnpm: 8.6.3 - ~/Library/pnpm/pnpm
    bun: 1.0.30 - ~/.bun/bin/bun
  Browsers:
    Chrome: 126.0.6478.127
    Safari: 17.2.1
  npmPackages:
    @vitest/browser: ^2.0.1 => 2.0.1 
    vitest: ^2.0.1 => 2.0.1
```

### Used Package Manager

npm

### Validations

sheremet-va:
### Description

Closes #6082

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/guide/browser/interactivity-api.md
+++ b/docs/guide/browser/interactivity-api.md
@@ -163,7 +163,7 @@ test('trigger keystrokes', async () => {
 
 References:
 
-- [Playwright `locator.press` API](https://playwright.dev/docs/api/class-locator#locator-press)
+- [Playwright `Keyboard` API](https://playwright.dev/docs/api/class-keyboard)
 - [WebdriverIO `action('key')` API](https://webdriver.io/docs/api/browser/action#key-input-source)
 - [testing-library `type` API](https://testing-library.com/docs/user-event/utility/#type)
 
@@ -194,7 +194,7 @@ test('tab works', async () => {
 
 References:
 
-- [Playwright `locator.press` API](https://playwright.dev/docs/api/class-locator#locator-press)
+- [Playwright `Keyboard` API](https://playwright.dev/docs/api/class-keyboard)
 - [WebdriverIO `action('key')` API](https://webdriver.io/docs/api/browser/action#key-input-source)
 - [testing-library `tab` API](https://testing-library.com/docs/user-event/convenience/#tab)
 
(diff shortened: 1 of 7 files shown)


sheremet-va:
@edoardocavazza I am actually not sure about having a global state - it leaks to other tests (I had to add `.setup` to some tests to not break them because previous tests were releasing buttons). What do you think?

sheremet-va:
> Consider this test:

`Shift` will not be released in your example - not before, not after this PR

edoardocavazza:
Sorry, I may have misunderstood the question.
What I wanted to demonstrate with that test is that having a global state may still be necessary to clean the state and avoid leaks in subsequent tests.

sheremet-va:
Yes, I was saying in my previous question that I also found those leaks, but I am not sure what to do with them. Do we resolve them automatically? If you had more tests in between, or `test.each` it would be harder to find why `shift` was pressed, for example. Automatic cleanup might cause unexpected issues because it will call `keyup` handlers that might throw an error, for example.

edoardocavazza:
Ok, now I understand your concerns. My opinion is that however we need to deviate slightly from the functioning of the original library. Automatic cleaning at the end of each test seems to me to be the most desired solution. It should be the developer's responsibility to make sure to clean up the test state by using `afterEach` and deleting added listeners. This is just my preference, but I am open to other solutions.

Issue #6087, opened 2024-07-11
Failed to load module created in test


Kaciras:
### Describe the bug

After upgrading Vitest to 2.0.2, I get the error when import a file created in test:

```javascript
import { afterEach, expect, it } from "vitest";
import { pathToFileURL } from "url";
import { unlinkSync, writeFileSync } from "fs";

const filename = "bar.js";

afterEach(() => unlinkSync(filename));

it("should works", async () => {
	writeFileSync(filename, "export default 123");
	const mod = await import(pathToFileURL(filename).href);
	expect(mod.default).toBe(123);
});
```

```text
Error: Failed to load url /D:/Coding/JavaScript/vite-svg-sfc/bar.js (resolved id: /D:/Coding/JavaScript/vite-svg-sfc/bar.js). Does the file exist?
    at loadAndTransform (file:///D:/Coding/JavaScript/vite-svg-sfc/node_modules/.pnpm/vite@5.3.3_@types+node@20.14.10/node_modules/vite/dist/node/chunks/dep-CzJTQ5q7.js:52400:17)
    at async ViteNodeServer._transformRequest (file:///D:/Coding/JavaScript/vite-svg-sfc/node_modules/.pnpm/vite-node@2.0.2_@types+node@20.14.10/node_modules/vite-node/dist/server.mjs:524:16)
    at async ViteNodeServer._fetchModule (file:///D:/Coding/JavaScript/vite-svg-sfc/node_modules/.pnpm/vite-node@2.0.2_@types+node@20.14.10/node_modules/vite-node/dist/server.mjs:486:17)
    at async Proxy.fetch (file:///D:/Coding/JavaScript/vite-svg-sfc/node_modules/.pnpm/vitest@2.0.2_@types+node@20.14.10/node_modules/vitest/dist/vendor/cli-api.FtQZKvQk.js:8093:22)
    at async EventEmitter.<anonymous> (file:///D:/Coding/JavaScript/vite-svg-sfc/node_modules/.pnpm/vitest@2.0.2_@types+node@20.14.10/node_modules/vitest/dist/vendor/index.BpSiYbpB.js:69:20)
```

It works in Vitest 1.6.0

### Reproduction

See above

### System Info

```shell
System:
    OS: Windows 11 10.0.22621
    CPU: (12) x64 AMD Ryzen 5 5625U with Radeon Graphics
    Memory: 6.20 GB / 15.31 GB
  Binaries:
[...]

PR #6109, opened 2024-07-12
fix(vite-node): remove suffix slash on file protocol for window

syi0808:
### Description
In window platform, dynamic import with file protocol are convert to request id and request id normalized by normalizeRequestId function.
normalizeRequestId function's return value with file protocol is like '/C:Users/user/workspace/file.js' in window, so fs can't resolve this absolute path because of suffix forward slash.

For that reason, when isWindow is true, i decided to remove the suffix slash. 

### before this pr
IMG_9441-min

### after this pr
IMG_9440-min

### Tests

### Documentation

### Changesets

Fix: #6087


syi0808:
Additional, vite-node hasn't test, so i don't add test for this changes. but is it better to add?

sheremet-va:
I don't think this is the root of the problem, this code wasn't touched for a long time. We need to know why the issue happens when it happens. It worked in 1.6.0. We also need a new test case of course.

syi0808:
> I don't think this is the root of the problem, this code wasn't touched for a long time. We need to know why the issue happens when it happens. It worked in 1.6.0. We also need a new test case of course.

I have trouble to add test. https://github.com/vitest-dev/vitest/pull/6109/commits/b6a34fa21bffe7a33f4000bda2aaf004cc1d8209
I dynamic import vite-node/utils, but isWindow not changed. It fixed by first process.platform.

And let's look for the reason why it worked in 1.6.0. The information I checked while debugging was generated from 2.0.0-beta.1, but the changes in 2.0.0-beta.1 did not seem to have much effect. I checked variously, such as node version and vite, but the only thing I found was the corresponding PR.

sheremet-va:
> I dynamic import vite-node/utils, but isWindow not changed. It fixed by first process.platform.

1. You don't need to mock the platform, we are running tests on all platforms
2. You shouldn't test the function because that's not what this PR fixes - you need to test the dynamic import

syi0808:
> > I dynamic import vite-node/utils, but isWindow not changed. It fixed by first process.platform.
> 
> 1. You don't need to mock the platform, we are running tests on all platforms
> 2. You shouldn't test the function because that's not what this PR fixes - you need to test the dynamic import with an absolute path (vite-node is just an implementation detail)

Thanks for guide. I remove util unit test and add integration test about write file and dynamic import.
And I've been debugging, but I can't find a good reason. What's certain is that 'fs.readFile('/C:/Users/user/workspace/file.js')' throws an error.

syi0808:
I think failed test is not depend on this pr changes. So new added test is works as well and I'll keep looking for reasons.

syi0808:
I found it @sheremet-va. this happen is because of https://github.com/vitest-dev/vitest/pull/5177.
In version 1.6.0 `server.watch.ignored` is exist. So it not cache fs method results that vite implements.
Bug version 2 `server.watch.ignored` is not exist because remove watchExclude. So it cache fs method results that vite implements.
File path resolving logic in vite fsUtils implemenet so, I don't know the details of the implementation but, file path resolving logic is certainly changed by `server.watch.ignored`.

Screenshot 2024-07-13 at 6 45 56 AM
Screenshot 2024-07-13 at 6 40 52 AM
Screenshot 2024-07-13 at 7 11 15 AM

syi0808:
So I think this change can be resolved separately from the cause, so why not reflect it? Basically, the path with the slash in window is wrong.

sheremet-va (approved):

sheremet-va:
LGTM!

PR #6120, opened 2024-07-14
fix(coverage): v8 to support source maps with multiple sources

AriPerkkio:
### Description

While setting up Vitest monorepo's coverage, noticed that V8 shows only the first file of bundled sources.

### Tests

### Documentation

### Changesets


AriPerkkio:
Looks like bug in Vite's SSR transform:

```js
import { createServer } from "vite";

const server = await createServer({
  configFile: false,
  root: fileURLToPath(new URL(".", import.meta.url)),
  server: { port: 1337 },
});

await server.listen();

const ssr = await server.transformRequest("./dist/bundle.js", { ssr: true });
const web = await server.transformRequest("./dist/bundle.js", { ssr: false });

console.log({ ssr: ssr.map.sources, web: web.map.sources });
// {
//   ssr: [ '../src/first.ts' ],
//   web: [ '../src/first.ts', '../src/second.ts' ]
// }

server.close();
```

AriPerkkio:
V8 coverage works fine with https://github.com/vitejs/vite/pull/17677 applied locally. Generated the snapshots with that patch.

package.json
```diff
@@ -61,7 +61,7 @@
     "rollup-plugin-license": "^3.5.2",
     "tsx": "^4.16.5",
     "typescript": "^5.5.4",
-    "vite": "^5.3.3",
+    "vite": "^5.4.0",
```
AriPerkkio, line 64: Needed for https://github.com/vitejs/vite/pull/17677

packages/coverage-v8/src/provider.ts
AriPerkkio, line 1: Verified that all our old tests pass with these changes when used with older Vite versions. ✅

packages/coverage-v8/src/provider.ts
```diff
@@ -457,39 +457,44 @@ export class V8CoverageProvider extends BaseCoverageProvider implements Coverage
...
 
-    const sources = [url]
-    if (map.sources && map.sources[0] && !url.endsWith(map.sources[0])) {
-      sources[0] = new URL(map.sources[0], url).href
+    const sources = (map.sources || [])
+      .filter((source): source is NonNullable<typeof source> => !!source)
+      .map(source => new URL(source, url).href)
+
+    if (sources.length === 0) {
+      sources.push(url)
```
AriPerkkio, line 485: I'm not sure if this is even needed. If it's removed, all tests pass just fine. Maybe let's try removing this in future and leave it here for now.

packages/coverage-v8/src/provider.ts
```diff
@@ -457,39 +457,44 @@ export class V8CoverageProvider extends BaseCoverageProvider implements Coverage
...
+        source: code || sourcesContent[0],
+        originalSource: sourcesContent[0],
       }
     }
 
-    const sources = [url]
-    if (map.sources && map.sources[0] && !url.endsWith(map.sources[0])) {
-      sources[0] = new URL(map.sources[0], url).href
+    const sources = (map.sources || [])
+      .filter((source): source is NonNullable<typeof source> => !!source)
```
sheremet-va, line 481: you can do `source != null` and you don't even need to have `source is` then :P

this was added in the latest TS version

sheremet-va (approved):

packages/coverage-v8/src/provider.ts
AriPerkkio: That works, nice! I though it would detect `!!` as well, and thought we were using older TS version. 💯

sheremet-va: Yeah, unfortunately it doesn't work with `!!` because it's not an explicit assertion 😢

sheremet-va (approved):

PR #6121, opened 2024-07-14
docs: improve extending matchers docs

vorant94:
### Description

This PR improves the docs for setting up custom matchers. With the current guide I wasn't able to properly configure TypeScript, so I needed to take a look at [Testing LIbrary Jest DOM matchers](https://github.com/testing-library/jest-dom/blob/main/types/vitest.d.ts) how they handle it. Their approach worked for me, so I suggest updating the docs accordingly.

- change ambient declaration file to how it is configured inside [testing lib](https://github.com/testing-library/jest-dom/blob/main/types/vitest.d.ts) since the current way didn't work for me
- rename MatcherResult to ExpectationResult as this is the current name of custom matcher return type according to source code (see MatchersObject > RawMatcherFn > ExpectationResult)

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/guide/extending-matchers.md
+++ b/docs/guide/extending-matchers.md
@@ -26,11 +26,13 @@ expect.extend({
 If you are using TypeScript, you can extend default `Assertion` interface in an ambient declaration file (e.g: `vitest.d.ts`) with the code below:
 
 ```ts
+import 'vitest'
+
 interface CustomMatchers<R = unknown> {
   toBeFoo: () => R
 }
 
-declare module '@vitest/expect' {
+declare module 'vitest' {
   interface Assertion<T = any> extends CustomMatchers<T> {}
   interface AsymmetricMatchersContaining extends CustomMatchers {}
 }
--- a/docs/guide/extending-matchers.md
+++ b/docs/guide/extending-matchers.md
@@ -26,13 +26,11 @@ expect.extend({
 If you are using TypeScript, you can extend default `Assertion` interface in an ambient declaration file (e.g: `vitest.d.ts`) with the code below:
 
 ```ts
-import type { Assertion, AsymmetricMatchersContaining } from 'vitest'
-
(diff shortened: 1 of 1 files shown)


docs/guide/extending-matchers.md
```diff
@@ -26,13 +26,11 @@ expect.extend({
...
 
 ```ts
-import type { Assertion, AsymmetricMatchersContaining } from 'vitest'
-
 interface CustomMatchers<R = unknown> {
   toBeFoo: () => R
 }
 
-declare module 'vitest' {
+declare module '@vitest/expect' {
```
sheremet-va, line 33: This is not the correct way to augment it. `@vitest/expect` is an implementation detail and should not be exposed. Testing Library gives a bad example

sheremet-va (changes requested):

docs/guide/extending-matchers.md
vorant94: aggre about `export {}` though in my case it wasn't needed since my `global.d.ts` file anyway imports some of mine types.

regarding `vitest` instead of `@vitest/expect` I played around a little bit more and here what I found:

-  both types from here `import type { Assertion, AsymmetricMatchersContaining } from 'vitest'` are considered unused and hence cleaned up by linters/formatters
- without importing something from 'vitest` `declare module 'vitest'` doesn't work for me

so what do you think about the following example? no `export` is needed since there is `import` and `@vitest/expect` is hidden as internal implementation detail

```ts
import 'vitest';

interface CustomMatchers<R = unknown> {
  toBeFoo: () => R
}

declare module 'vitest' {
  interface Assertion<T = any> extends CustomMatchers<T> {}
  interface AsymmetricMatchersContaining extends CustomMatchers {}
}
```

sheremet-va: Sure, looks good to me 👍🏻

PR #6123, opened 2024-07-14
perf(utils): optimize parse stack trace regexp

### LINKED ISSUE #6039: Performance issue when throwing errors that have stackTraceLimit = 0
### Describe the bug

When setting `Error.stackTraceLimit = 0` in a custom error and throwing it inside a test, the time it takes for vitest to display the error is proportional to the size of the object passed (size of the message string).
In the example below, you can change the value of object size and see in the console that it takes more and more time to return. You can also comment the line `Error.stackTraceLimit = 0` to see that the problem disappears.
On my machine, vitest just hangs after a certain size and I have to kill the process manually.
Note that I am using an external library that throws errors with a stackTraceLimit of 0, so I do not control this on my end.

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-edo7vt?file=test%2Fbasic.test.ts

### System Info
from stackblitz:
```shell
System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    @vitest/ui: latest => 1.6.0 
    vite: latest => 5.3.3 
    vitest: latest => 1.6.0
```

### Used Package Manager

npm

### Validations

userquin:
### Description

closes #6039

Check https://github.com/userquin/vitest-stacktrace-issue-6039

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/utils/src/source-map.ts
+++ b/packages/utils/src/source-map.ts
@@ -90,17 +90,17 @@ export function parseSingleFFOrSafariStack(raw: string): ParsedStack | null {
     return null
   }
 
-  // eslint-disable-next-line regexp/no-super-linear-backtracking,regexp/no-misleading-capturing-group,regexp/no-unused-capturing-group
-  const matches = /(at\s+)?(async\s+)?([^\s@]+)?(?:@|\s+\()?([^\s()]+):(\d+):(\d+)\)?/.exec(line)
+  // eslint-disable-next-line regexp/no-useless-assertions,regexp/no-unused-capturing-group
+  const matches = /^(\bat\s*\b)?(?:\b\s*async\s*\b)?(\b\w+(?!([/:@]|(https?)|(file)))\b)?(?:@|\s+\()?([^\s()]+):(\d+):(\d+)\)?$/.exec(line)
   if (!matches) {
     return null
   }
 
   return {
-    file: prepareUrl(matches[4]),
-    method: matches[3] || '',
-    line: Number.parseInt(matches[5]),
-    column: Number.parseInt(matches[6]),
+    file: prepareUrl(matches[6]),
+    method: matches[2] || '',
+    line: Number.parseInt(matches[7]),
+    column: Number.parseInt(matches[8]),
   }
 }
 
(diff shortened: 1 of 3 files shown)


packages/utils/src/source-map.ts
```diff
@@ -87,23 +90,17 @@ export function parseSingleFFOrSafariStack(raw: string): ParsedStack | null {
...
-  const functionNameRegex = /((.*".+"[^@]*)?[^@]*)(@)/
-  const matches = line.match(functionNameRegex)
-  const functionName = matches && matches[1] ? matches[1] : undefined
-  const [url, lineNumber, columnNumber] = extractLocation(
-    line.replace(functionNameRegex, ''),
-  )
-
-  if (!url || !lineNumber || !columnNumber) {
+  // eslint-disable-next-line regexp/no-super-linear-backtracking,regexp/no-misleading-capturing-group,regexp/no-unused-capturing-group
+  const matches = /(at\s+)?(async\s+)?([^\s@]+)?(?:@|\s+\()?([^\s()]+):(\d+):(\d+)\)?/.exec(line)
```
AriPerkkio, line 94: Disabling these ESLint rules is not convincing. Can we fix these errors as well?

AriPerkkio (changes requested):
We should add test case from https://github.com/vitest-dev/vitest/issues/6039 that timeouts without this fix.

packages/utils/src/source-map.ts
userquin: can you check @AriPerkkio  ? check also discord server, some hints there

userquin:
> We should add test case from #6039 that timeouts without this fix.

Added a test in the test/browser: rn we cannot test the timeout since we need to add PW/WDIO test, and we have only PW/WDIO tests using happy-dom to test the ui.

(check `test/ui/test` folder)

sheremet-va:
Seems like the current implementation breaks inline snapshots in the browser

sheremet-va:
> Seems like the current implementation breaks inline snapshots in the browser

By `current`, I mean this PR. It works fine in main.

jrmdayn:
@userquin any luck fixing this issue? 🙏 🙏 🙏

AriPerkkio:
https://github.com/vitest-dev/vitest/pull/8531

Issue #6089, opened 2024-07-11
Expect mock factories to have a return type compatible with the original module


acidoxee:
### Clear and concise description of the problem

Currently, when using mock factories, the expected return type is `any`. This makes it possible to return broken mocks that do not comply with the original module's signature.

```ts
// @filename: src/basic.ts
export const squared = (n: number) => n * n

// @filename: test/mocks.test.ts
import { expect, test, vi } from 'vitest';

test('Mocked module with broken contract', async () => {
  vi.doMock(import('../src/basic.js'), () => ({
    // When not manually typing the result, we can return a broken mock from the factory, as its expected return type is `any`
    sqrd: () => 1,
  }));

  // This throws the following error: Error: [vitest] No "squared" export is defined on the "../src/basic.js" mock. Did you forget to return it from "vi.mock"?
  const { squared } = await import('../src/basic.js');
  expect(squared(2)).toBe(1);
});
```

### Suggested solution

It would be great if the mock factory's return type was expected to be the same as the original module's, instead of using `any`.

### Alternative

Meanwhile, we've got hundreds of return type overrides in our codebase for each mock with `satisfies Awaited<ReturnType<typeof importOriginal>>` to ensure the mock's type is correct, but this is tedious and error-prone.

```ts
// @filename: src/basic.ts
export const squared = (n: number) => n * n

// @filename: test/mocks.test.ts
import { expect, test, vi } from 'vitest';

test('Mocked module', async () => {
  vi.doMock(
    import('../src/basic.js'),
    (importOriginal) =>
      // This ensures the value returned from the factory is compatible with the original module.
      // I would like for the whole `satisfies` statement to not be necessary
      ({ squared: () => 1 } satisfies Awaited<
        ReturnType<typeof importOriginal>
      >)
  );

  const { squared } = await import('../src/basic.js');
  expect(squared(2)).toBe(1);
});
```

### Additional context

Here's a reproduction: https://stackblitz.com/edit/vitest-dev-vitest-fkhseh?file=test%2Fmocks.test.ts

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #6139, opened 2024-07-15
fix: add return type and promisable mockFactory

syi0808:
### Description
Adjust MockFactoryWithHelper type and add PromiseMockFactoryWithHelper.

### screenshot
스크린샷 2024-07-16 오전 12 00 07
스크린샷 2024-07-16 오전 12 20 11

### Tests

### Documentation

### Changesets

Resolve: #6089

### DIFF
--- a/packages/vitest/src/types/mocker.ts
+++ b/packages/vitest/src/types/mocker.ts
@@ -1,6 +1,9 @@
 export type MockFactoryWithHelper<M = unknown> = (
-  importOriginal: <T extends M>() => Promise<T>
-) => any
+  importOriginal: <T extends M = M>() => Promise<T>
+) => Partial<Record<keyof M, any>>
+export type PromiseMockFactoryWithHelper<M = unknown> = (
+  importOriginal: <T extends M = M>() => Promise<T>
+) => Promise<Partial<Record<keyof M, any>>>
 export type MockFactory = () => any
 
 export type MockMap = Map<string, Record<string, string | null | MockFactory>>
--- a/packages/vitest/src/integrations/vi.ts
+++ b/packages/vitest/src/integrations/vi.ts
@@ -3,7 +3,7 @@ import { assertTypes, createSimpleStackTrace } from '@vitest/utils'
 import { parseSingleStack } from '../utils/source-map'
 import type { VitestMocker } from '../runtime/mocker'
 import type { ResolvedConfig, RuntimeConfig } from '../types'
-import type { MockFactoryWithHelper } from '../types/mocker'
+import type { MockFactoryWithHelper, PromiseMockFactoryWithHelper } from '../types/mocker'
 import { getWorkerState } from '../utils/global'
 import { resetModules, waitForImportsToResolve } from '../utils/modules'
 import { isChildProcess } from '../utils/base'
@@ -194,6 +194,8 @@ export interface VitestUtils {
   mock(path: string, factory?: MockFactoryWithHelper): void
   // eslint-disable-next-line ts/method-signature-style
   mock<T>(module: Promise<T>, factory?: MockFactoryWithHelper<T>): void
+  // eslint-disable-next-line ts/method-signature-style
+  mock<T>(module: Promise<T>, factory?: PromiseMockFactoryWithHelper<T>): void
 
   /**
    * Removes module from mocked registry. All calls to import will return the original module even if it was mocked before.
@@ -219,6 +221,8 @@ export interface VitestUtils {
   doMock(path: string, factory?: MockFactoryWithHelper): void
   // eslint-disable-next-line ts/method-signature-style
   doMock<T>(module: Promise<T>, factory?: MockFactoryWithHelper<T>): void
+  // eslint-disable-next-line ts/method-signature-style
+  doMock<T>(module: Promise<T>, factory?: PromiseMockFactoryWithHelper<T>): void
   /**
    * Removes module from mocked registry. All subsequent calls to import will return original module.
    *


syi0808:
I removed generic in importOriginal because of a type error, which could be a breaking change. I guess it's a typescript problem, but let's look at it more.

packages/vitest/src/types/mocker.ts
```diff
@@ -1,6 +1,9 @@
 export type MockFactoryWithHelper<M = unknown> = (
-  importOriginal: <T extends M>() => Promise<T>
-) => any
+  importOriginal: () => Promise<M>
+) => Partial<Record<keyof M, any>>
+export type PromiseMockFactoryWithHelper<M = unknown> = (
+  importOriginal: () => Promise<M>
+) => Promise<Partial<Record<keyof M, any>>>
```
sheremet-va, line 6: 1. Why is it partial?
2. Why do you need a separate `Promise` one and won't just do `| Promise<T>`?

syi0808: 1. I don't think everyone wants to mock all the exports.
2. The union type like `| Promise<T>` make wrong type like below screenshot. 
스크린샷 2024-07-16 오전 12 48 54

sheremet-va: > I don't think everyone wants to mock all the exports.

If they want typechecking, they need to mock it fully. We already throw an error if something is not returned.

syi0808: > If they want typechecking, they need to mock it fully. We already throw an error if something is not returned.

hmm... I think so. Because if you want to implement partially, they can call importOriginal.

> Seems fine to me

That is not async function but, typescript provides promise method.

sheremet-va: > That is not async function

You can return a `new Promise` from every function - in fact any object with `then` method can be awaited, so TypeScript is not wrong here

syi0808: I think I was wrong, but I think it might not be normal to be offered auto-completion for promise method, so I'm submitting PR in typescript.
But if it's okay, I think we can revise it as you suggested.

packages/vitest/src/types/mocker.ts
```diff
@@ -1,6 +1,8 @@
+type Promisable<T> = T | Promise<T>
+
 export type MockFactoryWithHelper<M = unknown> = (
-  importOriginal: <T extends M>() => Promise<T>
-) => any
+  importOriginal: <T extends M = M>() => Promise<T>
+) => Promisable<Record<keyof M, any> | null>
```
acidoxee, line 5: My apologies if my initial issue wasn't clear, but what I meant with a "compatible" mock type was not only regarding the keys in the export map, but their values as well.

Ensuring compatible top-level keys would already be a welcome improvement of course, but this doesn't address the correctness of the exports themselves. I'd still be able to provide a completely false (type-wise) mock from the factory, and TypeScript wouldn't tell me anything about it. I would still need to manually add the `Awaited<ReturnType<typeof importOriginal>>` return type to each of my factories to ensure my mocks are correct.

kettanaito:
I think this is a good change. `any` certainly doesn't belong in the return type of the mock factory function. It has to be `T`, where it stands for the module you've imported (I'm okay if that works only with dynamic imports). 

Another related issue is that Vitest may not provide correct mock implementations for functions out of the box. So while you technically don't have to provide any factory function, the default Vitest behavior will result in a broken module. 

Example:

```ts
// api.js
export const foo = { 
  bar: async () => 123
}
```

Here, `foo.bar` is a function that returns a Promise that resolves to `123`. However, when mocked vi `vi.mock()`/`vi.doMock()`, it doesn't do that:

```ts
vi.doMock(import('./api.js'))

const { api } = await import('./api.js')
api.bar() // returns nothing, and is NOT an async function.
```

Vitest converts all functions to `vi.fn()` regardless if they are sync or async. In this scenario, I'd expect `foo.bar` to translate to `vi.fn().mockResolvedValue(undefined)`.

Since the mock function is synchronous, assertions like this will fail:

```ts
await expect(foo.bar()).resolves.toBeUndefined()
```

sheremet-va:
> I think this is a good change. any certainly doesn't belong in the return type of the mock factory function. It has to be T, where it stands for the module you've imported (I'm okay if that works only with dynamic imports).

So, do you think users should be required to return every export even if it's not used?

> Another related issue is that Vitest may not provide correct mock implementations for functions out of the box. So while you technically don't have to provide any factory function, the default Vitest behavior will result in a broken module.

I'd say it is an intentionally broken module, this is how automocking works (and it's even documented!)

> Vitest converts all functions to vi.fn() regardless if they are sync or async. In this scenario, I'd expect foo.bar to translate to vi.fn().mockResolvedValue(undefined).

it is not possible to know if the function returns a promise without calling it which defeats the purpose of automocking. (You can look at the function type for AsyncFunction, but regular functions can also return Promises)

kettanaito:
> So, do you think users should be required to return every export even if it's not used?

Yes and no. Yes, because I like being explicit. I can do `await importOriginal()` and return/modify the original module in my factory. No, because Vitest provides automocking for exports, so, technically, all exports are already returned even if I haven't returned anything from the factory function.

I'm leaning toward No here, on the assumption that Vitest also provides the correct mock implementations for my exports (see the async function issue above). If it doesn't, I'd actually expect my factory function to type-error, letting me know that certain exports need my attention. 

> I'd say it is an intentionally broken module, this is how automocking works (and it's even documented!)

It's a nice gotcha! 

> it is not possible to know if the function returns a promise without calling it which defeats the purpose of automocking.

I presumed that was the reason. Thanks for clarifying!

sheremet-va:
> No, because Vitest provides automocking for exports, so, technically, all exports are already returned even if I haven't returned anything from the factory function.

This is not true, Vitest doesn't automock anything returned from the factory. That would require importing the original module. Vitest returns a proxy that throws an error if you try to access a property that you didn't return from the factory.

kettanaito:
Sorry if I phrased it wrong. Here's what I meant: calling `vi.doMock(import('./foo.js'))` will automatically replace all exports in `foo.js` with Vitest mocks. This is the behavior I experience on 2.x.

packages/vitest/src/types/mocker.ts
```diff
@@ -1,6 +1,8 @@
+type Promisable<T> = T | Promise<T>
+
 export type MockFactoryWithHelper<M = unknown> = (
-  importOriginal: <T extends M>() => Promise<T>
-) => any
+  importOriginal: <T extends M = M>() => Promise<T>
+) => Promisable<M | null>
```
sheremet-va, line 5: 1. `null` is never allowed
2. Let's return `Partial<M>` - it's allowed to return only a part of the module. I think it's better to throw an error with a nice stack trace like we already do if you forgot to return an export

sheremet-va (changes requested):

packages/vitest/src/types/mocker.ts
syi0808: Thank you for the review. It's all reflected.

sheremet-va:
Looks like you have type errors now. You need to add `ts-expect-error` to

https://github.com/vitest-dev/vitest/blob/a7be50dbceea31611403542caca7ebf44e4e014a/test/core/test/mocking/factory.test.ts#L58

sheremet-va (changes requested):
I think the documentation should reflect the change. Both in type and description

syi0808:
> I think the documentation should reflect the change. Both in type and description

I've drafted the document.

sheremet-va (approved):

acidoxee:
Thanks a lot for solving this! DX is going to be great 🎉

kwojcik:
Just had a multi-hour debug session with a coworker because of a bad mock seemingly crashing vitest `vi.mock('...', () => [])` instead of `vi.mock('...', () => ({default: () => []}))`, this is awesome!

PR #6150, opened 2024-07-16
fix(expect): prefer `unknown` in certain assertions

robertrossmann:
When `expect.Any(...)` returns `any` type, some ESLint configurations will flag this as potentially dangerous due to the fact that `any` simply turns off type checking for this object completely.

Screenshot 2024-07-16 at 14 55 23

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/expect/src/types.ts
+++ b/packages/expect/src/types.ts
@@ -98,19 +98,19 @@ export interface ExpectStatic
   AsymmetricMatchersContaining {
   <T>(actual: T, message?: string): Assertion<T>
   extend: (expects: MatchersObject) => void
-  anything: () => any
-  any: (constructor: unknown) => any
+  anything: () => unknown
+  any: (constructor: unknown) => unknown
   getState: () => MatcherState
   setState: (state: Partial<MatcherState>) => void
   not: AsymmetricMatchersContaining
 }
 
 export interface AsymmetricMatchersContaining {
-  stringContaining: (expected: string) => any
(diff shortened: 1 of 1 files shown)


sheremet-va:
This is expected behavior. Configure your eslint rule to not give an error for `expect` matchers.

These matchers need to return the broadest type so TypeScript doesn't fail. Consider this:

```ts
interface MyType {
  someValue: number
  otherValue: {}
  thirdValue: string
}

const myTypeValue: MyType = {
  someValue: expect.any(Number),
  otherValue: expect.anything(),
  thirdValue: expect.stringContaining('string')
}
```

If any of them return `unkown`, then the shape of the object is not assignable according to TypeScript.

robertrossmann:
@sheremet-va but this is not what the `expect.Any()` method returns. I feel like the pattern you mention should not be a valid approach to writing assertions. 🤔

sheremet-va:
> @sheremet-va but this is not what the `expect.Any()` method returns. I feel like the pattern you mention should not be a valid approach to writing assertions. 🤔

The type is meant for how it is used. If you don't like this pattern, then why do you use it in your example? It will fail if this PR is merged.

robertrossmann:
My understanding was that `t.expect()` is a drop-in replacement for Jest's `expect` assertions that is fully compatible. Perhaps this understanding is not correct and I will update my usage patterns. But I thought this change would benefit others that might come to the same conclusion.

I think those test failures could potentially be fixed. 🤔

Issue #6104, opened 2024-07-12
`Intl.RelativeTimeFormat` and `Intl.ListFormat` can't be spied upon


amannn:
### Describe the bug

When using `vi.spyOn(Intl, 'RelativeTimeFormat')`  or `vi.spyOn(Intl, 'ListFormat')`, the constructor can no longer be called due to:

```
Constructor Intl.RelativeTimeFormat requires 'new'
```

Note that `Intl.DateTimeFormat` works flawlessly.

### Reproduction

[StackBlitz](https://stackblitz.com/edit/vitest-dev-vitest-ye9hke?file=src%2Fbasic.ts,test%2Fbasic.test.ts&initialPath=__vitest__/)

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    CPU: (6) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 18.20.3 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.2.3 - /usr/local/bin/npm
    pnpm: 8.15.6 - /usr/local/bin/pnpm
  npmPackages:
    @vitest/ui: latest => 2.0.2 
    vite: latest => 5.3.3 
    vitest: latest => 2.0.2 
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

sheremet-va:
Respecting `new` in a constructor would be a breaking change as it will change the value of `mock.instances`. However, you can already use `tinyspy` directly:

```ts
[...]

PR #6160, opened 2024-07-18
fix(spy)!: support spying on classes

sheremet-va:
### Description

Closes [#6104](https://github.com/vitest-dev/vitest/issues/6104)

This is a breaking change.

This is not supported anymore:
```ts
vi.spyOn(mock, 'HelloWorld').mockImplementationOnce(() => {
  const Mock = vi.fn()
  Mock.prototype.hello = vi.fn(() => 'hello world')
  return new Mock()
})
const hw = new mock.HelloWorld()
```

But you can replace it with a `function`:
```diff
+ vi.spyOn(mock, 'HelloWorld').mockImplementationOnce(function () {
  const Mock = vi.fn()
  Mock.prototype.hello = vi.fn(() => 'hello world')
  return new Mock()
})
const hw = new mock.HelloWorld()
```

But ideally you should instead do something like this:

```ts
const HelloWorld = vi.spyOn(mock, 'HelloWorld')
HelloWorld.prototype.hello = vi.fn(() => 'hello world')
const hw = new mock.HelloWorld()
```

Or this (previously not supported):

```ts
const HelloWorld = vi.spyOn(mock, 'HelloWorld').mockImplementation(class Mock {
  hello = vi.fn(() => 'hello world')
})
```

### Tests

### Documentation

### Changesets


packages/spy/src/index.ts
```diff
@@ -562,16 +585,31 @@ function enhanceSpy<T extends Procedure>(
...
+        const result = Reflect.construct(impl, args, new.target)
+        instances.push(result)
+        contexts.push(result)
+        return result
+      }
+      catch (e) {
+        instances.push(undefined)
+        contexts.push(undefined)
+        throw e
+      }
```
AriPerkkio: Can we detect the incorrect usage here? If yes, the error message could point to documentation too. 

```ts
if (e instanceof TypeError && e.message.endsWith('is not a constructor')) {
  e.message = `vi.spyOn() for class did not use 'function' or 'class', see https://vitest.dev/api/vi.html#vi-spyon for examples.\n\n${e.message}`
}
```

AriPerkkio (approved):

PR #6169, opened 2024-07-19
docs: add information to the workspace docs on how to use extends

CoolGoose:
### Description

This should make it a bit clearer on how to reuse a base config file in our workspaces since implicitly they don't inherit from anything.

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/guide/workspace.md
+++ b/docs/guide/workspace.md
@@ -172,7 +172,7 @@ export default mergeConfig(
 ```
 :::
 
-At the `defineWorkspace` level you can use the `extends` option to inherit from your root-level config.
+At the `defineWorkspace` level you can also use the `extends` option instead to inherit from your root-level config.
 ::: code-group
 ```ts [packages/a/vitest.config.ts]
 import { defineWorkspace } from 'vitest/config'
--- a/docs/guide/workspace.md
+++ b/docs/guide/workspace.md
@@ -172,31 +172,29 @@ export default mergeConfig(
 ```
 :::
 
-
 At the `defineWorkspace` level you can use the `extends` option to inherit from your root-level config.
 ::: code-group
 ```ts [packages/a/vitest.config.ts]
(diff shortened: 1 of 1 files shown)


sheremet-va (changes requested):
Please, lint your edits according to the project eslint config

CoolGoose:
@sheremet-va done senpai, didn't realize the CI broke 😅

docs/guide/workspace.md
```diff
@@ -172,6 +172,30 @@ export default mergeConfig(
 ```
 :::
 
+At the `defineWorkspace` level you can use the `extends` option to inherit from your root-level config.
```
sheremet-va, line 175: Just noticed, but how is this different from 

https://github.com/CoolGoose/vitest/blob/266330c3dee2ff55ec1fc1aad1fdc03aff2a9b6b/docs/guide/workspace.md?plain=1#L62

CoolGoose: It's not fully different, I wrote it mostly in the context of 'please make sure you understand there's no base config automatically extended when you're doing workspaces' a bit localized to the section that talks about Configuration specifically.

Looking back on how I wrote it it's a bit clunky indeed.

The problem is that conceptually, I think it was more of a case of expecting the workspaces to automatically extend the vitest.config.ts (especially in the context that not all options are available in the workspace config, so you'll need to use extends). and also the difference of `mergeConfig` for vitest and `.extends` in the workspace 😅 

Happy to just close this if you think it's just a bit more confusing.

sheremet-va: I see, I guess it makes sense to also duplicate an example with `extends` here alongside `mergeConfig` as an alternative

sheremet-va, line 175:
```suggestion
At the `defineWorkspace` level you can also use the `extends` option instead to inherit from your root-level config.
```

sheremet-va (approved):

Issue #6057, opened 2024-07-08
`hoisted` does not work with istanbul coverage when overriding coverage excludes


kwojcik:
### Describe the bug

If you use istanbul coverage (required for browser mode), override the coverage `excludes`, and use `vi.hoisted`, then your test runs fail with cryptic syntax errors.

This worked on `v2.0.0-beta.12` and is broken on `v2.0.1`.

```
 FAIL  test/basic.test.tsx [ test/basic.test.tsx ]
SyntaxError: Unexpected token ')'
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯

 Test Files  1 failed (1)
      Tests  no tests
   Start at  12:06:27
   Duration  0ms
```
```
export default defineConfig({
  test: {
    coverage: {
      provider: "istanbul",
      exclude: [],
    },
  },
});
```
```
import { vi } from "vitest";

const mocks = vi.hoisted(() => ({}));
``` 

### Reproduction

https://github.com/kwojcik/vitestsourcemapbug/tree/hoistedBug

```
git checkout hoistedBug
npm install
npm run test
```

### System Info

```shell
System:
    OS: macOS 13.6.7
    CPU: (12) arm64 Apple M2 Max
    Memory: 11.03 GB / 64.00 GB
[...]

PR #6184, opened 2024-07-21
fix(vitest): `vi.hoisted` syntax errors on instrumented code

AriPerkkio:
### Description

- Fixes https://github.com/vitest-dev/vitest/issues/6057

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/hoistMocks.ts
+++ b/packages/vitest/src/node/hoistMocks.ts
@@ -241,7 +241,8 @@ export function hoistMocks(
     if (
       init
       && (init === node
-      || (init.type === 'AwaitExpression' && init.argument === node))
+      || (init.type === 'AwaitExpression' && init.argument === node)
+      || (init.type === 'SequenceExpression' && init.expressions.includes(node)))
     ) {
       return declarationNode
     }
@@ -432,6 +433,8 @@ export function hoistMocks(
   const hoistedCode = hoistedNodes
     .map((node) => {
       const end = getBetterEnd(code, node)
+      let replaceNodesCode = ''
+
       /**
        * In the following case, we need to change the `user` to user: __vi_import_x__.user
        * So we should get the latest code from `s`.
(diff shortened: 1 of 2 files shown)


AriPerkkio:
Current implementation:

```js
import { expect, it, vi } from 'vitest'

let coverageCounter = 0
const nestedHoist = (coverageCounter++, vi.hoisted(() => {
  return `Count is ${coverageCounter}`
}))

it('nesting vi.hoisted doesnt cause syntax errors', () => {
  expect(nestedHoist).toBe('Count is 1')
})
```

becomes:

```js
const { expect, it, vi } = await __vite_ssr_dynamic_import__("/@fs/x/vitest/packages/vitest/dist/index.js")
vi.hoisted(() => {
  return `Count is ${coverageCounter}`;
})

let coverageCounter = 0;
const nestedHoist = (coverageCounter++, );                  // <-- Syntax error
it("nesting vi.hoisted doesnt cause syntax errors", () => {
  expect(nestedHoist).toBe("Count is 1");
});
```

I think correct would be

```js
const { expect, it, vi } = await __vite_ssr_dynamic_import__("/@fs/x/vitest/packages/vitest/dist/index.js")
const nestedHoist = vi.hoisted(() => {
  return `Count is ${coverageCounter}`;
})

let coverageCounter = 0;
coverageCounter++;
it("nesting vi.hoisted doesnt cause syntax errors", () => {
  expect(nestedHoist).toBe("Count is 1");
});
```

If implementing this via AST modify becomes too difficult I think we could instead add `/* istanbul ignore next */` ignore hints in mocking methods before the instrumentation.

AriPerkkio:
This is getting quite complex. I think better and more maintainable solution is to:
- Prevent `@vitest/coverage-istanbul` from instrumenting `vi.mock|hoisted|etc` method calls. This is done by adding `/* istanbul ignore next */` ignore hints before detected method calls.
- Throw an error around here if anything else than `VariableDeclaration` is found, e.g. `SequenceExpression`

https://github.com/vitest-dev/vitest/blob/073a50c9740f75ba25ec1fabba73b6dae73e66f4/packages/vitest/src/node/hoistMocks.ts#L330

AriPerkkio:
https://github.com/vitest-dev/vitest/issues/6057#issuecomment-2813403136

Issue #6185, opened 2024-07-21
Bug: `testNamePattern` option unnecessarily requires a leading space for an exact match.


senkenn:
### Describe the bug

There is a example test.
```js
it("foo", () => {
  expect(1).toBe(1);
});
```

We require a leading space for an exact match by `testNamePattern`.
```sh
vitest -t '^ foo$' # '^foo$' skips it.
```

This is also the same `describe` name.

On the other hand, Jest can run it by '^foo$'. Is this behavior as expected in Vitest?

### Reproduction

1. Create a JavaScript file like the above, and `pnpm i -D vitest`
2. Run `pnpm vitest run -t '^foo$' ` and get the skipped result.
```
❯ pnpm vitest run -t '^foo$' 
The CJS build of Vite's Node API is deprecated. See https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.

 RUN  v2.0.3 /home/foo/bar

 ↓ index.test.js (1) [skipped]
   ↓ foo [skipped]

 Test Files  1 skipped (1)
      Tests  1 skipped (1)
   Start at  01:46:52
   Duration  270ms (transform 21ms, setup 0ms, collect 5ms, tests 0ms, environment 0ms, prepare 78ms)
```
3. Run `pnpm vitest run -t '^ foo$' ` and get the success result.
```
❯ pnpm vitest run -t '^ foo$'
The CJS build of Vite's Node API is deprecated. See https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.

 RUN  v2.0.3 /home/foo/bar

 ✓ index.test.js (1)
   ✓ foo

 Test Files  1 passed (1)
      Tests  1 passed (1)
   Start at  01:50:46
   Duration  266ms (transform 19ms, setup 0ms, collect 5ms, tests 1ms, environment 0ms, prepare 68ms)
```
### System Info

```shell
System:
    OS: Linux 6.5 Ubuntu 22.04.3 LTS 22.04.3 LTS (Jammy Jellyfish)
    CPU: (16) x64 AMD Ryzen 7 5700U with Radeon Graphics
    Memory: 2.68 GB / 14.94 GB
    Container: Yes
    Shell: 5.8.1 - /usr/bin/zsh
  Binaries:
    Node: 20.14.0 - ~/.nvm/versions/node/v20.14.0/bin/node
[...]

PR #6186, opened 2024-07-21
fix: `testNamePattern` adds leading space

AriPerkkio:
### Description

- Fixes https://github.com/vitest-dev/vitest/issues/6185

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/runner/src/collect.ts
+++ b/packages/runner/src/collect.ts
@@ -77,6 +77,14 @@ export async function collectTests(
 
     calculateSuiteHash(file)
 
+    file.tasks.forEach((task) => {
+      // task.suite refers to the internal default suite object
+      // it should not be reported
+      if (task.suite?.id === '') {
+        delete task.suite
+      }
+    })
+
     const hasOnlyTasks = someTasksAreOnly(file)
     interpretTaskModes(
       file,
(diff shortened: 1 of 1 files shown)


packages/runner/src/utils/collect.ts
```diff
@@ -57,7 +57,9 @@ export function interpretTaskModes(
 }
 
 function getTaskFullName(task: TaskBase): string {
-  return `${task.suite ? `${getTaskFullName(task.suite)} ` : ''}${task.name}`
```
sheremet-va, line 60: Why is this not working? We removed the empty suite from tasks. Maybe the filtering happens too early?

AriPerkkio: Oh that's right! The empty suite removing is happening only after this. At this point we have:

```js
{
  'task.suite': {
    id: '',
    type: 'suite',
    name: '',
    mode: 'run',
    each: undefined,
    file: <ref *1> {
      id: '677341536',
      name: 'test/first.test.ts',
      type: 'suite',
      mode: 'run',
      filepath: '/x/examples/packages/vitest-example/test/first.test.ts',
      tasks: [Array],
      meta: [Object: null prototype] {},
      projectName: undefined,
      file: [Circular *1],
      pool: 'forks',
      setupDuration: 0,
      collectDuration: 21
    },
    shuffle: undefined,
    tasks: [ [Object], [Object], [Object] ],
    meta: [Object: null prototype] {},
    concurrent: undefined
  }
}
```

I'll re-order those parts and revert changes of `getTaskFullName`.

sheremet-va (approved):

PR #6227, opened 2024-07-26
docs: Update common-errors to include dynamic import aliases

kamilogorek:
Since Vite 2.0 `vi.mock` can accept `import()` as the first argument to detect which module to be mock. This however won't work with aliased paths, which can confuse some people.

I believe this is not solvable at the moment, as this is not a Vitest, but rather TypeScript shortcoming.

### DIFF
--- a/docs/api/vi.md
+++ b/docs/api/vi.md
@@ -52,6 +52,10 @@ vi.mock(import('./path/to/module.js'), async (importOriginal) => {
 
 Under the hood, Vitest still operates on a string and not a module object.
 
+If you are using TypeScript with `paths` aliases configured in `tsconfig.json` however, the compiler won't be able to correctly resolve import types.
+In order to make it work, make sure to replace all aliased imports, with their corresponding relative paths.
+Eg. use `import('./path/to/module.js')` instead of `import('@/module')`.
+
 ::: warning
 `vi.mock` is hoisted (in other words, _moved_) to **top of the file**. It means that whenever you write it (be it inside `beforeEach` or `test`), it will actually be called before that.
 
--- a/docs/guide/common-errors.md
+++ b/docs/guide/common-errors.md
@@ -43,6 +43,32 @@ export default defineConfig({
 })
 ```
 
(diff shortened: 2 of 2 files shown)


kamilogorek:
Not sure if inline diff-view can be used here, as CI fails to parse the code correctly. Might need to go back to 2 separate blocks.

sheremet-va:
This is not true, it can be used with aliased paths

kamilogorek:
Here's a repro for this issue https://github.com/kamilogorek/vitest-dynamic-import-mock/

I'm more than happy to be corrected, as it'd be lovely to use it in such form :)

image

sheremet-va:
> Here's a repro for this issue [kamilogorek/vitest-dynamic-import-mock](https://github.com/kamilogorek/vitest-dynamic-import-mock/)
> 
> I'm more than happy to be corrected, as it'd be lovely to use it in such form :)

The `Cannot find module` issue in the documentation refers to the runtime error. It is not related to types, so adding this paragraph is confusing to me - maybe the documentation does a poor job at explaining it.

The issue you are experiencing is the opposite - TypeScript shows a type error, but code works at runtime, This is related to how TypeScript works and not to how Vitest works, so I don't think it has a place in our documentation

kamilogorek:
Correct, that's why I mentioned

> this is not a Vitest, but rather TypeScript shortcoming.

I'm perfectly fine closing this PR if the team decides so, no worries :)

sheremet-va:
> Correct, that's why I mentioned
> 
> > this is not a Vitest, but rather TypeScript shortcoming.
> 
> I'm perfectly fine closing this PR if the team decides so, no worries :)

I think we can add this note to the mocking page (https://vitest.dev/api/vi.html#vi-mock) when `vi.mock(import())` is introduced

kamilogorek:
Reworded and moved to the `vi.mock` docs instead. Didnt use a `warning` block as there would be 3 consecutive warnings which looked odd.

sheremet-va (approved):

sheremet-va:
Thank you!

PR #6238, opened 2024-07-29
docs: update error.ts for inlining a package

IAmLuisJ:
### Description

Update recommended workaround for inlining a package that has a syntax error when resolving a bundled dependency, since the current workaround uses deprecated config options

close vitest-dev#6237

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/error.ts
+++ b/packages/vitest/src/node/error.ts
@@ -284,11 +284,14 @@ function printModuleWarningForPackage(
         + '\n'
         + c.green(`export default {
   test: {
-    server: {
-      deps: {
-        inline: [
-          ${c.yellow(c.bold(`"${name}"`))}
-        ]
+    deps: {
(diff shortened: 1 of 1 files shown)


sheremet-va:
`server.deps.inline` is not deprecated, `deps.inline` is

Using optimizer actually won't do anything unless it's enabled

IAmLuisJ:
Oh I see. I believe my particular use case was alias configuration in a bundled dependency, which didn't work with the recommended solution. Would it be worth extending the error then instead of replacing it?

sheremet-va:
> Oh I see. I believe my particular use case was alias configuration in a bundled dependency, which didn't work with the recommended solution. Would it be worth extending the error then instead of replacing it?

How exactly do you want to extend it?

sheremet-va:
We still recommend inlining

PR #6243, opened 2024-07-29
fix(browser): Improve unique CSS selector generation

### LINKED ISSUE #6211: Browser Interactivity API should escape CSS IDs when building selectors
### Describe the bug

The issue is here: https://github.com/vitest-dev/vitest/blob/a48be6ffc427aacccf19ed4c69d007fd0c7c8010/packages/browser/src/client/tester/context.ts#L47 — the raw ID is inserted into the CSS selector string as-is, but that may not always produce valid CSS, and causes an error during the test.

You can reproduce this issue just by defining a `<button id=":foo">` and trying to use `userEvent.click()` on it.

Almost any string can be a valid HTML ID (see [here](https://html.spec.whatwg.org/multipage/dom.html#the-id-attribute)), but that does not mean it can be used unescaped in a CSS selector. It is possible to escape characters outside the regular range by following [these instructions](https://www.w3.org/TR/CSS21/syndata.html#characters). One could write a small function, e.g., `sanitizeHtmlIdForCss(htmlId: string): string`, and just call that on `el.id` before passing it into the string interpolation in the line of code I linked to above.

If this seems worth addressing I can submit this PR myself, just let me know. Thanks!

### Reproduction

This is hard/impossible to reproduce in Stackblitz because it only applies in the browser, and I don't think Stackblitz can run those. But if you have a local project that runs tests in the browser, try creating an element with an ID of `:foo`, and call any `userEvent` method on it.

### System Info

```shell
System:
    OS: macOS 14.5
    CPU: (10) arm64 Apple M1 Max
    Memory: 653.55 MB / 64.00 GB
    Shell: 3.7.0 - /opt/homebrew/bin/fish
  Binaries:
    Node: 21.7.1 - /opt/homebrew/bin/node
    npm: 10.5.0 - /opt/homebrew/bin/npm
  Browsers:
    Chrome: 126.0.6478.183
    Safari: 17.5
    Safari Technology Preview: 18.0
  npmPackages:
    @vitejs/plugin-react: ^4.2.1 => 4.2.1
    @vitest/browser: ^2.0.4 => 2.0.4
    vite: ^5.3.4 => 5.3.4
    vitest: ^2.0.4 => 2.0.4
```

### Used Package Manager

npm

### Validations

zacharyvoase:
Fixes #6211 

### Description

Currently the User Interactivity API, which is required to interact with any elements via the browser, generates a unique CSS selector based on the element requested. However, there are two main cases where this fails:

* When an element on the path (anywhere from the top-level HTML down to the target element itself) has an ID containing special characters that need escaping, such as `:`; and
* When an element on the path has an ID containing uppercase letters. This is because CSS selectors are case-sensitive for HTML IDs and class names, yet the entire selector is currently lowercased before being sent to the browser engine.

I've added a test here that fails with the current code, and passes with the changes.

### Tests

### Documentation

### Changesets


zacharyvoase:
Here's the error before the fix is applied:

```
 FAIL  test/userEvent.test.ts > userEvent.click > clicks a button with complex HTML ID
Error: locator.click: Unexpected token "#" while parsing selector "html > body > #:r3: > #a:button"
Call log:
  - waiting for [data-vitest="true"] >> internal:control=enter-frame >> css=html > body > #:r3: > #a:button
```

zacharyvoase:
I'm not sure what happened to the macos-14 webkit test, it looks like it may have been a timeout?

sheremet-va (approved):

sheremet-va:
Thank you! The test on Windows is flaky, ignore it for now

sheremet-va:
Note: in the future we will be using Playwright locators directly: https://github.com/vitest-dev/vitest/pull/6084

PR #6267, opened 2024-08-02
fix: don't panic when coverage.reporter is a string

### LINKED ISSUE #6260: `coverage.reporter` with a single value causes error while running tests without coverage (v2.0.x)
### Describe the bug

The documentation of `coverage.reporter` on both [V2](https://vitest.dev/config/#coverage-reporter) and [V1](https://v1.vitest.dev/config/#coverage-reporter) say that we can use `string` as a valid value.

So a minimal `vitest.config` like this, should work
```js
export default defineConfig({
  test: {
    coverage: {
      reporter: 'json',
    },
  },
})
```
This works fine with 1.6, but not in 2.0.x.

Error printed while executing `pnpm run test`. The error **is not** present while running `pnpm run test --coverage`

Screenshot 2024-08-01 at 18 22 23

### Reproduction

A minimal reproduction case in this repository [ildede/vitest-single-reporter-error](https://github.com/ildede/vitest-single-reporter-error)

### System Info

```shell
System:
    OS: macOS 14.5
    CPU: (8) arm64 Apple M1 Pro
    Memory: 116.05 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.9.0 - ~/Library/pnpm/node
    npm: 10.1.0 - ~/Library/pnpm/npm
    pnpm: 9.5.0 - ~/Library/pnpm/pnpm
  Browsers:
    Chrome: 127.0.6533.88
    Safari: 17.5
  npmPackages:
    vitest: ^2.0.5 => 2.0.5
```

### Used Package Manager

pnpm

### Validations

sheremet-va:
### Description

Fixes #6260

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/config/resolveConfig.ts
+++ b/packages/vitest/src/node/config/resolveConfig.ts
@@ -24,6 +24,7 @@ import { RandomSequencer } from '../sequencers/RandomSequencer'
 import type { BenchmarkBuiltinReporters } from '../reporters'
 import { builtinPools } from '../pool'
 import type { Logger } from '../logger'
+import type { BaseCoverageOptions, CoverageReporterWithOptions } from '../types/coverage'
 
 function resolvePath(path: string, root: string) {
   return normalize(
@@ -229,7 +230,7 @@ export function resolveConfig(
     )
   }
 
-  resolved.coverage.reporter = toArray(resolved.coverage.reporter)
+  resolved.coverage.reporter = resolveCoverageReporters(resolved.coverage.reporter)
 
   if (resolved.coverage.enabled && resolved.coverage.reportsDirectory) {
     const reportsDirectory = resolve(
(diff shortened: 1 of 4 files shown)


AriPerkkio (dismissed):
Sure, this will fix the crash but `htmlReporter` resolving parts will cause false results. Both `['html', 'json']` and `[['html', { subdir: 'tmp' }], 'json']` will pass the `toArray` check. Reporter resolving will then break for one of the cases:

https://github.com/vitest-dev/vitest/blob/081cfe039955f11102bd116e260a062afff324b0/packages/vitest/src/node/config/serializeConfig.ts#L49-L53

I started to do similar fix yesterday (https://github.com/AriPerkkio/vitest/commit/b1ad1e52d6e45af1a8a993ca64b4b12324a4cc8d) but realized that this kind of fixes should not be needed in the first place if typings were correct. Reporter config should be normalized when configuration is resolved.

sheremet-va:
> I started to do similar fix yesterday ([AriPerkkio@b1ad1e5](https://github.com/AriPerkkio/vitest/commit/b1ad1e52d6e45af1a8a993ca64b4b12324a4cc8d)) but realized that this kind of fixes should not be needed in the first place if typings were correct. Reporter config should be normalized when configuration is resolved.

Ok, I see, I moved `resolveReporters` to `resolveConfig` now without removing the call from `coverage` packages (to not accidentally break someone yet because of the usual package manager shenanigans - maybe with 2.1). I am still planning to refactor the coverage resolution in the future.

AriPerkkio (approved):
> maybe with 2.1

Sounds good!

PR #6273, opened 2024-08-04
feat(browser): support v8 coverage

AriPerkkio:
### Description

- Adds support for using V8 coverage in Browser mode
- Works only on Playwright + Chromium at the moment. We can later extend support for other providers and chromium based browsers.

```ts

export default defineConfig({
  test: {
    coverage: {
      provider: "v8",
    },
    browser: {
      provider: "playwright",
      name: "chromium",
    },
  },
});
```

### Tests

### Documentation

### Changesets


test/coverage-test/vitest.workspace.custom.ts
```diff
@@ -78,6 +78,29 @@ export default defineWorkspace([
...
+        '**/import-attributes.test.ts',
+        '**/multi-suite.test.ts',
+        '**/setup-files.test.ts',
+        '**/results-snapshot.test.ts',
+        '**/reporters.test.ts',
+        '**/temporary-files.test.ts',
+        '**/test-reporter-conflicts.test.ts',
+
+        // TODO: Enable, last source map issue here
+        // '**/vue.test.ts',
```
AriPerkkio, line 100: Need to figure out what's going on here. The `Defined.vue` is the only file that is transformed differently on Node vs Browser.

https://github.com/vitest-dev/vitest/blob/7b37d27ba7ce9e75fce385126e61d85d575f82ad/test/coverage-test/fixtures/src/Vue/Defined.vue#L1-L22

On browser mode it returns:

```js
import { defineComponent as _defineComponent } from "/node_modules/.vite/deps/vue.js?v=256f6d3b";
const defined = "hello";
const _sfc_main = /* @__PURE__ */ _defineComponent({
  __name: "Defined",
  setup(__props, { expose: __expose }) {
    __expose();
    if (defined) {
      console.log("Covered condition");
    } else {
      console.log("Uncovered condition");
    }
    const __returned__ = { defined };
    Object.defineProperty(__returned__, "__isScriptSetup", { enumerable: false, value: true });
    return __returned__;
  }
});
import { toDisplayString as _toDisplayString } from "/node_modules/.vite/deps/vue.js?v=256f6d3b";
function _sfc_render(_ctx, _cache, $props, $setup, $data, $options) {
  return _toDisplayString($setup.defined);
}
import "/fixtures/src/Vue/Defined.vue?vue&type=style&index=0&scoped=f7f04e08&lang.css";
import _export_sfc from "/@id/__x00__plugin-vue:export-helper";
export default /* @__PURE__ */ _export_sfc(_sfc_main, [["render", _sfc_render], ["__scopeId", "data-v-f7f04e08"], ["__file", "/Users/x/vitest/test/coverage-test/fixtures/src/Vue/Defined.vue"]]);
```

On Node with `happy-dom` it returns:
```js
const __vite_ssr_import_0__ = await __vite_ssr_import__("/@fs/Users/x/vitest/node_modules/.pnpm/vue@3.4.35_typescript@5.5.4/node_modules/vue/index.mjs", {"importedNames":["defineComponent"]});
const __vite_ssr_import_1__ = await __vite_ssr_import__("/@fs/Users/x/vitest/node_modules/.pnpm/vue@3.4.35_typescript@5.5.4/node_modules/vue/index.mjs", {"importedNames":["toDisplayString"]});
const __vite_ssr_import_2__ = await __vite_ssr_import__("/fixtures/src/Vue/Defined.vue?vue&type=style&index=0&scoped=f7f04e08&lang.css");
const __vite_ssr_import_3__ = await __vite_ssr_import__("/@id/__x00__plugin-vue:export-helper", {"importedNames":["default"]});

const defined = "hello";
const _sfc_main = /* @__PURE__ */ __vite_ssr_import_0__.defineComponent({
  __name: "Defined",
  setup(__props, { expose: __expose }) {
    __expose();
    if (defined) {
      console.log("Covered condition");
    } else {
      console.log("Uncovered condition");
    }
    const __returned__ = { defined };
    Object.defineProperty(__returned__, "__isScriptSetup", { enumerable: false, value: true });
    return __returned__;
  }
});

function _sfc_render(_ctx, _cache, $props, $setup, $data, $options) {
  return __vite_ssr_import_1__.toDisplayString($setup.defined);
}


__vite_ssr_exports__.default = /* @__PURE__ */ __vite_ssr_import_3__.default(_sfc_main, [["render", _sfc_render], ["__scopeId", "data-v-f7f04e08"], ["__file", "/Users/x/vitest/test/coverage-test/fixtures/src/Vue/Defined.vue"]]);

//# sourceMappingURL=data:application/json;charset=utf-8;base64,eyJ2ZXJzaW9uIjozLCJtYXBwaW5ncyI6IkFBQUEsQ0FBQTs7Ozs7QUFDQSxNQUFNLFVBQVU7Ozs7O0FBR2hCLFFBQUksU0FBUztBQUNYLGNBQVEsSUFBSSxtQkFBbUI7QUFBQSxJQUNqQyxPQUNLO0FBQ0gsY0FBUSxJQUFJLHFCQUFxQjtBQUFBLElBQ25DOzs7Ozs7OzsrQ0FJSyxjQUFPIiwibmFtZXMiOltdLCJpZ25vcmVMaXN0IjpbXSwic291cmNlcyI6WyJEZWZpbmVkLnZ1ZSJdLCJzb3VyY2VzQ29udGVudCI6WyI8c2NyaXB0IHNldHVwIGxhbmc9XCJ0c1wiPlxuY29uc3QgZGVmaW5lZCA9ICdoZWxsbydcblxuLyogZXNsaW50LWRpc2FibGUgbm8tY29uc29sZSAqL1xuaWYgKGRlZmluZWQpIHtcbiAgY29uc29sZS5sb2coJ0NvdmVyZWQgY29uZGl0aW9uJylcbn1cbmVsc2Uge1xuICBjb25zb2xlLmxvZygnVW5jb3ZlcmVkIGNvbmRpdGlvbicpXG59XG48L3NjcmlwdD5cblxuPHRlbXBsYXRlPlxuICB7eyBkZWZpbmVkIH19XG48L3RlbXBsYXRlPlxuXG48IS0tIFN0eWxlIGJsb2NrIHRyaWdnZXJzIGEgc3BlY2lmaWMgY29uZGl0aW9uIHdoZXJlIHNvdXJjZW1hcHMgdXNlZCB0byBicmVhayByYW5kb21seSAtLT5cbjxzdHlsZSBsYW5nPVwiY3NzXCIgc2NvcGVkPlxuYm9keSB7XG4gIGJhY2tncm91bmQtY29sb3I6ICNGRkY7XG59XG48L3N0eWxlPlxuIl0sImZpbGUiOiIvZml4dHVyZXMvc3JjL1Z1ZS9EZWZpbmVkLnZ1ZSJ9',
```

That looks like SSR transformed even though `happy-dom` should set `transformMode: 'web'` 🤔

packages/vitest/src/node/config/resolveConfig.ts
```diff
@@ -220,16 +220,6 @@ export function resolveConfig(
...
-  if (
-    resolved.coverage.provider === 'v8'
-    && resolved.coverage.enabled
-    && isBrowserEnabled(resolved)
-  ) {
-    throw new Error(
-      '@vitest/coverage-v8 does not work with --browser. Use @vitest/coverage-istanbul instead',
-    )
-  }
-
```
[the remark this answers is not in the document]
userquin: Shouldn't match, each browser has its own name in PW and WDIO.

Edge is now chromium based and CDP should be available.

PW now supports Edge via `msedge`, WDIO using `edge`: https://playwright.dev/docs/browsers

We'll need to review the logic, check browser names here:
https://github.com/vitest-dev/vitest/blob/7b37d27ba7ce9e75fce385126e61d85d575f82ad/packages/vitest/src/create/browser/creator.ts#L33-L42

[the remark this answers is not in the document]
userquin: We should also check https://playwright.dev/docs/browsers#chromium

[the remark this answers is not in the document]
AriPerkkio: Looks like currently CDP is only supported with `playwright` + `chromium` in Vitest. All other combinations throw runtime errors when `getCDPSession()` is called.

So I'll add here an check that verifies **built-in** providers:
- Throw if provider is `preview` or `webdriverio`
- Throw if provider is `playwright` and not `chromium`

packages/coverage-v8/src/browser.ts
```diff
@@ -0,0 +1,69 @@
...
+    return loadProvider()
+  },
+}
+
+function filterResult(coverage: ScriptCoverage['result'][number]): boolean {
+  if (!coverage.url.startsWith(window.location.origin)) {
+    return false
+  }
+
+  if (coverage.url.includes('@fs')) {
```
sheremet-va, line 52: Why are these excluded?

packages/coverage-v8/src/provider.ts
```diff
@@ -497,27 +486,41 @@ export class V8CoverageProvider extends BaseCoverageProvider implements Coverage
...
+      ? project.vitenode.fetchCaches[transformMode === 'browser' ? 'web' : transformMode]
+      : project.vitenode.fetchCache
     const transformResults = normalizeTransformResults(fetchCache)
 
-    const scriptCoverages = coverage.result.filter(result =>
-      this.testExclude.shouldInstrument(fileURLToPath(result.url)),
-    )
+    async function onTransform(filepath: string) {
+      if (transformMode === 'browser' && project.browser) {
+        const result = await project.browser.vite.transformRequest(filepath)
```
sheremet-va, line 499: If filepath is absolute, this will load a duplicate module because IDs are different - if you want to take the value from the cache, the path should be relative to root (`/src/t.ts` instead of `/home/project/src/t.ts`)

packages/coverage-v8/src/provider.ts
```diff
@@ -497,27 +486,41 @@ export class V8CoverageProvider extends BaseCoverageProvider implements Coverage
...
+    project: WorkspaceProject = this.ctx.getCoreWorkspaceProject(),
+    transformMode?: keyof CoverageFilesByTransformMode,
   ): Promise<CoverageMap> {
-    const viteNode
-      = this.ctx.projects.find(project => project.getName() === projectName)
-        ?.vitenode || this.ctx.vitenode
     const fetchCache = transformMode
-      ? viteNode.fetchCaches[transformMode]
-      : viteNode.fetchCache
+      ? project.vitenode.fetchCaches[transformMode === 'browser' ? 'web' : transformMode]
```
sheremet-va, line 493: `fetchCache` is not populated for the browser, this assumes that jsdom test and browser tests have the same transforms, but browser tests don't apply the SSR transform (`__vite_ssr_1__` and such)

sheremet-va (changes requested):

packages/coverage-v8/src/provider.ts
AriPerkkio: Yup, for browser mode there won't be matches for `fetchCache`. Maybe it's safer to add empty/dummy fetchCache here so that they don't mix when browser mode is used via `poolMatchGlobs`.

packages/coverage-v8/src/browser.ts
AriPerkkio: To exclude files like below:

```
http://localhost:5173/@fs/Users/x/vitest/packages/browser/dist/client.js
http://localhost:5173/@fs/Users/x/vitest/packages/browser/dist/client/__vitest_browser__/preload-helper-Cyh_YkkL.js
```

sheremet-va: But it's not used only for those things. It's used by Vite for other files. And you already exclude `__vitest_browser__` later. The `browser/dist` will be in `node_modules` so it will also be excluded

packages/coverage-v8/src/provider.ts
sheremet-va: Yeah, I don't think using web is safe enough

packages/coverage-v8/src/provider.ts
AriPerkkio: Good catch! Added change to use root relative paths here.

packages/coverage-v8/src/browser.ts
AriPerkkio: Alright, removing this should be fine. Though I'd like to add test case that ensures paths with `@fs` are handled correctly when remapping coverage. Any ideas how to create such test case? When would `@fs` appear in source file imports?

AriPerkkio: Files loaded in `test/allow-external.test.ts` seem to trigger this case - perfect! I'll add the required changes and enable that test for browser mode.

packages/coverage-v8/src/provider.ts
```diff
@@ -443,16 +437,11 @@ export class V8CoverageProvider extends BaseCoverageProvider implements Coverage
...
-      | Awaited<ReturnType<typeof this.ctx.vitenode.transformRequest>>
-      = transformResults.get(filePath)
+    let transformResult: FetchResult | TransformResult | undefined = transformResults.get(filePath)
 
     if (!transformResult) {
       isExecuted = false
-      transformResult = await this.ctx.vitenode
-        .transformRequest(filePath)
-        .catch(() => null)
+      transformResult = await onTransform(url.replace('file://', '')).catch(() => undefined)
```
sheremet-va, line 444: I am a bit concerned with transforms like `replace('a', 'b')`. Can we use `fileURLToPath` and `pathToFileURL`?

packages/coverage-v8/src/provider.ts
```diff
@@ -497,27 +486,49 @@ export class V8CoverageProvider extends BaseCoverageProvider implements Coverage
...
+    }
+
     const transformResults = normalizeTransformResults(fetchCache)
 
-    const scriptCoverages = coverage.result.filter(result =>
-      this.testExclude.shouldInstrument(fileURLToPath(result.url)),
-    )
+    async function onTransform(filepath: string) {
+      if (transformMode === 'browser' && project.browser) {
+        const result = await project.browser.vite.transformRequest(filepath.replace(project.config.root, ''))
```
sheremet-va, line 502: Can we have a utility with something like this? (This is from Vite source)

```ts
if (filepath.startsWith(withTrailingSlash(root))) {
    // in root: infer short absolute path from root
    url = filepath.slice(root.length)
  }
```

packages/coverage-v8/src/provider.ts
```diff
@@ -497,27 +486,49 @@ export class V8CoverageProvider extends BaseCoverageProvider implements Coverage
...
+      }
+      return project.vitenode.transformRequest(filepath)
+    }
+
+    const scriptCoverages = []
+
+    for (const result of coverage.result) {
+      if (transformMode === 'browser') {
+        if (result.url.includes('@fs/')) {
+          result.url = `file://${result.url.replace('@fs/', '')}`
```
sheremet-va, line 516: Can we use `pathToFileURL` instead? In that case, we also need to replace `/@fs/` with `isWindows ? '' : '/'`

packages/coverage-v8/src/provider.ts
```diff
@@ -497,27 +486,49 @@ export class V8CoverageProvider extends BaseCoverageProvider implements Coverage
...
+        }
+        else {
+          result.url = `file://${project.config.root}${result.url}`
```
sheremet-va, line 519: `pathToFileURL(join(project.config.root, result.url))` - join doesn't behave like `resolve` and will keep all segments

sheremet-va (dismissed):

packages/coverage-v8/src/browser.ts
```diff
@@ -0,0 +1,65 @@
+import { cdp } from '@vitest/browser/context'
+import type { V8CoverageProvider } from './provider'
+import { loadProvider } from './load-provider'
+
+interface ScriptCoverage {
+  result: Array<{ url: string }>
+}
+
+const session = cdp() as { send: (...options: any[]) => any }
```
sheremet-va, line 9: By the way, maybe this type casting is not needed if you add `@vitest/browser/playwright` to coverage-v8 tsconfig?

packages/coverage-v8/src/provider.ts
AriPerkkio: In this specific case we cannot. Here the `url` is a String and may contain query parameters, e.g. in Vue. Using `fileURLToPath` would remove those. If `url` was an instance of `URL` instead of String we could access the `.search` but it's just String here.

```sh
$ pnpm run test --project v8-browser test/vue.test.ts
```

```js
console.log(url)
> 'file:///Users/x/vitest/test/coverage-test/fixtures/src/Vue/Defined.vue?vue&type=style&index=0&scoped=f7f04e08&lang.css',
 
console.log(fileURLToPath(url))
> '/Users/x/vitest/test/coverage-test/fixtures/src/Vue/Defined.vue'

console.log(url.replace('file://', ''))
> '/Users/x/vitest/test/coverage-test/fixtures/src/Vue/Defined.vue?vue&type=style&index=0&scoped=f7f04e08&lang.css'
> ^^This is what we want
```

sheremet-va: Can we slice it then instead of replacing?

AriPerkkio: Sure! I guess `url.slice(7)` (7 for `file://`.length) would be faster than find-and-replace.

packages/coverage-v8/src/provider.ts
AriPerkkio: `pathToFleURL` escapes query parameters so it cannot be used here. 

```sh
$ node
> url.pathToFileURL("/fixtures/src/Vue/Counter/Counter.component.ts?vue&type=script&src=true&lang.ts")

URL {
  href: 'file:///fixtures/src/Vue/Counter/Counter.component.ts%3Fvue&type=script&src=true&lang.ts',
//                                                            ^^
  pathname: '/fixtures/src/Vue/Counter/Counter.component.ts%3Fvue&type=script&src=true&lang.ts',
  search: '',
}
```

And `join('file://', project.config.root, result.url)` doesn't work either as it converts to `file:/path` instead of `file:///path`

packages/coverage-v8/src/provider.ts
AriPerkkio: Same reason as below - `pathToFleURL` escapes query parameters.

packages/coverage-v8/src/provider.ts
AriPerkkio: `removeRoot` utility added and used in all cases where `.replace` was.

packages/coverage-v8/src/provider.ts
```diff
@@ -443,16 +438,11 @@ export class V8CoverageProvider extends BaseCoverageProvider implements Coverage
...
-      | Awaited<ReturnType<typeof this.ctx.vitenode.transformRequest>>
-      = transformResults.get(filePath)
+    let transformResult: FetchResult | TransformResult | undefined = transformResults.get(filePath)
 
     if (!transformResult) {
       isExecuted = false
-      transformResult = await this.ctx.vitenode
-        .transformRequest(filePath)
-        .catch(() => null)
+      transformResult = await onTransform(removeRoot(url, FILE_PROTOCOL)).catch(() => undefined)
```
sheremet-va, line 445: Sorry for being pedantic about this, but the function is called `removeRoot` 🤔 This feels a bit confusing to me that you pass down file protocol then 🤔

AriPerkkio: No worries! Any ideas for better name? `removeStartsWith`, `trimStartsWith`, `removeIfStartsWith`?

sheremet-va: `removeStartsWith` sounds fine to me

sheremet-va (dismissed):

sheremet-va (approved):

Issue #6147, opened 2024-07-16
Improve Vitest Public API


sheremet-va:
### Clear and concise description of the problem

When working with the current public API, these are the pain points that I encountered:

- impossible to watch specific files or tests (possible by overriding private methods with a getter)
- impossible to run tests with/without coverage, the value is resolved in the config (possible by overriding with a getter)
- impossible to run specific tests even if I know its `id` I have to construct a long `testNamePattern` regexp
- project API is slapped on top of task API without much thought (sorry my bad), it can be hard to associate a task and a project it was running in
- impossible to run files only in specific projects (`vitest.start` only accepts a file filter)
- `collect` method accepts only file filters, and can't pass down a project
- reporter API is hard to understand: no clear lifecycle, working with tasks is unintuitive

### Suggested solution

Develop a separate public API or improve our internal API to fix listed pain points.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

sheremet-va:
All points are addressed at this point. See https://vitest.dev/advanced/api/ for the comprehensive documentation on the public API.

PR #6281, opened 2024-08-05
feat(ui): add dark theme to istambul html reporter

userquin:
### Description

This PR adds custom dark theme css for istambul html reporter modified from this https://userstyles.world/style/14631/istanbul-lcov-material-theme-darker

Check also https://github.com/istanbuljs/istanbuljs/pull/553#issuecomment-1934930904

I need to override the css files when using standalone, rn only work when running Vitest UI (shouldn't work when running `npx server coverage`)

imagen

imagen

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/ui/node/utils.ts
+++ b/packages/ui/node/utils.ts
@@ -1,0 +1,22 @@
+export function transformCoverageEntryPoint(
+  entryPoint: string,
+) {
+  return entryPoint.replace('</head>', `<script>
+(() => {
+ const theme = 'vueuse-color-scheme'
+ const preference = localStorage.getItem(theme)
+ const prefersDark = window.matchMedia('(prefers-color-scheme: dark)').matches
+ if (!preference || preference === 'auto' ? prefersDark : preference === 'dark')
+   document.documentElement.classList.add('dark')
+
+ window.addEventListener('storage', (e) => {
+   if (e.key === theme) {
+     document.documentElement.classList.remove('dark')
+     if (!e.newValue || e.newValue === 'auto' ? prefersDark : e.newValue === 'dark')
+        document.documentElement.classList.add('dark')
+   }
+ })
+})();
+</script>
+</head>`)
+}
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -31,7 +31,9 @@
   "types": "./dist/index.d.ts",
   "files": [
     "*.d.ts",
-    "dist"
+    "dist",
+    "istambul-base.css",
+    "istambul-prettify.css"
   ],
   "scripts": {
     "build": "rimraf dist && pnpm build:node && pnpm build:client",
--- a/examples/basic/package.json
+++ b/examples/basic/package.json
@@ -7,6 +7,8 @@
   "scripts": {
     "test": "vitest",
     "test:ui": "vitest --ui",
+    "test:ui:coverage": "vitest --ui --open --coverage.enabled=true --coverage.reporter=html",
+    "test:ui:coverage:html": "vitest --ui --open --coverage.enabled=true --coverage.reporter=html --reporter=html",
     "test:run": "vitest run"
   },
   "devDependencies": {
--- a/examples/basic/src/basic.ts
+++ b/examples/basic/src/basic.ts
@@ -1,1 +1,3 @@
 export const squared = (n: number) => n * n
+
+export const doubleNumber = (n: number) => n * 2
--- a/examples/basic/src/uncovered.ts
+++ b/examples/basic/src/uncovered.ts
@@ -1,0 +1,1 @@
+export const squareRoot = (n: number) => Math.sqrt(n)
--- a/packages/ui/client/components/Coverage.vue
+++ b/packages/ui/client/components/Coverage.vue
@@ -10,7 +10,7 @@ defineProps<{
       <div class="i-carbon:folder-details-reference" />
       <span pl-1 font-bold text-sm flex-auto ws-nowrap overflow-hidden truncate>Coverage</span>
     </div>
-    <div flex-auto py-1 bg-white>
+    <div flex-auto bg-transparent>
       <iframe id="vitest-ui-coverage" :src="src" />
     </div>
   </div>
--- a/packages/ui/istambul-base.css
+++ b/packages/ui/istambul-base.css
@@ -1,0 +1,412 @@
+:root {
+  --color-text-light: #000;
+  --color-text-dark: #ddd;
+  --color-text: var(--color-text-light);
+  --background-color: white;
+  --anchor-color: #6b38fb;
+}
+
+html.dark {
+  --color-text: var(--color-text-dark);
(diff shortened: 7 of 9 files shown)


userquin:
Coverage including html folder when running standalone report (`--reporter=html`), no idea what's happening. 

Added `onFinishedReportCoverage` to the html reporter to allow copy/paste custom css after coverage finish (update vitest core to call it), cannot use `onFinished` since the coverage run after it.

AriPerkkio (commented):
Instead of making this UI specific, could we jus create a [custom coverage reporter](https://vitest.dev/guide/coverage.html#custom-coverage-reporter) that extends Istanbul's `html` reporter with theme toggle? And when ever user defines `html` reporter, we use that custom built-in one instead?

userquin:
Html reporter is the custom reporter you are asking for. Vitest will check for html reporter in the configuration to include the reporter from the ui package: check coverage docs (for example, will not work with html-spa reporter).

This reporter is only used with Vitest UI or Browser Mode and will work only with the istambul html reporter (both istambul and v8 using this reporter).

IIRC Vlad asked in discord about changing the reporter name to web.

packages/ui/package.json
```diff
@@ -31,7 +31,9 @@
   "types": "./dist/index.d.ts",
   "files": [
     "*.d.ts",
-    "dist"
+    "dist",
+    "istambul-base.css",
+    "istambul-prettify.css"
```
userquin, line 36: rename these files to `istanbul-*.css`

packages/ui/node/utils.ts
```diff
@@ -0,0 +1,80 @@
...
+}
+
+export async function prepareReportCoverageFolder(ctx: Vitest) {
+  const [folder] = resolveCoverageFolder(ctx)!
+  const root = resolve(fileURLToPath(import.meta.url), '../')
+  await Promise.all([
+    fs.copyFile(resolve(root, '../istambul-base.css'), resolve(folder, 'base.css')),
+    fs.copyFile(resolve(root, '../istambul-prettify.css'), resolve(folder, 'prettify.css')),
+  ])
+  await Promise.all(fg.sync('**/*.html', { cwd: folder, absolute: true }).map(async (file) => {
```
userquin, line 39: add the list to previous `Promise.all`

AriPerkkio:
> Html reporter is the custom reporter you are asking for. Vitest will check for html reporter in the configuration to include the reporter from the ui package: check coverage docs (for example, will not work with html-spa reporter).
> 
> This reporter is only used with Vitest UI or Browser Mode and will work only with the istambul html reporter (both istambul and v8 using this reporter).

Sorry I'm not sure I'm following here. Istanbul's [`html`](https://github.com/istanbuljs/istanbuljs/tree/main/packages/istanbul-reports/lib/html) reporter is used to generate the report that Vitest UI shows. This PR wants to add custom CSS/JS for that report.

Why does this need to be Vitest UI specific only?

> IIRC Vlad asked in discord about changing the reporter name to web.

Sure, we can do that. For the public API we would require users to specify `web` reporter, and then internally convert that to Istanbul's `html` reporter.

userquin:
> Sorry I'm not sure I'm following here. Istanbul's [`html`](https://github.com/istanbuljs/istanbuljs/tree/main/packages/istanbul-reports/lib/html) reporter is used to generate the report that Vitest UI shows. This PR wants to add custom CSS/JS for that report.

Vitest UI will show the coverage link in the UI only when html reporter added to the configuration and `coverage.enabled` flag enabled. This PR changes the css styles and the html pages generated by Istanbul html reporter to support dark theme inside Vitest UI (rn, when switching Vitest UI to dark, the html from the coverage folder doesn't support dark theme).

This PR is about integrating Istanbul html reporter in Vitest UI: it is out of scope how can we integrate dark theme outside Vitest UI. For example, some users may want to use system theme, others may want dark or light only (with a custom color palette like https://userstyles.world/style/14631/istanbul-lcov-material-theme-darker). 

Vitest UI using vueuse to switch between dark and light themes, this PR adds the corresponding script to all html pages to access and watch the corresponding local storage theme key and updates the css files to match Vitest UI color palette.

Check linked Istanbul issue, there is a PR to add system theme to the html reporter.

NOTE: the Vitest UI html reporter also generating standalone output when enabling it via `--reporter=html` (check last tip in the docs: https://vitest.dev/guide/ui.html#vitest-ui).

packages/vitest/src/node/core.ts
```diff
@@ -1097,6 +1097,9 @@ export class Vitest {
         if (reporter instanceof WebSocketReporter) {
           reporter.onFinishedReportCoverage()
         }
+        else if ('onFinishedReportCoverage' in reporter && typeof reporter.onFinishedReportCoverage === 'function') {
+          await reporter.onFinishedReportCoverage()
```
userquin, line 1101: should we find a better way to do this?

packages/ui/node/reporter.ts
```diff
@@ -59,6 +60,10 @@ export default class HTMLReporter implements Reporter {
     this.start = Date.now()
   }
 
+  async onFinishedReportCoverage() {
+    await prepareReportCoverageFolder(this.ctx)
+  }
+
   async onFinished() {
     const result: HTMLReportData = {
```
userquin, line 68: This logic should be moved to `onFinishedReportCoverage` since the coverage still not generated:
- running the standalone html reporter script will include ui client in the coverage report
- we should remove the html folder on every run using standalone html reporter, otherwise the ui client from previous run will be included in the coverage report

Running this script from `examples/basic` folder:
`pnpm vitest --ui --open --coverage.enabled=true --coverage.reporter=html --reporter=html --coverage.reportsDirectory=./html/coverage/`

Running first time standalone html reporter (all good):
imagen

Running second time without cleanup html folder:
imagen

packages/vitest/src/node/core.ts
hi-ogawa: Is it possible to check `reporter instanceof HTMLReporter` just like `WebSocketReporter` above?

In https://github.com/vitest-dev/vitest/pull/5242#discussion_r1495440749, we discussed and we didn't go with `onFinishedReportCoverage` at `interface Reporter` level for now (though it might change in reporter API rework https://github.com/vitest-dev/vitest/issues/6147).

sheremet-va: `HTMLReporter` is defined in `@vitest/ui`, not in `vitest`, so I don't think we can do that

userquin: How about adding a new type like BuildReporter?

sheremet-va (changes requested):
Dev UI seems to work properly, but HTML reporter just shows a blank white page. (Note that I am also editing the html reporter in https://github.com/vitest-dev/vitest/pull/7953)

osulli:
Any chance on a revival of this @userquin @sheremet-va?

PR #6284, opened 2024-08-06
fix: allow inlining vite's cached dependencies

chriswheeldon-peakon:
### Description

Allow the user to explicitly inline vite's cached dependencies. Either by specifying `inline: true` or providing a path that matches the vite cache directory.

This allows e.g. mocking calls made by a cached dependency.

This was briefly discussed [here on discord](https://discord.com/channels/917386801235247114/1263777870455640076).

### Tests

### Documentation

### Changesets

### DIFF
--- a/test/vite-node/test/server.test.ts
+++ b/test/vite-node/test/server.test.ts
@@ -229,3 +229,55 @@ describe('server correctly caches data', () => {
     expect(webFiles).toHaveLength(1)
   })
 })
+
+describe('externalize', () => {
+  describe('by default', () => {
+    test('should externalize vite\'s cached dependencies', async () => {
+      const vnServer = new ViteNodeServer({
+        config: {
+          root: '/',
+          cacheDir: '/node_modules/.vite',
+        },
+      } as any, {})
+
+      const externalize = await vnServer.shouldExternalize('/node_modules/.vite/cached.js')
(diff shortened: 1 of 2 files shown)


test/vite-node/test/server.test.ts
```diff
@@ -229,3 +229,55 @@ describe('server correctly caches data', () => {
...
+        },
+      } as any, {})
+
+      const externalize = await vnServer.shouldExternalize('/node_modules/.vite/cached.js')
+      expect(externalize).toBeTruthy()
+    })
+  })
+
+  describe('with server.deps.inline: true', () => {
+    test('should not externalize vite\'s cached dependencies', async () => {
```
chriswheeldon-peakon, line 249: @sheremet-va, what do you think about `inline: true` also inlining the cached dependencies? I worry that this makes this change potentially much more impactful for existing users of vitest who use `inline: true`. I could perhaps make it so when evaluating if the id matches the cache dir we only consider paths and patterns from `deps.inline` i.e. require the user to be explicit about inlining the cache deps.

sheremet-va: The only impact they'd have is slower tests. `inline: true` is already very slow by design, so it should be fine

sheremet-va (approved):

PR #6294, opened 2024-08-07
chore: remove unused execa devDependencies

benmccann:
### Description

Removes a few usages of execa from the project to make an eventual transition to `tinyexec` easier

### Tests

### Documentation

### Changesets


sheremet-va:
Any reason why? We don't want to upgrade execa because `@antfu/install-pkg` in our pinned version relies on the same version so updating it will cause two execas to be bundled, and we are waiting on using [`tinyexec`](https://github.com/tinylibs/tinyexec):  https://github.com/vitest-dev/vitest/pull/5713

packages/vitest/src/node/git.ts
```diff
@@ -12,7 +12,7 @@ export class VitestGit {
   constructor(private cwd: string) {}
 
   private async resolveFilesWithGitCommand(args: string[]): Promise<string[]> {
-    let result: ExecaReturnValue
+    let result: Result
```
AriPerkkio, line 15: This should narrow doens the type of `result.stdout` below to `string|undefined`:

```suggestion
    let result: Result<{}>
```

benmccann:
Ok, great! Thanks for those added details. I've reverted the upgrade, but left this PR open with the other commit it contained which removes a few unused instances of `execa`. I've update the title and description of the PR accordingly. This PR should make the eventual planned transitions to `ezspawn` or `tinyexec` easier

benmccann:
Hmm. It's weird that the CI is failing. It was passing with my prior commit. It's complaining about declining test coverage for a vue test and I haven't touched that. This looks like a problem with the CI setup rather than the PR

AriPerkkio:
Not sure what's going on there. The error itself is not related to coverage at all. All `expect` assertions in this test fixture failed. 

https://github.com/vitest-dev/vitest/blob/eae3bf4a72c1ab0b0b8ea0e13a3ed8a415ebfc66/test/coverage-test/fixtures/test/vue-fixture.test.ts#L1-L52

I haven't seen these failures before so cannot say that it's flaky - especially now that all 4 CI machines with different OS/Node failed at the same time. 🤔

benmccann:
Thanks! It seems not to be flaky, but also not related to this PR. I opened another PR with an unrelated change and it is also failing in the same manner: https://github.com/vitest-dev/vitest/pull/6298

AriPerkkio:
I think the `main` CI broke somewhere during here: https://github.com/vitest-dev/vitest/commits/main/?since=2024-08-03&until=2024-08-08. Some merges have skipped the CI I guess, cc. @sheremet-va.

PR #6317, opened 2024-08-10
fix(coverage): warn if `vitest` and `@vitest/*` versions don't match

AriPerkkio:
### Description

- Warn if test run loads different versions of `vitest` and `@vitest/coverage-*` packages
- This is just a warning as the packages still may work, but we don't support these cases. Vitest requires users to use 100% exact same versions for `vitest` and `@vitest/*` packages. This has been discussed quite much in the past. 

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/coverage-istanbul/src/provider.ts
+++ b/packages/coverage-istanbul/src/provider.ts
@@ -26,11 +26,12 @@ import type { CoverageMap } from 'istanbul-lib-coverage'
 import libCoverage from 'istanbul-lib-coverage'
 import libSourceMaps from 'istanbul-lib-source-maps'
 import { type Instrumenter, createInstrumenter } from 'istanbul-lib-instrument'
-// @ts-expect-error @istanbuljs/schema has no type definitions
+// @ts-expect-error missing types
 import { defaults as istanbulDefaults } from '@istanbuljs/schema'
-
 // @ts-expect-error missing types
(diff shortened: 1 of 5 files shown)


packages/coverage-istanbul/src/provider.ts
```diff
@@ -26,11 +26,12 @@ import type { CoverageMap } from 'istanbul-lib-coverage'
...
 import libSourceMaps from 'istanbul-lib-source-maps'
 import { type Instrumenter, createInstrumenter } from 'istanbul-lib-instrument'
-// @ts-expect-error @istanbuljs/schema has no type definitions
+// @ts-expect-error missing types
 import { defaults as istanbulDefaults } from '@istanbuljs/schema'
-
 // @ts-expect-error missing types
 import _TestExclude from 'test-exclude'
+
+import { version } from '../package.json' with { type: 'json' }
```
AriPerkkio, line 34: Verified that both providers inline this in `dist/provider.js`:

> ```js
> var version = "2.1.0-beta.4";
> ```

AriPerkkio: Is this done by Rollup automatically or do we configure it? @sheremet-va

sheremet-va: this is done by a json plugin

PR #6349, opened 2024-08-14
chore(deps): bump expect-type to 0.20.0-0

mmkal:

### Tests

### Documentation

### Changesets


mmkal:
@sheremet-va just testing out whether the latest expect-free changes are safe to release as v0.20.0, which will likely become v1 rc. 

I don't understand the failure though, I can't seem to see any logs associated?

sheremet-va:
> @sheremet-va just testing out whether the latest expect-free changes are safe to release as v0.20.0, which will likely become v1 rc.
> 
> I don't understand the failure though, I can't seem to see any logs associated?

It's just flaky. Don't worry.

mmkal:
@sheremet-va oh, interesting, I noticed that there's actually a change to LICENSE.md locally, some automated process picked up one what I inadvertently did which is move it from devDependencies to dependencies.

Is there a reason it's currently bundled rather than a regular dependency? If it's something about how expect-type is set up, I would be open to addressing any issues you might have found, I think since it is a "prod" dependency it'd be better if it's listed that way.

sheremet-va:
> Is there a reason it's currently bundled rather than a regular dependency?

I don't really remember why. Maybe it's an artifact from the time we tried - I think we had to re-implement something. Right now I think it's just re-exported.

> I think since it is a "prod" dependency it'd be better if it's listed that way.

Not sure about this argument - what is a "prod" dependency here and why is it better to list "prod" dependencies? I am fine with moving `expect-type` to `dependencies`, but I don't understand the argument you are trying to make here specifically.

mmkal:
Yeah I should have said regular rather than prod. Just meant that it's actually part of vitest rather than a build/test/lint tool in this case. I'll do some testing to see if I hit any issues- can try arethetypewrong. I'll mark as ready/ ping you when ready.

mmkal:
@sheremet-va I think this is ready. I did a sneaky pkg-pr-new release from my fork, so feel free to try it out: `pnpm add https://pkg.pr.new/mmkal/vitest@c05bcfe`. I think it's good to go.

image

sheremet-va:
Merged in https://github.com/vitest-dev/vitest/pull/6654#issue-comment-box

Thank you for the new version!

mmkal:
The other part of this PR was moving it to a regular dependency. I can open a new one which does only that. It's better that way because it allows for minor changes, security fixes (even if they're unlikely), and advanced users could even fully customise their version if they like.

Happy to see that the version was bumped by renovate but I still think it should be moved to dependencies. I'll create another PR that does *just* the move and we can discuss there if there are any reasons not to.

Issue #6377, opened 2024-08-21
Tasks (tests, suites, hooks) that are too fast show a duration of 0ms in Task results


JessicaSachs:
### Clear and concise description of the problem

I was writing a custom reporter and noticed that many of the tasks I was working with were showing 0ms as their duration. Upon further digging, I noticed that Vitest uses Date.now for a lot of measurements related to timing.

I asked @sheremet-va why that is, and he mentioned that `global.performance` required an import from `node:performance` at the time of writing. Importing from `node:performance` would've broken browser mode for Vitest, so the team elected to use the more compatible (but less precise) `Date.now` to take measurements.

### Suggested solution

Now that `global.performance` is available in all node versions that Vitest supports, we can go through everything to do with timings and swap from using `Date.now` to using `global.performance`

### Alternative

We could decide not to do this.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

LuciNyan:
I'd be happy to take on this issue. Could you assign it to me? 👀

PR #6382, opened 2024-08-23
perf: Use global.performance.now instead of Date.now

LuciNyan:
### Description

Close: #6377

### Tests

### Documentation

### Changesets


LuciNyan:
Hi team! 👋 
I'm seeing some flaky tests here. 🤔 Same code, different results – sometimes it passes, sometimes it fails. Not sure if we've got some unstable tests on our hands or if there's something else going on. Anyone else run into this?

packages/runner/src/run.ts
```diff
@@ -181,7 +181,7 @@ export async function runTest(test: Test | Custom, runner: VitestRunner): Promis
 
   test.result = {
     state: 'run',
-    startTime: start,
+    startTime: Date.now(),
```
sheremet-va, line 184: Why?

packages/runner/src/run.ts
```diff
@@ -376,7 +376,7 @@ export async function runSuite(suite: Suite, runner: VitestRunner): Promise<void
 
   suite.result = {
     state: 'run',
-    startTime: start,
+    startTime: Date.now(),
```
sheremet-va, line 379: Why?

sheremet-va (changes requested):

packages/runner/src/run.ts
LuciNyan: Hi! @sheremet-va 👋

Date.now() gives us milliseconds since 1970, but performance.now() doesn't. 
The name 'startTime' implies it might be used as a specific timestamp, like in `test/cli/test/reported-tasks.test.ts:106`.
```
const diagnostic = passedTest.diagnostic()!
expect(diagnostic).toBeDefined()
expect(diagnostic.heap).toBeGreaterThan(0)
expect(diagnostic.duration).toBeGreaterThan(0)
expect(date(new Date(diagnostic.startTime))).toBe(date(now))
expect(diagnostic.flaky).toBe(false)
expect(diagnostic.repeatCount).toBe(0)
expect(diagnostic.repeatCount).toBe(0)
```
If we switch to performance.now(), we'll probably need to update that test file. So, do I need to do this?

sheremet-va: Yes, I would expect it to use the new `now`

LuciNyan: Thanks for the review! I've made the changes you suggested.

LuciNyan: Hey @sheremet-va 👋,

I've noticed something important about our test execution, particularly for files like `test/reporters/tests/merge-reports.test.ts`. We're using `Tinypool` from `packages/vitest/src/node/pools/forks.ts` to spawn new processes for these tests.

The issue is with our use of `performance.now()`. While it works fine for timing within a single process, it becomes problematic across multiple processes. Each process has its own time reference for `performance.now()`, starting from when that process begins.

This causes issues when we try to use these `startTime` values for sorting in `packages/vitest/src/node/reporters/blob.ts`. The times from different processes aren't comparable.

```
  const files = blobs
    .flatMap(blob => blob.files)
    .sort((f1, f2) => {
      const time1 = f1.result?.startTime || 0
      const time2 = f2.result?.startTime || 0
      return time1 - time2
    })
```

Given this, I still think we should use `Date.now()` for `startTime`. It provides a consistent timestamp across all processes.

What are your thoughts on this? Do you see any potential drawbacks to switching to `Date.now()`?

LuciNyan: Ready for review! 🚀

sheremet-va (approved):

PR #6389, opened 2024-08-24
fix: cjs build of vite node server

AkaraChen:
### Description

Repro link: https://github.com/AkaraChen/vite-node-repro

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vite-node/rollup.config.js
+++ b/packages/vite-node/rollup.config.js
@@ -32,6 +32,9 @@ const external = [
   'node:events',
   'node:vm',
 ]
+const esmOnly = [
+  'tinyrainbow',
+]
 
 const plugins = [
   resolve({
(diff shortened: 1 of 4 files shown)


packages/vite-node/rollup.config.js
```diff
@@ -32,6 +32,9 @@ const external = [
   'node:events',
   'node:vm',
 ]
+const esmOnly = [
```
sheremet-va, line 35: It needs to be moved to `devDependencies` to be bundled.

AkaraChen: fixed

sheremet-va (approved):

packages/vite-node/rollup.config.js
AkaraChen: > It needs to be moved to `devDependencies` to be bundled, not just filtered. Otherwise consumers of CJS package will install a package that's not being used

Just check the bundle artifacts, it is ok, because `rollup-plugin-esbuild` bundle it, and  the esm build still use the original dependency, only cjs build use the bundled version.

But move to `devDependency` maybe a better approach.

PR #6392, opened 2024-08-25
feat(cli): Extend existing list command to output only a list of file names

### LINKED ISSUE #6374: Extend existing `list` command to output only a list of file names

https://github.com/vitest-dev/vitest/issues/2901 lead to the introduction of the [`vitest list`](https://vitest.dev/guide/cli.html#vitest-list) command, but unfortunately this is pretty slow and requires some processing if you're only interested in a list of _files_ that will run with the given filters & flags (i.e. `--changed`), since it parses all test files and returns a list containing every individual test case. What I ideally would want is a command that returns which files are going to be executed given the provided filters & flags without drilling down to individual test cases, requiring only a lookup and eliminating the need for a parse, thus being a lot quicker.

My main use-case for this is dynamic sharding, where I dynamically determine the amount of shards and agents I want to spin up based on a predetermined number `TESTS_PER_SHARD` (i.e. `100`), which serves to keep the test performance stable, irregardless of number of tests. Jest for example can achieve this with [`--listTests [--json]`](https://jestjs.io/docs/cli#--listtests), which runs in under a second for ~1k test files and leads to the following output:

**`jest --listTests`**

```
/path/to/testA.spec.tsx
/path/to/testB.spec.tsx
```
**`jest --listTests --json`**

```
["/path/to/testA.spec.tsx", "/path/to/testB.spec.tsx"]
```

For comparison, running `vitest list --json` on an M1 MBP for the same ~1k files takes 3.5 minutes.

### Suggested solution

Since there is already a dedicated `list` command I suppose it would make sense to provide a new optional flag, i.e. `--filesOnly`.

### Alternative

_No response_

### Additional context

_No response_

### Validations

Ma-hawaj:
- **feat(cli): add the files only option to the list command (unfuncitonal)**
- **feat(cli): add implementation for listing files only**
- **feat(cli): add testing for list --filesOnly**
- **feat(cli): add documentation for list --filesOnly**

### Description

Extend existing list command to output only a list of file names

closes #6374

- [x] It's really useful if your PR references an issue where it is discussed ahead of time. If the feature is substantial or 
introduces breaking changes without a discussion, PR might be closed.

### Tests

### Documentation

### Changesets


sheremet-va (changes requested):
This implementation seems a bit pointless because you can do the same yourself on `vitest list` result.

What we need is to _not collect_ tests at all to make this very fast. This line should just not be called:

https://github.com/vitest-dev/vitest/blob/95f0203f27f5659f5758638edc4d1d90283801ac/packages/vitest/src/node/core.ts#L369

Ma-hawaj:
OK, 
so will adding the following
```
async collectFiles(filters?: string[]) {
    const files = await this.filterTestsBySource(
      await this.globTestFiles(filters),
    )
}
```
in the vitest class and checking the option in cac.ts:collect() and calling it will do the job.

packages/vitest/src/node/cli/cli-api.ts
```diff
@@ -184,6 +189,51 @@ export function processCollected(ctx: Vitest, files: File[], options: CliOptions
...
+  })
+}
+
+function formatFilesAsString(files: WorkspaceSpec[]) {
+  return files.map((file) => {
+    let name = file.moduleId.substring(String(file[0].runner.root).length + 1)
+    if (file.project.name) {
+      name = `[${file.project.name}] ${name}`
+    }
+    console.log(name)
```
sheremet-va, line 232: Why does the `format` function has `console.log`?

sheremet-va (changes requested):

packages/vitest/src/node/cli/cli-api.ts
Ma-hawaj: It also prints the output to the user at the same time

packages/vitest/src/node/cli/cli-api.ts
```diff
@@ -184,6 +189,50 @@ export function processCollected(ctx: Vitest, files: File[], options: CliOptions
   return formatCollectedAsString(files).forEach(test => console.log(test))
 }
 
+export function outputFileList(files: WorkspaceSpec[], options: CliOptions) {
+  // const paths = files.map((file) => file.moduleId)
```
sheremet-va, line 193:
```suggestion
```

packages/vitest/src/node/cli/cli-api.ts
```diff
@@ -184,6 +189,48 @@ export function processCollected(ctx: Vitest, files: File[], options: CliOptions
...
+    if (file.project.name) {
+      result.projectName = file.project.name
+    }
+    return result
+  })
+}
+
+function formatFilesAsString(files: WorkspaceSpec[]) {
+  return files.map((file) => {
+    let name = file.moduleId.substring(String(file[0].runner.root).length + 1)
```
sheremet-va, line 226:
```suggestion
    let name = resolve(options.root || process.cwd(), file.moduleId)
```

sheremet-va (changes requested):

packages/vitest/src/node/cli/cli-api.ts
```diff
@@ -184,6 +189,48 @@ export function processCollected(ctx: Vitest, files: File[], options: CliOptions
...
+    }
+
+    if (file.project.name) {
+      result.projectName = file.project.name
+    }
+    return result
+  })
+}
+
+function formatFilesAsString(files: WorkspaceSpec[]) {
```
Ma-hawaj, line 224:
```suggestion
function formatFilesAsString(files: WorkspaceSpec[], options: CliOptions) {
```

packages/vitest/src/node/cli/cli-api.ts
```diff
@@ -184,6 +189,48 @@ export function processCollected(ctx: Vitest, files: File[], options: CliOptions
   return formatCollectedAsString(files).forEach(test => console.log(test))
 }
 
+export function outputFileList(files: WorkspaceSpec[], options: CliOptions) {
+  if (typeof options.json !== 'undefined') {
+    return outputJsonFileList(files, options)
+  }
+
+  return formatFilesAsString(files).map(file => console.log(file))
```
Ma-hawaj, line 197:
```suggestion
  return formatFilesAsString(files, options).map(file => console.log(file))
```

packages/vitest/src/node/cli/cli-api.ts
Ma-hawaj: > 

Seems this didn't work, it prints the whole path now.

packages/vitest/src/node/cli/cli-api.ts
```diff
@@ -184,6 +189,48 @@ export function processCollected(ctx: Vitest, files: File[], options: CliOptions
...
+
+  return formatFilesAsString(files, options).map(file => console.log(file))
+}
+
+function outputJsonFileList(files: WorkspaceSpec[], options: CliOptions) {
+  if (typeof options.json === 'boolean') {
+    return console.log(JSON.stringify(formatFilesAsJSON(files), null, 2))
+  }
+  if (typeof options.json === 'string') {
+    const jsonPath = resolve(options.root || process.cwd(), options.json)
```
sheremet-va, line 205:
```suggestion
    const jsonPath = relative(options.root || process.cwd(), options.json)
```

should be imported from `pathe`, can't add the comment there unfortunetly

Ma-hawaj: did it in 99afa71

sheremet-va:
Thanks!

Issue #6033, opened 2024-07-03
[Feature]: Allow Callbacks Passed to `before*`/`after*` to Return Anything


ITenthusiasm:
### Clear and concise description of the problem

Right now, I have an `async` function that I use to create and start a server:

```ts
async function startServer(port) {
  // ...
}
```

Within a test file, I have a setup that looks something like this

```ts
describe("Description", () => {
  let server: Awaited<ReturnType<typeof startServer>>;

  beforeAll(async () => {
    server = await startServer(port);
  });

  afterAll(() => {
    server.close();
  });

  // ...
});
```

For brevity, I'd like to write my code like this instead:
```ts
describe("Description", () => {
  let server: Awaited<ReturnType<typeof startServer>>;

  beforeAll(async () => (server = await startServer(port)));
  afterAll(() => server.close());
  // ...
});
```

But I can't, because the current types for `before*`/`after*` forbid returning an actual value.

### Suggested solution

It would be great if these return types (for `before*` and `after*`) could be updated so that developers could shorthand their setup/teardown blocks. For simple setup/teardown steps, this would reduce the lines of code by 2 per function call (assuming a formatter like `Prettier` is being used).

### Alternative

Currently, developers can circumvent the problem by using [`void`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/void). But the appearance of this operator in a codebase is usually considered a smell.

### Additional context

I _think_ that this change would make Vitest "more compatible" with Jest users (from a TypeScript perspective). However, I noticed from the [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/void) that `beforeAll` accepts a cleanup function. So if this feature request was accepted, there'd probably need to be a proper `typeof` check put in place. (There might already be one in place.)

---

Search Words: `before after promise`
Related Issues: #2149

I can understand the reasoning for restricting `before*`, but I think it would still be nice if `after*` could be looser in its type (assuming that its return value is not operated on outside of a `Promise` check). Note: The `after*` consideration is why I'm keeping this issue separate instead of commenting on the old one.

<details>
  <summary>However, if I may plead my case:</summary>

> I think it's better to be strict in this case, because hooks implementation differs from jest, and if we don't throw for non-functions, users might wrongly think it is the same, and accidentally return a teardown function.

[...]

PR #6393, opened 2024-08-25
fix(types): Allow Callbacks Passed to before*/after* to Return Anything

LuciNyan:
### Description

Hey team! 👋

- Current implementation of callCleanupHooks already accepts non-function values as cleanupCallback
- Type definition currently restricts cleanupCallback to functions only
- This PR updates the type definition to allow non-function values
- No changes to the actual implementation, only type definitions are modified

```
async function callCleanupHooks(cleanups: HookCleanupCallback[]) {
  await Promise.all(
    cleanups.map(async (fn) => {
      if (typeof fn !== 'function') {
        return
      }
      await fn()
    }),
  )
}
```

Close: #6033

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/runner/src/types/tasks.ts
+++ b/packages/runner/src/types/tasks.ts
@@ -498,28 +498,31 @@ export type HookListener<T extends any[], Return = void> = (
   ...args: T
 ) => Awaitable<Return>
 
-export type HookCleanupCallback = (() => Awaitable<unknown>) | void
+/**
+ * @deprecated
+ */
+export type HookCleanupCallback = unknown
 
 export interface BeforeAllListener {
-  (suite: Readonly<Suite | File>): Awaitable<HookCleanupCallback>
+  (suite: Readonly<Suite | File>): Awaitable<unknown>
 }
 
 export interface AfterAllListener {
-  (suite: Readonly<Suite | File>): Awaitable<void>
+  (suite: Readonly<Suite | File>): Awaitable<unknown>
 }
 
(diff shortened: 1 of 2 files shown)


packages/runner/src/types/tasks.ts
```diff
@@ -498,14 +498,14 @@ export type HookListener<T extends any[], Return = void> = (
   ...args: T
 ) => Awaitable<Return>
 
-export type HookCleanupCallback = (() => Awaitable<unknown>) | void
+export type HookCleanupCallback = (() => Awaitable<unknown>) | any
```
sheremet-va, line 501: I would prefer if we used `unknown` instead.

`(() => Awaitable<unknown>) | any` or `(() => Awaitable<unknown>) | unknown` also don't make sense because both `unknown` and `any` have higher priority, so a single `unknown` should be fine

sheremet-va (changes requested):

packages/runner/src/types/tasks.ts
LuciNyan: Hi @sheremet-va 👋
Do we still need to keep the 'export type HookCleanupCallback = unknown' type, or can we just replace all `HookCleanupCallback` with 'unknown'?

sheremet-va: We should just deprecate it because removing the export is a breaking change

LuciNyan: Ready for review! 🚀

PR #332, opened 2021-12-27
feat(tests): testNamePattern

Aslemammad:

PR #6411, opened 2024-08-27
feat(cli): Support specifying a line number when filtering tests

### LINKED ISSUE #5445: Support specifying a line number when filtering tests

Right now it's quite cumbersome to filter out tests when using a Vitest CLI. You need to know the name and the file:

```
vitest test/basic.test.ts -t "some-name"
```

It fails if there are multiple tests that have the same pattern, so you are forced to use the full name manually:

```
vitest test/basic.test.ts -t "test full some-name"
```

### Suggested solution

Vitest 1.4.0 introduced [`includeTaskLocation`](https://vitest.dev/config/#includeTaskLocation) option that adds a `location` property to every test. With this information, we can filter tests by the line number:

```
vitest test/basic.test.ts:20
```

### Alternative

_No response_

### Additional context

I am not sure what should happen when the line doesn't exist or there is no test at that line. Currently, we only know the location of a test, and we don't know when the body of a test ends, so it's tricky to implement a range support.

I think we should:

- Only run a test if the line is exactly equal to a location line
- If there is no test at that location, throw an error
- If range is used, throw an error (`test/basic.test.ts:20-30`)

### Validations

mzhubail:
### Description

closes #5445.

1. Added a function to parse filters into `{ filename, lineNumber? }`.  The function attempts to parse each filter with the format `<filename>:<line_number>`.  If succssful, `lineNumber` is set, and we have a "location filter", otherwise line number is not set and the filter is treated as a regular filter.

1. Added `locationFilters` to `UserConfig` and to `VitestRunnerConfig`.

    Note that the we only store filters that have `lineNumber`, in order to use them in `interpretTaskModes`.

1. The parsing is done in `cac.ts:start` because that's the place where we have access to both the options and filters.

I think I know where to do stuff for the most part, but I need help with some stuff:

- I'm not sure how to report errors in case the user included a range instead of a line number.
- I'm not happy with my use of `Required` in `Required<Filter[]>`, but I'm not sure what would be a better way to do it.

- I think we should match filename strictly when location filter is provided, but I'm not sure how to.  Should I be using something similar to `WorkspaceProject.filterFiles()`?

    ```ts
    if (isAbsolute(f) && t.startsWith(f)) {
        return true
    }
    ```

- ~~As for printing an error when the user provides a wrong line number, it's a bit tough with `interpretTaskModes` being a recursive a function.  What I have in mind is to keep track of all the filters and wether it was matched or not, and print out errors once we've been through all the tests.~~

    ~~The part I'm having an issue with is _where_ to keep track of that.  That would involve major changes to how `interpretTaskModes` is written.  I could move the~~

    See 36c88e9574eafcdc6be2aa582056b118d4ebb33c

### Tests

### Documentation

### Changesets


sheremet-va:
Vitest has specs that describe the test behaviour:

https://github.com/vitest-dev/vitest/blob/83638eb9ad58b9addfe74121f7ef2952518aeb96/packages/vitest/src/node/core.ts#L1083

I would prefer if we passed down locations there to keep the scope clean and avoid polluting the config.

sheremet-va:
> As far as I can tell, `globTestSpecs` is run before we have access to tests (not collected yet), and I need access to the tests to filter based on line number.

We don't need to collect tests for this to work, spec defines what tests to run. This is the perfect place to tell Vitest to limit what test functions to run.

> I went with this approach as it is more similar to how `-t, --testNamePattern` is handled (#332)

I know. This doesn't really explain why it should follow the same implementation.

mzhubail:
I might be missing something.  What kind of information do I have access to in a spec?  How would I access the line numbers of a test?

I see `location` in [`TaskBase`](https://github.com/vitest-dev/vitest/blob/36c88e9574eafcdc6be2aa582056b118d4ebb33c/packages/runner/src/types/tasks.ts#L64-L72).  Do I have access to it in `globTestSpecs`?

sheremet-va:
> I might be missing something. What kind of information do I have access to in a spec? How would I access the line numbers of a test?
> 
> I see `location` in [`TaskBase`](https://github.com/vitest-dev/vitest/blob/36c88e9574eafcdc6be2aa582056b118d4ebb33c/packages/runner/src/types/tasks.ts#L64-L72). Do I have access to it in `globTestSpecs`?

Spec is a specification, it doesn't have any information about the test or a file. It _tells_ what file to run and how to run it. `TestCase` is the result. You need to pass down this information somewhere in the pool:

https://github.com/vitest-dev/vitest/blob/83638eb9ad58b9addfe74121f7ef2952518aeb96/packages/vitest/src/node/pools/threads.ts#L105

Maybe you can populate the provided context somehow or inject the config value, that's up to you, but the processing should happen there.

mzhubail:
>  You can keep the filtering in the test like it is, but change how the values are passed down from the main process.

Do you mean that the issue at hand is not filtering inside `interpretTaskModes`, but rather the use of `VitestRunnerConfig` and `UesrConfig` to pass the values to `interpretTaskModes`?

> Maybe you can populate the provided context somehow or inject the config value, that's up to you, but the processing should happen there.

Is the issue that we should provide each worker with only the values that are relevant values to its files (unlike `testNamePatter` which is global for all files)?

sheremet-va:
> Do you mean that the issue at hand is not filtering inside `interpretTaskModes`, but rather the use of `VitestRunnerConfig` and `UesrConfig` to pass the values to `interpretTaskModes`?

Yes. The `interpretTaskModes` code is fine.

> Is the issue that we should provide each worker with only the values that are relevant values to its files (unlike testNamePatter which is global for all files)?

Yes, that's the idea.

mzhubail:
@sheremet-va 

I gave it a try.  See 76bd9212b4f3abecce3e4cbc5178ec00d2069348.  Previous code will be removed of course.

In `globTestSpecs`, the filters will be matched against the files, and if there is a location in the filter, it will be included with its file, in a `FileSpec: { path, testLocations? }`.

I thought it would be a good idea to include the files with the location in the same object, instead of having them in their own attribute in the config and then matching based on the path or something.  But I ended up having to modify types everywhere to get this to work.

I had to modify `groupFilesByEnv` to include the `testLocations` in its output, and also modify pools to pass the test locations to `runFiles`.

Much of the code in this commit is hacked together, but I had to get something that somewhat works, sometimes, just to see if I'm on the right track.

Would appreciate inputs on this before I commit to making more changes.

packages/vitest/src/node/core.ts
```diff
@@ -374,7 +377,7 @@ export class Vitest {
     }
   }
 
-  async start(filters?: string[]) {
+  async start(filters?: Filter[]) {
```
sheremet-va, line 380: We cannot change the type of this because it would be a big breaking change

packages/vitest/src/node/core.ts
```diff
@@ -1080,18 +1083,22 @@ export class Vitest {
     return this.globTestSpecs().then(specs => specs.map(spec => spec.moduleId))
   }
 
-  public async globTestSpecs(filters: string[] = []) {
+  public async globTestSpecs(filters: Filter[] = []) {
```
sheremet-va, line 1086: We cannot change the type of this because it would be a big breaking change

packages/vitest/src/node/logger.ts
```diff
@@ -141,7 +141,8 @@ export class Logger {
     return code
   }
 
-  printNoTestFound(filters?: string[]) {
+  printNoTestFound(filters_?: Filter[]) {
```
sheremet-va, line 144: We cannot change the type of this because it would be a big breaking change

packages/vitest/src/node/workspace.ts
```diff
@@ -153,8 +153,12 @@ export class WorkspaceProject {
     }
   }
 
-  public createSpec(moduleId: string, pool: string): DeprecatedWorkspaceSpec {
-    return new TestSpecification(this, moduleId, pool) as DeprecatedWorkspaceSpec
+  public createSpec(
+    moduleId: string,
+    pool: string,
+    testLocations: number[] | undefined,
```
sheremet-va, line 159:
```suggestion
    testLocations?: number[] | undefined,
```

packages/runner/src/collect.ts
```diff
@@ -20,14 +20,15 @@ import { runSetupFiles } from './setup'
 const now = Date.now
 
 export async function collectTests(
-  paths: string[],
+  specs: FileSpec[],
```
sheremet-va, line 23: We cannot change the type of this because it would be a big breaking change

sheremet-va (changes requested):
The general direction seems fine to me, but there are some changes that we cannot make because it would be a breaking change. Unless you want to have this merged in a year,

mzhubail:
> The general direction seems fine to me, but there are some changes that we cannot make because it would be a breaking change. 

One thing I shouldn't be doing is parsing the filters in `cac.ts:start`, this is forcing me to change types more than I have to.

https://github.com/vitest-dev/vitest/blob/76bd9212b4f3abecce3e4cbc5178ec00d2069348/packages/vitest/src/node/cli/cac.ts#L265

If I were to do the parsing inside `globTestSpecs` it will solve the type change related to `Filter`.


How about Instead of using an incompatible type like `FileSpec`, we use a superset type, maybe something like `string[] | FileSpec[]` or `(FileSpec | string)[]`, would that do for now?  Do you have any other suggestions?

> Unless you want to have this merged in a year,

This has been labelled a "nice to have", but I'm not sure I'll be here in a year :)

sheremet-va:
> How about Instead of using an incompatible type like `FileSpec`, we use a less restrictive superset type, maybe something like `string[] | FileSpec[]` or `(FileSpec | string)[]`, would that do for now? Do you have any other suggestions?

I guess that would work but I don't understand why you need this if you will still parse it inside, right?

mzhubail:
> I guess that would work but I don't understand why you need this if you will still parse it inside, right?

I Forgot to add the declaration of `FileSpec`, my bad.

```ts
export interface FileSpec {
  file: string
  testLocations: number[] | undefined
}
```

The intention of this is to pass both the filepath and the `testLocations` togther, instead of just passing the filepath as `string`.

mzhubail:
Cleaned up a bit

- I'm not sure if I should update these types to `string [] | FileSpec[]` too.

    https://github.com/vitest-dev/vitest/blob/8f49e9e544086544f4efa4bccc1256148d560606/packages/runner/src/run.ts#L516

    https://github.com/vitest-dev/vitest/blob/8f49e9e544086544f4efa4bccc1256148d560606/packages/runner/src/run.ts#L498

    https://github.com/vitest-dev/vitest/blob/8f49e9e544086544f4efa4bccc1256148d560606/packages/vitest/src/runtime/runBaseTests.ts#L21

    https://github.com/vitest-dev/vitest/blob/8f49e9e544086544f4efa4bccc1256148d560606/packages/vitest/src/runtime/runVmTests.ts#L25-L26

- In `interpretTaskModes`, I've renamed the argument to `file`.  It seems to me that only a file gets passed.  Please confirm.

    https://github.com/vitest-dev/vitest/blob/8f49e9e544086544f4efa4bccc1256148d560606/packages/runner/src/utils/collect.ts#L9

    and I'm not sure how to handle error reporting in case a wrong line number was provided.  Pointers would be appreciated.

- I've tried to minimize the changes in `packages/vitest/src/node/pools/forks.ts`.  I will replicate them for all pool types.

    https://github.com/vitest-dev/vitest/blob/8f49e9e544086544f4efa4bccc1256148d560606/packages/vitest/src/node/pools/forks.ts#L280

    Moved away from using a variable, because I wasn't sure what to name.


----

I hope I'm in the right direction.  Thank you for you time @sheremet-va!

packages/runner/src/utils/collect.ts
```diff
@@ -6,53 +6,84 @@ import type { File, Suite, TaskBase } from '../types/tasks'
...
+  traverseSuite(file)
+
+  const nonMatching = testLocations?.filter(loc => !matchedLocations.includes(loc))
+  if (nonMatching && nonMatching.length !== 0) {
+    const message = nonMatching.length === 1
+      ? `line ${nonMatching[0]}`
+      : `lines ${nonMatching.join(', ')}`
+
+    // TODO: figure out how to report errors
+    throw new Error(`No test found in ${file.name} in ${message}`)
```
sheremet-va, line 86: This function shouldn't throw errors. We can add a serialized error to `file.result.errors`

packages/vitest/src/node/cli/cli-api.ts
```diff
@@ -237,6 +237,62 @@ export function formatCollectedAsString(files: File[]) {
...
+      const [filename, filters] = entry
+      const testLocations = filters.map(f => f.lineNumber)
+
+      // `testLocations` should be all of same type, mixing means the file was
+      // specified with and without test locations
+      if (
+        testLocations.length !== 0
+        && testLocations.some(l => typeof l !== typeof testLocations[0])
+      ) {
+        console.error(`ERR: ${filename} was provided with and without test location`)
```
sheremet-va, line 283: The warning is confusing. Logs should also use `vitest.logger.*` methods

packages/runner/src/run.ts
```diff
@@ -512,10 +513,12 @@ export async function startTests(paths: string[], runner: VitestRunner): Promise
   return files
 }
 
-async function publicCollect(paths: string[], runner: VitestRunner): Promise<File[]> {
+async function publicCollect(specs: FileSpec[], runner: VitestRunner): Promise<File[]> {
```
sheremet-va, line 516: This is still a public API, we need to allow `string[]` even if Vitest always passed down `FileSpec`

packages/vitest/src/node/core.ts
```diff
@@ -1081,21 +1088,45 @@ export class Vitest {
...
+        testLocMatches[file] = true
+        const spec = project.createSpec(file, 'typescript', loc)
         this.ensureSpecCached(spec)
         files.push(spec)
       })
     }))
+
+    Object.entries(testLocations).forEach(([filepath, loc]) => {
+      if (loc.length !== 0 && !testLocMatches[filepath]) {
+        console.error(`Couldn\'t find file "${filepath}".\n`
```
sheremet-va, line 1125: All logs should be printed via `ctx.logger.*`

packages/vitest/src/node/core.ts
```diff
@@ -390,6 +391,12 @@ export class Vitest {
       await this.globTestFiles(filters),
     )
 
+    if (!this.config.includeTaskLocation && files.some(spec => spec.testLocations)) {
```
sheremet-va, line 394: This should be an actual `throw new Error` instead of a warning

packages/vitest/src/types/worker.ts
```diff
@@ -26,7 +27,7 @@ export interface ContextRPC {
   workerId: number
   config: SerializedConfig
   projectName: string
-  files: string[]
+  files: string[] | FileSpec[]
   environment: ContextTestEnvironment
```
sheremet-va, line 31: I wonder, wouldn't it be better if we used `SerializedTestSpecification` for this and added an additional `specs` field instead for better backwards compatibility?

packages/vitest/src/node/cli/cli-api.ts
mzhubail: I agree.  Seems pointless now that I think about it.

packages/vitest/src/types/worker.ts
mzhubail: Sounds like it, but I'm not sure.  I looked it up in the codebase, it seems to be a bit away from the files I've dealt with.

Seems better to me to use a type we already have (which also seems to be doing what we need).  But I'll have to do some mess around a bit so that I'll be able to access it in `interpretTaskModes`.

The call is yours.

sheremet-va: Ok, let's just keep it like you already did

mzhubail:
squashed and rebased.  Should be up to date with main, will check it later.

For now, I've added an error `IncludeTaskLocationDisabledError`, it is ok?  I'm thinking of adding a new one for range line number (location filters don't accept ranges).

mzhubail:
@sheremet-va Should be good, if you can take a look.

Thanks you!

sheremet-va:
> @sheremet-va Should be good, if you can take a look.
> 
> Thanks you!

There are issues in CI it seems. Something with benchmarking?

packages/vitest/src/node/cli/cac.ts
```diff
@@ -1,11 +1,11 @@
 import type { VitestRunMode } from '../types/config'
-import type { CliOptions } from './cli-api'
 import type { CLIOption, CLIOptions as CLIOptionsConfig } from './cli-config'
 import { toArray } from '@vitest/utils'
 import cac, { type CAC, type Command } from 'cac'
 import { normalize } from 'pathe'
 import c from 'tinyrainbow'
 import { version } from '../../../package.json' with { type: 'json' }
+import { type CliOptions, collectAndProcess } from './cli-api'
```
sheremet-va, line 8: The `'./cli-api'` should be imported dynamically

packages/runner/src/utils/collect.ts
```diff
@@ -6,53 +6,95 @@ import { relative } from 'pathe'
...
-      else if (t.mode === 'only') {
-        checkAllowOnly(t, allowOnly)
-        t.mode = 'run'
+      else if (t.type === 'suite') {
+        if (t.mode === 'skip') {
+          skipAllTasks(t)
+        }
+        else {
+          traverseSuite(t)
+          // interpretTaskModes(t, namePattern, locationFilters, onlyMode, includeTask, allowOnly)
```
sheremet-va, line 64: comment

mzhubail:
> There are issues in CI it seems. Something with benchmarking?

Aha, thought it had to do with stuff not related to me, will look into it.

packages/vitest/src/node/cli/cli-api.ts
```diff
@@ -163,6 +173,45 @@ export async function prepareVitest(
...
+      processCollected(ctx, tests, options)
+    }
+    else {
+      const files = await ctx.listFiles(cliFilters.map(normalize))
+      outputFileList(files, options)
+    }
+
+    await ctx.close()
+  }
+  catch (e) {
```
sheremet-va, line 202: This is interesting part. If we add the additional handling, we need to add it for every error like on line 102. I think how it works right now in `main` is a bug, but it is out of scope of this PR because we also need to add more tests for this. Let's keep it like it was before and track this in another issue

packages/vitest/src/node/errors.ts
```diff
@@ -13,3 +13,19 @@ export class GitNotFoundError extends Error {
...
+  constructor() {
+    super('Recieved line number filters while `includeTaskLocation` option is disabled')
+  }
+}
+
+export class RangeLocationFilterProvidedError extends Error {
+  code = 'VITEST_RANGE_LOCATION_FILTER_PROVIDED'
+
+  constructor(filter: string) {
+    super(`Location filters are not allowed: ${filter}`)
```
sheremet-va, line 29: This error is thrown when `-` is used in the location, but it doesn't explain why it's wrong or how to fix it. Maybe add more explanation?

packages/vitest/src/node/cli/cli-api.ts
```diff
@@ -282,6 +331,53 @@ export function formatCollectedAsString(files: File[]) {
   }).flat()
 }
 
+export function parseFilter(filter: string): Filter {
```
sheremet-va, line 334: It doesn't seem like these two functions belong here because they are not used for the CLI specifically. Maybe create a separate file for them?

packages/vitest/src/node/core.ts
```diff
@@ -1086,21 +1088,58 @@ export class Vitest {
   }
 
   public async globTestSpecs(filters: string[] = []) {
+    const dir = this.config.dir || this.config.root
+    const parsedFilters = filters.map(f => parseFilter(f))
+    const testLocations = groupFilters(parsedFilters.map(
+      f => ({ ...f, filename: resolve(dir, f.filename) }),
```
sheremet-va, line 1094: I wonder if it makes sense to validate that the path exists. Filters can be anything, you can provide a part of the file

I think we should throw an error if:
- There is a location in filters
- File is resolved with `process.cwd(), file`, not `root` (because filters are provides via a CLI normally)
- File doesn't exist, or it's not a file at all (a directory for example)

packages/vitest/src/node/spec.ts
```diff
@@ -20,20 +20,22 @@ export class TestSpecification {
   public readonly project: TestProject
   public readonly moduleId: string
   public readonly pool: Pool
+  public readonly testLocations: number[] | undefined
```
sheremet-va, line 23: let's mark it with `/** @private */` for now, I have plans for it in the future

packages/vitest/src/runtime/runVmTests.ts
```diff
@@ -21,7 +22,8 @@ import { getWorkerState } from './utils'
 
 export async function run(
   method: 'run' | 'collect',
-  files: string[],
+  // TODO consider type
```
sheremet-va, line 25: todo

packages/vitest/src/typecheck/collect.ts
```diff
@@ -203,6 +203,8 @@ export async function collectTests(
   interpretTaskModes(
     file,
     ctx.config.testNamePattern,
+    // TODO: make sure you're not screwing stuff up
```
sheremet-va, line 206: todo

sheremet-va (changes requested):
Looks good! We need more tests, all the tests now are using `list` command

packages/vitest/src/node/core.ts
mzhubail: Interesting.

What came to mind is to limit matching to full filename, instead of part of it, in order not to match multiple files at the same time.  Does that sound right to you?

Note that I'm passing the filters as is (before resolving) to `project.globTestFiles`, and then I narrow the matching for filters with location specified.  Later at line 1143, I'm skipping over reporting errors for filters that don't have location specified.

---

For now I'm just reporting error and continuing.  Now that I'm thinking about it, this has weird behavior for example `pnpm exec vitest basic.test.ts:3` would report an error about "not finding the file", but will run the tests anyways.

It seems I might be able to fix this by only resolving filters with test location and then passing on to `project.globTestFiles`, does that sound right to you?

Good point on resolving with `process.cwd()`.

sheremet-va: I don't know if I follow what you want to do instead of what I proposed

packages/vitest/src/runtime/runVmTests.ts
mzhubail: Just wasn't sure if this was a public interface or not

packages/vitest/src/node/cli/cli-api.ts
mzhubail: What do you think of `vitest/src/node/filter.ts`?

packages/vitest/src/node/pools/forks.ts
```diff
@@ -1,3 +1,4 @@
+import type { FileSpec } from '@vitest/runner/types/runner'
```
sheremet-va, line 1: `@vitest/runner/types/runner` doesn't exist when built. you need to expose it in `@vitest/runner`

mzhubail:
@sheremet-va should be mostly good.

Build is good.  Test is good, but not for windows (or sometimes macos), idk why

Changed code to throw an error when a given filter doesn't match -- covered this in test cases by having `math.test.ts` and `math-with-dashes-in-name.test.ts`; specifying `math:99` should result in an error.

sheremet-va:
> Test is good, but not for windows (or sometimes macos), idk why

It's probably related to `/` and `\` difference. MacOS is just flaky, but your Windows tests are clearly failing because there is a bug somewhere

sheremet-va (dismissed):

sheremet-va:
Looks good to me! Do you think you can fix the Windows errors?

docs/guide/filtering.md
```diff
@@ -24,6 +24,22 @@ basic/foo.test.ts
 
 You can also use the `-t, --testNamePattern <pattern>` option to filter tests by full name. This can be helpful when you want to filter by the name defined within a file rather than the filename itself.
 
+You can also specify the test by filename and line number:
```
sheremet-va, line 27:
```suggestion
Since Vitest 2.2, you can also specify the test by filename and line number:
```

docs/guide/filtering.md
```diff
@@ -24,6 +24,22 @@ basic/foo.test.ts
...
+```
+
+::: warning
+Note that you have to specify the full filename, and specify the exact line number, i.e. you can't do
+
+```bash
+$ vitest foo:10
+$ vitest basic/foo.test.ts:10-25
+```
+
```
sheremet-va, line 40:
```suggestion
```

sheremet-va (approved):

sheremet-va:
Thank you for keeping this PR up-to-date!

Issue #5090, opened 2024-02-01
debugger still not working while test in browser mode


gezilinll:
### Describe the bug

I commit issue yesterday: https://github.com/vitest-dev/vitest/issues/5085 , but @sheremet-va closed it and mark it to duplicate to https://github.com/vitest-dev/vitest/issues/4497 .
I have tried ways of https://github.com/vitest-dev/vitest/issues/4497 metioned, but still not working, and I don't think they are same problem, please try my demo in Reproduction, it still can not debugger in browser mode, THANKS~

### Reproduction

https://github.com/gezilinll/vitest-browser-debugger

### System Info

```shell
System:
    OS: macOS 13.5.2
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 288.28 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.15.0 - ~/Library/Caches/fnm_multishells/56721_1706690459670/bin/node
    Yarn: 1.22.19 - ~/Library/Caches/fnm_multishells/56721_1706690459670/bin/yarn
    npm: 9.5.0 - ~/Library/Caches/fnm_multishells/56721_1706690459670/bin/npm
    pnpm: 7.30.0 - ~/Library/Caches/fnm_multishells/56721_1706690459670/bin/pnpm
  Browsers:
    Chrome: 121.0.6167.139
    Safari: 16.6
```

### Used Package Manager

pnpm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

AriPerkkio:
It's possible if you run Vitest in watch-mode and open browser's dev-tools.

[...]

PR #6433, opened 2024-09-01
feat(browser): support `--inspect`

AriPerkkio:
### Description

- Closes https://github.com/vitest-dev/vitest/issues/5090

Adds support for opening debugger session on Browser mode via `vitest --inspect --browser`. This is only supported in Chromium browsers with Playwright. 

[vitest-inspect.webm](https://github.com/user-attachments/assets/ad5a1afc-756e-4c0b-a2c2-db7c350f3ce4)

Support for `--inspect-brk`, where debugger stops at the first test file is possible but not yet implemented. There's some weird going on with source maps in browser mode. Let's leave that feature to separate PR.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/browser/src/node/providers/playwright.ts
+++ b/packages/browser/src/node/providers/playwright.ts
@@ -71,6 +71,18 @@ export class PlaywrightBrowserProvider implements BrowserProvider {
         headless: options.headless,
       } satisfies LaunchOptions
 
+      if (this.ctx.config.inspector.enabled) {
+        // NodeJS equivalent defaults: https://nodejs.org/en/learn/getting-started/debugging#enable-inspector
+        const port = this.ctx.config.inspector.port || 9229
+        const host = this.ctx.config.inspector.host || '127.0.0.1'
+
+        launchOptions.args ||= []
+        launchOptions.args.push(`--remote-debugging-port=${port}`)
+        launchOptions.args.push(`--remote-debugging-address=${host}`)
+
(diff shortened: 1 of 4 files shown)


packages/browser/src/node/providers/playwright.ts
```diff
@@ -71,6 +71,18 @@ export class PlaywrightBrowserProvider implements BrowserProvider {
         headless: options.headless,
       } satisfies LaunchOptions
 
+      if (this.ctx.config.inspector.enabled) {
+        // NodeJS equivalent defaults: https://nodejs.org/en/learn/getting-started/debugging#enable-inspector
+        const port = this.ctx.config.inspector.port || 9229
+        const host = this.ctx.config.inspector.host || '127.0.0.1'
+
+        launchOptions.args ||= []
+        launchOptions.args.push(`--remote-debugging-port=${port}`)
```
sheremet-va, line 80: Does this work for all browsers? If not, do browsers just accept invalid args?

AriPerkkio: This works only on Chromium. Webkit seems to hang when this is passed. I'll add checks to `node/config/resolveConfig.ts` for browser+inspect and throw error when unsupported combination is used. 

Firefox has different flag, `--start-debugger-server=${port}`. I wasn't able to get my VSCode stop at correct lines with that one though so I'll leave it out of scope now.

sheremet-va (approved):

PR #6434, opened 2024-09-01
feat(browser): support `--inspect-brk`

AriPerkkio:
### Description

- Continues https://github.com/vitest-dev/vitest/pull/6433

Adds support for `--browser --inspect-brk`. Compared to `--inspect`, this one waits for debugger to attach and then stops on top of the first test file. Similar experience as `vitest --inspect-brk` does on Node side. Requires CDP so only Playwright and Chromium browser are supported.

Demo below has no initial breakpoints set:

[vitest-inspect-brk.webm](https://github.com/user-attachments/assets/d6aa1045-af53-4c11-a7e7-8fcca7fa9c69)

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/browser/src/node/pool.ts
+++ b/packages/browser/src/node/pool.ts
@@ -37,6 +37,23 @@ export function createBrowserPool(ctx: Vitest): ProcessPool {
       )
     }
 
+    async function setBreakpoint(contextId: string, file: string) {
+      if (!project.config.inspector.waitForDebugger) {
+        return
+      }
+
+      if (!provider.getCDPSession) {
+        throw new Error('Unable to set breakpoint, CDP not supported')
+      }
+
+      const session = await provider.getCDPSession(contextId)
+      await session.send('Debugger.enable', {})
+      await session.send('Debugger.setBreakpointByUrl', {
+        lineNumber: 0,
+        urlRegex: escapePathToRegexp(file),
+      })
+    }
+
     const filesPerThread = Math.ceil(files.length / threadsCount)
 
     // TODO: make it smarter,
@@ -83,7 +100,7 @@ export function createBrowserPool(ctx: Vitest): ProcessPool {
         const url = new URL('/', origin)
         url.searchParams.set('contextId', contextId)
         const page = provider
-          .openPage(contextId, url.toString())
+          .openPage(contextId, url.toString(), () => setBreakpoint(contextId, files[0]))
           .then(() => waitPromise)
         promises.push(page)
       }
@@ -144,4 +161,8 @@ export function createBrowserPool(ctx: Vitest): ProcessPool {
     runTests: files => runWorkspaceTests('run', files),
     collectTests: files => runWorkspaceTests('collect', files),
   }
+}
+
+function escapePathToRegexp(path: string): string {
+  return path.replace(/[/\\.?*()^${}|[\]+]/g, '\\$&')
 }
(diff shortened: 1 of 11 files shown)


packages/browser/src/node/pool.ts
```diff
@@ -37,6 +37,23 @@ export function createBrowserPool(ctx: Vitest): ProcessPool {
...
+      if (!provider.getCDPSession) {
+        throw new Error('Unable to set breakpoint, CDP not supported')
+      }
+
+      const session = await provider.getCDPSession(contextId)
+      await session.send('Debugger.enable', {})
+      await session.send('Debugger.setBreakpointByUrl', {
+        lineNumber: 0, // TODO: Need to make sure test files have source map location on first line
+        urlRegex: escapePathToRegexp(file),
+      })
```
AriPerkkio, line 54: This one doesn't yet work. The demo video on description is recorded by using `lineNumber: 3` here.

Some solutions I've already tried without much luck but might we worth to retry with more time:

- Try forcing source map location in `packages/browser/src/node/plugin.ts
` like we do on Node side:

https://github.com/vitest-dev/vitest/blob/372a2c26a8e5ac2cfe6db961ed40529f09464e82/packages/vite-node/src/source-map.ts#L64-L68

- Or try using `Debugger.getPossibleBreakpoints` and `Debugger.setBreakpoint`. At first try this didn't work, as `Debugger.getPossibleBreakpoints` returned positions of **transpiled module** - we want to set breakpoints for the remapped file instead.

AriPerkkio: Using `urlRegex` instead of `url` as the files have query paramters like `?browser` etc.

AriPerkkio: It was timing issue! Breakpoint has to be set before page is navigated to.

packages/vitest/src/node/config/resolveConfig.ts
```diff
@@ -213,9 +213,8 @@ export function resolveConfig(
       && resolved.poolOptions?.threads?.singleThread
     const isSingleFork
       = resolved.pool === 'forks' && resolved.poolOptions?.forks?.singleFork
-    const isBrowser = resolved.browser.enabled
 
-    if (resolved.fileParallelism && !isSingleThread && !isSingleFork && !isBrowser) {
+    if (resolved.fileParallelism && !isSingleThread && !isSingleFork) {
```
AriPerkkio, line 217: The requirement for `--no-file-parallelism` when used with `--browser` was removed in https://github.com/vitest-dev/vitest/pull/6433 but during this PR I noticed that the flag actually affects browser mode as well. We need it in order to successfully set breakpoint to the first test file.

test/browser/specs/inspect.test.ts
```diff
@@ -0,0 +1,114 @@
...
+
+test('--inspect-brk stops at test file', async () => {
+  const { vitest, waitForClose } = await runVitestCli(
+    '--root',
+    'fixtures/inspect',
+    '--browser',
+    '--no-file-parallelism',
+    '--inspect-brk',
+    REMOTE_DEBUG_URL,
+  )
```
AriPerkkio, line 19: This test is based on Node pool's equivalent, but adding browser mode tests there didn't look right.

https://github.com/vitest-dev/vitest/blob/8d883cf05832241e6d8f97d099f6d1fdd4e231d5/test/cli/test/inspect.test.ts#L1-L92

sheremet-va (approved):

PR #6447, opened 2024-09-03
fix(vite-node): fix when dumped module file name is greater than max file name

rluvaton:
### Description

When using the `test.server.debug` option (or the corresponding env var) and the module file path that is about to be dumped + the hash is greater than the file name max limit the dump fails (this is 255 chars in most systems)

I did not add any test as currently there is no test on the debug functionality (did not find now and in the original PR #1723)

**How did I fix it:**

I just hashed twice using 2 different starting values for the djb hash function (use 2 to decrease the chance for hash collision as I suspect that the original path was part of the file name because of that)

### Tests

### Documentation

### Changesets


rluvaton:
@sheremet-va Can you please review?

sheremet-va:
We no longer have this option, replaced by [`experimental.fsModuleCache`](https://vitest.dev/config/experimental.html#experimental-fsmodulecache).

PR #6452, opened 2024-09-04
repro: vmForks bug

sheremet-va:
### Description

To run the reproduction: `pnpm run node-error-repro`. This should run tests and create files with all the output from those tests.

What happens:

- Vitest spawns several workers to run tests inside of them. One worker will run several tests one after another when it's free.
- At one point the worker (`child_process` in the reproduction, because `worker_thread` crashes the whole process) crashes while trying to import something. I've added `console` logs in our module runner implementation - it only logs the module if it's not cached already. Every time it crashes it imports the same file it looks like (because the end of the debug is always the same in all files - `node_modules/loupe/lib/helpers.js`). I don't think that file is the problem, but one of its parents maybe? Usually, it happens at least on the second test run after we create a new executor:
https://github.com/vitest-dev/vitest/blob/0499a31520243beb10243ff7d288b35acedc5fb3/packages/vitest/src/runtime/workers/vm.ts#L77
Vitest never manually clears the cache, we rely on gc, but we also reuse `fileMap` cache:
https://github.com/vitest-dev/vitest/blob/0499a31520243beb10243ff7d288b35acedc5fb3/packages/vitest/src/runtime/workers/vm.ts#L15
- The problem is most likely with how we implement VM executor - it was based on Jest's with some modifications, but I am not smart enough to figure out why. The error started happening in Node 22.1, but the changelog doesn't really mention related changes


AriPerkkio:
The `test/cli/test/vm-threads.test.ts` is also failing on CI and locally now. The failure does not happen with Node `22.0`, but happens with all versions above `22.1`.

```sh
ari ~/x/vitest/test/cli (main) $ pnpm run test vm-threads --run

> @vitest/test-cli@ test /Users/x/x/vitest/test/cli
> vitest "vm-threads" "--run"


 RUN  v2.1.1 /Users/x/x/vitest/test/cli

 ❯ test/vm-threads.test.ts (1)
 ❯ test/vm-threads.test.ts (1)
   ⠏ importing files in restricted fs works correctly

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Unhandled Errors ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Unhandled Error ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
Error: Worker exited unexpectedly
 ❯ ChildProcess.<anonymous> ../../node_modules/.pnpm/tinypool@1.0.0/node_modules/tinypool/dist/index.js:136:34
 ❯ ChildProcess.emit node:events:531:35
 ❯ ChildProcess._handle.onexit node:internal/child_process:293:12

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯

 Test Files   (1)
      Tests   (1)
     Errors  1 error
   Start at  09:34:59
   Duration  683ms (transform 43ms, setup 0ms, collect 161ms, tests 0ms, environment 0ms, prepare 46ms)

```

AriPerkkio:
Oh wow, does Node `22.10` fix this? Can't crash locally with that. `22.9` keeps crashing.

PR #6514, opened 2024-09-15
fix(browser): exports of context in package.json

IdanGoor:
### Description

I tried to run vitest in browser mode, but it failed for the error `The requested module '@vitest/browser/context' does not provide an export named 'page'`. Then I noticed that in the `package.json` we're exporting the wrong (empty) file.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/browser/package.json
+++ b/packages/browser/package.json
@@ -26,7 +26,7 @@
     },
     "./context": {
       "types": "./context.d.ts",
-      "default": "./context.js"
+      "default": "./dist/context.js"
     },
     "./client": {
       "default": "./dist/client.js"


AriPerkkio (commented):
> Then I noticed that in the `package.json` we're exporting the wrong (empty) file.

And it should:

https://github.com/vitest-dev/vitest/blob/699055eb93909287e1542fdfb99d97f2a38965ba/packages/browser/context.js#L1-L2

Importing `@vitest/browser/context` does more than just importing `dist/context.js`:

https://github.com/vitest-dev/vitest/blob/699055eb93909287e1542fdfb99d97f2a38965ba/packages/browser/src/node/plugins/pluginContext.ts#L31-L90

sheremet-va:
context is a virtual file, we put an empty file to not break static analysers

you cannot import it outside of Vitest Browser Mode

sheremet-va:
but we should probably throw our own error there

PR #6634, opened 2024-10-03
fix(browser): not.toBeInTheDocument works with locators API

sheremet-va:
### Description

Related #6560

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/browser/matchers.d.ts
+++ b/packages/browser/matchers.d.ts
@@ -16,7 +16,7 @@ declare module 'vitest' {
   type PromisifyDomAssertion<T> = Promisify<Assertion<T>>
 
   interface ExpectStatic {
-    element: <T extends Element | Locator>(element: T, options?: ExpectPollOptions) => PromisifyDomAssertion<Awaited<Element>>
+    element: <T extends Element | Locator>(element: T, options?: ExpectPollOptions) => PromisifyDomAssertion<Awaited<Element | null>>
   }
 }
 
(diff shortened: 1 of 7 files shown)


hi-ogawa (commented):
Chai flag approach is very clever, but is it necessary to support this use case?
From the discussion https://github.com/vitest-dev/vitest/discussions/6560, I feel it would also make sense to emphasize the difference between `element` and `query` like jest-dom instead of special casing `toBeInTheDocument`.

I'm sure if I'm missing some nuance, but by discussion OP's argument, I would be probably equally surprised if this happens:

```ts
// this pass
await expect.element(page.getByText('empty')).not.toBeInTheDocument()

// but this throws
await expect.element(page.getByText('empty')).not.toBeVisible()
```

Clarifying `expect.element`'s behavior and being consistent sounds better to me to confusion.

sheremet-va:
> Clarifying expect.element's behavior and being consistent sounds better to me to confusion.

Even if we just clarify how it works, there is no way to assert that the element is not in the DOM with `expect.element`. 

`not.toBeInTheDocument` is a known convention in testing-library, so that's why I am prioritizing it here

> `await expect.element(page.getByText('empty')).not.toBeVisible()`

This is a bit of a different story because it requires the change to the query itself, not the the way we retrieve the element:

```diff
- await expect.element(page.getByText('empty')).not.toBeVisible()
+ await expect.element(page.getByText('empty', { force: true })).not.toBeVisible()
```

This can just be clarified in the docs.

hi-ogawa (approved):
> Even if we just clarify how it works, there is no way to assert that the element is not in the DOM with `expect.element`.
> 
> `not.toBeInTheDocument` is a known convention in testing-library, so that's why I am prioritizing it here

Okay, sounds good. `expect.element` expecting `Element` seemed simpler mental model, but I guess it's worth having this.

PR #6405, opened 2024-08-26
refactor: proposal for the public Vitest API

sheremet-va:
sneaky draft proposal, look away

related #6147

```ts
// The main Vitest class
export interface Vitest {
  // the main config from the root of the project
  readonly config: ResolvedConfig
  // instances of projects defined in the workspace config
  // always has at least one project
  readonly projects: Map<string, TestProject>
  // global Vite server instance
  readonly vite: ViteDevServer
  // class that runs tests and stores results
  readonly runner: VitestRunner
  // global inject/provide context
  readonly context: VitestContext
  // snapshot state
  readonly snapshot: VitestSnapshot
  readonly logger: Logger

  onClose(cb: () => void): void

  close(): Promise<void>
  exit(): Promise<void>
}

interface TestProject {
  readonly config: ResolvedProjectConfig
  readonly globalConfig: ResolvedConfig // same as vitest.config
  readonly name: string
  // local project context
  readonly context: VitestContext
  readonly vite: ViteDevServer
  // config that can be sent as a json
  readonly serializedConfig: SerializedConfig
  // browser server instance if tests are running in the browser
  readonly browser: BrowserServer | null
  // an easy way to create a spec
  createSpecification(moduleId: string): TestSpecification
}

export interface VitestRunner {
  // Vitest starts a standalone runner, will react on watch changes, it doesn't run tests
  start: () => void
  stop: () => void

  // Map<filepath, spec[]>
  // unique by project+pool, creating a new spec overrides the old one
  readonly specifications: Map<string, Array<TestSpecification>>

  run: () => Promise<TestRunResult>
  collect: () => Promise<TestRunResult>
  // run module in every project
  runModules: (moduleNames: string[]) => Promise<TestRunResult>
  // run specific tests
  runTests: (specifications: Array<TestSpecification>) => Promise<TestRunResult>
  mergeReports: () => Promise<TestRunResult>
}

interface TestRunResult {
  testModules: TestModule[]
  errors: TestError[]
}
```

PR #6641, opened 2024-10-04
feat(ui): allow run individual tests/suites from the UI

### LINKED ISSUE #6638: [UI] Allow running individual tests/suites

At the moment, Vitest UI allows re-running all tests in a file.
However, it is commonly useful to re-run just a single test or a suite of tests.
The option to do so is currently disabled in the UI (as this capability is not yet implemented).

### Suggested solution

On however on a test/suite in the sidebar, show a button to re-run a given test/suite.
This would be similar functionality to the "Testing" panel in VS Code.

### Alternative

- Run all tests in a file (slow if file has many tests)
- Stop Vitest, and re-run it with `-t 'somePattern'` argument passed to run only matching tests (clunky)

### Additional context

Original discussion: https://github.com/vitest-dev/vitest/pull/5948#discussion_r1786614993

### Validations

userquin:
### Description

**We need this PR https://github.com/vitest-dev/vitest/pull/6405**

Since we don't have the full tree in the state (missing parent in task), the client will send the task id, the children ids and all parent ids: parent ids are required to avoid mark the test not being run as skipped and so their state will not change.

We also need to reset `configOverride.testNamePattern` when re-running some file (including all files), otherwise we can only re-run previous test/suite run, maybe we can have some side effect here (I need to do some additional tests).

>[!CAUTION]
> Right now the command line will skip other tests/suites not matching the pattern (using t + regexp): the problem with the UI is about enabling/disabling re-run button, disabled when test/suite without state or state with `skip` or `todo` value (check _command line filter behaviour_ screenshot below)

>[!WARNING]
> We should add the `test "path"` when filtering tests/suites (check _filtering from command line_ screenshot below)

resolves #6638

<details>
<summary><i>filtering from command line</i></summary>

imagen
</details>

<details>
<summary><i>command line filter behaviour</i></summary>

imagen
</details>

### Tests

### Documentation

### Changesets


packages/ui/client/composables/client/index.ts
```diff
@@ -68,6 +69,21 @@ export function runAll() {
   return runFiles(client.state.getFiles()/* , true */)
 }
 
+function clearTaskResult(patterns: string[], task: Task) {
```
userquin, line 72: rename `patterns` to `ids`

userquin: not required

packages/vitest/src/node/core.ts
```diff
@@ -688,11 +693,44 @@ export class Vitest {
...
+    }
+  }
+
+  async rerunTestOrSuite(id: string, filename: string) {
+    const patterns: string[] = []
+    const task = this.state.idMap.get(id)
+    let trigger = 'rerun test file'
+    if (task) {
+      trigger = this.isSuite(task) ? 'rerun suite' : 'rerun test'
+      // this.collectTasks(task, patterns)
```
userquin, line 717: cleanup this

packages/ui/client/components/explorer/ExplorerItem.vue
```diff
@@ -219,12 +238,12 @@ const projectNameTextColor = computed(() => {
...
         v-if="!isReport"
-        v-tooltip.bottom="'Run current test'"
+        v-tooltip.bottom="runButtonTitle"
         data-testid="btn-run-test"
-        title="Run current test"
+        :title="runButtonTitle"
         icon="i-carbon:play-filled-alt"
         text-green5
-        :disabled="type !== 'file'"
+        :disabled="state === 'todo'"
```
userquin, line 246: review this including `standalone` check, we shouldn't allow run skipped tests (check `ui.spec.ts` and the `standalone` suite: all files when using `standalone` initialized with `state` set to `skip`)

packages/ui/client/components/explorer/ExplorerItem.vue
```diff
@@ -68,12 +69,22 @@ function toggleOpen() {
 }
 
 async function onRun(task: Task) {
+  if (state === 'todo') {
```
sheremet-va, line 72: state can be dynamic (you can call `ctx.skip`), so let's never assume that the `result.state` dictates if we can attempt to run it

it is just a reported state, it shouldn't block an execution attempt

packages/ui/client/components/explorer/ExplorerItem.vue
```diff
@@ -68,12 +69,22 @@ function toggleOpen() {
...
+  }
+
   onItemClick?.(task)
   if (coverageEnabled.value) {
     disableCoverage.value = true
     await nextTick()
   }
-  await runFiles([task.file])
+
+  if (type === 'file') {
```
sheremet-va, line 82: Can't we also run folders?

packages/ui/client/components/explorer/ExplorerItem.vue
sheremet-va: we should ignore any state, `disabled` should never be `true` here

packages/vitest/src/api/setup.ts
```diff
@@ -70,8 +70,11 @@ export function setup(ctx: Vitest, _server?: ViteDevServer) {
           }
           return fs.writeFile(id, content, 'utf-8')
         },
-        async rerun(files) {
-          await ctx.rerunFiles(files)
+        async rerun(files, resetTestNamePattern) {
+          await ctx.rerunFiles(files, undefined, resetTestNamePattern)
+        },
+        async rerunTestOrSuite(id, filename) {
```
sheremet-va, line 76: `testOrSuite` can be shortened to `task`

sheremet-va (changes requested):

packages/ui/client/components/explorer/ExplorerItem.vue
userquin: we don't have folders structure in the ui, we only have the tasks from the rpc call (collect and getTestFiles/getFiles): if the new api has a new folder type we'll need to change the ui

packages/vitest/src/node/core.ts
```diff
@@ -688,11 +693,28 @@ export class Vitest {
...
+    return Object.hasOwnProperty.call(task, 'tasks')
+  }
+
+  async rerunTask(id: string) {
+    const task = this.state.idMap.get(id)
+    if (task) {
+      await this.changeNamePattern(
+        task.name,
+        [task.file.filepath],
+        this.isSuite(task) ? 'rerun suite' : 'rerun test',
```
sheremet-va, line 706: `task.type === 'suite'` is how you check for suite

packages/vitest/src/node/core.ts
```diff
@@ -688,11 +693,28 @@ export class Vitest {
     await this.report('onWatcherStart', this.state.getFiles(files))
   }
 
+  private isSuite(task: RunnerTask): task is RunnerTestSuite {
+    return Object.hasOwnProperty.call(task, 'tasks')
+  }
+
+  async rerunTask(id: string) {
+    const task = this.state.idMap.get(id)
+    if (task) {
```
sheremet-va, line 702: maybe we should throw an error if task it not found?

sheremet-va (changes requested):

packages/vitest/src/node/core.ts
userquin: yeah, I'll ask you about this... or maybe we can rerun all files or passing the filename back again and rerun all files

sheremet-va: I think an error is better

sheremet-va (dismissed):

sheremet-va:
Looks good! There are a few conflicts though

sheremet-va:
Sorry to ask, but can you resolve the conflict again?

Karthik0521:
🏆 You decided: These are the winners of Boosting Cycle #3

PR #6676, opened 2024-10-10
docs: global type recommendation

privatenumber:

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -353,7 +353,7 @@ Vitest uses Vite SSR primitives to run tests which has [certain pitfalls](https:
 - **Default:** `false`
 - **CLI:** `--globals`, `--globals=false`
 
-By default, `vitest` does not provide global APIs for explicitness. If you prefer to use the APIs globally like Jest, you can pass the `--globals` option to CLI or add `globals: true` in the config.
+By default, `vitest` does not provide global APIs for testing to encourage explicitness and reduce potential conflicts. If you prefer using global APIs like Jest, you can enable them by passing the `--globals` option in the CLI or setting `globals: true` in your config file.
 
 ```ts
 // vitest.config.ts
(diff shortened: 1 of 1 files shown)


docs/config/index.md
```diff
@@ -366,18 +366,37 @@ export default defineConfig({
...
-```
+To ensure TypeScript recognizes the global APIs, you have two options:
+
+1. Import the global types
+
+    Do a type import from `vitest/globals` in any `.ts` file (e.g., your [setup file](#setupfiles)). This registers the global APIs types automatically. Ensure the file is included in your `tsconfig.json` so TypeScript knows it's apart of the same project.
+
+    ```ts
+    // test-setup.ts
+    import type {} from 'vitest/globals'
```
privatenumber, line 379: If this can be updated to also include an empty file for `default`:
https://github.com/vitest-dev/vitest/blob/main/packages/vitest/package.json#L37-L39

We can simplify the recommendation to `import 'vitest/globals'`

docs/config/index.md
```diff
@@ -366,18 +366,45 @@ export default defineConfig({
...
-```
+To register global API types, follow one of the following methods:
+
+1. Use a `@types/vitest.d.ts` file
+
+    Create a `@types/vitest.d.ts` file in your project root that imports `vitest/globals`:
+
+    ```ts
+    import 'vitest/globals'
+    ```
```
privatenumber, line 379: In my experiments, this is the best method because it's clear, doesn't require any hack like `import type {}`, doesn't pollute source code, and works even if `types`/`typeRoots` are overridden.

docs/config/index.md
```diff
@@ -366,18 +366,45 @@ export default defineConfig({
...
+    ```ts
+    // test-setup.ts
+    import type {} from 'vitest/globals'
+    ```
+
+3. Configure it in [`tsconfig.json#compilerOptions.types`](https://www.typescriptlang.org/tsconfig/#types)
+
+    Alternatively, add `vitest/globals` to the `types` field in `tsconfig.json`.
+
+    Warning: This disables TypeScript's automatic type detection, which could break your setup if your project relies on it. If you encounter types missing from other packages, you'll now need to manually register them here as well.
```
privatenumber, line 394: I think we should not even include this method as many beginners have been following this without understanding how disruptive it is and makes TS harder for them to work with.

docs/config/index.md
sheremet-va: Maybe we should also mention it here in the docs?

docs/config/index.md
sheremet-va: Vitepress has a `::: danger` block we can use here

sheremet-va:
We are going to align with Vite here: https://github.com/vitejs/vite/pull/20132 (so, keep what we have already)

TypeScript also plans to set `types` to `[]` by default, so it seems like TS team is moving into that direction: https://github.com/microsoft/TypeScript/issues/62195

PR #6679, opened 2024-10-10
fix(ui): restore splitpanes sizes

### LINKED ISSUE #6678: UI splitpanes sizes are no longer persisted to local storage
### Describe the bug

It seems that there was a regression of https://github.com/vitest-dev/vitest/pull/5166

- `vitest --ui`
- Resize splitpanes on the left
- reload page
- the left pane resets to the default size

I think that the `initOnMounted` option no longer takes effect since they were moved outside of a component. Maybe they should be exposed as functions instead of variables `const useDetailsSizes = () => useLocalStorage(...)`

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-enxjhj?file=README.md&initialPath=__vitest__/

(default repro, might have outdated deps but I tested locally too)

### System Info

```shell
na
```

### Used Package Manager

pnpm

### Validations

userquin:
### Description

Cannot test browser ui on my local Windows laptop, we need to check the side effect opening UI (test/core) and then the Browser UI (test/browser running any any `test:browser:*` script)

resolves #6678

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/ui/client/pages/index.vue
+++ b/packages/ui/client/pages/index.vue
@@ -42,12 +42,14 @@ const resizingMain = useDebounceFn((event: { size: number }[]) => {
 function resizeMain() {
   const width = window.innerWidth
   const panelWidth = Math.min(width / 3, 300)
-  mainSizes.value[0] = (100 * panelWidth) / width
-  mainSizes.value[1] = 100 - mainSizes.value[0]
-  recordMainResize([
-    { size: mainSizes.value[0] },
-    { size: mainSizes.value[1] },
-  ])
+  if (browserState) {
+    mainSizes.value[0] = (100 * panelWidth) / width
+    mainSizes.value[1] = 100 - mainSizes.value[0]
+  }
+  else {
+    detailSizes.value[0] = (100 * panelWidth) / width
+    detailSizes.value[1] = 100 - detailSizes.value[0]
+  }
 }
 
 function recordMainResize(event: { size: number }[]) {


packages/ui/client/pages/index.vue
```diff
@@ -42,12 +42,14 @@ const resizingMain = useDebounceFn((event: { size: number }[]) => {
 function resizeMain() {
   const width = window.innerWidth
   const panelWidth = Math.min(width / 3, 300)
-  mainSizes.value[0] = (100 * panelWidth) / width
-  mainSizes.value[1] = 100 - mainSizes.value[0]
-  recordMainResize([
-    { size: mainSizes.value[0] },
-    { size: mainSizes.value[1] },
-  ])
+  if (browserState) {
```
sheremet-va, line 45: Why did it suddenly stop recording the main resize? `recordMainResize`

sheremet-va:
> Cannot test browser ui on my local Windows laptop, we need to check the side effect opening UI (test/core) and then the Browser UI (test/browser running any `test:browser:*` script)

Try deleting the `node_modules` folder and reinstalling dependencies

packages/ui/client/pages/index.vue
userquin: We drop the middle panel in the default ui

sheremet-va: But we don't drop it in the browser

sheremet-va: Are you saying it is ok to break the browser UI?

userquin: No, I need to test the changes in the browser ui, this change just fix the ui.

This shouldn't break the browser ui, the logic is the same, we only initilize the ui panels properly.

The problem is about using the same storage key for both, the ui and the browser ui.

userquin:
Supersede by #7265

PR #6703, opened 2024-10-14
perf: use `hash` to replace `createHash`

btea:
### Description
`crypto.hash` seems to be faster than `crypto.createHash`.

https://nodejs.org/docs/latest-v20.x/api/crypto.html#cryptohashalgorithm-data-outputencoding)

I did a simple test and `crypto.hash` seems to be 1.6 times faster.

image

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/utils/index.ts
+++ b/packages/vitest/src/utils/index.ts
@@ -1,3 +1,4 @@
+import crypto from 'node:crypto'
 import { relative } from 'pathe'
 import { getWorkerState } from '../utils'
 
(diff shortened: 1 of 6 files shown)


sheremet-va (dismissed):

packages/vitest/src/utils/index.ts
```diff
@@ -1,3 +1,4 @@
+import crypto from 'node:crypto'
```
sheremet-va, line 1: Ah, actually. This file can be imported in any environment. The utility should be in node-specific utils file

btea: Should I create a separate file or does such a file already exist and I need to move this function into it?

sheremet-va: Creating a separate file should be fine

btea: Ok.

sheremet-va (approved):

PR #6803, opened 2024-10-28
feat(vitest): add `onTestsRerun` method to global setup context

### LINKED ISSUE #6417: Adding custom watch commands

We use [testcontainers](https://testcontainers.com/) to run our integration tests with a live PostgreSQL database, in 5 parallel workers

All these DBs must be migrated using Prisma before starting `vitest --watch`, we do this in our global setup:

```typescript
import { PostgreSqlContainer } from "@testcontainers/postgresql";
import { exec as execCb } from "child_process";
import { promisify } from "util";
import type { GlobalSetupContext } from "vitest/node";

const exec = promisify(execCb);

export default async function setup({ provide, config }: GlobalSetupContext) {
  const stopsFunctions: (() => Promise<void>)[] = [];
  const postgresUrls: Record<number, string> = {};

  await prismaGenerate();

  await Promise.all(
    [...Array(config.maxWorkers).keys()].map(async (index) => {
      const postgres = await buildPostgresContainer();
      postgresUrls[index + 1] = postgres.url;
      stopsFunctions.push(postgres.stop);
    }),
  );

  provide("postgresUrls", postgresUrls);

  return async () => {
    await Promise.all(
      stopsFunctions.map(async (stop) => {
        await stop();
      }),
    );
  };
}

const prismaGenerate = async () => {
  console.log("[Prisma] Generating...");
  await exec(`npx prisma generate`);
  console.log("[Prisma] Generated successfully!");
};

const buildPostgresContainer = async () => {
  console.log("[Postgres] Starting...");
  const postgres = await new PostgreSqlContainer("postgres:16")
    .withTmpFs({ "/dev/shm/postgres": "rw,noexec,nosuid,size=65536k" })
    // https://www.postgresql.org/docs/current/non-durability.html
    .withCommand([
      "postgres",
      "-c",
      "fsync=off",
      "-c",
      "synchronous_commit=off",
      "-c",
      "full_page_writes=off",
      "-c",
      "autovacuum=off",
    ])
    .start();

  const url = postgres.getConnectionUri();

  console.log("[Postgres] Migrating...");
  await exec(`DB_URL=${url} npx prisma migrate deploy`);
  console.log("[Postgres] Migrated successfully!");

  return {
    url,
    stop: async () => {
      await postgres.stop();
    },
  };
};

// Type safe access to `provide/inject` methods:
declare module "vitest" {
  export interface ProvidedContext {
    postgresUrls: Record<number, string>;
  }
}
```

If I write a new migration, it is not applied to my live test databases, I must end the Vitest process and start again (in this case, this takes ~30s because of all the containers)

### Suggested solution

Would be nice if we could add a new quick command to the watch menu, such as `m` for triggering migrations inside the databases

### Alternative

We could setup a parallel watching process which runs the migrations every time the migrations folder changes, but that seems flaky

Perhaps related to https://github.com/vitest-dev/vitest/issues/842

### Additional context

_No response_

### Validations

sheremet-va:
### Description

Fixes #6417

### Tests

### Documentation

### Changesets

### DIFF
--- a/test/watch/test/global-setup-rerun.test.ts
+++ b/test/watch/test/global-setup-rerun.test.ts
@@ -3,7 +3,7 @@ import { editFile, runVitest } from '../../test-utils'
 
 const testFile = 'fixtures/math.test.ts'
 
-test('global setup calls hooks correctly', async () => {
+test('global setup calls hooks correctly when file changes', async () => {
   process.env.TEST_GLOBAL_SETUP = 'true'
   const { vitest, ctx } = await runVitest({
     root: 'fixtures',
@@ -17,6 +17,29 @@ test('global setup calls hooks correctly', async () => {
   expect(calls).toEqual(['start'])
 
   editFile(testFile, testFileContent => `${testFileContent}\n\n`)
+
+  await vitest.waitForStdout('RERUN')
+  expect(calls).toEqual(['start', 'rerun'])
+
+  await ctx?.close()
+
+  expect(calls).toEqual(['start', 'rerun', 'end'])
+})
+
+test('global setup calls hooks correctly with a manual rerun', async () => {
+  process.env.TEST_GLOBAL_SETUP = 'true'
+  const { vitest, ctx } = await runVitest({
+    root: 'fixtures',
+    watch: true,
+    include: ['math.test.ts'],
+  })
+
+  await vitest.waitForStdout('Waiting for file changes')
+
+  const calls = (globalThis as any).__CALLS as string[]
+  expect(calls).toEqual(['start'])
+
+  vitest.write('r')
 
   await vitest.waitForStdout('RERUN')
   expect(calls).toEqual(['start', 'rerun'])
--- a/test/watch/test/global-setup-rerun.test.ts
+++ b/test/watch/test/global-setup-rerun.test.ts
@@ -1,0 +1,27 @@
+import { expect, test } from 'vitest'
+import { editFile, runVitest } from '../../test-utils'
+
+const testFile = 'fixtures/math.test.ts'
+
+test('global setup calls hooks correctly', async () => {
+  process.env.TEST_GLOBAL_SETUP = 'true'
+  const { vitest, ctx } = await runVitest({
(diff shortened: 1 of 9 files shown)


test/watch/test/global-setup-rerun.test.ts
```diff
@@ -0,0 +1,27 @@
...
+    watch: true,
+    include: ['math.test.ts'],
+  })
+
+  await vitest.waitForStdout('Waiting for file changes')
+
+  const calls = (globalThis as any).__CALLS as string[]
+  expect(calls).toEqual(['start'])
+
+  editFile(testFile, testFileContent => `${testFileContent}\n\n`)
```
AriPerkkio, line 19: I think we could use `'r'` keypress here instead. Editing files is heavy on CI.

```ts
  vitest.write('r')

  await vitest.waitForStdout('RERUN')
```

sheremet-va: But we are testing that _watcher_ triggered a callback

sheremet-va: There should be two tests then, not one

AriPerkkio: If we are testing file editing here, then yes. If it's just about testing whether rerun calls this callback, then it's not important.

sheremet-va: We are testing everything that affects this, I don't think about implementation details when testing it. When user does something, something else should happen

AriPerkkio (dismissed):
Looks good to me, I like this idea! 💯 

I wonder if at some point we'll have "plugins" that are registered in the global setup for more hooks. We'll see where this API lands eventually 🤔

hi-ogawa (approved):
I tested this to make arktype/attest's global setup rerun-able https://github.com/hi-ogawa/reproductions/tree/main/vitest-attest-snapshot It looks working nicely :clap:

AriPerkkio (approved):

hi-ogawa (approved):

PR #6868, opened 2024-11-06
docs: remove X links

patak-dev:
### Description

We're seeing more activity in the Bluesky account. I think it is better for Vitest users to use that profile (that is public, even without an account) instead of the X one.

This PR removes links to X in the docs. We're also going to remove the link the GitHub org.

I'm also removing the X link in my team profile. @vitest-dev/team let me know with a comment if you'd like the same for yours so we batch most of them in this PR (or feel free to directly PR the change).


sheremet-va:
You can remove mine

userquin:
You can remove mine

Aslemammad:
Let's do it for mine too!

antfu:
Same for mine, thanks!

AriPerkkio:
Let's remove mine too, thanks!

PR #6906, opened 2024-11-14
refactor: merge `TestProject` with `WorkspaceProject`

sheremet-va:
### Description

In preparation for the public API. Previously `TestProject` was a prototype for the public API, but it was hard to work with internally because it was too restrictive. In this PR I decided to merge them together and introduce a sort of convention:

- All public methods starting with `_` and marked with `@internal` are internal and not part of the public API. They are striped from the public interface during bundling
- All those methods were renamed, and the previous names stayed with a `@deprecated` tag - they will be removed in V3 (we can remove them right away, they shouldn't have been used at all)
- The `WorkspaceProject` was renamed to `TestProject` - all properties from the test project are now present there. The file was also renamed from `workspace` to `project`
- Added a getter to throw an error if `.vite` or `.config` was accessed before the project was initialised. This should catch unexpected errors with a nice stack trace

The public test project API now looks like this (excluding deprecated and private APIs):

```ts
export interface TestProject {
  readonly name: string
  readonly serializedConfig: SerializedConfig
  readonly globalConfig: ResolvedConfig
  readonly vitest: Vitest
  readonly config: ResolvedConfig // project config
  readonly vite: ViteDevServer
  browser?: BrowserServer
  
  provide(key: string | symbol, value: unknown): void
  getProvidedContext(): ProvidedContext
  createSpecification(moduleId: string, pool?: string): TestSpecification
  isRootProject(): boolean
  globTestFiles(filters?: string[]): Promise<{ testFiles, typecheckTestFiles }>
  matchesTestGlob(filepath: string, code?: string): boolean
  close(): Promise<void>
}
```

image

TODO:
- [ ] Better docs

### Tests

### Documentation

### Changesets


docs/advanced/api.md
```diff
@@ -123,15 +127,220 @@ declare module 'vitest' {
...
+]
+```
+```ts [vitest.workspace.js]
+export default [
+  './packages/server', // has package.json with "@pkg/server"
+  './utils', // doesn't have a package.json file
+  {
+    // doesn't customize the name
+    test: {
+      pool: 'node',
```
AriPerkkio, line 170: There is no `pool: 'node'`

packages/vitest/src/node/core.ts
```diff
@@ -204,42 +205,38 @@ export class Vitest {
...
+    return this.projects.find(p => p.name === projectName)
+      || this.getRootTestProject()
       || this.projects[0]
   }
 
   public getProjectByName(name: string = '') {
-    return this.projects.find(p => p.getName() === name)
-      || this.getCoreWorkspaceProject()
+    return this.projects.find(p => p.name === name)
+      || this.getRootTestProject()
```
AriPerkkio, line 239: This public method can now throw so we should use direct access instead:

```suggestion
      || this.coreWorkspaceProject
```

packages/vitest/src/node/logger.ts
```diff
@@ -163,8 +163,7 @@ export class Logger {
     }
     this.ctx.projects.forEach((project) => {
       const config = project.config
-      const name = project.getName()
-      const output = project.isCore() || !name ? '' : `[${name}]`
+      const output = project.isRootProject() || !project.name ? '' : `[${project.name}]`
```
AriPerkkio, line 166: Does this require brackets?

```suggestion
      const output = (project.isRootProject() || !project.name) ? '' : `[${project.name}]`
```

packages/vitest/src/node/project.ts
```diff
@@ -38,95 +38,68 @@ import { loadGlobalSetupFiles } from './globalSetup'
...
-  await createViteServer(config)
+  /**
+   * Browser instance if the browser is enabled. This is initialized when the tests run for the first time.
+   */
+  public browser?: BrowserServer
 
-  return project
-}
+  /** @deprecated use `vitest` instead */
+  public ctx: Vitest
```
AriPerkkio, line 63: Nice 👍

packages/vitest/src/node/project.ts
```diff
@@ -138,26 +111,108 @@ export class WorkspaceProject {
         },
       )
     }
+    // casting `any` because the devault type is `never` since `ProvidedContext` is empty
```
AriPerkkio, line 114:
```suggestion
    // casting `any` because the default type is `never` since `ProvidedContext` is empty
```

packages/vitest/src/node/project.ts
```diff
@@ -138,26 +111,108 @@ export class WorkspaceProject {
...
+
+  /** @deprecated use `vite` instead */
+  public get server(): ViteDevServer {
+    return this._vite!
+  }
+
+  /**
+   * Is this the root project. The root project is the one that has the root config.
+   */
+  public isRootProject(): boolean {
```
AriPerkkio, line 195: Consider:

```suggestion
  /**
   * Check if this is the root project. The root project is the one that has the root config.
   */
  public isRootProject(): boolean {
```

AriPerkkio (commented):
The overall direction is great! Thanks for working on this 🙌

docs/advanced/api.md
sheremet-va:
```suggestion
      pool: 'threads',
```

hi-ogawa (commented):
I wasn't aware of the much of API, but I like the idea of merging and simplifying `WorkspaceProject` and `TestProject` :+1:

Issue #6913, opened 2024-11-15
Inline workspace


sheremet-va:
### Clear and concise description of the problem

It would be nice to support inline workspace configuration:

```ts
export default defineConfig({
  test: {
    workspace: [
      {
        test: { 
          name: 'node', 
          environment: 'node'
        }, 
      },
      {
        test: { 
          name: 'jsdom', 
          environment: 'jsdom'
        }, 
      }
    ]
  }
})
```

This feature is required for https://github.com/vitest-dev/vitest/issues/6912 to make the migration easier.

### Suggested solution

Alllow inlined `workspace` in the main config.

### Additional context

Some questions:

1. Should inline workspace override the workspace file? Or should they be merged instead?
2. Is there a way to reduce the nesting?

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #6923, opened 2024-11-18
feat: allow inline workspace configuration

sheremet-va:
### Description

Fixes https://github.com/vitest-dev/vitest/issues/6913

This PR allows specifying an array in `test.workspaces` option. This will override the `vitest.workspace` file if there is one:

```ts
import { defineConfig } from 'vitest'
export default defineConfig({
  test: {
    workspace: [
      './packages/*',
      {
        test: {
          name: 'inline',
        },
      },
    ],
  },
})
```

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/advanced/api.md
+++ b/docs/advanced/api.md
@@ -140,7 +140,7 @@ export default function setup({ provide }) {
 ```
 :::
 
-## TestProject <Version>2.2.0</Version>
+## TestProject <Version>2.2.0</Version> {#testproject}
 
 - **Alias**: `WorkspaceProject` before 2.2.0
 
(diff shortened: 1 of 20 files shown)


hi-ogawa (approved):

AriPerkkio (dismissed):
Looks good, this will be really handy!

In some comments the `extends: true` says it makes project inherit all options from root even when it inherits only the ones that can be used in project level settings. Maybe that could be made more specific. 🤔

sheremet-va:
> Looks good, this will be really handy!
> 
> 
> 
> In some comments the `extends: true` says it makes project inherit all options from root even when it inherits only the ones that can be used in project level settings. Maybe that could be made more specific. 🤔 

I don't understand what you are saying. `extends` will merge the project config with root config inheriting all properties. If some of them are root-only, then it doesn't really matter when user doesn't override them (and they can't override because they are root only)

PR #6932, opened 2024-11-19
fix(reporters): write buffered stdout/stderr on process exit

AriPerkkio:
### Description

- When `logger.ts` writes errors for unhandled rejections, it exits immediatelly. The windowed-renderer intercepts all `process.stdout/err` writes and buffers them. The buffered content must be written before exit.

https://github.com/vitest-dev/vitest/blob/78f07c0537c63d9adb2bf24a59afea988890ae4d/packages/vitest/src/node/logger.ts#L306-L317

- Moves `signal-exit` from transitive deps to direct deps 
- I wasn't able to come up with failing test case for this. We are overriding `process.exit` during tests. 🤔 

### Tests

### Documentation

### Changesets


packages/vitest/src/node/reporters/renderers/windowedRenderer.ts
```diff
@@ -54,6 +55,9 @@ export class WindowRenderer {
     restoreCursor()
     this.write(HIDE_CURSOR, 'output')
 
+    // Write buffered content on unexpected exits, e.g. direct `process.exit()` calls
+    onExit(() => this.flushBuffer())
```
sheremet-va, line 59: doesn't `process.on('exit')` work? 🤔

AriPerkkio: `exit` is not called on SIGTERM, e.g. `ctrl + c`. 

But now we can also replace `restore-cursor` with direct usage of `signal-exit`.

sheremet-va: `process.on('SIGTERM')` should be called on SIGTERM 🤔

sheremet-va: I just don't like overriding `process.reallyExit`

sheremet-va: `onExit` watches these signals: 

```
signals.push('SIGHUP', 'SIGINT', 'SIGTERM');
29if (process.platform !== 'win32') {
30    signals.push('SIGALRM', 'SIGABRT', 'SIGVTALRM', 'SIGXCPU', 'SIGXFSZ', 'SIGUSR2', 'SIGTRAP', 'SIGSYS', 'SIGQUIT', 'SIGIOT'
31    // should detect profiler and enable/disable accordingly.
32    // see #21
33    // 'SIGPROF'
34    );
35}
36if (process.platform === 'linux') {
37    signals.push('SIGIO', 'SIGPOLL', 'SIGPWR', 'SIGSTKFLT');
38}
```

sheremet-va (dismissed):
I'll be happy if we can remove `signal-exit`, but it's also fine to merge it as is

AriPerkkio:
Do we want to implement our own `signal-exit`? We are already using it via `log-update > cli-cursor > restore-cursor > signal-exit`.

sheremet-va:
> Do we want to implement our own `signal-exit`? We are already using it via `log-update > cli-cursor > restore-cursor > signal-exit`.

I was hoping we would be able to drop it when `log-update` is removed 😄

sheremet-va:
If we need it for stdout handling, then sure let's keep it

AriPerkkio:
I can take a look at removing it completely once rest of `log-update` usage is removed. Though I'm not familiar with `process.reallyExit` and don't know why `signal-exit` overwrites it. Ideally we would not override anything, just hook into `exit` and the signals.

AriPerkkio:
Looking at how `picospinner` does this, I think we can follow their implementation: https://github.com/tinylibs/picospinner/pull/2/files (cc. @43081j)

I'll try in this PR to replace `signal-exit` with zero-deps solution.

packages/vitest/src/node/reporters/renderers/windowedRenderer.ts
```diff
@@ -175,6 +174,38 @@ export class WindowRenderer {
   private write(message: string, type: 'output' | 'error' = 'output') {
     (this.streams[type] as Writable['write'])(message)
   }
+
+  private addProcessExitListeners() {
```
AriPerkkio, line 178: At this point it's ok to have this here in `windowedRenderer`. But later on it might make sense to move this to `logger.ts` instead. 🤔

packages/vitest/src/node/reporters/renderers/windowedRenderer.ts
```diff
@@ -175,6 +174,38 @@ export class WindowRenderer {
...
+    const onExit = (signal?: string | number) => {
```
AriPerkkio, line 179: Use exit code that's passed here by Node:

```suggestion
    const onExit = (signal?: string | number, code?: number) => {
```

hi-ogawa (approved):

Issue #6736, opened 2024-10-17
Wrong coverage when using isolate: false


RodrigoFerretti:
### Describe the bug

This is the same bug as mentioned in https://github.com/vitest-dev/vitest/issues/3846, but now with minimal reproduction.

I've read somewhere that vitest orders the execution by the file's size, descending.

The coverage is ignored for the smaller tests when the method being tested is called in one of the "setupFiles" configured, even though everything should be covered

### Reproduction

https://github.com/RodrigoFerretti/vitest-coverage-w-isolate

### System Info

```shell
System:
    OS: Linux 5.15 Ubuntu 22.04.4 LTS 22.04.4 LTS (Jammy Jellyfish)
    CPU: (8) x64 Intel(R) Core(TM) i7-10610U CPU @ 1.80GHz
    Memory: 5.28 GB / 15.23 GB
    Container: Yes
    Shell: 5.8.1 - /usr/bin/zsh
  Binaries:
    Node: 18.19.0 - ~/.nvm/versions/node/v18.19.0/bin/node
    Yarn: 1.22.22 - ~/.yarn/bin/yarn
    npm: 10.2.3 - ~/.nvm/versions/node/v18.19.0/bin/npm
  Browsers:
    Chrome: 127.0.6533.72
  npmPackages:
    @vitest/coverage-v8: 2.0.3 => 2.0.3 
    vite: 5.3.4 => 5.3.4 
    vitest: 2.0.3 => 2.0.3
```

### Used Package Manager

yarn

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

AriPerkkio:
Does this happen when setup files are not importing source files?

RodrigoFerretti:
[...]

PR #6935, opened 2024-11-19
fix(coverage): `isolate: false` with istanbul provider

AriPerkkio:
### Description

- Related to https://github.com/vitest-dev/vitest/issues/6736 but does not fix `@vitest/coverage-v8`

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/coverage-istanbul/src/index.ts
+++ b/packages/coverage-istanbul/src/index.ts
@@ -1,33 +1,48 @@
+import type { CoverageMapData } from 'istanbul-lib-coverage'
 import type { IstanbulCoverageProvider } from './provider'
 import { COVERAGE_STORE_KEY } from './constants'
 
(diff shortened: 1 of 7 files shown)


packages/coverage-istanbul/src/index.ts
```diff
@@ -1,33 +1,48 @@
...
+  startCoverage() {
+    // @ts-expect-error -- untyped global
+    const coverageMap = globalThis[COVERAGE_STORE_KEY] as CoverageMapData
 
-  return coverage
-}
+    // When isolated, there are no previous results
+    if (!coverageMap) {
+      return
+    }
```
AriPerkkio, line 19: Doing the reset on `startCoverage` instead of `stopCoverage` should have better perf, as isolated tests won't ever run into this. Only non-isolated tests can have previous results in global scope.

PR #6943, opened 2024-11-21
fix(reporters): rewrite `dot` reporter without `log-update`

AriPerkkio:
### Description

- Fixes issues where `log-update` hides `process.stdout` calls that come outside Vitest, e.g. NodeJS deprecation warnings. This is visible for example in Vite repo's `test-serve` command. 
- Fixes issues where `log-update` got confused of row count and started to duplicate rows
- Improves performance as now we don't render finished tests all the time. There's some logic in `DotSummary`'s `onTestFileFinished` that optimizes this.
- Removes usage of `log-update` from `dot` reporter.

### Tests

### Documentation

### Changesets


packages/vitest/src/node/reporters/dot.ts
```diff
@@ -1,53 +1,182 @@
...
+    this.ctx.logger.log(finalLog)
+
+    this.tests.clear()
+    this.renderer.finish()
   }
 
-  async onWatcherRerun(files: string[], trigger?: string) {
-    await this.stopListRender()
-    super.onWatcherRerun(files, trigger)
+  onTestFilePrepare(file: File): void {
```
sheremet-va, line 74: I just noticed we are still using the old API here... We need to experiment with the new one to see what's missing. Otherwise, it will never be out of the experimental

I don't know if we should do it here, but keep in mind that it needs to be rewritten before 2.2 (I can do it myself, if no one has resources)

sheremet-va (approved):

packages/vitest/src/node/reporters/dot.ts
AriPerkkio: You mean the `File` part, as in we should use `state.getReportedEntity` instead?

When building the summary reporter I had at first that one in use and also some new Reporter API hooks. But as nothing was working well I started to reduce amount of changes. First fix all reporter flickering and scrolling issues and then look into new APIs.

PR #6968, opened 2024-11-26
fix(watch): don't indicate exit when no matching files

AriPerkkio:
### Description

- Fixes https://github.com/vitest-dev/vitest/issues/6949
- Fixes https://github.com/vitest-dev/vitest/issues/7076

### Tests

### Documentation

### Changesets


packages/vitest/src/node/core.ts
```diff
@@ -426,7 +426,10 @@ export class Vitest {
 
       this.logger.printNoTestFound(filters)
 
-      if (!this.config.watch || !(this.config.changed || this.config.related?.length)) {
+      if (this.config.watch) {
+        this.report('onFinished', [], [])
```
sheremet-va, line 430: I feel like this breaks the lifecycle. There is usually at least `onPathsCollected`/`onSpecsCollected` call and `onCollected`

sheremet-va: I think that would be a big breaking change, the assumption now is that we always call `onInit`. Maybe we shouldn't start timers in `onInit`?

AriPerkkio: Not sure what test reporters do with `onInit` in cases where test run didn't start.  If we started timers in next `onPathsCollected` hook, we would get faster runs times in summary 😄 

Looks like the internal junit reporter opens fs handle in `onInit` and closes it in `onFinish`. That logic sounds good to me. It's quite unexpected that `onFinished` is never called after `onInit`.

AriPerkkio:
Maybe after https://github.com/vitest-dev/vitest/pull/7069 this could be solved

PR #6985, opened 2024-11-29
fix(junit): fix testsuites time to be sum of all testsuite items

saitonakamura:
not the total time of execution

### Description

This addresses 2 problems, one that is general to junit reporter itself, but more debatable, and the other one is merge-reports specific, but is clearly a bug. Let's start with the first one.

1. I haven't found an official junit.xml spec (and I'm not sure if it exists at all), but this document https://github.com/testmoapp/junitxml?tab=readme-ov-file#complete-junit-xml-example suggests that `testsuites.time` is an
> time        Aggregated time of all tests in this file in seconds

which makes sense
But the current implementation uses `new Date().getTime() - this._timeStart.getTime()`, so like, total execution time, including transform, prepare and so on. Which IMO should be changed to aggregated sum

2. When using `--merge-reports --reporter=junit` the `testsuites.time` would be completely wrong (due to it being execution time of reports merging, not the tests times aggregation). This is not a junit-only problem, it also affect text reporters, but I'm not sure what is the correct behaviour for them, so I focused on junit here.

I specifically found it when merging reports, since we calculate mean test time by doing `testsuites.time / (testsuites.total - testsuites.skipped)`, and noticed that mean time suddenly dropped down to nanoseconds (I wish it would, though!)

Xml tests are just best-effort, I'd like to use a proper parser, so let me know if they will suffice or not

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/reporters/junit.ts
+++ b/packages/vitest/src/node/reporters/junit.ts
@@ -335,6 +335,7 @@ export class JUnitReporter implements Reporter {
       (stats, file) => {
         stats.tests += file.tasks.length
         stats.failures += file.stats.failures
+        stats.time += file.result?.duration || 0
         return stats
       },
       {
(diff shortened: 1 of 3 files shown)


sheremet-va (approved):
This seems correct to me

saitonakamura:
Anything else required from my side?

Issue #5895, opened 2024-06-16
Benchmark output can be improved


guoyunhe:
### Describe the bug

If you write multiple benchmark group in a single file, the output isn't very clear to separate different test cases:

```
   ✓ setAlpha (2) 7444ms
     name                        hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · @ctrl/tinycolor  18,716,735.25  0.0000  0.0122  0.0001  0.0001  0.0001  0.0001  0.0001  ±0.02%  9358368
   · fastcolor        19,874,881.13  0.0000  0.0291  0.0001  0.0001  0.0001  0.0001  0.0001  ±0.03%  9937441   fastest
   ✓ toHexString (2) 3935ms
     name                        hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · @ctrl/tinycolor   5,342,122.96  0.0002  0.6116  0.0002  0.0002  0.0002  0.0002  0.0005  ±0.90%  2671062
   · fastcolor        10,865,390.93  0.0001  1.1889  0.0001  0.0001  0.0001  0.0001  0.0002  ±0.50%  5432696   fastest
   ✓ toRgbString (2) 7394ms
     name                        hz     min     max    mean     p75     p99    p995    p999     rme   samples
   · @ctrl/tinycolor  17,450,610.85  0.0000  0.0307  0.0001  0.0001  0.0001  0.0001  0.0001  ±0.03%   8725306
   · fastcolor        20,120,086.75  0.0000  0.7499  0.0000  0.0001  0.0001  0.0001  0.0001  ±0.29%  10060044   fastest
```

If you add some blank lines between cases, it should be more readable:

```
   ✓ setAlpha (2) 7444ms
     name                        hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · @ctrl/tinycolor  18,716,735.25  0.0000  0.0122  0.0001  0.0001  0.0001  0.0001  0.0001  ±0.02%  9358368
   · fastcolor        19,874,881.13  0.0000  0.0291  0.0001  0.0001  0.0001  0.0001  0.0001  ±0.03%  9937441   fastest

   ✓ toHexString (2) 3935ms
     name                        hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · @ctrl/tinycolor   5,342,122.96  0.0002  0.6116  0.0002  0.0002  0.0002  0.0002  0.0005  ±0.90%  2671062
[...]

PR #7019, opened 2024-12-04
fix: rewrite benchmark reporter without `log-update`

AriPerkkio:
### Description

- Fixes https://github.com/vitest-dev/vitest/issues/5895
- ~~Patches `tinyexec` for https://github.com/tinylibs/tinyexec/issues/40.~~ Alternatively I could remove the patch and use `node:child_process` in test cases.
- Removes `log-update` and `cli-truncate` -> -92 kB from bundle
- Uses summary-reporter so users can now scroll up and see previous benchmark results while tests are running.

### Test preview release with:
- [x] https://github.com/radashi-org/radashi/tree/main/benchmarks
  - They have 98 benchmark files. In `vitest@v2` they could not see previous results during the run as `log-update` limits the rows to the terminal height. Now you can actually scroll up during the run and see previous results. No need to wait for that ~8min run to finish.
  - Having the summary reporter on bottom of terminal helps here a lot as you can actually see how many benchmark files are left.
- [x] https://github.com/toss/es-toolkit/tree/main/benchmarks/performance
  - Noticed tests being reported as `'pass'` multiple times, fixed. 
- [x] https://github.com/vuejs/core/blob/main/packages/reactivity/__benchmarks__/ref.bench.ts

### Tests

### Documentation

### Changesets


packages/vitest/src/node/reporters/renderers/utils.ts
```diff
@@ -258,3 +236,11 @@ export function withLabel(color: 'red' | 'green' | 'blue' | 'cyan' | 'yellow', l
...
   return c.dim(`${str.padStart(11)} `)
 }
+
+export function truncateString(text: string, maxLength: number): string {
+  if (stripVTControlCharacters(text).length <= maxLength) {
+    return text
+  }
+
+  return `${sliceAnsi(text, 0, maxLength - 1)}…`
+}
```
AriPerkkio, line 246: `slice-ansi` is only required here. We could remove it if we didn't care about coloring when error messages' lines are truncated. Thoughts @hi-ogawa, @sheremet-va?

image

hi-ogawa: I don't mind either way, but unless it's buggy or slow, I think It's good to have it.

AriPerkkio: Not buggy or slow. I was mostly thinking about reducing install size, but maybe it's not worth it. Though this truncating is only done for these errors so it's not used most of the time. https://pkg-size.dev/slice-ansi

sheremet-va: I also don't mind

packages/vitest/src/node/logger.ts
```diff
@@ -303,6 +307,42 @@ export class Logger {
     this.log(c.red(divider()))
   }
 
+  getColumns() {
+    return 'columns' in this.outputStream ? this.outputStream.columns : 80
+  }
+
+  onExit(listener: Listener) {
```
sheremet-va, line 314: From the outside perspective, I am a bit confused by the `logger` is responsible for `onExit` callback 👀

sheremet-va: I like `onTerminalCleanup` 👍🏻

packages/vitest/src/node/reporters/renderers/utils.ts
AriPerkkio: I'll remove `slice-ansi` from here as its usage is quite an edge case.

AriPerkkio:
Looks like benchmark runner is reporting tests of nested suites twice. Need to resolve this, otherwise summary is showing too high test count.

sheremet-va:
Does this also add a project name to the output? I just noticed we don't print `|name|`

AriPerkkio:
Project name is visible now:

AriPerkkio:
It's not in table output though, should it be here too?

sheremet-va:
> It's not in table output though, should it be here too?
> 
> 

Yeah

packages/vitest/src/runtime/runners/benchmark.ts
```diff
@@ -135,16 +137,6 @@ async function runBenchmarkSuite(suite: Suite, runner: NodeBenchmarkRunner) {
     suite.result!.duration = performance.now() - start
     suite.result!.state = 'pass'
 
-    tasks
```
sheremet-va, line 138: Did you move the sorting? Why is it removed from here?

AriPerkkio: It's moved to main thread's side, so that we don't need to report tasks twice as `'pass'`.

- Created failing test case first: https://github.com/vitest-dev/vitest/pull/7019/commits/d4223fd55f46c9ef34d0e82326da8e523a110d3b
- Then moved sorting to the other side of RPC: https://github.com/vitest-dev/vitest/pull/7019/commits/9ed93b56752033409c4146c5e1a69c0d94281830

There's `test/benchmark/test/reporter.test.ts` that fails if this sorting and `result.benchmark.rank` setting is wrong.

hi-ogawa (approved):
Looks good! Tested radashi locally and the improvement is awesome :clap: 

About `tinyexec` patch, it's not a blocker, but the situation is not ideal since patch is only applied for our local dev, but `tinyexec` is also used for external dependencies in a few places.

AriPerkkio:
> About `tinyexec` patch, it's not a blocker, but the situation is not ideal since patch is only applied for our local dev, but `tinyexec` is also used for external dependencies in a few places.

I agree. I'll replace `tinyexec` usage+patching with `node:child_process` or maybe even `startVitest` when using `preserveAnsi` option from https://github.com/vitest-dev/vitest/pull/7090/files#diff-eee947831abd0685dc52e26f42f2c7c3adfb802e478ca8e252ba649f1a1ab2f6R5-R13.

hi-ogawa (approved):
Thanks!

Issue #6956, opened 2024-11-23
Simplify `coverage.exclude` patterns, remove `coverage.all`


AriPerkkio:
### Clear and concise description of the problem

Setting good defaults for `coverage.exclude` is difficult. The `coverage.include` is even more difficult. By default `coverage.include` is `['*']` as we don't know where users' source code are. 

Currently we have listed complex `coverage.exclude` pattern that includes all kinds of framework and tool specific patterns that aren't really Vitest related. We constantly get new requests to add there more tool specific files, like MSW, postcss, tailwind, etc. This is not maintainable in long run.

What (most) users don't realize is that they shouldn't be using `coverage.exclude` in the first place. Instead of defining every single `*.config.*` file of your project's root, you should be defining `coverage.include`.

```
├── vitest.config.js
├── tailwind.config.js
├── postcss.config.js
|
├── public
|  └── mockServiceWorker.js
|
├── src
|  ├── math.js
|  └── utils
|     └── geometry.js
└── test
   └── math.test.js
```

Here users shouldn't do `coverage.exclude: ['tailwind.config.js', 'postcss.config.js, 'public/mockServiceWorker.js']`. They should do just `coverage.include: ['src']`. 

### Suggested solution

https://github.com/vitest-dev/vitest/issues/6956#issuecomment-2519917038

We could do similar option as Jest has: https://jestjs.io/docs/configuration#collectcoveragefrom-array

Their option forces users to define `coverage.include`, as by default they have `coverage.all: false`. By default Jest shows only files that were imported by test files. Users have to define uncovered files in `collectCoverageFrom`.

💡 We could remove `coverage.all` and by default only show files that were imported by tests. Then if `coverage.includes` is defined, show all covered+uncovered files that match those. 

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

AriPerkkio:
> ### Alternative
> Alternatively we could do similar option as Jest has: https://jestjs.io/docs/configuration#collectcoveragefrom-array
> 
[...]

PR #7020, opened 2024-12-04
add storybook coverage exclude entries

ndelangen:
### Description

Storybook stories can be considered test, especially with the storybook test plugin for vitest in the works, they are converted into test files.

When enabling coverage in that scenario, it's undesired to have the story files count for measuring coverage.

Therefore is useful for those files to be excluded.

It would be possible for vitest users to configure this, however we think it could be useful to add this to the default set of coverage exclude patterns inside of vitest itself.

More context here:
https://github.com/storybookjs/storybook/issues/29613

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/defaults.ts
+++ b/packages/vitest/src/defaults.ts
@@ -39,7 +39,7 @@ const defaultCoverageExcludes = [
   '**/\x00*',
   'cypress/**',
   'storybook-static/**',
-  '*.stories.*,
+  '*.stories.*',
   'test?(s)/**',
   'test?(-*).?(c|m)[jt]s?(x)',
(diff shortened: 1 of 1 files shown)


AriPerkkio:
We are considering removing all non-Vitest related defaults from `coverage.exclude`, see:
- https://github.com/vitest-dev/vitest/issues/6956

AriPerkkio (commented):
Hey @ndelangen, thanks for the PR.

Currently we won't be accepting any new values to `coverage.excludes` as we are planning to remove most of the current ones in future releases. 
- https://github.com/vitest-dev/vitest/issues/6956

PR #7044, opened 2024-12-07
fix(ui): add errors and draft state (*) to the code editor

userquin:
### Description

We have a regression somewhere, maybe updating Vue dependency, I'm not sure. The problem is that changes included in these 2 PR missing in current version:
- https://github.com/vitest-dev/vitest/pull/1131
- https://github.com/vitest-dev/vitest/pull/1039

This PR adds the draft indicator and the line error to the code editor: tested with test/core and test/browser (only preview).

This PR also includes:
- codemirror optimizations when registering/unregistering the errors entries to remove memory leaks: looks like `destroy` method is missing in v6
- changes the cursor to `wait` when saving the file: [Video showing the cursor](https://streamable.com/27q83g)
- registers the span listener to open the file in the editor (at line and column error), missing in the original PR #1131

imagen

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/ui/client/components/views/ViewEditor.vue
+++ b/packages/ui/client/components/views/ViewEditor.vue
@@ -18,13 +18,17 @@ const code = ref('')
 const serverCode = shallowRef<string | undefined>(undefined)
 const draft = ref(false)
 const loading = ref(true)
-// finished.value is true when saving the file, we need it to restore the caret position
-const saving = ref(true)
+const saving = ref(false)
 const currentPosition = ref<CodeMirror.Position | undefined>()
 
 watch(
   () => props.file,
   async () => {
+    // this watcher will be called multiple times when saving the file in the view editor
+    // since we are saving the file and changing the content inside onSave we just return here
+    if (saving.value) {
+      return
+    }
     loading.value = true
     try {
       if (!props.file || !props.file?.filepath) {
@@ -46,21 +50,27 @@ watch(
 
     // fire focusing editor after loading
     loading.value = false
-    saving.value = false
   },
   { immediate: true },
 )
 
-watch(() => [loading.value, props.file, lineNumber.value] as const, ([loadingFile, _, l]) => {
-  if (!loadingFile) {
+watch(() => [loading.value, saving.value, props.file, lineNumber.value] as const, ([loadingFile, s, _, l]) => {
+  if (!loadingFile && !s) {
     if (l != null) {
       nextTick(() => {
-        const line = { line: l ?? 0, ch: 0 }
-        codemirrorRef.value?.scrollIntoView(line, 100)
-        nextTick(() => {
-          codemirrorRef.value?.focus()
-          codemirrorRef.value?.setCursor(line)
-        })
+        const cp = currentPosition.value
+        const line = cp ?? { line: l ?? 0, ch: 0 }
+        // restore caret position: the watchDebounced below will use old value
+        if (cp) {
+          currentPosition.value = undefined
+        }
+        else {
+          codemirrorRef.value?.scrollIntoView(line, 100)
+          nextTick(() => {
+            codemirrorRef.value?.focus()
+            codemirrorRef.value?.setCursor(line)
+          })
+        }
       })
     }
     else {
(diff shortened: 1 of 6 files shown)


packages/ui/client/components/views/ViewEditor.vue
```diff
@@ -134,54 +139,111 @@ function createErrorElement(e: ErrorWithDiff) {
...
+    if (!hasBeenEdited.value) {
+      cmValue.clearHistory()
+    }
+
+    cmValue.on('changes', codemirrorChanges)
+
+    // restore caret position
+    const { ch, line } = currentPosition.value ?? {}
+    console.log('WTF: ', currentPosition.value)
+    console.error('WTF', new Error('WTF'))
```
userquin, line 188: remove these 2 lines

packages/ui/client/components/views/ViewEditor.vue
```diff
@@ -134,54 +139,108 @@ function createErrorElement(e: ErrorWithDiff) {
...
+    // the server will send 3 events in a row
+    // await first change in the state
+    await until(finished).toBe(false, { flush: 'sync', timeout: 1000, throwOnTimeout: false })
+    // await second change in the state
+    await until(finished).toBe(true, { flush: 'sync', timeout: 1000, throwOnTimeout: false })
+    // await last change in the state
+    await until(finished).toBe(false, { flush: 'sync', timeout: 1000, throwOnTimeout: false })
+  }
+  catch (e) {
+    console.error('error saving file', e)
```
userquin, line 224: we can iterate this later to add some popup to show the error

packages/ui/client/components/views/ViewEditor.vue
```diff
@@ -134,54 +139,108 @@ function createErrorElement(e: ErrorWithDiff) {
...
+    serverCode.value = content
+    // update draft indicator in the tab title (</> * Code)
+    draft.value = false
+    // the server will send 3 events in a row
+    // await first change in the state
+    await until(finished).toBe(false, { flush: 'sync', timeout: 1000, throwOnTimeout: false })
+    // await second change in the state
+    await until(finished).toBe(true, { flush: 'sync', timeout: 1000, throwOnTimeout: false })
+    // await last change in the state
+    await until(finished).toBe(false, { flush: 'sync', timeout: 1000, throwOnTimeout: false })
```
userquin, line 221: maybe we should throw error, I'll check it tmr

packages/ui/client/components/views/ViewEditor.vue
```diff
@@ -134,54 +139,108 @@ function createErrorElement(e: ErrorWithDiff) {
...
   listeners.push([span, el, () => destroyTooltip(span)])
   handles.push(codemirrorRef.value!.addLineClass(stack.line - 1, 'wrap', 'bg-red-500/10'))
   widgets.push(codemirrorRef.value!.addLineWidget(stack.line - 1, div))
 }
 
-watch(
-  [cm, failed],
-  ([cmValue]) => {
+const { pause, resume } = watch(
+  [codemirrorRef, failed, finished, saving] as const,
```
userquin, line 150: add hint here: the watcher receiving `saving` with true even when not changed, that's why we use pause and resume it in `onSave`.

The problem without pausing this watcher when saving the file => cursor position lost (the caret will go to the first line in the test)

userquin: The watcher loading the file content being called multiple times when the server re-run the file, splitting he logic for loading and saving will prevent reloading the content while saving from the view editor.

The cursor will be preserved only when saving in the file in the ui view editor, saving the file in the IDE will lost the position (will use the line in the url)

userquin: Note: we should lock the editor or check if there is something changed when saving, but we cannot set the editor in readonly (maybe refreshing  the editor works), rn you can save at any time using the keyboard.

I Will check it again later since latest commit will prevent reloading the race condition in loading watcher.

userquin: looks like using `setOption('readOnly', true/false)` + `refresh()` we can prevent saving the file while saving it, updated logic and added protection to prevent firing `save` event when editor in readOnly mode (also for static html reporter)

sheremet-va (approved):

PR #7048, opened 2024-12-08
feat(cli): Support location filters for suites

### LINKED ISSUE #5445: Support specifying a line number when filtering tests

mzhubail:
### Description

Closes #5445.

This is a continuation on my previous PR #6411, with support for matching location of suite.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/runner/src/utils/collect.ts
+++ b/packages/runner/src/utils/collect.ts
@@ -51,10 +51,8 @@ export function interpretTaskModes(
         else if (parentMatchedWithLocation) {
           t.mode = 'run'
         }
-        else {
-          if (t.type === 'test') {
-            t.mode = 'skip'
-          }
+        else if (t.type === 'test') {
+          t.mode = 'skip'
         }
       }
 
--- a/packages/runner/src/utils/collect.ts
+++ b/packages/runner/src/utils/collect.ts
@@ -15,7 +15,7 @@ export function interpretTaskModes(
 ): void {
   const matchedLocations: number[] = []
 
-  const traverseSuite = (suite: Suite, parentIsOnly?: boolean) => {
+  const traverseSuite = (suite: Suite, parentIsOnly?: boolean, parentMatchedWithLocation?: boolean) => {
     const suiteIsOnly = parentIsOnly || suite.mode === 'only'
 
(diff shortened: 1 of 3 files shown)


mzhubail:
- The jest of it is that we have a stricter matching mode, when a test location is passed:

  ```ts
  if (testLocations !== undefined && testLocations.length !== 0) {
  ```

  We have three cases to deal with:

  1. Parent did match via location filter, and as such this test is marked to run. No Questions asked.

  2. This test / suite does match via location filter, and as such mark it to run, and do the same for its children.

  3. No match. In this case I'd mark as skipped only it is a test. (if it's a suite, I still want to go through its children)

- https://github.com/vitest-dev/vitest/blob/a1b7384916db2ab224fe87e78964964067107117/packages/runner/src/utils/collect.ts#L46-L51

  Note that I set the case of test matching before `parentMatchedWithLocation` to avoid printing wrong output on an edge case.  See https://github.com/vitest-dev/vitest/blob/a1b7384916db2ab224fe87e78964964067107117/test/cli/test/location-filters.test.ts#L52  If we had the condition `parentMatchedWithLocation` before, this test cases would report test not found.

sheremet-va (approved):

sheremet-va:
Thanks!

PR #7061, opened 2024-12-10
feat(coverage): `thresholds` to support maximum uncovered items

### LINKED ISSUE #7056: Absolute coverage thresholds

As a developer using Vitest, I want a way to specify absolute coverage thresholds instead of percentage coverage thresholds. The benefit of absolute thresholds is that you can freeze the code coverage at a specific number of uncovered lines, which then ensures that test coverage must be considered for any new additions to the code.

For example, let's say I start working on a package that has 1000 lines, 750 of which are covered (75% coverage), so I set the coverage threshold to 75% to "preserve" this amount of coverage. I then add a new feature with 10 lines, 8 of which are covered. The new coverage percentage is 75.05%, so I pass coverage. However, if only 7 lines of the new feature are covered, then the new coverage percentage is 74.95%, and I fail. This feels a bit arbitrary - I've only enforced that new code additions must have as good of coverage as the existing code.

With absolute coverage thresholds, I would specify my coverage threshold as "250 uncovered lines." When adding a new feature with 10 lines, all 10 lines must be covered (or marked to be ignored in coverage).

In my organization, our approach to code coverage is to strive for 100% coverage unless there's a reasonable tradeoff (e.g. a particular is too hard to reproduce in testing to make it worthwhile, in which case we explicitly mark that case to be ignored). This is easy to do for code that already has 100% coverage. But for any legacy code that doesn't, we can't really implement this policy very well without absolute coverage thresholds.

We are currently migrating from jest (which supports this feature) to vitest. We are loving vitest but are sad to lose this important piece of functionality.

### Suggested solution

Jest supports [configuring a coverage threshold](https://jestjs.io/docs/configuration#coveragethreshold-object) with a positive number, in which case its interpreted as a percentage, or a negative number, in which case its interpreted as an absolute threshold.

The advantage of this approach is that it doesn't add any new configuration options. The disadvantage is that it's not totally clear what's happening unless you read the docs.

I think it would also make sense to have a more explicit configuration option to use absolute or percentage thresholds in the coverage config. E.g. something like `coverage.thresholdType: 'percentage' | 'absolute'`.

I haven't yet looked into how to implement this in the code base, but I'd be happy to take the lead on researching and implementing the feature once its agreed that it would be a useful addition.

### Alternative

_No response_

### Additional context

Thank you for your time and consideration!

### Validations

jonahkagan:
### Description

Closes #7056 

Match [Jest's API for coverage thresholds](https://jestjs.io/docs/configuration#coveragethreshold-object), treating a positive number as the minimum covered percentage (`lines: 90` means 90% of lines must be covered) and a negative number as the maximum uncovered count (`lines: -10` means no more than 10 lines may be uncovered).

When using negative coverage thresholds with the `autoUpdate` flag, set the new thresholds based on the reported number of uncovered entities. If there are no uncovered entities, set the threshold to 100%.

### Tests

### Documentation

### Changesets


jonahkagan:
I see one test failure in CI that appears to be an unrelated test: `test/timeout.spec.ts`. I don't think I have permission to re-run the job to see if this test might be flaky.

AriPerkkio (changes requested):
Looks good! Glad to see that `thresholds.autoUpdate` didn't require that much complex logic. Also test cases seem to cover all cases - nice! 💯 

Left some minor improvement ideas.

test/coverage-test/test/threshold-absolute-failure.test.ts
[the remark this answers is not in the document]
jonahkagan: Merged!

test/coverage-test/test/threshold-failure.test.ts
```diff
@@ -28,6 +28,35 @@ test('failing thresholds', async () => {
...
+  if (isV8Provider()) {
+    expect(stderr).toContain(`ERROR: Uncovered lines (6) exceed "**/fixtures/src/math.ts" threshold (5)`)
+    expect(stderr).toContain(`ERROR: Uncovered functions (3) exceed "**/fixtures/src/math.ts" threshold (2)`)
+    expect(stderr).toContain('ERROR: Uncovered statements (6) exceed "**/fixtures/src/math.ts" threshold (1)')
+  }
+  else {
+    expect(stderr).toContain(`ERROR: Uncovered functions (3) exceed "**/fixtures/src/math.ts" threshold (2)`)
+    expect(stderr).toContain('ERROR: Uncovered statements (3) exceed "**/fixtures/src/math.ts" threshold (1)')
+  }
+})
```
AriPerkkio, line 58:
```suggestion
  expect(exitCode).toBe(1)

  if (isV8Provider()) {
    expect(stderr).toContain('ERROR: Uncovered lines (6) exceed "**/fixtures/src/math.ts" threshold (5)')
    expect(stderr).toContain('ERROR: Uncovered functions (3) exceed "**/fixtures/src/math.ts" threshold (2)')
    expect(stderr).toContain('ERROR: Uncovered statements (6) exceed "**/fixtures/src/math.ts" threshold (1)')
  }
  else {
    expect(stderr).toContain('ERROR: Uncovered functions (3) exceed "**/fixtures/src/math.ts" threshold (2)')
    expect(stderr).toContain('ERROR: Uncovered statements (3) exceed "**/fixtures/src/math.ts" threshold (1)')
  }
})
```

AriPerkkio: One last small refactor and it's good to go.

Looks like I cannot apply this change myself, the `Maintainers are allowed to edit this pull request.` is not enabled on this PR. Probably due to origin of the PR being from `votingworks` organization. 🤔

jonahkagan: Yeah, I don't see any way to enable maintainers to edit in the UI or in these docs: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

AriPerkkio (approved):
Looks good, thanks!

PR #7069, opened 2024-12-11
feat: introduce the new reporter API

sheremet-va:
### Description

Initial draft (decided to not implement `onError` for now):
<details>

```ts
interface Reporter {
  /**
   * This method is called when the error happens outside of the test.
   *
   * Possible errors:
   * - unhandled error thrown during the test run
   * - error thrown during collection phase when importing a module (global throw)
   * - error thrown during collection phase when calling `describe` callback
   * - error thrown inside `beforeAll`/`afterAll` hooks since it can't be attributed to a specific test
   */
  onError(error: SerialisedError, suite?: TestSuite): void

  /**
   * This method is called when all tests inside a single file were collected.
   * The `file` object contains `children()` method with all suites and tests.
   * Tasks won't have `result` property yet.
   *
   * **Note:** This method can be called in parallel if multiple files are collected at the same time (`--single-worker` is used)
   */
  onFileCollected(file: TestModule): void

  /**
   * This method is called when the runner is ready to start running a test. The `result` will always be `undefined`.
   * **Note:** If test is marked as `todo` or `skip`, this method will still be called.
   */
  onTestPrepare(test: TestCase): void
  /**
   * This method is called when the test finished running. It will always have a `result` property.
   */
  onTestFinished(test: TestCase): void
  /**
   * This method is called only after the test has finished running and the `result.type` is `failed`.
   */
  onTestFailed(test: TestCase): void

  /**
   * This method is called when user logs something to the console.
   * The order of logs is not guaranteed.
   */
  onConsoleLog(type: 'stderr' | 'stdout', log: UserConsoleLog): void

  /**
   * This is called when all tests have finished running.
   */
  onTestRunFinished(files: TestModule[], errors: SerialisedError[], reason: 'passed' | 'failed' | 'timedout' | 'interrupted'): void

  /**
   * Coverage is performed for all files that were executed.
   * This is called only if `--coverage` flag is used.
   */
  onCoverage(summary: any): void
}
```
</details>

The current lifecycle ("run tests" has its own lifecycle):

image

image

tenor

### Tests

### Documentation

### Changesets


packages/vitest/src/node/test-run.ts
```diff
@@ -0,0 +1,53 @@
+import type { TaskResultPack } from '@vitest/runner'
+import type { ArgumentsType } from '../types/general'
+import type { Vitest } from './core'
+import type { Reporter } from './reporters'
+import type { TestModule } from './reporters/reported-tasks'
+import type { TestSpecification } from './spec'
+
+export class TestRun {
```
AriPerkkio, line 8: I can start helping out with this. Did you already have an idea how this should be integrated into the `core.ts` and RPC calls?

sheremet-va: I wanted to have parsing logic here, and use higher level methods in `core` (`this.testRun.start(specs)`) and low lever in rpc calls (like `onTaskUpdate`)

This is also the place where we call reporter methods related to the test run.

The inspiration for method names is https://vshaxe.github.io/vscode-extern/vscode/TestRun.html

sheremet-va: I also think that proposed events are not enough, we will probably need to add more

sheremet-va: I also added `statistics` to the reported tasks, so we can modify them in place

docs/advanced/guide/lifecycle.md
```diff
@@ -0,0 +1,3 @@
+# Test Lifecycle
+
+<!-- TODO: lifecyle diagram and reporter API -->
```
AriPerkkio, line 3: Could we use `vitepress-plugin-mermaid` for this? 🤔 

https://emersonbottero.github.io/vitepress-plugin-mermaid/guide/styles.html

sheremet-va:
> * `onHookStart(...)` - `before*` hooks is called

if we report suite hooks (`beforeAll`/`beforeEach`), should report `onSuiteStart`?

> * `onTestFinish` / `onTestFail`

I think it's better to have a single hook and check the result there - I updated the `TestCase` to always return a `result()` (now return `pending` if no result is set)

> * `onTestFile*`

Should be `onTestModule*`

docs/advanced/guide/lifecycle.md
sheremet-va: Maybe when the lifecycle is finalised

sheremet-va:
@AriPerkkio should we update the `onTaskUpdate` payload to include the type of action to make it easier to parse? (`test-started`, `file-finished`, etc)

AriPerkkio:
> if we report suite hooks (`beforeAll`/`beforeEach`), should we report `onSuiteStart`?

Do we need that anywhere? I don't think `onTaskUpdate` reports this atm. Currently slowly running `beforeAll` etc. hooks are showed by reporters so we need to support those.

> Should be `onTestModule*`. The idea with the name is that I expect us to implement support for virtual modules.

Yup, let's use `onTestModule*` and `onTestCase*` naming convention in the hooks. It matches the TS interfaces too.

> should we update the `onTaskUpdate` payload to include the type of action to make it easier to parse? (`test-started`, `file-finished`, etc)

I'm not yet sure. The `entity.type` and `entity.state()` should be enough. :thinking:

sheremet-va:
> I'm not yet sure. The `entity.type` and `entity.state()` should be enough. 🤔

entity is a user facing API. I am asking if we should send the type to ourselves when we trigger `onTaskUpdate` in the test runtime to parse it more easily on the server. Instead of sending `updateTask(task)`, we do something like `updateTask('test-start', task)`

sheremet-va:
> Do we need that anywhere? I don't think `onTaskUpdate` reports this atm. Currently slowly running `beforeAll` etc. hooks are showed by reporters so we need to support those.

If it doesn't report it, we can start reporting it. I just don't understand why some hooks are reported and some not - I see it from the practical standpoint, but philosophically - why? What makes them more special?

AriPerkkio:
`beforeAll/afterAll` are showed by reporters, let's keep supporting them like before. I feel like I'm missing something here. 🤔

- https://github.com/vitest-dev/vitest/pull/6893#discussion_r1839800573. 

https://github.com/vitest-dev/vitest/blob/8764f5c15760c596f408c30ae1849b3284555c81/packages/vitest/src/node/reporters/task-parser.ts#L66-L75

sheremet-va:
> `beforeAll/afterAll` are showed by reporters, let's keep supporting them like before. I feel like I'm missing something here. 🤔
> 
> * [feat(reporters): `summary` option for `verbose` and `default` reporters #6893 (comment)](https://github.com/vitest-dev/vitest/pull/6893#discussion_r1839800573).
> 
> https://github.com/vitest-dev/vitest/blob/8764f5c15760c596f408c30ae1849b3284555c81/packages/vitest/src/node/reporters/task-parser.ts#L66-L75

`beforeEach` hook is tied to a test case, it's obvious what reporter hooks will be called and in what order. I feel like with `beforeAll` it's not so obvious because there is no corresponding `onTestSuite*` hook

packages/vitest/src/node/test-run.ts
```diff
@@ -118,7 +123,8 @@ export class TestRun {
     await Promise.all(finishedTestCases.map(testCase => this.vitest.report('onTestCaseFinished', testCase)))
     await Promise.all(finishedTestModules.map(module => this.vitest.report('onTestModuleFinished', module)))
 
-    await Promise.all(queuedTestModules.map(module => this.vitest.report('onTestModuleQueued', module)))
+    // TODO: Not needed as RPC has onQueued already. Do we want to remove that and centralize resolving here instead?
```
sheremet-va, line 126: queued state cannot be reported in `onTaskUpdate` because the task is not created yet, so we always have to use `onQueued`

packages/vitest/src/node/core.ts
```diff
@@ -94,6 +95,7 @@ export class Vitest {
   /** @internal */ reporters: Reporter[] = undefined!
   /** @internal */ vitenode: ViteNodeServer = undefined!
   /** @internal */ runner: ViteNodeRunner = undefined!
+  /** @internal */ testRun: TestRun = undefined!
```
sheremet-va, line 98: We should use the `_*` convention for public internal fields, the old ones are kept as is for backwards compatibility

packages/vitest/src/node/test-run.ts
AriPerkkio: Makes sense, I'll update this part. 👍

sheremet-va:
I'd like to remove the public `state` in the future (we can keep errors and stuff there) and replace it with a `testRun`. Maybe we should make the `testRun` public, but keep the methods internal?

packages/vitest/src/node/test-run.ts
```diff
@@ -0,0 +1,161 @@
...
+        if (event === 'test-prepare' && entity.type === 'test') {
+          runningTestCases.push(entity)
+        }
+
+        if (event === 'test-finished' && entity.type === 'test') {
+          finishedTestCases.push(entity)
+        }
+
+        if ((event === 'suite-hook-start' || event === 'suite-hook-end') && entity.task.result?.hooks) {
+          for (const hook of Object.keys(entity.task.result.hooks)) {
```
sheremet-va, line 91: Wouldn't it report all hooks that it has? For example, after `afterAll`, the task can have `beforeAll`, `beforeEach`, `afterEach` and `afterAll` in this object

packages/vitest/src/node/test-run.ts
```diff
@@ -0,0 +1,161 @@
...
+        if (event === 'suite-prepare' && entity.type === 'module') {
+          runningTestModules.push(entity)
+        }
+
+        if (event === 'suite-finished' && entity.type === 'module') {
+          finishedTestModules.push(entity)
+
+          // Skipped tests need to be reported manually once test module has finished
+          for (const test of entity.children.allTests()) {
+            if (test.result().state === 'skipped') {
```
sheremet-va, line 76: do we need to report skipped tests? they won't have `start` and calling them arbitrary at the end seems weird to me

packages/vitest/src/node/test-run.ts
AriPerkkio: I haven't been able to get this part working fine yet. Feel free to change it, the test case in `test-run.test.ts` should print tasks correctly once fixed.

packages/vitest/src/node/test-run.ts
sheremet-va: We can report them when any `suite-finished` is emitted so it's at least timed with other tests, although the order is still wrong, - and they won't have `start` event. Also if suite skips all tests, `suite-finished` is never fired 🤔 

I don't know what to do here, honestly (a new hook?)

AriPerkkio: I've now implemented this so that skipped test cases don't have any special rules. They are reported exactly as other test cases too. This will be easier for reporters to implement.

If this is not something we want, it can still be changed. 

https://github.com/vitest-dev/vitest/blob/41985f3d90f5d1de10233122f794ae2a1fbf8354/test/reporters/tests/test-run.test.ts#L24-L62

userquin: IIRC the ui having some problems with todo tests since they are not reported in current api version, we need to change mode to `todo` on received end signal:

https://github.com/vitest-dev/vitest/blob/main/packages/ui/client/composables/explorer/collector.ts#L109

Maybe a bug in the ui where we set to running all tests when we click any run button.

sheremet-va: how can test trigger an event called "start" and "end" when it neither started nor ended?

sheremet-va: I green lit the names "start" and "end" when skipped tests weren't reported. the "start" event was called "prepare" because I assumed that skipped tests are reported and it doesn't make sense to call it start

We need to find a better name if we report them like that

sheremet-va: Would be nice to also have a test for the reverse:

```ts
       test.skip('skipped', () => {}); 
       test('running', () => {}); 
```

userquin: > how can test trigger events called "start" and "end" when it neither started nor ended?

I guess we should re-check the idle logic in the client module, since we set the state to running when we receive on
TaskUpdated rpc callback.

We have onFinished, shouldn't be the end event?

AriPerkkio: > how can test trigger events called "start" and "end" when it neither started nor ended?

In `test("skipped test", ctx => { ctx.skip(); })` for example. Though would it have "end" then?

sheremet-va: > Though would it have "end" then?

To clarify, `test("skipped test", ctx => { ctx.skip(); })` always has a start event, so not having an "end" wouldn't make sense

`test.skip()` doesn't have a start event, we just skip it

So, I think we should either:
- rename methods
- add different methods
- don't report skipped tests

AriPerkkio: Any of those options sound good to me.

When building the summary reporter, I had difficulties showing correct test count when tests were skipped. The reporter keeps increasing test count when a test is reported as finished. But as skipped tests are not, I had to manually extract the count of skipped tests when test module finished. So the active test counter logic needs to be in multiple places.

sheremet-va: Maybe we need to have `queued/`dequeued` for tests like [Node has](https://nodejs.org/api/test.html#custom-reporters)? They will be called for every test, but `start`/`end` will only be called for unskipped tests

sheremet-va: (I think they need queued/dequeued because they run as the collect(?))

sheremet-va: Interestingly, `'test:start'` is emitted for tests _and_ subtests which is basically a suite?

sheremet-va:
The current proposal looks like this:

image

Events with dashed border are optional

packages/vitest/src/node/reporters/dot.ts
```diff
@@ -1,115 +1,91 @@
...
-    this.renderer.finish()
+    this.renderer?.finish()
+
+    super.onFinished(files, errors)
   }
 
-  onTestFilePrepare(file: File): void {
-    for (const test of getTests(file)) {
+  onTestModuleCollected(module: TestModule): void {
+    for (const test of module.children.tests()) {
```
sheremet-va, line 56: shouldn't it be `allTests()`?

packages/vitest/src/node/core.ts
```diff
@@ -715,13 +729,12 @@ export class Vitest {
         }
       }
       finally {
-        // can be duplicate files if different projects are using the same file
-        const files = Array.from(new Set(specs.map(spec => spec.moduleId)))
-        const errors = this.state.getUnhandledErrors()
+        // TODO: wait for coverage only if `onFinished` is defined
```
AriPerkkio, line 732: We do need it for `this.reportCoverage` below

docs/advanced/api/reporters.md
```diff
@@ -0,0 +1,306 @@
...
+- `afterAll`
+- `beforeEach`
+- `afterEach`
+
+If `beforeAll` or `afterAll` are started, the `entity` will be either [`TestSuite`](/advanced/api/test-suite) or [`TestModule`](/advanced/api/test-module).
+
+If `beforeEach` or `afterEach` are started, the `entity` will always be [`TestCase`](/advanced/api/test-case).
+
+::: warning
+`onHookStart` method will not be called if the hook did not run during the test run.
```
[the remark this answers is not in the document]
AriPerkkio: Or is it here due to `collect` not running hooks?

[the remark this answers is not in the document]
sheremet-va: As I found out after years of working on Vitest, some people think differently than me. I can imagine someone thinking that this hook is called with empty duration (start/end in the same tick) just to keep the lifecycle

packages/vitest/src/node/core.ts
sheremet-va: `this.reportCoverage` happens after `onFinished`, so I don't see anything that would stop this being a reality

I would go even further and move the printing logic to reporters

docs/advanced/api/reporters.md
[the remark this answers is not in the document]
AriPerkkio: I would imagine this being the same as suite hooks. Only called when it exists 🤔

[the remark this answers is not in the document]
sheremet-va: And we specify this in suites documentation 😄

packages/vitest/src/node/core.ts
sheremet-va: I do wonder of course if it is a good idea to separate test run end from the coverage reporting. Playwright doesn't report coverage at all, for example

packages/vitest/src/node/core.ts
```diff
@@ -447,51 +450,61 @@ export class Vitest {
...
         if (task.logs) {
           logs.push(...task.logs)
         }
-        taskPacks.push([task.id, task.result, task.meta])
+        if (task.type === 'test') {
+          taskPacks.push([task.id, task.result, task.meta])
+          if (task.mode !== 'skip' && task.mode !== 'todo') {
+            events.push([task.id, 'test-prepare'], [task.id, 'test-finished'])
+          }
+        }
```
AriPerkkio, line 484: Is ignoring skipped tests here intentional?

sheremet-va: Of course. They are ignored by the runner, we report them at the end of suite: https://github.com/vitest-dev/vitest/pull/7069/files#diff-00f8b4e68ee1b24f202a98a83759158b273f298a31033bad9ac50e8ea887055fR97

sheremet-va: The suite skip on the next line seems unintentional though

packages/vitest/src/node/test-run.ts
```diff
@@ -0,0 +1,168 @@
...
+      // by this point, the run will be marked as failed if there are any errors,
+      // should it be done by testRun.end?
+      : process.exitCode
+        ? 'failed'
+        : 'passed'
+
+    try {
+      await Promise.all([
+        this.vitest.report('onTestRunEnd', modules, [...errors] as SerializedError[], state),
+        // TODO: in a perfect world, the coverage should be done in parallel to `onFinished`
```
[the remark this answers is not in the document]
sheremet-va: This is the same comment as https://github.com/vitest-dev/vitest/pull/7069#discussion_r1912494255

AriPerkkio (dismissed):
Looks good to me

AriPerkkio:
frodo-its-done

PR #7101, opened 2024-12-18
fix: process sourcemaps for stack traces from `globalSetup` files

aleclarson:
### Description

When a `globalSetup` file (or a module imported as a result of a `globalSetup` file's execution) has an error at runtime, the sourcemaps were not being used, because globalSetup files are executed with `node:vm` (which doesn't read sourcemaps).

In my manual testing, calling `installSourcemapsSupport` before loading the globalSetup files will fix the issue.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/project.ts
+++ b/packages/vitest/src/node/project.ts
@@ -26,6 +26,7 @@ import mm from 'micromatch'
 import { isAbsolute, join, relative } from 'pathe'
 import { ViteNodeRunner } from 'vite-node/client'
 import { ViteNodeServer } from 'vite-node/server'
+import { installSourcemapsSupport } from 'vite-node/source-map'
 import { setup } from '../api/setup'
 import { isBrowserEnabled, resolveConfig } from './config/resolveConfig'
 import { serializeConfig } from './config/serializeConfig'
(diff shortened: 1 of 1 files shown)


hi-ogawa (commented):
Thanks for the PR :wave:
I had a few concerns about this, but while thinking about it, they might be actually a non issue. Let me write this down.

The first one is what happens when multiple projects have `globalSetup` and do `installSourcemapsSupport`. This appears to be fine because Vite-node supports registering multiple handlers properly for each `installSourcemapsSupport` call. I was just testing this with pkg.pr.new in my repro and error stack is working after this PR https://github.com/hi-ogawa/reproductions/tree/main/vitest-global-setup-error-stack

The 2nd one is mostly about our internal tests. We often runs vitest inside vitest via `runVitest` of `test-utils` and I'm not sure whether `installSourcemapsSupport` referenced from inside and outside are the same thing. I'll check this later.

For the tests, there is a test asserting global setup error stack, which just happens to pass since the code is plain js: https://github.com/vitest-dev/vitest/blob/e9a61c3ae6eebe7c0018fd9b1bcee6037768f04f/test/config/test/unhandled-rejections.test.ts#L15
Can you make `setup-unhandled-rejections.ts` more complicated (something like https://github.com/vitest-dev/vitest/pull/7096) and see if error position maps correctly?

PR #7103, opened 2024-12-19
feat: make json reporter easier to extend

barraponto:
### Description

Separates concerns : generating results and writing results should be separate functions.
This way we can extend the logtasks function to change the coverage map results without touching the write logic.
Or the other way, change the writing logic without touching the coverage map logic.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/reporters/json.ts
+++ b/packages/vitest/src/node/reporters/json.ts
@@ -88,7 +88,7 @@ export class JsonReporter implements Reporter {
     this.start = Date.now()
   }
 
-  protected async logTasks(files: File[], coverageMap?: CoverageMap | null): JsonTestResults {
+  protected async logTasks(files: File[], coverageMap?: CoverageMap | null): Promise<JsonTestResults> {
     const suites = getSuites(files)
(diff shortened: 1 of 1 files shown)


packages/vitest/src/node/reporters/json.ts
```diff
@@ -88,7 +88,7 @@ export class JsonReporter implements Reporter {
     this.start = Date.now()
   }
 
-  protected async logTasks(files: File[], coverageMap?: CoverageMap | null) {
+  protected async logTasks(files: File[], coverageMap?: CoverageMap | null): Promise<JsonTestResults> {
```
sheremet-va, line 91: `logTasks` no longer logs tasks. Should we change a name?

sheremet-va (changes requested):

packages/vitest/src/node/reporters/json.ts
barraponto: yeah... it generates the logs, it just leaves writing up to the caller.

sheremet-va: Change the name then

Issue #7008, opened 2024-12-02
Expand locators implementation to support common chaining cases for better parity with Playwright locators


xeger:
### Clear and concise description of the problem

Playwright supports an `nth()` [method](https://playwright.dev/docs/api/class-locator#locator-nth) on its Locator which derives a new locator targeting a specific index in a multi-element set.

We can accomplish something similar in vitest using `expect.poll` of course:

```
await expect.poll( ()=> page.getByRole('button').elements()[3] ).toBeInTheDocument();
```

But this can get annoying if we want to perform an interaction e.g. `click` on the element -- we need to wrap the element in a locator again.

### Suggested solution

- [x] #7137
- [ ] Implement other high-value Playwright locator methods

### Alternative

As mentioned, there is always _some_ way to accomplish what we need using `expect.poll`, but the Playwright idioms of `nth`, `and`, `or` etc are highly productive and readable in comparison.

### Additional context

My testing team is interested in this feature and we may be willing to try implementing it, but I wanted to check on the maintainers' opinions first to make sure this is a reasonable direction for ivya + vitest to move in.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

xeger:
I've added support for nth; let me know what you think @sheremet-va:
https://github.com/vitest-dev/vitest/pull/7137

xeger:
On the topic of which other locator methods to support, `.filter` looks [extremely useful](https://playwright.dev/docs/other-locators#parent-element-locator) as it's recommended to solve a class of problem i.e. find parent element without relying on xpath.

sheremet-va:
> On the topic of which other locator methods to support, `.filter` looks [extremely useful](https://playwright.dev/docs/other-locators#parent-element-locator) as it's recommended to solve a class of problem i.e. find parent element without relying on xpath.
> 
> I'd want to do `filter` in a separate PR.
> 
> What about `and` / `or`? Need to look into their semantics more to see how useful they'd be to me.

We can support all of them. The engine already does, we just need a wrapper that constructs the selector

PR #7137, opened 2024-12-27
feat: Implement Locator.nth()

xeger:
### Description

This adds more parity with Playwright locators as discussed in https://github.com/vitest-dev/vitest/issues/7008.

Compared to `elements()[n]`, the `nth` modifier promotes easier-to-debug tests by throwing an obvious error message instead of letting a matcher fail on `undefined` or allowing the caller to try and call some additional locator method.

### Tests

### Documentation

### Changesets


packages/browser/src/client/tester/locators/index.ts
```diff
@@ -185,6 +185,15 @@ export abstract class Locator {
     return this.elements().map(element => this.elementLocator(element))
   }
 
+  public nth(index: number): Locator {
+    const element = this.elements()[index]
```
sheremet-va, line 189: Playwright supports `nth=` selector: https://github.com/microsoft/playwright/blob/3ec8ee7a9b9ee682f63ff155a36ea868951d8ae5/packages/playwright-core/src/client/locator.ts#L211

sheremet-va (changes requested):

packages/browser/src/client/tester/locators/index.ts
xeger: So it does. I reimplemented `nth()` to call the protected `locator` ... I admit that I am ignorant of the Playwright locator syntax or its role in the `@vitest/browser`, but I copied the `>> nth=` notation from Playwright and the blog tests still pass, so it appears successful. Let me know if that's what you had in mind.

The exception message is still a bit misleading i.e. it does not mention the `nth` chain when translating back to javascript:

```
Cannot find element with locator: locator('body').getByRole('listitem')
```

It looks like Ivya might be responsible for translation. Is it worth looking into this for clearer error reporting?

test/browser/fixtures/locators/blog.test.tsx
```diff
@@ -16,6 +16,16 @@ test('renders blog posts', async () => {
   expect(firstPost.element()).toHaveTextContent(/molestiae ut ut quas/)
   expect(firstPost.getByRole('heading').element()).toHaveTextContent(/occaecati excepturi/)
 
+  expect(screen.getByRole('listitem').nth(0).element()).toHaveTextContent(/molestiae ut ut quas/)
+  var thrown
+  try {
+    screen.getByRole('listitem').nth(666)
+  } catch(err) {
+    thrown = err
```
sheremet-va, line 24: This doesn't actually fail, `locator` just creates a locator, it doesn't search for elements so it cannot fail

sheremet-va (changes requested):

packages/browser/src/client/tester/locators/index.ts
sheremet-va: You can look into it if you want if you think it's important. It's also just a fork of playwright's locators implementation

test/browser/fixtures/locators/blog.test.tsx
xeger: Weird; I later verify that `thrown` is an Error with a suitable message (all synchronously) ... will investigate more closely.

xeger:
The exception message is 100% accurate now that I am using the `nth=` engine instead of reinventing ivya/playwright 😬 .

I feel a bit bad because I am writing selector notation by hand, using string interpolation, whereas the other methods are importing from ivya `locator` e.g. `getByRoleSelector` -- so, other methods don't pollute `vitest/browser` with selector semantics. However, ivya provides no helper method for `nth` engine, nor does [playwright main](https://github.com/microsoft/playwright/blob/4819747c85faa8cad0f33d770e1e63f6fb894d49/packages/playwright-core/src/utils/isomorphic/locatorUtils.ts#L58) at this time, so it's probably fine for now.

I also added `first` / `last` which are sugar, but useful. I think I'm happy with this PR. There are other useful locator methods e.g. `and` / `or` but I want to gain more experience actually using vitest + playwright to write tests before I blindly add features. If you know of any high-value or popular missing features, LMK and I will work on them in a separate PR.

docs/guide/browser/locators.md
```diff
@@ -709,6 +709,18 @@ Internally, this method calls `.elements` and wraps every element using [`page.e
 
 - [See `locator.elements()`](#elements)
 
+### nth, first, last
```
sheremet-va, line 712: these need to be separate and have examples

they should also come after all locator-returning methods

sheremet-va (changes requested):

docs/guide/browser/locators.md
xeger: I'll separate them and add examples.

They currently come after `all` which is a locator-returning method of `Locator`. I figured that the `getByXyz` were in their own group at the top because, although they are methods, they are also root locators of `Page` whereas `all`, `nth`, etc are only methods.

Was your goal for me to move `nth`, `first` and `last` to the top, below the `getXyz`, but leave `all` where it is? Or should they remain below `all`? (Or something else? Just LMK.)

sheremet-va: just move yours bellow `get`. `all` returns an array, not a locator

docs/guide/browser/locators.md
```diff
@@ -387,6 +387,69 @@ It is recommended to use this only after the other locators don't work for your
 
 - [testing-library's `ByTestId`](https://testing-library.com/docs/queries/bytestid/)
 
+## nth
+
+```ts
+function nth(index: number): Locator
+```
+
+This method returns a new locator that matches only a specific index within a multi-element query result. This is similar to calling `elements()[n]` but easier to use with asynchronous retry.
```
sheremet-va, line 396:
```suggestion
This method returns a new locator that matches only a specific index within a multi-element query result. Unlike `elements()[n]`, the `nth` locator will be retries until the element is present.
```

docs/guide/browser/locators.md
```diff
@@ -387,6 +387,69 @@ It is recommended to use this only after the other locators don't work for your
...
+page.getByRole('textbox').nth(4) // ❌
+```
+
+::: tip
+Before resorting to `nth`, you may find it useful to use chained locators to narrow down your search.
+Sometimes there is no better way to distinguish than by element position; although this can lead to flake, it's better than nothing.
+:::
+
+```tsx
+page.getByLabel('two').getByRole('input') // ✅ better alternative to page.nth(3)
```
sheremet-va, line 414: `page.nth()` is not a thing (and shouldn't be). I do wonder if we should expose `page.body` as `Locator`, but this is out of scope of this PR

sheremet-va (changes requested):

docs/guide/browser/locators.md
xeger: D'oh! Good catch. Fixing. (I assume you like the idea of annotating the ✅/❌ to explain the nuance.)

Re `page.body`, sounds like an excellent discussion to have on the Discord, or in an issue. (Along with `and` / `or` and the idea of an extension API...)

sheremet-va (approved):

docs/guide/browser/locators.md
```diff
@@ -387,6 +387,69 @@ It is recommended to use this only after the other locators don't work for your
...
+
+This method returns a new locator that matches only the first index of a multi-element query result.
+It is sugar for `nth(0)`.
+
+```html
+<input/> <input/> <input/>
+```
+
+```tsx
+page.getByRole('textbox').first() // ✅
```
Smrtnyk, line 433: I am trying to use this in my project
 await expect.element(screen.getByText("Org 1").first()).toBeVisible();
 
 test succeeds but there is a type error
 ```text
src/pages/Admin/PageAdminIssueReports.spec.ts:178:56 - error TS2339: Property 'first' does not exist on type 'Locator'.

178         await expect.element(screen.getByText("Org 1").first()).toBeVisible();

 ```
 Do I need to include some special types for this?

sheremet-va: Good point. All public methods need to be defined also here: https://github.com/vitest-dev/vitest/blob/main/packages/browser/context.d.ts

xeger: Whoops; sorry about that. Addressed in #7176

PR #7146, opened 2024-12-29
fix (core): filter different projects in watch mode

### LINKED ISSUE #7131: Project filter is includes shared tests across projects
### Describe the bug

I tried to use workspace on a single package that contains both server side and client side code, as well as code that can run on both environments.
I created one Vitest project config for each platform.

When filtering tests using project filter in watch mode, it run the shared tests on both projects instead of the project that I tried to filter.

```sh
src
  client
  server
  shared
vitest.config.browser.ts # run tests in `client` and `shared` folder
vitest.config.node.ts # run tests is `server` and `shared` folder
vitest.workspace.ts
```

`vitest run --project=node` is working correctly.
This issue only occurs in watch mode.

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-project-filter-issue

press `w` and enter `node` to filter the `node` project.

you will observe `|browser| src/shared/some.test.ts` is also executed.

### System Info

```shell
System:
    OS: Linux 5.15 Ubuntu 24.04.1 LTS 24.04.1 LTS (Noble Numbat)
    CPU: (16) x64 AMD Ryzen 7 5800X 8-Core Processor
    Memory: 25.32 GB / 31.30 GB
    Container: Yes
    Shell: 5.9 - /usr/bin/zsh
  Binaries:
    Node: 20.18.0 - /run/user/1000/fnm_multishells/56615_1735161007494/bin/node
    npm: 10.8.2 - /run/user/1000/fnm_multishells/56615_1735161007494/bin/npm
    pnpm: 9.12.2 - /run/user/1000/fnm_multishells/56615_1735161007494/bin/pnpm
  npmPackages:
    @vitest/coverage-v8: ^2.1.8 => 2.1.8 
    vitest: ^2.1.8 => 2.1.8
```

### Used Package Manager

pnpm

### Validations

DevJoaoLopes:
### Description

closes #7131 

- This PR resolves a bug in watch (-watch) mode where the filter (w key) is not filtering only the tests of a given project

### Tests

### Documentation

### Changesets


packages/vitest/src/node/core.ts
```diff
@@ -830,11 +830,12 @@ export class Vitest {
     }
 
     const specifications = files.flatMap(file => this.getModuleSpecifications(file))
```
sheremet-va, line 832: This should already return only filtered projects because it uses `vitest.projects` and not `vitest.resolvedProjects`. Why doesn't it?

test/filterWatchMode/package.json
```diff
@@ -0,0 +1,18 @@
+{
```
sheremet-va, line 1: Can you move the test to `test/watch`?

sheremet-va (changes requested):

packages/vitest/src/node/core.ts
DevJoaoLopes: Yes, but in those cases where we have different projects defined by `defineProject`, the filter of `files.flatMap(file => this.getModuleSpecifications(file))` is returning TestProjects beyond what it should.
Because of this, one of the possible solutions would be to apply the filter by name.

sheremet-va: You just rephrased what I asked. Why does it do that? `getModuleSpecification` shouldn't include filtered out projects

test/filterWatchMode/package.json
DevJoaoLopes: moved for `test/watch` @sheremet-va

packages/vitest/src/node/core.ts
DevJoaoLopes: @sheremet-va Do you have any idea what paths we can take to solve the problem? Insert validation into `specificationsWithFilterByProject` or remove cache validation for `getModuleSpecification` ?

sheremet-va: We can invalidate all cache if projects array is changed

DevJoaoLopes: @sheremet-va  done, if you can review please

DevJoaoLopes (commented):
Removed cache so we have the projects filtered properly.

sheremet-va:
> Removed cache so we have the projects filtered properly.

I said "invalidate" the cache, not remove the if check

DevJoaoLopes:
> > Removed cache so we have the projects filtered properly.
> 
> I said "invalidate" the cache, not remove the if check

okay, then I'll have to think about how to do that

DevJoaoLopes:
> > > Removed cache so we have the projects filtered properly.
> > 
> > 
> > I said "invalidate" the cache, not remove the if check
> 
> okay, then I'll have to think about how to do that

@sheremet-va One of the ways to maintain this cache validation was to enter a condition to compare filtered projects by invalidating the cache when there are differences.

PR #7152, opened 2024-12-31
fix: only use locator.element on last expect.element attempt (fix #7139)

### LINKED ISSUE #7139: Memory Leak on prettyDOM calls while waiting for expect.element
### Describe the bug

While waiting for expect.element to resolve, prettyDOM will be called on every attempt and its going to send the whole body as `dom` parameter. This makes memory usage go wild and at some point it goes OOM. Can confirm removing `prettyDom` usage in following line fixes the issue:

https://github.com/vitest-dev/vitest/blob/f4406f152490f06f68153dd632770b0bd11c2a74/packages/browser/src/client/tester/public-utils.ts#L69

IDK how Im the first person to notice this bug... I wonder if Im missing a config param or something.

### Reproduction

Had to reproduce it on private codebase so I cant really share code here.

image

image

image

After removing the `prettyDOM` call, all tests pass successfully:

image

### System Info

```shell
System:
    OS: macOS 15.1.1
    CPU: (14) arm64 Apple M3 Max
    Memory: 8.26 GB / 36.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.12.2 - ~/.nvm/versions/node/v20.12.2/bin/node
    Yarn: 4.5.1 - ~/.nvm/versions/node/v20.12.2/bin/yarn
    npm: 10.5.0 - ~/.nvm/versions/node/v20.12.2/bin/npm
  Browsers:
    Chrome: 131.0.6778.205
    Safari: 18.1.1
```

### Used Package Manager

yarn

### Validations

tsirlucas:
### Description

Fixes memory leak caused by multiple prettyDOM calls during expect.element retries.

ref: https://github.com/vitest-dev/vitest/issues/7139

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/browser/src/client/tester/expect-element.ts
+++ b/packages/browser/src/client/tester/expect-element.ts
@@ -18,6 +18,9 @@ export async function setupExpectDom() {
 
       const isNot = chai.util.flag(this, 'negate') as boolean
       const name = chai.util.flag(this, '_name') as string
+      // element selector uses prettyDOM under the hood, which is an expensive call
+      // that should not be called on each failed locator attempt to avoid memory leak:
+      // https://github.com/vitest-dev/vitest/issues/7139
       const isLastPollAttempt = chai.util.flag(this, '_isLastPollAttempt')
       // special case for `toBeInTheDocument` matcher
       if (isNot && name === 'toBeInTheDocument') {
--- a/packages/browser/src/client/tester/expect-element.ts
+++ b/packages/browser/src/client/tester/expect-element.ts
@@ -18,11 +18,23 @@ export async function setupExpectDom() {
 
       const isNot = chai.util.flag(this, 'negate') as boolean
       const name = chai.util.flag(this, '_name') as string
+      const isLastPollAttempt = chai.util.flag(this, "_isLastPollAttempt")
       // special case for `toBeInTheDocument` matcher
       if (isNot && name === 'toBeInTheDocument') {
         return elementOrLocator.query()
       }
-      return elementOrLocator.element()
+
+      if (isLastPollAttempt) {
+        return elementOrLocator.element()
+      }
+  
+      const result = elementOrLocator.query()
+  
+      if (!result) {
+        throw new Error(`Cannot find element with locator: ${JSON.stringify(elementOrLocator)}`)
+      }
+  
+      return result
     }, options)
   }
 }
--- a/packages/browser/src/client/tester/expect-element.ts
+++ b/packages/browser/src/client/tester/expect-element.ts
@@ -18,7 +18,7 @@ export async function setupExpectDom() {
 
       const isNot = chai.util.flag(this, 'negate') as boolean
       const name = chai.util.flag(this, '_name') as string
-      const isLastPollAttempt = chai.util.flag(this, "_isLastPollAttempt")
+      const isLastPollAttempt = chai.util.flag(this, '_isLastPollAttempt')
       // special case for `toBeInTheDocument` matcher
       if (isNot && name === 'toBeInTheDocument') {
         return elementOrLocator.query()
@@ -27,13 +27,13 @@ export async function setupExpectDom() {
       if (isLastPollAttempt) {
         return elementOrLocator.element()
       }
-  
+
       const result = elementOrLocator.query()
-  
+
       if (!result) {
         throw new Error(`Cannot find element with locator: ${JSON.stringify(elementOrLocator)}`)
       }
-  
+
       return result
     }, options)
   }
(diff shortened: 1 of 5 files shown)


test/browser/specs/runner.test.ts
```diff
@@ -28,8 +28,8 @@ describe('running browser tests', async () => {
       console.error(stderr)
     })
 
-    expect(browserResultJson.testResults).toHaveLength(19 * instances.length)
-    expect(passedTests).toHaveLength(17 * instances.length)
+    expect(browserResultJson.testResults).toHaveLength(20 * instances.length)
```
tsirlucas, line 31: is this correct? i assumed i needed to bump those since i added new test cases

packages/browser/src/client/tester/expect-element.ts
```diff
@@ -18,11 +18,23 @@ export async function setupExpectDom() {
 
       const isNot = chai.util.flag(this, 'negate') as boolean
       const name = chai.util.flag(this, '_name') as string
+      const isLastPollAttempt = chai.util.flag(this, '_isLastPollAttempt')
```
sheremet-va, line 21: let's also add a comment why we are doing this with a link to an issue

packages/vitest/src/integrations/chai/poll.ts
```diff
@@ -90,9 +80,24 @@ export function createExpectPoll(expect: ExpectStatic): ExpectStatic['poll'] {
...
-                intervalId = setTimeout(check, interval)
+                if (!chai.util.flag(assertion, '_isLastPollAttempt')) {
+                  intervalId = setTimeout(check, interval)
+                }
               }
             }
+            timeoutId = setTimeout(async () => {
+              clearTimeout(intervalId)
+              chai.util.flag(assertion, '_isLastPollAttempt', true)
+              await check()
```
sheremet-va, line 91: this check can succeed or throw an error, this needs special handling

packages/vitest/src/integrations/chai/poll.ts
```diff
@@ -90,9 +80,24 @@ export function createExpectPoll(expect: ExpectStatic): ExpectStatic['poll'] {
               }
               catch (err) {
                 lastError = err
-                intervalId = setTimeout(check, interval)
+                if (!chai.util.flag(assertion, '_isLastPollAttempt')) {
+                  intervalId = setTimeout(check, interval)
+                }
               }
             }
+            timeoutId = setTimeout(async () => {
```
sheremet-va, line 88: let's not make it async

test/browser/test/expect-element.test.ts
```diff
@@ -0,0 +1,26 @@
...
+// https://github.com/vitest-dev/vitest/issues/7139
+test('should only use element selector on last expect.element attempt', async () => {
+  const div = document.createElement('div')
+  const spanString = '<span>test</span>'
+  div.innerHTML = spanString
+  document.body.append(div)
+
+  const locator = page.getByText('non-existent')
+  const locatorElementMock = vi.fn(locator.element)
+  locator.element = locatorElementMock
```
sheremet-va, line 15: you can just do `vi.spyOn()`

sheremet-va (changes requested):

packages/vitest/src/integrations/chai/poll.ts
tsirlucas: hey

from what i can see the `check` function declaration is already using a try/catch block and handling the error.

if the assertion succeeds, it will resolve the promise successfuly
if the assertion fails, it will end in the catch block and wont throw an error

so the only scenario check can throw an error is if `chai.util.flag` access for some reason throw. if you want me to handle that case, can you provide more details on how? you want specific behavior or just logging?

im adding test cases for both scenarios: sucess/failure on last attempt

i added two new test cases to cover such scenarios

packages/vitest/src/integrations/chai/poll.ts
tsirlucas: i need to be able to await the `check` call. do you want me to use `Promise.then` instead?
can do it NP, just want to make sure we are aligned

sheremet-va: Yes, I meant using `promise.then.catch`

packages/vitest/src/integrations/chai/poll.ts
sheremet-va: I see, I guess the name is a bit misleading. We don't need an error handling here then

sheremet-va:
LGMT! Thank you!

PR #7168, opened 2025-01-03
fix: `cancelCurrentRun` awaits `runningPromise`

sheremet-va:

### Tests

### Documentation

### Changesets


AriPerkkio (changes requested):
~~I think we need to remove `await` here now. Otherwise double `CTRL+c` doesn't force exit if test threads are stuck/slow.~~

https://github.com/vitest-dev/vitest/blob/847d322119290541debc5769099eaf0069b46d35/packages/vitest/src/node/stdin.ts#L78-L82

E: Tested locally, works fine. The `_keypressHandler` is not awaited so no need to remove that other `await ctx.cancelCurrentRun`.

AriPerkkio (approved):

PR #7217, opened 2025-01-12
feat(ui): allow hide/show node_modules in module graph tab

### LINKED ISSUE #5363: Hidden node_modules dependencies in Module Graph

I found some dependencies(math.js etc.) will display in Module Graph, and it was regarded as inline modules。

image
image

### Suggested solution

Hidden those dependencies in Module Graph

### Alternative

_No response_

### Additional context

_No response_

### Validations

userquin:
### Description

closes #5363

By default the new `Hide node_modules` check checked. Check screenshots below using https://github.com/vitest-tests/browser-examples/tree/main/examples/react (using ui and browser package from this PR and latest Vitest 3 alpha version).

This PR removes the modules from node_modules in both inlined and externalized entries when `Hide node_modules` enabled before loading the ui (ignores inlined and externalize check states when filtering, `Hide node_modules` is global).

<details>
<summary>Screenshots</summary>

_Hide node_modules_
image

_Show node_modules_
image
</details>

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/ui/client/components/FileDetails.vue
+++ b/packages/ui/client/components/FileDetails.vue
@@ -18,6 +18,7 @@ const draft = ref(false)
 const hasGraphBeenDisplayed = ref(false)
 const loadingModuleGraph = ref(false)
 const currentFilepath = ref<string | undefined>(undefined)
+const hideNodeModules = ref(true)
 
 const graphData = computed(() => {
   const c = current.value
@@ -61,10 +62,12 @@ function onDraft(value: boolean) {
   draft.value = value
 }
 
-async function loadModuleGraph() {
+const nodeModuleRegex = /[/\\]node_modules[/\\]/
+
+async function loadModuleGraph(force = false) {
   if (
     loadingModuleGraph.value
-    || graphData.value?.filepath === currentFilepath.value
+    || (graphData.value?.filepath === currentFilepath.value && !force)
   ) {
     return
   }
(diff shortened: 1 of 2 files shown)


userquin:
We should check html reporter with and without browser mode, looks like node_modules not being added to the html.meta.json.gz (checked only with browser mode), I'll check the reporter logic tmr.

packages/ui/client/components/views/ViewModuleGraph.vue
```diff
@@ -21,6 +21,8 @@ const props = defineProps<{
   projectName: string
 }>()
 
+const modelValue = defineModel<boolean>({ default: true })
```
userquin, line 24: use `const modelValue = defineModel<boolean>({ required: true })` instead

userquin: and rename `modelValue` to `hideNodeModules`

packages/ui/client/components/FileDetails.vue
```diff
@@ -76,20 +79,32 @@ async function loadModuleGraph() {
...
       || (!graph.value.nodes.length && !graph.value.links.length)
     ) {
+      const moduleGraph = await client.rpc.getModuleGraph(
+        gd.projectName,
+        gd.filepath,
+        !!browserState,
+      )
+      // remove node_modules from the graph when enabled
+      if (hideNodeModules.value) {
+        moduleGraph.inlined = moduleGraph.inlined.filter(
```
userquin, line 99: We need to clone the data when using static html reporter, the metadata is global.

PR #7250, opened 2025-01-14
fix: prevent infinite loop on prettyDOM calls

tsirlucas:
### Description

Fixed infinite loop described in ~https://github.com/vitest-dev/vitest/pull/7152~ https://github.com/vitest-dev/vitest/issues/7249

Thought about adding a test case but it would just enter the loop and timeout the pipeline instead of failing (this was happening to my setup in the private repo haha). WDYT?

### Tests

### Documentation

### Changesets


packages/utils/src/display.ts
```diff
@@ -72,9 +72,11 @@ export function stringify(
       ...options,
     })
   }
-
+  
+  // Prevents infinite loop https://github.com/vitest-dev/vitest/issues/7249
+  const nextMaxDepth = maxDepth === Infinity ? Number.MAX_VALUE : Math.floor(maxDepth / 2)
```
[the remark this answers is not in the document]
tsirlucas: This is stil a pretty big recursion so I wonder how it will behave in CI env with limited resources. Will run a test and get back here.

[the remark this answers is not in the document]
tsirlucas: Ok I have pretty similar results time-wise doesnt matter if I use `Number.MAX_VALUE` or just `10`.

hi-ogawa:
I think we can avoid infinity for `prettyDOM`. I'd imagine practically it wouldn't matter whether the dom is truncated at 100, or even just 10. 

https://github.com/vitest-dev/vitest/blob/f9a628438a5462436b59dd9bdeffddada19a9e81/packages/browser/src/client/tester/public-utils.ts#L58

Pretty-format's maxDepth is quite scary and, for example, we reduced from 10 to 8 in other cases https://github.com/vitest-dev/vitest/pull/7133

Can you add test cases with repro from https://github.com/vitest-dev/vitest/issues/7249#issuecomment-2591651590?

packages/browser/src/node/providers/playwright.ts
```diff
@@ -185,6 +185,12 @@ export class PlaywrightBrowserProvider implements BrowserProvider {
       })
     }
 
+    // unhandled page crashes will hang vitest process
```
tsirlucas, line 188: this was making the tests hang https://github.com/vitest-dev/vitest/issues/7249#issuecomment-2592373055

packages/utils/src/display.ts
```diff
@@ -73,8 +73,10 @@ export function stringify(
     })
   }
 
+  // Prevents infinite loop https://github.com/vitest-dev/vitest/issues/7249
+  const nextMaxDepth = maxDepth === Infinity ? Number.MAX_SAFE_INTEGER : Math.floor(maxDepth / 2)
```
tsirlucas, line 77: MAX_NUMBER would still crash the browser, so Im using MAX_SAFE_INTEGER which works fine. If you want me to use any specific number, please send it here so I can replace

test/browser/test/browser-crash.test.ts
```diff
@@ -0,0 +1,15 @@
+import { it } from 'vitest';
+
+it.skip('fails gracefully when browser crashes', async () => {
```
tsirlucas, line 3: I dont know what to do with this test. It reproduces the issue where browser crashes, but if I enable it it will break the pipeline

sheremet-va: You can add this test to fixtures and create a new test file in test/browser/specs to run the fixture and assert that it fails correctly

tsirlucas:
ok i think im hitting a wall with the window tests... any idea what it may be?

tsirlucas:
ok... the old 20k nodes test was messing up with perf/stability in CI.

i adjusted the test case to test a simpler scenario that would still break without the changes in PR

also adjusted how we simulate crashes for same reason as above.

packages/utils/src/display.ts
```diff
@@ -73,8 +73,11 @@ export function stringify(
     })
   }
 
+  // Prevents infinite loop https://github.com/vitest-dev/vitest/issues/7249
+  const MAX_DEPTH = Number.MAX_SAFE_INTEGER
+  const nextMaxDepth = maxDepth > MAX_DEPTH ? MAX_DEPTH : Math.floor(maxDepth / 2)
   return result.length >= MAX_LENGTH && maxDepth > 1
-    ? stringify(object, Math.floor(maxDepth / 2))
+    ? stringify(object, nextMaxDepth, { maxLength, ...options })
```
tsirlucas, line 80: i noticed that maxLength and options were not being sent to recursive child calls before so i added it here

hi-ogawa, line 80: Maybe one liner?

```suggestion
    ? stringify(object, Math.floor(Math.min(maxDepth, Number.MAX_SAFE_INTEGER) / 2), { maxLength, ...options })
```

sheremet-va (approved):
LGTM!

PR #7272, opened 2025-01-16
Changed readHelloWorld in exampele 3 to return a string

pjotrsavitski:
### Description

File system mocking example would not run as `readHelloWorld` returns a buffer with test assertions expecting a text instead. It seems that the returned value should be a string, which is a simpler solution that to call `.toString()` in the tests. The name of the variable also suggests that text is expected to be returned.

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/guide/mocking.md
+++ b/docs/guide/mocking.md
@@ -385,7 +385,7 @@ module.exports = fs.promises
 import { readFileSync } from 'node:fs'
 
 export function readHelloWorld(path) {
-  return readFileSync(path)
+  return readFileSync(path, { encoding: 'utf8', flag: 'r' })
 }
 ```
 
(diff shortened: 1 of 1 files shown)


docs/guide/mocking.md
```diff
@@ -385,7 +385,7 @@ module.exports = fs.promises
 import { readFileSync } from 'node:fs'
 
 export function readHelloWorld(path) {
-  return readFileSync(path)
+  return readFileSync(path, { encoding: 'utf8', flag: 'r' })
```
sheremet-va, line 388:
```suggestion
  return readFileSync(path, 'utf-8')
```

sheremet-va (changes requested):

PR #7326, opened 2025-01-21
docs: clarify that locator.nth is zero based

patak-dev:
### Description

Add a note as Playwright has in [their docs for `nth`](https://playwright.dev/docs/api/class-locator#locator-nth).

I think that `nth()` should start from 1, in the same way that CSS `:nth-child` and usage in JS like `nthCalledWith`. I thought of adding an `at()` alias to avoid confusion, but I think aligning with playwright locators is more important here. Maybe the alias could first be proposed to Playwright, but I doubt they will be willing to change this.

### DIFF
--- a/docs/guide/browser/locators.md
+++ b/docs/guide/browser/locators.md
@@ -393,7 +393,7 @@ It is recommended to use this only after the other locators don't work for your 
 function nth(index: number): Locator
 ```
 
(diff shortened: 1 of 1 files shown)


sheremet-va (approved):

sheremet-va:
> Maybe the alias could first be proposed to Playwright, but I doubt they will be willing to change this.

We don't depend on Playwright, we can build our own locators.

PR #7349, opened 2025-01-23
feat: add "configureVitest" plugin hook

sheremet-va:
### Description

See discussion: https://github.com/vitest-dev/vitest/discussions/7104

### Tests

### Documentation

### Changesets


hi-ogawa (commented):
The code looks good, but I'm not familiar with use case. For example, the purpose of `force` flag doesn't look obvious to me.

If this serves storybook's need, I'm good to proceed with this.

docs/guide/workspace.md
```diff
@@ -231,6 +231,21 @@ bun test --project e2e --project unit
 ```
 :::
 
+Since Vitest 3.1, you can define a project that will always run even if you filter it out with a `--project` flag. Specify `force: true` flag when defining the project to always include it in your test run (note that this is only available in inline configs):
+
+```ts
+export default defineWorkspace([
+  {
+    // this project will always run even if it was filtered out
+    force: true,
```
AriPerkkio, line 240: Same confusion as @hi-ogawa in https://github.com/vitest-dev/vitest/pull/7349#pullrequestreview-2692775570.

Couldn't this same be achieved via plugin that always pushes the project into project filters?

```ts
{
  name: "vitest:always-include-unit-tests",
  configureVitest(context: VitestPluginContext) {
    if(!context.project.vitest.config.project.includes("unit")) {
      context.project.vitest.config.project.push("unit")
    }
  },
},
```

AriPerkkio (commented):
Looks good, great work! Once tests have been added this should be ready to go as is.

I tested this with two real-world use cases that developers building on top of Vitest have asked before:

> #### Check if coverage is enabled via `coverage.enabled` or `--coverage` (or resolved by both)
> 
> ```ts
> import { defineConfig } from "vitest/config";
> import type { Vite, VitestPluginContext } from "vitest/node";
> 
>   plugins: [
>     {
>       name: "vitest:is-coverage-enabled",
>       configureVitest(context: VitestPluginContext) {
>         console.log({ coverage: context.vitest.config.coverage.enabled });
>       },
>     } satisfies Vite.Plugin,
>   ],
> });
> ```
> 
> #### Configure `coverage.excludes` from workspace project's plugins
> 
> ```ts
> import { defineConfig } from "vitest/config";
> import type { VitestPluginContext } from "vitest/node";
> 
>   test: {
>     workspace: [
>       {
>         test: {
>           name: "Headless unit tests",
>           include: ["test/*.test.ts"],
>         },
>       },
>       {
>         test: {
>           name: "Stories",
>           include: ["src/**.stories.*ts*"],
>         },
>         plugins: [
>           {
>             name: "vitest:configure-coverage-for-storybook",
>             configureVitest(context: VitestPluginContext) {
>               // At this point we can assume `config.coverage` is resolved and defined to non-nullable
>               if (context.project.config.coverage.enabled) {
>                 // No need to check if `coverage.exclude` is array, just push.
>                 // No need to care about default values like "**/node_modules/**", push doesnt overwrite anything
>                 context.project.config.coverage.exclude.push("**/*.stories.*");
>               }
>             },
>           },
>         ],
>       },
>     ],
>   },
> });
> ```
> 
>

docs/guide/workspace.md
sheremet-va: > Couldn't this same be achieved via plugin that always pushes the project into project filters?

No, because the filtering happens when the project is resolved. This flag exists to bypass this. See:

https://github.com/vitest-dev/vitest/blob/8ff519a25714aef2d329e33d6452c6a5e30f062f/packages/vitest/src/node/plugins/workspace.ts#L88

sheremet-va: Hm, no, actually. The source of truth is `_projectFilters`, not `config.project`

PR #7357, opened 2025-01-24
fix(coverage): Yarn PnP virtual module resolving

AriPerkkio:
### Description

- Some package managers require rest of the ecosystem to add runtime checks for them

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/coverage-istanbul/src/provider.ts
+++ b/packages/coverage-istanbul/src/provider.ts
@@ -1,9 +1,12 @@
+import type { CoverageMap, CoverageMapData } from 'istanbul-lib-coverage'
+import type { SourceMap } from 'rollup'
 import type { CoverageProvider, ReportContext, ResolvedCoverageOptions, Vitest } from 'vitest/node'
 import { promises as fs } from 'node:fs'
+import { createRequire } from 'node:module'
 // @ts-expect-error missing types
 import { defaults as istanbulDefaults } from '@istanbuljs/schema'
 import createDebug from 'debug'
-import libCoverage, { type CoverageMap } from 'istanbul-lib-coverage'
+import libCoverage from 'istanbul-lib-coverage'
 import { createInstrumenter, type Instrumenter } from 'istanbul-lib-instrument'
 import libReport from 'istanbul-lib-report'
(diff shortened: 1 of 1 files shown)


packages/coverage-istanbul/src/provider.ts
```diff
@@ -204,3 +218,24 @@ async function transformCoverage(coverageMap: CoverageMap) {
...
+
+function resolveYarnPnpPaths(coverage: CoverageMapData) {
+  if (!pnp) {
+    return
+  }
+
+  for (const file of Object.keys(coverage) as (keyof typeof coverage)[]) {
+    const fileCoverage = coverage[file]
+
+    const filename = pnp.resolveVirtual(file)
```
mrginglymus, line 230: This should only be called on virtual paths, otherwise it will return `null`.

mrginglymus: Also, on windows this will return a `\` separated path, where the coverage map contains `/` separated paths (even on windows), so you'll need to apply your favourite `slash` function (or just `.replaceAll('\\', '/')`)

AriPerkkio:
Instead of adding Yarn PnP runtime checks in Vitest, let's recommend users to use Vite plugin that fixes PnP's virtual modules. 


```ts
import { createRequire } from "node:module";
import MagicString from "magic-string";

const pnp = createRequire(import.meta.url)("pnpapi");

export default defineConfig({
  plugins: [
    {
      /**
       * Yarn PnP loads virtual modules when user imports source files.
       * It doesn't add any indicators about the original source files, so we'll
       * need to add source maps that point to the original file manually.
       */
      name: "yarn-pnp-source-maps",
      transform(code, id) {
        if (id.includes(".yarn/__virtual__")) {
          const s = new MagicString(code);
          const map = s.generateMap({ hires: true });
          const filename = pnp.resolveVirtual(id).replaceAll("\\", "/");
          map.sources = [filename];
          return { code, map };
        }
      },
    },
  ],
  test: {
    coverage: {
      enabled: true,
      provider: "istanbul",
      include: ["packages", ".yarn/**"],
      exclude: ["**/node_modules/**"],
    },
  },
});
```

PR #7360, opened 2025-01-25
docs: fix missing eslint and prettier in exclude default config

kimulaco:
### Description

Thanks for the great testing framework!

I found a mismatch between docs and code for the default `exclude` config.

The current documentation is missing `eslint` and `prettier` in the default exclude patterns. This PR updates the documentation to match the actual implementation.

Reference to actual implementation:
https://github.com/vitest-dev/vitest/blob/v3.0.4/packages/vitest/src/defaults.ts#L13-L19

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -124,7 +124,7 @@ When using coverage, Vitest automatically adds test files `include` patterns to 
 ### exclude
 
 - **Type:** `string[]`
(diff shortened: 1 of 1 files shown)


AriPerkkio (approved):
Thanks, looks good.

(Though it doesn't make sense that our `test.exclude` has patterns like `eslint.config.*` as exclusion, when default `test.include` is `['**/*.{test,spec}.?(c|m)[jt]s?(x)']`.)

kimulaco:
@AriPerkkio Thanks for the review!

Indeed, we should probably reconsider these default `test.exclude` patterns.

AriPerkkio:
Clean up for the `test.exclude` will be done in https://github.com/vitest-dev/vitest/pull/6287.

Issue #7398, opened 2025-02-01
watcher throws an error after removing existing test files


soc221b:
### Describe the bug

In watch mode, it will throws an error when we remove an existing test files and then trigger a rerun because existing test file cache is not deleted after the unlink function is invoked.

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-ng9dneil

1. wait until "Waiting for file changes..." appears
2. rename `test/basic.test.ts` to `test/renamed.test.ts`
3. wait until "Waiting for file changes..." appears again
4. press `a` to run all tests

### System Info

```shell
System:
  OS: macOS 14.5
  CPU: (8) x64 Intel(R) Core(TM) i5-1038NG7 CPU @ 2.00GHz
  Memory: 1.00 GB / 16.00 GB
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 20.18.1 - ~/.nvm/versions/node/v20.18.1/bin/node
  Yarn: 1.22.15 - ~/.yarn/bin/yarn
  npm: 10.8.2 - ~/.nvm/versions/node/v20.18.1/bin/npm
  pnpm: 9.15.4 - ~/Library/pnpm/pnpm
  bun: 1.2.0 - ~/Library/pnpm/bun
  Watchman: 2024.01.22.00 - /usr/local/bin/watchman
Browsers:
  Chrome: 132.0.6834.160
  Safari: 17.5
npmPackages:
  @vitest/browser: workspace:* => 3.0.4 
  @vitest/coverage-istanbul: workspace:* => 3.0.4 
  @vitest/coverage-v8: workspace:* => 3.0.4 
  @vitest/ui: workspace:* => 3.0.4 
  vite: ^5.4.0 => 5.4.0 
  vitest: workspace:* => 3.0.4
```

### Used Package Manager

npm

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vitest.dev/guide/).
- [x] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
[...]

PR #7399, opened 2025-02-01
fix(watch): properly remove cache after removing existing test files

soc221b:
### Description

Fixes https://github.com/vitest-dev/vitest/issues/7398

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/project.ts
+++ b/packages/vitest/src/node/project.ts
@@ -390,7 +390,8 @@ export class TestProject {
     this.testFilesList?.push(testPath)
   }
 
-  removeCachedTestFile(testPath: string): void {
+  /** @internal */
+  _removeCachedTestFile(testPath: string): void {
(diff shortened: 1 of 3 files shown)


packages/vitest/src/node/project.ts
```diff
@@ -390,6 +390,12 @@ export class TestProject {
     this.testFilesList?.push(testPath)
   }
 
+  removeCachedTestFile(testPath: string): void {
```
sheremet-va, line 393: this is an internal method

it should start with _ and have `@internal` jsdoc

sheremet-va (changes requested):

packages/vitest/src/node/project.ts
soc221b: Thanks! I've updated it.

sheremet-va (approved):

PR #7419, opened 2025-02-04
feat(reporters): customize project name colors through options

kevinmarrec:
### Description

Fixes https://github.com/vitest-dev/vitest/issues/6481
Follows #7371 

Introduce a `colors` reporter option to be able to customize what colors are used when formatting reporter outputs.

Maybe what we would to introduce instead is the whole `formatProjectName` function override.

### Tests

I'd like some help to writing tests for this.

### Documentation

Needs to be discussed, the option name is subject to change.

### DIFF
--- a/packages/vitest/src/node/reporters/benchmark/reporter.ts
+++ b/packages/vitest/src/node/reporters/benchmark/reporter.ts
@@ -56,7 +56,7 @@ export class BenchmarkReporter extends DefaultReporter {
     const duration = task.result.duration
 
     if (benches.length > 0 && benches.every(t => t.result?.state !== 'run' && t.result?.state !== 'queued')) {
-      let title = `\n ${getStateSymbol(task)} ${formatProjectName(task.file.projectName)}${getFullName(task, c.dim(' > '))}`
+      let title = `\n ${getStateSymbol(task)} ${formatProjectName(task.file.projectName, { colors: this.colors })}${getFullName(task, c.dim(' > '))}`
 
       if (duration != null && duration > this.ctx.config.slowTestThreshold) {
         title += c.yellow(` ${Math.round(duration)}${c.dim('ms')}`)
--- a/packages/vitest/src/node/reporters/default.ts
+++ b/packages/vitest/src/node/reporters/default.ts
@@ -1,17 +1,21 @@
 import type { Vitest } from '../core'
 import type { TestSpecification } from '../spec'
 import type { BaseOptions } from './base'
+import type { ReporterColors } from './renderers/utils'
 import type { ReportedHookContext, TestCase, TestModule } from './reported-tasks'
 import { BaseReporter } from './base'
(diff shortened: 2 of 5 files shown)


packages/vitest/src/node/reporters/renderers/utils.ts
```diff
@@ -223,7 +232,7 @@ export function formatProjectName(name: string | undefined, suffix = ' ') {
     .split('')
     .reduce((acc, v, idx) => acc + v.charCodeAt(0) + idx, 0)
 
-  const colors = [c.black, c.yellow, c.cyan, c.green, c.magenta]
+  colors = typeof colors === 'function' ? colors(c) : colors
```
kevinmarrec, line 235: Provides `tinyrainbow` :

```ts
{ colors: (c) => [c.blue, c.red] }
```

packages/vitest/src/node/reporters/benchmark/reporter.ts
```diff
@@ -56,7 +56,7 @@ export class BenchmarkReporter extends DefaultReporter {
     const duration = task.result.duration
 
     if (benches.length > 0 && benches.every(t => t.result?.state !== 'run' && t.result?.state !== 'queued')) {
-      let title = `\n ${getStateSymbol(task)} ${formatProjectName(task.file.projectName)}${getFullName(task, c.dim(' > '))}`
+      let title = `\n ${getStateSymbol(task)} ${formatProjectName(task.file.projectName, { colors: this.colors })}${getFullName(task, c.dim(' > '))}`
```
kevinmarrec, line 59: Sounds like little cumbersome, I'm open for refactoring for something like `formatOptions`

packages/vitest/src/node/reporters/default.ts
```diff
@@ -1,18 +1,22 @@
...
 export interface DefaultReporterOptions extends BaseOptions {
+  colors?: ReporterColors
   summary?: boolean
 }
 
 export class DefaultReporter extends BaseReporter {
   private options: DefaultReporterOptions
   private summary?: SummaryReporter
 
+  protected colors: DefaultReporterOptions['colors']
```
sheremet-va, line 18: 1. Colors should also be inherited by UI, so limiting it to reporters won't work
2. Since UI also requires colors, we can't use ANSII or functions, it should be hard coded names

sheremet-va (changes requested):

packages/vitest/src/node/reporters/default.ts
kevinmarrec: Thank you for the feedback, I'll work on another solution.

kevinmarrec:
> Colors should also be inherited by UI, so limiting it to reporters won't work

Closing this, will eventually work on a more suitable solution that would also work for Vitest UI.

The option should be configured at upper level therefore `test` object and not `test.reporters`

PR #7444, opened 2025-02-07
VoidZero reference information

alexted:
add VoidZero reference info


patak-dev:
Hey @alexted, thanks for the PR but these kind of changes are initiated from between the team of OSS projects. Please avoid sending further PRs like this one to OSS projects.

Issue #7403, opened 2025-02-01
Invalid config from browser creator


nickserv:
### Describe the bug

I was confused when trying to use `vitest init browser` because it writes a nonexistent `test.browser.configs` config option, producing an invalid config. It seems like this should be `test.browser.instances`, though the documentation and TypeScript types are correct.

### Reproduction

Run `vitest init browser` with any options, then view the generated config.

#### Config patch
```diff
--- vitest.bad.config.js
+++ vitest.bad.config.js
@@ -1,9 +1,9 @@
 import { defineConfig } from "vitest/config";
 
 export default defineConfig({
        test: {
                browser: {
-                       configs: [],
+                       instances: [],
                },
        },
 });
```

#### Engineering note
There should probably be an end-to-end, smoke, or type test that asserts generated configs are valid.

### System Info

```shell
System:
    OS: macOS 12.7.6
    CPU: (4) x64 Intel(R) Core(TM) i5-5257U CPU @ 2.70GHz
    Memory: 190.06 MB / 8.00 GB
    Shell: 3.7.1 - /usr/local/bin/fish
  Binaries:
    Node: 23.6.0 - /usr/local/bin/node
    npm: 11.1.0 - /usr/local/bin/npm
    pnpm: 9.11.0 - /usr/local/bin/pnpm
  Browsers:
    Chrome: 132.0.6834.160
    Safari: 17.6
  npmPackages:
    @vitest/browser: ^3.0.4 => 3.0.4 
    playwright: ^1.50.1 => 1.50.1
```

### Used Package Manager

npm

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vitest.dev/guide/).
- [x] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

AriPerkkio:
[...]

PR #7475, opened 2025-02-12
fix(init): invalid browser config

AriPerkkio:
### Description

- Fixes https://github.com/vitest-dev/vitest/issues/7403

### Tests

### Documentation

### Changesets


test/cli/test/init.test.ts
```diff
@@ -0,0 +1,90 @@
...
+import { runVitestCli } from '../../test-utils'
+
+const ARROW_DOWN = '\u001B[B'
+const ENTER = '\n'
+
+const cwd = 'fixtures/browser-init'
+
+beforeEach(async () => {
+  await cleanup()
+  return cleanup
```
sheremet-va, line 13: why can't you just have the `cleanup` content in `beforeEach`?

sheremet-va (approved):

test/cli/test/init.test.ts
AriPerkkio: I'm not sure I understand - could you provide example?

sheremet-va: Ah, I see now you are calling the cleanup twice 🤔 

Why do you need to call it twice though?

AriPerkkio: Well it should be enough to call it just as `afterEach`. But while I was developing this test case I had to hit `CTRL+c` couple of times and it always left the files on fs in that immediate termination.

PR #7509, opened 2025-02-17
feat: support rolldown-vite

sheremet-va:
### Description

Good thing is that Vitest works with vite-rolldown for the most part (99%)

- Removed sveltekit example for now because pnpm didn't like vite override and I don't want to deal with it 👀 
- Issues:
  - vite-rolldown doesn't expose `Rollup.Program` (or `Rolldown.Program`?)
  - rolldown doesn't support `using` syntax - we are transpiling the code down to node18, but it still doesn't work (although we use the `esbuild` field 🤔 ) - the test is skipped for now
    - there is no way to change the transform plugin if `experimental.enableNativePlugin` is `true` (the default)
    - `examples/lit` decorators don't seem to work correctly:
> target was modified to include ES2021 because useDefineForClassFields is set to false and oxc does not support transforming useDefineForClassFields=false for ES2022+ yet 
  - ✅ vite-rolldown uses `'' + import.meta.url` when transforming `new URL` to opt-out of rolldown's transform (this is an easy regexp change). Vitest also changes this transform to make the URL closer to the one that will be resolved in the browser (in node.js `import.meta.url` refers to a file, but we need a server URL there)
  - ✅ does `optimizeDeps.rollupOptions.resolve.alias` work? Browser Mode aliases `vue` packages to CJS versions testing-library is CJS (and it should import the same version of test-utils), and also we want to have the ability to use string slots and string templates
  - ✅ seems like I see this message even if I didn't specify any `esbuildOptions` (we always check for an empty stderr and this breaks some tests):
> You've set "optimizeDeps.esbuildOptions" in your config. This is deprecated and vite already use rollup to optimize packages. Please use "optimizeDeps.rollupOptions" instead.

The value of `esbuildOptions` is `{ preserveSymlinks: false }` and the environment name is `client` 🤔 

This comes probably from here:

https://github.com/rolldown/vite/blob/1eee1f56ee90a08d82f465900caea6f2a618d358/packages/vite/src/node/config.ts#L1139

  - ✅ styles in the UI seem to be completely broken when built with vite-rolldown

Screenshot 2025-02-17 at 16 07 07

  - ~source maps sometimes are not correct. See `test/config/unhandled-rejections.test.ts` - the column should be 42 (*this MIGHT BE an issue with custom Vitest plugins*, but it was not a problem before):~

```
  ❯ setup-unhandled-rejections.ts:2:41
       1| export function setup() {
       2|   void new Promise((_, reject) => reject(new Error('intentional unhand…
        |                                         ^ (this is incorrect position)
        |                                          ^ (it should be here where new Error starts)
       3| }
```

-----

> [!NOTE]
> Seems like we always had "Re-optimizing dependencies because vite config has changed" because of incorrect usage of `cacheDir`


hi-ogawa:
> Good thing is that Vitest works with vite-rolldown for the most part (99%)

Awesome progress already! 🎉

> rolldown doesn't support `using` syntax - we are transpiling the code down to node18, but it still doesn't work (although we use the `esbuild` field 🤔 ) - the test is skipped for now
> * ...
> * `examples/lit` decorators don't seem to work correctly:

These two are oxc transform issues and they are tracked in https://github.com/oxc-project/oxc/issues/9168 and https://github.com/oxc-project/oxc/issues/9129.

> styles in the UI seem to be completely broken when built with vite-rolldown

Probably this is because of unocss plugin not working https://github.com/unocss/unocss/issues/4403.

sapphi-red:
Great!

> * does `optimizeDeps.rollupOptions.resolve.alias` work? Browser Mode aliases `vue` packages to CJS versions testing-library is CJS (and it should import the same version of test-utils), and also we want to have the ability to use string slots and string templates

I've fixed that yesterday so maybe it'll work if you upgrade rolldown-vite.
https://github.com/rolldown/vite/pull/43/commits/1eee1f56ee90a08d82f465900caea6f2a618d358

> * seems like I see this message even if I didn't specify any `esbuildOptions` (we always check for an empty stderr and this breaks some tests):
> 
> > You've set "optimizeDeps.esbuildOptions" in your config. This is deprecated and vite already use rollup to optimize packages. Please use "optimizeDeps.rollupOptions" instead.

I'll take a look this one.

> * optimizer unexpectedly re-optimizes dependencies with a message:

Is it actually re-optimized? or is it just the message?

sapphi-red:
I've fixed the things related to warnings.

I also took a look the sourcemap one. I found that the sourcemaps output by OXC are slightly different from esbuild (https://github.com/oxc-project/oxc/issues/9187). But I'm not sure if that's the reason.
I extracted the result here
https://github.com/vitest-dev/vitest/blob/6cc408d66445de5472f61ea527ceea6e0dddb16e/packages/vite-node/src/client.ts#L507
and tried to reproduce it minimally by doing
```js
// run.mjs
import 'source-map-support/register.js'
globalThis.__vite_ssr_exports__ = {}
await import('./run2.mjs')
globalThis.__vite_ssr_exports__.setup()

// ---------------------------
// run2.mjs
'use strict';function setup() {
	void new Promise((_, reject) => reject(new Error("intentional unhandled rejection")));
}
Object.defineProperty(__vite_ssr_exports__, "setup", { enumerable: true, configurable: true, get(){ return setup }});

//# sourceMappingURL=data:application/json;charset=utf-8;base64,eyJ2ZXJzaW9uIjozLCJtYXBwaW5ncyI6IkFBQU8sU0FBUyxRQUFRO0FBQ3RCLE1BQUssSUFBSSxRQUFRLENBQUMsR0FBRyxXQUFXLE9BQU8sSUFBSSxNQUFNLG1DQUFtQztBQUNyRiIsIm5hbWVzIjpbXSwiaWdub3JlTGlzdCI6W10sInNvdXJjZXMiOlsic2V0dXAtdW5oYW5kbGVkLXJlamVjdGlvbnMudHMiXSwic291cmNlc0NvbnRlbnQiOlsiZXhwb3J0IGZ1bmN0aW9uIHNldHVwKCkge1xuICB2b2lkIG5ldyBQcm9taXNlKChfLCByZWplY3QpID0+IHJlamVjdChuZXcgRXJyb3IoJ2ludGVudGlvbmFsIHVuaGFuZGxlZCByZWplY3Rpb24nKSkpXG59XG4iXSwiZmlsZSI6IkQ6L2RvY3VtZW50cy9HaXRIdWIvdml0ZXN0L3Rlc3QvY29uZmlnL2ZpeHR1cmVzL3VuaGFuZGxlZC1yZWplY3Rpb25zL3NldHVwLXVuaGFuZGxlZC1yZWplY3Rpb25zLnRzIn0=
```
But the error message showed `setup-unhandled-rejections.ts:2:42`.

hi-ogawa:
Ah, I should've mentioned earlier. Source map for global setup is not supported due to Vitest side https://github.com/vitest-dev/vitest/pull/7101. The current assertion `setup-unhandled-rejections.ts:2:42` works because transform output miraculously 
 matched original source.

sapphi-red:
Ah, I see.

hi-ogawa:
@sheremet-va How did you run this PR locally? If I added rolldown-vite overrides, some dts build is failing, for example:

```sh
$ pnpm -C packages/vite-node build
src/source-map.ts(36,37): error TS7006: Parameter 'source' implicitly has an 'any' type.

[!] (plugin dts) RollupError: [plugin dts] src/source-map.ts: Failed to compile. Check the logs above.
src/source-map.ts
```

sheremet-va:
> @sheremet-va How did you run this PR locally? If I added rolldown-vite overrides, some dts build is failing, for example:
> 
> 
> 
> ```sh
> 
> $ pnpm -C packages/vite-node build
> 
> src/source-map.ts(36,37): error TS7006: Parameter 'source' implicitly has an 'any' type.
> 
> 
> 
> [!] (plugin dts) RollupError: [plugin dts] src/source-map.ts: Failed to compile. Check the logs above.
> 
> src/source-map.ts
> 
> ```
> 
> 
> 
> (The odd thing is that I don't see type error red squiggles on vscode and I have no idea what's the issue)

I added `link:` override manually to overrides in the top level package.json

Did you build vite package?

hi-ogawa:
Hmm, I was using `file:` and that was failing. Switched to `link:` and now it's working on Vitest :+1: I can continue with this, but it might come back when using released package :thinking:

sheremet-va:
> Ah, I should've mentioned earlier. Source map for global setup is not supported due to Vitest side #7101. The current assertion `setup-unhandled-rejections.ts:2:42` works because transform output miraculously matched original source.

There is another difference with source maps. I am not sure if it's correct or not, I think it is caused by a difference in boundaries. This code previously had a different column:

```ts
test.each([1, 2])('custom %s', () => {
//                ^ previously returned this column
//               ^ now returns this column
})
```

Raw JavaScript returns the column that Rolldown returns. We had to do `column-1` to get the correct column here, now this code is gone.

sheremet-va:
> I've fixed that yesterday so maybe it'll work if you upgrade rolldown-vite.

I used the same version yesterday. Still doesn't work. I also tried `^@vue/test-utils$` alias, but it still doesn't seem to work. I am testing in this repo - https://github.com/vitest-dev/vitest-browser-vue (the last test expects `template` to work)

sapphi-red:
> > I've fixed that yesterday so maybe it'll work if you upgrade rolldown-vite.
> 
> I used the same version yesterday. Still doesn't work (tried again today). I also tried `^@vue/test-utils$` alias, but it still doesn't seem to work. I am testing in this repo - https://github.com/vitest-dev/vitest-browser-vue (the last test expects `template` to work)

I think the value needs to be a resolved path instead of a unresolved specifier here.
https://github.com/vitest-dev/vitest/pull/7509/files#diff-929cb96abb1c2e297c6711af0875db4ecdc700c54b824cc4ce7e0716657b28aeR537-R543
But given that the `importer` is not static, I guess it needs to be a plugin like it was in esbuild.
It's confusing that `resolve.alias` of rolldown works differently from vite or plugin-alias's `resolve.alias`. I'll make a feedback to them.

sheremet-va:
> Ah, no, it's not that `resolve.alias` working differently, it's because Vite's internal plugin runs before `resolve.alias`. I think the workaround for now would be to be using a plugin instead as the custom plugins runs before the internal plugin.

A rolldown plugin seems to work correctly.

sapphi-red:
I merged https://github.com/rolldown/vite/pull/82 now, so the UI should be working now.

sheremet-va:
> I merged [rolldown/vite#82](https://github.com/rolldown/vite/pull/82) now, so the UI should be working now.

Can confirm it does work now:

Screenshot 2025-02-19 at 15 08 52

sheremet-va:
> I merged [rolldown/vite#82](https://github.com/rolldown/vite/pull/82) now, so the UI should be working now.

Is there a way to specify rolldown-vite as a dependency in this PR so CI can run with it?

sapphi-red:
> vite-rolldown doesn't expose `Rollup.Program` (or `Rolldown.Program`?)

It seems `Rollup.Program` was not meant to be public, but was import-able due to https://github.com/microsoft/TypeScript/issues/38592
https://github.com/rollup/rollup/blob/8f667b7c15b176728449a4917cb29fe5ee3a1c0c/src/rollup/types.d.ts#L1049

> * if `experimental.enableNativePlugin` is `true` (the default)

That option is disable by default. rolldown-vite uses OXC regardless of that option.

sheremet-va:
> It seems `Rollup.Program` was not meant to be public, but was import-able due to

How do you type the parse function then 🤔

sheremet-va:
> That option is disable by default. rolldown-vite uses OXC regardless of that option.

From what I can see in the code, the transform plugin doesn’t accept any options 🤔 JSDoc also says it’s true by default

sapphi-red:
> > It seems `Rollup.Program` was not meant to be public, but was import-able due to
> 
> How do you type the parse function then 🤔

I'm not sure, just that the `export` is not written there. It could be not intended. You can use `ReturnType<typeof this.parse>` (it's hacky though).

> > That option is disable by default. rolldown-vite uses OXC regardless of that option.
> 
> From what I can see in the code, the transform plugin doesn’t accept any options 🤔 JSDoc also says it’s true by default

Ah, the JSDoc is incorrect. To clarify, you can pass an option via `config.oxc` to OXC tranform plugin, but it won't unblock you.

test/cli/test/public-api.test.ts
```diff
@@ -94,6 +94,7 @@ it.each([
     expect(files[0].tasks[index + 1].name).toBe(`custom ${name}`)
     expect(files[0].tasks[index + 1].location).toEqual({
       line: 18,
+      // FIXME: rolldown (possibly) gives incorrect location
```
hi-ogawa, line 97: I'm suspecting that this might the bug on esbuild transform, which required the following workaround

https://github.com/vitest-dev/vitest/blob/41cc1d8a48bf866fa3bd4eed1fabc0dc5cf7383a/packages/runner/src/suite.ts#L851-L857

For rolldown vite / oxc, we can remove the extra `+ 1` and the result should matches as plain js stack.

Here is a comparison of source map:
(btw, I made this quick dirty stackblitz app in https://github.com/hi-ogawa/reproductions/tree/main/oxc-sourcemap-debug)

image

test/config/test/unhandled-rejections.test.ts
```diff
@@ -12,5 +12,6 @@ test('unhandled rejections of main thread are reported even when no reporter is
   expect(exitCode).toBe(1)
   expect(stderr).toContain('Unhandled Rejection')
   expect(stderr).toContain('Error: intentional unhandled rejection')
+  // FIXME: this should be the correct location, but rolldown (possibly) shows 2:41
```
hi-ogawa, line 15: I think this one is also alright. It should be fixed once we figure out global setup source map https://github.com/vitest-dev/vitest/pull/7101.
While esbuild preserved column exactly, oxc replaced two spaces with one tab, so the transformed output became one character off from the original one.

<details><summary>source map comparison</summary>

image

</details>

hi-ogawa:
> > Now we get the same error with pkg.pr.new 😄 Seems like it doesn't pick up rolldown types 🤔
> 
> Slight progress. I added `"rolldown": "1.0.0-beta.3-commit.b546e53"` (the same one used by current rolldown-vite) to root and `pnpm -C packages/vite-node build` got through. Now it's failing with `pnpm -C packages/browser/ build:node` but this looks like there are duplicate vite somewhere.
> 
> I pushed my hack and build passed on CI 😀

I found the culprit of dts build error. This is because `rollup-plugin-dts` sets `compilerOptions.preserveSymlinks: true` by default https://github.com/Swatinem/rollup-plugin-dts/blob/deddf70ccc09345f41b3129004ce3321b0529a54/src/program.ts#L25-L26. This likely requires externalized (even transitively) dep to be visible directly from project, so `rolldown` is required to be in the root. The same error can be seen on IDE by setting `preserveSymlinks: true` inside our `tsconfig.json`.

While I have mostly same setup for rolldown-vite testing PR on my plugin https://github.com/hi-ogawa/vite-plugins/pull/673 where I use `rollup-plugin-dts` through `tsup`, it turns out `tsup` overwrites such "likely bad" default back to `preserveSymlinks: false` https://github.com/egoist/tsup/blob/0328fd64a771adb0789968775c22b4317683fafe/src/rollup.ts#L141-L145, so people using `tsup` won't see this issue.

sheremet-va:
So we don't see this issue in main because we have rollup in dependencies and it makes it visible?

hi-ogawa:
> So we don't see this issue in main because we have rollup in dependencies and it makes it visible?

Yes, I think that's the reason. It's not possible to reproduce the exact same scenario with rollup because `rollup-plugin-dts` obviously requires `rollup` to be installed in root :upside_down_face: 

For our case, I think either we can install `rolldown` in root or manually add `dts({ compilerOptions: { preserveSymlinks: false } })`.

sheremet-va:
Seems like all tests are passing, except:

- `examples/lit`: need to update `rolldown-vite` to include the fix from https://github.com/oxc-project/oxc/issues/9129
- `test/config/test/unhandled-rejections.test.ts`: not sure what to do with this one, the previous test was kind of useless because it relied on the wrong transform
- `esnext.test.ts`: rolldown doesn't support `using` because `oxc` didn't implement it yet: https://github.com/oxc-project/oxc/issues/9168
- vitepress doesn't recognize `blog.test.ts` factory (not related to vitest)

And that's it 😄

docs/.vitepress/blog.data.ts
```diff
@@ -9,7 +9,10 @@ interface Post {
   }
 }
 
-declare const data: Post[]
+// TODO: the exact typed data loader pattern is not supported by rolldown/oxc transform
+// https://vitepress.dev/guide/data-loading#typed-data-loaders
+// eslint-disable-next-line import/no-mutable-exports
+let data!: Post[]
 export { data }
```
hi-ogawa, line 16: Vitepress's typed data loader pattern looks like a bit broken concept. Esbuild-based config loader somehow eliminated to `export {}`, but rolldown/oxc transpiles it to `export { data }` and that causes a runtime error.

- esbuild https://esbuild.github.io/try/#YgAwLjI1LjEAeyBmb3JtYXQ6ICdlc20nIH0AZQBlbnRyeS50cwBkZWNsYXJlIGNvbnN0IGRhdGE6IFBvc3RbXQpleHBvcnQgeyBkYXRhIH0
- rolldown https://stackblitz.com/edit/github-ga7ubbtm?file=src%2Ftest.ts,rolldown.config.mjs

sapphi-red: It seems it's related to https://github.com/rolldown/rolldown/issues/3048.

hi-ogawa: Right, that one! It looks so familiar but I didn't remember where I saw. Indeed, this workaround looks cooler :smile: 

```ts
export declare const data: Post[]
```

sheremet-va:
@sapphi-red can we merge this in any capacity for rolldown-vite? Some APIs have changed completely (like `rollupOptions` instead of `esbuildOptions` for optimized dependencies). Is there a way to know which `vite` is loaded?

sapphi-red:
`rolldown-vite` has `this.meta.rolldownVersion` so you should be able to differentiate with the existence of that property. The other way would be to check `import * as vite from 'vite'; vite.rolldownVersion`.

examples/lit/test/basic.test.ts
```diff
@@ -3,7 +3,8 @@ import { beforeEach, describe, expect, it } from 'vitest'
 
 import '../src/my-button.js'
 
-describe('Button with increment', async () => {
+// TODO: wait until rolldown has https://github.com/oxc-project/oxc/issues/9129
+describe.skip('Button with increment', async () => {
```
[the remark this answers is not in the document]
sapphi-red: Which rolldown-vite version did you try?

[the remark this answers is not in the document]
sheremet-va: `@beta`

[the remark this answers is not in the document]
sapphi-red: Made an issue here: https://github.com/oxc-project/oxc/issues/10147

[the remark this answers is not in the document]
sheremet-va: oh, I didn't realise the target is related this issue 🤔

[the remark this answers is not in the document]
sheremet-va: Can confirm it works!

packages/runner/src/suite.ts
```diff
@@ -846,21 +846,15 @@ function formatTemplateString(cases: any[], args: any[]): any[] {
   return res
 }
 
-function findTestFileStackTrace(error: string, each: boolean) {
```
sheremet-va, line 849: not sure how we can support both vite and rolldown-vite here 🤔 @hi-ogawa

docs/.vitepress/blog.data.ts
```diff
@@ -9,8 +9,10 @@ interface Post {
   }
 }
 
-declare const data: Post[]
-export { data }
+// TODO: the exact typed data loader pattern is not supported by rolldown/oxc
+// https://vitepress.dev/guide/data-loading#typed-data-loaders
+// https://github.com/rolldown/rolldown/issues/3048#issuecomment-2547371389
+export declare const data: Post[]
```
sapphi-red: This is now fixed.

test/cli/test/stacktraces.test.ts
```diff
@@ -102,6 +102,8 @@ it('stacktrace in vmThreads', async () => {
 })
 
 function removeLines(log: string) {
+  // strip vite oxc transform warning
+  log = log.replace(/target was modified to include ES2021.*\n/g, '')
```
sapphi-red: This warning is now removed.

test/core/vite.config.ts
```diff
@@ -59,7 +59,12 @@ export default defineConfig({
     includeSource: [
       'src/in-source/*.ts',
     ],
-    exclude: ['**/fixtures/**', ...defaultExclude],
+    exclude: [
+      '**/fixtures/**',
+      ...defaultExclude,
+      // FIXME: wait for support in rolldown
+      '**/esnext.test.ts',
```
sapphi-red: If this is for `using` syntax, it can be removed as the support landed.

packages/runner/src/suite.ts
hi-ogawa: What does it break when column is off by 1?
I think the same happens currently when the files are not transformed by esbuild like plain js file. For example, two test files `repro.test.ts` and `repro.test.js` with the same content:

```js
import { test } from "vitest"

test.each([1])("foo", () => {})
//            ^15
//             ^16
```

This is the result of `vitest list --json --includeTaskLocation`

```js
[
  {
    "name": "foo",
    "file": "/home/hiroshi/code/others/vitest/examples/basic/test/repro.test.js",
    "location": {
      "line": 3,
      "column": 16
    }
  },
  {
    "name": "foo",
    "file": "/home/hiroshi/code/others/vitest/examples/basic/test/repro.test.ts",
    "location": {
      "line": 3,
      "column": 15
    }
  }
]
```

If maybe off by +1 is acceptable in practice, then can we keep `each ? stack.column + 1 : stack.column` for now and just deal with `public-api.test.ts` test cases for ecosystem ci by checking `vite.rolldownVersion`?

test/core/vite.config.ts
sheremet-va: I did remove it, but it keeps failing. Do we need to update the target somewhere ourselves?

packages/runner/src/suite.ts
sheremet-va: The `js` seems to be wrong here, actually. There might be other transforms at play, but in a real browser the result column should be `(`. It is probably wrong because the `(` is reported correctly, but we always add +1

sheremet-va: Let's just keep the column there because the issue is with source maps, and not with the runner

sheremet-va:
Seems like there is a small issue with Vue coverage again, could you help with that, @AriPerkkio?

packages/runner/src/suite.ts
hi-ogawa: > The js seems to be wrong here, actually

Yes, that's what I meant. It's currently adding +1 to account for esbuild's source map bug, so doing the same for plain js file makes it off.

AriPerkkio:
> Seems like there is a small issue with Vue coverage again, could you help with that, @AriPerkkio?

Looks like this is failing only on the old `v8` provider. Some generated code is increasing function and branch coverages. I think it's fine to leave as is, as we are planning to remove whole old `v8-to-istanbul` in next major. 

You can just update the snapshots from the tests.

sheremet-va:
rolldown-vite now passes all tests, I am changing the dependency back to 6.3.5 and making it work there 😄

test/core/vite.config.ts
hi-ogawa: `using` is transpiled but decorator is left, so node gets syntax error. I think this is tracked in https://github.com/oxc-project/oxc/issues/9170.

```
 FAIL   forks  test/esnext.test.ts [ test/esnext.test.ts ]
SyntaxError: Invalid or unexpected token
```

<details><summary>transform output</summary>

```js
// /home/hiroshi/code/others/vitest/test/core/test/esnext.test.ts
const __vite_ssr_import_0__ = await __vite_ssr_import__("/@fs/home/hiroshi/code/others/vitest/packages/vitest/dist/index.js", {"importedNames":["expect","it"]});
const __vite_ssr_import_1__ = await __vite_ssr_import__("/@fs/home/hiroshi/code/others/vitest/node_modules/.pnpm/@oxc-project+runtime@0.72.2/node_modules/@oxc-project/runtime/src/helpers/usingCtx.js", {"importedNames":["default"]});
const [version] = process.version.slice(1).split(".");
__vite_ssr_import_0__.it.skipIf(Number(version) < 20)("\"v\" flag in regexp", () => {
	const regexp = new RegExp("\\p{RGI_Emoji}|\\P{Mark}\\p{Mark}*", "gv");
	(0,__vite_ssr_import_0__.expect)("👍🏼👍🏼👍🏼".match(regexp)).toEqual([
		"👍🏼",
		"👍🏼",
		"👍🏼"
	]);
});
(0,__vite_ssr_import_0__.it)("decorators work", () => {
	(0,__vite_ssr_import_0__.expect)(Sut.mocked).toBe(true);
	(0,__vite_ssr_import_0__.expect)(new Sut()).toBeInstanceOf(Sut);
});
(0,__vite_ssr_import_0__.it)("new \"using\" feature", () => {
	let getResource = () => {
		throw new Error("don't call me");
	};
	try {
		var _usingCtx = (0,__vite_ssr_import_1__.default)();
		const resource = _usingCtx.u(resourceful("foo"));
		getResource = () => resource.resource;
	} catch (_) {
		_usingCtx.e = _;
	} finally {
		_usingCtx.d();
	};
	(0,__vite_ssr_import_0__.expect)(getResource()).toBe(null);
});
// @ts-expect-error - readonly symbol, but might not be assigned
Symbol.dispose ??= Symbol("dispose");
function resourceful(resourceDefault) {
	let resource = resourceDefault;
	return {
		get resource() {
			return resource;
		},
		[Symbol.dispose]: () => {
			resource = null;
		}
	};
}
function exampleDecorator(ClassExample, context) {
	if (context.kind !== "class") {
		throw new Error("not a class to decorate");
	};
	ClassExample.mocked = true;
	return ClassExample;
}
@exampleDecorator class Sut {
	static mocked = false;
}

```

</details>

hi-ogawa (approved):
Looks good! I did a minor cleanup of package.json.

AriPerkkio (commented):
We have couple of places where we use `parseAst` and `parseAstAsync` from Vite. Are those now using `oxc-parser` or the SWC parser via rollup?

sapphi-red:
> We have couple of places where we use `parseAst` and `parseAstAsync` from Vite. Are those now using `oxc-parser` or the SWC parser via rollup?

`parseAst` and `parseAstAsync` exposed from rolldown-vite uses Oxc's parser.

sheremet-va:
Should we add a separate CI pipeline for rolldown-vite?

PR #7578, opened 2025-02-27
feat(reporter): add retryCount and flaky to json reporter

saitonakamura:
### Description

I've added additional info about retryCount and flakiness to json reporter to be able to take advantage of it for monitoring tests (we send metrics for each test run to influx database and we would like to monitor retries as well)

I've specifically used the same model as in https://github.com/vitest-dev/vitest/blob/main/packages/vitest/src/node/reporters/reported-tasks.ts#L197 since it looks more useful to me. However jest does a different thing and reports tests invocations https://github.com/jestjs/jest/blob/main/packages/jest-types/src/TestResult.ts#L40. I'm not sure what's better, to be more aligned with jest format (cause initial vitest json report clearly took inspiration from it) or not. So please let me know how I should proceed.

### Tests

### Documentation

### Changesets


packages/vitest/src/node/reporters/json.ts
```diff
@@ -152,6 +160,8 @@ export class JsonReporter implements Reporter {
             t.result?.errors?.map(e => e.stack || e.message) || [],
           location: t.location,
           meta: t.meta,
+          retryCount: t.result?.retryCount ?? 0,
+          flaky: !!t.result?.retryCount && t.result?.state === 'pass' && t.result?.retryCount > 0,
```
AriPerkkio, line 164: We should probably re-use the logic from `diagnostics()` here:

```diff
      const assertionResults = tests.map((t) => {
+       const entity = this.ctx.state.getReportedEntity(t)
+       const diagnostic = entity?.type === 'test' ? entity.diagnostic() : null
+       const flaky = diagnostic?.flaky
+       const retryCount = diagnostic?.retryCount
```

saitonakamura: Thanks for it, fixed

saitonakamura:
I've added some context to the PR and mainly the question about jest report compatibility

sheremet-va:
We think it's better to introduce a new `json` reporter with a different structure that better represents how Vitest handles tests. See https://github.com/vitest-dev/vitest/issues/7586

sheremet-va:
Closing this in favor of #7586

PR #7590, opened 2025-03-01
chore: use pnpm catalog

btea:
### Description
Using catalog makes it easier for multiple packages to manage the same dependencies. https://pnpm.io/catalogs
And renovate also supports catalog. https://github.com/renovatebot/renovate/pull/33376

### Tests

### Documentation

### Changesets


pnpm-workspace.yaml
```diff
@@ -4,3 +4,42 @@ packages:
   - examples/*
   - test/*
   - test/config/fixtures/conditions-pkg
+catalog:
```
AriPerkkio, line 7: Could you sort these alphabetically

btea: Done.

sheremet-va:
Seems like CI fails because `vue` was updated in the lockfile. We need to pin the version or fix the error in coverage

btea:
> Seems like CI fails because `vue` was updated in the lockfile. We need to pin the version or fix the error in coverage

Ok, I have pinned the vue version.

AriPerkkio (changes requested):
We should probably use the same convention in each `/test/*/package.json` too? 

https://github.com/vitest-dev/vitest/blob/0c2924b7aeb7eaf1b1fa0e78d2da76a7f3197b50/test/core/package.json#L1-L42

btea:
Some dependency versions in the test directory are inconsistent with the versions in pnpm-workspace.yaml. If they are changed to the same, it seems that CI will fail. No changes have been made for the time being.

sheremet-va (approved):

AriPerkkio:
Looks like this broke ecosystem-ci: https://github.com/vitest-dev/vitest-ecosystem-ci/actions/runs/13738553952

btea:
It seems that the pnpm version used by `vitest-ecosystem-ci` does not support the catalog protocol yet. 🤔

Issue #7521, opened 2025-02-18
CLI --browser.headless=false doesn't load Vitest UI when using workspaces


silvenon:
### Describe the bug

When using (inline) workspaces the CLI option `--browser.headless=false`  doesn't map to the project's `browser.headless` configuration, it seems to map to the instances instead because my instance opens, but without the Vitest UI, which is how setting `headless: false` on the instance behaves in combination with top-level `headless: true`.

```
npx vitest --project=browser --browser.headless=false
```

```ts
// vite.config.ts

// ...

{
  test: {
    workspace: [
      {
        test: {
          name: 'browser',
          browser: {
            enabled: true,
            headless: true,
            instances: [
              {
                browser: 'chromium',
              },
            ],
          },
        },
      },
    ],
  },
}
```

The CLI option works fine when not using workspaces, in that case the top-level `browser.headless` is successfully set to `false`.

Probably related to #7311.

### Reproduction

https://github.com/silvenon/repro/tree/vitest-browser-mode-headless-workspaces

```
gh repo clone silvenon/repro -- --branch vitest-browser-mode-headless-workspaces
cd repro
npm install
npm test
```

### System Info

```shell
System:
    OS: macOS 15.3.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 233.84 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.18.1 - ~/.n/bin/node
    Yarn: 1.22.22 - /opt/homebrew/bin/yarn
    npm: 10.8.2 - ~/.n/bin/npm
    pnpm: 10.1.0 - /opt/homebrew/bin/pnpm
    bun: 1.1.29 - ~/.bun/bin/bun
  Browsers:
    Chrome: 133.0.6943.99
    Edge: 133.0.3065.69
    Safari: 18.3
```

### Used Package Manager

npm

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vitest.dev/guide/).
- [x] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
[...]

PR #7604, opened 2025-03-04
fix(workspace): forward `--browser.<option>` CLI args to workspace projects

AriPerkkio:
### Description

- Fixes https://github.com/vitest-dev/vitest/issues/7521

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/workspace/resolveWorkspace.ts
+++ b/packages/vitest/src/node/workspace/resolveWorkspace.ts
@@ -48,6 +48,7 @@ export async function resolveWorkspace(
     'inspect',
     'inspectBrk',
     'fileParallelism',
+    'browser',
   ] as const
 
   const cliOverrides = overridesOptions.reduce((acc, name) => {
@@ -69,6 +70,12 @@ export async function resolveWorkspace(
       : options.extends === true
         ? (vitest.vite.config.configFile || false)
         : false
+
+    // Merge --browser.<option> with projects' test.browser options
+    const browser = (cliOptions.browser || options.test?.browser)
+      ? { ...options.test?.browser, ...cliOptions.browser }
+      : undefined
+
     // if `root` is configured, resolve it relative to the workspace file or vite root (like other options)
     // if `root` is not specified, inline configs use the same root as the root project
     const root = options.root
@@ -77,7 +84,7 @@ export async function resolveWorkspace(
     projectPromises.push(concurrent(() => initializeProject(
       index,
       vitest,
-      { ...options, root, configFile, test: { ...options.test, ...cliOverrides } },
+      { ...options, root, configFile, test: { ...options.test, ...cliOverrides, browser } },
     )))
   })
 
@@ -311,8 +318,7 @@ function cloneConfig(project: TestProject, { browser, ...config }: BrowserInstan
       testerHtmlPath: testerHtmlPath ?? currentConfig.testerHtmlPath,
       screenshotDirectory: screenshotDirectory ?? currentConfig.screenshotDirectory,
       screenshotFailures: screenshotFailures ?? currentConfig.screenshotFailures,
-      // TODO: test that CLI arg is preferred over the local config
-      headless: project.vitest._options?.browser?.headless ?? headless ?? currentConfig.headless,
+      headless: headless ?? currentConfig.headless,
       name: browser,
       providerOptions: config,
       instances: undefined, // projects cannot spawn more configs
--- a/test/cli/fixtures/browser-workspace/example.test.ts
+++ b/test/cli/fixtures/browser-workspace/example.test.ts
@@ -1,0 +1,11 @@
+import { expect, test } from 'vitest';
+import { getWorkerState } from '../../../test-utils/runtime';
+
+test('overriden options are found on runtime', () => {
+  const config = getWorkerState().config;
+
+  expect(config.browser.isolate).toBe(false)
+  expect(config.browser.headless).toBe(true)
+  expect(config.browser.name).toBe("chromium")
+
+})
--- a/test/cli/test/cli-options-workspace.test.ts
+++ b/test/cli/test/cli-options-workspace.test.ts
@@ -1,0 +1,15 @@
+import { expect, test } from 'vitest'
+import { runVitestCli } from '../../test-utils'
+
+test('workspace project options are overriden by CLI options', async () => {
+  const { stderr, stdout } = await runVitestCli(
+    { nodeOptions: { cwd: './fixtures/browser-workspace' } },
+    '--run',
+    '--browser.provider=playwright',
+    '--browser.isolate=false',
+  )
+
+  expect(stderr).toBe('')
+  expect(stdout).toContain('✓  browser (chromium)  example.test.ts (1 test)')
+  expect(stdout).toContain('Test Files  1 passed (1)')
+})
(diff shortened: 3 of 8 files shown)


test/cli/test/cli-options-workspace.test.ts
```diff
@@ -0,0 +1,15 @@
+import { expect, test } from 'vitest'
+import { runVitestCli } from '../../test-utils'
+
+test('workspace project options are overriden by CLI options', async () => {
+  const { stderr, stdout } = await runVitestCli(
+    { nodeOptions: { cwd: './fixtures/browser-workspace' } },
+    '--run',
+    '--browser.provider=playwright',
+    '--browser.isolate=false',
+  )
```
AriPerkkio, line 10: I didn't want to test the reported `browser.headless` option here, as it has different resolving logic on CI.  Reverting the fix makes these options fail so it should be equivalent.

packages/vitest/src/node/config/resolveConfig.ts
```diff
@@ -798,7 +798,7 @@ export function resolveConfig(
 
   resolved.browser ??= {} as any
   resolved.browser.enabled ??= false
-  resolved.browser.headless ??= isCI
+  resolved.browser.headless ??= resolved.browser.enabled ? isCI : true
```
sheremet-va, line 801: Why does it matter if it's headless or now when `browser` is disabled?

sheremet-va (approved):

packages/vitest/src/node/config/resolveConfig.ts
AriPerkkio: I think it's some other bug, but something keeps opening browser when this line is reverted. The `browser.ui` is `false` and `browser.enabled` is `false`.

test/cli/fixtures/browser-workspace/example.test.ts
```diff
@@ -0,0 +1,10 @@
+import { expect, test } from 'vitest';
+import { getWorkerState } from 'vitest/src/runtime/utils.js';
```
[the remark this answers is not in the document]
AriPerkkio: Just re-export it there?

This was also used here:

https://github.com/vitest-dev/vitest/blob/eab5c37547b34decc6cb6c3d10bad29693f6604e/test/core/test/vi.spec.ts#L7

https://github.com/vitest-dev/vitest/blob/eab5c37547b34decc6cb6c3d10bad29693f6604e/test/core/test/vi.spec.ts#L180-L182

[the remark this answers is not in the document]
sheremet-va: > Just re-export it there?

Yeah

packages/vitest/src/node/config/resolveConfig.ts
sheremet-va: So does it still open them, but in headless mode? 🤔

AriPerkkio: I don't know what opening in headless mode means. When I'm using `headless: false, ui: false`, no browsers are opened. But without this change, an actual browser opens on desktop.

sheremet-va: > I don't know what opening in headless mode means

It means the browser is opened in the background without a visual shell. I have a suspicious that this change just hides the issue because you can't see the browser window

hi-ogawa (dismissed):

packages/vitest/src/node/workspace/resolveWorkspace.ts
```diff
@@ -67,6 +68,12 @@ export async function resolveWorkspace(
       : options.extends === true
         ? (vitest.vite.config.configFile || false)
         : false
+
+    // Merge --browser.<option> with projects' test.browser options
+    const browser = (cliOptions.browser || options.test?.browser)
+      ? { ...options.test?.browser, ...cliOptions.browser }
+      : undefined
```
sheremet-va, line 75: this will break if config has the browser field

```ts
/// config file
export default { test: { browser } }

/// cli
createVite({
  test: {
    browser: undefined
  }
})
```

it should pass down `{}` instead

should there be a test for this?

sheremet-va (changes requested):

AriPerkkio:
Passing `--browser.<option>` that are intended for workspace projects, accidentally activates root project browser too. We need more logic to prevent that.

test/cli/test/list.test.ts
```diff
@@ -9,7 +9,10 @@ test.each([
   ['--browser.enabled'],
 ])('correctly outputs all tests with args: "%s"', async (...args) => {
   const { stdout, exitCode } = await runVitestCli('list', '-r=./fixtures/list', ...args)
-  expect(stdout).toMatchSnapshot()
+
+  const lines = stdout.split('\n').filter(line => !line.includes('Re-optimizing dependencies because vite config has changed')).join('\n')
```
sheremet-va, line 13: that's a bug, it should not reoptimize dependencies

packages/vitest/src/node/workspace/resolveWorkspace.ts
```diff
@@ -311,8 +318,7 @@ function cloneConfig(project: TestProject, { browser, ...config }: BrowserInstan
       testerHtmlPath: testerHtmlPath ?? currentConfig.testerHtmlPath,
       screenshotDirectory: screenshotDirectory ?? currentConfig.screenshotDirectory,
       screenshotFailures: screenshotFailures ?? currentConfig.screenshotFailures,
-      // TODO: test that CLI arg is preferred over the local config
-      headless: project.vitest._options?.browser?.headless ?? headless ?? currentConfig.headless,
+      headless: headless ?? currentConfig.headless,
```
sheremet-va, line 321: this doesn't seem to be related

AriPerkkio: It is related.

test/cli/test/list.test.ts
AriPerkkio: This is on `main` too - I can revert this test fix from this PR.

sheremet-va: If it's on main, then there is a bug on main. Hiding the bug is not the solution

packages/vitest/src/node/workspace/resolveWorkspace.ts
sheremet-va: I don't see how it is related. Wouldn't this actually break the use case and you won't be able to override the headless mode?

test/cli/fixtures/browser-workspace/example.test.ts
```diff
@@ -0,0 +1,11 @@
+import { expect, test } from 'vitest';
+import { getWorkerState } from '../../../test-utils/runtime';
+
+test('overriden options are found on runtime', () => {
+  const config = getWorkerState().config;
+
+  expect(config.browser.isolate).toBe(false)
+  expect(config.browser.headless).toBe(true)
```
sheremet-va, line 8: this option is not overriden like the test assumes

test/cli/fixtures/browser-workspace/example.test.ts
```diff
@@ -0,0 +1,11 @@
...
+  expect(config.browser.name).toBe("chromium")
```
sheremet-va, line 9: this option is not overriden like the test assumes (according to the test name)

AriPerkkio:
Closing this PR as it seems too complex. The issue https://github.com/vitest-dev/vitest/issues/7521 has good repro that demonstrates weird behaviour when CLI flags are used with workspace+browser mode.

PR #7608, opened 2025-03-04
Mark `skip` as never returning

sirlancelot:
Since `skip` throws an error, its return type should be set to `never` so that TypeScript & linters can determine that this code path ends.

### Tests

### Documentation

### Changesets


AriPerkkio (commented):
How does this change come up in practice when writing test cases?

sirlancelot:
When I was writing a test that used `skip()` I didn't realize that it would terminate that code path, so I instead wrote `return skip()` to indicate that the code path is done. With a type signature of `never` I (as well TypeScript) would've known this already.

AriPerkkio (approved):

PR #7618, opened 2025-03-06
test: add test to monitor vitest exports

sheremet-va:

### Tests

### Documentation

### Changesets


test/core/test/exports.test.ts
```diff
@@ -0,0 +1,205 @@
...
+        ".": {
+          "afterAll": "function",
+          "afterEach": "function",
+          "assert": "function",
+          "assertType": "function",
+          "beforeAll": "function",
+          "beforeEach": "function",
+          "bench": "function",
+          "chai": "object",
+          "createExpect": "function",
```
sheremet-va, line 23: should this one be exposed?

test/core/test/exports.test.ts
```diff
@@ -0,0 +1,205 @@
...
+          "describe": "function",
+          "expect": "function",
+          "expectTypeOf": "function",
+          "getRunningMode": "function",
```
sheremet-va, line 27: this one is not documented, alongside `runOnce`, `isFirstRun` and `isWatchMode` - should we get rid of them?

test/core/test/exports.test.ts
```diff
@@ -0,0 +1,205 @@
...
+          "getBenchFn": "function",
+          "getBenchOptions": "function",
+          "getCurrentSuite": "function",
+          "getCurrentTest": "function",
+          "getFn": "function",
+          "getHooks": "function",
+          "setFn": "function",
+          "setHooks": "function",
+        },
+        "./utils": {
```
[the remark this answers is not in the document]
AriPerkkio: We should remove the unnecessary ones. It might also speed up the build times as internal ones don't need type declarations.

test/core/test/exports.test.ts
```diff
@@ -0,0 +1,205 @@
...
+          "should": "function",
+          "suite": "function",
+          "test": "function",
+          "vi": "object",
+          "vitest": "object",
+        },
+        "./browser": {
+          "SpyModule": "object",
+          "collectTests": "function",
+          "getCoverageProvider": "function",
```
AriPerkkio, line 44: This shouldn't be exposed. It's only used on main thread / node side.

sheremet-va: Seems like this also leads to types leaking to `@vitest/browser` in local dev (https://github.com/vitest-dev/vitest/pull/7619)

test/core/test/exports.test.ts
sheremet-va: let's remove it

test/core/test/exports.test.ts
sheremet-va: let's remove it

test/core/test/exports.test.ts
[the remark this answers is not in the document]
sheremet-va: let's remove the whole entry point

packages/vitest/package.json
```diff
@@ -74,10 +74,6 @@
       "types": "./dist/environments.d.ts",
       "default": "./dist/environments.js"
     },
-    "./utils": {
-      "types": "./dist/utils.d.ts",
-      "default": "./dist/utils.js"
-    },
```
remcohaszing, line 80: This is a breaking change. This may or may not be intentional, but I want to make sure you’re aware. I had to make code changes because of this in https://github.com/remcohaszing/monaco-languageserver-types/commit/7183fd50eabd4a34de780bb11c6bfb11a51eca5d.

sheremet-va: It was never documented, so we don't consider it a breaking change

PR #7627, opened 2025-03-07
refactor: coverage types to avoid mixing node and runtime types

sheremet-va:
### Description

Cherry-picked from https://github.com/vitest-dev/vitest/pull/7619

I think we need to refactor the whole coverage import process - split it into separate entry points, maybe. But for now this should keep node types from leaking into `@vitest/browser`

### Tests

### Documentation

### Changesets


AriPerkkio (approved):
Looks good, thanks for the refactoring.

I do agree we should refactor these into clear "main thread entrypoint" and "runtime entrypoint". The current architecture is as old as https://github.com/vitest-dev/vitest/pull/1676, and it hasn't been improved since.

PR #7636, opened 2025-03-09
chore: revert "chore: use pnpm catalog (#7590)"

AriPerkkio:
### Description

- Reverts https://github.com/vitest-dev/vitest/pull/7590 as it accidentally broke vitest-ecosystem-ci
- Upstream issues https://github.com/pnpm/pnpm/issues/8999

I tested to use `link:` instead of `file:` protocol for local overrides but Vitest's browser mode has never worked locally with that.

- https://github.com/vitest-dev/vitest-ecosystem-ci/pull/32

### Tests

### Documentation

### Changesets


AriPerkkio:
Let's first try if using locally packed packages works: https://github.com/vitest-dev/vitest-ecosystem-ci/pull/32#issuecomment-2708774896

gluxon:
Hey @AriPerkkio @btea — I'm taking a look at the [ecosystem CI failures here](https://github.com/vitest-dev/vitest-ecosystem-ci/actions/runs/13738553952/job/38425120826). Was directed here from @btea's message at https://github.com/pnpm/pnpm/issues/7072#issuecomment-2708761003.

```
  /home/runner/work/vitest-ecosystem-ci/vitest-ecosystem-ci/workspace/nuxt/nuxt/packages/kit:
   ERR_PNPM_SPEC_NOT_SUPPORTED_BY_ANY_RESOLVER  chai@catalog: isn't supported by any available resolver.
  
  This error happened while installing the dependencies of vitest@3.0.8
  
  An external package outside of the pnpm workspace declared a dependency using
  the catalog protocol. This is likely a bug in that external package. Only
  packages within the pnpm workspace may use catalogs. Usages of the catalog
  protocol are replaced with real specifiers on 'pnpm publish'.
  
  This is likely a bug in the publishing automation of this package. Consider filing
  a bug with the authors of:
  
    vitest@3.0.8
```

---

I see `vitest` packages are set up in each repo test is through the `file:` protocol.

https://github.com/vitest-dev/vitest-ecosystem-ci/blob/195b3bbcfb43c9b3aa05d556b9af9f54aae7e8a7/utils.ts#L272

```ts
overrides.vitest ||= `${options.vitestPath}/packages/vitest`
overrides['vite-node'] ||= `${options.vitestPath}/packages/vite-node`
```

https://github.com/vitest-dev/vitest-ecosystem-ci/blob/195b3bbcfb43c9b3aa05d556b9af9f54aae7e8a7/utils.ts#L476

```ts
overrides = Object.fromEntries(
	Object.entries(overrides)
		//eslint-disable-next-line @typescript-eslint/no-unused-vars
		.filter(([key, value]) => typeof value === 'string')
		.map(([key, value]) => [key, useFileProtocol(value as string)]),
)
```

Usages of the `catalog:` protocol tend not to work cross-repo in this way. Generally the `catalog:` protocol is replaced at `pnpm publish` time. It's similar to how the `workspace:` protocol doesn't work across different repositories. **I think you would have the same issue if any vitest packages use the `workspace:` protocol.**

A few options.

- You could try running `pnpm pack` on each `vitest` package when linking them in `vitest-ecosystem-ci`. The resulting `.tgz` would be more similar to what the publicly published `vitest` packages would look like. Both the `catalog:` and `workspace:` protocols would be replaced in the `package.json` for each `vitest` package.
- You could try using `link:packages/vitest` instead of `file:packages/vitest`. With the `link:` protocol, pnpm won't try to read the dependencies of the linked package and install them itself. But you'll have to run `pnpm install` in the cloned vitest repo in order to populate its `node_modules`.

gluxon:
I see I ended up making the same suggestion as @hi-ogawa. Nice. 😊

- https://github.com/vitest-dev/vitest-ecosystem-ci/pull/32#issuecomment-2708774896

AriPerkkio:
Resolved by https://github.com/vitest-dev/vitest-ecosystem-ci/pull/32.

PR #7659, opened 2025-03-11
feat(vitest): allow conditional `context.skip(boolean)`

sheremet-va:
### Description

I was writing tests and notice that it would be nice to have this during the test in case users provide `text.extend` overrides and can't use `test.skipIf()` during collection:

```ts
test('some test', { custom, skip } => {
  skip(custom === 'value', 'not supported')
})
```

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/runner/src/context.ts
+++ b/packages/runner/src/context.ts
@@ -104,10 +104,18 @@ export function createTestContext(
 
   context.task = test
 
-  context.skip = (note?: string) => {
+  context.skip = (condition?: boolean | string, note?: string): never => {
+    if (typeof condition === 'boolean' && !condition) {
+      // do nothing
+      return undefined as never
+    }
     test.result ??= { state: 'skip' }
     test.result.pending = true
-    throw new PendingError('test is skipped; abort execution', test, note)
+    throw new PendingError(
+      'test is skipped; abort execution',
+      test,
+      typeof condition === 'string' ? condition : note,
+    )
   }
 
   context.onTestFailed = (handler, timeout) => {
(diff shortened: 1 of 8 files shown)


packages/runner/src/context.ts
```diff
@@ -104,10 +104,18 @@ export function createTestContext(
 
   context.task = test
 
-  context.skip = (note?: string) => {
+  context.skip = (condition?: boolean | string, note?: string): never => {
+    if (typeof condition === 'boolean' && !condition) {
```
AriPerkkio, line 108: Isn't this equivalent to:

```suggestion
    if (condition === false) {
```

AriPerkkio (commented):
So is this any different from following:

```diff
test('skipped test', (context) => {
-  Math.random() < 0.5 && context.skip('optional message')
+  context.skip(Math.random() < 0.5, 'optional message')
  // Test skipped, no error
  assert.equal(Math.sqrt(4), 3)
})
```

sheremet-va:
> So is this any different from following:

It is not different, just a syntactic sugar. See playwright for example: https://playwright.dev/docs/api/class-testinfo#test-info-skip-2

We already support `test.skipIf` instead of asking people to use `if(condition) { test }`, so I don't see any issue here 😄

AriPerkkio:
If playwright has it, sounds good to me.

> We already support `test.skipIf` instead of asking people to use `if(condition) { test }`, so I don't see any issue here 😄

That would not be collected as skipped test though, so not equivalent.

hi-ogawa (approved):
Yeah, I've seen on playwright before and I thought why not :+1:

PR #7695, opened 2025-03-18
fix(reporter): `--hideSkippedTests` should hide suites too

AriPerkkio:
### Description

- Skipped suites should be hidden when `--hideSkippedTests` is used.

Noticed this verbose noise in my project even when using `hideSkippedTests: true`:

With this fix it becomes much easier to read:

### Tests

### Documentation

### Changesets


hi-ogawa (approved):
Thanks!
Btw, I haven't notice the issue with the default reporter, so is this only for verbose reporter?

AriPerkkio:
Yes, I think this affects only verbose reporter. The default/base reporter renders tests and their suites on the same line, so hiding affects whole row.

This issue came up only after https://github.com/vitest-dev/vitest/pull/7503.

Issue #6157, opened 2024-07-17
Scoped Fixtures


V-Mann-Nick:
### Clear and concise description of the problem

When setting up a test, fixtures can have a different scope. Suppose we have two fixtures:
1) `db` - connection to a database possibly with migrations and an initial seed (expensive setup) global to the test module at hand (`beforeAll`/`afterAll`)
2) `user` - a user created in the db which should be setup and torn down on every test (`beforeEach`/`afterEach`)

Scope is also [reflected in vitest-fixture](https://github.com/larsthorup/vitest-fixture/blob/main/doc/api.md#fixture-scope). I don't know the exact relation to this project, but I was very happy to see that it was incorporated into `vitest` directly. But it does seem that there's no feature parity with `vitest-fixture` as I can't find a way to scope fixtures to the `worker`.

This makes it really hard to use if you have an expensive setup like described above that you want to only setup and tear down once per test module.

### Suggested solution

I think the original [vitest-fixture had a good approach](https://github.com/larsthorup/vitest-fixture/blob/main/doc/api.md#fixture-scope) at least in terms of API:

```js
import { test as base } from "vitest-fixture";

export const test = base.extend<{}, {resource: string}>({
  resource: [async ({}, use) => {
    const resource = ...; // setup resource
    await use(resource, async () => {
      ...; // teardown resource
    });
  }, { scope: "worker" }],
});
```

**I also would be interested in contributing this feature.**

### Additional context

There exists a similar issue with an overlapping scope:
https://github.com/vitest-dev/vitest/issues/4953

@jakebiesinger-storyhealth writes:
> Playwright has a notion of [automatic fixtures](https://playwright.dev/docs/test-fixtures#automatic-fixtures), where the fixture is executed even if it is not part of the fixture dependency list in a test. This enables test writers to implicitly set up fixtures that are common to all tests of a type.

He writes about "common to all tests". While the `{ auto: true }` option did solve this problem, it does not solve the issue that these fixtures are setup and torn down for each test (`beforeEach`/`afterEach`) which can potentially be costly.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
[...]

PR #7704, opened 2025-03-19
feat(vitest): allow per-file and per-worker fixtures

sheremet-va:
### Description

Closes https://github.com/vitest-dev/vitest/issues/6157

Test plan:
1. the file fixture is provided as a primitive value and is accessible by tests
2. the file fixture is provided as a factory and is initialised once in all suites, teardown is called once
3. the file fixture is provided as a factory and is initialised once per file when isolation is disabled, teardown is called once per file
4. the worker fixture is provided as a primitive and is accessible by tests
5. the worker fixture is provided as a factory and is initialised once in all suites, teardown is called once by default
6. the worker fixture is provided as a factory and initialised once per worker when isolation is disabled, teardown is called once per file in `threads`
7. the worker fixture is provided as a factory and initialised once _per file_ when isolation is disabled in `vmThreads`, teardown is called by the same rules
8. worker fixture cannot import from the file or test fixture except static
9. file fixture cannot import from test fixture except static
10. file fixture with auto is initialised always on the first test
11. worker fixture with auto is initialised always on the first test
12. file and worker fixtures are available in beforeEach and afterEach

### Tests

### Documentation

### Changesets


sheremet-va:
@AriPerkkio can you help me with the worker teardown? I cannot make it work at all. `beforeExit` is not called, and calling `cleanup` before `recycleWorkers` kills Vitest process 🤔

AriPerkkio (commented):
> `beforeExit` is not called, and calling `cleanup` before `recycleWorkers` kills Vitest process 🤔

Is there a test that can be used to reproduce this?

packages/vitest/src/node/pools/forks.ts
```diff
@@ -280,10 +284,23 @@ export function createForksPool(
...
   }
 
+  async function cleanupWorkers() {
+    const emitter = new EventEmitter()
+    const events = { message: 'message', response: 'response' }
+    const channel: TinypoolChannel = {
+      onMessage: callback => emitter.on(events.message, callback),
+      postMessage: message => emitter.emit(events.response, message),
+    }
+    await pool.run({}, { name: 'cleanup', channel })
```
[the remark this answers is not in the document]
sheremet-va: We want to run them before the worker is terminated. I don't think it makes sense to have two implementations for isolate/no-isolate. We can't run them in `execute` because that's now where we would run them in non-isolated mode

sheremet-va:
> > `beforeExit` is not called, and calling `cleanup` before `recycleWorkers` kills Vitest process 🤔
> 
> Is there a test that can be used to reproduce this?

Here is a test file:

https://github.com/vitest-dev/vitest/blob/2cc955786cb5f14530314849d894849de83364b6/test/cli/test/scoped-fixtures.test.ts

`test.skip` are for worker scoped tests

test/cli/test/scoped-fixtures.test.ts
```diff
@@ -0,0 +1,771 @@
...
+    >> fixture | teardown worker | test1"
+  `)
+  expect(tests).toMatchInlineSnapshot(`
+    " ✓ 1-basic.test.ts > test1 <time>
+     ✓ 2-basic.test.ts > test1 <time>"
+  `)
+})
+
+// TODO: no-isolate breaks tests
+test.skip('worker fixture initiates and torn down in different workers', async () => {
```
AriPerkkio, line 289: This is breaking since message ports that are used to communicate in rpc channel are closed before teardown is called. 

- When `isolate: false` is used, teardown is called when `pool.destroy()` is called. 
- When `isolate: true` is called, teardown is called when `pool.run()` resolves.

Rpc cannot send messages after this part:

https://github.com/vitest-dev/vitest/blob/c2106b337b386773f6e6a9f0a64c8049cc540c37/packages/vitest/src/node/pools/threads.ts#L147-L150

packages/vitest/src/node/pools/vmThreads.ts
```diff
@@ -131,33 +131,33 @@ export function createVmThreadsPool(
...
         }
         else {
           throw error
         }
       }
       finally {
-        port.close()
-        workerPort.close()
+        port.unref()
+        workerPort.unref()
```
AriPerkkio, line 154:
```suggestion
        port.close()
        workerPort.close()
```

packages/vitest/src/runtime/cleanup.ts
```diff
@@ -0,0 +1,14 @@
...
+export function addCleanupListener(listener: () => unknown): void {
+  listeners.add(listener)
+}
+
+export function removeCleanupListener(listener: () => unknown): void {
+  listeners.delete(listener)
+}
+
+export async function cleanup(): Promise<void> {
+  const promises = [...listeners].map(l => l())
```
AriPerkkio, line 12:
```suggestion
  const promises = listeners.map(l => l())
```

packages/runner/src/fixture.ts
```diff
@@ -86,6 +93,25 @@ export function mergeContextFixtures<T extends { fixtures?: FixtureItem[] }>(
           ({ prop }) => prop !== fixture.prop && usedProps.includes(prop),
         )
       }
+      // test can access anything, so we ignore it
+      if (fixture.scope !== 'test') {
+        fixture.deps?.forEach((dep) => {
+          if (!dep.isFn) {
+            // non fn fixtures are always resolved an available to anyone
```
AriPerkkio, line 100:
```suggestion
            // non fn fixtures are always resolved and available to anyone
```

packages/vitest/src/runtime/cleanup.ts
AriPerkkio: Ohh its `Set`

sheremet-va: `listeners` is a Set

AriPerkkio (approved):

PR #7707, opened 2025-03-20
fix (ui): rerun individually tests with special chars in name

### LINKED ISSUE #7671: Tests with Special Characters in Test Name Don't Rerun Individually in Vitest UI
### Describe the bug

When using the Vitest UI, tests that have special characters like (), [], or similar in their test name cannot be rerun individually by clicking the "Run" button. Instead, only the "Rerun All" button at the top works. Instead the following error message appears:

```
RERUN  rerun test 
        Test name pattern: /Square root of nine (9)/

No test files found. You can change the file name pattern by pressing "p"

include: **/*.{test,spec}.?(c|m)[jt]s?(x)
exclude:  **/node_modules/**, **/dist/**, **/cypress/**, **/.{idea,git,cache,output,temp}/**, **/{karma,rollup,webpack,vite,vitest,jest,ava,babel,nyc,cypress,tsup,build,eslint,prettier}.config.*
```

This issue seems similar to vitest-dev/vscode#93, where the problem was resolved by escaping testNamePattern using:
> testNamePattern.replace(/[$^+?()[\]]/g, '\\$&')

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-8b9barmz?file=test%2Fbasic.test.ts

### System Info

```shell
System:
    OS: Linux 6.11 Ubuntu 24.04.2 LTS 24.04.2 LTS (Noble Numbat)
  Binaries:
    Node: 23.6.1
    npm: 10.9.2
  Browsers:
    Brave Browser: 134.1.76.73
  npmPackages:
    @vitejs/plugin-react-swc: ^3.8.0 => 3.8.0
    @vitest/browser: ^3.0.8 => 3.0.8
    playwright: ^1.51.0 => 1.51.0
    vite: ^6.2.0 => 6.2.1
    vitest: ^3.0.8 => 3.0.8
    vitest-browser-react: ^0.1.1 => 0.1.1
```

### Used Package Manager

npm

### Validations

DevJoaoLopes:
### Description

Closes #7671 

Fixes behavior where it was not possible to click 'rerun' in the ui in tests with special characters.

### Tests

### Documentation

### Changesets


DevJoaoLopes:
@hi-ogawa For some reason that I haven't discovered yet, the pipeline jobs are breaking (I've even tried to go back to the implementation and it didn't work)

sheremet-va:
> @hi-ogawa For some reason that I haven't discovered yet, the pipeline jobs are breaking (I've even tried to go back to the implementation and it didn't work)

You are adding tests to UI, but don't test them in the actual spec. UI tests expect a specific number of tests.

packages/vitest/src/node/core.ts
```diff
@@ -906,7 +906,17 @@ export class Vitest {
       this.filenamePattern = undefined
     }
 
-    const testNamePattern = pattern ? new RegExp(pattern) : undefined
+    const validatePatternWithReplaceChars = () => {
+      if (!pattern) {
+        return undefined
+      }
+
+      return /[()[\]{}]/.test(pattern)
```
sheremet-va, line 914: 1. This is the correct regexp:

```ts
function escapeRegex(str: string) {
  return str.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
}
```

2. It should be applied on the UI side in case someone wants to use the actual regexp in CLI

sheremet-va (changes requested):

packages/vitest/src/node/core.ts
DevJoaoLopes: Got it, thanks for the help @sheremet-va 

What do you think about the following approach:

Instead of trying to change task.name directly in ExplorerItem.vue, change the parameter passed in rerunTask using a ternary?

```ts
async rerunTask(id: string): Promise<void> {
    const task = this.state.idMap.get(id)
    if (!task) {
      throw new Error(`Task ${id} was not found`)
    }

    const taskNamePattern = this.config.ui ? task.name.replace(/[.*+?^${}()|[\]\\]/g, '\\$&') : task.name

    await this.changeNamePattern(
      taskNamePattern,
      [task.file.filepath],
      'tasks' in task ? 'rerun suite' : 'rerun test',
    )
  }
  ```

DevJoaoLopes:
> > @hi-ogawa For some reason that I haven't discovered yet, the pipeline jobs are breaking (I've even tried to go back to the implementation and it didn't work)
> 
> You are adding tests to UI, but don't test them in the actual spec. UI tests expect a specific number of tests.

@sheremet-va I don't know if I understood very well, should I also change the spec.ts files?

sheremet-va:
> > > @hi-ogawa For some reason that I haven't discovered yet, the pipeline jobs are breaking (I've even tried to go back to the implementation and it didn't work)
> > 
> > 
> > You are adding tests to UI, but don't test them in the actual spec. UI tests expect a specific number of tests.
> 
> @sheremet-va I don't know if I understood very well, should I also change the spec.ts files?

Your spec file doesn't test anything. To test that test is running, you need to write a test that clicks on the run button

packages/vitest/src/node/core.ts
sheremet-va: Why can't we just do it in UI?

DevJoaoLopes:
> > > > @hi-ogawa For some reason that I haven't discovered yet, the pipeline jobs are breaking (I've even tried to go back to the implementation and it didn't work)
> > > 
> > > 
> > > You are adding tests to UI, but don't test them in the actual spec. UI tests expect a specific number of tests.
> > 
> > 
> > @sheremet-va I don't know if I understood very well, should I also change the spec.ts files?
> 
> Your spec file (https://github.com/vitest-dev/vitest/pull/7707/files#diff-5759d01c34f2a4853324b490d41dfd7a0d497eb9ebe315ec5ab1df335964c6a9R10) doesn't test anything.
> 
> To test that regexp is applied correctly, you need to write a test that clicks on the run button

sure... I removed these scenarios (which were not validating anything) and I will rewrite more assertive tests.

packages/vitest/src/node/core.ts
DevJoaoLopes: Done

sheremet-va (approved):

sheremet-va:
Thanks! Seems to work fine

PR #7710, opened 2025-03-20
fix(browser): Prioritise instance headless option over root option (fix #7661)

### LINKED ISSUE #7661: Playwright browser instance `headless` setting is ignored in favour of root setting
### Describe the bug

# TL;DR
Vitest Browser's Playwright Provider does not seem to resolve configuration values for headless mode correctly. The root value is favoured over the instance value.

| `BrowserConfigOptions.headless` Root Value | `BrowserConfigOptions.instances[].headless` Instance Value | Actual Behaviour | Expected Behaviour |
|-------------|-------------------------------------|--------------------------------------------------|---------------------
`true` | `undefined` | Headless | Headless
`undefined` | `true` | Headed | Headless
`true` | `false` | Headless | Headed

# Detailed Issue

Here is a more detailed description of the issue.

## Headless Configuration Settings

In `vitest.config.ts`, the `headless` option can be set in two different places.

1. In [`BrowserConfigOptions.headless`](https://github.com/vitest-dev/vitest/blob/470cbec1f91bd3cb0aa604077fa288c4a6e1c2b9/packages/vitest/src/node/types/browser.ts#L123-L128), the root value.

```ts
export default defineConfig({
  test: {
    browser: {
      enabled: true,
      provider: "playwright",

      headless: true,
    },
  },
});
```

2. In `BrowserConfigOptions.instances[]`, which is an array of [`BrowserInstanceOption`](https://github.com/vitest-dev/vitest/blob/470cbec1f91bd3cb0aa604077fa288c4a6e1c2b9/packages/vitest/src/node/types/browser.ts#L71).

```ts
export default defineConfig({
  test: {
    browser: {
      enabled: true,
      provider: "playwright",

      instances: [
        {
          browser: "chromium",
          headless: true,
        },
      ],
    },
  },
});
```

Both of the approaches above will pass type-checking, but only Scenario 1 will launch the browser with headless mode enabled. Scenario 2 will not work as expected, and the tests will run in headed mode.

## Reproduction Steps

Please see https://github.com/delilahw/vitest-playwright-headless-demo for a minimal example. Clone the repository and use `pnpm i` to install dependencies.

Then, you can verify the behaviour by enabling debug logs in Playwright and inspecting the output for the `headless` option. It will be printed after the line containing `"browserType.launch started"`.

```sh
# Launch Scenario 2
$ DEBUG='pw:*' pnpm test:browser -c vitest.config.headless-instance.ts &| tee test-instance.log

$ cat test-instance.log | grep -A3 'browserType.launch started'
# Notice the `headless` option being set to false in the below JSON.
2025-03-12T04:04:06.823Z pw:api => browserType.launch started
2025-03-12T04:04:06.823Z pw:channel SEND> {"id":1,"guid":"browser-type@6b4a60839944b8891fb59ce7131b8330","method":"launch","params":{"args":["--start-maximized"],"ignoreAllDefaultArgs":false,"headless":false}}

# Browser command/args and its PID will also be printed here.
# Notice that `--headless` is not present in the below args.
2025-03-12T04:04:06.826Z pw:browser <launching> ms-playwright/chromium-1161/chrome-mac/Chromium.app/Contents/MacOS/Chromium --disable-field-trial-config --disable-background-networking --disable-background-timer-throttling --disable-backgrounding-occluded-windows --disable-back-forward-cache --disable-breakpad --disable-client-side-phishing-detection --disable-component-extensions-with-background-pages --disable-component-update --no-default-browser-check --disable-default-apps --disable-dev-shm-usage --disable-extensions --disable-features=AcceptCHFrame,AutoExpandDetailsElement,AvoidUnnecessaryBeforeUnloadCheckSync,CertificateTransparencyComponentUpdater,DeferRendererTasksAfterInput,DestroyProfileOnBrowserClose,DialMediaRouteProvider,ExtensionManifestV2Disabled,GlobalMediaControls,HttpsUpgrades,ImprovedCookieControls,LazyFrameLoading,LensOverlay,MediaRouter,PaintHolding,ThirdPartyStoragePartitioning,Translate --allow-pre-commit-input --disable-hang-monitor --disable-ipc-flooding-protection --disable-popup-blocking --disable-prompt-on-repost --disable-renderer-backgrounding --force-color-profile=srgb --metrics-recording-only --no-first-run --enable-automation --password-store=basic --use-mock-keychain --no-service-autorun --export-tagged-pdf --disable-search-engine-choice-screen --unsafely-disable-devtools-self-xss-warnings --enable-use-zoom-for-dsf=false --no-sandbox --start-maximized --user-data-dir=/tmp/1/playwright_chromiumdev_profile --remote-debugging-pipe --no-startup-window
2025-03-12T04:04:06.827Z pw:browser <launched> pid=88938
```

You can check the `headless: true | false` output for various scenarios by swapping out the `-c <config file>` argument with the different config files below.
| Config File | `BrowserConfigOptions.headless` Root Value | `BrowserConfigOptions.instances[].headless` Value | Actual Behaviour | Expected Behaviour | Notes |
|-------------|-------------------------------------|--------------------------------------------------|---------------------|--------------------|-|
[vitest.config.headless-browser.ts](https://github.com/delilahw/vitest-playwright-headless-demo/blob/main/vitest.config.headless-browser.ts) | `true` | `undefined` | Headless | Headless | Scenario 1
[vitest.config.headless-instance.ts](https://github.com/delilahw/vitest-playwright-headless-demo/blob/main/vitest.config.headless-instance.ts) | `undefined` | `true` | Headed | Headless | Scenario 2
[vitest.config.headless-conflicting-values.ts](https://github.com/delilahw/vitest-playwright-headless-demo/blob/main/vitest.config.headless-conflicting-values.ts) | `true` | `false` | Headless | Headed | -

## Cause

What's causing this discrepancy? Simply put, the `headless` option in `BrowserConfigOptions.instances[]` is not being used in the Playwright Provider.

Let's take a look at [`vitest/packages/browser/src/node/providers/playwright.ts:47-54`](https://github.com/vitest-dev/vitest/blob/470cbec1f91bd3cb0aa604077fa288c4a6e1c2b9/packages/browser/src/node/providers/playwright.ts#L47-L54), containing the following constructor.

```ts
export class PlaywrightBrowserProvider implements BrowserProvider {
  initialize(
    project: TestProject,
    { browser, options }: PlaywrightProviderOptions,
  ): void {
    this.project = project
    this.browserName = browser
    this.options = options as any
  }
}
```

The `this.options.headless` property will be populated with `BrowserConfigOptions.instances[].headless`. However, it isn't consumed.

Let's take a look at [Lines 66-73](https://github.com/vitest-dev/vitest/blob/470cbec1f91bd3cb0aa604077fa288c4a6e1c2b9/packages/browser/src/node/providers/playwright.ts#L66-L73), within the `openBrowser()` method. This snippet determines the final headless setting.

```ts
private async openBrowser() {
  ...

  const options = this.project.config.browser

  const launchOptions = {
    ...this.options?.launch,
    headless: options.headless,
  } satisfies LaunchOptions

  ...
}
```

We can see that the `headless` option is set to `options.headless`, which is populated from `this.project.config.browser.headless`, which itself is populated from `BrowserConfigOptions.headless`.

There is no usage of `this.options.headless`, so the value from `BrowserConfigOptions.instances[].headless` is ignored.

### A. Use `BrowserConfigOptions.headless` only

If the preferred behaviour is to always use the value from `BrowserConfigOptions.headless`, then the current implementation is correct.

In this case, we should remove the `headless` option from the [`BrowserInstanceOption` type](https://github.com/vitest-dev/vitest/blob/470cbec1f91bd3cb0aa604077fa288c4a6e1c2b9/packages/vitest/src/node/types/browser.ts#L71). This would make the following configuration invalid:

```ts
export default defineConfig({
  test: {
    browser: {
      enabled: true,
      provider: "playwright",

      instances: [
        {
          browser: "chromium",
          headless: true, // Invalid
        },
      ],
    },
  },
});
```

### B. `BrowserInstanceOption.headless` takes precedence over `BrowserConfigOptions.headless`

According to the [Browser Config Reference Docs](https://github.com/vitest-dev/vitest/blob/470cbec1f91bd3cb0aa604077fa288c4a6e1c2b9/docs/guide/browser/config.md?plain=1#L67-L88), “every browser config inherits from the root config”.

We should replace the resolution logic with

```ts
const options = this.project.config.browser

// BrowserConfigOptions.headless
const headlessFromRoot = options?.headless

// BrowserConfigOptions.instances[].headless
const headlessFromInstance = this.options?.headless

const launchOptions = {
  ...this.options?.launch,
  // Prefer headlessFromInstance if it's set
  headless: headlessFromInstance ?? headlessFromRoot,
} satisfies LaunchOptions
```

### Other Solutions

If you have any other ideas, please feel free to discuss!

I'm happy to implement a solution and make a PR. Let me know what we wanna do. 😇

### Reproduction

https://github.com/[delilahw/vitest-playwright-headless-demo](https://github.com/delilahw/vitest-playwright-headless-demo)

### System Info

```shell
System:
    OS: macOS 15.3.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 105.13 MB / 64.00 GB
    Shell: 3.7.1 - /Users/delilahmsft/.nix-profile/bin/fish
  Binaries:
    Node: 22.14.0 - ~/.n/bin/node
    Yarn: 4.6.0 - ~/.npm/bin/yarn
    npm: 10.9.2 - ~/.n/bin/npm
    pnpm: 9.12.3 - ~/.npm/bin/pnpm
  Browsers:
    Chrome: 134.0.6998.45
    Edge: 134.0.3124.51
    Safari: 18.3
  npmPackages:
    @vitest/browser: ^3.0.8 => 3.0.8 
    playwright: ^1.51.0 => 1.51.0 
    vitest: ^3.0.8 => 3.0.8
```

### Used Package Manager

pnpm

### Validations

delilahw:
### Description

Fixes #7661.

Currently, project/workspace resolution will generate a child project for each browser instance. However, the headless config value is incorrectly resolved because it does not prioritise the instance's `headless` value over the root value:
https://github.com/vitest-dev/vitest/blob/63949b1d735914dbccfc8d56ac1a4fdc2b3f1553/packages/vitest/src/node/workspace/resolveWorkspace.ts#L315

The logic should instead prioritise the instance's `headless` value over the root value. So we need to swap the order of the first two values:
```ts
headless: headless ?? project.vitest._options?.browser?.headless ?? currentConfig.headless,
```

#### Changes
 - Add tests to prevent regressions
 - Fix workspace config resolution logic for headless value

### Tests
Use [headless-config-inheritance.test.ts](https://github.com/delilahw/vitest/blob/27bdc767b7f784a2199c79521609d968961535c2/test/browser/specs/headless-config-inheritance.test.ts#L1) with various providers to demonstrate the issue. It will fail without the fix in 27bdc767b7f784a2199c79521609d968961535c2.

```sh
$ gh pr checkout 7710

# Revert the fix, but keep the tests
$ git revert -n 27bdc767b7f784a2199c79521609d968961535c2

# Tests will fail
$ cd test/browser
$ pnpm test:webdriverio ./specs/headless-config-inheritance.test.ts
$ pnpm test:safaridriver ./specs/headless-config-inheritance.test.ts
$ pnpm test:playwright ./specs/headless-config-inheritance.test.ts

# Apply the fix
$ git revert --abort

# Re-run tests and they will pass
```

### Documentation
No new functionality; inheritance should already be documented as it is the expected behaviour.

### Changesets


delilahw:
Keeping it as draft for now; Still got some todo items on my list 😇

AriPerkkio (commented):
This might be fixed in https://github.com/vitest-dev/vitest/pull/7604/files#r1989283690 too. Noticed similar weirdness with `headless` value there. 🤔

sheremet-va:
> instance's `headless` value over the root value

`project.vitest._options?.browser?.headless` is not the root value, it's supposed to be the CLI value which should have a priority over anything else. Why is it there if CLI wasn't called with `--browser.headless` flag?

test/browser/specs/headless-config-inheritance.test.ts
```diff
@@ -0,0 +1,98 @@
...
+    name: 'both set to the same value',
+    rootHeadless: false,
+    instanceHeadless: false,
+    expectedResolvedInstanceHeadless: false,
+  },
+  {
+    name: 'instance headless overrides root headless',
+    rootHeadless: true,
+    instanceHeadless: false,
+    expectedResolvedInstanceHeadless: false,
```
sheremet-va, line 52: this is not correct within your test. you define headless as a CLI argument, not as a value that will be read from the config

in this case, the value _need_ to be `true`, otherwise it's not possible to override with CLI flags

sheremet-va (changes requested):

test/browser/specs/headless-config-inheritance.test.ts
delilahw: oh ok, will circle back with updated test next week. probably just going to duplicate the `vitest.config.ts` file and specify different paths pointing to `vitest-root-true.config.ts`/`vitest-root-false.config.ts`. unless there is api to override the config during test?

sheremet-va: there is `runInlinedTests` utility that creates them during the test and then deletes the files

delilahw: thanks for the pointer! i've changed the test to use the `runInlineTests` utility in https://github.com/vitest-dev/vitest/pull/7710/commits/2df36e492fdd844598c1e100b09249cab6e6bcc8.

the test isn't passing `headless` as a CLI argument anymore, so i think it's correct now?

delilahw:
> `project.vitest._options?.browser?.headless` is not the root value, it's supposed to be the CLI value which should have a priority over anything else.

@sheremet-va I'm not sure what's going on here, but when I rollback the "fix" (https://github.com/vitest-dev/vitest/pull/7710/commits/27bdc767b7f784a2199c79521609d968961535c2) to reset the order priority back to the version in `main`, then the test starts failing. This is assuming that the new implementation of the test is correct.

test/browser/specs/headless-config-inheritance.test.ts
```diff
@@ -0,0 +1,128 @@
...
+            browser: {
+              enabled: true,
+              provider: '${provider}',
+              headless: ${rootHeadless},
+              instances: [
+                instance,
+              ],
+            },
+          },
+          cacheDir: fileURLToPath(new URL('./node_modules/.vite', import.meta.url)),
```
delilahw, line 87: `cacheDir` probably not needed because `runInlineTests` will place each attempt inside its own folder

sheremet-va:
Superseded by #10851, which adds focused regression coverage for the behavior already fixed by #7858.

Issue #6300, opened 2024-08-07
coverage-v8 issues with calculation of branch coverage


stereobooster:
### Describe the bug

I have following code

```js
export function comp(a, b) {
  if (a === b) {
    return 0
  } else if (a > b) {
    return 1
  } else {
    return -1
  }
}
```

and following test

```js
test("compares", () => {
  expect(comp(1, 1)).toBe(0);
  expect(comp(1, 0)).toBe(1);
});
```

it shows branch coverage 2/3 - which is **correct**. But If I add missing test

```js
expect(comp(0, 1)).toBe(-1);
```

It shows branch coverage 4/4. Which seems to be **incorrect**

If I change code to 

```js
export function comp(a, b) {
  if (a === b) {
    return 0
  } else if (a > b) {
    return 1
  } 
  return -1
}
```

It shows branch coverage 5/5. Which seems to be **incorrect**.

**PS 1** I checked it with istanbul, and istanbul numbers are more consistent. So I guess this is the issue with `c8`

**PS 2** I checked using `c8` (`^10.1.2`) directly and results are different, so there is a bug

### Reproduction

https://github.com/stereobooster/vitest-coverage-v8

### System Info

```shell
System:
    OS: macOS 14.1.1
    CPU: (8) x64 Intel(R) Core(TM) i5-1038NG7 CPU @ 2.00GHz
    Memory: 196.43 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.17.1 - ~/.nvm/versions/node/v18.17.1/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 9.6.7 - ~/.nvm/versions/node/v18.17.1/bin/npm
    pnpm: 9.2.0 - ~/Library/pnpm/pnpm
  Browsers:
    Chrome: 127.0.6533.99
    Safari: 17.1
  npmPackages:
    @vitest/coverage-v8: ^2.0.3 => 2.0.3
    @vitest/ui: ^2.0.3 => 2.0.3
    vite: ^5.3.4 => 5.3.4
    vitest: ^2.0.3 => 2.0.
```

### Used Package Manager

pnpm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
[...]

Issue #5783, opened 2024-05-28
coverage-v8 reports incorrect coverage when using multiple environments


jerrythomas:
### Describe the bug

The following code runs differently in different environments (jsdom/windows)

```js
export const runningOn = typeof window === 'undefined' ? 'server' : 'browser'
```

If I write tests which tests these two paths shouldn't the coverage reported be 100%?

```js
/**
 * @vitest-environment jsdom
 */
describe('Logger on browser', () => {
	it('should identify running_on as "browser"', () => {
		console.log('Running on environment:', runningOn)
		expect(runningOn).toEqual('browser')
	})
})
```

```js
/**
 * @vitest-environment node
 */
describe('Logger on browser', () => {
	it('should identify running_on as "server"', () => {
		console.log('Running on environment:', runningOn)
		expect(runningOn).toEqual('server')
	})
})
```

Actual reported coverage is 
```text
stdout | src/run_on_server.spec.js > Logger on browser > should identify running_on as "server"
Running on environment: server

stdout | src/run_on_browser.spec.js > Logger on browser > should identify running_on as "browser"
Running on environment: browser

 ✓ src/run_on_server.spec.js (1)
 ✓ src/run_on_browser.spec.js (1)

 Test Files  2 passed (2)
      Tests  2 passed (2)
   Start at  11:32:26
   Duration  1.27s (transform 66ms, setup 0ms, collect 31ms, tests 8ms, environment 717ms, prepare 293ms)

 % Coverage report from v8
--------------|---------|----------|---------|---------|-------------------
File          | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
--------------|---------|----------|---------|---------|-------------------
All files     |     100 |        0 |     100 |     100 |                   
 constants.js |     100 |        0 |     100 |     100 | 1                 
--------------|---------|----------|---------|---------|-------------------
```

### Reproduction

[repro-vitest-v8-coverage](https://github.com/jerrythomas/repro-vitest-v8-coverage/blob/main/package.json)

### System Info

```shell
Need to install the following packages:
envinfo@7.13.0
Ok to proceed? (y) y

  System:
    OS: macOS 14.4.1
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 20.21 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
[...]

PR #7736, opened 2025-03-25
feat(coverage): v8 experimental AST-aware remapping

AriPerkkio:
### Description

- Adds support for experimental AST-aware V8 coverage remapping
- Vitest v3 will default to old `v8-to-istanbul`, but switches to `ast-v8-to-istanbul` when `experimentalAstAwareRemapping: true` is set.
- If all goes well, we can remove `v8-to-istanbul` in Vitest v4 (https://github.com/vitest-dev/vitest/issues/7928)
- Right now the `provider.ts` is a bit of mess as it supports both remappers. Once `v8-to-istanbul` can be dropped, it will simplify quite much
- Tests start and run as fast as with previous `@vitest/coverage-v8`. Report accuracy is as good as with `@vitest/coverage-istanbul`. Report generation is slightly slower than with previous `@vitest/coverage-v8`, but faster than overall istanbul provider run. 
- Fixes https://github.com/vitest-dev/vitest/issues/6300
- Fixes https://github.com/vitest-dev/vitest/issues/5783
- Fixes https://github.com/vitest-dev/vitest/issues/7130
- Fixes https://github.com/vitest-dev/vitest/discussions/7587

Testing: 

- `vuejs/core`: 100% match with `@vitest/coverage-istanbul`, except branch coverage is `87.69% 12059/13751` vs `87.46% 12047/13773`. There are couple of uncovered function default arguments that are not reported by V8 at all. These are marked as covered.
- `vitejs/vite`: Close to 100% match with `@vitest/coverage-istanbul`. Differences coming from V8 limitations, e.g. uncovered default parameters.
- Self-test: 100% match https://github.com/AriPerkkio/ast-v8-to-istanbul/pull/31

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/coverage-v8/src/provider.ts
+++ b/packages/coverage-v8/src/provider.ts
@@ -209,19 +209,11 @@ export class V8CoverageProvider extends BaseCoverageProvider<ResolvedCoverageOpt
           transform,
         )
 
-        const converter = v8ToIstanbul(
+        coverageMap.merge(await this.v8ToIstanbul(
           filename.href,
           0,
           sources,
-          undefined,
-          this.options.ignoreEmptyLines,
-        )
-
-        await converter.load()
-
-        try {
-          // Create a made up function to mark whole file as uncovered. Note that this does not exist in source maps.
-          converter.applyCoverage([{
+          [{
             ranges: [
               {
                 startOffset: 0,
@@ -232,13 +224,8 @@ export class V8CoverageProvider extends BaseCoverageProvider<ResolvedCoverageOpt
             isBlockCoverage: true,
             // This is magical value that indicates an empty report: https://github.com/istanbuljs/v8-to-istanbul/blob/fca5e6a9e6ef38a9cdc3a178d5a6cf9ef82e6cab/lib/v8-to-istanbul.js#LL131C40-L131C40
             functionName: '(empty-report)',
-          }])
-        }
-        catch (error) {
-          this.ctx.logger.error(`Failed to convert coverage for uncovered ${filename.href}.\n`, error)
-        }
-
-        coverageMap.merge(converter.toIstanbul())
+          }],
+        ))
 
         if (debug.enabled) {
           clearTimeout(timeout)
@@ -251,6 +238,26 @@ export class V8CoverageProvider extends BaseCoverageProvider<ResolvedCoverageOpt
     }
 
     return coverageMap
+  }
+
+  private async v8ToIstanbul(filename: string, wrapperLength: number, sources: Awaited<ReturnType<typeof this.getSources>>, functions: Profiler.FunctionCoverage[]) {
+    const converter = v8ToIstanbul(
+      filename,
+      wrapperLength,
+      sources,
+      undefined,
+      this.options.ignoreEmptyLines,
+    )
+    await converter.load()
+
+    try {
+      converter.applyCoverage(functions)
+    }
+    catch (error) {
+      this.ctx.logger.error(`Failed to convert coverage for ${filename}.\n`, error)
+    }
+
+    return converter.toIstanbul()
   }
 
   private async getSources<TransformResult extends (FetchResult | Awaited<ReturnType<typeof this.ctx.vitenode.transformRequest>>)>(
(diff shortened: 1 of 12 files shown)


packages/coverage-v8/src/provider.ts
```diff
@@ -396,13 +413,17 @@ function excludeGeneratedCode(
...
+    && !source.match(VITE_IMPORTS_LINE_PATTERN)
     && !source.match(DECORATOR_METADATA_PATTERN)
   ) {
     return map
   }
 
   const trimmed = new MagicString(source)
   trimmed.replaceAll(VITE_EXPORTS_LINE_PATTERN, '\n')
+  trimmed.replaceAll(VITE_IMPORTS_LINE_PATTERN, '')
+  trimmed.replaceAll(VITE_EXPORTS_LINE_PATTERN_2, '')
```
AriPerkkio, line 426: Let's move these to a separate `fix:` PR.

test/coverage-test/test/pre-transpiled-source.test.ts
```diff
@@ -14,13 +14,32 @@ test('pre-transpiled code with source maps to original (#5341)', async () => {
...
+  else {
+    expect(fileCoverage).toMatchInlineSnapshot(`
+      {
+        "branches": "3/6 (50%)",
+        "functions": "2/2 (100%)",
+        "lines": "6/8 (75%)",
+        "statements": "6/8 (75%)",
+      }
+    `)
+  }
```
AriPerkkio, line 44: Let's move these test snapshot changes to separate `test: refactor` PR.

packages/coverage-v8/src/provider.ts
```diff
@@ -218,19 +212,70 @@ export class V8CoverageProvider extends BaseCoverageProvider<ResolvedCoverageOpt
...
+            type === 'statement'
+            && node.type === 'ExpressionStatement'
+            && node.expression.type === 'AssignmentExpression'
+            && node.expression.left.type === 'MemberExpression'
+            && node.expression.left.object.type === 'Identifier'
+            && node.expression.left.object.name === '__vite_ssr_exports__'
+          ) {
+            return true
+          }
+        },
```
AriPerkkio, line 255: @hi-ogawa this approach should provide us an easier way to control how Vite's generated code is excluded from coverage reports. We no longer need to modify source maps - instead we ignore the nodes while remapping coverage.

hi-ogawa (dismissed):
I checked `ast-v8-to-istanbul` too and it looks all good to me. Awesome stuff!

> Report generation is slightly slower than with previous @vitest/coverage-v8, but faster than overall istanbul provider run.

I'm curious how you understand the performance characteristic leading to this result.

I think `ast-v8-to-istanbul` being slower than `v8-to-istanbul` is probably obvious since it now parses and traverse AST and remap node by node while `v8-to-istanbul` just goes through v8 coverage once.

What about `@vitest/coverage-istanbul`? I guess istanbul's instrumentation transform can be slower than `ast-v8-to-istanbul`'s ast manipulation? Also do you take into account instrumented code being running slower? Or maybe other aspect of coverage processing logic?

AriPerkkio:
> I think `ast-v8-to-istanbul` being slower than `v8-to-istanbul` is probably obvious since it now parses and traverse AST and remap node by node while `v8-to-istanbul` just goes through v8 coverage once.

Yes, exactly. It is expected that AST based remapping is slower than the original `v8-to-istanbul`. The original one is very naive and produces many false positives, but it's fast. It's not expected that `ast-v8-to-istanbul` would reach the speed of `v8-to-istanbul`, but I'm hoping we could get very close to that. Both of these tools process a lot of text (JSON, code, source maps) so there's plenty of optimization possibilities. Looking at the API of `ast-v8-to-istanbul`, I think in future all of this could even run on Rust based tooling like oxc. 

> What about `@vitest/coverage-istanbul`? I guess istanbul's instrumentation transform can be slower than `ast-v8-to-istanbul`'s ast manipulation? Also do you take into account instrumented code being running slower? Or maybe other aspect of coverage processing logic?

This speed difference reproduces on `vuejs/core` repository really well. When using Istanbul's pre-instrumented coverage, it's actually visible for end-users to see that files remain in `[queued] filename.test.ts` state for a while when transform is running. Also executing this pre-instrumented code is a lot slower - I think there would be good optimization opportunities in `istanbul-lib-instrumentation`'s code transforms. Pre-instrumenting increases transpiled code and combined source maps a lot. 

When comparing AST parsing, visiting and analysing of `istanbul-lib-instrument` and `ast-v8-to-istanbul`, we should see similar speed results. 

> I'm curious how you understand the performance characteristic leading to this result.

I've been comparing `@vitest/coverage-istanbul`, `@vitest/coverage-v8` and `@vitest/coverage-v8 + ast-v8-to-istanbul` in:
- `vitejs/vite`
- `vuejs/code`
- https://github.com/vitest-tests/coverage-large/pull/1
- https://github.com/AriPerkkio/ast-v8-to-istanbul/pull/31

In these tests I've checked overall speed with `time vitest --run --coverage` and compared coverage summaries between each run. I'm also planning to add debug timer for coverage generation via `DEBUG=vitest:coverage` in separate PR at some point.

sheremet-va (approved):
This looks incredible! Thank you for working on it 🙏🏻

PR #7741, opened 2025-03-26
fix: fix vm tests flakiness

sheremet-va:
### Description

Vitest did not properly cache synthetic modules. After adding the cache, flakiness disappears. I wasn't able to reproduce this with a simple setup, unfortunately, so I am still not sure what exactly caused the crashes to happen.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/runtime/vm/commonjs-executor.ts
+++ b/packages/vitest/src/runtime/vm/commonjs-executor.ts
@@ -1,12 +1,14 @@
 import type { FileMap } from './file-map'
-import type { ImportModuleDynamically, VMModule } from './types'
+import type { ImportModuleDynamically, VMSyntheticModule } from './types'
 import { Module as _Module, createRequire } from 'node:module'
 import vm from 'node:vm'
 import { basename, dirname, extname } from 'pathe'
 import { isNodeBuiltin } from 'vite-node/utils'
+import { interopCommonJsModule, SyntheticModule } from './utils'
 
 interface CommonjsExecutorOptions {
   fileMap: FileMap
+  interopDefault?: boolean
   context: vm.Context
   importModuleDynamically: ImportModuleDynamically
 }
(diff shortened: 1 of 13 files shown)


packages/vitest/src/runtime/vm/esm-executor.ts
```diff
@@ -148,15 +148,17 @@ export class EsmExecutor {
...
       exports.map(({ name }) => name),
-      async () => {
+      async function () {
         const importsObject: WebAssembly.Imports = {}
         for (const { module, name } of imports) {
           if (!importsObject[module]) {
             importsObject[module] = {}
           }
-          await this.evaluateModule(moduleLookup[module])
+          await evaluateModule(moduleLookup[module])
```
sheremet-va, line 161: This one looks suspicious, is async function actually allowed? 🤔

sheremet-va:
> Can we include these tests too now:

For sure!

AriPerkkio (approved):

packages/vitest/src/runtime/vm/commonjs-executor.ts
```diff
@@ -132,7 +137,7 @@ export class CommonjsExecutor {
       static SourceMap = _Module.SourceMap
       static syncBuiltinESMExports = _Module.syncBuiltinESMExports
 
-      static _cache = executor.moduleCache
+      static _cache = executor.builtinCache
```
sheremet-va, line 140: this should be publicRequireCache

packages/vitest/src/runtime/vm/commonjs-executor.ts
```diff
@@ -33,10 +35,13 @@ export class CommonjsExecutor {
 
   private fs: FileMap
   private Module: typeof _Module
+  private cjsNamedExportsMap = new Map<string, Set<string>>()
```
sheremet-va, line 38: this is not needed

packages/vitest/src/runtime/vm/commonjs-executor.ts
```diff
@@ -24,7 +26,7 @@ export class CommonjsExecutor {
   private requireCache = new Map<string, NodeJS.Module>()
```
sheremet-va, line 26: we dont need a private cache, it's harder to keep them in sync

PR #7809, opened 2025-04-08
feat: support custom colors for `test.name`

AriPerkkio:
### Description

- Fixes https://github.com/vitest-dev/vitest/issues/6481

```ts

export default defineConfig({
  test: {
    workspace: [
      {
        test: {
          name: { label: "Unit", color: "green" },
        },
      },
      {
        test: {
          name: { label: "Browser" }, // Uses default color picking
        },
      },
      {
        test: {
          name: "Database", // Uses default color picking
        },
      },
    ],
  },
});
```

_Note: Text color on CLI is terminal specific. Setting both background color and text color doesn't work at least in VSCode terminal. You can set either background or text color, not both._ 

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -146,9 +146,11 @@ When defined, Vitest will run all matched files with `import.meta.vitest` inside
 
 ### name
 
-- **Type:** `string`
+- **Type:** `string | { label: string, color?: LabelColor }`
 
-Assign a custom name to the test project or Vitest process. The name will be visible in the CLI and available in the Node.js API via [`project.name`](/advanced/api/test-project#name).
+Assign a custom name to the test project or Vitest process. The name will be visible in the CLI and UI, and available in the Node.js API via [`project.name`](/advanced/api/test-project#name).
+
+Color used by CLI and UI can be changed by providing an object with `color` property.
 
 ### server {#server}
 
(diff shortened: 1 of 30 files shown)


packages/ui/client/components/FileDetails.vue
```diff
@@ -137,7 +137,7 @@ debouncedWatch(
 )
 
 const projectNameColor = computed(() => {
-  return getProjectNameColor(current.value?.file.projectName)
+  return getProjectNameColor(current.value?.file.projectName, undefined)
```
sheremet-va, line 140: shouldn't we pass down the color somehow?

sheremet-va (changes requested):

packages/vitest/src/node/config/serializeConfig.ts
```diff
@@ -42,7 +42,7 @@ export function serializeConfig(
     retry: config.retry,
     disableConsoleIntercept: config.disableConsoleIntercept,
     root: config.root,
-    name: config.name,
+    name: config.name.name,
```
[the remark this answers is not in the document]
AriPerkkio: I agree, let's use `name.label`. Another option could be `name.text` too. 🤔

packages/ui/client/components/FileDetails.vue
AriPerkkio: Yes we should. I left this intentionally as `undefined` for now, so that I can pick up from here later. My Vue skill are very limited so it wasn't obvious where to make changes.

AriPerkkio: I'm Vue developer now 💯

sheremet-va (dismissed):

PR #7824, opened 2025-04-11
docs: suggest `vitest run` in `--merge-reports` usage

kettanaito:
## The problem

The docs suggest running `vitest --merge-reports` but since the `vitest` command runs test in _watch mode_ by default, you are getting the following error:

```
Error: Cannot merge reports with --watch enabled
```

The solution here is either use explicit `vitest run` or make the `--merge-reports` flag turn off the watch mode implicitly. 

I am generally not a fan of implicit magic, so I'm proposing an easier solution to update the docs so merging the reports works as expected. This is also an incremental solution to the broken suggestion, and it doesn't require any work on Vitest itself.

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/guide/improving-performance.md
+++ b/docs/guide/improving-performance.md
@@ -88,7 +88,7 @@ vitest run --reporter=blob --shard=3/3 # 3rd machine
 Collect the results stored in `.vitest-reports` directory from each machine and merge them with [`--merge-reports`](/guide/cli#merge-reports) option:
 
 ```sh
-vitest --merge-reports
+vitest run --merge-reports
 ```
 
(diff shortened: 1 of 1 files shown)


AriPerkkio (approved):
Makes sense to me. I think we could also even flip this automatically in 

https://github.com/vitest-dev/vitest/blob/5652bf9201c4bc5cfb9d4bc3b0e21b93b3bc58d3/packages/vitest/src/node/config/resolveConfig.ts#L112-L116

kettanaito:
@AriPerkkio, let's update the docs first, then address the config changes.

PR #7833, opened 2025-04-14
docs: improve mocking classes section

rodgobbi:
There are no issues open for this PR, I opened it directly to suggest some changes to the docs, I hope this approach is ok.  

### Description

When I was checking the documentation for mocking classes, I found out that some recommendations are incorrect, and in general, I found it a bit more convoluted than the recommendation from Jest.  
Because of that, I always refer to the [Jest docs for mocking classes](https://jestjs.io/docs/es6-class-mocks).  
I opened this PR to try to improve the docs.

When I mention "convoluted", it's more a matter of taste, and I could remove the changes from this PR if the maintainers don't think it's useful.   
What I changed is removing references to `this` and `prototype` as much as possible, which can be confusing topics in JS in general, and used a more explicit and inline recommendation for mocking classes.

The other part of the PR is fixing some incorrect suggestions in https://github.com/vitest-dev/vitest/blob/main/docs/guide/mocking.md, the following part:

```js
const Dog = vi.fn(function (name) {
  this.name = name
})

// notice that static methods are mocked directly on the function,
// not on the instance of the class
Dog.getType = vi.fn(() => 'mocked animal')

// mock the "speak" and "feed" methods on every instance of a class
// all `new Dog()` instances will inherit these spies
Dog.prototype.speak = vi.fn(() => 'loud bark!')
Dog.prototype.feed = vi.fn()
```

The correct recommendation is that static methods are added to the `prototype` and the instance methods are directly set as properties of the object instance.

Adding a property to the original function and not the `prototype` will not be shared across the class instances, and trying to call it throws an error.

Adding the instance methods to the `prototype` will cause those methods to be shared across all instances, which can cause issues if trying to assert mock calls between different tests.

### DIFF
--- a/docs/guide/mocking.md
+++ b/docs/guide/mocking.md
@@ -703,19 +703,19 @@ test('can feed dogs', () => {
 Now, when we create a new instance of the `Dog` class its `speak` method (alongside `feed` and `greet`) is already mocked:
 
 ```ts
-const dog = new Dog('Cooper')
-dog.speak() // loud bark!
-dog.greet() // Hi! My name is Cooper!
+const Cooper = new Dog('Cooper')
+Cooper.speak() // loud bark!
+Cooper.greet() // Hi! My name is Cooper!
 
 // you can use built-in assertions to check the validity of the call
-expect(dog.speak).toHaveBeenCalled()
-expect(dog.greet).toHaveBeenCalled()
+expect(Cooper.speak).toHaveBeenCalled()
+expect(Cooper.greet).toHaveBeenCalled()
 
-const dog2 = new Dog('Max')
+const Max = new Dog('Max')
 
 // methods assigned to the prototype are shared between instances
-expect(dog2.speak).toHaveBeenCalled()
-expect(dog2.greet).not.toHaveBeenCalled()
+expect(Max.speak).toHaveBeenCalled()
+expect(Max.greet).not.toHaveBeenCalled()
 ```
 
 We can reassign the return value for a specific instance:
--- a/docs/guide/mocking.md
+++ b/docs/guide/mocking.md
@@ -656,7 +656,7 @@ const IncorrectDogClass = vi.fn(name => ({
 const Marti = new CorrectDogClass('Marti')
 const Newt = new IncorrectDogClass('Newt')
 
-Marti instanceof CorrectDogClass // true
+Marti instanceof CorrectDogClass // ✅ true
 Newt instanceof IncorrectDogClass // false!
 ```
 :::
--- a/docs/guide/mocking.md
+++ b/docs/guide/mocking.md
@@ -602,10 +602,15 @@ class Dog {
 
   constructor(name: string) {
     this.name = name
+    this.greet = this.greet.bind(this)
   }
 
   static getType(): string {
     return 'animal'
+  }
+
+  greet(): string {
+    return `Hi! My name is ${this.name}!`
   }
 
   speak(): string {
@@ -622,6 +627,8 @@ We can re-create this class with ES5 functions:
 ```ts
 const Dog = vi.fn(function (name) {
   this.name = name
+  // mock bound methods in the constructor, each instance will have its own spy
+  this.greet = vi.fn(() => `Hi! My name is ${name}!`)
 })
 
 // notice that static methods are mocked directly on the function,
@@ -629,10 +636,30 @@ const Dog = vi.fn(function (name) {
 Dog.getType = vi.fn(() => 'mocked animal')
 
 // mock the "speak" and "feed" methods on every instance of a class
-// all `new Dog()` instances will inherit these spies
+// all `new Dog()` instances will inherit and share these spies
 Dog.prototype.speak = vi.fn(() => 'loud bark!')
 Dog.prototype.feed = vi.fn()
 ```
+
+::: warning
+If a non-primitive is returned from the constructor function, that value will become the result of the new expression. In this case the `[[Prototype]]` may not be correctly bound:
+
+```ts
+const CorrectDogClass = vi.fn(function (name) {
+  this.name = name
+})
+
+const IncorrectDogClass = vi.fn(name => ({
+  name
+}))
+
+const Marti = new CorrectDogClass('Marti')
+const Newt = new IncorrectDogClass('Newt')
+
+Marti instanceof CorrectDogClass // true
+Newt instanceof IncorrectDogClass // false!
+```
+:::
 
 ::: tip WHEN TO USE?
 Generally speaking, you would re-create a class like this inside the module factory if the class is re-exported from another module:
@@ -673,14 +700,22 @@ test('can feed dogs', () => {
 ```
 :::
 
-Now, when we create a new instance of the `Dog` class its `speak` method (alongside `feed`) is already mocked:
+Now, when we create a new instance of the `Dog` class its `speak` method (alongside `feed` and `greet`) is already mocked:
 
 ```ts
 const dog = new Dog('Cooper')
 dog.speak() // loud bark!
+dog.greet() // Hi! My name is Cooper!
 
 // you can use built-in assertions to check the validity of the call
 expect(dog.speak).toHaveBeenCalled()
+expect(dog.greet).toHaveBeenCalled()
+
+const dog2 = new Dog('Max')
+
+// methods assigned to the prototype are shared between instances
+expect(dog2.speak).toHaveBeenCalled()
+expect(dog2.greet).not.toHaveBeenCalled()
 ```
 
 We can reassign the return value for a specific instance:
--- a/docs/guide/mocking.md
+++ b/docs/guide/mocking.md
@@ -620,18 +620,19 @@ class Dog {
 We can re-create this class with ES5 functions:
 
 ```ts
-const Dog = vi.fn(function (name) {
-  this.name = name
-})
+// constructor parameters are passed to the mock implementation
+const Dog = vi.fn(name => ({
+  // assign class fields to the returned object
+  name,
+  // mock the "speak" and "feed" methods on every instance of a class
+  speak: vi.fn(() => 'loud bark!'),
+  feed: vi.fn(),
+}))
 
-// notice that static methods are mocked directly on the function,
+// notice that static methods are mocked directly on the function prototype,
 // not on the instance of the class
-Dog.getType = vi.fn(() => 'mocked animal')
-
-// mock the "speak" and "feed" methods on every instance of a class
-// all `new Dog()` instances will inherit these spies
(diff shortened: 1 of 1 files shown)


sheremet-va:
> When I was checking the documentation for mocking classes, I found out that some recommendations are incorrect, and in general, I found it a bit more convoluted than the recommendation from Jest.

The recommendations in Vitest documentation are more correct than in Jest. Jest examples return an object which is not how classes work in JavaScript. 

This is not a matter of taste, this is how the JS engine works, returning an object is objectively an incorrect thing to do. The only thing in the current documentation that can be improved is moving the `vi.fn()` from `prototype` to `this.speak = vi.fn()` so they are recreated for every instance instead of being shared:

```diff
const Dog = vi.fn(function (name) {
  this.name = name
+ this.speak = vi.fn(() => 'loud bark!')
+ this.feed = vi.fn()
})

// notice that static methods are mocked directly on the function,
// not on the instance of the class
Dog.getType = vi.fn(() => 'mocked animal')

// mock the "speak" and "feed" methods on every instance of a class
// all `new Dog()` instances will inherit these spies
- Dog.prototype.speak = vi.fn(() => 'loud bark!')
- Dog.prototype.feed = vi.fn()
```

The static example that you are proposing is completely incorrect, this is not how static methods work, instances don't have access to a method:

```ts
class Dog {
  static getType() { return 'animal' }
}

const dog = new Dog()
dog.getType() // error! getType is not a function
Dog.getType() // 'animal'
```

There is an example of a class right before the mocking code, you can validate how it works in any JS environment.

docs/guide/mocking.md
```diff
@@ -765,9 +769,12 @@ export class SomeClass {}
...
 
-vi.mock(import('./example.js'), () => {
-  const SomeClass = vi.fn()
-  SomeClass.prototype.someMethod = vi.fn()
+vi.mock('./example.js', () => {
+  const SomeClass = vi.fn(someConstructorParameter => ({
+    someInstanceField: someConstructorParameter,
+    someInstanceMethod: vi.fn(),
+  }))
+  SomeClass.prototype.someStaticMethod = vi.fn()
```
sheremet-va, line 777: this is not a static method

docs/guide/mocking.md
```diff
@@ -620,18 +620,19 @@ class Dog {
...
+// notice that static methods are mocked directly on the function prototype,
 // not on the instance of the class
-Dog.getType = vi.fn(() => 'mocked animal')
-
-// mock the "speak" and "feed" methods on every instance of a class
-// all `new Dog()` instances will inherit these spies
-Dog.prototype.speak = vi.fn(() => 'loud bark!')
-Dog.prototype.feed = vi.fn()
+// all `new Dog()` instances will inherit and share this mock
+Dog.prototype.getType = vi.fn(() => 'mocked animal')
```
sheremet-va, line 635: methods on a prototype are not static

docs/guide/mocking.md
```diff
@@ -620,18 +620,19 @@ class Dog {
 We can re-create this class with ES5 functions:
 
 ```ts
-const Dog = vi.fn(function (name) {
-  this.name = name
-})
+// constructor parameters are passed to the mock implementation
+const Dog = vi.fn(name => ({
```
sheremet-va, line 624: returning an object is not a correct thing to do

docs/guide/mocking.md
```diff
@@ -640,12 +641,13 @@ Generally speaking, you would re-create a class like this inside the module fact
...
 import { Dog } from './dog.js'
 
-vi.mock(import('./dog.js'), () => {
-  const Dog = vi.fn()
-  Dog.prototype.feed = vi.fn()
-  // ... other mocks
-  return { Dog }
-})
+vi.mock('./dog.js', () => ({
+  Dog: vi.fn(name => ({
```
sheremet-va, line 645: This is not the correct thing to do

docs/guide/mocking.md
```diff
@@ -659,8 +661,10 @@ function feed(dog: Dog) {
 import { expect, test, vi } from 'vitest'
 import { feed } from '../src/feed.js'
 
-const Dog = vi.fn()
-Dog.prototype.feed = vi.fn()
+const Dog = vi.fn(name => ({
+  name,
```
sheremet-va, line 665: This is not how methods on a class work

sheremet-va (changes requested):

rodgobbi:
Omg, sorry about the confusion, thanks @sheremet-va for clarifying!

I confused the methods in the prototype with static methods, sorry for that.

Indeed, Vitest is more correct than Jest, and I see now it's a bit more convoluted to properly mock a class.
It's simpler to do inline and may work for most use cases, but eventually can cause issues.
Thanks for the `instanceof` example.

> This is not a matter of taste, this is how the JS engine works, returning an object is objectively an incorrect thing to do. The only thing in the current documentation that can be improved is moving the vi.fn() from prototype to this.speak = vi.fn() so they are recreated for every instance instead of being shared:

Does it make sense to explicitly say one case is mocking a method bound to the instance and leave one example for a prototype method (which are all the examples in the current docs)?

sheremet-va:
> Does it make sense to explicitly say one case is mocking a method bound to the instance and leave one example for a prototype method (which are all the examples in the current docs)?

Sure, more examples are always better. We can even show the pitfalls of returning an object

rodgobbi:
> Sure, more examples are always better. We can even show the pitfalls of returning an object

Good point, I fell into this trap.
Ok, I'll update the PR and write such pitfall example, then request review. 👍

docs/guide/mocking.md
```diff
@@ -622,18 +627,40 @@ We can re-create this class with ES5 functions:
...
 
 // mock the "speak" and "feed" methods on every instance of a class
-// all `new Dog()` instances will inherit these spies
+// all `new Dog()` instances will inherit and share these spies
 Dog.prototype.speak = vi.fn(() => 'loud bark!')
 Dog.prototype.feed = vi.fn()
 ```
 
+::: warning
+If a non-primitive is returned from the constructor function, that value will become the result of the new expression. In this case the `[[Prototype]]` may not be correctly bound:
```
rodgobbi, line 645: @sheremet-va 
I just copied and pasted this note from the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Inheritance_and_the_prototype_chain#constructors), let me know if it's helpful to also add a link to this docs

docs/guide/mocking.md
```diff
@@ -602,12 +602,17 @@ class Dog {
 
   constructor(name: string) {
     this.name = name
+    this.greet = this.greet.bind(this)
```
sheremet-va, line 605: Why do we need this in the example? Doesn't it already work like this?

docs/guide/mocking.md
```diff
@@ -622,18 +627,40 @@ We can re-create this class with ES5 functions:
...
+
+const IncorrectDogClass = vi.fn(name => ({
+  name
+}))
+
+const Marti = new CorrectDogClass('Marti')
+const Newt = new IncorrectDogClass('Newt')
+
+Marti instanceof CorrectDogClass // true
+Newt instanceof IncorrectDogClass // false!
```
sheremet-va, line 660:
```suggestion
Newt instanceof IncorrectDogClass // ❌ false!
```

docs/guide/mocking.md
```diff
@@ -673,14 +700,22 @@ test('can feed dogs', () => {
...
 
 // you can use built-in assertions to check the validity of the call
 expect(dog.speak).toHaveBeenCalled()
+expect(dog.greet).toHaveBeenCalled()
+
+const dog2 = new Dog('Max')
+
+// methods assigned to the prototype are shared between instances
+expect(dog2.speak).toHaveBeenCalled()
+expect(dog2.greet).not.toHaveBeenCalled()
```
rodgobbi, line 718: @sheremet-va 
I added this extra part to extend the example, going it's helpful

docs/guide/mocking.md
```diff
@@ -622,18 +627,40 @@ We can re-create this class with ES5 functions:
...
+})
+
+const IncorrectDogClass = vi.fn(name => ({
+  name
+}))
+
+const Marti = new CorrectDogClass('Marti')
+const Newt = new IncorrectDogClass('Newt')
+
+Marti instanceof CorrectDogClass // true
```
sheremet-va, line 659:
```suggestion
Marti instanceof CorrectDogClass // ✅ true
```

docs/guide/mocking.md
```diff
@@ -673,14 +700,22 @@ test('can feed dogs', () => {
 ```
 :::
 
-Now, when we create a new instance of the `Dog` class its `speak` method (alongside `feed`) is already mocked:
+Now, when we create a new instance of the `Dog` class its `speak` method (alongside `feed` and `greet`) is already mocked:
 
 ```ts
 const dog = new Dog('Cooper')
```
sheremet-va, line 706: let's rename him to `Cooper` so we don't need to have numbers (`dog2`) later on

sheremet-va (changes requested):

docs/guide/mocking.md
rodgobbi: Done 👍

docs/guide/mocking.md
sheremet-va: I am not sure why the whole line is needed. The method is already defined as a public class method

sheremet-va: Basically, my question is why can't we just have this:

```ts
class Dog {
  name: string

  constructor(name: string) {
    this.name = name
  }

  static getType(): string {
    return 'animal'
  }

  greet(): string {
    return `Hi! My name is ${this.name}!`
  }

  speak(): string {
    return 'bark!'
  }

  isHungry() {}
  feed() {}
}
```

rodgobbi: > Why do we need to repeat it in the constructor? This code is how the source code looks like - people usually don't duplicate this

I'm a bit biased for sure, binding class methods like that was pretty common in React when classes was the default API to use React.

With the example, I want to simulate a method that only exists in the object instance.
Would it be clearer if I changed to a class field with arrow function like mentioned above:
```
  greet = () => {
    return `Hi! My name is ${this.name}!`
  }
```
would be equivalent to
```
    constructor(name) {
        this.greet = () => {
            return `Hi! My name is ${this.name}!`;
        };
    }
```

sheremet-va: Sure, this makes more sense to me in this example

rodgobbi: Awesome, makes sense, thanks for the thorough review 👍 

Finished updating based on your comment.

sheremet-va (approved):

sheremet-va:
Thank you! Looks good to me

PR #7837, opened 2025-04-15
fix!(coverage): remove `.all`, reworked `include/exclude`

AriPerkkio:
### Description

- Fixes https://github.com/vitest-dev/vitest/issues/6956
- Replaces `test-exclude` with `micromatch` and `tinyglobby`
  - `test-exclude` is 3.6MB and depends on `glob` and `minimatch`
  - `vitest` already depends on `picomatch` and `tinyglobby` so we can get those free in coverage packages

### Breaking changes:

- `coverage.all` option is removed
- `coverage.extensions` option is removed
- By default Vitest will show coverage only for the files that were imported during test run
- To see coverage for untested files, users need to configure `coverage.include` with a pattern.
- As `coverage.extensions` is removed, it is recommended to define extensions in `include` pattern, e.g. `include: ["src/**.{js,ts,tsx}"]`.

### Tests

### Documentation

### Changesets


packages/vitest/src/node/coverage.ts
```diff
@@ -128,6 +132,64 @@ export class BaseCoverageProvider<Options extends ResolvedCoverageOptions<'istan
...
+    let includedFiles = await glob(this.options.include, {
+      cwd: this.ctx.config.root,
+      ignore: [...this.options.exclude, ...testedFiles.map(file => slash(file))],
+      absolute: true,
+      dot: true,
+      onlyFiles: true,
+    })
+
+    // Run again through picomatch as tinyglobby's exclude pattern is different ({ "exclude": ["math"] } should ignore "src/math.ts")
+    includedFiles = includedFiles.filter(file => this.isIncluded(file))
```
AriPerkkio, line 184: We use `picomatch` when checking covered files, and `tinyglobby` when checking uncovered files from file system. To make `coverage.include` and `coverage.exclude` behave similarly, we need to run `tinyglobby`'s results through `picomatch` again here. This feels like duplicate work.

Minimal repro below. @SuperchupuDev any suggestions how to make `tinyglobby` behave identical to `picomatch` here?

```js
import { writeFileSync, rmSync, readdirSync, mkdirSync } from "node:fs";
import { resolve } from "node:path";
import { glob } from "tinyglobby";
import pm from "picomatch";

{
  // Setup
  rmSync("./src", { force: true, recursive: true });
  mkdirSync("./src");

  writeFileSync("./src/first.ts", "export const a = 1;", "utf8");
  writeFileSync("./src/second.ts", "export const a = 2;", "utf8");
}

// Include everything inside src, but exclude anything that contains word "second"
const options = {
  include: ["src"],
  exclude: ["second"],
};

// ['/<root>/repros/tinyglobby/src/first.ts', '/<root>/repros/tinyglobby/src/second.ts']
const files = readdirSync(options.include[0]).map((file) => resolve(options.include[0], file));

const picomatchResult = files.filter((file) =>
  pm.isMatch(file, options.include, {
    ignore: options.exclude,
    contains: true,
  })
);

const tinyglobbyResult = await glob(options.include, {
  ignore: options.exclude,
  absolute: true,
});

console.log({ picomatchResult, tinyglobbyResult });
// {
//   picomatchResult: [
//     '/<root>/repros/tinyglobby/src/first.ts'
//   ],
//
//   tinyglobbyResult: [
//     '/<root>/repros/tinyglobby/src/first.ts',
//     '/<root>/repros/tinyglobby/src/second.ts'
//   ]
// }
```

SuperchupuDev: hi! i just woke up. will investigate in some hours. off the top of my head you might want to use `expandDirectories: false` as i believe vitest uses that everywhere else? if it doesn't match after that then it's likely a picomatch issue, as iirc behavior between picomatch.isMatch and picomatch(pattern)(file) can differ (or at least i've seen behavior differences at least once in the past). still have to test and confirm though

sheremet-va (dismissed):

packages/vitest/src/node/coverage.ts
AriPerkkio: We'll merge this for Vitest v4 as is, but in the future it would be nice to have part working just by using `tinyglobby`, without needing `picomatch`.

sheremet-va (approved):

packages/vitest/src/node/coverage.ts
SuperchupuDev: @AriPerkkio sorry for the wait. after testing locally it looks like the reason why your example gives different results between tinyglobby and picomatch is because you are using the `contains` picomatch option, which tinyglobby does not expose as an option and neither does fast-glob. i suppose you can achieve the same functionality by wrapping the patterns with `**/*<pattern>*/**`. or you could move off `contains` and its behavior altogether

PR #7858, opened 2025-04-18
fix(browser): correctly inherit CLI options

### LINKED ISSUE #7521: CLI --browser.headless=false doesn't load Vitest UI when using workspaces

sheremet-va:
### Description

Fixes  #7521

Let's complicate config resolution even further!

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/config/resolveConfig.ts
+++ b/packages/vitest/src/node/config/resolveConfig.ts
@@ -234,7 +234,12 @@ export function resolveConfig(
 
   const browser = resolved.browser
 
-  if (browser.enabled) {
+  // if browser was enabled via CLI and it's configured by the user, then validate the input
+  if (browser.enabled && viteConfig.test?.browser) {
+    if (!browser.name && !browser.instances) {
+      throw new Error(`Vitest Browser Mode requires "browser.name" (deprecated) or "browser.instances" options, none were set.`)
+    }
+
     const instances = browser.instances
     if (browser.name && browser.instances) {
       // --browser=chromium filters configs to a single one
--- a/packages/vitest/src/node/config/resolveConfig.ts
+++ b/packages/vitest/src/node/config/resolveConfig.ts
@@ -235,25 +235,17 @@ export function resolveConfig(
   const browser = resolved.browser
 
   if (browser.enabled) {
-    if (!browser.name && !browser.instances) {
-      // CLI can enable `--browser.*` flag to change config of workspace projects
-      // the same flag will be applied to the root config that doesn't have to have "name" or "instances"
-      // in this case we just disable the browser mode
-      browser.enabled = false
+    const instances = browser.instances
+    if (browser.name && browser.instances) {
+      // --browser=chromium filters configs to a single one
+      browser.instances = browser.instances.filter(instance => instance.browser === browser.name)
     }
-    else {
-      const instances = browser.instances
-      if (browser.name && browser.instances) {
-        // --browser=chromium filters configs to a single one
-        browser.instances = browser.instances.filter(instance => instance.browser === browser.name)
-      }
 
-      if (browser.instances && !browser.instances.length) {
-        throw new Error([
-          `"browser.instances" was set in the config, but the array is empty. Define at least one browser config.`,
-          browser.name && instances?.length ? ` The "browser.name" was set to "${browser.name}" which filtered all configs (${instances.map(c => c.browser).join(', ')}). Did you mean to use another name?` : '',
-        ].join(''))
-      }
+    if (browser.instances && !browser.instances.length) {
+      throw new Error([
+        `"browser.instances" was set in the config, but the array is empty. Define at least one browser config.`,
+        browser.name && instances?.length ? ` The "browser.name" was set to "${browser.name}" which filtered all configs (${instances.map(c => c.browser).join(', ')}). Did you mean to use another name?` : '',
+      ].join(''))
     }
   }
 
--- a/test/config/test/browser-configs.test.ts
+++ b/test/config/test/browser-configs.test.ts
@@ -413,7 +413,7 @@ describe('[e2e] workspace configs are affected by the CLI options', () => {
       // headless was set in the config
       headless: true,
       browser: true,
-      // UI is false necause headless is enabled
+      // UI is false because headless is enabled
       ui: false,
       parent: {
         name: 'browser',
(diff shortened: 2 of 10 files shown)


test/config/test/browser-configs.test.ts
```diff
@@ -303,3 +306,143 @@ test('can enable browser-cli options for multi-project workspace', async () => {
...
+                    return {
+                      name: p.name,
+                      headless: p.config.browser.headless,
+                      browser: p.config.browser.enabled,
+                      ui: p.config.browser.ui,
+                    }
+                  })
+                }))
+                // throw an error to avoid running tests
+                throw new Error('stop')
```
AriPerkkio, line 343: This utility looks great - I've thought about building something like this before. Sometimes it's not enough to check resolved options just via `resolveConfig` as plugins and other parts can change them before tests run. Bailing out like this before pools are running saves time.

test/config/test/browser-configs.test.ts
```diff
@@ -303,3 +306,143 @@ test('can enable browser-cli options for multi-project workspace', async () => {
...
+          name: 'browser',
+          browser: {
+            enabled: true,
+            headless: true,
+            instances: [
+              {
+                browser: 'chromium',
+              },
+            ],
+          },
```
AriPerkkio, line 387: This config should actually throw an error that `preview` provider doesn't support `headless: true`. 🤔 

Using `instances: [{ browser: 'chromium' }]`  makes this look like `playwright` was intended but it actually resolves to `preview`.

AriPerkkio (changes requested):
Looks good so far, great work 💯 

I wrote couple of manual test cases earlier that fail with these changes:

This config correctly opens Vitest UI when run with `vitest --browser.headless=false` ✅

```ts

export default defineConfig({
  test: {
    browser: {
      provider: "playwright",
      enabled: true,
      headless: true,
      instances: [{ browser: "chromium" }],
    },
  },
});
```

This config opens browser without Vitest UI when run with `vitest --browser.headless=false` ❌ 

```ts

export default defineConfig({
  test: {
    workspace: [
      {
        test: {
          name: "browser",
          browser: {
            provider: "playwright",
            enabled: true,
            headless: true,
            instances: [{ browser: "chromium" }],
          },
        },
      },
    ],
  },
});
```

This config does not run test cases in Browser Mode when run with `vitest --browser.enabled` ❌ 

```ts

export default defineConfig({
  test: {
    workspace: [
      {
        test: {
          name: "browser",
          browser: {
            provider: "playwright",
            enabled: false,
            headless: true,
            instances: [{ browser: "chromium" }],
          },
        },
      },
    ],
  },
});
```

test/config/test/browser-configs.test.ts
sheremet-va: The preview-headless error is thrown by the provider when tests start to run, this test bails out before that

sheremet-va:
> I wrote couple of manual test cases earlier that fail with these changes:

I cannot confirm this. All configs you provided open the UI (automatic tests also check that UI is set, tests fail without these changes)

packages/vitest/src/node/config/resolveConfig.ts
```diff
@@ -234,26 +234,23 @@ export function resolveConfig(
 
   const browser = resolved.browser
 
-  if (browser.enabled) {
+  // if browser was enabled via CLI and it's configured by the user, then validate the input
+  if (browser.enabled && viteConfig.test?.browser) {
```
sheremet-va, line 238: The main change is that we ignore all browser CLI flags if workspace doesn't have the `browser` configured

test/config/test/browser-configs.test.ts
```diff
@@ -303,3 +307,231 @@ test('can enable browser-cli options for multi-project workspace', async () => {
...
+      ui: true,
+      parent: null,
+    })
+
+    expect(config.workspace[1]).toEqual({
+      name: 'browser (chromium)',
+      // headless was set in the config
+      headless: true,
+      browser: true,
+      // UI is false necause headless is enabled
```
AriPerkkio, line 416:
```suggestion
      // UI is false because headless is enabled
```

AriPerkkio:
> I cannot confirm this. All configs you provided open the UI (automatic tests also check that UI is set, tests fail without changes in this PR)

You are right, it seems to work properly now! 💯

packages/vitest/src/node/config/resolveConfig.ts
sheremet-va: Note: we throw an error if `--browser` is provided, but there are no configurations for it

AriPerkkio (approved):

silvenon:
Thanks a million, this will reduce the amount of people acidentally comitting changing `headless: true` to `false` because it will be possible to turn it off from the CLI as needed 😌

PR #7860, opened 2025-04-18
docs: use `extends` instead of `configFile` in `injectTestProjects`

romhml:
### Description

This fixes a typo in the `injectTestProjects` documentation which was referencing the `configFile` attribute instead of `extends` in the instructions to inherit the user's config.

See: https://github.com/vitest-dev/vitest/blob/main/packages/vitest/src/node/types/config.ts#L1134

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/advanced/api/plugin.md
+++ b/docs/advanced/api/plugin.md
@@ -93,10 +93,10 @@ This methods accepts a config glob pattern, a filepath to the config or an inlin
 ```ts
 // inject a single project with a custom alias
 const newProjects = await injectTestProjects({
-  // you can inherit the current project config by referencing `configFile`
+  // you can inherit the current project config by referencing `extends`
   // note that you cannot have a project with the name that already exists,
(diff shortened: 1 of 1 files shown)


sheremet-va:
> This fixes a typo in the `injectTestProjects` documentation which was referencing the `configFile` attribute instead of `extends` in the instructions to inherit the user's config.

Why do you think it's better to document `extends` instead of the `configFile`?

romhml:
I've been experimenting with `injectTestProjects` here: https://github.com/romhml/compodium/blob/feat/vitest-integration/packages/core/src/plugins/tests.ts#L47

The `configFile` option from the documentation raises a typescript error in this context and does not work as expected. Using `extend` on the other hand works, so I thought this was a typo. Am I missing something?

I can work on a minimal reproduction if needed.

sheremet-va (approved):

Issue #7825, opened 2025-04-11
Total test duration is wrong in `--merge-reports`


kettanaito:
### Describe the bug

When merging reports from a sharded test run, the merged report prints an incorrect total duration:

```
npx vitest run --merge-reports

 ✓ src/one.test.ts (500 tests) 268ms
 ✓ src/two.test.ts (500 tests) 266ms

 Test Files  2 passed (2)
      Tests  1000 passed (1000)
   Start at  11:48:26
   Duration  6ms (transform 0ms, setup 0ms, collect 25ms, tests 534ms, environment 0ms, prepare 86ms)
             ^^^
```

The tests have clearly taken more time than 6ms to run. It is also apparent from the detailed timings in parenthesis. 

It seems that the total time here is of the `--merge-report` command itself, not the total time of the merged tests, which can be misleading.

### Reproduction

You can reproduce this issue by following the [Sharing](https://main.vitest.dev/guide/improving-performance.html#sharding) recipe from the Vitest docs. No additional changes were made. 

### System Info

```shell
—
```

### Used Package Manager

npm

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vitest.dev/guide/).
- [x] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

AriPerkkio:
I don't think the total duration is saved anywhere. It's reporter specifid metric only:

https://github.com/vitest-dev/vitest/blob/412866556b7193d3c1a1acb1295f79c07e885758/packages/vitest/src/node/reporters/base.ts#L23-L25

https://github.com/vitest-dev/vitest/blob/412866556b7193d3c1a1acb1295f79c07e885758/packages/vitest/src/node/reporters/base.ts#L484-L509

We should probably add similar to `blob` reporter and use it here:

[...]

PR #7877, opened 2025-04-23
fix(reporters): `--merge-reports` to show each total run times

AriPerkkio:
### Description

- Fixes https://github.com/vitest-dev/vitest/issues/7825

Adds total test run times of each blob in test report of `--merge-reports` run. The `Duration` is sum of all blobs. Below it we show `Per blob` durations.

### Tests

### Documentation

### Changesets


packages/vitest/src/node/core.ts
```diff
@@ -510,7 +510,7 @@ export class Vitest {
     }
 
     this._checkUnhandledErrors(errors)
-    await this._testRun.end(specifications, errors).catch(noop)
+    await this._testRun.end(specifications, errors, undefined, executionTimes).catch(noop)
```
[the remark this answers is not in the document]
AriPerkkio: As `TestRun` is private, I can make it into a `{ coverage, executionTimes }` object. 

But reporter `onFinished` will just receive new variable - we can convert that into object in next major if wanted.

packages/vitest/src/node/types/reporter.ts
```diff
@@ -35,7 +35,8 @@ export interface Reporter {
   onFinished?: (
     files: File[],
     errors: unknown[],
-    coverage?: unknown
+    coverage?: unknown,
+    executionTimes?: number[]
```
sheremet-va, line 39: if we have to add something here, let's keep it `@internal`

packages/vitest/src/node/test-run.ts
```diff
@@ -74,12 +74,12 @@ export class TestRun {
       await Promise.all([
         this.vitest.report('onTestRunEnd', modules, [...errors] as SerializedError[], state),
```
sheremet-va, line 75: shouldn't times be passed down somehow to new hooks? At one point, `onFinished` won't be part of the API

packages/vitest/src/node/test-run.ts
```diff
@@ -57,7 +57,7 @@ export class TestRun {
     }
   }
 
-  async end(specifications: TestSpecification[], errors: unknown[], coverage?: unknown): Promise<void> {
+  async end(specifications: TestSpecification[], errors: unknown[], meta: { coverage?: unknown; executionTimes?: number[] }): Promise<void> {
```
sheremet-va, line 60: why would times be optional?

sheremet-va (commented):
Generally, I dislike `executionTimes?: number[]` as a public API. 

What are these numbers? Why are the optional? Do we just keep adding new arguments to hooks? I think this needs to be standardized _somehow_

Maybe have something like a `SubTestRun`? Or `ShardTestRun`

AriPerkkio:
Do you have any suggestions what would be better way to pass data from `blob` reporter's `readBlob` to other reporters? Should we instead store it in `ctx` or `ctx.state`?

packages/vitest/src/node/test-run.ts
AriPerkkio: Only present when `--merge-reports` is used.

AriPerkkio:
Storing blobs in `ctx.blobs` now. Base reporter is accessing `executionTimes` from state instead of reporter hooks.

sheremet-va (approved):

PR #7878, opened 2025-04-23
feat: provide `ctx.signal`

### LINKED ISSUE #7647: AbortSignal in tests on run cancel

When tests that depend on other processes, like during loadtesting or long-running tests, tests that have a teardown process (using `afterAll` or `afterEach`) - it would be great to have an AbortSignal in the test itself that would abort when the run is cancelled (by pressing Q or CTRL+C) that would be used to abort the dependant processes.

### Suggested solution

Roughly
```ts
import { it, expect } from 'vitest';
import { startServer, loadtest } from './my-tools';

it('should pass this long running test', async (ctx) => {
  const server  = await startServer();

  const highestMemoryUsage = await loadtest({
    server,
    // the loadtest runs for a longer period of time, it'll use this signal to stop loadtesting earlier on run cancel
    signal: ctx.signal,
  });

  expect(highestMemoryUsage).toBeLessThan(100);
});
```

### Alternative

```ts
import { it, expect } from 'vitest';
import { onCancel } from 'vitest/runners';
import { startServer, loadtest } from './my-tools';

const ctrl = new AbortController();
onCancel(() => ctrl.abort());

it('should pass this long running test', async (ctx) => {
  const server  = await startServer();

  const highestMemoryUsage = await loadtest({
    server,
    // the loadtest runs for a longer period of time, it'll use this signal to stop loadtesting earlier on run cancel
    signal: ctx.signal,
  });

  expect(highestMemoryUsage).toBeLessThan(100);
});
```

### Additional context

If terminating abruptly (double CTRL+C), the teardown wont happen leaving processes or services (like Docker containers) running after the tests end.

### Validations

sheremet-va:
### Description

Fixes #7647

This PR provides a `signal` value to the test context. Signal can be aborted by Vitest when:
- Current run is aborted with CTRL+C
- Test times out
- `--bail` cancels test run

Test plan:
1. timeout in a normal test aborts the signal
2. timeout in multiple concurrent tests aborts all signals
3. `cancelCurrentRun` aborts the signal in a single test
3. `cancelCurrentRun` aborts the signal in a multiple concurrent tests


packages/runner/src/run.ts
```diff
@@ -588,21 +599,35 @@ export async function runFiles(files: File[], runner: VitestRunner): Promise<voi
 }
 
 export async function startTests(specs: string[] | FileSpecification[], runner: VitestRunner): Promise<File[]> {
-  const paths = specs.map(f => typeof f === 'string' ? f : f.filepath)
-  await runner.onBeforeCollect?.(paths)
+  const cancel = runner.cancel?.bind(runner)
+  runner.cancel = (reason) => {
```
sheremet-va, line 603: Ideally, we _need_ to have an event listener for this, but only have a runner here. Adding another `onCancel` felt wrong (maybe it needs to be refactored)

packages/runner/src/run.ts
```diff
@@ -274,6 +282,7 @@ export async function runTest(test: Test, runner: VitestRunner): Promise<void> {
   }
   updateTask('test-prepare', test, runner)
 
+  const cleanupRunningTest = addRunningTest(test)
```
sheremet-va, line 285: This is needed for concurrent tests

docs/guide/test-context.md
```diff
@@ -65,6 +65,21 @@ it('math is hard', ({ skip }) => {
 })
 ```
 
+#### `context.signal` <Version>3.2.0</Version> {#context-signal}
+
+A signal object that can be aborted by Vitest. The signal is aborted in these situations:
```
AriPerkkio, line 70: Maybe:

```suggestion
An [`AbortSignal`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) that can be aborted by Vitest. The signal is aborted in these situations:
```

packages/runner/src/context.ts
```diff
@@ -102,14 +106,34 @@ export function withTimeout<T extends (...args: any[]) => any>(
...
   const context = function () {
     throw new Error('done() callback is deprecated, use promise instead')
-  } as unknown as TestContext
+  } as unknown as WriteableTestContext
+
+  const abortController = abortControllers.get(context) || (() => {
+    const abortController = new AbortController()
+    abortControllers.set(context, abortController)
+    return abortController
+  })()
```
AriPerkkio, line 134: Nitpick: Personally I would like this IIFE to be just good old `if`. It's okey to be verbose to reduce complexity 😄 

```suggestion
  let abortController = abortControllers.get(context)
  
   if(!abortController) {
    abortController = new AbortController()
    abortControllers.set(context, abortController)
  }
```

packages/runner/src/errors.ts
```diff
@@ -9,3 +9,8 @@ export class PendingError extends Error {
     this.taskId = task.id
   }
 }
+
+export class AbortError extends Error {
+  name = 'AbortError'
+  code = 20
```
AriPerkkio, line 15: What's the `20` for?

packages/runner/src/test-state.ts
```diff
@@ -9,3 +9,15 @@ export function setCurrentTest<T extends Test>(test: T | undefined): void {
 export function getCurrentTest<T extends Test | undefined>(): T {
   return _test as T
 }
+
+const tests: Array<Test> = []
+export function addRunningTest(test: Test): () => void {
+  const index = tests.push(test)
+  return () => {
+    tests.splice(index)
+  }
```
AriPerkkio, line 18: Is it possible that this is not called serially, e.g. in `test.concurrent`? The stored `index` could become stale.

```suggestion
  return () => {
    tests.splice(tests.indexOf(test))
  }
```

test/cli/test/signal.test.ts
AriPerkkio, line 1: Great test cases! 🤩

packages/runner/src/errors.ts
sheremet-va: The code for abort error in DOMException: https://developer.mozilla.org/en-US/docs/Web/API/DOMException/code

packages/runner/src/test-state.ts
sheremet-va: True, this code won't work with concurrent tests

PR #7885, opened 2025-04-24
fix: extraneous coverage for --project (fix #6331)

### LINKED ISSUE #6331: Allow project filtered test runs to report coverage of only filtered projects

Currently running `vitest run --coverage --project <project_a>` will result in all other projects reporting no coverage.

### Suggested solution

Report only the coverage of the filtered projects when using the `--project` flag.

### Alternative

_No response_

### Additional context

Getting accurate project coverage while testing in a monorepo is hard and you have to resort to `npm run cover -w <project_a>` which generates a `coverage` folder that isn't ignored by default at the project level.

### Validations

gtbuchanan:
### Description

Coverage was reported for projects outside of the specified project filter. For example, given a monorepo with two projects, p1 and p2, `vitest run --coverage --project=p1` reported coverage for code from p2 at 0% because no tests from p2 were run.

Resolves: #6331

### Tests

### Documentation

### Changesets


gtbuchanan:
@AriPerkkio Does this align with what you were thinking? I still need to familiarize myself with how the tests are written, but I wanted to ensure I was on the right track first.

Also, should I move the duplicate code to a common location (like `BaseCoverageProvider`)?

AriPerkkio (commented):
This change would only filter out uncovered files outside the workspace project's root. We filter out covered files too.

To do this, we'll need to make `this.testExclude` an array, and create one for each project with their roots. 

https://github.com/vitest-dev/vitest/blob/9be32fb74dfc990c40024e3ca979b92cae6e4fdd/packages/coverage-istanbul/src/provider.ts#L34-L35

gtbuchanan:
That should do it. It seems to work for my project. I will add tests when I figure that out 🙂 

As an aside, I wasn't able to move the duplicate code to `BaseCoverageProvider` because `vitest/node` doesn't reference `test-exclude`. I figured it wasn't worth adding the extra reference for this fix.

packages/coverage-istanbul/src/provider.ts
```diff
@@ -22,23 +22,33 @@ import { COVERAGE_STORE_KEY } from './constants'
...
+      excludeNodeModules: true,
+      extension: options.extension,
+      relativePath: !options.allowExternal,
+    })
+    return { root: config.root, exclude }
+  }
+  return ctx.config.project.length
+    ? ctx.projects.map(project => create(project.config))
+    : [create(ctx.config)]
+}
```
AriPerkkio, line 40: We can inline this in constructor:

```ts
    this.testExcludes = roots.map(root => ({
      root,
      exclude: new TestExclude({
        cwd: root,
        include: this.options.include,
        exclude: this.options.exclude,
        excludeNodeModules: true,
        extension: this.options.extension,
        relativePath: !this.options.allowExternal,
      }),
    }))
```

AriPerkkio (changes requested):
Looks good so far.

> As an aside, I wasn't able to move the duplicate code to BaseCoverageProvider because vitest/node doesn't reference test-exclude. I figured it wasn't worth adding the extra reference for this fix.

That's okey for now. In https://github.com/vitest-dev/vitest/pull/7837 we'll remove `test-exclude` completely and utilize packages that are available in `vitest` package. That allows re-using logic from `BaseProvider`.

For test cases we'll need to verify following:
- When `--project` filter is passed, files outside that project are excluded. This means covered and uncovered files.
- When `--project` filter is not passed, files in project and in the root should be visible.

You can add fixtures in `test/coverage-test/fixtures/workspaces` and create test cases in new file `test/coverage-test/test/workspace.test.ts` for example.

AriPerkkio:
Hey @gtbuchanan, do you have any time to finalize this PR? I'm also happy to take over and finish it if you prefer that. Thanks for all the work so far! 💯

gtbuchanan:
@AriPerkkio I would have liked to finish it myself, but I'm running into some issues:

1. I figured out how to write the tests, but the one testing my fix is actually failing. I'm trying to debug the coverage test in VS Code using [these](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md#vs-code) steps, but clicking "Run and Debug" immediately stops debugging. Do you have any tips? Is there a specific file I'm supposed to have open?
2. ~Obviously, my branch is now outdated. Out of curiosity, I added my unit tests on top of `main` to make sure one would fail, but they actually passed. I'm assuming my test is just wrong since `main` doesn't appear to have a fix for this. Will you glance at my tests to see if something stands out?~

EDIT:
Disregard 2. There must have been some caching involved. I picked the PR back up on my work computer and the test fails there as expected. Now to reapply the fix...

gtbuchanan:
I didn't figure out how to debug the coverage tests in VS Code, but I was able to get everything working! It seems to work without updating `isIncluded`, but let me know if I'm missing something.

packages/vitest/src/node/coverage.ts
```diff
@@ -187,7 +192,19 @@ export class BaseCoverageProvider<Options extends ResolvedCoverageOptions<'istan
...
-    return includedFiles.map(file => slash(path.resolve(this.ctx.config.root, file)))
+    return includedFiles.map(file => slash(path.resolve(root, file)))
+  }
+
+  async getUntestedFiles(testedFiles: string[]): Promise<string[]> {
+    if (this.options.include == null) {
+      return []
+    }
+
+    const rootMapper = this.getUntestedFilesByRoot.bind(this, testedFiles, this.options.include)
```
gtbuchanan, line 210: This strategy avoids assigning `this.options.include` to a local variable or ignoring the nullability warning in a lambda.

test/coverage-test/test/workspace.test.ts
AriPerkkio, line 1: Let's rename this to `workspace.project-filter.test.ts` to align with other `workspace.**.test.ts` tests.

test/coverage-test/fixtures/configs/vitest.config.workspace.ts
```diff
@@ -0,0 +1,20 @@
...
+      },
+      {
+        test: {
+          name: "project2",
+          root: "fixtures/workspaces/project/project2",
+        },
+      },
+    ]
+  }
+});
```
AriPerkkio: Let's add a `project-shared` that contains `project-shared/src/utils.ts`, that's used by `id.ts` and `konst.ts`. Then make sure that `utils.ts` does not show up in report when project filter is used even though it is covered file.

AriPerkkio (changes requested):
> There must have been some caching involved. I picked the PR back up on my work computer and the test fails there as expected. Now to reapply the fix...

The test cases are running against built assets so you'll need to run `pnpm build` or `pnpm dev` before running the test cases. That might explain the caching issue. 

> I didn't figure out how to debug the coverage tests in VS Code

These test cases are running "Vitest inside Vitest" so it's likely that VS Code extension is unable to inspect those.

test/coverage-test/fixtures/configs/vitest.config.workspace.ts
gtbuchanan: Good call. Turns out I did need to update `isIncluded`.

packages/vitest/src/node/coverage.ts
```diff
@@ -130,12 +131,18 @@ export class BaseCoverageProvider<Options extends ResolvedCoverageOptions<'istan
       this.options.reportsDirectory,
       tempDirectory,
     )
+
+    this.roots = ctx.config.project?.length
+      ? ctx.projects.map(project => project.config.root)
+      : [ctx.config.root]
```
AriPerkkio: Let's add a comment to specify this logic:

```suggestion

    // If --project filter is set pick only roots of resolved projects
    this.roots = ctx.config.project?.length
      ? ctx.projects.map(project => project.config.root)
      : [ctx.config.root]
```

packages/vitest/src/node/coverage.ts
```diff
@@ -144,50 +151,64 @@ export class BaseCoverageProvider<Options extends ResolvedCoverageOptions<'istan
...
+    const included = roots.some((root) => {
+      const options: pm.PicomatchOptions = {
+        contains: true,
+        dot: true,
+        cwd: root,
+        ignore: this.options.exclude,
+      }
+
+      return pm.isMatch(filename, glob, options)
+    }) && existsSync(cleanUrl(filename))
```
AriPerkkio: Format changes to improve readability:

```suggestion
    let included = roots.some(root =>
      pm.isMatch(filename, glob, {
        contains: true,
        dot: true,
        cwd: root,
        ignore: this.options.exclude,
      }))

    included = included && existsSync(cleanUrl(filename))
```

packages/vitest/src/node/coverage.ts
```diff
@@ -130,12 +131,18 @@ export class BaseCoverageProvider<Options extends ResolvedCoverageOptions<'istan
       this.options.reportsDirectory,
       tempDirectory,
     )
+
+    this.roots = ctx.config.project?.length
+      ? ctx.projects.map(project => project.config.root)
```
AriPerkkio: Just realized that projects can have identical `root` too (e.g. https://github.com/AriPerkkio/aria-live-capture/blob/cc0b9d87e3d8d70e2f115c6a9a1ef6bc03195e89/vitest.config.ts#L20-L40) when they are used just to test different settings. 

Let's make sure we don't hold duplicate entries in `this.roots` to avoid unnecessary globbing in other methods:

```suggestion
      ? [...new Set(ctx.projects.map(project => project.config.root))]
```

AriPerkkio (changes requested):
Small improvements, otherwise this is looking good!

Btw as this PR is from a Github organization account (`energyworldnet`), Vitest authors cannot push any changes to your PR branch. Usually I would fix small errors like this myself and push changes to your branch directly, but now it's not possible due to permission requirements.

AriPerkkio (approved):
Looks good, thanks for the PR @gtbuchanan! 💯

PR #7915, opened 2025-04-30
feat: Implement connect option for playwright browser provider

egfx-notifications:
### Description

This PR adds the ability to use the Playwright browser provider with remote Playwright servers. Playwright itself doesn't support running on all distributions, it is primarily tested against Debian-based distributions. Notably Playwright doesn't support running on Red Hat Enterprise Linux and derived distributions. To solve this use case Playwright allows its client and server to run on different hosts, so the server may be launched on a supported distribution (see https://playwright.dev/docs/docker#remote-connection). By offering the client connect mode as an alternative to the launch mode we make browser mode available for more distributions.

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/guide/browser/playwright.md
+++ b/docs/guide/browser/playwright.md
@@ -66,7 +66,7 @@ Vitest will ignore `launch.headless` option. Instead, use [`test.browser.headles
 Note that Vitest will push debugging flags to `launch.args` if [`--inspect`](/guide/cli#inspect) is enabled.
 :::
 
-## connect
+## connect <Version>3.2.0</Version> {#connect}
 
 These options are directly passed down to `playwright[browser].connect` command. You can read more about the command and available arguments in the [Playwright documentation](https://playwright.dev/docs/api/class-browsertype#browser-type-connect).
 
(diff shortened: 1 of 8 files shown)


packages/browser/src/node/providers/playwright.ts
```diff
@@ -76,6 +81,13 @@ export class PlaywrightBrowserProvider implements BrowserProvider {
 
       const playwright = await import('playwright')
 
+      if (this.options?.connect) {
```
sheremet-va, line 84: Can we also add a warning if `launch` options exists that they are ignored?

sheremet-va (approved):

docs/guide/browser/playwright.md
```diff
@@ -65,6 +66,14 @@ Vitest will ignore `launch.headless` option. Instead, use [`test.browser.headles
 Note that Vitest will push debugging flags to `launch.args` if [`--inspect`](/guide/cli#inspect) is enabled.
 :::
 
+## connect
```
sheremet-va, line 69:
```suggestion
## connect <Version>3.2.0</Version> {#connect}
```

sheremet-va (dismissed):

PR #7927, opened 2025-05-03
feat: Explicit Resource Management support in mocked functions

EskiMojo14:
### Description

Support [explicit resource management](https://github.com/tc39/proposal-explicit-resource-management) with mocked functions, meaning mockRestore can be called automatically once scope is exited.

```ts
using consoleSpy = vi.spyOn(console, "spy")
// is like
const consoleSpy = vi.spyOn(console, "spy")
try {
  // ...
} finally {
  consoleSpy.mockRestore()
}
```

This is very convenient, but also good for Jest parity - Jest added it here https://github.com/jestjs/jest/pull/14895 and released in [v30.0.0-alpha.3](https://github.com/jestjs/jest/releases/tag/v30.0.0-alpha.3).

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/api/vi.md
+++ b/docs/api/vi.md
@@ -460,6 +460,19 @@ expect(spy).toHaveReturnedWith(1)
 ```
 
 ::: tip
+In environments that support [Explicit Resource Management](https://github.com/tc39/proposal-explicit-resource-management), you can use `using` instead of `const` to automatically call `mockRestore` on any mocked function when the containing block is exited. This is especially useful for spied methods:
+
+```ts
+it('calls console.log', () => {
+  using spy = vi.spyOn(console, 'log').mockImplementation(() => {})
+  debug('message')
+  expect(spy).toHaveBeenCalled()
+})
+// console.log is restored here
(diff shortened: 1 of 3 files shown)


hi-ogawa (commented):
Thanks! I think this is nice to have. Vitest should already support `using` at least on typescript out of the box since esbuild transpiles it https://github.com/vitest-dev/vitest/pull/4409, so you can probably add a test with `using` like

https://github.com/vitest-dev/vitest/blob/30b709caf5281f13cc3a326c76770a7ba207acb6/test/core/test/esnext.test.ts#L15-L24

sheremet-va:
This looks good to me, but seems like tests are failing (ignore the browser ones)

EskiMojo14:
I don't really understand why tests are failing - is there a way to get a more useful failure message?

EskiMojo14:
tests are fixed - the dispose was holding onto the function reference instead of using the key we can spy on.

sheremet-va (approved):

PR #7959, opened 2025-05-11
chore: remove deprecated type

iiio2:
`ErrorWithDiff` is showing deprecated in vitest.


sheremet-va:
`ErrorWithDiff` is deprecated publically, internal usage is not affected until we remove the types in #7619

iiio2:
Oh, I see. Thanks @sheremet-va .

PR #7982, opened 2025-05-15
fix(docs): remove falsy comment on test execution speed

christian-bromann:
### Description

This PR removes a wrong statement from the docs that WebdriverIO doesn't support parallel tests and is slower than PW. Both statements are wrong. You can run as many parallel browser sessions as your system allows you. Furthermore, WebdriverIO is mostly relying on WebDriver Bidi these days which is the same communication method than CDP is using. You may technically argue that WebDriver Bidi messages go through a browser driver which is another hop and can potentially be slower than sending a socket message directly to the browser but I would argue that this doesn't impact test execution time a lot and I would love to see someone challenging this.

On contrary I can put in a statement saying that Playwright browser are not real browser used by users in the real world given they are patched to accomodate Playwrights proprietary automation approach.

I think generally we should stay away from making judgements about what is better for the user to use. 

### Tests

### Documentation

### Changesets


sheremet-va:
We did remove the second part of that statement in #8823. The first part still stands. I am going to close the PR since there is no traction here.

PR #7985, opened 2025-05-15
fix(coverage): in-source test cases excluded

AriPerkkio:
### Description

- Exclude in-source tests' `if (import.meta.vitest)` contents from coverage
- Applied to Istanbul and AST-aware V8 only
  - the old V8 would require hacky solutions and is being deprecated soon so leaving it out of scope

Before-after:

image

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/coverage-v8/src/provider.ts
+++ b/packages/coverage-v8/src/provider.ts
@@ -292,6 +292,20 @@ export class V8CoverageProvider extends BaseCoverageProvider<ResolvedCoverageOpt
           ) {
             return true
           }
+
+          // Browser mode's "import.meta.env ="
+          if (
+            type === 'statement'
+            && node.type === 'ExpressionStatement'
+            && node.expression.type === 'AssignmentExpression'
+            && node.expression.left.type === 'MemberExpression'
+            && node.expression.left.object.type === 'MetaProperty'
+            && node.expression.left.object.meta.name === 'import'
+            && node.expression.left.object.property.name === 'meta'
+            && node.expression.left.property.type === 'Identifier'
+            && node.expression.left.property.name === 'env') {
+            return true
+          }
         },
       },
       )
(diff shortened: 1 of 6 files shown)


packages/coverage-v8/package.json
```diff
@@ -56,7 +56,7 @@
   "dependencies": {
     "@ampproject/remapping": "catalog:",
     "@bcoe/v8-coverage": "^1.0.2",
-    "ast-v8-to-istanbul": "^0.3.1",
+    "ast-v8-to-istanbul": "https://pkg.pr.new/AriPerkkio/ast-v8-to-istanbul@50",
```
AriPerkkio, line 59: https://github.com/AriPerkkio/ast-v8-to-istanbul/pull/50

packages/coverage-v8/src/provider.ts
```diff
@@ -283,6 +283,19 @@ export class V8CoverageProvider extends BaseCoverageProvider<ResolvedCoverageOpt
...
             return true
           }
+
+          // in-source test with "if (import.meta.vitest)"
+          if (
+            (type === 'branch' || type === 'statement')
+            && node.type === 'IfStatement'
+            && node.test.type === 'MemberExpression'
+            && node.test.object.type === 'Identifier'
+            && node.test.object.name === '__vite_ssr_import_meta__'
```
sheremet-va, line 293: Would also be nice to support `import.meta.vitest` directly (the browser mode doesn't transform it into `__vite_ssr_import_meta__`)

AriPerkkio: Good point! We should enable AST-aware remapping for browser mode in `test/coverage-test` to catch this. So far it's only been enabled for Node.

sheremet-va (approved):

PR #7994, opened 2025-05-17
fix(dot): allow dot reporter to work in non interactive terminals

### LINKED ISSUE #3932: vitest dot reporter is not used in CI (jenkins)
### Describe the bug

with viteconfig 
```js
test: {
    globals: true,
    environment: 'jsdom',
    setupFiles: './tests/setup.ts',
    reporters: ['dot'],
    coverage: {
        provider: 'v8',
        reportsDirectory: './reports/coverage-vitest',
        reporter: [['lcov']],
        lines: 3,
        functions: 3,
        branches: 3,
        statements: 3,
    },
},
  ```
  
  or command line 
  ```bash
  npx vitest --reporter dot
  ```
  
  work on "my machine" (osx)
  
  but it doesn't in CI (jenkins)
  
Screen Shot 2023-08-11 at 10 20 01

  

### Reproduction

hard to build env with jenkins and share it,

### System Info

```shell
10:28:09  + npx envinfo --system --npmPackages '{vitest,@vitest/*,vite,@vitejs/*}' --binaries --browsers
10:28:10  npm WARN exec The following package was not found and will be installed: envinfo@7.10.0
10:28:11  
10:28:11    System:
10:28:11      OS: Linux 5.4 Amazon Linux 2
10:28:11      CPU: (16) x64 Intel(R) Xeon(R) Platinum 8259CL CPU @ 2.50GHz
10:28:11      Memory: 121.03 GB / 124.36 GB
10:28:11      Container: Yes
10:28:11      Shell: 4.2.46 - /bin/bash
10:28:11    Binaries:
10:28:11      Node: 16.20.0 - /usr/local/lib/nodejs/node-v16.20.0-linux-x64/bin/node
10:28:11      npm: 8.19.4 - /usr/local/lib/nodejs/node-v16.20.0-linux-x64/bin/npm
1
```

### Used Package Manager

npm

### Validations

bstephen1:
### Description

Updates dot reporter to support non-TTY terminals instead of silently switching to basic reporter. Closes #3932

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/reporters/dot.ts
+++ b/packages/vitest/src/node/reporters/dot.ts
@@ -1,4 +1,5 @@
 import type { File, Test } from '@vitest/runner'
+import type { Writable } from 'node:stream'
 import type { Vitest } from '../core'
 import type { TestCase, TestModule } from './reported-tasks'
 import c from 'tinyrainbow'
@@ -30,11 +31,8 @@ export class DotReporter extends BaseReporter {
     }
   }
 
-  printTestModule(testModule: TestModule): void {
-    if (!this.isTTY) {
-      super.printTestModule(testModule)
-    }
-  }
+  // ignore base reporter
+  printTestModule(): void {}
 
   onWatcherRerun(files: string[], trigger?: string): void {
     this.tests.clear()
@@ -46,6 +44,9 @@ export class DotReporter extends BaseReporter {
     if (this.isTTY) {
       const finalLog = formatTests(Array.from(this.tests.values()))
       this.ctx.logger.log(finalLog)
+    }
+    else {
+      this.ctx.logger.log()
     }
 
     this.tests.clear()
@@ -70,10 +71,15 @@ export class DotReporter extends BaseReporter {
   }
 
   onTestCaseResult(test: TestCase): void {
+    const result = test.result().state
+    if (!this.isTTY && result !== 'pending') {
+      (this.ctx.logger.outputStream as Writable).write(formatTests([result]))
+    }
+
     super.onTestCaseResult(test)
 
     this.finishedTests.add(test.id)
-    this.tests.set(test.id, test.result().state || 'skipped')
+    this.tests.set(test.id, result || 'skipped')
     this.renderer?.schedule()
   }
 
--- a/test/reporters/tests/dot.test.ts
+++ b/test/reporters/tests/dot.test.ts
@@ -56,9 +56,8 @@ describe('{ isTTY: false }', () => {
       typecheck: undefined,
     })
 
-    expect(stdout).toContain('✓ fixtures/ok.test.ts')
+    expect(stdout).toContain('\n·\n')
     expect(stdout).toContain('Test Files  1 passed (1)')
-    expect(stdout).not.toContain('·')
 
     expect(stderr).toBe('')
   })
@@ -70,11 +69,7 @@ describe('{ isTTY: false }', () => {
       typecheck: undefined,
     })
 
-    expect(stdout).toContain('❯ fixtures/some-failing.test.ts (2 tests | 1 failed)')
-    expect(stdout).toContain('✓ 2 + 3 = 5')
-    expect(stdout).toContain('× 3 + 3 = 7')
-    expect(stdout).not.toContain('\n·x\n')
-
+    expect(stdout).toContain('\n·x\n')
     expect(stdout).toContain('Test Files  1 failed (1)')
     expect(stdout).toContain('Tests  1 failed | 1 passed')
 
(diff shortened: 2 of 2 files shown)


packages/vitest/src/node/reporters/dot.ts
```diff
@@ -70,10 +71,15 @@ export class DotReporter extends BaseReporter {
   }
 
   onTestCaseResult(test: TestCase): void {
+    const result = test.result().state
+    if (!this.isTTY && result !== 'pending') {
+      (this.ctx.logger.outputStream as Writable).write(formatTests([result]))
```
AriPerkkio, line 76: This is breaking the newlines of other logging quite badly. See how Node's logging and Vitest's `stderr`/`stdout` test logs are shown here:

image

Instead it should look like this:

image

test/reporters/tests/dot.test.ts
AriPerkkio, line 1: These duplicate tests can now be merged into a single `describe.each` block:


```diff
- describe('{ isTTY: true }', () => {
-   const isTTY = true
+ describe.each([true, false])('{ isTTY: %s }', (isTTY) => {

...

- describe('{ isTTY: false }', () => {
-   const isTTY = false
```

AriPerkkio (changes requested):

AriPerkkio (changes requested):
We definitely do not want to make changes to `logger.ts` just for `dot` reporter.

Instead, the `dot` reporter should maybe print multiple test results at once instead of one-by-one, when in non-TTY.

bstephen1:
I don't think there's a way to ensure the next console log will print with a prepending newline unless there's a wrapper like that in the logger to determine if the last write was a dot. If you print all the tests in a module at once then the next console log will still print on the same line since writing dots will never include a newline. The only way to avoid it completely would be to print all the tests at once at the end, where you could explicitly add the newline since the tests are done.

I would assume if you're running the dot reporter in CI you probably are suppressing console logs to have minimal output so I'm not sure if it would really be an issue to print them on the same line. I would think the typical output would just be the dots if everything passed. I tried it with the playwright dot reporter and it prints console logs inline too. 
playwright no html

AriPerkkio:
We should not try to intercept the console logs at all here. Instead, let's not write test results one-by-one on `onTestCaseResult` - let's instead write test cases in batches at `onTestModuleEnd`.

Removing this `isTTY` condition should make the reporter print tests as soon as there are enough finished tests to fill a single terminal row:

https://github.com/vitest-dev/vitest/blob/c00983c5a23497e92936535938b353cf0a7cf1af/packages/vitest/src/node/reporters/dot.ts#L80-L87

bstephen1:
Is that worth the tradeoff of not printing tests as they come in? You won't be able to check in on the test runner as it's running. It can take a long time to get through a whole row of dots, especially if you're testing browser code. You might not even have enough tests for a full row of dots.

I did some demos with other dot reporters and I couldn't find any that work like that. They all print tests as they finish and either print console logs in line or don't print them at all.

node -- ignores all console logs
node dot test

mocha -- prints inline
mocha dot test

cypress -- ignores all console logs
cypress dot reporter

playwright -- prints inline
playwright no html

japa -- prints inline
japa dot reporter

I would propose the behavior from the first commit that doesn't touch the logger, prints tests as they finish, and has no extra handling for console logs.

AriPerkkio:
> I would propose the behavior from the first commit that doesn't touch the logger, prints tests as they finish, and has no extra handling for console logs.

Yeah, you are right. There's no easy way to make this work perfectly. When we rewrote Vitest's reporters, this part was something we spent a lot of time optimizing in. But that was just for the TTY mode where we can control the terminal completely. 

Let's do this how it was in the first commit. It seems to be inline with other test runners. Thanks for testing other tools too! 💯

packages/vitest/src/node/reporters/dot.ts
```diff
@@ -70,10 +71,15 @@ export class DotReporter extends BaseReporter {
   }
 
   onTestCaseResult(test: TestCase): void {
+    const result = test.result().state
+    if (!this.isTTY && result !== 'pending') {
```
AriPerkkio:
```suggestion
    const result = test.result().state
    
    // On non-TTY the finished tests are printed immediately
    if (!this.isTTY && result !== 'pending') {
```

packages/vitest/src/node/reporters/dot.ts
```diff
@@ -30,11 +31,8 @@ export class DotReporter extends BaseReporter {
     }
   }
 
-  printTestModule(testModule: TestModule): void {
-    if (!this.isTTY) {
-      super.printTestModule(testModule)
-    }
-  }
+  // ignore base reporter
```
AriPerkkio:
```suggestion
  // Ignore default logging of base reporter
```

packages/vitest/src/node/reporters/dot.ts
```diff
@@ -70,10 +71,17 @@ export class DotReporter extends BaseReporter {
   }
 
   onTestCaseResult(test: TestCase): void {
+    const result = test.result().state
+
```
AriPerkkio:
```suggestion

```

PR #8012, opened 2025-05-22
feat: add iframe content access support

### LINKED ISSUE #6966: Browser Mode - Add an API for accessing content rendered in iframes

As a develoepr using Vitest browser mode, I want to be able to search for elements within an iframe that is rendered by tested code.
If my code under test renders an iframe, I currently have no way of accessing elements within that iframe since the `page` object is scoped to the global iframe where the component under test is rendered and any iframe rendered within this will not have it's content visible to the `page` object.

### Suggested solution

It would be good to have the ability to create a new `BrowserPage` from an iframe element:
```tsx
const iframeContent = '<p>text in the iframe</p>';
const ReactComponentThatRendersIframe = () => {
  return ( <iframe data-testid="the=iframe" srcdoc={iframeContent} /> )
}

it('renders an iframe with the correct text', async () => {
  render(<ReactComponentThatRendersIframe />);

  const myFrameElement = page.getByTestId('the-iframe');
  await expect.element(myFrameElement).toBeInTheDocument();
  
  const myFramePage: BrowserPage = makeMyFrameABrowserPage(myFrameElement); // Magic here

  // Now `myFramgePage` can be used in the exact same manner as `page`
  const textInIframe = myFramePage.getByText('text in the iframe');
  await expect.element(textInIframe).toBeInTheDocument();
})
```

### Alternative

The only way to get elements within the iframe at the minute would be to create a custom command and access the playwright APIs directly. There is no way of passing this element back to the test code however and the command would likely have to contain logic that should belong in the test.

### Additional context

_No response_

### Validations

starc007:
## add iframe content access support

- Add evaluate method to BrowserPage interface
- Implement evaluate in Playwright provider
- Add test cases for iframe functionality

fix #6966

### DIFF
--- a/packages/browser/context.d.ts
+++ b/packages/browser/context.d.ts
@@ -578,6 +578,18 @@ export interface BrowserPage extends LocatorSelectors {
    * @see {@link https://vitest.dev/guide/browser/locators}
    */
   elementLocator(element: Element): Locator
+  /**
+   * Create a new BrowserPage instance from an iframe element
+   * @param iframeElement The iframe element to create a page from
+   * @returns A new BrowserPage instance that can interact with the iframe's content
+   */
+  createIframePage(iframeElement: Element): Promise<BrowserPage>
+  /**
+   * Evaluates a function in the page context.
+   * @param fn Function to evaluate in the page context
+   * @returns The result of the function evaluation
+   */
+  evaluate<T>(fn: () => T): Promise<T>
 }
 
 export interface BrowserLocators {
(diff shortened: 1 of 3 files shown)


sheremet-va:
Did you just generate this with AI? This is not how Browser Mode works, there is nothing to review here really. You are mixing browser context with node context, and it seems like you never even ran the code.

starc007:
> Did you just generate this with AI? This is not how Browser Mode works, there is nothing to review here really. You are mixing browser context with node context, and it seems like you never even ran the code.
> 
> Why do you need the `evaluate` method when you already run code in the browser?

yeah, sorry. I misunderstood it, was thinking from playwright's page perspective

closing it !!

PR #8015, opened 2025-05-22
feat(reporter): add `onWritePath` option to `github-actions`

nwalters512:
### Description

This PR adds support for an `onWritePath` option for the GitHub Actions reporter. This can be used to customize the path that's written in [GitHub's annotation command format](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-an-error-message). This names was taken from `vitest-sonar-reporter` as suggested by this comment: https://github.com/vitest-dev/vitest/issues/8008#issuecomment-2896741513.

This is useful for situations where tests might be run such that their paths don't match the paths that GitHub sees. For instance, if the tests are being run in a container, the paths might start with `/app/`, which GitHub can't map to a specific file in order to attach the annotation.

Closes https://github.com/vitest-dev/vitest/issues/8008.

### Tests

### Documentation

### Changesets

### DIFF
--- a/test/reporters/tests/github-actions.test.ts
+++ b/test/reporters/tests/github-actions.test.ts
@@ -1,17 +1,41 @@
 import { resolve } from 'pathe'
-import { expect, test } from 'vitest'
+import { describe, expect, it } from 'vitest'
 import { GithubActionsReporter } from '../../../packages/vitest/src/node/reporters'
 import { runVitest } from '../../test-utils'
 
-test(GithubActionsReporter, async () => {
-  let { stdout, stderr } = await runVitest(
-    { reporters: new GithubActionsReporter(), root: './fixtures', include: ['**/some-failing.test.ts'] },
-  )
-  stdout = stdout.replace(resolve(__dirname, '..').replace(/:/g, '%3A'), '__TEST_DIR__')
-  expect(stdout).toMatchInlineSnapshot(`
+describe(GithubActionsReporter, () => {
+  it('uses absolute path by default', async () => {
+    let { stdout, stderr } = await runVitest(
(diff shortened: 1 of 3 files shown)


nwalters512:
Sorry, I hit Cmd-Enter before I was done updating the PR description. Working on that now!

test/reporters/tests/github-actions.test.ts
```diff
@@ -1,17 +1,41 @@
...
+        include: ['**/some-failing.test.ts'],
+        env: { GITHUB_WORKSPACE: '/home/user/workspace' },
+      },
+    )
+    stdout = stdout.replace(resolve(__dirname, '..').replace(/:/g, '%3A'), '__TEST_DIR__')
+    expect(stdout).toMatchInlineSnapshot(`
+      "
+      ::error file=__TEST_DIR__/fixtures/some-failing.test.ts,title=some-failing.test.ts > 3 + 3 = 7,line=8,column=17::AssertionError: expected 6 to be 7 // Object.is equality%0A%0A- Expected%0A+ Received%0A%0A- 7%0A+ 6%0A%0A ❯ some-failing.test.ts:8:17%0A%0A
+      "
+    `)
```
AriPerkkio, line 38: This assertion is 100% identical with `it('uses absolute path by default'`. We don't really see the usage of `onWritePath` here. Does this test pass locally as it's relying on `process.env.GITHUB_WORKSPACE`?

I think we could instead verify that some non-existing path written in `onWritePath` ends up in the report.

AriPerkkio (changes requested):

AriPerkkio (approved):
Looks good! Thanks for the PR @nwalters512! 💯

Issue #6265, opened 2024-08-02
Support visual regression testing in the browser


sheremet-va:
### Clear and concise description of the problem

There is no built-in way to compare images in the browser mode.

### Suggested solution

Implement a snapshot-style assertion to compare previously stored images. I am open to API ideas.

This issue is opened to start a discussion around the topic.

### Alternative

Related: https://github.com/vitest-dev/vitest/discussions/690

### Additional context

Consider using https://github.com/mapbox/pixelmatch as it is small and fast

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

AriPerkkio:
I'm not sure if this should be part of the core. Or at least it should be marked as very-very experimental all the time.

Comparing images is very hard. There are companies building products that especially focus on this, and using those can be expensive. Open source tools like `jest-image-snapshot` and `cypress-image-snapshot` (that use `pixelmatch`) are very flaky, especially when comparing images that are generated on different operating systems. For example pixel-by-pixel comparison fails in as simple cases as rendering fonts. 

If we can find a way to implement this confidently then I'm all in. 👍 

@JessicaSachs any thoughts? Do you have experience on this area?

herrlegno:
I tried to implement using `jest-image-snapshot` but no luck so far, it throws an error: 
![image](https://github.com/user-attachments/assets/c2c37e4e-d717-4202-9bf0-ae916b6a0dbb)

Here is the repo: https://github.com/herrlegno/vitest-browser-snapshot

m-lyon:
@herrlegno if you use node polyfills, e.g. via the [vite-plugin-node-polyfills](https://www.npmjs.com/package/vite-plugin-node-polyfills) plugin, then you should bypass that error. Though, the current outstanding issue with using `jest-image-snapshot` here is the use of `__dirname` as can be seen [here](https://github.com/americanexpress/jest-image-snapshot/blob/da49e578c9a02c50aa841d05ac64e9ffd4b725fd/src/diff-snapshot.js#L420), which gives the `ReferenceError: __dirname is not defined` error.

tsirlucas:
[...]

PR #8041, opened 2025-05-27
feat(browser): introduce `toMatchScreenshot` for Visual Regression Testing

macarie:
### Description

This PR introduces **initial support for Visual Regression Testing** for Vitest via a new `toMatchScreenshot` assertion.

**Related issue**: #6265

In this initial iteration:

- The feature supports PNG screenshots and uses `pixelmatch` as the comparator.
- The architecture is extensible: new comparators or codecs can be added easily, as long as they implement the expected interface.
- Comparators and codecs can be asynchronous.
- Screenshot comparison is executed in Node as a browser command, which allows for future adoption of native codecs or comparators.

The logic to get a stable screenshot follows Playwright's approach (with some differences):

1. Uses as baseline an optional reference screenshot or captures a new screenshot.
2. Takes a screenshot and compares it to the baseline.
3. If they match, the page is considered stable and the function returns.
4. If not, it continues with the latest screenshot as the baseline.
5. Repeats until stability is reached or a timeout is hit.

The command has 6 possible outcomes:

| Outcome | Description | Result |
| :------ | :---------- | :----: |
| `#01` | couldn't get a stable screenshot within timeout | `fail` |
| `#02` | stable screenshot taken, but no reference exists and update is not allowed | `fail` |
| `#03` | stable screenshot taken, no reference exists, but updates are allowed | `fail` |
| `#04` | stable screenshot taken on first try and matches reference | `pass` |
| `#05` | stable screenshot taken matches reference after retries | `pass` |
| `#06` | stable screenshot taken doesn't match reference (fallback case) | `fail` |

The client matcher expects an `Element` or `Locator` along with:

- An optional name for the screenshot
- An optional options object that includes:
  - Screenshot options, excluding conflicting ones
  - Comparator name and options
  - Timeout value

#### To-do
- [x] Fix screenshot paths and naming
- [x] Add global config for matcher, including `resolveScreenshotPath` and `resolveDiffPath` functions
- [x] Handle default config in one place
- [x] Use `annotation` API (#7953) to show screenshot paths
- [x] Support thresholds in `pixelmatch` comparator
- [x] Add documentation
- [x] Write test cases

### Tests

### Documentation

### Changesets


packages/browser/src/client/tester/expect/toMatchScreenshot.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  if (this.currentTestName === undefined) {
+    throw new Error('\'toMatchScreenshot\' cannot be used without test context')
+  }
+
+  // @todo add a counter after the name
+  const name
+    = typeof nameOrOptions === 'string' ? nameOrOptions : this.currentTestName
+
+  const result = await ensureAwaited(error =>
```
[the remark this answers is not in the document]
macarie: As the matcher itself is async, I thought it would still work if the `expect` is not awaited.

I wanted a safeguard, if possible, as `toMatchSnapshot` is not async, while this one is, so it might be easy to forget to `await` it.

packages/browser/src/node/commands/screenshotMatcher/index.ts
```diff
@@ -0,0 +1,312 @@
...
+  name,
+  screenshotOptions,
+}: {
+  codec: Codec
+  context: BrowserCommandContext
+  element: string
+  name: string
+  screenshotOptions: ScreenshotMatcherOptions['screenshotOptions']
+}) {
+  return (screenshot(
```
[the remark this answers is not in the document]
macarie: I agree, but unfortunately a conversion from base64 is needed somewhere, as the comparators usually work with raw bytes.

Do you think we could deprecate the `base64` option in favor of, let's say, `output?: 'buffer' | 'base64'`?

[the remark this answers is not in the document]
sheremet-va: To return a buffer back to the browser you need to convert it from the buffer to something else, which is again wasteful. And the browser can only receive typed array, the buffer is not implemented in browsers

packages/browser/src/client/tester/expect/toMatchScreenshot.ts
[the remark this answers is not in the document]
macarie: I ran into an issue here, @sheremet-va.

In `packages/expect/src/jest-extend.ts` (according to the debugger), it checks if `result` is an `instanceof Promise`. But the value returned by `ensureAwaited` is not, so it's assumed to be synchronous. Even if we made it return a proper `Promise`, the code still calls `.then` on it, which breaks the assumption that it was awaited by the user 😕

I'm thinking of dropping this change for now. The fix doesn't seem straightforward, and it's probably outside the scope of this PR. What do you think?

A possible solution might be to return an `ensureAwaited` inside the `.then` that's already being called, this way async matchers would be handled automatically. The downside is that this changes behavior as it would start the promise execution even if it's not explicitly awaited by the user.

macarie:
I moved the types around to what starts to hopefully make some sense. I created a `shared` folder for the types used in both browser and Node environments.

I've also rebased and resolved the conflicts from main.

packages/browser/jest-dom.d.ts
```diff
@@ -1,6 +1,7 @@
 // Disable automatic exports.
 
 import { ARIARole } from './aria-role.ts'
+import { ComparatorRegistry, ScreenshotMatcherOptions } from './context.js'
```
macarie, line 4: I know this creates a circular dependency between this file and `context.d.ts`, was wondering if it's fine tho.

To fix it I would have to move `ScreenshotOptions` out of `context.d.ts` and, as it's using `Locator`, at least half of the other types declared in it 😅

packages/vitest/src/node/types/browser.ts
```diff
@@ -236,6 +300,27 @@ export interface BrowserConfigOptions {
    * @default 30000
    */
   connectTimeout?: number
+
+  expect?: {
```
macarie, line 304: These will have to go somewhere else I believe, it was just to make TypeScript happy for now.

packages/browser/src/node/commands/screenshotMatcher/utils.ts
```diff
@@ -0,0 +1,281 @@
...
+    }
+  } {
+  if (context.testPath === undefined) {
+    throw new Error('`resolveOptions` has to be used in a test file')
+  }
+
+  const resolvedOptions = deepMerge<GlobalOptions>(
+    Object.create(null),
+    defaultOptions,
+    context.project.config.browser.expect?.toMatchScreenshot ?? {},
```
macarie, line 99: I don't think these are actually read, but will work on it later, I mostly wanted to move all the defaults and stuff in one place instead of having them spread across the project.

packages/browser/src/node/commands/screenshotMatcher/utils.ts
```diff
@@ -0,0 +1,281 @@
...
+ * @param input - The raw string to clean.
+ */
+function sanitizeArg(input: string): string {
+  return sanitize(relative(sep, join(sep, input)), true)
+}
+
+/**
+ * Retrieves the name of the browser from a given context.
+ */
+function getBrowserName(context: BrowserCommandContext): string {
```
sheremet-va, line 221: You can just get the name from `provider.config.browser.name`

packages/browser/src/client/tester/expect/toMatchScreenshot.ts
```diff
@@ -0,0 +1,105 @@
...
+        element: convertElementToCssSelector(
+          getElementFromUserInput(actual, toMatchScreenshot, this),
+        ),
+        ...options,
+      },
+    ] satisfies ScreenshotMatcherArguments,
+  )
+
+  if (result.pass === false) {
+    const test = getCurrentTest()
```
sheremet-va: You already have the current test in `getWorkerState().current`, you don't need to call for it again

packages/browser/context.d.ts
```diff
@@ -44,6 +45,129 @@ export interface ScreenshotOptions {
   save?: boolean
 }
 
+export interface ComparatorRegistry {
```
sheremet-va: Maybe this should have a `Screenshot` as a prefix?

packages/browser/jest-dom.d.ts
sheremet-va: If it works, it works 😄

packages/vitest/src/node/config/resolveConfig.ts
```diff
@@ -56,7 +56,7 @@ function parseInspector(inspect: string | undefined | boolean | number) {
   return { host, port: Number(port) || defaultInspectPort }
 }
 
-export function resolveApiServerConfig<Options extends ApiConfig & UserConfig>(
+export function resolveApiServerConfig<Options extends ApiConfig & Omit<UserConfig, 'expect'>>(
```
macarie, line 60: This is provided with the `browser` object, the new `browser.expect` property conflicts with `expect`. I noticed `options.expect` is not used tho, so instead of removing it at runtime I removed it from the type.

packages/vitest/src/node/types/browser.ts
```diff
@@ -236,6 +236,21 @@ export interface BrowserConfigOptions {
...
+      {
+        /**
+         * The name of the comparator to use for visual diffing.
+         *
+         * @defaultValue `'pixelmatch'`
+         */
+        comparatorName?: ComparatorName
+        comparatorOptions?: ToMatchScreenshotComparators[ComparatorName]
+      }
+    }[keyof ToMatchScreenshotComparators] & ToMatchScreenshotOptions
```
macarie, line 250: This type is a bit chaotic, but without generics the only solution is to create discriminated unions.

The interface is compatible with `ScreenshotComparatorRegistry`, so I used that to extend it in the providers. In the future this will allow to add more comparators and have them typed correctly in the config by augmenting `ScreenshotComparatorRegistry`.

packages/browser/context.d.ts
```diff
@@ -1,4 +1,5 @@
 import { SerializedConfig } from 'vitest'
+import type pm from 'pixelmatch'
```
sheremet-va: I don't think this is used anywhere in this file

sheremet-va (commented):
Code looks good to me, but we need a lot more tests for this feature

test/browser/fixtures/expect-dom/toMatchScreenshot.test.ts
```diff
@@ -16,18 +16,77 @@ const renderTestCase = (colors: readonly [string, string, string]) =>
...
+        .*?[\\/]rgb-box-actual-\\w+\\.png
+    `)))
+  })
+
+  test('throws when creating a screenshot for the first time', async ({ onTestFinished }) => {
+    let path: string | undefined
+
+    onTestFinished(async () => {
+      if (path) {
+        server.commands.removeFile(path)
```
sheremet-va: This needs to be awaited, by the way

macarie:
I changed the testing approach: instead of relying on pre-generated screenshots, the tests now create them dynamically during execution.

This makes the tests more stable and avoids failures across different OSes, browser versions, or rendering environments. All artifacts are cleaned up automatically after the tests run.

~~One thing still missing is a test for watch mode. I ran into some issues with it and will take another shot at getting it to work tomorrow.~~ ➡️ Done.

docs/guide/browser/assertion-api.md
```diff
@@ -1046,3 +1046,216 @@ await expect.element(queryByTestId('parent')).toHaveSelection('ected text')
...
+However, browser rendering can vary across:
+
+- Different browsers and browser versions
+- Operating systems (Windows, macOS, Linux)
+- Screen resolutions and pixel densities
+- GPU drivers and hardware acceleration
+- Font rendering and system fonts
+
+It is recommended to read the
+[Visual Regression Testing guide](/) to
```
macarie: 🔔 Reminder: change this back once the guide is available

packages/browser/src/node/commands/screenshotMatcher/index.ts
```diff
@@ -0,0 +1,248 @@
...
+    )).pass
+
+    decodedBaseline = image2
+
+    if (comparatorResult) {
+      break
+    }
+
+    retries += 1
+  }
```
macarie, line 240: Playwright uses a backoff mechanism based on predefined timeouts (`[0, 100, 250, 500]`, fallbacks to `1_000`) when retrying.

I'm not sure if I should implement that, if each tests waits 100ms to take a screenshot again, it adds up.

Right now, it's surprisingly hard **not to** get a stable screenshot in a few retries, but this doesn't mean the timing is consistent across runs. When using Playwright as a provider, consistency should not be an issue because they allow disabling animations. With WebdriverIO I don't know as I have no experience with it.

pnpm-lock.yaml
macarie, line 1: I'm not sure why there are so many changes here. I tried resetting this file to `main`, but as soon as I add any of the dependencies I use, all these changes show up.

test/browser/specs/to-match-screenshot.test.ts
macarie, line 1: Is this the correct location for this test?

sheremet-va: Yes!

docs/guide/browser/assertion-api.md
```diff
@@ -1046,3 +1046,216 @@ await expect.element(queryByTestId('parent')).toHaveSelection('ected text')
 await expect.element(queryByTestId('prev')).not.toHaveSelection()
 await expect.element(queryByTestId('next')).toHaveSelection('ne')
 ```
+
+## toMatchScreenshot <Badge type="warning">experimental</Badge> <Version>3.3.0</Version> {#tomatchscreenshot}
```
macarie: I guess this version will have to change 😛

Will do before merging, note that this version is written in `config.md` too.

macarie:
I will have to debug the failing Webdriverio test on my Windows PC, looks like `arg` is not getting sanitized for some reason. I also don't understand why it fails with `ENOENT` 🙃

packages/browser/src/node/commands/screenshotMatcher/index.ts
```diff
@@ -0,0 +1,246 @@
...
+      diff: null,
+      message: `Could not capture a stable screenshot within ${timeout}ms.`,
+    }
+  }
+
+  const { updateSnapshot } = context.project.serializedConfig.snapshotOptions
+
+  // if there's no reference or if we want to update snapshots, we have to finish the comparison early
+  if (reference === null || updateSnapshot === 'all') {
+    const shouldCreateReference = updateSnapshot !== 'none'
```
macarie, line 69: Another question from the notes: should this check if ref and actual match and eventually set the flag to false? Would it even make sense to update screenshots if they still match?

docs/guide/browser/visual-regression-testing.md
macarie, line 1: Will add a CI section here, either today or tomorrow. Got a minimalistic workflow that can work fine by running on GitHub without relying on any third party services.

docs/guide/browser/visual-regression-testing.md
```diff
@@ -0,0 +1,546 @@
...
+```yaml [.github/workflows/update-screenshots.yml]
+name: Update Visual Regression Screenshots
+
+on:
+  workflow_dispatch: # manual trigger only
+
+env:
+  AUTHOR_NAME: 'github-actions[bot]'
+  AUTHOR_EMAIL: '41898282+github-actions[bot]@users.noreply.github.com'
+  COMMIT_MESSAGE: 'test: update visual regression screenshots'
```
macarie: todo: add `co-authored-by`

docs/guide/browser/visual-regression-testing.md
```diff
@@ -0,0 +1,546 @@
...
+- Whether you're running headless or not
+- Browser settings and versions
+- ...and honestly, sometimes just the phase of the moon
+
+That's why Vitest includes the browser and platform in screenshot names (like
+`button-chromium-darwin.png`).
+
+For stable tests, use the same environment everywhere. We **strongly recommend**
+cloud services like
+[Microsoft Playwright Testing](https://azure.microsoft.com/en-us/products/playwright-testing)
```
sheremet-va, line 48: Can you actually run Vitest tests there?

macarie: @sheremet-va, yes, the config looks like this:
```ts
instances: [
	{
		browser: "chromium",
		connect: {
			wsEndpoint: `${env.PLAYWRIGHT_SERVICE_URL}?cap=${JSON.stringify({
				os: "linux",
				runId: `Vitest ${env.CI ? 'CI' : 'local'} run @${new Date.toISOString()}`,
			})}`,
			options: {
				exposeNetwork: "<loopback>",
				headers: {
					"x-mpt-access-key": env.PLAYWRIGHT_SERVICE_ACCESS_TOKEN,
				},
				timeout: 30_000,
			},
		},
	},
],
```

Playwright (well, Vitest in this case) keeps running locally and exposes the server over the network. The service only offers access to browser instances, but everything else is local, which is the main reason we chose Playwright at work 😄

sheremet-va: I didn't know it's possible! I wonder if we can mention this somewhere (unless I missed it?)

macarie: No, I didn't mention it explicitly anywhere. Where would you place it? As new section beneath "Running on GitHub Actions"?

sheremet-va: > As new section beneath "Running on GitHub Actions"?

Sounds good to me. Maybe use a `::: group` for these examples?

macarie: Is this what you had in mind? Maybe I misunderstood, but I think it looks great with the two tabs explaining the two approaches 😄

macarie:
Just tested this PR by replacing Playwright's component testing with Vitest in _one of_ our visual regression test suites at work. Pretty impressed with how it went!

The speed difference is huge:

- Vitest: 4m46s
  Vitest's result showing it run in 4m46s

- Playwright: 34m41s
  Playwright's result showing it run in 34m41s

Not sure why Playwright is taking so long here. This was with just 1 browser instance, we usually run 20 in parallel. Our runs take less than 2 minutes to complete, but the consumed _browser-time_ amounts to ~35m each time, so this result aligns well with what we've always observed. Worth noting: Playwright runs a few more tests, those are non-visual tests and only add 1m40s.

About the failing tests in Vitest:

- half failed because this PR handles some _special_ characters in filenames differently
- the other half looks like an actual bug, I'm gonna dig into this one a bit more. Seems like the tests are running _too fast_ and not waiting for the screenshots to complete. It's happening with our button tests where we capture default and hover states, but the first screenshot is always showing the hover state tho.


Overall tho, really minimal changes needed to get this working!

docs/guide/browser/visual-regression-testing.md
```diff
@@ -0,0 +1,716 @@
...
+Free for open source, works with any browser provider, and you control
+everything.
+
+The downside? That "works on my machine" conversation when someone generates
+screenshots locally and they don't match CI expectations anymore.
+
+Microsoft's service shines when you need developers to run visual tests locally.
+
+Maybe you've got designers who want to verify their changes, or developers who
+prefer catching issues before pushing. Yes, it costs money, but compared to the
```
sheremet-va: This reads like we are selling it. Can we rephase it?

docs/guide/browser/visual-regression-testing.md
```diff
@@ -0,0 +1,716 @@
+---
+title: Visual Regression Testing
+outline: [2, 3]
+---
+
+# Visual Regression Testing
+
+Since version 3.3, Vitest can run visual regression tests out of the box. It
```
sheremet-va: This will be part of Vitest 4 release. In major versions we don't mention "since ..."

docs/guide/browser/config.md
```diff
@@ -325,3 +325,152 @@ The timeout in milliseconds. If connection to the browser takes longer, the test
 ::: info
 This is the time it should take for the browser to establish the WebSocket connection with the Vitest server. In normal circumstances, this timeout should never be reached.
 :::
+
+## browser.expect
+
+- **Type:** `ExpectOptions`
+
+### browser.expect.toMatchScreenshot <Version>3.3.0</Version> {#browser-expect-tomatchscreenshot}
```
sheremet-va: Version can be ommited

docs/guide/browser/assertion-api.md
```diff
@@ -1046,3 +1046,216 @@ await expect.element(queryByTestId('parent')).toHaveSelection('ected text')
 await expect.element(queryByTestId('prev')).not.toHaveSelection()
 await expect.element(queryByTestId('next')).toHaveSelection('ne')
 ```
+
```
sheremet-va, line 1070: Version can be omitted

packages/browser/src/client/tester/expect/toMatchScreenshot.ts
```diff
@@ -0,0 +1,102 @@
+import type { AsyncExpectationResult, MatcherState } from '@vitest/expect'
```
sheremet-va, line 1: `@vitest/expect` is not part of Vitest browser dependencies, you can only import from `vitest` directly

packages/browser/src/client/tester/expect/toMatchScreenshot.ts
```diff
@@ -0,0 +1,102 @@
...
+import type { ScreenshotMatcherOptions } from '../../../../context'
+import type { ScreenshotMatcherArguments, ScreenshotMatcherOutput } from '../../../shared/screenshotMatcher/types'
+import type { Locator } from '../locators'
+import { getCurrentTest } from '@vitest/runner'
```
sheremet-va: `@vitest/runner` is not part of Vitest browser dependencies, you can only import from `vitest` directly

packages/browser/src/node/commands/screenshotMatcher/utils.ts
```diff
@@ -0,0 +1,266 @@
...
+            arg: `${resolvePathData.arg}-actual`,
+          }),
+          diff: resolvedOptions.resolveDiffPath({
+            ...resolvePathData,
+            arg: `${resolvePathData.arg}-diff`,
+          }),
+        }
+
+        // @ts-expect-error we set the value back right beneath
+        delete this.diffs
```
sheremet-va: Let's avoid using `delete`, can we just use `Object.defineProperty` here?

test/test-utils/index.ts
```diff
@@ -209,7 +209,10 @@ async function runCli(command: 'vitest' | 'vite-node', _options?: CliOptions | s
   if (args[0] !== 'list' && (args.includes('--watch') || args[0] === 'watch')) {
     if (command === 'vitest') {
       // Wait for initial test run to complete
-      await cli.waitForStdout('Waiting for file changes')
+      await Promise.race([
+        cli.waitForStdout('Waiting for file changes'),
+        cli.waitForStdout('Watching for file changes'),
```
sheremet-va: What happened here? `Watching for file changes` is printed _after_ the test has rerun, but this check waits for the first run. This looks like an attempt to cover a bug

test/browser/fixtures/expect-dom/toMatchScreenshot.test.ts
```diff
@@ -0,0 +1,303 @@
...
+      <div style="background-color: ${colors[1]}; width: var(--size);"></div>
+      <div style="background-color: ${colors[2]}; width: var(--size);"></div>
+    </div>
+  `)
+
+describe('.toMatchScreenshot', () => {
+  test('compares screenshots correctly', async ({ onTestFinished }) => {
+    const filename = globalThis.crypto.randomUUID()
+    const path = join(
+      import.meta.dirname,
```
sheremet-va: `import.meta.dirname` is `undefined` in the browser

test/browser/fixtures/expect-dom/toMatchScreenshot.test.ts
```diff
@@ -0,0 +1,303 @@
...
+  render(`
+    <div style="--size: ${blockSize}px; display: flex; justify-content: center; height: var(--size); width: calc(var(--size) * ${blocks});" data-testid="${dataTestId}">
+      <div style="background-color: ${colors[0]}; width: var(--size);"></div>
+      <div style="background-color: ${colors[1]}; width: var(--size);"></div>
+      <div style="background-color: ${colors[2]}; width: var(--size);"></div>
+    </div>
+  `)
+
+describe('.toMatchScreenshot', () => {
+  test('compares screenshots correctly', async ({ onTestFinished }) => {
```
sheremet-va, line 30: this test doesn't seem to test anything, it will always suceed because the path is always new

packages/browser/src/client/tester/expect/toMatchScreenshot.ts
```diff
@@ -0,0 +1,102 @@
...
+      {
+        element: convertElementToCssSelector(
+          getElementFromUserInput(actual, toMatchScreenshot, this),
+        ),
+        ...options,
+      },
+    ] satisfies ScreenshotMatcherArguments,
+  )
+
+  if (result.pass === false) {
```
sheremet-va: These annotations have a wrong location. Culprint: https://github.com/macarie/vitest/blob/0600986af7d753b50df2db671358b13c069d6e0f/packages/runner/src/context.ts#L207

It should look more like this: https://github.com/macarie/vitest/blob/0600986af7d753b50df2db671358b13c069d6e0f/packages/runner/src/suite.ts#L891

sheremet-va: After this change, the annotations are shown correctly, but the content is not properly visible (might be bug in annotations)

test/browser/fixtures/expect-dom/toMatchScreenshot.test.ts
macarie: The idea here is to test default behavior once a screenshot exists (and also that the function checks the correct reference, as the value is random). Takes a screenshot that will be used as reference, then checks that the element still matches it.

I've done it this way because with a stored reference the pipeline was failing, and with UI enabled it wasn't working either with a reference created in headless (or viceversa).

packages/browser/src/client/tester/expect/toMatchScreenshot.ts
macarie: Now that I checked, this is true, but all the files in `packages/browser/src/client/tester/expect` import from `@vitest/expect`.

Where should I import these types from?

test/test-utils/index.ts
macarie: I'm testing that the test fails, and the CLI output is this:

```
 DEV  v4.0.0-beta.2 /.../vitest/test/browser/vitest-test-b7af4dc4-e2e5-4bc8-9f78-1ebbbcaa26b2

 × |chromium| basic.test.ts > screenshot-snapshot 165ms
   → expect(element).toMatchScreenshot()

No existing reference screenshot found; a new one was created. Review it before running tests again.

Reference screenshot:
  /.../vitest/test/browser/vitest-test-b7af4dc4-e2e5-4bc8-9f78-1ebbbcaa26b2/__screenshots__/basic.test.ts/screenshot-snapshot-1-chromium-darwin.png

   ❯ ../../../packages/browser/dist/expect-element.js:26:2329 notice
     ↳ Reference screenshot


 Test Files  1 failed (1)
      Tests  1 failed (1)
   Start at  00:15:29
   Duration  761ms (transform 0ms, setup 0ms, collect 9ms, tests 166ms, environment 0ms, prepare 603ms)

 FAIL  Tests failed. Watching for file changes...

```

Am I using the wrong reporter or is the function only meant to run tests that succeed?

docs/guide/browser/visual-regression-testing.md
macarie: True, I might have gotten a bit too carried away 😄 I love the service and I think it has one of the best price/value ratios ever, solves so many problems.

I toned it down a bit, is it better?

packages/browser/src/node/commands/screenshotMatcher/utils.ts
macarie: Done, but I'm re-evaluating the need for a lazy getter here. Supposedly, it's concatenating three strings, do you think it makes sense?

packages/browser/src/client/tester/expect/toMatchScreenshot.ts
macarie: Sorry, not sure I follow here. Is this a _note to self_ or should I change it?

packages/browser/src/node/commands/screenshotMatcher/utils.ts
sheremet-va: I don't know 🤷 If they are always used, just return them right away

packages/browser/src/client/tester/expect/toMatchScreenshot.ts
sheremet-va: Interesting! Since they are bundled, it doesn't really matter what is used here, so I guess we can keep it as is

test/test-utils/index.ts
sheremet-va: Interesting. It is ok to await stdout then, but the string should include "Tests failed", and let's have a comment like `// waiting for either success or failure`

test/browser/fixtures/expect-dom/toMatchScreenshot.test.ts
sheremet-va: Ok, now I see what this test does. Can you add a small comment explaining the test? Why are we calling `locator.screenshot` with these values

packages/browser/src/client/tester/expect/toMatchScreenshot.ts
sheremet-va: It is both. I will fix the file path in a separate PR, but I don't know why the images are not visible in UI, which I defer to you 😄

sheremet-va: Annotations are now visible in the UI where the `expect().toMatchSnapshot` is called (#8315)

packages/browser/src/node/commands/screenshotMatcher/utils.ts
macarie: They're only used when the test fails, so hopefully most of the time they should not get created. I guess we can keep it, if it becomes an issue it can easily be removed anyway.

test/browser/fixtures/expect-dom/toMatchScreenshot.test.ts
macarie: I added a comment explaining how the tests work above the `describe` and a small explanation of what's going on in the first and second tests.

Let me know if it works.

macarie:
Let me know if you don't want the merge commit in the PR's history, I will rebase and remove it once we're done with the review.

Force-pushing after a rebase moves comments up in the "load more" hidden section every time I update from `main` 😄

sheremet-va (dismissed):
This looks good to me! Thanks for working on this 👍

sheremet-va (approved):

PR #8071, opened 2025-06-02
fix(vite-node): unable to handle errors where sourcemap mapping empty

blake-newman:
https://github.com/vitejs/vite/pull/19226 introduced a change where by empty source mappings are not combined. As such when vitest has an internal error caused such as using jsdom feature that is not implemented which logs an error it tries to map the source map when the mapping is empty which results in an error such as:

```
Cannot read properties of undefined (reading 'map')
 ❯ VirtualConsole.<anonymous> node_modules/.pnpm/jsdom@26.1.0/node_modules/jsdom/lib/jsdom/virtual-console.js:29:53
 ❯ module.exports node_modules/.pnpm/jsdom@26.1.0/node_modules/jsdom/lib/jsdom/browser/not-implemented.js:12:26
 ❯ HTMLAudioElementImpl.load node_modules/.pnpm/jsdom@26.1.0/node_modules/jsdom/lib/jsdom/living/nodes/HTMLMediaElement-impl.js:112:5
 ❯ HTMLAudioElement.load node_modules/.pnpm/jsdom@26.1.0/node_modules/jsdom/lib/jsdom/living/generated/HTMLMediaElement.js:109:34
```

When the error originates from a file without source mapping such as `x.vue` it errors instead of logging original error.

---

### Tests

### Documentation

### Changesets


blake-newman:
Looks like this could be related:

https://github.com/vitest-dev/vitest/issues/5380

hi-ogawa (commented):
Good catch. I think I have some idea how this goes wrong, but reproduction would be great :pray: Is the error happening during `new TraceMap({ mappings: '' })`?

Also I'm not sure whether we should fix it here or maybe we can change Vite `ssrTransform` to return more robust `map`.

blake-newman:
@hi-ogawa here is reproduction. Sorry took me a while to reproduce myself:

https://github.com/blake-newman/vitest-dev-vitest-se4padzq

Turns out it's specific to when using TSX

hi-ogawa (approved):
Thanks for the fix! I added a minimal repro as a test case. The same issue didn't reproduce on Vite SSR / module runner because Vite's handling was internally more robust (e.g. checking `"version" in map` and also fallback `sources ?? []` to avoid `undefined.map`).

Issue #8018, opened 2025-05-23
The test name can differ in the base reporter in the task errors


lafeuil:
### Clear and concise description of the problem

I would like to display some information in the report with the [test metadata](https://vitest.dev/advanced/metadata.html#task-metadata). Therefore, when overloading the default reporter to rename the test by overriding the `getTestName` function in the `BaseReporter` class, I can easily add some metadata in the console.

But the `getTestName` function is not used in the `printTaskErrors` function. So, the metadata information are not displayed in the errors summary. The [getFullname](https://github.com/vitest-dev/vitest/blob/ac88181a046c9dbe4684327187411e9c21ec5208/packages/vitest/src/node/reporters/base.ts#L613) function from the `@vitest/runner/utils` module.

### Suggested solution

Use the `getTestName` function in the `BaseReporter` class in the `printTaskErrors` function. 

Or, create a new `getFullName` function in the `BaseReporter` class to use it in the `printTaskErrors` function. 

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vitest.dev/guide/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

sheremet-va:
We can add `getFullName` as a method on the base reporter.

PR #8115, opened 2025-06-06
fix(reporter): `task.meta` should be available in custom reporter's errors

AriPerkkio:
### Description

- Fixes https://github.com/vitest-dev/vitest/issues/8018

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/reporters/base.ts
+++ b/packages/vitest/src/node/reporters/base.ts
@@ -249,6 +249,10 @@ export abstract class BaseReporter implements Reporter {
     return getTestName(test, separator)
   }
 
+  protected getFullName(test: Task, separator?: string): string {
+    return getFullName(test, separator)
+  }
+
   protected formatShortError(error: ErrorWithDiff): string {
     return `${F_RIGHT} ${error.message}`
   }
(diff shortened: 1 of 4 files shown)


test/reporters/tests/default.test.ts
```diff
@@ -238,6 +239,21 @@ describe('default reporter', async () => {
...
+        return `${separator}{ name: ${test.name}, meta: ${test.meta.custom} } (Custom getFullName here)`
+      }
+    }
+
+    const { stderr } = await runVitest({
+      root: 'fixtures/metadata',
+      reporters: new Custom(),
+    })
+
+    expect(stderr).toMatch('FAIL   > { name: fails, meta: Failing test added this } (Custom getFullName here')
```
AriPerkkio, line 255: @lafeuil from https://github.com/vitest-dev/vitest/issues/8018 👋 

Is this the kind of use case you are trying to achieve?

lafeuil: Yes, it is exactly that :+1:

packages/vitest/src/node/reporters/base.ts
```diff
@@ -249,6 +249,10 @@ export abstract class BaseReporter implements Reporter {
     return getTestName(test, separator)
   }
 
+  protected getFullName(test: Task, separator?: string): string {
```
sheremet-va, line 252: One thing I don't like here is the use of `Task` since it's an experimental API, but to have the reported entity, we would have to rewrite the reporter which is probably outside of the scope of this PR

AriPerkkio: Yeah, this is inline with `protected getTestName(test: Task, separator?: string)` above.

sheremet-va: Yeah, but the previous one was there before the new API

PR #8118, opened 2025-06-07
chore: update deprecated workspace file log

btea:
### Description
I think it is more friendly to specify the specific level of the `projects` field in the printed log information.

However, I also wonder if some users will misunderstand `test.projects` as a field. Look forward to your opinions.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/core.ts
+++ b/packages/vitest/src/node/core.ts
@@ -477,7 +477,7 @@ export class Vitest {
       : 'the root config file'
 
     this.logger.deprecate(
-      `The workspace file is deprecated and will be removed in the next major. Please, use the \`projects\` field in ${configFile} instead.`,
+      `The workspace file is deprecated and will be removed in the next major. Please, use the \`test.projects\` field in ${configFile} instead.`,
     )
 
     const workspaceModule = await this.import<{


AriPerkkio (commented):
If we want to use this style, it should be applied here too:

https://github.com/vitest-dev/vitest/blob/c69be1fc1ab5d413661a7bead5bffd7147f4aa97/packages/vitest/src/node/core.ts#L433-L435

https://github.com/vitest-dev/vitest/blob/c69be1fc1ab5d413661a7bead5bffd7147f4aa97/packages/vitest/src/node/core.ts#L448-L450

btea:
Thanks for reviewing. Looking forward to other people's comments, if you agree with the changes, I will update them.

sheremet-va:
> If we want to use this style, it should be applied here too:
> 
> https://github.com/vitest-dev/vitest/blob/c69be1fc1ab5d413661a7bead5bffd7147f4aa97/packages/vitest/src/node/core.ts#L433-L435
> 
> https://github.com/vitest-dev/vitest/blob/c69be1fc1ab5d413661a7bead5bffd7147f4aa97/packages/vitest/src/node/core.ts#L448-L450

I agree

PR #8145, opened 2025-06-11
fix(expect): handle async errors in expect.soft

### LINKED ISSUE #8092: expect.soft does not work together with an async expectation result
### Describe the bug

When using `expect.soft(actual)` combined with a `.to<Something>` that returns an `AsyncExpectationResult`, the test gets stopped after the first failure.

Initially found this trying to use `vitest-plugin-vis` matcher `toMatchImageSnapshot` together with `expect.soft` to enable getting multiple image comparison result files for a single test case.

A little bit of research (though not sure if it would be the right place to fix things) lead me to [jest-extend.ts, line 83](https://github.com/vitest-dev/vitest/blob/main/packages/expect/src/jest-extend.ts#L85) where a promise is augmented to throw an expectation error on failure, instead of handling it similar to [wrapAssertion](https://github.com/vitest-dev/vitest/blob/main/packages/expect/src/utils.ts#L76) where the `test.result.errors` is augmented when using `expect.soft`.

### Reproduction

StackBlitz here: https://stackblitz.com/edit/vitest-dev-vitest-pvde5zau

Reproduction in basic.test.ts (two tests, comparing an async assertion with a sync one)

### System Info

```shell
Windows and linux (probably any system), with vitest 3.2.1
```

### Used Package Manager

pnpm

### Validations

lzl0304:
### Description

Fixes #8092 

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/expect/src/utils.ts
+++ b/packages/expect/src/utils.ts
@@ -1,5 +1,6 @@
 import type { Test } from '@vitest/runner/types'
 import type { Assertion } from './types'
+import { noop } from '@vitest/utils'
 import { processError } from '@vitest/utils/error'
 
 export function createAssertionMessage(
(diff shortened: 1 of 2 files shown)


packages/expect/src/utils.ts
```diff
@@ -95,13 +102,18 @@ export function wrapAssertion(
     }
 
     try {
-      return fn.apply(this, args)
+      const result = fn.apply(this, args)
+
+      if (result && typeof result === 'object' && result instanceof Promise) {
```
sheremet-va: I thought I already fixed that, but maybe I forgot. Can you change the check from `instanceof Promise` to `typeof result.then === 'function'` here and in `jest-extend.ts`? 

And use `then(noop, (err) => { /* ... */ })` here instead

This is needed to support `PromiseLike` (we already do this in all other places except here)

sheremet-va (changes requested):
Looks very good, thank you! Just one little thing

packages/expect/src/utils.ts
lzl0304: The updates have been submitted. Thank you for your suggestion!

sheremet-va (approved):

sheremet-va:
LGTM! Thank you!

PR #8159, opened 2025-06-13
feat: provide entity to onConsoleLog

### LINKED ISSUE #8156: Add name of the test case and test module in `test.onConsoleLog` callbacks

I want to easily silence `console.error`'s in `vitest.config.ts`. I know the test files that can cause these errors. Now I'm parsing test filename from stacktrace from `log`.

### Suggested solution

Not yet sure how the arguments should be passed. Maybe a `meta: { filename: string, testname?: string }`?

```ts

export default defineConfig({
  test: {
    onConsoleLog(log, type, meta) {
      if (meta.filename === 'math.test.ts' && meta.testName === 'sum should add 1 + 1') {
        return false
      }
    },
})
```

### Alternative

-

### Additional context

-

### Validations

sheremet-va:
### Description

Fixes #8156

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/reporters/base.ts
+++ b/packages/vitest/src/node/reporters/base.ts
@@ -433,11 +433,13 @@ export abstract class BaseReporter implements Reporter {
       return false
     }
 
-    const task = log.taskId ? this.ctx.state.idMap.get(log.taskId) : undefined
-    const entity = task && this.ctx.state.getReportedEntity(task)
-    const shouldLog = this.ctx.config.onConsoleLog?.(log.content, log.type, entity)
-    if (shouldLog === false) {
-      return shouldLog
+    if (this.ctx.config.onConsoleLog) {
+      const task = log.taskId ? this.ctx.state.idMap.get(log.taskId) : undefined
+      const entity = task && this.ctx.state.getReportedEntity(task)
+      const shouldLog = this.ctx.config.onConsoleLog(log.content, log.type, entity)
+      if (shouldLog === false) {
+        return shouldLog
+      }
     }
     return true
   }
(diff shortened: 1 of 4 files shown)


packages/vitest/src/node/reporters/base.ts
```diff
@@ -433,7 +433,9 @@ export abstract class BaseReporter implements Reporter {
       return false
     }
 
-    const shouldLog = this.ctx.config.onConsoleLog?.(log.content, log.type)
+    const task = log.taskId ? this.ctx.state.idMap.get(log.taskId) : undefined
+    const entity = task && this.ctx.state.getReportedEntity(task)
+    const shouldLog = this.ctx.config.onConsoleLog?.(log.content, log.type, entity)
```
AriPerkkio: Nitpick, but could avoid `getReportedEntity` and `idMap.get` calls:

```ts
let shouldLog = true;

if(this.ctx.config.onConsoleLog) {
  const task = log.taskId ? this.ctx.state.idMap.get(log.taskId) : undefined
  const entity = task && this.ctx.state.getReportedEntity(task)
  shouldLog = this.ctx.config.onConsoleLog(log.content, log.type, entity)
}

sheremet-va: We need to decide in what version we will release this. Technically this is a feature, but I am not planning to have 3.3 (or I guess we can have a mini release?)

AriPerkkio: Maybe just leave it for 4.0

AriPerkkio (approved):

PR #8168, opened 2025-06-16
fix: extend `birpc` timeout if `testTimeout` exceeds the default

AriPerkkio:
### Description

- Fixes https://github.com/vitest-dev/vitest/issues/8164

Users should be able to block test runner thread for longer than `birpc`'s default timeout when `testTimeout` is set to a higher value.

### Tests

### Documentation

### Changesets


sheremet-va:
rpc messaging has nothing to do with the test, it should not be configured by the same value

AriPerkkio:
> rpc messaging has nothing to do with the test, it should not be configured by the same value

Sure it does. As rpc handles events of test, its timeouts affect code that runs in tests.

sheremet-va:
> > rpc messaging has nothing to do with the test, it should not be configured by the same value
> 
> 
> 
> Sure it does. As rpc handles events of test, its timeouts affect code that runs in tests. 

It handles events of everything, not just tests. Only onTaskUpdate is related to the test itself, and it happens outside of the test body even

AriPerkkio:
> It handles events of everything, not just tests. 

Exactly, and test is part of this "everything". Any configurable timeout should be used when configurating test runner's RPC. I'll add other timeouts to `Math.max` too.

sheremet-va:
> > It handles events of everything, not just tests. 
> 
> 
> 
> Exactly, and test is part of this "everything". Any configurable timeout should be used when configurating test runner's RPC. I'll add other timeouts to `Math.max` too. 

Test is only a part of it, you can't use testTimeout to cover it, it makes zero sense. What is in the issue is an edge case that needs special handling, not a general solution.

AriPerkkio:
> Test is only a part of it, you can't use testTimeout to cover it, it makes zero sense.

Any configurable timeout should affect RPC timeouts. I feel like I'm missing something essential here. 🤔

sheremet-va:
> > Test is only a part of it, you can't use testTimeout to cover it, it makes zero sense.
> 
> 
> 
> Any configurable timeout should affect RPC timeouts. I feel like I'm missing something essential here. 🤔 

The `testTimeout` affects only the test body, and there is no calls that are called inside of test body (except annotations). When test is sync and blocks the thread, it is an edge case that needs more investigation

PR #8183, opened 2025-06-18
feat(cli): restart browsers with ‘b’ shortcut

### LINKED ISSUE #7678: Use `b` in watch mode to restart the browser server

When running in watch mode, sometimes the browser instance is in a bad state, causing all browser tests to time out.

I have to quit the watch mode and restart.

### Suggested solution

Instead of having `b` to do `press b to start the browser server if not started yet`,
use it to restart the browser instance(s) so we don't need to restart watch mode.

### Alternative

Automatically detect dead browser and restart. To do that need to add a special `heartbeat` command to ensure the browser is still in good shape. However, determine what does "good shape" means could be challenging.

### Additional context

_No response_

### Validations

pengooseDev:
### Description
Closes #7678

- [x] Update the `b` shortcut handler in `registerConsoleShortcuts` to close and then restart all browser servers without restarting watch mode
- [x] Add a regression test to ensure pressing `b` in watch mode:
  - Closes all active browser server instances via `project.browser.close()`
  - Works correctly with multiple browser servers registered

Based on the [previous discussion](https://github.com/vitest-dev/vitest/issues/7678), three approaches were proposed:

1. Restart only the dead (unresponsive) servers  
2. Close and restart all servers  
3. Add a new shortcut key  

Currently, the `b` shortcut is implemented to restart all servers.
If teams think we should instead restart only the dead servers or add a new shortcut, please share feedback. :)

---

### Tests

### Documentation

### Changesets


test/cli/test/console-shortcut.test.ts
```diff
@@ -0,0 +1,30 @@
+import { PassThrough } from 'node:stream'
+import { noop } from '@vitest/utils'
+import { expect, test, vi } from 'vitest'
+import { registerConsoleShortcuts } from 'vitest/node'
+
+test('pressing "b" restarts the browser server', async () => {
```
sheremet-va: this test doesn't test anything, you mocked all APIs

packages/vitest/src/node/stdin.ts
```diff
@@ -119,6 +119,11 @@ export function registerConsoleShortcuts(
       return inputFilePattern()
     }
     if (name === 'b') {
+      await Promise.all(ctx.projects.map(async (project) => {
+        if (project.browser) {
```
sheremet-va, line 123: This will just close the browser, it won't be recreated since init checks for existence of browser, not on its state

pengooseDev:
```ts
// FIXME: should we close the orchestrators and testers?
// project.browser.state.orchestrators.forEach(orchestrator => orchestrator.$close())
// project.browser.state.orchestrators.clear()
// project.browser.state.testers.forEach(tester => tester.$close())
// project.browser.state.testers.clear()
```

 Should we also clear the `orchestrators and testers`? Right now, clearing the orchestrators poses no issues, but because we don't have a type declaration for testers, adding support for clearing the would require a large refactor.
I look forward to team's feedback.

test/watch/test/shortcut.test.ts
```diff
@@ -0,0 +1,56 @@
...
+
+const _options = { root: 'fixtures', watch: true }
+
+it('should restart browser server', async () => {
+  const { vitest, ctx } = await runVitest(_options, ['math'])
+  const ctx2 = ctx as any
+
+  ctx2._initBrowserServers = async () => {
+    ctx2.projects.forEach((project: any) => {
+      project.browser = {
```
sheremet-va: This test overrides everything that needs to be tested. Why can't you just _run_ tests?

sheremet-va (changes requested):

test/watch/test/shortcut.test.ts
pengooseDev: I’m struggling to test the browser server in watch mode in our regression suite:
```ts
  printBrowserBanner(project: TestProject): void {
    if (!project.browser) {
      return
    }
 ```
 Whenever I try to verify messages that depend on the browser server, nothing shows up. I’ve looked through existing regression tests and I can’t find any that assert on context logs emitted from inside the browser. I even tried working around this with fixture settings, but I still can’t capture any log messages generated by the browser. How can I fix this? I need some guidance.
 
---

### Another Available direction to verify server restarted
 ```ts
it('should restart multiple browser servers', async () => {
  const root = resolve(import.meta.dirname, '../fixtures/browser-multiple')
  const { vitest, ctx } = await runVitest({ root, watch: true })
  const ctx2 = ctx as any
  const oldBrowsers = ctx2.projects.map((p: any) => p.browser)

  vitest.write('b')

  await waitFor(() => {
    return ctx2.projects.every((p: any, i: number) => p.browser !== oldBrowsers[i])
  })

  ctx2.projects.forEach((project: any) => {
    expect(project.browser).toBeDefined()
  })
})

 function waitFor(predicate: () => boolean, timeout = 5000) {
  return new Promise<void>((resolve, reject) => {
    const start = Date.now()
    const iv = setInterval(() => {
      if (predicate()) {
        clearInterval(iv)
        resolve()
      }
      else if (Date.now() - start > timeout) {
        clearInterval(iv)
        reject(new Error('timeout waiting for browser restart'))
      }
    }, 50)
  })
}
```

pengooseDev:
I don’t like tying up maintainers’ resources with too many review cycles.
Therefore, if you find the code changes or regression tests insufficiently reliable, please feel free to close the PR, and I’ll submit one for another issue.
Thank you, as always, for your excellent reviews.

pengooseDev:
I don’t like tying up maintainers’ resources with too many review cycles.
Therefore, if you find the code changes or regression tests insufficiently reliable, please feel free to close the PR, and I’ll submit one for another issue.
Thank you, as always, for your excellent reviews. :)

packages/vitest/src/node/stdin.ts
```diff
@@ -119,6 +119,18 @@ export function registerConsoleShortcuts(
...
+          ctx._browserSessions.sessionIds.clear()
+          project.browser.state.orchestrators.forEach((orchestrator) => {
```
sheremet-va: I don't think all of this cleanup is required? The WS cache is removed when the connection is severed

`cleanupTesters` is also async, it can call orchestrator again

Even if some state cleanup is required, it shouldn't be done here, it should be part of `browser.close`

packages/vitest/src/node/stdin.ts
```diff
@@ -119,6 +119,18 @@ export function registerConsoleShortcuts(
...
+            orchestrator.cleanupTesters()
+            orchestrator.$close()
+          })
+          project.browser.state.orchestrators.clear()
+          await project.browser.close()
```
sheremet-va, line 125: I don't see anywhere the closing of actual browsers? `project.browser.close` just closes Vite server

test/cli/test/shortcut.test.ts
```diff
@@ -0,0 +1,49 @@
...
+      enabled: true,
+      provider: 'playwright',
+    }, reporters: [{
+      onBrowserInit(project) {
+        projects.push(project)
+      },
+    }] })
+    const RESTART_COUNT = 5
+
+    const initSpies = projects.map(project => vi.spyOn(project, '_initBrowserServer'))
```
sheremet-va: We really should not care if the internal function was called

test/cli/test/shortcut.test.ts
```diff
@@ -0,0 +1,49 @@
...
+      enabled: true,
+      provider: 'playwright',
+    } })
+    const originalBrowsers = ctx?.projects.map(p => p.browser)
+    await vitest.waitForStdout('Waiting for file changes')
+
+    vitest.write('b')
+    await new Promise(resolve => setTimeout(resolve, 10))
+
+    expect(ctx?.projects.map(p => p.browser)).not.toBe(originalBrowsers)
```
sheremet-va: toBe checks array reference, and you always create a new array with `.map`, so this test will always pass

test/cli/test/shortcut.test.ts
```diff
@@ -0,0 +1,49 @@
...
+describe('restart browser servers with watch', () => {
+  it('should restart multiple browser servers', async () => {
+    const root = resolve(import.meta.dirname, '../fixtures/browser-multiple')
+    const projects: TestProject[] = []
+    const { vitest, ctx } = await runVitest({ root, watch: true, browser: {
+      enabled: true,
+      provider: 'playwright',
+    }, reporters: [{
+      onBrowserInit(project) {
+        projects.push(project)
```
sheremet-va: it's a good idea to check this, since this hooks _should_ be called several times - this is what you need to test

PR #8188, opened 2025-06-19
fix(types): ensure Chai declaration merge works with TS-Go

LukeAbby:
### Description

Context: Currently TypeScript accepts code like this:
```ts
interface Foo {
    foo(x: any): void;
    foo: (x: any) => void;
}
```

While it does not accept code like this:
```ts
interface Foo {
    foo: (x: any) => void;
    foo(x: any): void;
}
```

The only difference being the ordering.

By comparison, in TypeScript Go neither of these are accepted. See https://github.com/microsoft/typescript-go/issues/1192#issuecomment-2976459707 for official confirmation by Anders Hejlsberg (lead architect of TypeScript). I noticed this effected Vitest due to its merge of `Chai.Assert.containSubset` as a callback instead of as a method, leading to a conflict with `@types/chai`.

No associated issue because it's a small types-only fix.

### Tests

### Changesets

### DIFF
--- a/packages/vitest/src/types/global.ts
+++ b/packages/vitest/src/types/global.ts
@@ -7,11 +7,17 @@ import type { UserConsoleLog } from './general'
 declare global {
   // eslint-disable-next-line ts/no-namespace
   namespace Chai {
-    interface Assertion {
-      containSubset: (expected: any) => Assertion
+    interface ContainSubset {
+      (expected: any): Assertion
     }
+
+    interface Assertion {
+      containSubset: ContainSubset
+    }
+
     interface Assert {
-      containSubset: (val: any, exp: any, msg?: string) => void
+      // eslint-disable-next-line ts/method-signature-style
+      containSubset(val: any, exp: any, msg?: string): void
     }
   }
 }


packages/vitest/src/types/global.ts
```diff
@@ -7,11 +7,17 @@ import type { UserConsoleLog } from './general'
...
   // eslint-disable-next-line ts/no-namespace
   namespace Chai {
+    interface ContainSubset {
+      (expected: any): Assertion
+    }
+
     interface Assertion {
-      containSubset: (expected: any) => Assertion
+      containSubset: ContainSubset
     }
```
LukeAbby, line 16: This change is not critically necessary for supporting TS-Go. However while touching this area of the code I noticed that `@types/chai` does this:
```ts
        interface ContainSubset {
            (expected: any): Assertion;
        }

        interface Assertion extends LanguageChains, NumericComparison, TypeComparison {
            ...
            containSubset: ContainSubset;
            ...
        }
```

And this change makes the polyfill here more compliant with what `@types/chai` is doing. Currently if someone changed `ContainSubset` it would break `vitest`.

packages/vitest/src/types/global.ts
```diff
@@ -7,11 +7,17 @@ import type { UserConsoleLog } from './general'
...
+
     interface Assert {
-      containSubset: (val: any, exp: any, msg?: string) => void
+      // eslint-disable-next-line ts/method-signature-style
+      containSubset(val: any, exp: any, msg?: string): void
```
LukeAbby, line 20: This is the change that's critical for TS-Go. You can test this locally by running these commands in `packages/vitest`:
```ts
pnpm add -D @typescript/native-preview
pnpm exec tsgo
```

Before this change you would get these errors:
```
src/types/global.ts:20:7 - error TS2300: Duplicate identifier 'containSubset'.

20       containSubset: (val: any, exp: any, msg?: string) => void
         ~~~~~~~~~~~~~

  ../../node_modules/.pnpm/@types+chai@5.2.2/node_modules/@types/chai/index.d.ts:538:13 - 'containSubset' was also declared here.
    538             containSubset(val: any, exp: any, msg?: string): void;
```

After, these errors go away.

LukeAbby:
Could I get a rerun on "Browsers: node-20, macos-latest" and "Build&Test: node-20, windows-latest"?
The macos error is:
```
Error: Command failed: "/Applications/Firefox.app/Contents/MacOS/Firefox" --version
```

The windows error is:
```
@vitest/test-coverage : test test\coverage-test
  test/coverage-test test$ vitest --workspace=vitest.workspace.custom.ts
   ELIFECYCLE  Command failed with exit code 134.
```

I don't see how either of these could be caused by this change and I see these checks failing elsewhere.

Issue #5722, opened 2024-05-14
Text reporter still creates an empty coverage folder


xixixao:
### Describe the bug

Run
```sh
npx vitest run --coverage --coverage.reporter=text
```
You will get an empty `coverage` folder

### Reproduction

1. Use the default playground: https://stackblitz.com/fork/github/vitest-dev/vitest/tree/main/examples/basic?initialPath=__vitest__/
2. Quit the watcher
3. Run the command
4. Observer the folder

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 18.18.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.2.3 - /usr/local/bin/npm
    pnpm: 8.15.6 - /usr/local/bin/pnpm
  npmPackages:
    @vitest/coverage-v8: ^1.6.0 => 1.6.0 
    @vitest/ui: latest => 1.6.0 
    vite: latest => 5.2.11 
    vitest: latest => 1.6.0 

(it reproes on my mac too)
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

AriPerkkio:
During test run Vitest needs to create `coverage/.tmp` where raw coverage results from tests are stored. After the coverage report has been generated it removes the `.tmp` directory.

I think we could check if `coverage` directory is completely empty after that, and then delete that one as well.

xixixao:
[...]

PR #8212, opened 2025-06-23
fix(coverage): do not throw if reports directory is a mount

orgads:
### Description

When the reports directory is mounted, it is possible to delete all the files and dirs within it, but not the directory itself.

On this case, the purpose of clean is already achieved, and there is no reason to fail.

### Tests

### Documentation

### Changesets


orgads:
The failure is unclear. Is is possible to rerun?

orgads:
Rebased. ping.

test/config/test/failures.test.ts
```diff
@@ -604,3 +605,49 @@ test('cannot set the `workspace` options', async () => {
   })
   expect(stderr).toContain('The `test.workspace` option was removed in Vitest 4. Please, migrate to `test.projects` instead. See https://vitest.dev/guide/projects for examples.')
 })
+
+test('coverage.clean does not throw if reportsDirectory cannot be deleted but is empty', async () => {
```
AriPerkkio, line 609: This test passes in `main` too, without any other changes from this PR.

packages/vitest/src/node/coverage.ts
```diff
@@ -326,7 +340,11 @@ export class BaseCoverageProvider<Options extends ResolvedCoverageOptions<'istan
 
     // Remove empty reports directory, e.g. when only text-reporter is used
     if (readdirSync(this.options.reportsDirectory).length === 0) {
-      await fs.rm(this.options.reportsDirectory, { recursive: true })
+      try {
+        await fs.rm(this.options.reportsDirectory, { recursive: true })
+      }
+      catch {
+      }
```
AriPerkkio, line 347: We want to throw the errors from failed `fs.rm` here.

packages/vitest/src/node/coverage.ts
```diff
@@ -232,11 +232,25 @@ export class BaseCoverageProvider<Options extends ResolvedCoverageOptions<'istan
...
-      })
+      try {
+        await fs.rm(this.options.reportsDirectory, {
+          recursive: true,
+          force: true,
+          maxRetries: 10,
+        })
+      }
+      catch (error) {
+        // If the directory is empty or cannot be deleted (e.g., mounted), allow it
```
AriPerkkio, line 243: Could this case be identified from `error.code` instead?

test/config/test/failures.test.ts
orgads: Yes, it's quite strange. If I run it with:
```
npx vitest -t coverage.clean
```

then the mock is called and it fails (without the fix), but if I run it through pnpm the mock is never called :/

Can you assist?

packages/vitest/src/node/coverage.ts
orgads: The error code is `EBUSY`, but it is not necessarily emitted on the directory itself, it can be from one of the files/directories inside it.

Do you want to change the condition to:
```ts
if (error.code !== 'EBUSY' || readdirSync(this.options.reportsDirectory).length > 0) {
```
?

packages/vitest/src/node/coverage.ts
orgads: Why do you care if you cannot delete an empty directory? Should it fail the whole operation?

AriPerkkio: Users do care: https://github.com/vitest-dev/vitest/issues/5722

We should be able to catch regression if we break this functionality.

AriPerkkio (commented):
I'm not sure I fully understand why this fix is needed. Users could just define `coverage: { reportsDirectory: './mounted-directory/coverage' }`, so that `mounted-directory` would not cause the errors. This same issue should happen with other test result directories too, not just coverage.

test/config/test/failures.test.ts
AriPerkkio: I think the mocking happens in different execution context here, as these tests are running Vitest inside Vitest.

orgads:
I think I can make that work. My use case involves running coverage inside a container and mounting the `coverage` directory so the results are accessible on the host. Since we run coverage both inside the container and directly on the host, I'll need to write a script that detects when it's running in the container and writes results to something like `coverage/foo` (or `/tmp/coverage`), then moves them to `coverage`. It's doable, but definitely a bit messy.

PR #8256, opened 2025-07-03
feat(ui): [PoC] bundle reporter ui's HTML and it's resources into single HTML file

### LINKED ISSUE #6425: Option to save report in a single combined HTML file

In our CI pipeline, we generate HTML reports using @vitest/ui. Our pipeline is designed to handle self-contained files, meaning that all assets (scripts, styles, etc.) must be embedded directly within the HTML report. Currently, the generated reports include external asset references, which our pipeline cannot process correctly due to its requirement for single-file artifacts.

### Suggested solution

We would like to request an option in @vitest/ui to generate fully self-contained HTML reports. This feature would inline all necessary assets into the HTML file, producing a single self-contained file that is more easily portable.

### Alternative

- Trying to make the multi-file structure work in our CI - with the config available in our CI, I've attempted to upload all the report files (i.e. all of `html/`), but the the original directory structure isn't preserved; all assets end up at `$s3Host.$file`. Potentially, an option to fully customize the `src` attribute to the needed assets could also solve this, but I feel that a self-contained report is ultimately the better option. 

### Additional context

Here's a similar issue for a `jest` plugin: https://github.com/Hazyzh/jest-html-reporters/issues/184

### Validations

Shinyaigeek:
### Description

Close: https://github.com/vitest-dev/vitest/issues/6425

**In current(25/07/04) status, this is just PoC. so I want feedback about interface, and design from maintainers 🙇**

Still this is just PoC, you can test this with the below vitest option

```ts
// vitest.config.ts
export default defineConfig({
  test: {
    reporters: [
        [
            'html',
            {
                bundleSingleHtml: true,
            }
        ]
    ],
  },
})
```

As mentioned above, this is still PoC but I would like to refine and complete this implementation if maintainers feels this is acceptable.

---

I implemented `bundleSingleHtml` option for html reporter. This enable users to output reporter's html as single HTML file instead of HTML file with it's subresources like css or js.

In my understanding, HTML reporter's frontend is implemented in `@vitest/ui`. This is implemented as SPA and built as HTML + CSS + JS + favicon in the build process of vitest itself. When a user invoke `vitest` with `--ui` or `--provider=html` option, vitest will embed user's test data as `METADATA_PATH`, which is path for test data's gzip and in frontend, vitest's ui application resolve it's `METADATA_PATH` test data and display that.

In this PoC, I implemented like below in order to accomplish to serve reporter's ui as single HTML file

- bundle `@vitest/ui` SPA application as a single HTML file with [vite-plugin-singlefile](https://www.npmjs.com/package/vite-plugin-singlefile) as separated build process of the normal `@vitest/ui` SPA application's build process(outputs HTML + CSS + JS)
- if `bundleSingleHTML` option is specified, vitest will embed user's test data as base64 encoded data instead of gzip path
- Additionaly, favicon is abandoned in single bundled HTML because it cannot be bundle into HTML

Please feel free to point out if you feel something unclear or this way is not accepatable!

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/ui/client/composables/client/static.ts
+++ b/packages/ui/client/composables/client/static.ts
@@ -95,18 +95,30 @@ export function createStaticClient(): VitestClient {
   }
 
   async function registerMetadata() {
-    const res = await fetch(window.METADATA_PATH!)
-    const contentType = res.headers.get('content-type')?.toLowerCase() || ''
-    if (
-      contentType.includes('application/gzip')
-      || contentType.includes('application/x-gzip')
-    ) {
-      const compressed = new Uint8Array(await res.arrayBuffer())
-      const decompressed = strFromU8(decompressSync(compressed))
+    if (window.METADATA_BASE64) {
+      // Handle embedded metadata for single HTML mode
+      const compressedData = Uint8Array.from(atob(window.METADATA_BASE64), c => c.charCodeAt(0))
+      const decompressed = strFromU8(decompressSync(compressedData))
       metadata = parse(decompressed) as HTMLReportMetadata
     }
+    else if (window.METADATA_PATH) {
+      // Handle external metadata file for multi-file mode
+      const res = await fetch(window.METADATA_PATH)
+      const contentType = res.headers.get('content-type')?.toLowerCase() || ''
+      if (
+        contentType.includes('application/gzip')
+        || contentType.includes('application/x-gzip')
+      ) {
+        const compressed = new Uint8Array(await res.arrayBuffer())
+        const decompressed = strFromU8(decompressSync(compressed))
+        metadata = parse(decompressed) as HTMLReportMetadata
+      }
+      else {
+        metadata = parse(await res.text()) as HTMLReportMetadata
+      }
+    }
     else {
-      metadata = parse(await res.text()) as HTMLReportMetadata
+      throw new Error('No metadata found. Expected either METADATA_PATH or METADATA_BASE64.')
     }
     const event = new Event('open')
     ctx.ws.dispatchEvent(event)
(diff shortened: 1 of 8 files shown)


sheremet-va:
It would be nice if we didn't need to bundle twice, doubling the size of the package for a single option seems strange.

Shinyaigeek:
@sheremet-va Thank you for feedback!(and sorry for really late 🙇 I was too busy to develop any OSS)

> It would be nice if we didn't need to bundle twice, doubling the size of the package for a single option seems strange.

I agree. On the other hand, I couldn't think of any way to provide the vitest-ui frontend as a single HTML file other than building it twice. vitest-ui generates static HTML, CSS, and JS resources for the frontend by building them before the package is published, rather than at package startup (which I think is a legitimate approach for improving startup speed). Therefore, to switch between distributing a single HTML file or multiple HTML, JS, and CSS files via an option, it seems necessary to build both before publishing the package...

sheremet-va:
> @sheremet-va Thank you for feedback!(and sorry for really late 🙇 I was too busy to develop any OSS)
> 
> > It would be nice if we didn't need to bundle twice, doubling the size of the package for a single option seems strange.
> 
> I agree. On the other hand, I couldn't think of any way to provide the vitest-ui frontend as a single HTML file other than building it twice. vitest-ui generates static HTML, CSS, and JS resources for the frontend by building them before the package is published, rather than at package startup (which I think is a legitimate approach for improving startup speed). Therefore, to switch between distributing a single HTML file or multiple HTML, JS, and CSS files via an option, it seems necessary to build both before publishing the package...

I think if we can provide a single file output, then we should probably just only do that.

Issue #7914, opened 2025-04-30
`browser.instances[].includes` is ignored


AriPerkkio:
### Describe the bug

This configuration ends up running all tests with Firefox:

```ts
import { defineConfig } from 'vitest/config';

export default defineConfig({
  test: {
    browser: {
      enabled: true,
      provider: 'playwright',

      instances: [
        // Run all tests with Chromium, it's stable
        { browser: "chromium" },

        // Firefox is flaky, run just one test 
        { browser: "firefox", include: ["test/something-simple.test.ts"] },
      ],
    },
  },
});
```

### Reproduction

https://stackblitz.com/~/edit/vitest-browser-intances-includes-ignroed

Happens on latest too, but due to https://github.com/vitest-dev/vitest/issues/7913 reproducing the issue with `3.1.1` on Stackblitz. 

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    CPU: (4) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 20.19.1 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.8.2 - /usr/local/bin/npm
    pnpm: 8.15.6 - /usr/local/bin/pnpm
  npmPackages:
    @vitest/browser: latest => 3.1.2 
    @vitest/ui: latest => 3.1.2 
    vite: latest => 6.3.4 
    vitest: latest => 3.1.2
```

### Used Package Manager

pnpm

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vitest.dev/guide/).
- [x] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

[...]

PR #8260, opened 2025-07-04
fix(vitest): override `config.include` option with `config.browser.instances[].include` option if it is specified

Shinyaigeek:
### Description

Close: #7914 

Hi Team 👋 

As described in https://github.com/vitest-dev/vitest/issues/7914#issuecomment-2847037046, I understand that #7914 happens because `mergeConfig` merges array in config object instead of overriding it.

As mentioned in [docs](https://vitest.dev/config/#include), vitest config root's `include` option has `['**/*.{test,spec}.?(c|m)[jt]s?(x)']` as default value, and config which is calculated per browser instance configuration will be like below:

```typescript
// input vitest.config.ts

export default defineConfig({
  test: {
    browser: {
      enabled: true,
      provider: 'playwright',

      instances: [
        // Run all tests with Chromium, it's stable
        { browser: "chromium" },

        // Firefox is flaky, run just one test 
        { browser: "firefox", include: ["test/something-simple.test.ts"] },
      ],
    },
    // include: ['**/*.{test,spec}.?(c|m)[jt]s?(x)'] as default value
  },
});

// output config for chromium
export default defineConfig({
  test: {
    ...config,
    include: ['**/*.{test,spec}.?(c|m)[jt]s?(x)']
});

// output config for firefox
export default defineConfig({
  test: {
    ...config,
    include: ['**/*.{test,spec}.?(c|m)[jt]s?(x)', "test/something-simple.test.ts"]
});
```

and so, `'**/*.{test,spec}.?(c|m)[jt]s?(x)'` pattern in firefox's config will match test files which should be ignored by `config.browsers.instances[firefox].include`.

I thought vitest user expect include config in deeper will be prioritized than the root that, so I fixed this issue by ignoreing project root's include value if `config.browser.instances[].include` is specified.
 
### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/projects/resolveProjects.ts
+++ b/packages/vitest/src/node/projects/resolveProjects.ts
@@ -322,8 +322,10 @@ function cloneConfig(project: TestProject, { browser, ...config }: BrowserInstan
       providerOptions: config,
       instances: undefined, // projects cannot spawn more configs
     },
-    // If there is no include pattern in browser.instances[], we should use the include pattern from the parent project
+    // If there is no include or exclude or includeSource pattern in browser.instances[], we should use the that's pattern from the parent project
     include: (overrideConfig.include && overrideConfig.include.length > 0) ? [] : clonedConfig.include,
+    exclude: (overrideConfig.exclude && overrideConfig.exclude.length > 0) ? [] : clonedConfig.exclude,
+    includeSource: (overrideConfig.includeSource && overrideConfig.includeSource.length > 0) ? [] : clonedConfig.includeSource,
     // TODO: should resolve, not merge/override
   } satisfies ResolvedConfig, overrideConfig) as ResolvedConfig
 }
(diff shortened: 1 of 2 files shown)


packages/vitest/src/node/projects/resolveProjects.ts
```diff
@@ -321,6 +322,8 @@ function cloneConfig(project: TestProject, { browser, ...config }: BrowserInstan
       providerOptions: config,
       instances: undefined, // projects cannot spawn more configs
     },
+    // If there is no include pattern in browser.instances[], we should use the include pattern from the parent project
+    include: (overrideConfig.include && overrideConfig.include.length > 0) ? [] : clonedConfig.include,
```
sheremet-va, line 326: Let's also have `exclude` and `includeSource` here, they usually all go together as a single source of test files

sheremet-va (changes requested):
LGTM! Let's also apply this to `exclude` and `includeSource`

packages/vitest/src/node/projects/resolveProjects.ts
Shinyaigeek: Make sense! Thank you!  d59b896

sheremet-va (approved):

Issue #8175, opened 2025-06-16
Support standalone + a filename pattern in the CLI


mjewell:
### Clear and concise description of the problem

I want to use `--standalone` as the default behaviour in our test script because it normally doesn't make sense for people to run the entire suite, e.g.

```sh
"test": "vitest --standalone --coverage"
```

But I still want people to be able to pass a path if they want to run a subset of tests. e.g.

```sh
yarn test src/whatever

# file path is ignored, no tests run
```

It would be great if we could default to standalone but still allow people to pass a path. I know you can also press `p` and type a path, but

- people aren't as familiar with that
- when you're typing a path there it doesn't let you tab to complete the file path

Is there some existing way to do this?

### Suggested solution

When given a file path in the CLI vitest could make that the default filename pattern. It would behave the same as when you do

- `vitest --standalone`
- press `p`
- type the path and press enter

### Alternative

We could have two scripts in our app, one that has standalone enabled and one that doesn't, but we never really want the version that runs all the tests for local development, and when we do we could just run standalone and press `a`.

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vitest.dev/guide/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

AriPerkkio:
The point of `--standalone` is to not start any tests. I'm not sure if it would be unexpected if it suddenly started running tests when filename pattern is also given.

Would it work better if filename pattern was set automatically when it was passed as CLI argument? Then you could just press enter and see the test run.

mjewell:
I think that might be a bit better but still feels a bit awkward in my opinion. Agree that it feels a bit confusing with the way things are set up but I do think the workflow I suggested above seems like a pretty good default when a project gets to be large enough that running all the tests is slow so it'd be nice if there was a way to support it

AriPerkkio:
This was accepted by the team as breaking change.

So when user passes a filename filter with `--standalone`, we discard the standalone mode and just start the passed files.

PR #8262, opened 2025-07-05
fix!: ignore `--standalone` when CLI filename filter is used

AriPerkkio:
### Description

- Fixes https://github.com/vitest-dev/vitest/issues/8175

This is a breaking change to improve overall user experience. It makes sense to start the test run when `--standalone` is used with CLI filename filter.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/cli/cli-api.ts
+++ b/packages/vitest/src/node/cli/cli-api.ts
@@ -57,6 +57,7 @@ export async function startVitest(
     options,
     viteOverrides,
     vitestOptions,
+    cliFilters,
   )
 
   if (mode === 'test' && ctx.config.coverage.enabled) {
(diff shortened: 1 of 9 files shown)


packages/vitest/src/node/cli/cli-api.ts
```diff
@@ -148,6 +150,10 @@ export async function prepareVitest(
     options.watch = false
   }
 
+  if (options.standalone && (cliFilters?.length || 0) > 0) {
+    options.standalone = false
+  }
```
AriPerkkio, line 155: This seemed like the best place to do this. The `resolveConfig` doesn't have `cliFilters` in the scope.

sheremet-va (dismissed):
Can we also add a note to the migration guide?

sheremet-va (approved):

PR #8266, opened 2025-07-06
feat(types): add better promise support in expects and matchers

### LINKED ISSUE #6279: Missing type argument for Assertions<T> after Promisify (resolves/rejects)
### Describe the bug

When you chain an assertion using the `.resolves`/`.rejects` properties, the chained assertions lose any existing type arguments. 

### Reproduction

```ts
await expect(fn()).resolves.toEqual<number>(1)
//                                 ^^^^^^^^
// Says given 1 type argument, expected 0.
```

All the while this is working as expected:

```ts
expect(fn()).toEqual<number>(1)
```

The issue is likely caused by the `Promisify<T>` type dropping any type arguments from the `Assertions<T>` it accepts in `resolves` and `rejects` properties: 

https://github.com/vitest-dev/vitest/blob/74ca11a4b88289828cd3bfd54cb1271c181a29ba/packages/expect/src/types.ts#L174-L182

### System Info

```shell
I believe system info is irrelevant here, it's a types issue.
```

### Used Package Manager

pnpm

### Validations

samchungy:
### Description

Resolves https://github.com/vitest-dev/vitest/issues/6279

At the moment, when you use `.resolves` or `.rejects` the type generics get lost due to the Promisify function.

```ts
import { expect } from "vitest";

interface FooBar {
  foo: string;
}

expect({ foo: "bar" }).resolves.toStrictEqual<FooBar>({ foo: "bar" });
expect({ foo: "bar" }).toStrictEqual<FooBar>({ foo: "bar" });
```

<details>
  <summary>See error</summary>
	image
</details>

This PR changes the types and allows us to also type-check the resolves and reject matchers as well as properly types custom matchers as promises.

### Tests

### Documentation

### Changesets


test/typescript/test-d/expect-extend.test-d.ts
```diff
@@ -27,6 +27,7 @@ test('infers matcher declaration type from a custom matcher type', () => {
   })
 
   expect({ a: 1, b: '2' }).toMatchSchema({ a: '1' })
+  await (expect({ a: '1' }).resolves.toMatchSchema({ a: '1' }) satisfies Promise<{ a: string }>)
```
samchungy: This previously wasn't returning a Promise.

packages/browser/matchers.d.ts
```diff
@@ -3,18 +3,11 @@ import type { TestingLibraryMatchers } from './jest-dom.js'
 import type { Assertion, ExpectPollOptions } from 'vitest'
 
 declare module 'vitest' {
-  interface JestAssertion<T = any> extends TestingLibraryMatchers<void, T> {}
```
samchungy, line 6: Technically breaking but this aligns more with how Jest works and makes more sense?

See: https://github.com/vitest-dev/vitest/pull/8266/files#diff-810c1bb9ffac48cfe14f7dbbaf10fdf5725e06feecb99c657d49f3c7712bcd00L197-R197

image

I think we may be incorrectly setting these as it seems like the intention is for `void` to be R and T to be set to E here.

samchungy:
Tests seem flakey and out of control of my PR

sheremet-va (changes requested):
`expect(val).resolves.eq(val)` no longer returns a Promise

samchungy:
Added a test and fixed that! Thanks for catching! There's so many cases 🥲

sheremet-va:
We decided that with the amount of breaking changes in Vitest 4 already, we need to postpone this at least to Vitest 5 because the impact is quite big, but it fixes a niche use case

samchungy:
Fair enough! feel free to tag me when it's time to revisit

sheremet-va:
We are considering to merge this for 5.0. Could you resolve the conflicts, please?

sheremet-va (approved):

PR #8273, opened 2025-07-08
Fix: nested test projects

fightZy:
### Description

Fix the repeated execution of test cases when Test Projects have multi-level nesting, and eliminate unexpected and undesired duplicate content;

For example, regarding the added test cases, I expect the content under business/ to execute the test cases using its own directory configuration, just like the directories under packages/. Therefore, I added packages/business/*. 
image
image
However, this currently results in duplicate execution directories: packages/business and packages/business/pkg, and the execution of packages/business is not as expected.

- [x] It's really useful if your PR references an issue where it s discussed ahead of time. If the feature is substantial or introduces breaking changes without a discussion, PR might be closed.

### Tests

### Documentation

### Changesets


sheremet-va:
I don't like this. The `projects` glob is intentionally dead simple. `packages/*` matches `packages/business`, so it _should_ create a project there. If you don't want it to be treated as a project, add a negated glob `!packages/business`

fightZy:
Thank you for your feedback! I understand your consideration of keeping the project glob simple.

## Regarding the Complexity of Manual Exclusion

You mentioned using the negative glob `!packages/business`, but in reality, this won't work in the current situation because `'packages/business/*'` will also be excluded. To correctly exclude nested projects, users need to use:
```
['packages/!(business)', 'packages/business/*']
```

This syntax is not intuitive for many developers, especially in large monorepos where multiple such exclusion rules may need to be maintained, which imposes a certain mental burden.

## The Significance of the Current Solution

This modification mainly addresses the issue where **developers' expectations do not match the actual behavior**:

1. **Intuitiveness**: When users configure `packages/business/*`, their expectation is to "only run sub-projects under business" rather than "run business itself + sub-projects".
2. **Avoid Duplicate Execution**: In our actual projects, such duplicate execution has led to doubled testing time and confusing reports.
3. **Consistency with Common Usage Patterns**: In monorepos, there are usually root-level configurations and sub-project-specific configurations.

## Improvement Suggestions

I fully understand your concern about maintaining simplicity. Could we consider making this feature an **optional one**? For example:

```
export default {
  projects: {
    patterns: ['packages/*', 'packages/business/*'],
    deduplication: true // Defaults to false for backward compatibility
  }
}
```

This way, the simplicity of the existing behavior is preserved, while providing an option for users who need such intelligent handling.

What do you think of this optional approach? Looking forward to your feedback.

JesusTheHun:
@sheremet-va What about inline configurations ?

Let's say the root config is `projects: ['packages/*']` and each package configure several inline projects to run unit tests or e2e tests, for example.

In CI, you may want to run e2e tests using the docker image provided playwright along with `vitest --run --project='*:e2e'`. And run the non e2e tests on the node image.

Currently, projects do not run nested projects, inline or not.

sheremet-va:
> This syntax is not intuitive for many developers, especially in large monorepos where multiple such exclusion rules may need to be maintained, which imposes a certain mental burden.

Then perhaps adding the example to the documentation is enough. 

Magically removing files from the glob you specified is more confusing than the current behaviour where we process everything that your glob matches.

> @sheremet-va What about inline configurations ?

This PR has nothing to do with inline configuration.

fightZy:
That’s true, you’ve got a point.

Issue #8244, opened 2025-07-01
`No test files found` error with sharing and just a few shards


Fox32:
### Describe the bug

We are using sharding a lot, it works well in general. But lately we had some issues with edge cases: We have a package with just a few test suites, just enough to have a single test on each shard, however the last shard fails with `No test files found`. Maybe there is an off-by-one error somewhere?

### Reproduction

Use this example with **four** test suites: https://stackblitz.com/edit/vitest-dev-vitest-wmhw4hys?file=test%2Fa.test.ts,test%2Fb.test.ts&initialPath=__vitest__/

What works:
* Running all tests via `vitest --no-watch` works
* Running with two shards works: `vitest --no-watch --shard 1/2 && vitest --no-watch --shard 1/2`
* Running with five shards does not work, as expected: `vitest --no-watch --shard 1/5 && vitest --no-watch --shard 2/5 && vitest --no-watch --shard 3/5 && vitest --no-watch --shard 4/5 && vitest --no-watch --shard 5/5` results in `Error: --shard <count> must be a smaller than count of test files. Resolved 4 test files for --shard=1/5.`
* Running with four shards works as expected too: `vitest --no-watch --shard 1/4 && vitest --no-watch --shard 2/4 && vitest --no-watch --shard 3/4 && vitest --no-watch --shard 4/4`
* **However, running with only three shards fails:**: `vitest --no-watch --shard 1/3 && vitest --no-watch --shard 2/3 && vitest --no-watch --shard 3/3`, the third shard fails with `No test files found, exiting with code 1`. Two tests are executed on the first shard, two on the second and the last shard has not tests left.

I think this problem wasn't noticeable before introducing the new error message that you need at least as many tests as shards.

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 20.19.1 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.8.2 - /usr/local/bin/npm
    pnpm: 8.15.6 - /usr/local/bin/pnpm
  npmPackages:
    @vitest/ui: latest => 3.2.4 
    vite: latest => 7.0.0 
    vitest: latest => 3.2.4
```

### Used Package Manager

npm

### Validations

[...]

PR #8288, opened 2025-07-10
fix: distribute test files to shards more evenly

Shinyaigeek:
### Description

Fixes: #8244

In current implementation, test files are distributed with `Math.ceil(files.length / count)` in `{index}/{count}` shard setting.
This causes #8244 because this ceil based distribution logic can consume all test files by shards without the ending shard.
In addition to this, I believe this logic causes a bit lower performance in an edge-case like:

- test files: 10
- `--shard index/3`

In the above case, test files will be distributed as `[3, 3, 3, 1]` but I think it is more ideal to be distributed as `[3, 3, 2, 2]` in the point of parallelism.

This PR fixes this issue by introducing distribution logic based on `Math.floor(files.length / count)` instead of `Math.ceil`

### Tests

### Documentation

### Changesets

### DIFF
--- a/test/config/test/shard.test.ts
+++ b/test/config/test/shard.test.ts
@@ -67,7 +67,7 @@ test('--shard=3/3 should distribute files evenly', async () => {
   expect(paths).toEqual(['3.test.js'])
 })
 
-test('4 files with 3 shards should distribute as [2,1,1] not [2,2,0]', async () => {
+test('4 files with 3 shards should distribute evenly', async () => {
   const { stdout: stdout1 } = await runVitest({ shard: '1/3' }, './fixtures/shard-4-files')
   const { stdout: stdout2 } = await runVitest({ shard: '2/3' }, './fixtures/shard-4-files')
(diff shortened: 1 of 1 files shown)


test/config/test/shard.test.ts
```diff
@@ -40,6 +40,52 @@ test('--shard=2/2', async () => {
...
+test('--shard=3/3 should distribute files evenly', async () => {
+  const { stdout } = await runVitest({ shard: '3/3' })
+
+  const paths = parsePaths(stdout)
+
+  // With 3 files and 3 shards, should get 1 file per shard
+  expect(paths).toEqual(['3.test.js'])
+})
+
+test('4 files with 3 shards should distribute as [2,1,1] not [2,2,0]', async () => {
```
AriPerkkio:
```suggestion
test('4 files with 3 shards should distribute evenly', async () => {
```

AriPerkkio (approved):
Looks good, thanks @Shinyaigeek! 💯

PR #8290, opened 2025-07-11
fix: Remove Nullable type.

cpojer:
### Description

I usually have a `Nullable` type in my projects. When I use vitest in the same project, VS Code/TypeScript's auto-import always tries to pull the one from vitest. This is kinda fine but also somewhat annoying as I don't want the entire type system of an app to depend on types provided by a testing framework.

I think it's best not to export such a generic type that is available to a whole app through Vite. I checked, and there were actually only two minor usages of this type inside of vitest itself. In this PR, I'm removing the type and inlining the usage in the two places.

If you still want to keep it (and export it), I would prefer renaming it to something like `ViteNullable`.

### Tests
Typechecking passes, also ran the tests.


packages/utils/src/index.ts
```diff
@@ -47,7 +47,6 @@ export type {
   Constructable,
   DeepMerge,
   MergeInsertions,
-  Nullable,
```
sheremet-va: This is an internal package, TypeScript doesn't provide these types in user projects

packages/vitest/src/types/general.ts
```diff
@@ -1,7 +1,6 @@
 export type { ParsedStack, TestError } from '@vitest/utils'
 
 export type Awaitable<T> = T | PromiseLike<T>
-export type Nullable<T> = T | null | undefined
```
sheremet-va, line 4: This is the only change that we should keep from this PR (meaning let's remove this line)

sheremet-va (changes requested):
The `Nullable` type from `vitest` was already removed in Vitest 4 beta, your changes don't affect the public API at all

packages/vitest/src/types/general.ts
cpojer: 👍

cpojer:
I just inlined the other uses since it didn't seem like it was widely used in the codebase.

Updated the PR to only remove the unused exported type from the main package.

PR #8291, opened 2025-07-11
feat(coverage-swc): add SWC coverage provider for Vitest

travzhang:

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/coverage-swc/src/provider.ts
+++ b/packages/coverage-swc/src/provider.ts
@@ -1,0 +1,220 @@
+import type { WasmPlugin } from '@swc/core'
+import type { CoverageMap } from 'istanbul-lib-coverage'
+import type { ProxifiedModule } from 'magicast'
+import type { CoverageProvider, ReportContext, ResolvedCoverageOptions, Vitest } from 'vitest/node'
+import { promises as fs } from 'node:fs'
+import { transformSync } from '@swc/core'
+import createDebug from 'debug'
+import libCoverage from 'istanbul-lib-coverage'
+import libReport from 'istanbul-lib-report'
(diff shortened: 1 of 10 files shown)


travzhang:
https://github.com/vitest-dev/vitest/issues/8292

packages/coverage-swc/src/provider.ts
```diff
@@ -0,0 +1,220 @@
...
+
+export class IstanbulCoverageProvider extends BaseCoverageProvider<ResolvedCoverageOptions<'swc'>> implements CoverageProvider {
+  name = 'swc' as const
+  version: string = version
+
+  initialize(ctx: Vitest): void {
+    this._initialize(ctx)
+  }
+
+  onFileTransform(sourceCode: string, id: string, pluginCtx: any): { code: string; map: any } | undefined {
```
AriPerkkio, line 30: Apart from `onFileTransform`, is everything else in this PR 100% copy of `@vitest/coverage-istanbul`?

What if SWC had `istanbul-lib-instrument` compatible instrumenter that downstream projects like Vitest could use. 🤔

travzhang: Yes, apart from the `onFileTransform` implementation, the rest of the PR currently mirrors the logic from `@vitest/coverage-istanbul` to ensure compatibility while keeping the changes minimal.

Regarding instrumentation, since there isn’t a direct drop-in alternative yet, I propose exposing this as an **optional provider** (e.g., `istanbul-swc`) for now. This would allow users to opt in without disrupting existing workflows, and we can revisit the implementation once a more robust solution (like SWC-native instrumentation) becomes available.

Would this approach work for the time being?

Issue #7904, opened 2025-04-29
expect `toBeNullish`


MetRonnie:
### Clear and concise description of the problem

Sometimes we don't care whether a value is `undefined` or `null`, we just want to test that it is [nullish](https://developer.mozilla.org/en-US/docs/Glossary/Nullish).

### Suggested solution

I am proposing a new assertion, `toBeNullish()`. This would test `value == null`, i.e. that the value is either `null` or `undefined`.

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vitest.dev/guide/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #8294, opened 2025-07-11
feat(expect): support `toBeNullable` expect function to check provided value is nullish

Shinyaigeek:
### Description

Close: https://github.com/vitest-dev/vitest/issues/7904

In this PR, I implemented `toBeNullable` expect function to check provided value is nullish. 
I named `toBeNullable` instead of `toBeNullish`, which is proposed in original issue, because there is already similar function `toBeNullable` in `expectTypeOf` ref: https://vitest.dev/api/expect-typeof.html#tobenullable

### Tests

### Documentation

### Changesets


packages/expect/src/types.ts
```diff
@@ -338,6 +338,14 @@ export interface JestAssertion<T = any> extends jest.Matchers<void, T>, CustomMa
    */
   toBeNull: () => void
 
+  /**
+   * Used to check that a variable is nullable(null or undefined).
```
sheremet-va:
```suggestion
   * Used to check that a variable is nullable (null or undefined).
```

sheremet-va (dismissed):

sheremet-va:
Thank you!

Issue #4060, opened 2023-09-01
Display Snapshot of the component in Vitest UI


IsraelDCastro:
### Clear and concise description of the problem

When I create a snapshot of the component in the test, I have to look for the file I try to figure out which is the component I want to see (When I make a few snapshots), because they are all in the same file. The style of the file doesn't let me check very well. Most of when the component is a form, I want to check if the fields are filled correctly.

### Suggested solution

Maybe it could be a way to display the component snapshot in vitest-ui, to check the component snapshot in that place, so the users won't need to search the snapshot file and try to figure out the bad style of the component.

Display this in vitest UI I think will improve the development of the tests and it will be more efficient to check if everything is ok.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #8335, opened 2025-07-22
feat(ui): display test's snapshot result on vitest-ui

Shinyaigeek:
### Description

Close: https://github.com/vitest-dev/vitest/issues/4060

スクリーンショット 2025-07-23 2 08 54

In this ticket, I implemented to display test's snapshot result on vitest-ui. Please feel free to provide any feedback about UserInterface(=vitest/ui itself) and also data flow design.

In this implementation, vitest/ui can display added/matched/unmatched/unchecked snapshots result. Vitest node environment(=vm context) executes tests and generate snapshot result, and passes that result to vitest core context via `rpc()`, and vitest-ui fetches some data from vitest core context. However, vitest node environment passes only snapshot diff count information to vitest core context. in current implementation. So I changed to pass more detailed snapshot result to vitest core context to display these in vitest-ui.

If you have any concern or something unclear, please feel free to pointed out!

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/ui/client/components/views/ViewSnapshot.vue
+++ b/packages/ui/client/components/views/ViewSnapshot.vue
@@ -1,0 +1,287 @@
+<script setup lang="ts">
+import type { RunnerTestFile } from 'vitest'
+import { client } from '~/composables/client'
+
+interface SnapshotData {
+  filepath: string
+  snapshots: {
+    added: Array<{
+      name: string
+      content: string
+      testName?: string
+    }>
+    matched: Array<{
+      name: string
+      content: string
+      testName?: string
+    }>
+    unmatched: Array<{
+      name: string
+      content: string
+      testName?: string
+    }>
+    unchecked: Array<{
+      name: string
+      content: string
+      testName?: string
+    }>
+  }
+}
+
+const props = defineProps<{
+  file: RunnerTestFile
+}>()
+
+const snapshotData = ref<SnapshotData | void>()
+const loading = ref(false)
+const error = ref<string | null>(null)
+
+async function loadSnapshot() {
+  if (!props.file?.filepath) {
+    return
+  }
+  loading.value = true
+  error.value = null
+
+  try {
+    const data = await client.rpc.getSnapshotData?.(props.file.filepath)
+    snapshotData.value = data
+  }
+  catch (err) {
+    error.value = err instanceof Error ? err.message : 'Failed to load snapshot data'
+  }
+  finally {
+    loading.value = false
+  }
+}
+
+// Copy to clipboard function
+async function copyToClipboard(text: string) {
+  try {
+    await navigator.clipboard.writeText(text)
+  }
+  catch {
+    // Fallback for older browsers
+    const textArea = document.createElement('textarea')
+    textArea.value = text
+    document.body.appendChild(textArea)
+    textArea.select()
+    document.execCommand('copy')
+    document.body.removeChild(textArea)
+  }
+}
+
+// Load snapshot when file changes
+watch(() => props.file?.filepath, loadSnapshot, { immediate: true })
+</script>
+
+<template>
+  <div class="h-full flex flex-col">
+    <div v-if="loading" class="flex-1 flex items-center justify-center">
+      <div class="text-secondary">
+        Loading snapshots...
+      </div>
+    </div>
+
+    <div v-else-if="error" class="flex-1 flex items-center justify-center flex-col gap-2">
+      <div class="text-red-500">
+        {{ error }}
+      </div>
+      <button class="text-primary underline" @click="loadSnapshot">
+        Try again
+      </button>
+    </div>
+
+    <div v-else-if="!snapshotData" class="flex-1 flex items-center justify-center">
+      <div class="text-secondary">
+        No snapshots found for this test file
+      </div>
+    </div>
+
+    <div
+      v-else-if="!snapshotData.snapshots.added.length
+        && !snapshotData.snapshots.matched.length
+        && !snapshotData.snapshots.unmatched.length
+        && !snapshotData.snapshots.unchecked.length"
+      class="flex-1 flex items-center justify-center"
+    >
+      <div class="text-secondary">
+        Snapshot file found but no snapshots exist
+      </div>
+    </div>
+
+    <div v-else class="flex-1 overflow-auto p-4 space-y-6">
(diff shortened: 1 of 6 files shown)


packages/snapshot/src/manager.ts
```diff
@@ -17,10 +18,12 @@ export class SnapshotManager {
 
   clear(): void {
     this.summary = emptySummary(this.options)
+    this.result = undefined
   }
 
   add(result: SnapshotResult): void {
     addSnapshotResult(this.summary, result)
+    this.result = result
```
sheremet-va, line 26: Why do we need a separate variable? What does `result` represent?

packages/ui/client/components/views/ViewSnapshot.vue
```diff
@@ -0,0 +1,287 @@
...
+            </div>
+          </div>
+        </details>
+      </div>
+
+      <!-- Unmatched Snapshots -->
+      <div v-if="snapshotData.snapshots.unmatched.length > 0" class="space-y-2">
+        <details class="group" open>
+          <summary class="cursor-pointer p-2 bg-red-50 dark:bg-red-950 border border-red-200 dark:border-red-800 rounded text-red-700 dark:text-red-300 font-medium text-sm">
+            ❌ Unmatched Snapshots ({{ snapshotData.snapshots.unmatched.length }})
```
sheremet-va, line 190: What is "unmatched" snapshot?

packages/ui/client/components/views/ViewSnapshot.vue
```diff
@@ -0,0 +1,287 @@
...
+    <div v-else class="flex-1 overflow-auto p-4 space-y-6">
+      <div class="text-xs text-secondary opacity-70 font-mono mb-4">
+        {{ snapshotData.filepath }}
+      </div>
+
+      <!-- Added Snapshots -->
+      <div v-if="snapshotData.snapshots.added.length > 0" class="space-y-2">
+        <details class="group" open>
+          <summary class="cursor-pointer p-2 bg-green-50 dark:bg-green-950 border border-green-200 dark:border-green-800 rounded text-green-700 dark:text-green-300 font-medium text-sm">
+            🆕 Added Snapshots ({{ snapshotData.snapshots.added.length }})
```
sheremet-va, line 122: Let's not use emojis, we have [iconify](https://iconify.design/) icons available (see dependencies for from which library)

packages/vitest/src/api/setup.ts
```diff
@@ -79,6 +80,37 @@ export function setup(ctx: Vitest, _server?: ViteDevServer): void {
           }
           return fs.writeFile(id, content, 'utf-8')
         },
+        async getSnapshotData(): Promise<SnapshotData | void> {
+          try {
+            // Get snapshot result from the Vitest context
+            const snapshotResult = ctx.snapshot.result
```
sheremet-va, line 86: Seems like this is sensitive to race conditions? `result` doesn't represent all of snapshots, or? Only for a specific file?

packages/vitest/src/api/types.ts
```diff
@@ -4,6 +4,32 @@ import type { SerializedConfig } from '../runtime/config'
 import type { SerializedTestSpecification } from '../runtime/types/utils'
 import type { Awaitable, LabelColor, ModuleGraphData, UserConsoleLog } from '../types/general'
 
+export interface SnapshotData {
+  filepath: string
+  snapshots: {
+    added: Array<{
```
sheremet-va, line 10: This can probably be its own type, so we don't need to repeat it

sheremet-va (changes requested):
I wonder if we can display them in a similar way to annotations (but hidden by default, unless there is an error). This will require storing more information in [`Test`](https://github.com/vitest-dev/vitest/blob/02da20864c661939858e55a16af1368f897685d3/packages/runner/src/types/tasks.ts#L265) (maybe snapshot names and location?) Also regarding existing code, maybe if the snapshot failed, we should show the diff?

Shinyaigeek:
@sheremet-va Thank you for your feedback!

> I wonder if we can display them in a similar way to annotations (but hidden by default, unless there is an error). This will require storing more information in [Test](https://github.com/vitest-dev/vitest/blob/02da20864c661939858e55a16af1368f897685d3/packages/runner/src/types/tasks.ts#L265) (maybe snapshot names and location?) 

Is my understanding correct that, in this PR, instead of displaying Snapshots within a dedicated Snapshot Tab, you're proposing to have an expandable Annotation Block within the Code Tab, similar to [Test Annotations in an HTML Reporter](https://vitest.dev/guide/test-annotations.html#html), at the location where Snapshot comparisons are performed in the code, and expanding it would allow users to view the Snapshot? If so, that sounds good! Also, in that case, InlineSnapshot might be ignored since its content is already present within the code itself.


> Also regarding existing code, maybe if the snapshot failed, we should show the diff?

That's true! Failed snapshots are displayed in Report Tab in current implementation, so maybe we don't need to display failed snapshot results in Code(or Snapshot) Tab duplicatedly.

sheremet-va:
> Is my understanding correct that, in this PR, instead of displaying Snapshots within a dedicated Snapshot Tab, you're proposing to have an expandable Annotation Block within the Code Tab, similar to [Test Annotations in an HTML Reporter](https://vitest.dev/guide/test-annotations.html#html), at the location where Snapshot comparisons are performed in the code, and expanding it would allow users to view the Snapshot? If so, that sounds good! Also, in that case, InlineSnapshot might be ignored since its content is already present within the code itself.

Yes, that is what I am proposing. We can probably reuse annotations for these. If needed, the type can be extended. Or we can also introduce a separate field that is only available after the test has finished (annotations appear while the test is running and have a separate reporter event which can make the reporting slower). I think I would prefer to reduce the amount of text that we have to pass down between UI and Node because it's `JSON.stringified` which takes slower the longer the string is. (Maybe we should store the name of the snapshot and load the content separately)

> That's true! Failed snapshots are displayed in Report Tab in current implementation, so maybe we don't need to display failed snapshot results in Code(or Snapshot) Tab duplicatedly.

Makes sense, let's not duplicate it.

Shinyaigeek:
Thank you! I could grasp what I should do! I will implement the snapshot visualization feature according to what should be done, so please let me more time to implement this. (perhaps, I will re-create PullRequest if the most of implementation in this PR is not needed, but I cannot judge at least now)

PR #8405, opened 2025-08-07
feat 8404 Add metadata to 

rebasecase:

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/runner/src/suite.ts
+++ b/packages/runner/src/suite.ts
@@ -325,7 +325,11 @@ function createSuiteCollector(
           : options.todo
             ? 'todo'
             : 'run',
-      meta: options.meta ?? Object.create(null),
+      meta: {
+        ...Object.create(null),
+        ...(suiteOptions?.meta || {}),
+        ...(options.meta || {}),
+      },
       annotations: [],
     }
     const handler = options.handler
--- a/packages/runner/src/types/tasks.ts
+++ b/packages/runner/src/types/tasks.ts
@@ -485,6 +485,10 @@ export interface TestOptions {
    * Whether the test is expected to fail. If it does, the test will pass, otherwise it will fail.
    */
   fails?: boolean
+  /**
+   * Custom metadata for the task. This will be merged with any meta property defined in the test.
+   */
+  meta?: Record<string, unknown>
 }
 
 interface ExtendedAPI<ExtraContext> {
--- a/test/core/test/test-meta-options.test.ts
+++ b/test/core/test/test-meta-options.test.ts
@@ -1,0 +1,69 @@
+import { beforeEach, describe, expect, test } from 'vitest'
+
+describe('TestOptions meta property functionality', { meta: { suiteLevel: 'test-suite', priority: 'medium' } }, () => {
+  let beforeEachMeta: Record<string, unknown>
+
+  beforeEach(({ task }) => {
+    beforeEachMeta = { ...task.meta }
+  })
+
+  test('should merge suite and test meta properties', { meta: { testLevel: 'individual-test', priority: 'high' } }, ({ task }) => {
+    // Test should have both suite and test meta
+    expect(task.meta).toMatchObject({
+      suiteLevel: 'test-suite',
+      testLevel: 'individual-test',
+      priority: 'high', // test meta should override suite meta
+    })
+
+    // beforeEach should have access to merged meta
+    expect(beforeEachMeta).toMatchObject({
+      suiteLevel: 'test-suite',
+      testLevel: 'individual-test',
+      priority: 'high',
+    })
+  })
+
+  test('should inherit suite meta when no test meta provided', ({ task }) => {
+    // Test should only have suite meta
+    expect(task.meta).toMatchObject({
+      suiteLevel: 'test-suite',
+      priority: 'medium',
+    })
+
+    // beforeEach should have access to suite meta
+    expect(beforeEachMeta).toMatchObject({
+      suiteLevel: 'test-suite',
+      priority: 'medium',
+    })
+  })
+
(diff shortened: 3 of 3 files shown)


packages/runner/src/types/tasks.ts
```diff
@@ -485,6 +485,10 @@ export interface TestOptions {
    * Whether the test is expected to fail. If it does, the test will pass, otherwise it will fail.
    */
   fails?: boolean
+  /**
+   * Custom metadata for the task. This will be merged with any meta property defined in the test.
+   */
+  meta?: Record<string, unknown>
```
rebasecase: This needs to be a type that is JSON serializable?

rebasecase:
- [X] Need to consider assertion of `ctx.task.suite.meta !== ctx.task.meta`
Now can distinguish between suite and task meta

rebasecase:
- [X] Need to consider nested describes, how would that work? How does the `meta` cascade, if at all?

Now cascades from grandparents

packages/runner/src/types/tasks.ts
sheremet-va: This needs to be `Partial<TaskMeta>`

packages/runner/src/suite.ts
```diff
@@ -325,7 +347,11 @@ function createSuiteCollector(
           : options.todo
             ? 'todo'
             : 'run',
-      meta: options.meta ?? Object.create(null),
+      meta: {
+        ...Object.create(null),
```
sheremet-va: What is the purpose of this line? If meta should be kept with a null prototype, you need to create it first and then assign properties with `Object.assign`, for example. This line just does nothing

packages/runner/src/suite.ts
```diff
@@ -457,7 +483,10 @@ function createSuiteCollector(
       file: undefined!,
       shuffle: suiteOptions?.shuffle,
       tasks: [],
-      meta: Object.create(null),
+      meta: {
+        ...Object.create(null),
```
sheremet-va: See https://github.com/vitest-dev/vitest/pull/8405/files#r2279369625

test/core/test/test-meta-options.test.ts
```diff
@@ -0,0 +1,125 @@
...
+    // beforeEach should have access to suite meta
+    expect(beforeEachMeta).toMatchObject({
+      suiteLevel: 'test-suite',
+      priority: 'medium',
+    })
+  })
+
+  test('should allow adding meta at runtime', { meta: { testLevel: 'runtime-test' } }, ({ task }) => {
+    // Add meta at runtime
+    (task.meta as any).runtimeAdded = 'added-during-test'
```
sheremet-va: Let's add a local type instead of assigning `any`

sheremet-va (changes requested):

rebasecase:
@sheremet-va Thank you for your review!

The comments on `Object.create(null),` do not make much sense to me yet - I am rather new to TS. I kept it there because it was there before.

`Object.create(null)` sets a prototype to an object which "is" null, but has all the available interfaces to an "Object"? The spread operator here does nothing since this "null object" has no properties _to_ spread? 

So in this instance it can be left out?

I shall fix this 😸

packages/runner/src/types/tasks.ts
rebasecase: Some places it is `Record<string, unknown>` should these all be `Partial<TaskMeta>`?

rebasecase: e.g. https://github.com/vitest-dev/vitest/blob/c1ac15c6b22903cf17b5b105dffd0b8aabf083c3/packages/vite-node/src/types.ts#L99

test/core/test/test-meta-options.test.ts
rebasecase: With meta now being `Partial<TaskMeta>` we can now assign the property directly without any types? is that okay?

rebasecase: I was wrong 😿

packages/runner/src/types/tasks.ts
rebasecase: I don't know what to do about this. Setting `Partial<TaskMeta>` cause a lot of errors now 😢  Does `TaskMeta` want to be `extends Record<string, any>` ?

I don't understand complex typing much yet.

packages/runner/src/types/tasks.ts
```diff
@@ -637,7 +641,7 @@ export interface TaskCustomOptions extends TestOptions {
   /**
    * Custom metadata for the task that will be assigned to `task.meta`.
    */
-  meta?: Record<string, unknown>
```
rebasecase, line 640: @sheremet-va This is what I meant here

packages/runner/src/suite.ts
```diff
@@ -325,7 +347,10 @@ function createSuiteCollector(
           : options.todo
             ? 'todo'
             : 'run',
-      meta: options.meta ?? Object.create(null),
+      meta: {
```
sheremet-va, line 350: Let's keep the meta with a null prototype. Just assign `options.meta` to meta collected with `collectAncestorMeta`:

```ts
const testMeta = collectAncestorMeta(collectorContext.currentSuite?.suite)
if(options.meta) {
  Object.assign(testMeta, options.meta)
}

{
  // ...
  meta: testMeta,
}
```

packages/runner/src/suite.ts
```diff
@@ -457,7 +482,9 @@ function createSuiteCollector(
       file: undefined!,
       shuffle: suiteOptions?.shuffle,
       tasks: [],
-      meta: Object.create(null),
+      meta: {
```
sheremet-va, line 485: Same here

sheremet-va (changes requested):

PR #8456, opened 2025-08-19
feat(coverage): `autoUpdate` to support percentage formatting

### LINKED ISSUE #8424: Ability to edit number of decimal places in threshold autoupdtes

As a developer using Vitest I want to have the ability to round autoupdates to thresholds down to the nearest whole number so that it... 
- looks nicer in the config file.
- doesn't require as many extra commits for tiny changes to the tresholds
- doesn't have issues where multiple coverage runs with the same code return slightly different results (less than 0.03% from what I've seen).

I would very much like to submit a PR for this, it's a pretty quick fix.

### Suggested solution

So as to not make this a breaking change I'd like to add a new config option: `coverage.thresholds.roundedAutoUpdate`. If this is set to true, I would call `Math.floor()` on the thresholds returned by `BaseCoverageProvider.resolvedThreshold()`.

### Alternative

I poked around and found one other discussion regarding this topic that did not contain any workarounds: https://github.com/vitest-dev/vitest/discussions/6143.

My solution would work great for me but I am open to suggestions.

### Additional context

_No response_

### Validations

Battjmo:
close #8424

### Description

I've found that when I have the boolean `autoUpdate` flag enabled and run coverage reports on the same code multiple times, the coverage amounts vary but up to 0.03%. This means code will arbitrarily fail to pass the coverage requirements. Also I don't really care about my coverage percentage being accurate down to two decimal places, which is how V8 reports them.

To address this, I've update the `autoUpdate` flag to accept either a boolean or a function. If it's a boolean it will keep working as it did before. If it's a function, that function will run on each coverage percentage returned by the engine and return the result, which should be a number or undefined.

So now you can do something like...

```
// in your vitest.config.mjs
{
"coverage": {
  "thresholds": {
      thresholds: {
        statements: 94,
        branches: 90,
        lines: 94,
        functions: 90,
        autoUpdate: (value) => {
          if (value) {
            // do something to format the value
            return Math.floor(value);
          }
        },
      },
}
}
}
```

Now when you run `test:coverage` or whatever your coverage script is, your coverage thresholds will come back rounded down to the nearest whole number.

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -1450,13 +1450,30 @@ Check thresholds per file.
 
 ##### coverage.thresholds.autoUpdate
 
-- **Type:** `boolean`
+- **Type:** `boolean | function`
 - **Default:** `false`
 - **Available for providers:** `'v8' | 'istanbul'`
 - **CLI:** `--coverage.thresholds.autoUpdate=<boolean>`
 
 Update all threshold values `lines`, `functions`, `branches` and `statements` to configuration file when current coverage is better than the configured thresholds.
 This option helps to maintain thresholds when coverage is improved.
+
+You can also pass a function for formatting the updated threshold values:
+
+<!-- eslint-skip -->
+```ts
+{
+  coverage: {
+    thresholds: {
+      // Update thresholds without decimals
+      autoUpdate: (newThreshold) => Math.floor(newThreshold),
+
+      // 95.85 -> 95
+      functions: 95,
+    }
+  }
+}
+```
 
 ##### coverage.thresholds.100
 
--- a/docs/guide/cli-generated.md
+++ b/docs/guide/cli-generated.md
@@ -191,7 +191,7 @@ Check thresholds per file. See `--coverage.thresholds.lines`, `--coverage.thresh
 
 ### coverage.thresholds.autoUpdate
 
-- **CLI:** `--coverage.thresholds.autoUpdate`
+- **CLI:** `--coverage.thresholds.autoUpdate <boolean|function>`
 - **Config:** [coverage.thresholds.autoUpdate](/config/#coverage-thresholds-autoupdate)
 
 Update threshold values: "lines", "functions", "branches" and "statements" to configuration file when current coverage is above the configured thresholds (default: `false`)


packages/vitest/src/node/cli/cli-config.ts
```diff
@@ -109,6 +109,16 @@ function transformNestedBoolean(value: unknown) {
   return value
 }
 
+function transformBoolean(value: unknown) {
```
AriPerkkio: Let's inline this into `transform` of `autoUpdate`

packages/vitest/src/node/types/coverage.ts
```diff
@@ -279,10 +279,11 @@ interface Thresholds {
 
   /**
    * Update threshold values automatically when current coverage is higher than earlier thresholds
+   * Also can accept a function to format the new threshold values
    *
    * @default false
    */
-  autoUpdate?: boolean
+  autoUpdate?: boolean | ((newThreshold: number | undefined) => number | undefined)
```
```suggestion
  autoUpdate?: boolean | ((newThreshold: number) => number)
```

packages/vitest/src/node/coverage.ts
```diff
@@ -578,13 +578,16 @@ export class BaseCoverageProvider<Options extends ResolvedCoverageOptions<'istan
 
       updatedThresholds = true
 
+      const thresholdFormatter = typeof this.options.thresholds?.autoUpdate === 'function' ? this.options.thresholds?.autoUpdate : (value: number | undefined) => value
```
AriPerkkio:
```suggestion
      const thresholdFormatter = typeof this.options.thresholds?.autoUpdate === 'function' ? this.options.thresholds?.autoUpdate : (value: number) => value
```

test/core/test/cli-test.test.ts
```diff
@@ -127,6 +127,17 @@ test('fails when an array is passed down for a single value', async () => {
...
+  expect(getCLIOptions('--coverage.thresholds.autoUpdate true').coverage.thresholds.autoUpdate).toBe(true)
+  expect(getCLIOptions('--coverage.thresholds.autoUpdate false').coverage.thresholds.autoUpdate).toBe(false)
+  expect(getCLIOptions('--coverage.thresholds.autoUpdate yes').coverage.thresholds.autoUpdate).toBe(true)
+  expect(getCLIOptions('--coverage.thresholds.autoUpdate no').coverage.thresholds.autoUpdate).toBe(false)
+})
+
+test('coverage autoUpdate preserves non-boolean values from CLI', async () => {
+  expect(getCLIOptions('--coverage.thresholds.autoUpdate someFunction').coverage.thresholds.autoUpdate).toBe('someFunction')
+})
+
```
AriPerkkio: This test doesn't make sense:

```suggestion
```

test/coverage-test/test/threshold-auto-update.unit.test.ts
```diff
@@ -130,11 +130,79 @@ export default config
   await expect(updateThresholds(config)).rejects.toThrowErrorMatchingInlineSnapshot(`[Error: Failed to update coverage thresholds. Configuration file is too complex.]`)
 })
 
+test('autoUpdate function rounds down decimal coverage to whole numbers', async () => {
+  // Create a config with autoUpdate function that rounds down to whole numbers
+  const configWithFunction = JSON.stringify(defineConfig({
+    test: {
+      coverage: {
+        thresholds: {
+          autoUpdate: (value: number | undefined) => value ? Math.floor(value) : value,
```
AriPerkkio:
```suggestion
          autoUpdate: value => Math.floor(value),
```

AriPerkkio (changes requested):

AriPerkkio (approved):
Looks good, thanks for the feature! 💯

Battjmo:
Thanks @AriPerkkio!

PR #8461, opened 2025-08-20
perf: remove chai as a direct dependency, keep it in `@vitest/expect`

sheremet-va:
### Description

We were also installing chai in both `@vitest/expect` and `vitest`. This duplicated the amount of work Node.js needs to do. Removing `chai` from `vitest` (but keeping it in `expect`) removes ~10ms at startup for every test file.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/expect/package.json
+++ b/packages/expect/package.json
@@ -36,10 +36,10 @@
     "@types/chai": "catalog:",
     "@vitest/spy": "workspace:*",
     "@vitest/utils": "workspace:*",
+    "chai": "catalog:",
     "tinyrainbow": "catalog:"
   },
   "devDependencies": {
-    "@vitest/runner": "workspace:*",
-    "chai": "catalog:"
+    "@vitest/runner": "workspace:*"
   }
 }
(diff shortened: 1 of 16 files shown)


packages/expect/package.json
```diff
@@ -36,11 +36,10 @@
     "@types/chai": "catalog:",
     "@vitest/spy": "workspace:*",
     "@vitest/utils": "workspace:*",
-    "chai": "catalog:",
     "tinyrainbow": "catalog:"
   },
   "devDependencies": {
     "@vitest/runner": "workspace:*",
-    "rollup-plugin-copy": "^3.5.0"
+    "chai": "catalog:"
```
AriPerkkio: Latest `^5` and `6` versions should be good now, no need for bundling.

```js
const startTime = performance.now();

await import("chai");
// "chai": "5.2.1" > 4.39 ms
// "chai": "5.3.1" > 13.52 ms
// "chai": "5.3.2" > 2.92 ms
// "chai": "6.0.0" > 2.87 ms

console.log(`${(performance.now() - startTime).toFixed(2)} ms`);
```

sheremet-va: Sounds good. Let's keep `chai` in dependencies.

PR #8487, opened 2025-08-27
fix: correctly remove undefined values in options

TrevorBurnham:
### Description

This `removeUndefinedValues` utility function was added by #2281. However, there's a bug in the implementation: It performs an `in` loop over the array returned by `Object.keys`, which means that it iterates over the array indices `1, 2, 3...` instead of the actual object keys. That makes it effectively a no-op.

Since no one has noticed this in 3 years, simply removing the function would be one option. It's not clear to me why the `TypeError` described in #2153 occurred. However, this PR goes with the other option: Implementing the intended behavior by removing the `Object.keys` so that `in` iterates over the object's keys.

- [ ] Ideally, include a test that fails without this PR but passes with it. (**Author note:** Unfortunately, there's no existing test coverage for this part of the codebase.)

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/plugins/index.ts
+++ b/packages/vitest/src/node/plugins/index.ts
@@ -308,10 +308,11 @@ export async function VitestPlugin(
     ModuleRunnerTransform(),
   ].filter(notNullish)
 }
+
 function removeUndefinedValues<T extends Record<string, any>>(
   obj: T,
 ): T {
-  for (const key in Object.keys(obj)) {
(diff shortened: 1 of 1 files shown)


sheremet-va:
This is really funny, let's just remove the call altogether to keep the current behavior.

TrevorBurnham:
> This is really funny, let's just remove the call altogether to keep the current behavior.

Roger that! Closing in favor of #8501.

PR #8500, opened 2025-08-28
fix!: verbose reporter prints tests in a list, introduce `tree` reporter

sheremet-va:
### Description

Make the `verbose` output consistent. This PR:

- Makes `verbose` output print only test case right after they are finished
- Moves the `tree` view into a `--reporter=tree`
- Prints tests as a tree when there is only one test file running (previously it was printing them as a list)
- Removes inline error messages in `default` and `tree` view, they are kept only in `verbose` reporter 

TODO:
- [x] Tests
- [x] Document `tree` reporter


hi-ogawa (approved):

test/reporters/tests/verbose.test.ts
```diff
@@ -80,10 +72,7 @@ test('prints repeat count', async () => {
...
   })
 
-  expect(trimReporterOutput(stdout)).toMatchInlineSnapshot(`
-    "✓ fixtures/repeats.test.ts (1 test) [...]ms
-       ✓ repeat couple of times [...]ms (repeat x3)"
-  `)
+  expect(trimReporterOutput(stdout)).toMatchInlineSnapshot(`"✓ fixtures/repeats.test.ts > repeat couple of times (repeat x3) [...]ms"`)
 })
 
 test('renders tree when in TTY', async () => {
```
AriPerkkio: This is no longer a tree

docs/guide/migration.md
```diff
@@ -267,6 +267,7 @@ Vitest 4.0 removes some deprecated APIs, including:
 - `poolMatchGlobs` config option. Use [`projects`](/guide/projects) instead.
 - `environmentMatchGlobs` config option. Use [`projects`](/guide/projects) instead.
 - Reporter APIs `onCollected`, `onSpecsCollected`, `onPathsCollected`, `onTaskUpdate` and `onFinished`. See [`Reporters API`](/advanced/api/reporters) for new alternatives. These APIs were introduced in Vitest `v3.0.0`.
+- The `verbose` reporter now always prints tests cases even if the terminal is TTY (previously, it would fallback to the `default` reporter in that case).
```
AriPerkkio: Does this mean that in TTY, `--reporter verbose` is equal to `--reporter default`? I don't think it is.

sheremet-va: It basically is. Most of the rendering is done by the default reporter in that case

AriPerkkio: Default reporter does not print test cases, verbose does. https://vitest.dev/guide/reporters.html#verbose-reporter

Default:

```
 ✓  forks  test/mocking/factory.test.ts (7 tests) 10ms
```

Verbose:

```
✓  forks  test/mocking/factory.test.ts (7 tests) 7ms
   ✓ mocking with factory (7)
     ✓ missing exports on mock 4ms
     ✓ non-object return on factory gives error 2ms
     ✓ defined exports on mock 0ms
     ✓ successfully with actual 0ms
     ✓ successfully with factory helper 0ms
     ✓ mocks node_modules 0ms
     ✓ logger extended 0ms
```

sheremet-va: Yeah, but it's done by `verbose: true`, not by the reporter

sheremet-va: > Yeah, but it's done by `verbose: true`, not by the reporter

Which is basically the same, I guess

In any way, we need to have the same view for TTY and non-TTY - either remove tree view (which is done here) or remove the long names view

AriPerkkio: I think the internal `verbose` option of reporters affects only error printing and summary. The rendering of tests and suites is done by BaseReporter and how VerboseReporter overrides some default options.

> In any way, we need to have the same view for TTY and non-TTY - either remove tree view (which is done here) or remove the long names view

That makes sense. I've always wondered why TTY/non-TTY works differently.

docs/guide/reporters.md
```diff
@@ -141,9 +141,27 @@ Final output after tests have finished:
    Duration  1.26s (transform 35ms, setup 1ms, collect 90ms, tests 1.47s, environment 0ms, prepare 267ms)
 ```
 
+If there is only one test file running, Vitest will output the full test tree of that file, simillar to the [`tree`](#tree-reporter) reporter. The default reporter will also print the test tree if there is at least one failed test in the file.
```
AriPerkkio, line 144: `default` reporter does not use tree:

```ts
import { describe, expect, test, } from "vitest";
import { add } from "../src/add";

describe("first level", () => {
  describe("second level", () => {
    test("example", async () => {
      add(1, 0);
    });
  });
});
```

```sh
$ pnpm run test --reporter default test/add.test.ts 

 ✓ test/add.test.ts (1 test) 1ms
   ✓ first level > second level > example 1ms

$ pnpm run test --reporter verbose test/add.test.ts 

 ✓ test/add.test.ts (1 test) 1ms
   ✓ first level (1)
     ✓ second level (1)
       ✓ example 1ms
```

docs/guide/test-annotations.md
```diff
@@ -53,7 +53,7 @@ Error: thrown error
 
 ### verbose
 
-In a TTY terminal, the `verbose` reporter works similarly to the `default` reporter. However, in a non-TTY environment, the `verbose` reporter will also print annotations after every test.
+The `verbose` reporter is the only terminal reporter that reports annotation when the test doesn't fail.
```
AriPerkkio:
```suggestion
The `verbose` reporter is the only terminal reporter that reports annotations when the test doesn't fail.
```

docs/guide/reporters.md
AriPerkkio: Oh wait this PR actually makes `default` reporter to print tree now. Is that intentional? Previously it did not do that.

AriPerkkio: Mentioned in PR description, alright. ✅

AriPerkkio: I'm not sure if that's good though. Tree view should only be used in `tree` reporter.

sheremet-va: Yes, this is intentional. We discussed this during the meeting

test/reporters/tests/verbose.test.ts
```diff
@@ -106,35 +95,29 @@ test('renders tree when in TTY', async () => {
...
+     ↓ fixtures/verbose/example-1.test.ts > suite skip in root > suite in suite > test in nested suite
+     ↓ fixtures/verbose/example-1.test.ts > suite skip in root > suite in suite > test failure in nested suite of skipped suite
+     ✓ fixtures/verbose/example-2.test.ts > test 0.1 [...]ms
+     ↓ fixtures/verbose/example-2.test.ts > test 0.2
+     ✓ fixtures/verbose/example-2.test.ts > suite 1.1 > test 1.1 [...]ms"
   `)
 })
 
-test('does not render tree when in non-TTY', async () => {
+test('renders lications if enabled', async () => {
```
AriPerkkio:
```suggestion
test('renders locations if enabled', async () => {
```

test/reporters/tests/tree.test.ts
```diff
@@ -0,0 +1,141 @@
...
+             ✓ nested a failed 2 test [...]ms
+             ✓ nested a failed 3 test [...]ms
+         ✓ a skipped (3)
+           ↓ skipped with note [...]ms [reason]
+           ↓ condition [...]ms
+           ↓ condition with note [...]ms [note]"
+    `)
+  })
+
+  it('reports multiple tests when all passed as a tree', async () => {
```
AriPerkkio: Took me a while to realize why some tests are reported as skipped, let's specify it in name too:

```suggestion
  it('reports multiple tests when all passed with testname filter as a tree', async () => {
```

AriPerkkio (approved):

AriPerkkio (approved):

PR #8516, opened 2025-09-01
perf(vm): optimize memory management and module loading performance

TrevorBurnham:
### Description

This PR is an effort to improve VM performance around module loading.

The biggest change is adding an upper bound to the size of the `FileMap` cache to prevent OOM errors, using LRU to evict files if the cache approaches the limit. That upper bound is set to 50MB, which is an arbitrary number. This could be made configurable, though in practice I'd expect 50MB to be small enough to prevent OOM while large enough to provide adequate caching in the vast majority of cases. (The upper bound is also somewhat fuzzy because calculating the exact size of files in the cache would be expensive, so I opted to use `length * 1.2` as an approximation.)

As an optimization, this PR adds logic to prevent duplicate module loading in the ESM executor. It also applies several micro-optimizations, such as using `Map` instead of `Object` for cache lookups.

The benchmarks show no improvement on net, but this should be a win in terms of avoiding OOM errors, and the duplicate module loading check should produce gains in some scenarios.

### Tests
- [ ] Run the tests with `pnpm test:ci`. (**Author's note:** These tests don't seem to work on my Mac.)

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/runtime/vm/commonjs-executor.ts
+++ b/packages/vitest/src/runtime/vm/commonjs-executor.ts
@@ -26,7 +26,7 @@ export class CommonjsExecutor {
   private publicRequireCache = this.createProxyCache()
 
   private moduleCache = new Map<string, VMSyntheticModule>()
-  private builtinCache: Record<string, NodeJS.Module> = Object.create(null)
+  private builtinCache = new Map<string, NodeJS.Module>()
   private extensions: Record<
     string,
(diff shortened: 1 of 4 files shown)


sheremet-va:
Seems like CI is failing due to these changes.

TrevorBurnham:
Sorry, this one needs more time in the oven. 👨‍🍳

PR #8527, opened 2025-09-04
feat: add schema validation matchers

zirkelc:
### Description

This PR implements a simple proof of concept for the feature suggested in #8485 to use a [standard schema ](https://github.com/standard-schema/standard-schema) like Zod, Valibot and ArkType for validating data.

Three options have been implemented for further discussion:

- custom equality tester for `expect.toEqual()`
- custom matcher `expect.toEqualSchema()`
- asymmetric matcher `expect.schemaMatching()`

Usage:

```ts
const schema = z.object({
  name: z.string(),
  email: z.email(),
});
const validData = { name: 'John', email: 'john@example.com' };

// custom equality test using toEqual
expect(validData).toEqual(schema);
expect({ email: "john@example.com" }).toEqual({ email: z.email() });

// custom matcher using toEqualSchema
expect(validData).toEqualSchema(schema);
expect({ email: "john@example.com" }).toEqualSchema({ email: z.email() });

// asymmetric matcher
expect(validData).toEqual(expect.schemaMatching(schema));
expect({ email: "john@example.com" }).toEqual(expect.schemaMatching({ email: z.email() }));
```

### `expect.toEqual(schema)`

The custom equality tester for `expect.toEqual()` has the problem that a failing assertions prints the stringified schema as **Expected** value instead of the issues from the validation:

image

I assume there is no easy way to overwrite the message, please correct me if I'm wrong.

### `expect.toEqualSchema(schema)`

The custom matcher for `expect.toEqualSchema()` has the advantage that we have full control over the assertion message in case of a failure:

image

However, if `expect.toEqualSchema()` is used inside an asymmetric matcher like `expect.objectContaining()` and fails, it also prints the stringified schema as **Expected** value like `expect.toEqual()`:

image

### `expect.schemaMatching(schema)`

The asymmetric matcher `expect.schemaMatching(schema)` seems to be the best candidate since it gives us full control over the assertion messages, especially when used inside another asymmetric matcher like `expect.objectContaining`:

image

---

I would lean towards implementing it as a asymmetric matcher since it allows the most customisation options and is well encapsulated from the rest and therefore reduces the risk of unknown side-effects on existing matchers likes `toEqual()`.

Please let me know if and how you'd like to proceed.

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/advanced/api/test-suite.md
+++ b/docs/advanced/api/test-suite.md
@@ -197,14 +197,16 @@ Note that errors are serialized into simple objects: `instanceof Error` will alw
 function meta(): TaskMeta
 ```
 
-Custom [metadata](/advanced/metadata) that was attached to the suite during its execution or collection. The meta can be attached by assigning a property to the `task.meta` object during a test run:
+Custom [metadata](/advanced/metadata) that was attached to the suite during its execution or collection. The meta can be attached by assigning a property to the `suite.meta` object during a test run:
 
-```ts {5,10}
+```ts {7,12}
 import { test } from 'vitest'
+import { getCurrentSuite } from 'vitest/suite'
 
-describe('the validation works correctly', (task) => {
+describe('the validation works correctly', () => {
   // assign "decorated" during collection
-  task.meta.decorated = false
+  const { suite } = getCurrentSuite()
+  suite!.meta.decorated = true
 
   test('some test', ({ task }) => {
     // assign "decorated" during test run, it will be available
--- a/docs/guide/browser/config.md
+++ b/docs/guide/browser/config.md
@@ -492,3 +492,116 @@ For example, to store diffs in a subdirectory of attachments:
 resolveDiffPath: ({ arg, attachmentsDir, browserName, ext, root, testFileName }) =>
   `${root}/${attachmentsDir}/screenshot-diffs/${testFileName}/${arg}-${browserName}${ext}`
 ```
+
+#### browser.expect.toMatchScreenshot.comparators
+
+- **Type:** `Record<string, Comparator>`
+
+Register custom screenshot comparison algorithms, like [SSIM](https://en.wikipedia.org/wiki/Structural_similarity_index_measure) or other perceptual similarity metrics.
+
+To create a custom comparator, you need to register it in your config. If using TypeScript, declare its options in the `ScreenshotComparatorRegistry` interface.
+
+```ts
+import { defineConfig } from 'vitest/config'
+
+// 1. Declare the comparator's options type
+declare module 'vitest/browser' {
+  interface ScreenshotComparatorRegistry {
+    myCustomComparator: {
+      sensitivity?: number
+      ignoreColors?: boolean
+    }
+  }
+}
+
+// 2. Implement the comparator
+export default defineConfig({
+  test: {
+    browser: {
+      expect: {
+        toMatchScreenshot: {
+          comparators: {
+            myCustomComparator: async (
+              reference,
+              actual,
+              {
+                createDiff, // always provided by Vitest
+                sensitivity = 0.01,
+                ignoreColors = false,
+              }
+            ) => {
+              // ...algorithm implementation
+              return { pass, diff, message }
+            },
+          },
+        },
+      },
+    },
+  },
+})
+```
+
+Then use it in your tests:
+
+```ts
+await expect(locator).toMatchScreenshot({
+  comparatorName: 'myCustomComparator',
+  comparatorOptions: {
+    sensitivity: 0.08,
+    ignoreColors: true,
+  },
+})
+```
+
+**Comparator Function Signature:**
+
+```ts
+type Comparator<Options> = (
+  reference: {
+    metadata: { height: number; width: number }
+    data: TypedArray
+  },
+  actual: {
+    metadata: { height: number; width: number }
+    data: TypedArray
+  },
+  options: {
+    createDiff: boolean
+  } & Options
+) => Promise<{
+  pass: boolean
+  diff: TypedArray | null
+  message: string | null
+}> | {
+  pass: boolean
+  diff: TypedArray | null
+  message: string | null
+}
+```
+
+The `reference` and `actual` images are decoded using the appropriate codec (currently only PNG). The `data` property is a flat `TypedArray` (`Buffer`, `Uint8Array`, or `Uint8ClampedArray`) containing pixel data in RGBA format:
+
+- **4 bytes per pixel**: red, green, blue, alpha (from `0` to `255` each)
+- **Row-major order**: pixels are stored left-to-right, top-to-bottom
+- **Total length**: `width × height × 4` bytes
+- **Alpha channel**: always present. Images without transparency have alpha values set to `255` (fully opaque)
+
+::: tip Performance Considerations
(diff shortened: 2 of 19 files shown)


sheremet-va:
> The custom equality tester for `expect.toEqual()` has the problem that a failing assertions prints the stringified schema as **Expected** value instead of the issues from the validation:

The printing is done by `@vitest/pretty-format` package which is a fork of [`pretty-format`](http://npmjs.com/pretty-format). We can just add a plugin similar to the asymmetric matchers one:

https://github.com/vitest-dev/vitest/blob/de74785e052649c12a415d290dbce7ad6ad76be4/packages/pretty-format/src/plugins/AsymmetricMatcher.ts#L17

For the issues themselves, I think it's time to add something like `getLastErrors` that returns the `message()` from the asymmetric matcher:

```ts
const isEqual = equals(a, b) // keep boolean
const errors = getLastErrors() // returns errors generated during equals
```

Both are sync so there should be no race condition. Another thing we can do is add a new function. Something like `equalsWthErrors(a, b)` that returns `{ equals: boolean, errors: string[] }` (names are just examples).

sheremet-va:
One nice thing with wrapping it as a matcher is that we can keep the issues as the state and return the text from toString (as example):

```ts
class Matcher {
  asymmetricMatch() {
    const valid = isEqual(a, b)
    this.issues = valid.issues
    return valid.valid
  }

  toString() {
    return `Schema<${this.issues}>`
  }
}
```

zirkelc:
> > The custom equality tester for `expect.toEqual()` has the problem that a failing assertions prints the stringified schema as **Expected** value instead of the issues from the validation:
> 
> The printing is done by `@vitest/pretty-format` package which is a fork of [`pretty-format`](http://npmjs.com/pretty-format). We can just add a plugin similar to the asymmetric matchers one:
> 
> https://github.com/vitest-dev/vitest/blob/de74785e052649c12a415d290dbce7ad6ad76be4/packages/pretty-format/src/plugins/AsymmetricMatcher.ts#L17

I had the same idea, but wasn't sure if you'll want to add a schema-specific plugin to pretty-format. That's the main reason I'm preferring an asymmetric matcher like `expect.schemaMatching()` (or another name) because the `AsymmetricMatcher` already handles it.

> For the issues themselves, I think it's time to add something like `getLastErrors` that returns the `message()` from the asymmetric matcher:
> 
> ```ts
> const isEqual = equals(a, b) // keep boolean
> const errors = getLastErrors() // returns errors generated during equals
> ```
> 
> Both are sync so there should be no race condition. Another thing we can do is add a new function. Something like `equalsWthErrors(a, b)` that returns `{ equals: boolean, errors: string[] }` (names are just examples).

This would definitely make sense and be a good improvement for other matchers that need more information than just `pass=true/false`. However, I have no intuitive feeling how much effort this would be.


> One nice thing with wrapping it as a matcher is that we can keep the issues as the state and return the text from toString (as example):
> 
> ```ts
> class Matcher {
>   asymmetricMatch() {
>     const valid = isEqual(a, b)
>     this.issues = valid.issues
>     return valid.valid
>   }
> 
>   toString() {
>     return `Schema<${this.issues}>`
>   }
> }
> ```

I think that's what I already do with `expect.schemaMatching()`, or do you mean something specific? Except I'm using the `toAsymmetricMatcher` function to format the issues instead of `toString`. Would have to check why I went with `toAsymmetricMatcher`.

sheremet-va:
> I think that's what I already do with `expect.schemaMatching()`, or do you mean something specific? Except I'm using the `toAsymmetricMatcher` function to format the issues instead of `toString`. Would have to check why I went with `toAsymmetricMatcher`.

You are right, it is what you are already doing, I misremembered your implementation when I left a comment.

> This would definitely make sense and be a good improvement for other matchers that need more information than just pass=true/false. However, I have no intuitive feeling how much effort this would be.

You are right here, we do not need it for this implementation.

Overall, I agree that `expect.schemaMatching(schema)` seems to be the best API out of all of them.

zirkelc:
> > I think that's what I already do with `expect.schemaMatching()`, or do you mean something specific? Except I'm using the `toAsymmetricMatcher` function to format the issues instead of `toString`. Would have to check why I went with `toAsymmetricMatcher`.
> 
> You are right, it is what you are already doing, I misremembered your implementation when I left a comment.

I added the asymmetric matcher in a later commit, maybe when you checked the PR it was not committed yet.

> Overall, I agree that `expect.schemaMatching(schema)` seems to be the best API out of all of them.

What so you think naming-wise? I tried to align it with the existing asymmetric matchers, but something shorter like it has been discussed on the issue also has its appeal. Off the top of my head, I can think of `expect.satisfies()`, `expect.matching()/matches()`, and `expect.schema()`.

zirkelc:
@sheremet-va I just wanted to check how we should proceed with this PR. Should I remove all other matcher except the asymmetric one, or do you want to discuss the current options with the team internally first?

sheremet-va:
> @sheremet-va I just wanted to check how we should proceed with this PR. Should I remove all other matcher except the asymmetric one, or do you want to discuss the current options with the team internally first?

Yes, please only keep the asymmetric one.

zirkelc:
@sheremet-va I removed the matchers. 

What about the naming: something shorter like `expect.schema()` has it's appealing. What do you think?

sheremet-va:
> What about the naming: something shorter like `expect.schema()` has it's appealing. What do you think?

Maybe `expect.schemaMatching` to align with `expect.stringMatching`?

zirkelc:
> > @sheremet-va I just wanted to check how we should proceed with this PR. Should I remove all other matcher except the asymmetric one, or do you want to discuss the current options with the team internally first?
> 
> Yes, please only keep the asymmetric one.

Okay, then we can keep the current version!

I added the new matcher to the docs. Let me know if there is anything else to do! 🙏

zirkelc:
@sheremet-va the tests keep failing due to install errors. Is there something I can do?

test/core/test/expect.test.ts
```diff
@@ -477,3 +478,113 @@ describe('expect with custom message', () => {
...
+    })
+
+    test('should work with objects', () => {
+      expect({
+        email: 'john@example.com',
+      }).toEqual({
+        email: expect.schemaMatching(emailSchema),
+      })
+
+      expect(() => expect({
```
sheremet-va, line 559: Can we have more failing examples that also check the generated `error.diff`?

sheremet-va:
> @sheremet-va the tests keep failing due to install errors. Is there something I can do?

Your pnpm-lockfile is incorrect, reinstall dependencies

zirkelc:
> > @sheremet-va the tests keep failing due to install errors. Is there something I can do?
> 
> Your pnpm-lockfile is incorrect, reinstall dependencies

I tried that multiple times, but the actual issue was that I was using pnpm v9 and package.json was set at v10. Upgrading pnpm solved it.

Can you look at the current errors, it looks like it's a general CI error. Build and test runs locally.

sheremet-va:
> Can you look at the current errors, it looks like it's a general CI error. Build and test runs locally.

The error happens because you are updating all packages instead of only installing `@standard-schema/spec`. Try to keep the number of changes in the lockfile to a minimum.

zirkelc:
> The error happens because you are updating all packages instead of only installing `@standard-schema/spec`. Try to keep the number of changes in the lockfile to a minimum.

Thanks, I wasn't aware of that - I had merge conflicts on the lockfile when pulling in upstream changes and I just deleted it and re-installed it, probably causing many lockfile changes.

I added failing tests with `error.diff` snapshots. The failing browser tests look unrelated, but no idea regarding the failing macos build&test.

sheremet-va:
LGTM

Issue #8482, opened 2025-08-26
Feature Request: Enhanced `retry` options


matzeeable:
### Clear and concise description of the problem

Vitest supports a [`retry`](https://vitest.dev/config/#retry) mechanism, but currently it only retries a fixed number of times on **any kind of error**.

This limits flexibility in several ways:

- It’s not possible to retry **only on specific error types** (e.g. `TimeoutError`).
- There is no option to **delay retries** (useful for APIs, DBs, or rate-limited services).
- There is no option to **defer retries until the end of the test file**, which can improve flow and reduce interference with other tests.

Other test runners (e.g. Jest, WebdriverIO) provide these options:

- Webdriver.io and their [`specFileRetriesDelay` + `specFileRetriesDeferred`](https://webdriver.io/docs/configuration/#specfileretries) option.
- Jest and the [`retryTimes(numRetries, { waitBeforeRetry: 100, retryImmediately: false })`](https://jestjs.io/docs/jest-object#jestretrytimesnumretries-options) function.

### Suggested solution

Extend the `retry` API with the following features:

- `retryStrategy: boolean` (default `false`)
  - `immediate`: retry failed test immediately.
  - `test-file`: run retries until the end of the test file.
  - `deferred`: defer retries after all other tests have run across all test files.

- `retryDelay: number` (milliseconds, default `0`). Or even better, use https://github.com/vitest-dev/vitest/issues/8095

- `retryCondition: string | ((error: Error) => boolean)`: If `string` (Regexp), it should only retry when the occurred `Error` message matches the Regexp. If function, it should only retry when the function returns `true`.

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vitest.dev/guide/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

sheremet-va:
This sounds like a great idea. PR is welcome!

nehaaprasad:
hi @matzeeable 
implemented the enhanced retry options feature you requested - vitest now supports retryStrategy, retryDelay, and retryCondition with full backward compatibility.

jamesgpearce:
Not sure what happened to this PR but would have loved this feature!

matanshavit:
[...]

PR #8548, opened 2025-09-07
feat: add enhanced retry options with delay and condition support 

naaa760:
### Description

**This PR implements enhanced retry options for Vitest tests**, extending the current simple retry mechanism with advanced configuration capabilities.

fixes: #8482

### What This PR Solves

The current Vitest retry mechanism only supports a fixed number of retries on any kind of error. This PR adds:

#### **New Retry Features:**
- **`retryStrategy`**: Choose when retries are executed (`"immediate"`, `"test-file"`, `"deferred"`)
- **`retryDelay`**: Configurable delay between retries with exponential backoff
- **`retryCondition`**: Conditional retry based on error type (string patterns, RegExp, or custom functions)

#### **Backward Compatibility:**
- Fully maintains existing `retry: number` syntax
- No breaking changes to current API
- All existing tests continue to work

### Implementation Details

**Files Modified:**
- Configuration types in `packages/vitest/src/node/types/config.ts`
- Core retry logic in `packages/runner/src/run.ts` 
- CLI support in `packages/vitest/src/node/cli/cli-config.ts`
- Utility functions in `packages/vitest/src/utils/retry.ts` (new file)

**Key Features:**
-  Exponential backoff delay mechanism
-  Flexible error condition matching
-  TypeScript type safety
-  CLI configuration support
-  Comprehensive test coverage


packages/runner/src/types/runner.ts
```diff
@@ -35,7 +35,12 @@ export interface VitestRunnerConfig {
   maxConcurrency: number
   testTimeout: number
   hookTimeout: number
-  retry: number
+  retry: number | {
```
sheremet-va: I see that the type is repeated often. Maybe it should be a separate type in `./tasks`?

packages/runner/src/run.ts
```diff
@@ -340,13 +342,18 @@ export async function runTest(test: Test, runner: VitestRunner): Promise<void> {
...
-          else if (test.repeats && retry === retryCount) {
+          else if (test.repeats && retryConfig.count === retryCount) {
             test.result.state = 'pass'
           }
         }
       }
       catch (e) {
         failTask(test.result, e, runner.config.diffOptions)
+
+        const error = test.result.errors?.[0]?.error as Error
```
sheremet-va: Can you explain this line, please? Why do we need to go to `test.result.errors?.[0]?.error` when `e` is available already?

packages/runner/src/run.ts
```diff
@@ -401,18 +408,20 @@ export async function runTest(test: Test, runner: VitestRunner): Promise<void> {
...
         test.result.state = 'run'
         test.result.retryCount = (test.result.retryCount ?? 0) + 1
+
+        if (retryConfig.delay > 0) {
+          const delayMs = calculateBackoffDelay(retryConfig.delay, retryCount, 'exponential')
+          await sleep(delayMs)
+        }
       }
 
-      // update retry info
```
sheremet-va, line 411: Let's not remove comments not related to the changes made

packages/runner/src/run.ts
```diff
@@ -401,18 +408,20 @@ export async function runTest(test: Test, runner: VitestRunner): Promise<void> {
...
       updateTask('test-retried', test, runner)
     }
   }
 
-  // if test is marked to be failed, flip the result
```
sheremet-va: Let's not remove comments not related to the changes made

packages/vitest/src/node/cli/cli-config.ts
```diff
@@ -602,8 +602,26 @@ export const cliOptionsConfig: VitestCLIOptions = {
   },
   retry: {
     description:
-      'Retry the test specific number of times if it fails (default: `0`)',
-    argument: '<times>',
+      'Retry configuration. Can be a number or JSON string with advanced options (default: `0`)',
```
sheremet-va: It cannot be a JSON string, the value would be defined as `--retry.count=5 --retry.strategy=immediate`

packages/vitest/src/node/cli/cli-config.ts
```diff
@@ -602,8 +602,26 @@ export const cliOptionsConfig: VitestCLIOptions = {
...
+        description: 'Retry strategy: immediate, test-file, or deferred (default: immediate)',
+        argument: '<strategy>',
+      },
+      delay: {
+        description: 'Delay in milliseconds between retries (default: 0)',
+        argument: '<milliseconds>',
+      },
+      condition: {
+        description: 'Condition for retrying: RegExp pattern or custom function',
+        argument: '<pattern|function>',
```
sheremet-va: I don't think you can pass a function in the CLI, so maybe this should have a transform method. RegExp also cannot be passed down and automatically inferred as a RegExp

packages/vitest/src/node/cli/cli-config.ts
```diff
@@ -602,8 +602,26 @@ export const cliOptionsConfig: VitestCLIOptions = {
   },
   retry: {
```
sheremet-va, line 603: Please, add a test for CLI input to `test/core/cli-test.test.ts`

packages/vitest/src/node/cli/cli-config.ts
```diff
@@ -602,8 +602,26 @@ export const cliOptionsConfig: VitestCLIOptions = {
...
+      strategy: {
+        description: 'Retry strategy: immediate, test-file, or deferred (default: immediate)',
+        argument: '<strategy>',
+      },
+      delay: {
+        description: 'Delay in milliseconds between retries (default: 0)',
+        argument: '<milliseconds>',
+      },
+      condition: {
+        description: 'Condition for retrying: RegExp pattern or custom function',
```
sheremet-va: I think this needs an explanation where this pattern is applied. According to types, you can also pass down a string, but this message doesn't mention it

packages/vitest/src/node/types/config.ts
```diff
@@ -741,12 +741,12 @@ export interface InlineConfig {
    */
   bail?: number
 
-  /**
-   * Retry the test specific number of times if it fails.
-   *
-   * @default 0
-   */
-  retry?: number
+  retry?: number | {
```
sheremet-va, line 750: Let's keep comments, please

packages/vitest/src/utils/retry.ts
```diff
@@ -0,0 +1,70 @@
+import type { SerializedConfig } from '../runtime/config'
```
sheremet-va, line 1: This whole file should be in `packages/runner`. There is no reason to keep it in another package

packages/vitest/src/utils/retry.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+export function sleep(ms: number): Promise<void> {
+  return new Promise(resolve => setTimeout(resolve, ms))
+}
+
+export function calculateBackoffDelay(
+  baseDelay: number,
+  retryCount: number,
+  algorithm: 'linear' | 'exponential' = 'linear'
```
sheremet-va, line 61: I am a bit confused why this needs to be customizable when we always pass down `exponential`. Should this check be removed?

sheremet-va:
The tests are failing.

packages/runner/src/types/tasks.ts
```diff
@@ -442,18 +442,19 @@ type ChainableTestAPI<ExtraContext = object> = ChainableFunction<
...
    * Test timeout.
    */
   timeout?: number
-  /**
-   * Times to retry the test if fails. Useful for making flaky tests more stable.
-   * When retries is up, the last test error will be thrown.
-   *
-   * @default 0
-   */
-  retry?: number
```
flx-sta, line 456: Shouldn't the doc-comment be replaced/extended instead of removed completely?

PR #8574, opened 2025-09-14
chore: consolidate TypeScript configurations

TrevorBurnham:
### Description

Currently, there's a lot of variation in individual `tsconfig.json` files in this monorepo. There's a central `tsconfig.base.json` that sets `target` and `lib` to ESNext, which is [an imprecise target](https://www.typescriptlang.org/tsconfig/#target). This PR updates each individual `tsconfig.json` to extend from one of three base configurations:

* `tsconfig.browser.json` (for browsers code)
* `tsconfig.node.json` (for Node.js code)
* `tsconfig.node-dom.json` (for Node.js code that runs in a DOM environment)

Those new base configurations use precise ECMAScript targets: ES2023 for Node.js, ES2021 for browsers.

Per https://github.com/vitest-dev/vitest/discussions/8489, I've also updated all packages that target Node.js 14 to instead target Node.js 18, and added an explicit `engines` declaration to that effect where appropriate.

#### Build size and performance

These changes reduce build size across all packages by a total of 23.7KB, or about ~0.6%.

Benchmarks show no effect on runtime performance, though I'd expect the smaller Node.js modules to slightly improve startup time.

#### Type check accuracy

Previously, the `tsconfig.check.json` used by the `typecheck` script simply inherited the compiler options from `tsconfig.base.json`, causing mismatch when type checking packages with different requirements.

This PR switches to using [references](https://www.typescriptlang.org/tsconfig/#references) so that each package's individual `tsconfig.json` is respected when running the `typecheck` script. That should improve type check accuracy and reduce configuration duplication.

### Tests
- [ ] Run the tests with `pnpm test:ci`. (**Author's note:** Several tests fail on `main` on my Mac.)

### Changesets


TrevorBurnham:
Ugh, unfortunately I discovered that the [references](https://www.typescriptlang.org/tsconfig/#references) approach has a fatal flaw:

* References only work when using `--build` (as I've done here)
* But using `--build` is incompatible with `noEmit`, making it unsuitable for the `typecheck` script

I'd thought things were working because there were no build artifacts, but it turns out that the behavior of `--build` is to generate no build artifacts if (and only if) the cache is up-to-date. So if you make any changes to the source code, the `typecheck` script here would generate build artifacts. 😭 

I'll close this out and see what I can salvage from it.

PR #8611, opened 2025-09-22
docs: name of `unhandledRejection` event

SimonSimCity:
### Description

The documentation seems to be swapping the case-sensitive naming of the _unhandled rejection_ event. At first, I'll here do a quick overview of where I think this confusion comes from:

In the HTML spec, the handler for the `onunhandledrejection` event is defined as [`unhandledrejection`](https://html.spec.whatwg.org/multipage/webappapis.html#handler-window-onunhandledrejection), which is case-sensitive.

But since the tests of vitest do not (yet) fully support running in the browser, we have to stick to what NodeJS is giving us, where the event has a different capitalization. There it's defined as [`unhandledRejection`](https://nodejs.org/api/process.html#event-unhandledrejection).

I've still not yet mentioned other runtimes, where it's again different. Just see [deno](https://deno.com/blog/v1.24#unhandledrejection-event) or [bun](https://bun.com/blog/bun-v1.1.8#new-support-for-process-on-uncaughtexception-and-process-on-unhandledrejection), each creating something on their own (whereas it seems that deno's implementation seems best compatible with what's found in the HTML spec).

Regardless of the other runtimes, in my opinion, the documentation should (if not stated otherwise) follow what is possible in node - whereas the other options might be good to mention as well, but I've left it out as I don't have any experience with those.

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/guide/features.md
+++ b/docs/guide/features.md
@@ -295,24 +295,14 @@ If you need to test that error was not caught, you can create a test that looks 
 
 ```ts
 test('my function throws uncaught error', async ({ onTestFinished }) => {
+  const unhandledRejectionListener = vi.fn()
+  process.on('unhandledRejection', unhandledRejectionListener)
   onTestFinished(() => {
-    // if the event was never called during the test,
-    // make sure it's removed before the next test starts
-    process.removeAllListeners('unhandledRejection')
+    process.off('unhandledRejection', unhandledRejectionListener)
   })
 
-  return new Promise((resolve, reject) => {
-    process.once('unhandledRejection', (error) => {
-      try {
-        expect(error.message).toBe('my error')
-        resolve()
-      }
-      catch (error) {
-        reject(error)
-      }
-    })
+  callMyFunctionThatRejectsError()
 
-    callMyFunctionThatRejectsError()
-  })
+  await expect.poll(unhandledRejectionListener).toHaveBeenCalled()
 })
 ```


SimonSimCity:
> What about browser mode?

Can you please further elaborate what you mean by "browser mode"?

The MDN docs have a compatibility list, where all the major browsers are mentioned, where I cannot find any mentioning of something being special, with the exception of Firefox: https://developer.mozilla.org/en-US/docs/Web/API/Window/unhandledrejection_event#browser_compatibility

AriPerkkio:
> Can you please further elaborate what you mean by "browser mode"?

https://vitest.dev/guide/browser/#browser-mode

docs/guide/features.md
```diff
@@ -298,11 +298,11 @@ test('my function throws uncaught error', async ({ onTestFinished }) => {
   onTestFinished(() => {
     // if the event was never called during the test,
     // make sure it's removed before the next test starts
-    process.removeAllListeners('unhandledrejection')
+    process.removeAllListeners('unhandledRejection')
```
sheremet-va: Actually, this is a bad idea because it will also remove the vitest handler 🙈

We should probably recommend using off and passing down the function

Or just introduce `expect.unhandledRejection()` (which I think we agreed to do 🤔)

AriPerkkio: Ohh this is bad example 😄 
Using `proces.off()` would indeed be better

SimonSimCity:
@sheremet-va, I updated the example to use `process.off()`, and while I was on it, I also updated the code to use `vi.waitFor()` instead of returning a promise. Hope this makes the code also a bit clearer.

sheremet-va:
> I also updated the code to use `vi.waitFor()` instead of returning a promise. Hope this makes the code also a bit clearer.

I think in this case it's better to use `expect.poll`:

```ts
await expect.poll(spy).toHaveBeenCalled()
```

SimonSimCity:
@sheremet-va, thanks for the hint - didn't know of that function until now.

Issue #8630, opened 2025-09-26
"Test attachment requires body or path to be set. Both are missing." when specifying an empty string for `body`


Codesleuth:
### Describe the bug

Specifying `body: ""` when using annotations reports an error. However, this is a valid case (should annotate an empty string).
This appears to be caused by the simple `if (!attachment.body ...)` line here:

https://github.com/vitest-dev/vitest/blob/acbe0e973ba95a0eb1b337fe5650f945f4d51bba/packages/runner/src/context.ts#L171

This would be better replaced with:

```ts
      if (attachment.body == null && !attachment.path) {
        throw new TypeError(`Test attachment requires body or path to be set. Both are missing.`)
      }
```

### Reproduction

```js
await annotate("Anything", { body: "" })
```

### System Info

```shell
System:
    OS: macOS 15.6
    CPU: (10) arm64 Apple M1 Pro
    Memory: 199.09 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.19.0 - ~/.asdf/installs/nodejs/22.19.0/bin/node
    Yarn: 1.22.22 - ~/.asdf/installs/nodejs/22.19.0/bin/yarn
    npm: 10.9.3 - ~/.asdf/installs/nodejs/22.19.0/bin/npm
  Browsers:
    Chrome: 140.0.7339.213
    Edge: 140.0.3485.81
    Safari: 18.6
```

### Used Package Manager

yarn

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vitest.dev/guide/).
- [x] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #8640, opened 2025-09-29
chore: compare `attachment.body` for `null` & `undefined`

iiio2:
I think it resolves https://github.com/vitest-dev/vitest/issues/8630

### DIFF
--- a/packages/runner/src/context.ts
+++ b/packages/runner/src/context.ts
@@ -168,8 +168,8 @@ export function createTestContext(
       type: type || 'notice',
     }
     if (attachment) {
-      if (!attachment.body && !attachment.path) {
-        throw new TypeError(`Test attachment requires body or path to be set. Both are missing.`)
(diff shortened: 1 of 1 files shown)


packages/runner/src/context.ts
```diff
@@ -168,8 +168,8 @@ export function createTestContext(
       type: type || 'notice',
     }
     if (attachment) {
-      if (!attachment.body && !attachment.path) {
-        throw new TypeError(`Test attachment requires body or path to be set. Both are missing.`)
+      if ((attachment.body === null || attachment.body === undefined) && !attachment.path) {
```
sheremet-va:
```suggestion
      if (attachment.body == null && !attachment.path) {
```

sheremet-va (approved):

iiio2:
Thanks @sheremet-va . :blush:

PR #8647, opened 2025-09-30
fix(snapshot): consider comment out between Inline snapshot matcher invoker method and parentheses

### LINKED ISSUE #8632: Adding a comment between `toMatchInlineSnapshot` and `()` breaks snapshot updates
### Describe the bug

**Description:**

This is a bit of an edge case, but I thought I'd report it anyway:

It seems like Vitest is unable to write a snapshot to the file if a comment separates `toMatchInlineSnapshot` and `()`.

This had me confused for a minute, because the test runner does not report the failure unless the snapshot is incorrect.

### Reproduction

**Sandbox:** [StackBlitz](https://stackblitz.com/edit/vitest-inline-snapshot-comment-repro?file=test%2Frepro.test.ts&initialPath=__vitest__/)

**Example:**

```typescript
import * as vi from 'vitest';

// Repro: type 'u' in the terminal to see the snapshots fail to update

vi.test('snapshot w/ comment between assertion and target node does not write', () => {
  vi.expect(
    0
  ).toMatchInlineSnapshot 
     // This comment breaks snapshot updates
    ();
});
```

### System Info

```shell
System:
  OS: macOS 14.6.1
  CPU: (10) arm64 Apple M1 Pro
  Memory: 127.13 MB / 16.00 GB
  Shell: 3.3.1 - /usr/local/bin/fish
Binaries:
  Node: 24.4.1 - ~/.local/state/fnm_multishells/46602_1758709423661/bin/node
  npm: 11.4.2 - ~/.local/state/fnm_multishells/46602_1758709423661/bin/npm
  pnpm: 10.15.1 - ~/Library/pnpm/pnpm
Browsers:
  Chrome: 140.0.7339.214
  Safari: 17.6
npmPackages:
  @vitest/coverage-v8: catalog: => 3.2.4
  @vitest/ui: catalog: => 3.2.4
  vitest: catalog: => 3.2.4
```

### Used Package Manager

pnpm

### Validations

Shinyaigeek:
### Description

Fixes: https://github.com/vitest-dev/vitest/issues/8632

In the current implementation, the InlineSnapshot Matcher is built in a fairly primitive, string-based way, and it seems there was no consideration for commenting out code. Therefore I updated the regex to account for comments, and this issue has been resolved.

### Tests

### Documentation

### Changesets


test/unit/test/inline-snap.test.ts
```diff
@@ -101,6 +101,23 @@ ${indent}}\`)
     `)
   })
 
+  it('replaceInlineSnap with comment between method and parentheses', async () => {
+    const code = `
+  expect(0).toMatchInlineSnapshot
+    // This comment breaks snapshot updates
```
sheremet-va, line 107: Does this fail with `/* This comment */`? Or with `// comment\b// second comment` I feel like using regexp here is not the safest bet. Maybe we should just replace comments with an empty space and the apply the regexp? We do something similar with `filterOutComments` in `@vitest/runner` or `stripLiteral` from `strip-literal`

sheremet-va (changes requested):

test/unit/test/inline-snap.test.ts
Shinyaigeek: Sorry for late 🙇 You're right that the regex approach was fragile. I've reworked it to mask comments before
 matching, as you suggested.

[4b2f0b7](https://github.com/vitest-dev/vitest/pull/8647/commits/4b2f0b7aeb081308507750d86293983d2192c667)

sheremet-va (changes requested):
Can we have e2e test here? I am not very confident with this change when there is only a unit test of a single function

Shinyaigeek:
@sheremet-va Thank you! I added e2e test scenarios to verify this behavior!

PR #8699, opened 2025-10-12
fix(browser): use client-side timer for browser mode duration

userquin:
### Description

This PR measures the wall-clock time from the UI's perspective (from test run trigger to "finished" event), which is the most accurate measure of the perceived developer experience. It will naturally be higher than the duration reported in the terminal, as it includes the entire browser test orchestration:
- Network latency (UI -> Server -> UI)
- Browser iframe creation and orchestration
- Broadcast channel communication between the UI and the test iframe.

The terminal reporter, in contrast, only aggregates the pure test execution time reported by the browser orchestrator.

/cc @JessicaSachs 

### Tests

### Documentation

### Changesets


userquin:
Looks like we need to initialize the configuration correctly...

packages/ui/client/composables/explorer/tree.ts
```diff
@@ -53,11 +55,34 @@ export class ExplorerTree {
...
+    if (config.value.browser) {
+      // Note: This measures the wall-clock time from the UI's perspective (from test
+      // run trigger to "finished" event), which is the most accurate measure of the
+      // perceived developer experience. It will naturally be higher than the duration
+      // reported in the terminal, as it includes the entire browser test orchestration:
+      // - Network latency (UI -> Server -> UI)
+      // - Browser iframe creation and orchestration
+      // - Broadcast channel communication between the UI and the test iframe.
+      // The terminal reporter, in contrast, only aggregates the pure test execution
+      // time reported by the browser orchestrator.
```
sheremet-va, line 80: I don't see the point of this, we want to show the same duration as the terminal, not the time when you click on run

There are events like `onTestRunStart`/`onTestRunEnd` that should be used to calculate the different

I am also not sure why we need to separate browser duration and UI duration, it shouldn't matter

sheremet-va (changes requested):

packages/ui/client/composables/explorer/tree.ts
userquin: will check those hooks

userquin: will add a new hook (onStarted) and send the executionTime from the websocket reporter, maybe we can send the full data shown in the default reporter, but will require some more work since the websocket just implements the Reporter interface => change the logic at client to display that executionTime instead using actual logic for the total time (new PR):

image

Issue #8650, opened 2025-10-01
4.0.0-beta post `10` partially breaks for Bun runtime


nazarhussain:
### Describe the bug

Latest Beta `4.0.0-beta.13` which have `tinypool` version `2.0.0` supposed to support Bun runtime. Which it does but partially.

The `watch` and `dev` mode works but `run` mode does not run any tests. 

### Reproduction

Here the simple test case. 

```ts
// test/unit/simple.test.ts

import { describe, it, expect } from "vitest";

describe("simple test", () => {
  it("simple case", () => {
    expect(1).toBe(1);
  });
});
```

The configuration file. 

```ts
// vitest.config.ts
import { defineConfig } from "vitest/config";

export default defineConfig({
  test: {
    include: ["test/unit/**/*.test.ts"],
  }
});
```

When we run watch mode it shows 

```
bunx --bun vitest watch

 DEV  v4.0.0-beta.13 /vitest-bun-test

 ✓ test/unit/simple.test.ts (1 test) 2ms
   ✓ simple test (1)
     ✓ simple case 1ms

 Test Files  1 passed (1)
      Tests  1 passed (1)
   Start at  15:41:57
   Duration  118ms (transform 17ms, setup 0ms, collect 23ms, tests 2ms, environment 0ms, prepare 4ms)

 PASS  Waiting for file changes...
       press h to show help, press q to quit
```

But running those tests in run mode didn't execute any tests. 

```
bunx --bun vitest run

 RUN  v4.0.0-beta.13 /vitest-bun-test


 Test Files 0 passed (1)
      Tests 0 passed (0)
   Start at 15:42:33
   Duration 0ms
```

### System Info

```shell
npx envinfo --system --npmPackages '{vitest*,@vitest/*,vite,@vitejs/*,playwright,webdriverio}' --binaries --browsers

  System:
    OS: macOS 26.0.1
    CPU: (11) arm64 Apple M3 Pro
    Memory: 214.39 MB / 18.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.13.0 - ~/.asdf/installs/nodejs/22.13.0/bin/node
    Yarn: 1.22.22 - ~/.asdf/installs/nodejs/22.13.0/bin/yarn
    npm: 10.9.2 - ~/.asdf/plugins/nodejs/shims/npm
    pnpm: 10.11.1 - ~/.bun/bin/pnpm
[...]

Issue #8649, opened 2025-10-01
'Closing rpc while "fetch" was pending' on v4 beta


kleinfreund:
### Describe the bug

While trying to upgrade a project from Node 23.11.1 to 24.9.0, I initially ran into some issues with tests interacting with msw (specifically getting a bunch of `TypeError: RequestInit: Expected signal ("AbortSignal {}") to be an instance of AbortSignal.` which was fixed by https://github.com/vitest-dev/vitest/pull/8390 in v4 beta). However, I then also ran into different issues with the v4 beta that I'm not seeing in v3 and so I thought to report them.

With 4.0.0-beta.13 installed, the tests become flaky: they sometimes pass and sometimes fail. Regardless of whether they pass or fail, they always print error messages including the string `Closing rpc while "fetch" was pending` (from https://github.com/vitest-dev/vitest/blob/88e62c7586940b260a1e53efb21d736dd30bdbd8/packages/vitest/src/runtime/worker.ts#L59).

As far as I can tell, the tests always pass when turning off parallelization with `maxWorkers: 1` and _sometimes_ fail when leaving parallelization on.

With parallelization turned off, the following error message mentioned is printed and tests pass:

<details>
<summary>Error</summary>

```
stderr | triggerError (file:/Users/me/project/node_modules/vue-router/dist/vue-router.mjs:3623:21)
Error: [vitest-worker]: Closing rpc while "fetch" was pending
    at reviveInvokeError (file:///Users/me/project/node_modules/vite/dist/node/module-runner.js:473:14)
    at Object.invoke (file:///Users/me/project/node_modules/vite/dist/node/module-runner.js:489:33)
    at VitestModuleRunner.getModuleInformation (file:///Users/me/project/node_modules/vite/dist/node/module-runner.js:1052:7)
    at request (file:///Users/me/project/node_modules/vite/dist/node/module-runner.js:1069:83)
    at /Users/me/project/node_modules/package/Button.vue:2:31
    at VitestModuleEvaluator.runInlinedModule (file:///Users/me/project/node_modules/vitest/dist/module-evaluator.js:136:4)
    at VitestModuleRunner.directRequest (file:///Users/me/project/node_modules/vite/dist/node/module-runner.js:1112:59)
    at VitestModuleRunner.cachedRequest (file:///Users/me/project/node_modules/vite/dist/node/module-runner.js:1030:73)
    at /Users/me/project/src/views/profile/ProfileListView.vue:69:1
[...]

Issue #8564, opened 2025-09-12
Error: Terminating worker thread


geersch:
### Describe the bug

Since `vitest4-beta.9` we get an error when running test suites in **non-watch** mode. Issue does not occur in watch mode.

The tests run fine, but `tinypool` throws an error when terminating the worker threads.

```
⎯⎯⎯⎯ Unhandled Rejection ⎯⎯⎯⎯⎯
Error: Terminating worker thread
❯ Object.ThreadTermination ../../node_modules/tinypool/dist/index.js:354:27
❯ ../../node_modules/tinypool/dist/index.js:393:73
❯ processTicksAndRejections node:internal/process/task_queues:95:5
```

Since that version in non-watch mode `minWorkders` is automatically set to `0`

https://github.com/vitest-dev/vitest/pull/8454

Using forks, if I revert the change to `minThreads`, it does not throw an error.

```ts
const minThreads
    = poolOptions.minThreads ?? vitest.config.minWorkers ?? Math.min(recommendedCount, maxThreads)
```

### Reproduction

Could reproduce this on one of my repos. Made a branch `chore/vitest4` where I switched to vitest 4 beta 11.
I could look into making this branch more minimal in the weekend, but perhaps it already suffices to diagnose the issue. Had the same issue in other repositories as well.

https://github.com/geersch/retry/tree/chore/vitest4

To reproduce

```
git clone git@github.com:geersch/retry.git
git checkout chore/vitest4
yarn
yarn test
```

Or if you just want to run the tests for the `retry` package. Error only occurs for this one.

```
...
cd packages/retry
yarn test run
```

Output

```
❯ yarn test

Process started

 RUN  v4.0.0-beta.11
 ✓ tests/interceptors/retry-interceptor.spec.ts (4 tests) 35ms
 Test Files  1 passed (1)
      Tests  4 passed (4)
   Start at  07:39:06
   Duration  252ms (transform 45ms, setup 23ms, collect 134ms, tests 35ms, environment 0ms, prepare 3ms)

Process exited (exit code 0), completed in 0s 568ms

Process started

 RUN  v4.0.0-beta.11

 ✓ tests/strategies/fibonacci.backoff-strategy.spec.ts (1 test) 1ms
 ✓ tests/strategies/decorrelated-jitter.backoff-strategy.spec.ts (1 test) 2ms
 ✓ tests/strategies/exponential.backoff-strategy.spec.ts (1 test) 1ms
[...]

Issue #8484, opened 2025-08-26
Regression: vitest 3.2 hangs when using workspaces, bail and single threaded


nicojs:
### Describe the bug

When the following conditions are met, the process lives forever:

- Using the latest vitest (3.2)
- With workspaces
- Programmatic API (node API)
- Bail
- Single threaded
- And a test fails in one of the projects

See reproduction example.

### Reproduction

[reproduction.zip](https://github.com/user-attachments/files/21993938/reproduction.zip)

Run with `node run-vitest.ts`. See that the process hangs forever. Reverting back solves this problem `npm i -D vitest@3.1`.

The `run-vitest.ts` file looks like this:

```ts
import { createVitest } from "vitest/node";

async function runVitest() {
  const vitest = await createVitest("test", {
    pool: "threads",
    poolOptions: {
      threads: {
        maxThreads: 1,
        minThreads: 1,
      },
    },
    bail: 1,
  });
  await vitest.start();
  await vitest.close();
}

runVitest().catch((err) => {
  console.error("Vitest run failed:", err);
  process.exit(1);
});
```

I would expect the process to always exit, since we're calling `vitest.close()` (or `process.exit(1)` in case of an error).

### System Info

```shell
System:
    OS: macOS 15.6.1
    CPU: (14) arm64 Apple M4 Pro
    Memory: 730.77 MB / 48.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.18.0 - ~/.nvm/versions/node/v22.18.0/bin/node
    npm: 11.5.2 - ~/.nvm/versions/node/v22.18.0/bin/npm
  Browsers:
    Chrome: 139.0.7258.154
    Edge: 139.0.3405.111
    Safari: 18.6
  npmPackages:
    playwright: ^1.55.0 => 1.55.0 
    vitest: ^3.2.4 => 3.2.4
```

### Used Package Manager

npm

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vitest.dev/guide/).
- [x] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
[...]

PR #8705, opened 2025-10-13
feat!: rewrite pools without `tinypool`

AriPerkkio:
### Description

Removes Tinypool and adds new custom internal pool for Vitest only. Some previous thoughts I wrote ~2 months ago about this below.

- Fixes https://github.com/vitest-dev/vitest/issues/8484
- Fixes https://github.com/vitest-dev/vitest/issues/8564
- Fixes https://github.com/vitest-dev/vitest/issues/8201 (that one has no reproduction so can't say for sure 🤷)
- Fixes https://github.com/vitest-dev/vitest/issues/8649 (need to recheck)
- Fixes https://github.com/vitest-dev/vitest/issues/8650
- Enables https://github.com/vitest-dev/vitest/issues/8010 if we want it
- Enables https://github.com/vitest-dev/vitest/issues/3119, let's implement it later 🥳 

> Vitest uses [`tinypool`](https://github.com/tinylibs/tinypool) for orchestrating how test files are run. Originally Tinypool was a fork of [`piscina`](https://github.com/piscinajs/piscina) which implemented `node:worker_threads` pooling only. Codebase of Piscina is very complex. It has features Vitest users do not need or cannot even use. When ever we add new features to Tinypool, we need to make sure we don't break those unused features that some other dependent package might be using.
> 
> Tinypool has some flaws that are making it hard to use in Vitest. When user is running Vitest workspaces with multiple pools, we easily end up in a situation where we allocate too many resources at the same time. For example in a 4 core CPU we may end up spinning 4 threads AND 4 processes at the same time. This is twice the amount of resources we intended to use. Tinypool does not support splitting resourceses across `runtime`. Also mixing `isolate: false` with `isolate: true` is very complex from Vitest's point of view - and not supported properly at the moment.
> 
> Originally Tinypool did not support `node:child_process` runtime. That was added for Vitest in https://github.com/tinylibs/tinypool/pull/65. While this feature works fine, there's still some overhead that the original `node:worker_threads` runtime required. I think there's room for some optimization and performance gains. Note that adding this feature to Piscina's codebase was unnecessarily difficult.
> 
> For the past 1-2 years, whenever we've seen Tinypool related weird errors in Vitest I've thought about rewriting Tinypool completely. However Tinypool is nowadays used by large packages like `facebook/docusaurus`, `jest-light-runner` and `@rstest/core`. Rewriting Tinypool just for Vitest's needs might break other's use cases. It would also be better if we could make changes to Vitest's pool implementation quickly, without any major releases.
> 
> We should replace Tinypool usage in Vitest with a custom internal pool implementation that lives in Vitest monorepo.
> 
> ## Requirements
> 
> - Basic functionality of Tinypool that Vitest currently uses
> - `node:worker_threads` and `node:child_process` runtimes
> - :new: Interface for easily adding new runtimes in future, e.g. `browser` or even a `async function` pool
> - :new: Resource limits (`maxWorkers/maxThreads/maxForks`) that are shared across whole pool
> - :new: The `runtime` should be configurable on Task level, not on Pool level
>   - Queue that is not `runtime` (`forks/threads/custom`) specific
>   - A single task should describe what `runtime` it uses
>   - Instead of `new Pool({ runtime: 'worker_threads' })`, we do `pool = new Pool(); pool.run({ runtime: 'worker_threads' })`.
>   - Allows mixing `forks` and `threads` specific tests parallel
> - :new: `isolate` should be configurable on Task level, not on Pool level
>   - Pool/Queue sorts tasks based on `isolate` option, so that workers can be reused between tasks
> - Do not use `minThreads` option at all, no idle workers.
>

TODO:
- [x] Add Browser pool back
- [x] Check TODOs in code
- [x] Documentation
- [x] Migration guide
- [x] Fix tests
- [x] Add new tests 
- [x] Custom pool support - it might work already but needs types and tests
- [x] Verify manually that `vmMemory` works. Worker should be recycled (e.g. `pid` changes) after limit is reached

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/advanced/pool.md
+++ b/docs/advanced/pool.md
@@ -129,6 +129,9 @@ init({
   on: (callback) => {
     // Provide a way to listen CustomPoolRunner's "postMessage" calls
   },
+  off: (callback) => {
+    // Optional, provide a way to remove listeners added by "on" calls
+  },
   teardown: () => {
     // Optional, provide a way to teardown worker, e.g. unsubscribe all the `on` listeners
   },


packages/vitest/src/node/pools/pool.ts
```diff
@@ -0,0 +1,263 @@
...
+  return task.context.files.map(file => file.filepath).join(', ')
+}
+
+function isEqualRuntime(runtime: Runtime, task: Task) {
+  if (task.isolate) {
+    throw new Error('Isolated tasks should not share runtimes')
+  }
+
+  // TODO: Compare task.context.environment.name, add runtime.options.env
+  return runtime.name === task.runtime && runtime.options.project.name === task.project.name
```
sheremet-va: Could you explain why you also check the project name? I would imagine that it doesn't matter what project is running as long as the runtime is the same. for example, when there are two projects that have different environment, but the same pool. I would even image that most projects will share the same pool

AriPerkkio: I was probably thinking how projects can transform same files differently due to custom plugins. But modules are reset between test files, right? Looking at `main`, non-isolated pools are not separated by projects there either. 

So we should be able to remove the project name check from here.

packages/vitest/src/runtime/workers/init-threads.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  init({
+    send: response => parentPort!.postMessage(response),
+    subscribe: callback => parentPort!.on('message', callback),
+    off: callback => parentPort!.off('message', callback),
+
+    worker: {
+      post: v => parentPort!.postMessage(v),
+      on: (fn) => {
+        parentPort!.addListener('message', fn)
```
AriPerkkio: For https://github.com/vitest-dev/vitest/issues/8650#issuecomment-3398666165

```suggestion
        parentPort!.addEventListener('message', fn)
```

packages/vitest/src/node/pools/pool.ts
```diff
@@ -0,0 +1,263 @@
...
+
+      async function cancelTask() {
+        await runtime.stop()
+        resolver.reject(new Error('Cancelled'))
+      }
+
+      const onFinished = (message: WorkerResponse) => {
+        if (message?.__vitest_worker_response__ && message.type === 'testfileFinished') {
+          if (task.memoryLimit && message.usedMemory) {
+            isMemoryLimitReached = message.usedMemory >= task.memoryLimit
```
AriPerkkio, line 93: Tested memory limit manually, seems to work like before:

packages/vitest/src/node/pools/runtimes/forks.ts
```diff
@@ -0,0 +1,128 @@
...
+  async stop(): Promise<void> {
+    if (this.stopPromise) {
+      return this.stopPromise
+    }
+
+    this.stopPromise = (async () => {
+      const waitForExit = new Promise<void>(resolve => this.fork?.once('exit', resolve))
+      await super.stop()
+
+      // why does it need a timeout? what is it not dying in time? how to debug it?
```
AriPerkkio: If user's code does not terminate on `SIGTERM`, we'll send `SIGKILL`. We can't really control what code runs there. This is similar to what `jest-worker` and `tinypool` do.


```suggestion
```

sheremet-va: can we instead of removing the comments from the source code - add explanation there? I don't want to scan through PRs commits to see this

AriPerkkio: Sure. At the moment I haven't added any references to other projects here, but we can do it for the weirdly looking parts.

sheremet-va: Me leaving a "what is this" comment is a good place to leave an explanation

packages/vitest/src/node/pools/pool.ts
```diff
@@ -0,0 +1,278 @@
...
+  async close(): Promise<void> {
+    await this.cancel()
+  }
+
+  private getRuntime(task: PoolTask, method: 'run' | 'collect'): PoolRuntime {
+    if (task.isolate === false) {
+      const index = this.sharedRuntimes.findIndex(runtime => isEqualRuntime(runtime, task))
+
+      if (index !== -1) {
+        return this.sharedRuntimes.splice(index, 1)[0]
```
sheremet-va: why not just `this.sharedRuntimes[index]`?

AriPerkkio: When we pick shared runtime, we need to remove it from the `sharedRuntimes` list so that other workers don't pick it too. `splice` is perfect for this, no?

sheremet-va: Ah, makes sense then

packages/vitest/src/runtime/worker.ts
```diff
@@ -100,13 +68,21 @@ export async function execute(method: 'run' | 'collect', ctx: ContextRPC, worker
     await Promise.all(cleanups.map(fn => fn()))
 
     await rpcDone().catch(() => {})
-    environmentLoader?.close()
+    await environmentLoader?.close()
+    rpc.$close()
```
AriPerkkio, line 72: We should pass `off` method from Worker inits, so that `birpc` removes its listener here automatically.

- https://github.com/antfu-collective/birpc/blob/559f621162137194b9c0d4456b90a7f15ea7a135/src/index.ts#L299

packages/vitest/src/node/pools/pool.ts
```diff
@@ -0,0 +1,282 @@
...
+}
+
+function isEqualRunner(runner: PoolRunner, task: PoolTask) {
+  if (task.isolate) {
+    throw new Error('Isolated tasks should not share runners')
+  }
+
+  return (
+    runner.worker.name === task.worker
+    && runner.project === task.project
```
AriPerkkio, line 282: This project check was unnecessary, right?


```suggestion
```

sheremet-va: Let's keep it for now, previously it was also grouped by project. There are some things like `define` that define global variables and they will be leaking. I think we should iterate on this separately.

PR #8713, opened 2025-10-14
fix: change default generics type parameter of `CustomMatchers`

### LINKED ISSUE #7731: Suggesting `R = unknown` for custom matchers makes asymmetric matchers error
### Describe the bug

Here's how Vitest docs [recommend](https://vitest.dev/guide/extending-matchers.html) declaring types for custom matchers.

```ts
interface CustomMatchers<R = unknown> {
  toBeFoo: () => R
}
```

While this works with symmetric matchers (`expect(a).toBeFoo()`), when you apply this to an asymmetric one, the `unknown` return type of the matcher will result in a type error.

```ts
interface O {
  s: string
}

expect({ s: 'foo' }).toEqual<O>({ s: expect.toBeFoo() })
```

```
Type 'unknown' is not assignable to type '{ s string; }'.ts(2322)
```

## Suggested solutions

1. Consider defaulting to `any` instead of `unknown`. The two are not the same and are not interchangeable. Here, it looks like `any` may be a better fit.

In fact, I think it should've been `any` from the start. The matcher itself cannot infer or guarantee a type of the value (that'd be out of its scope). The best it can do is be `anything`. 

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-2narajwa?file=vite.config.ts,test%2Fbasic.test.ts&initialPath=__vitest__/

Go to `test/basic.test.ts` and see the type error when asserting on the `s` property. It must not be there. 

### System Info

```shell
Irrelevant.
```

### Used Package Manager

npm

### Validations

nozomuikuta:
### Description

Resolves #7731

I just changed default generics type parameter, from `unknown` to `any`, as described in the issue.

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/api/expect.md
+++ b/docs/api/expect.md
@@ -1773,7 +1773,7 @@ This function is compatible with Jest's `expect.extend`, so any library that use
 If you are using TypeScript, since Vitest 0.31.0 you can extend default `Assertion` interface in an ambient declaration file (e.g: `vitest.d.ts`) with the code below:
 
 ```ts
-interface CustomMatchers<R = unknown> {
+interface CustomMatchers<R = any> {
   toBeFoo: () => R
 }
 
--- a/docs/blog/vitest-3-2.md
+++ b/docs/blog/vitest-3-2.md
@@ -242,7 +242,7 @@ For example, to have a type-safe `toBeFoo` matcher, you can write something like
 ```ts twoslash
 import { expect } from 'vitest'
 
-interface CustomMatchers<R = unknown> {
+interface CustomMatchers<R = any> {
   toBeFoo: (arg: string) => R
 }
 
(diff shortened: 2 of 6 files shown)


nozomuikuta:
Hmm, I'm not sure why some CI fails...

sheremet-va:
I think we should instead recommend using `Matchers` interface, and deprecate this one. `Matchers` interface is applied to `expect.extend` and `expect.to*` too.

nozomuikuta:
If I'm correct, the latest documentation and the codebase seems to have already resolved #7731 (was it fixed during Vitest 4 development?).

@sheremet-va 

We could close this PR and #7731, couldn't we?

sheremet-va:
> If I'm correct, the latest documentation and the codebase seems to have already resolved #7731 (was it fixed during Vitest 4 development?).
> 
> @sheremet-va
> 
> We could close this PR and #7731, couldn't we?

We do have a new way to define matchers now via `Matchers` interface, and `CustomMatchers` is now a deprecated way to do it. Let's just close these two for now then,

Issue #7834, opened 2025-04-14
Run failed tests in a clean environment on retry


gxxcastillo:
### Clear and concise description of the problem

When using [retry](https://vitest.dev/config/#retry) on a failing test, tests are re-run within the same runtime environment. However, many intermittent or flaky tests (especially those involving side effects, DOM manipulation, async operations, or mocked dependencies) require a full environment reset to succeed. Re-running in the same context doesn’t help when the failure stems from shared or mutated state, event listeners, or global mocks that need cleanup.

### Suggested solution

Introduce a new option (e.g., --retry-isolated, --retry-clean, or --reload-on-retry) that allows Vitest to:
	1.	Detect failing tests on the first run.
	2.	For each failed test (or test file), spawn a new isolated Vitest process to re-run only those tests.
	3.	Report the combined result of both runs, clearly indicating which tests passed only on retry.

### Alternative

Without built in support, users must:
	•	Run the test suite with a reporter that outputs failed test names
	•	Parse the output manually
	•	Re-run Vitest, filtering for the tests that failed

However, where this would be most helpful is in CI environments, where these types of failures introduce a significant delay

### Additional context

This is particularly valuable in CI environments and with UI feature / integration tests that rely on mocked endpoints and DOM queries.

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vitest.dev/guide/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

AriPerkkio:
Some thoughts/ideas:

Extend `retry` to support `number | { count: number, isolated?: boolean }` 
- `--retry=2` and `test('...', { retry: 2 }, () => {})` should behave as before
- `--retry.count=2 --retry.isolated` and `test('...', { retry: { count: 2, isolated: true } }, () => {})` -> retry twice, recycle worker between retries.


When `retry.isolated: true` test fails inside test runner, do not re-run it. Instead report it as failed + add required info in task. When test file has finished, check whether it should be re-run. Re-run only the test cases that are required. Reporters can do their usual retry reporting logic.

gxxcastillo:
What I like about the 2 separate flags is that you could do this:

```
retry: 2
isolated-retry: 1
```

[...]

PR #8812, opened 2025-10-25
feat: add retryDelay, retryCondition, and retryStrategy options for enhanced test retry control (fix #8482)

### LINKED ISSUE #8482: Feature Request: Enhanced `retry` options

matanshavit:
## Summary

This PR implements enhanced retry options for Vitest tests, addressing feature request #8482. The current `retry` mechanism only retries a fixed number of times on any kind of error. This enhancement introduces a nested retry configuration object that provides greater flexibility and control over test retry behavior while maintaining full backward compatibility.

### New Features

The `retry` option now accepts either a **number** (for backward compatibility) or an **object** with the following properties:

1. **`count`** - Number of retry attempts
   - Type: `number`
   - Default: `0`
   - Equivalent to the existing `retry` number option

2. **`delay`** - Delay between retry attempts
   - Type: `number` (milliseconds)
   - Default: `0`
   - Useful for tests that interact with rate-limited APIs or need time to recover between attempts

3. **`condition`** - Conditional retry based on error type
   - Type: `string | ((error: Error) => boolean)`
   - Default: `undefined` (retry on all errors)
   - String: Treated as regex pattern to match against error message (case-insensitive)
   - Function: Called with error object, return `true` to retry
   - ⚠️ **Function form only works in test files** (see important notes below)

4. **`strategy`** - When to retry failed tests
   - Type: `'immediate' | 'test-file' | 'deferred'`
   - Default: `'immediate'`
   - `immediate`: Retry immediately after failure (current behavior, default)
   - `test-file`: Defer retries until after all tests in the current file complete
   - `deferred`: Defer retries until after all test files complete

### Usage Examples

```typescript
// Backward compatibility - number syntax still works
it('simple retry', { retry: 3 }, () => {
  // Retries up to 3 times on any error
})

// New nested syntax - retry with delay (useful for rate-limited APIs)
it('calls API', {
  retry: {
    count: 3,
    delay: 1000
  }
}, async () => {
  await callRateLimitedAPI()
})

// Retry only on specific error types (string regex)
it('handles timeout', {
  retry: {
    count: 5,
    condition: 'TimeoutError'
  }
}, () => {
  // Only retries if error message contains "TimeoutError"
})

// Retry only on specific error types (function)
it('handles timeout', {
  retry: {
    count: 5,
    condition: (error) => error.name === 'TimeoutError'
  }
}, () => {
  // Only retries if error name is exactly "TimeoutError"
})

// Defer retries until end of test file
it('flaky test', {
  retry: {
    count: 3,
    strategy: 'test-file'
  }
}, () => {
  // Retries happen after all other tests in this file complete
})

// Defer retries until all test files complete
it('global flaky test', {
  retry: {
    count: 3,
    strategy: 'deferred'
  }
}, () => {
  // Retries happen after all test files have run
})

// Combine all options
it('complex retry scenario', {
  retry: {
    count: 5,
    delay: 500,
    condition: 'NetworkError',
    strategy: 'test-file'
  }
}, () => {
  // Only retries on NetworkError, with 500ms delay between attempts,
  // and defers retries until end of file
})

// Options can be set on describe blocks
describe('flaky suite', {
  retry: {
    count: 3,
    condition: 'timeout'
  }
}, () => {
  it('test 1', () => { /* inherits retry options */ })
  it('test 2', () => { /* inherits retry options */ })
})
```

### Configuration

All options can be configured globally in `vitest.config.ts`:

```typescript
export default defineConfig({
  test: {
    // Simple number syntax
    retry: 3,
    
    // Or nested object syntax
    retry: {
      count: 3,
      delay: 1000,
      condition: 'TimeoutError',  // ⚠️ String only in config (see notes)
      strategy: 'test-file'
    }
  }
})
```

#### Function Conditions in Config

⚠️ **The function form of `retry.condition` cannot be used in `vitest.config.ts`** because configurations are serialized when passed to worker threads. If you need conditional retry logic, you have two options:

1. **Use a string pattern in your config** (works everywhere):
   ```typescript
   // vitest.config.ts
   export default defineConfig({
     test: {
       retry: {
         count: 3,
         condition: 'TimeoutError|NetworkError'  // ✅ Works
       }
     }
   })
   ```

2. **Use a function directly in test files** (test-level only):
   ```typescript
   // test file
   it('test', {
     retry: {
       count: 3,
       condition: (error) => error.name === 'TimeoutError'  // ✅ Works
     }
   }, () => {
     // ...
   })
   ```

### Core Changes
- **packages/runner/src/run.ts**: Implemented retry logic with delay, condition checking, and strategy handling
  - Helper functions to normalize retry options (number vs object)
  - `shouldRetryTest()`: Helper function to evaluate retry conditions
  - `collectDeferredRetryTests()`: Collects tests with deferred retry strategies
  - Modified `runTest()`: Handles immediate strategy with delay and condition
  - Modified `runSuite()`: Handles test-file strategy retries
  - Modified `runFiles()`: Handles deferred strategy retries

### Type Definitions
- **packages/runner/src/types/tasks.ts**: Updated `TaskBase` and `TestOptions` to support nested retry structure
- **packages/runner/src/types/runner.ts**: Updated `VitestRunnerConfig` to support nested retry structure
- **packages/vitest/src/node/types/config.ts**: Updated `InlineConfig` with proper documentation and serialization warnings
- **packages/vitest/src/runtime/config.ts**: Updated serialized config types

### Configuration Support
- **packages/runner/src/suite.ts**: Option inheritance handles both number and object retry syntax
- **packages/vitest/src/node/config/serializeConfig.ts**: Config serialization with function detection warning
- **packages/vitest/src/node/cli/cli-config.ts**: CLI support (basic number syntax only)

### Test Coverage
- **test/core/test/retry-condition.test.ts**: Tests for `retry.condition` (string regex and function)
- **test/core/test/retry-delay.test.ts**: Tests for `retry.delay` timing validation
- **test/core/test/retry-strategy.test.ts**: Tests for all three retry strategies
- **test/core/test/retry-nested-syntax.test.ts**: Tests for nested syntax and backward compatibility

## Backward Compatibility

✅ **Fully backward compatible** - The existing `retry: number` syntax continues to work exactly as before:
- `retry: 3` is equivalent to `retry: { count: 3 }`
- All new options default to their current behavior values:
  - `delay` defaults to `0` (no delay)
  - `condition` defaults to `undefined` (retry on all errors)
  - `strategy` defaults to `'immediate'` (retry immediately)

Existing tests with the `retry` option will continue to work exactly as before with no changes required.

## Test Plan

- ✅ Added comprehensive test coverage for all new features
- ✅ All existing tests pass
- ✅ Verified backward compatibility with number syntax
- ✅ Manual testing with various configurations
- ✅ Type safety verified with TypeScript

## Related Issues

Closes #8482

Related (future enhancement): #7834

## Comparison with Other Test Runners

This implementation provides similar functionality to:
- **Jest**: `retryTimes(numRetries, { waitBeforeRetry: 100 })` → Our `retry.delay`
- **WebdriverIO**: `specFileRetriesDelay` → Our `retry.delay`
- **WebdriverIO**: `specFileRetriesDeferred` → Our `retry.strategy: 'deferred'`

Our implementation goes beyond existing solutions by:
- Combining all these features in a single, cohesive API
- Adding conditional retry logic via `retry.condition`
- Providing a nested structure that allows for future enhancements (e.g., issue #7834)
- Maintaining full backward compatibility with existing syntax

### DIFF
--- a/packages/runner/src/run.ts
+++ b/packages/runner/src/run.ts
@@ -593,16 +593,12 @@ function collectDeferredRetryTests(suite: Suite, strategy: 'test-file' | 'deferr
   const tests: Test[] = []
 
   function collectFromTask(task: Task) {
-    if (task.type === 'test') {
+    if (task.type === 'test' && task.result?.state === 'fail') {
       const retry = getRetryCount(task.retry)
       const retryCount = task.result?.retryCount ?? 0
       const testStrategy = getRetryStrategy(task.retry)
 
-      if (
-        testStrategy === strategy
-        && task.result?.state === 'fail'
-        && retryCount < retry
-      ) {
+      if (testStrategy === strategy && retryCount < retry) {
         tests.push(task)
       }
     }


packages/runner/src/types/runner.ts
```diff
@@ -36,6 +36,9 @@ export interface VitestRunnerConfig {
   testTimeout: number
   hookTimeout: number
   retry: number
+  retryDelay?: number
+  retryCondition?: string | ((error: Error) => boolean)
+  retryStrategy?: 'immediate' | 'test-file' | 'deferred'
```
AriPerkkio: Should we instead have everyting inside `retry`? It would allow extending it later, like in https://github.com/vitest-dev/vitest/issues/7834.


```suggestion
  retry: number | {
    count?: number
    delay?: number
    condition?: string | ((error: Error) => boolean)
    strategy?: 'immediate' | 'test-file' | 'deferred'
  }
```

packages/runner/src/run.ts
```diff
@@ -266,6 +266,51 @@ async function callCleanupHooks(runner: VitestRunner, cleanups: unknown[]) {
...
+
+  // No condition means always retry
+  if (!condition) {
+    return true
+  }
+
+  // No errors means test passed, shouldn't get here but handle it
+  if (!errors || errors.length === 0) {
+    return false
+  }
```
AriPerkkio: I think `errors` check should be done before any check that returns `true`.


```suggestion
  // No errors means test passed, shouldn't get here but handle it
  if (!errors || errors.length === 0) {
    return false
  }

  // No condition means always retry
  if (!condition) {
    return true
  }
```

packages/runner/src/run.ts
```diff
@@ -468,6 +538,42 @@ function markTasksAsSkipped(suite: Suite, runner: VitestRunner) {
...
+ */
+function collectDeferredRetryTests(suite: Suite, strategy: 'test-file' | 'deferred'): Test[] {
+  const tests: Test[] = []
+
+  function collectFromTask(task: Task) {
+    if (task.type === 'test') {
+      const test = task as Test
+      const retry = test.retry ?? 0
+      const retryCount = test.result?.retryCount ?? 0
+      const testStrategy = test.retryStrategy || 'immediate'
```
AriPerkkio: Checking `type` narrows the typings too:


```suggestion
    if (task.type === 'test') {
      const retry = task.retry ?? 0
      const retryCount = task.result?.retryCount ?? 0
      const testStrategy = task.retryStrategy || 'immediate'
```

packages/runner/src/run.ts
```diff
@@ -468,6 +538,42 @@ function markTasksAsSkipped(suite: Suite, runner: VitestRunner) {
...
+        testStrategy === strategy
+        && test.result?.state === 'fail'
+        && retryCount < retry
+      ) {
+        tests.push(test)
+      }
+    }
+    else if (task.type === 'suite') {
+      const subSuite = task as Suite
+      for (const child of subSuite.tasks) {
```
AriPerkkio:
```suggestion
    else if (task.type === 'suite') {
      for (const child of task.tasks) {
```

packages/vitest/src/node/types/config.ts
```diff
@@ -788,6 +788,33 @@ export interface InlineConfig {
...
+  retryDelay?: number
+
+  /**
+   * Condition to determine if a test should be retried based on the error.
+   * - If a string, treated as a regular expression to match against error message
+   * - If a function, called with the error object; return true to retry
+   *
+   * @default undefined (retry on all errors)
+   */
+  retryCondition?: string | ((error: Error) => boolean)
```
AriPerkkio: We cannot pass functions from main thread to test runner threads, as everything needs to be serialized. So the `vitest.config.ts` cannot support functions in `retryCondition`.

packages/runner/src/types/runner.ts
matanshavit: Sure, done!

packages/runner/src/run.ts
matanshavit: done

packages/runner/src/run.ts
matanshavit: Ok, thanks!

packages/runner/src/run.ts
matanshavit: Yes, that makes sense

matanshavit:
@AriPerkkio Just checking in, did I address all comments?

test/core/test/retry-strategy.test.ts
```diff
@@ -0,0 +1,55 @@
...
+it('retry with deferred strategy', {
+  retry: {
+    count: 2,
+    strategy: 'deferred',
+  },
+}, () => {
+  deferredCount += 1
+  // This will fail on attempts 1 and 2, pass on attempt 3
+  expect(deferredCount).toBe(3)
+})
```
AriPerkkio, line 42: @matanshavit could you check why this test keeps failing?

> `Error: The snapshot state for '/Users/runner/work/vitest/vitest/test/core/test/retry-strategy.test.ts' is not found. Did you call 'SnapshotClient.setup()'?`

AriPerkkio (changes requested):

test/core/test/retry-strategy.test.ts
matanshavit: @AriPerkkio Thanks for pointing that out, I thought I fixed that already for some reason. I had to group the deferred tests by file. Now it is fixed

packages/runner/src/run.ts
```diff
@@ -615,6 +805,80 @@ export async function runFiles(files: File[], runner: VitestRunner): Promise<voi
...
+      // We need to check children only because the file's own state
+      // was set to 'fail' before deferred retries
+      if (hasFailed(file.tasks)) {
+        file.result.state = 'fail'
+      }
+      else {
+        file.result.state = 'pass'
+      }
+      // Send updated file state to the main process
+      updateTask('suite-finished', file, runner)
```
AriPerkkio, line 879: I think this is causing reporters to show the file twice. Why is this needed?

packages/runner/src/types/runner.ts
```diff
@@ -35,7 +35,12 @@ export interface VitestRunnerConfig {
   maxConcurrency: number
   testTimeout: number
   hookTimeout: number
-  retry: number
+  retry: number | {
+    count?: number
+    delay?: number
+    condition?: string | ((error: Error) => boolean)
+    strategy?: 'immediate' | 'test-file' | 'deferred'
```
sheremet-va, line 42: I am really concerned about `strategy`. Breaking the flow of tests doesn't seem worth to me, at least for a minor release.

This adds a lot of complexity that's hard to maintain. It breaks the order of [hooks](https://vitest.dev/api/advanced/reporters.html) and makes `afterEach` hook unpredictable. It's pretty hard to wrap your hand around how it _supposed_ to work if it's not `immediate`.

So I would remove this option entirely.

packages/runner/src/types/tasks.ts
```diff
@@ -433,12 +465,42 @@ export interface TestOptions {
...
-   * Times to retry the test if fails. Useful for making flaky tests more stable.
-   * When retries is up, the last test error will be thrown.
-   *
+   * Retry configuration for the test.
+   * - If a number, specifies how many times to retry
+   * - If an object, allows fine-grained retry control
    * @default 0
    */
-  retry?: number
+  retry?: number | {
```
sheremet-va, line 473: I think it would be better if we had a separate type for `retry` instead of copy-pasting it twice.

packages/runner/src/run.ts
```diff
@@ -266,6 +314,51 @@ async function callCleanupHooks(runner: VitestRunner, cleanups: unknown[]) {
...
+  if (typeof condition === 'string') {
+    // String condition is treated as regex pattern
+    const regex = new RegExp(condition, 'i')
+    return regex.test(error.message || '')
+  }
+  else if (typeof condition === 'function') {
+    // Function condition is called with error object
+    try {
+      // Convert TestError back to Error-like object for the condition function
+      const errorObj = Object.assign(new Error(error.message), {
```
sheremet-va, line 345: Why do we need to convert to an error-like object?

packages/vitest/src/node/reporters/reported-tasks.ts
```diff
@@ -500,7 +500,12 @@ export interface TaskOptions {
   readonly fails: boolean | undefined
   readonly concurrent: boolean | undefined
   readonly shuffle: boolean | undefined
-  readonly retry: number | undefined
+  readonly retry: number | {
+    count?: number
+    delay?: number
+    condition?: string | ((error: Error) => boolean)
```
sheremet-va, line 506: I don't think this is possible. Please add a test for this to verify how `retry` options are passed down.

packages/vitest/src/runtime/config.ts
```diff
@@ -77,7 +77,12 @@ export interface SerializedConfig {
     truncateThreshold?: number
   } | undefined
   diff: string | SerializedDiffOptions | undefined
-  retry: number
+  retry: number | {
+    count?: number
+    delay?: number
+    condition?: string | ((error: Error) => boolean)
```
sheremet-va, line 83: The function is not possible here

packages/runner/src/types/runner.ts
```diff
@@ -35,7 +35,12 @@ export interface VitestRunnerConfig {
   maxConcurrency: number
   testTimeout: number
   hookTimeout: number
-  retry: number
+  retry: number | {
+    count?: number
+    delay?: number
+    condition?: string | ((error: Error) => boolean)
```
sheremet-va, line 41: It is not possible to have a function here, it can only be defined on a task

sheremet-va (changes requested):

packages/runner/src/types/runner.ts
AriPerkkio: Yeah from reporter's perspective `test-file` and `deferred` introduce changes that could be considered as breaking changes. The same `TestModule` could be called twice in the finish hook.

Also I'm not sure how exactly should reporters print modules that they see twice 🤔 

Maybe it's better to drop `strategy` completely from the PR.

sheremet-va:
Implemented in #9370

PR #8813, opened 2025-10-25
feat(ui): add tabbed failure view for `toMatchScreenshot` with comparison slider

macarie:
### Description

This PR adds a new failure view for `toMatchScreenshot` in the report panel. The view displays diff, reference, and actual screenshots in separate tabs, with a slider tab for side-by-side comparison. Under the hood, this feature is implemented using the new artifacts API.

Preview:

https://github.com/user-attachments/assets/1b09cf6a-0523-46fe-8047-9585121d829a

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/ui/client/components/artifacts/visual-regression/SmallTabs.vue
+++ b/packages/ui/client/components/artifacts/visual-regression/SmallTabs.vue
@@ -6,10 +6,6 @@ import { idFor, SMALL_TABS_CONTEXT } from '~/composables/small-tabs'
 const activeTab = ref<string | null>(null)
 const tabs = ref<SmallTabsConfig[]>([])
 
-function setActive(key: string) {
-  activeTab.value = key
-}
-
 const id = useId()
 
 provide(SMALL_TABS_CONTEXT, {
@@ -21,7 +17,7 @@ provide(SMALL_TABS_CONTEXT, {
     }
 
     if (tabs.value.length === 1) {
-      setActive(tab.id)
+      activeTab.value = tab.id
     }
   },
   unregisterTab: (tab) => {
@@ -29,6 +25,10 @@ provide(SMALL_TABS_CONTEXT, {
 
     if (index > -1) {
       tabs.value.splice(index, 1)
+    }
+
+    if (activeTab.value === tab.id) {
+      activeTab.value = tabs.value[0]?.id ?? null
     }
   },
 })
@@ -52,7 +52,7 @@ provide(SMALL_TABS_CONTEXT, {
         :aria-controls="idFor.tabpanel(tab.id, id)"
         type="button"
         class="aria-[selected=true]:underline underline-offset-4"
-        @click="setActive(tab.id)"
+        @click="activeTab = tab.id"
       >
         {{ tab.title }}
       </button>
--- a/packages/ui/client/components/artifacts/visual-regression/VisualRegression.vue
+++ b/packages/ui/client/components/artifacts/visual-regression/VisualRegression.vue
@@ -29,16 +29,32 @@ const groups = computed(() => ({
       {{ regression.message }}
     </template>
     <SmallTabs>
-      <SmallTabsPane v-if="groups.diff" title="Diff">
+      <SmallTabsPane
+        v-if="groups.diff"
+        :key="groups.diff.path"
+        title="Diff"
+      >
         <VisualRegressionImage :attachment="groups.diff" />
       </SmallTabsPane>
-      <SmallTabsPane v-if="groups.reference" title="Reference">
+      <SmallTabsPane
+        v-if="groups.reference"
+        :key="groups.reference.path"
+        title="Reference"
+      >
         <VisualRegressionImage :attachment="groups.reference" />
       </SmallTabsPane>
-      <SmallTabsPane v-if="groups.actual" title="Actual">
+      <SmallTabsPane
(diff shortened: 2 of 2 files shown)


packages/ui/client/components/VisualRegressionSlider.vue
```diff
@@ -0,0 +1,72 @@
...
+const maxWidth = computed(() =>
+  Math.max(props.reference.metadata!.width as number, props.actual.metadata!.width as number),
+)
+const maxHeight = computed(() =>
+  Math.max(props.reference.metadata!.height as number, props.actual.metadata!.height as number),
+)
+
+const splitPercentage = ref(50)
+
+const inputId = crypto.randomUUID()
```
userquin: Dont use crypto here, use useId from vue

macarie: Oh, thanks!

I looked for that hook but only found the RFC issue with no updates, then I implemented it like this and forgot to check again when I was done 😅

userquin: Maybe we need to update vue at ui package 🤞

userquin: Cannot use props destructuring (too many props. at components)?

macarie: > Maybe we need to update vue at ui package 🤞

Don't think so, luckily `useId` seems to have been introduced in 3.5, the version used here 😄

> Cannot use props destructuring (too many props. at components)?

What do you mean?

userquin: `const { a, b } = defineProps<...>()`

macarie: I didn't know that was allowed now... my Vue knowledge is definitely not the most up-to-date. Thanks, _again_! 😄

sheremet-va (commented):
I like the direction of this, but I wonder if instead of adding a separate artifact entity, we can just extend attachments with additional fields, but have them as `@private`? The two share a lot of similarities and it's hard to justify the separation, to be honest. 

I feel like in the future we could also make the API public, the metadata seems useful

macarie:
> I wonder if instead of adding a separate artifact entity, we can just extend attachments with additional fields, but have them as `@private`? The two share a lot of similarities and it's hard to justify the separation, to be honest.

I do agree, @sheremet-va. I initially went for a unified API, but we need one "event" to store multiple attachments, and I'm not sure how this can be done while maintaining backwards compatibility.

Current type for annotations:

```ts
export interface TestAttachment {
  contentType?: string
  path?: string
  body?: string | Uint8Array
}

export interface TestAnnotation {
  message: string
  type: string
  location?: TestAnnotationLocation
  attachment?: TestAttachment
}
```

Proposed artifact types:

```ts
export interface TestArtifactAttachment extends TestAttachment {
  name?: string
  metadata?: Record<string, unknown>
}

export interface TestArtifact {
  title: string                          // new and required
  message?: string                       // existing, but now optional
  source: string                         // new and required
  type: string                           // different usage (?)
  attachments: TestArtifactAttachment[]  // new and required, `attachment` dropped
  location?: TestAnnotationLocation      // same as before
  metadata?: Record<string, unknown>     // new but optional
}
```

The `TestArtifactAttachment` extensions are backwards-compatible as all new fields are optional.

In `TestArtifact` we have these differences:
- **`title`** (new, required): displays in the section header (e.g., "Visual Regression" in the video)
- **`message`**: now optional instead of required
- **`source`** (new, required): identifies the artifact's "origin" (assertion name, selector, etc) so reporters know what to display
- **`type`**: same field but different semantics I believe. A `source` can produce multiple artifact types, this allows reporters to _specialize_ rendering. Example: `toMatchScreenshot` with `--update` _could_ attach `{ source: 'toMatchScreenshot', type: 'reference-update' }`
- **`attachments`** (new, required): enables multiple files per artifact (can be empty). Replaces the optional singular `attachment`

I believe these changes are needed to make the feature more versatile and support new use cases like snapshot diffs, listing multiple matched elements, etc. I'm open to merging them if we can find a way that maintains backwards compatibility while supporting these requirements (if they are necessary, of course).

sheremet-va:
Can't all of the new required fields just go into an `artifact` field? The `message` can be set to `''` until the next major. And `attachments` can be added to `TestAnnotation` alongside the `attachment` as an `@experimental` feature

packages/ui/vitest.config.ts
```diff
@@ -6,6 +6,9 @@ import viteConfig from './vite.config'
 export default mergeConfig(
   viteConfig,
   defineConfig({
+    optimizeDeps: {
+      include: ['vue-router', 'splitpanes', 'd3-graph-controller', 'vue-virtual-scroller'],
+    },
```
macarie, line 11: Not entirely sure about this, but without it Vite would trigger a reload and hang the test until the pipeline reaches the 30m timeout: https://github.com/vitest-dev/vitest/actions/runs/19538883275/job/55939684333?pr=8813

macarie:
Should be fixed now @sheremet-va. Noticed Vue reused the node when switching test results so "Reference" moved to the left, that should be fixed now too.

sheremet-va (approved):

sheremet-va:
Thank you!

PR #8836, opened 2025-10-27
docs: Fix mock call assertion in documentation

danielrentz:
### Description

Do not compare array literals

### Tests

### Documentation

### Changesets


docs/api/mock.md
```diff
@@ -7,7 +7,7 @@ import { vi } from 'vitest'
 
 const fn = vi.fn()
 fn('hello world')
-fn.mock.calls[0] === ['hello world']
+fn.mock.calls[0][0] === 'hello world'
```
sheremet-va, line 10: It's important to _show_ the array. If this is a confusing representation, I propose


```suggestion
fn.mock.calls[0] // == ['hello world']
```

sheremet-va (changes requested):

docs/api/mock.md
danielrentz: @sheremet-va Yeah this is done throughout the documentation (as I have seen after creating the PR), especially on the Mocks page.
I think it is better to use your proposed style with a comment to not advertise to use this in real tests. I will extend the PR to cover all other places when I am back from vacation :)

danielrentz: @sheremet-va I think the best way would be to change these examples to use real expect assertions:
```ts
epect(fn.mock.calls[0]).toEqual(['hello world'])
```
as done in most other examples in the documentation. If you agree, I would adapt similar examples in the docs as well.

sheremet-va: we don't want to introduce different concepts in this part of the guide, like assertions

danielrentz: Ok, then I would say it's best to leave it as it is.

PR #8842, opened 2025-10-27
feat: added chai style assertions

### LINKED ISSUE #8819: Enhance Chai assertions

Following [this discussion](https://github.com/vitest-dev/vitest/discussions/8789#discussioncomment-14771676), I think it would be good to provide Chai-like assertions for all Jest-like assertions.

Especially in the case of spies/mocks, the assertions could be similar to [sinon-chai](https://github.com/chaijs/sinon-chai) or consistent with the existing Jest assertions.

For examples:
- `toHaveBeenCalled()` -> `to.have.been.called` or `to.have.been.called()`
- `toHaveBeenCalledTimes(n)` -> `to.have.callCount(n)` or `to.have.been.calledTimes(n)`
- `toHaveBeenCalledWith(args)` ->  `to.have.been.calledWith(args)`
- ...

### Suggested solution

Assuming the new assertions should be consistent with the existing ones, the changes could be just adding new assertion names in [jest-expect.ts](https://github.com/vitest-dev/vitest/blob/main/packages/expect/src/jest-expect.ts). For example: 
```ts
def(['toHaveBeenCalled', 'toBeCalled', 'called'], function () { ...}
```

However, the file might have to be renamed to something like `jest-chai-expect.ts`, because it's not only about Jest anymore.

### Alternative

If we don't want to mix the the Jest and Chai assertions in the same file, another solution could be moving the assertion functions to a new shared module, then defining the Jest assertions in `jest-expect.ts` and the Chai assertions in `chai-expect.ts`.

### Additional context

I'd be glad to create a PR, once this feature request is accepted and an implementation is recommended.

### Validations

ronnakamoto:
### Description

This PR adds Chai-style assertion names for spy matchers, enabling users migrating from Mocha+Chai+Sinon to use familiar assertion syntax without rewriting their tests.

Resolves #8819 

Users migrating from Mocha/Chai/Sinon to Vitest currently need to rewrite all their spy assertions from Chai-style (e.g., `expect(spy).to.have.been.called`) to Jest-style (e.g., `expect(spy).toHaveBeenCalled()`). This creates friction during migration and requires significant test refactoring.

This implementation provides Chai-style assertion names that delegate to existing Jest-style implementations, allowing both styles to coexist.

Added the following assertions that delegate to existing Jest-style implementations:

**Property Assertions (no parentheses):**
- `called` -> `toHaveBeenCalled`
- `calledOnce` -> `toHaveBeenCalledOnce`
- `calledTwice` -> `toHaveBeenCalledTimes(2)`
- `calledThrice` -> `toHaveBeenCalledTimes(3)`
- `returned` -> `toHaveReturned`

**Method Assertions (with parentheses):**
- `callCount(n)` -> `toHaveBeenCalledTimes(n)`
- `calledWith(...args)` -> `toHaveBeenCalledWith(...args)`
- `calledOnceWith(...args)` -> `toHaveBeenCalledExactlyOnceWith(...args)`
- `lastCalledWith(...args)` -> `toHaveBeenLastCalledWith(...args)`
- `nthCalledWith(n, ...args)` -> `toHaveBeenNthCalledWith(n, ...args)`
- `returnedWith(value)` -> `toHaveReturnedWith(value)`
- `returnedTimes(n)` -> `toHaveReturnedTimes(n)`
- `lastReturnedWith(value)` -> `toHaveLastReturnedWith(value)`
- `nthReturnedWith(n, value)` -> `toHaveNthReturnedWith(n, value)`
- `calledBefore(spy)` -> `toHaveBeenCalledBefore(spy)`
- `calledAfter(spy)` -> `toHaveBeenCalledAfter(spy)`

### Tests
- Run the tests with `pnpm test:ci`.

### Documentation
- Updated documentation

### DIFF
--- a/docs/api/expect.md
+++ b/docs/api/expect.md
@@ -6,7 +6,7 @@ The following types are used in the type signatures below
 type Awaitable<T> = T | PromiseLike<T>
 ```
 
-`expect` is used to create assertions. In this context `assertions` are functions that can be called to assert a statement. Vitest provides `chai` assertions by default and also `Jest` compatible assertions built on top of `chai`. For spy/mock testing, Vitest also provides [Chai-style assertions](#chai-style-spy-assertions) (e.g., `expect(spy).to.have.been.called()`) alongside Jest-style assertions (e.g., `expect(spy).toHaveBeenCalled()`). Unlike `Jest`, Vitest supports a message as the second argument - if the assertion fails, the error message will be equal to it.
+`expect` is used to create assertions. In this context `assertions` are functions that can be called to assert a statement. Vitest provides `chai` assertions by default and also `Jest` compatible assertions built on top of `chai`. Since Vitest 4.1, for spy/mock testing, Vitest also provides [Chai-style assertions](#chai-style-spy-assertions) (e.g., `expect(spy).to.have.been.called()`) alongside Jest-style assertions (e.g., `expect(spy).toHaveBeenCalled()`). Unlike `Jest`, Vitest supports a message as the second argument - if the assertion fails, the error message will be equal to it.
 
 ```ts
 export interface ExpectStatic extends Chai.ExpectStatic, AsymmetricMatchersContaining {


docs/api/expect.md
```diff
@@ -1329,6 +1329,140 @@ test('spy function returns bananas on second call', async () => {
 })
 ```
 
+## Chai-Style Spy Assertions
```
sheremet-va: This is a guide, not an API.

For an API we need all methods to be described similar to other `to*` assertions

The current doc can be moved to guides

docs/guide/migration.md
```diff
@@ -602,3 +602,306 @@ export default defineConfig({
...
+  afterEach(() => { /* teardown */ })
+
+  it('test', () => {
+    // test code
+  })
+})
+```
+
+::: tip
+Vitest supports both `before`/`after` (Mocha-style) and `beforeAll`/`afterAll` (Jest-style) - they're aliases!
```
sheremet-va: Vitest does not support `before`/`after`

docs/guide/migration.md
```diff
@@ -602,3 +602,306 @@ export default defineConfig({
...
+:::
+
+### Assertions
+
+Vitest includes Chai assertions by default, so most Chai assertions work without changes:
```
sheremet-va: Let's remove the "most" part, all of them work the same way

docs/guide/migration.md
```diff
@@ -602,3 +602,306 @@ export default defineConfig({
...
+const spy = vi.spyOn(obj, 'method')
+obj.method('arg1', 'arg2')
+
+expect(spy).to.have.been.called()
+expect(spy).to.have.been.calledOnce()
+expect(spy).to.have.been.calledWith('arg1', 'arg2')
+```
+
+::: tip
+Notice the only difference: Chai-style assertions in Vitest use method calls `called()` instead of properties `called`. This is because Vitest implements these as methods that delegate to Jest-style matchers.
```
sheremet-va: I think we still want them to be defined in the same way that sinon-chai does:

```ts
expect(spy).to.have.been.called
expect(spy).to.have.been.calledOnce
expect(spy).to.have.been.calledWith('arg1', 'arg2')
```

docs/guide/migration.md
```diff
@@ -602,3 +602,306 @@ export default defineConfig({
...
+clock.restore()
+
+// Vitest
+import { vi } from 'vitest'
+vi.useFakeTimers()
+vi.advanceTimersByTime(1000)
+vi.useRealTimers()
+```
+
+### Async Testing
```
sheremet-va: I think this can be removed

docs/guide/migration.md
```diff
@@ -602,3 +602,306 @@ export default defineConfig({
...
+  expect(result).to.equal(42)
+})
+```
+
+### Key Differences
+
+1. **Globals**: Mocha provides globals by default. In Vitest, either import from `vitest` or enable [`globals`](/config/#globals) config
+2. **Assertion style**: You can use both Chai-style (`expect(spy).to.have.been.called()`) and Jest-style (`expect(spy).toHaveBeenCalled()`)
+3. **Method vs Property**: Chai-style assertions in Vitest use methods `called()` instead of properties `called`
+4. **Module mocking**: Vitest has built-in module mocking with `vi.mock()`, unlike Sinon
```
sheremet-va: I don't think module mocking should be mention at all anywhere in the migration guide since it wasn't support before the migration

docs/guide/migration.md
```diff
@@ -602,3 +602,306 @@ export default defineConfig({
...
+5. **Parallel execution**: Vitest runs tests in parallel by default, Mocha runs sequentially
+
+### Migration Checklist
```
sheremet-va: Let's remove that checklist, that's just a repetition for the sake of a bigger text

docs/guide/migration.md
```diff
@@ -602,3 +602,306 @@ export default defineConfig({
...
+- [ ] Import test functions from `vitest`: `import { describe, it, expect, vi } from 'vitest'`
+- [ ] Replace `sinon.spy()` with `vi.fn()` or `vi.spyOn()`
+- [ ] Replace `sinon.stub()` with `vi.spyOn().mockImplementation()`
+- [ ] Update Chai-style spy assertions to use methods: `called` → `called()`
+- [ ] Replace `sinon.useFakeTimers()` with `vi.useFakeTimers()`
+- [ ] Convert callback-style async tests to `async`/`await`
+- [ ] Update module mocking to use `vi.mock()`
+- [ ] Enable [`globals`](/config/#globals) config if you prefer not to import test functions
+
+### Example Migration
```
sheremet-va: The same thing, all of this is already described with examples in the same doc. It can be removed

packages/expect/src/chai-style-assertions.ts
```diff
@@ -0,0 +1,138 @@
...
+ * expect(spy).to.have.been.called
+ * expect(spy).to.have.been.calledWith('arg1', 'arg2')
+ * expect(spy).to.have.been.calledOnce
+ *
+ * // These delegate to Jest-style implementations:
+ * // toHaveBeenCalled, toHaveBeenCalledWith, toHaveBeenCalledOnce
+ */
+export const ChaiStyleAssertions: ChaiPlugin = (chai, utils) => {
+  function def(
+    name: keyof Assertion | (keyof Assertion)[],
```
sheremet-va: `name` is never passed down as an array here

packages/expect/src/chai-style-assertions.ts
```diff
@@ -0,0 +1,138 @@
...
+      name.forEach(n => addMethod(n))
+    }
+    else {
+      addMethod(name)
+    }
+  }
+
+  // Chai-style assertion: called
+  // Delegates to: toHaveBeenCalled
+  def('called', 'toHaveBeenCalled')
```
sheremet-va: this should be a property

packages/expect/src/chai-style-assertions.ts
```diff
@@ -0,0 +1,138 @@
...
+
+    if (Array.isArray(name)) {
+      name.forEach(n => addMethod(n))
+    }
+    else {
+      addMethod(name)
+    }
+  }
+
+  // Chai-style assertion: called
```
sheremet-va: All of the comments can be removed, they do not provide any value

packages/expect/src/chai-style-assertions.ts
```diff
@@ -0,0 +1,138 @@
...
+  // Delegates to: toHaveBeenCalledBefore
+  def('calledBefore', 'toHaveBeenCalledBefore')
+
+  // Chai-style assertion: calledAfter(spy)
+  // Delegates to: toHaveBeenCalledAfter
+  def('calledAfter', 'toHaveBeenCalledAfter')
+
+  // Chai-style assertion: calledTwice
+  // Wrapper that calls toHaveBeenCalledTimes(2)
+  utils.addMethod(
```
sheremet-va: Why can't we use `def` for this? Maybe enhance it to support a property with specified arguments

packages/expect/src/types.ts
```diff
@@ -793,6 +794,156 @@ export interface Assertion<T = any>
...
+   * expect(spy1).to.have.been.calledAfter(spy2)
+   */
+  calledAfter: (mock: MockInstance, failIfNoFirstInvocation?: boolean) => void
+
+  /**
+   * Checks that a spy was called exactly twice.
+   * Chai-style equivalent of `toHaveBeenCalledTimes(2)`.
+   *
+   * @example
+   * expect(spy).to.have.been.calledTwice
```
sheremet-va, line 935: All the examples show it as a property...

packages/expect/src/types.ts
```diff
@@ -793,6 +794,156 @@ export interface Assertion<T = any>
   rejects: PromisifyAssertion<T>
 }
 
+/**
+ * Chai-style assertions for spy/mock testing.
+ * These provide sinon-chai compatible assertion names that delegate to Jest-style implementations.
+ */
+export interface ChaiStyleAssertion {
```
sheremet-va: Let's name it something like `ChaiMockAssertion` or something

docs/guide/migration.md
ronnakamoto: Refactored to that format

docs/guide/migration.md
ronnakamoto: done

PR #8846, opened 2025-10-28
chore: bump node version to lts

iiio2:
actually node v24 is now lts. I've tried to change something. will it be ok ?


iiio2:
Thanks @sheremet-va .

Issue #8854, opened 2025-10-28
v4 coverage include/exclude logic breaks advanced use cases


clydin:
### Describe the bug

The refactor introduced in v4 to move away from `test-exclude`, while a significant improvement to on-disk package size, also introduced a change in behavior by performing direct filesystem access within the replacement.  The previous v3 `test-exclude` behavior operated solely on the path strings and matched them to include/exclude patterns.  The introduction of the direct file system access is problematic for several use cases such as:
* Vite plugins that provided load hooks for virtual files that may need to be processed for coverage.  (virtual files may contain relevant code from actual application files.  Source maps can map back to the files)
* Reporting coverage of node package/libraries that contain source maps. (first-party libraries being common but analysis of third-party package usage is also a possibility)

Additionally, there is the potential for performance degradation due to a synchronous file system operation being performed in a loop over every file included in the coverage.  For large projects this may be problematic.

Relevant `existsSync` call: https://github.com/vitest-dev/vitest/blob/9b75ec5f2fa46daaaa649493eb7f643a1cc9f51a/packages/vitest/src/node/coverage.ts#L175

### Reproduction

For the first above use case:
1) Create a new Angular application for next version: `npx @angular/cli@next new vitest-coverage --defaults`
2) Change into directory and install coverage: `npm i @vitest/coverage-v8`
3) Run tests via `ng test --coverage --no-watch` and note no coverage report information.
4) Force install vitest 3: `npm i --force vitest@3 @vitest/coverage-v8@3`
5) Re-run tests and note coverage report with files.
6) Reinstall vitest 4: `npm i vitest@4 @vitest/coverage-v8@4`
7) Manually edit the vitest package to remove the `existsSync` call here: `node_modules/vitest/dist/chunks/coverage.CSfvENE1.js` (line 2784)
8) Re-run tests and note coverage report with files.

### System Info

```shell
npmPackages:
  @vitest/coverage-v8: ^4.0.4 => 4.0.4 
  vitest: ^4.0.4 => 4.0.4
```

### Used Package Manager

npm

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vitest.dev/guide/).
[...]

PR #8860, opened 2025-10-29
fix(coverage): prevent filtering out virtual files before remapping to sources

AriPerkkio:
### Description

- Fixes https://github.com/vitest-dev/vitest/issues/8854

This change feels a bit wrong. Angular should really use the original filenames and do transforms on plugin. There is really no reason to use virtual files. 

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/coverage-istanbul/src/provider.ts
+++ b/packages/coverage-istanbul/src/provider.ts
@@ -2,7 +2,7 @@ import type { CoverageMap } from 'istanbul-lib-coverage'
 import type { Instrumenter } from 'istanbul-lib-instrument'
 import type { ProxifiedModule } from 'magicast'
 import type { CoverageProvider, ReportContext, ResolvedCoverageOptions, Vite, Vitest } from 'vitest/node'
-import { promises as fs } from 'node:fs'
+import { existsSync, promises as fs } from 'node:fs'
 // @ts-expect-error missing types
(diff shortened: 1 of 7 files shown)


AriPerkkio:
@clydin could you verify if the preview release works fine on Angular? I wasn't able to reproduce the `existSync` removal fix with your reproduction steps. The virtual module in this case, it's not one of those `ng://` files?

clydin:
Using the linked preview builds fixes the issue.

Thanks for the quick fix.

Also, a note on Angular's behavior in this case.  It is doing something similar to Vite's planned "full bundle mode".  Ideally, once this feature is available within Vite, Angular can leverage it directly and remove its own implementation.

packages/coverage-istanbul/src/provider.ts
```diff
@@ -59,7 +59,7 @@ export class IstanbulCoverageProvider extends BaseCoverageProvider<ResolvedCover
       return
     }
 
-    if (!this.isIncluded(removeQueryParameters(id))) {
+    if (!this.isIncluded(removeQueryParameters(id.replace('file://', '')))) {
```
sheremet-va: Can we use the normalization function for this instead? How does this work with windows paths?

AriPerkkio: This was actually unnecessary. Fixed the custom plugin and removed this change.

sheremet-va (approved):

PR #8879, opened 2025-10-30
feat: setup stackblitz codeflow

iiio2:
Actually I'm not sure about it. I've seen other projects have stackblitz codeflow, e.g, vite (https://github.com/vitejs/vite). That is why I have thought maybe we can use it here or not.


iiio2:
Ok @AriPerkkio . Thanks.

PR #8906, opened 2025-11-02
feat(expect): add Set support to toBeOneOf

### LINKED ISSUE #8902: Support sets as argument to expect.toBeOneOf

In Vitest tests you can do this:
```js
expect('b').toBeOneOf(['a','b','c','d']);
```
but not this:
```js
const values = new Set(['a','b','c','d']);
expect('b').toBeOneOf(values);
```
You get this error: `TypeError: You must provide an array to expect(received).toBeOneOf(expected), not 'object'.`

### Suggested solution

I think `toBeOneOf` is a classic Set operation and I think Vitest should support it. I think the implementation should be very straightforward and I am happy to provide a PR for it.

### Alternative

There are simple workarounds of course, as you can just do this:
```js
expect('b').toBeOneOf([...values]);
```
(I had to do this at work recently)
But I don't see a reason why Vitest should not just support this out of the box.

### Additional context

_No response_

### Validations

tim-we:
### Description

This PR adds Set support to the `.toBeOneOf` method.

```ts
const allowedFruits = new Set(['apple', 'banana', 'orange'])
expect(fruit).toBeOneOf(allowedFruits)
```

Resolves #8902 

I find it quite strange that the test passes if the given array is empty. For consistency the Set implementation behaves the same. Is this really intentional? If so I would adjust the documentation. If not I can change it in a follow up PR.

Also I did not use the vitest internal equals method so something like 
```ts
expect('apple').toBeOneOf(new Set([expect.any(String)]))
```
would fail. I don't think `toBeOneOf` will ever be used like that, therefore I just used the native `Set.has` method. Does that work for you?

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/api/expect.md
+++ b/docs/api/expect.md
@@ -387,7 +387,6 @@ test('getApplesCount has some unusual side effects...', () => {
 Providing a `Set` is an experimental feature and may change in a future release.
 :::
 
-
 ```ts
 import { expect, test } from 'vitest'
 


sheremet-va:
> would fail. I don't think `toBeOneOf` will ever be used like that, therefore I just used the native `Set.has` method. Does that work for you?

I would expect `toBeOneOf([expect.any()])` to work the same as `toBeOneOf(new Set([expect.any()]))`

tim-we:
> I would expect toBeOneOf([expect.any()]) to work the same as toBeOneOf(new Set([expect.any()]))

@sheremet-va I changed the implementation to behave the exact same way as the array. 

No idea why there are 4 checks failing, is that to be expected at the moment or is this somehow caused by my changes?

docs/api/expect.md
```diff
@@ -379,9 +379,9 @@ test('getApplesCount has some unusual side effects...', () => {
 
 ## toBeOneOf
 
-- **Type:** `(sample: Array<any>) => any`
+- **Type:** `(sample: Array<any> | Set<any>) => any`
 
-`toBeOneOf` asserts if a value matches any of the values in the provided array.
+`toBeOneOf` asserts if a value matches any of the values in the provided array or set.
```
sheremet-va, line 384: Let's mention that providing a set is an experimental feature so we can merge this sooner

sheremet-va (dismissed):

docs/api/expect.md
tim-we: Like this?

docs/api/expect.md
```diff
@@ -379,9 +379,12 @@ test('getApplesCount has some unusual side effects...', () => {
 
 ## toBeOneOf
 
-- **Type:** `(sample: Array<any>) => any`
+- **Type:** `(sample: Array<any> | Set<any>) => any`
+
+`toBeOneOf` asserts if a value matches any of the values in the provided array or set.
+
+> Providing a `Set` is an experimental feature and may change in a future release.
```
sheremet-va:
```suggestion
::: warning EXPERIMENTAL
Providing a `Set` is an experimental feature and may change in a future release.
:::
```

docs/api/expect.md
```diff
@@ -379,9 +379,14 @@ test('getApplesCount has some unusual side effects...', () => {
...
-- **Type:** `(sample: Array<any>) => any`
+- **Type:** `(sample: Array<any> | Set<any>) => any`
+
+`toBeOneOf` asserts if a value matches any of the values in the provided array or set.
+
+::: warning EXPERIMENTAL
+Providing a `Set` is an experimental feature and may change in a future release.
+:::
 
-`toBeOneOf` asserts if a value matches any of the values in the provided array.
```
sheremet-va:
```suggestion
```

sheremet-va (approved):

PR #8921, opened 2025-11-03
fix: support iterables in toBeOneOf matcher (fix #8902)

### LINKED ISSUE #8902: Support sets as argument to expect.toBeOneOf

pabloleonalcaide:
### Description

Adds support for iterables (Set, Map, generators, etc.) to the `toBeOneOf` matcher, expanding beyond just arrays as requested.

Currently, `toBeOneOf` only accepts arrays. However, Sets and other iterables are common in JavaScript and forcing users to convert them to arrays (`[...mySet]`) is unnecessary.

Resolves  #8902.

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/api/expect.md
+++ b/docs/api/expect.md
@@ -379,19 +379,23 @@ test('getApplesCount has some unusual side effects...', () => {
 
 ## toBeOneOf
 
-- **Type:** `(sample: Array<any>) => any`
+- **Type:** `(sample: Array<any> | Iterable<any>) => any`
 
-`toBeOneOf` asserts if a value matches any of the values in the provided array.
+`toBeOneOf` asserts if a value matches any of the values in the provided array or iterable (Set, Map, etc.).
 
 ```ts
 import { expect, test } from 'vitest'
 
 test('fruit is one of the allowed values', () => {
-  expect(fruit).toBeOneOf(['apple', 'banana', 'orange'])
+  expect('apple').toBeOneOf(['apple', 'banana', 'orange'])
+
+  // Also works with Sets and other iterables
+  const allowedFruits = new Set(['apple', 'banana', 'orange'])
+  expect('apple').toBeOneOf(allowedFruits)
 })
 ```
 
-The asymmetric matcher is particularly useful when testing optional properties that could be either `null` or `undefined`:
+The asymmetric matcher is particularly useful when testing optional properties:
 
 ```ts
 test('optional properties can be null or undefined', () => {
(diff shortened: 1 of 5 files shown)


tim-we:
Hi there, did you see that there was already [another MR](https://github.com/vitest-dev/vitest/pull/8906) for #8902 open?
If all iterables should be supported, this is the better implementation, but I am not sure if that makes sense. I wonder for example if putting a `Map` into this method makes sense and what the expected behavior would be. `Array.from(map)` would be an array of the map entries and compare the received value against that.

pabloleonalcaide:
> Hi there, did you see that there was already [another MR](https://github.com/vitest-dev/vitest/pull/8906) for #8902 open? If all iterables should be supported, this is the better implementation, but I am not sure if that makes sense. I wonder for example if putting a `Map` into this method makes sense and what the expected behavior would be. `Array.from(map)` would be an array of the map entries and compare the received value against that.

Hi!
Thanks for the feedback! My goal with this change was to allow any iterable (like Set, Map.values(), generators, etc.), not just Set. But you're right — the behavior with Map could not be very clear.

I can update the PR to make the supported iterable types explicit and add a test case for Map to define how it should behave. Does that sound good to you?

tim-we:
This is not for me to decide, I am just the one who opened the issue. I have a use case for supporting `Set`s and for other iterables besides Array I don't have one (right now). Maybe this should have been discussed in the issue before opening a second PR? 

Anyway I'm happy when either one of the PRs gets merged, as long as `Set`s are supported in the future :)

pabloleonalcaide:
Yes, probably. I didn't actually see the first PR when I saw the issue, and once it was open, I thought it would be interesting to share this second approach.
In any case, it will be great to see the functionality of tobeoneof expanded.

Cheers!

pabloleonalcaide:
After seeing that [this pull request](https://github.com/vitest-dev/vitest/pull/8906) has been merged, I suspect mine is no longer necessary, correct?
I'll keep it open until you confirm, thank you.

sheremet-va:
I think we can close this now, yes.

PR #8937, opened 2025-11-04
fix: When no argument is passed to the .click interaction command, the webdriver command should also have no argument

### LINKED ISSUE #8931: Webdrivers's click should be called with undefined as the default value
### Describe the bug

I noticed that `actionClick` is used instead of `elementClick` in webdriver (context: https://github.com/webdriverio/webdriverio/blob/e86efccd84e20b01713a43113d166d8deb397edc/packages/webdriverio/src/commands/element/click.ts#L109-L121), even when passing no option.

The problem is that an empty object is passed by default in:

https://github.com/vitest-dev/vitest/blob/ab9a690a2dcc1467345e8a44541a3f11e6b8f307/packages/browser-webdriverio/src/commands/click.ts#L4-L11

I noticed that because `actionClick` was giving me other errors later (probably a webdriverio or browser issue, I'll deal with that too).

### Reproduction

I don't think that's needed

### System Info

```shell
System:
    OS: Linux 6.7 Debian GNU/Linux 12 (bookworm) 12 (bookworm)
    CPU: (16) x64 11th Gen Intel(R) Core(TM) i9-11900H @ 2.50GHz
    Memory: 30.91 GB / 62.54 GB
    Container: Yes
    Shell: 5.2.15 - /bin/bash
  Binaries:
    Node: 24.8.0 - /home/julien/.nvm/versions/node/v24.8.0/bin/node
    Yarn: 1.22.22 - /usr/bin/yarn
    npm: 11.6.0 - /home/julien/.nvm/versions/node/v24.8.0/bin/npm
    pnpm: 10.9.0 - /home/julien/.local/share/pnpm/pnpm
    bun: 1.0.29 - /home/julien/.bun/bin/bun
    Deno: 2.0.0 - /home/julien/.cargo/bin/deno
    Watchman: 4.9.0 - /usr/bin/watchman
  Browsers:
    Chrome: 141.0.7390.107
    Chromium: 141.0.7390.65
    Firefox: 140.4.0esr
    Firefox Developer Edition: 140.4.0esr
  npmPackages:
    @vitest/browser-webdriverio: ^4.0.4 => 4.0.4 
    @vitest/coverage-istanbul: ^4.0.4 => 4.0.4 
    @vitest/eslint-plugin: ^1.3.26 => 1.3.26 
    vite: ^7.1.12 => 7.1.12 
    vitest: ^4.0.4 => 4.0.4 
    vitest-browser-lit: ^1.0.1 => 1.0.1 
    webdriverio: ^9.20.0 => 9.20.0
```

### Used Package Manager

npm

### Validations

julienw:
Indeed webdriver.io has a different behavior when the argument is not undefined.

Fixes #8931

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/guide/browser/interactivity-api.md
+++ b/docs/guide/browser/interactivity-api.md
@@ -59,7 +59,31 @@ test('clicks on an element', async () => {
   await userEvent.click(logo)
   // or you can access it directly on the locator
   await logo.click()
+
+  // With WebdriverIO, this uses either ElementClick (with no arguments) or
+  // actions (with arguments). Use an empty object to force the use of actions.
+  await logo.click({})
 })
+```
+
+### Clicking with a modifier
+
+With either WebdriverIO or Playwright:
+
+```ts
+await userEvent.keyboard('{Shift>}')
+// By using an empty object as the option, this opts in to using a chain of actions
+// instead of an ElementClick in webdriver.
+// Firefox has a bug that makes this necessary.
+// Follow https://bugzilla.mozilla.org/show_bug.cgi?id=1456642 to know when this
+// will be fixed.
+await userEvent.click(element, {})
+await userEvent.keyboard('{/Shift}')
+```
+
+With Playwright:
+```ts
+await userEvent.click(element, { modifiers: ['Shift'] })
 ```
 
 References:


packages/browser-webdriverio/src/commands/click.ts
```diff
@@ -4,10 +4,10 @@ import type { UserEventCommand } from './utils'
 export const click: UserEventCommand<UserEvent['click']> = async (
   context,
   selector,
-  options = {},
+  options,
```
sheremet-va, line 7: Did you check what is passed here when `= {}` is removed? I think we always normalize it in the `browser/*/locators/index.ts`

Maybe a better test could catch that 😄

julienw: ah good question, I'll double check at least by adding logs :)

I'm not sure how to do a good test, as they should essentially do the same thing in the end ... except with more work (possibly taking longer especially).

julienw:
FWIW this is the webdriver.io bug I was running into because of that: https://github.com/webdriverio/webdriverio/issues/14840

packages/browser-webdriverio/src/commands/click.ts
julienw: or maybe I could try to mock the call to see what is _really_ called...

test/browser/test/userEvent.test.ts
```diff
@@ -29,6 +29,10 @@ describe('userEvent.click', () => {
 
     expect(onClick).toHaveBeenCalled()
     expect(dblClick).not.toHaveBeenCalled()
+
+    onClick.mockClear()
+    await userEvent.click(button, { button: 'right' })
+    expect(onClick).toHaveBeenCalled()
```
julienw: Can you give me some hint about how to run this test only? I couldn't find it :/

sheremet-va: in `test/core`:

```
PROVIDER=webdriverio pnpm test-fixtures userEvent --run -t "correctly clicks a button"
```

packages/browser-webdriverio/src/commands/click.ts
julienw: Yeah you're right of course:

https://github.com/vitest-dev/vitest/blob/1f730373858a7d7f49785cca5d6c45d17afc77b6/packages/browser/src/client/tester/locators/index.ts#L77

julienw: Playwright defaults to `{}` too, and the `preview` one doesn't take any option, so I guess we can remove the default everywhere.

https://github.com/microsoft/playwright/blob/e7bff526433b6dcb02801763ab5b1c6407902d47/packages/playwright-core/src/client/locator.ts#L108

test/browser/test/userEvent.test.ts
julienw: I guess you meant `test/browser` :)

I see my new test wasn't passing either, but I don't see that in the CI, are they running there?

Now I can fix all of that, thanks!

julienw:
It's not clear to me why the tests are failing :/ I tested locally the individual files, they pass for me, but they don't seem to pass when testing the runner...

Though I'm on Linux, Linux passes too in CI, just not Windows and MacOS, so there might be something different there?

julienw:
I reverted the changes on playwright because this wasn't related to this issue, I just wanted to make things consistent and I thought this would change nothing. But maybe I misread the playwright code. Let's see if this fixes the tests.

__Update__: this didn't fix the test.

packages/browser-webdriverio/src/commands/click.ts
julienw: I checked using `WDIO_LOG_LEVEL=info` that `elementClick` is correctly called now

I don't have a good idea for a test, I think we'd be testing webdriver.io itself... and I'm also not sure how I'd correctly spy the right API.

julienw:
By preventing the context menu from appearing when clicking right, I fixed the tests, to make new failures appear, which make more sense. I wonder if that might be a bug in Firefox...

It's still pretty weird that only macos fails, and not linux. (I don't know for windows yet) __update__: Ah of course, browser tests are not run on Ubuntu.

packages/browser-webdriverio/src/commands/click.ts
sheremet-va: Sure, let's not add a new test specifically for this, but it's also important that already established tests don't fail 😄 I can see there is now only `click with modifier` on firefox failing for some reason 🤔

julienw: yeah I was looking into that too...

https://github.com/vitest-dev/vitest/blob/1f730373858a7d7f49785cca5d6c45d17afc77b6/test/browser/fixtures/user-event/keyboard.test.ts#L65-L67

I've been told from webdriver's firefox implementor that it's expected: this works with actions but not elementClick at the moment. Which is quite unfortunate :-( But they said also that they didn't implement it because the other browsers didn't either.

It's not clear to me if the test runs in Chrome to check this assumption is true? I don't see it in the test output, but also I see nothing in the file that would run it only in Firefox...

julienw: I'm thinking this could be an issue in webdriver.io too, they could/should track the keyboard and decide if they should do an elementClick or a chain of actions.

julienw: I pushed an update to fix the test. But this makes the change a breaking change, even if it's webdriver + firefox only.

julienw: I created a webdriver.io-based test to look at this:

https://github.com/julienw/webdriver-click-with-modifier/blob/ae33a2e1ff8a8697922ccbf4ec7080a5e1c81339/test/shift-click.test.js#L103-L107

It looks like there is a behavior difference in Firefox :/

julienw: https://bugzilla.mozilla.org/show_bug.cgi?id=1456642 turns out there's a bug open in Firefox, it looks like it's actively worked on these days.

Given there's an easy workaround (passing `{}` as a parameter to click), I'll leave it up to you to decide if you want to merge this.

sheremet-va: I think if we can add a documentation notice to the `click` event, it's fine to merge. The impact should not be big

julienw: I pushed some documentation, tell me if that looks like what you expected.

sheremet-va: Yeah, looks good to me, just make sure eslint doesn't fail

julienw: oh no, should be fixed now :-)

sheremet-va (approved):

sheremet-va:
Thank you!

julienw:
Thanks!

julienw:
Is there a changelog to update about this, or will you add it to the release notes ?

sheremet-va:
> Is there a changelog to update about this, or will you add it to the release notes ?

release notes are generated based on the squashed commit message

julienw:
Okay, just want to make sure the folks know what to do in case they see failures because of that.

julienw:
So this change actually avoided the bug I mentioned earlier https://github.com/webdriverio/webdriverio/issues/14840 in Firefox... but now I get a new webdriver bug in Chrome, related to a web component wrapping a button element. Nothing related to vitest IMO, but I wanted to mention it in case other folks got the issue.

julienw:
Not related to vitest directly, but mentioning it in case this can be useful to other folks.
Using ElementClick instead of the actions like previously hits this Chrome issue sometimes: https://groups.google.com/g/chromedriver-users/c/9oxCEdNUdYg/m/7A4DIbbnAQAJ :(

PR #8952, opened 2025-11-06
fix(browser): locator which does not locate any element is not visible

jnizet:
### Description

When using an assertion such as `expect(page.getByTestId('not-existing')).not.toBeVisible()`, the assertion used to fail because the assertion tried to get the element of the locator, which threw an exception. After this fix, the assertion will pass, because an element that does not exist (and which is thus not in the document) is de facto not visible. It allows the developer to use `not.toBeVisible()` without having to care if the element is hidden in the DOM or removed from the DOM.

See https://github.com/vitest-dev/vitest/discussions/8946#discussioncomment-14882890

### Tests
- [x] Run the tests with `pnpm test:ci`. **but so many tests are failing that can't possibly come from the tiny fix I introduced**

### Documentation

### Changesets

### DIFF
--- a/packages/browser/src/client/tester/expect-element.ts
+++ b/packages/browser/src/client/tester/expect-element.ts
@@ -21,7 +21,7 @@ function element<T extends HTMLElement | SVGElement | null | Locator>(elementOrL
     const isNot = chai.util.flag(this, 'negate') as boolean
     const name = chai.util.flag(this, '_name') as string
     // special case for `toBeInTheDocument` matcher
-    if (isNot && name === 'toBeInTheDocument') {
+    if (isNot && (name === 'toBeInTheDocument' || name === 'toBeVisible')) {
       return elementOrLocator.query()
     }
     if (name === 'toHaveLength') {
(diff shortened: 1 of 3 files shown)


jnizet:
I just realized that this fix only partially fixes the issue. 

An assertion such as `expect(page.getByTestId('not-existing')).not.toBeVisible()` will pass, but unfortunatly, an assertion such as `await expect.element(page.getByTestId('not-existing')).not.toBeVisible()` will still fail, because `expect.element(locator)` always tries to resolve the element of the locator, instead of repeating the assertion on the locator until it succeeds.

I guess I'll need to look into how the assertion `.not.toBeInTheDocument()` works, but I wonder why `expect.element(locator).toXxx()` is equivalent to `retryUntilItSucceeds(() => expect(locator.element()).toXxx())` instead of being equivalent to `retryUntilItSucceeds(() => expect(locator).toXxx())`.

jnizet:
I added a second commit which makes it work with `expect(page.getByTestId('not-existing')).not.toBeVisible()`

PR #8954, opened 2025-11-06
fix: incorrect error message for non-awaited `expect.element()`

### LINKED ISSUE #8953: Incorrect error message in non-awaited `expect.element()` calls
### Describe the bug

When not awaiting an `expect.element()` call, there seem to be an issue with the error text. it shows the error for `expect.poll`:

(image taken from https://github.com/vitest-dev/eslint-plugin-vitest/issues/683)

### Reproduction

```TS
test("renders name", async () => {
  const element = page.getByText("Hello Vitest!");

  expect.element(element).toBeInTheDocument();
});

```

### System Info

```shell
Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.21.1 - /Users/**/.local/state/fnm_multishells/15956_1762412239648/bin/node
    Yarn: 1.22.22 - /Users/**/.local/state/fnm_multishells/15956_1762412239648/bin/yarn
    npm: 10.9.4 - /Users/**/.local/state/fnm_multishells/15956_1762412239648/bin/npm
    pnpm: 10.20.0 - /Users/**/.local/state/fnm_multishells/15956_1762412239648/bin/pnpm
  Browsers:
    Chrome: 142.0.7444.60
    Firefox Developer Edition: 144.0
    Safari: 18.5
  npmPackages:
    @vitest/browser-playwright: ^4.0.7 => 4.0.7 
    vitest: ^4.0.7 => 4.0.7
```

### Used Package Manager

npm

### Validations

StyleShit:
### Description

Resolves #8953

Flagging the callback happened [inside the callback itself](https://github.com/vitest-dev/vitest/blob/97aed7454f09d352a9e71b3cd67c3584238cf455/packages/browser/src/client/tester/expect-element.ts#L19), which runs after [the check](https://github.com/vitest-dev/vitest/blob/97aed7454f09d352a9e71b3cd67c3584238cf455/packages/vitest/src/integrations/chai/poll.ts#L113) whether it's `poll` or `element`, so I flagged it outside the callback

I want to add some tests. Saw that there are tests for [`poll-no-awaited`](https://github.com/vitest-dev/vitest/blob/97aed7454f09d352a9e71b3cd67c3584238cf455/test/cli/fixtures/fails/poll-no-awaited.test.ts), but I'm assuming that `expect.element` tests shouldn't be there and I wasn't sure how to test it. Would be glad to get some help here

### DIFF
--- a/test/browser/specs/__snapshots__/expect-element-no-awaited.test.ts.snap
+++ b/test/browser/specs/__snapshots__/expect-element-no-awaited.test.ts.snap
@@ -1,0 +1,23 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`should fail for non-awaited expect.element 1`] = `
+"
+⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯
+
+ FAIL  |browser (chromium)| expect-element.test.js > not awaited
+Error: expect.element(locator).toBeInTheDocument() was not awaited. This assertion is asynchronous and must be awaited; otherwise, it is not executed to avoid unhandled rejections:
+
+await expect.element(locator).toBeInTheDocument()
+
+ ❯ expect-element.test.js:18:33
+     16|         const element = page.getByText("Hello Vitest!");
+     17| 
+     18|         expect.element(element).toBeInTheDocument();
+       |                                 ^
+     19|       })
+     20|       
+
+⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯
+
+"
+`;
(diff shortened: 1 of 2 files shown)


sheremet-va:
The error is correct, just the message is wrong.

sheremet-va:
I don't see how this fixes the message. Please, add a test somewhere in `test/browser`, probably

StyleShit:
> I don't see how this fixes the message

Determining the error message is based on a flag that's attached to the expect function reference (either `expect.poll()` or `expect.element()`).

What happened is that the flagging of `expect.element()` happens inside the function itself
(basically `function expectElement() { addFlag(this, 'expect.element', true) }`),
but the check against this flag runs _before_ this function even runs, so it's not flagged yet.

> add a test

Added. Took me some time to figure it out...

packages/browser/src/client/tester/expect-element.ts
```diff
@@ -47,6 +46,10 @@ function element<T extends HTMLElement | SVGElement | null | Locator>(elementOrL
 
     return result
   }, processTimeoutOptions(options))
+
+  chai.util.flag(expectElement, '_poll.element', true)
```
Omerisra6, line 50: Hi, new here :) 
How does this ensures the right message?

sheremet-va:
> but the check against this flag runs _before_ this function even runs, so it's not flagged yet.

I see, the `poll` is not called at all if the function is not awaited, so the flag is never set

PR #8971, opened 2025-11-07
fix(bail): fix race condition in bail

### LINKED ISSUE #8852: Race condition in new pool implementation when using bail
### Describe the bug

There is a race condition in the new worker pool implementation.

### Reproduction

See repro.zip.

The relevant part is the `run-vitest.ts` file.

```ts
import { createVitest } from "vitest/node";
const vitest = await createVitest("test", {
  pool: "threads",
  coverage: { enabled: false },
  maxWorkers: 1,
  maxConcurrency: 1,
  watch: false,
  bail: 1,
  reporters: [],
});

while (vitest.state.errorsSet.size === 0) {
  await vitest.start();
}
reportErrors();

function reportErrors() {
  if (vitest.state.errorsSet.size > 0) {
    const msg = [...vitest.state.errorsSet]
      .map((err) => err.message)
      .join("\n");
    console.error("Test run errored with:\n", msg);
  } else {
    console.log("No errors!");
  }
}
```

I would expect this program to run infinitely. What actually happens:
- It stops after ~1 second
- It logs: ` [vitest-pool]: Cannot run tasks while pool is cancelling`

The race condition concerns bail. The canceling of the run doesn't finish before the run itself finishes (there is 1 failing test).

[repro.zip](https://github.com/user-attachments/files/23184949/repro.zip)

### System Info

```shell
System:
    OS: macOS 26.0.1
    CPU: (14) arm64 Apple M4 Pro
    Memory: 989.08 MB / 48.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.18.0 - /Users/nicojs/.nvm/versions/node/v22.18.0/bin/node
    Yarn: 1.22.22 - /Users/nicojs/.nvm/versions/node/v22.18.0/bin/yarn
    npm: 11.5.2 - /Users/nicojs/.nvm/versions/node/v22.18.0/bin/npm
    pnpm: 10.19.0 - /Users/nicojs/Library/pnpm/pnpm
  Browsers:
    Chrome: 141.0.7390.123
    Edge: 141.0.3537.99
    Firefox: 144.0
    Safari: 26.0.1
  npmPackages:
    vitest: ^4.0.4 => 4.0.4
```

### Used Package Manager

npm

### Validations

nicojs:
### Description

Resolves #8852

### Tests

### Documentation

(no new functionality)

### Changesets

### DIFF
--- a/test/cli/test/bail-race.test.ts
+++ b/test/cli/test/bail-race.test.ts
@@ -32,6 +32,10 @@ test('cancels previous run before starting new one', async () => {
     rounds++
   }
 
+  if (vitest.state.errorsSet.size > 0) {
+    throw vitest.state.errorsSet.values().next().value
+  }
+
   expect(results).toMatchInlineSnapshot(`
     [
       {


nicojs:
I don't know why the tests are failing. @sheremet-va  any idea?

sheremet-va:
> I don't know why the tests are failing. @sheremet-va any idea?

They are flaky. If you didn't touch them, don't worry.

As a note, your test needs to be inside `test/cli`, tests are separated in groups, see test/README

nicojs:
Thanks for the review! 🙏

I've moved the test to `test/cli`.

nicojs:
@sheremet-va I've improved the test, made it silent, assert happy path and clarified with some comments. What do you think?

packages/vitest/src/node/pools/pool.ts
```diff
@@ -47,9 +47,9 @@ export class Pool {
   }
 
   async run(task: PoolTask, method: 'run' | 'collect'): Promise<void> {
-    // Prevent new tasks from being queued during cancellation
-    if (this._isCancelling) {
-      throw new Error('[vitest-pool]: Cannot run tasks while pool is cancelling')
+    // Wait for any ongoing cancellation to complete before accepting new tasks
+    if (this.cancellingPromise) {
+      await this.cancellingPromise
```
AriPerkkio, line 52: I wonder if this change will start some previous run's tests too. I have no idea why `this._isCancelling` was required before - @sheremet-va I think that's something you added?

packages/vitest/src/node/pools/pool.ts
```diff
@@ -167,28 +167,30 @@ export class Pool {
   }
 
   async cancel(): Promise<void> {
-    // Set flag to prevent new tasks from being queued
-    this._isCancelling = true
```
AriPerkkio, line 171: Pool should already have had all tests queued at this point. Or if there was another group, the `ctx.isCancelling` in `node/pool.ts` should have prevented them from being queued.

packages/vitest/src/node/pools/pool.ts
sheremet-va: > I have no idea why `this._isCancelling` was required before

The comment explains it:
> Prevent new tasks from being queued during cancellation

packages/vitest/src/node/pools/pool.ts
sheremet-va: If this was true, there wouldn't have been an issue https://github.com/vitest-dev/vitest/issues/8852

packages/vitest/src/node/pools/pool.ts
```diff
@@ -167,28 +167,30 @@ export class Pool {
...
 
-    const sharedRunners = this.sharedRunners.splice(0)
-    await Promise.all(sharedRunners.map(runner => runner.stop()))
+      await Promise.all(this.exitPromises.splice(0))
 
-    await Promise.all(this.exitPromises.splice(0))
+      this.workerIds.forEach((_, id) => this.freeWorkerId(id))
 
-    this.workerIds.forEach((_, id) => this.freeWorkerId(id))
+      this.cancellingPromise = null
```
sheremet-va: I would set it to null on the next line in `finally` in case there is an unexpected error

test/cli/test/bail-race.test.ts
```diff
@@ -0,0 +1,44 @@
+import { resolve } from 'pathe'
+import { expect, test } from 'vitest'
+import { createVitest } from 'vitest/node'
```
sheremet-va: We have `runVitest` in `../test-utils` that you can use instead. It sets all the flags (maxWorkers, watch) automatically

test/cli/test/bail-race.test.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    if (rounds >= 2) {
+      break
+    }
+
+    rounds++
+  }
+
+  expect(results).toMatchInlineSnapshot(`
+    [
```
sheremet-va, line 40: This structure is a bit confusing to me, it doesn't have test names. `test-utils` has a `buildTree` function somewhere, maybe use it here instead? I think at the moment it is only returned from `runInlinedTests`, but you can add a return in `runVitest`

sheremet-va (changes requested):

packages/vitest/src/node/pools/pool.ts
sheremet-va: These kinds of checks exist to catch any unintended behaviours

AriPerkkio (changes requested):
To me this looks more like race condition in `core.ts`: 

We start cancelling here:

https://github.com/vitest-dev/vitest/blob/26d2b60c133f22b8a4b917959f3c4adfba9fe066/packages/vitest/src/node/core.ts#L861

And it's then overwritten by

https://github.com/vitest-dev/vitest/blob/26d2b60c133f22b8a4b917959f3c4adfba9fe066/packages/vitest/src/node/core.ts#L707-L710

This is noticable with your reproduction. No idea why that happens.

packages/vitest/src/node/pools/pool.ts
AriPerkkio: Now that we just `await` here and then let it continue, doesn't it mean it's going to continue the `run` it was intended the cancel? Should it `return` instead?

nicojs: No, that was the previous cancellation (because of `bail`). So it needs to wait for that to finish.

packages/vitest/src/node/pools/pool.ts
nicojs: I've updated this

nicojs:
@AriPerkkio I now use your implementation with `rounds >= 2`, however the race condition will almost never happen within the first 3 rounds. Should we make it maybe 20 rounds to make sure?

I've implemented the changes requested by @sheremet-va 

@AriPerkkio I'm not familiar enough with the code base to know how cancellation in `core.ts` is supposed to work. Maybe @sheremet-va knows the answer?

test/cli/test/bail-race.test.ts
```diff
@@ -0,0 +1,57 @@
+import { resolve } from 'pathe'
+import { expect, test } from 'vitest'
+import { runVitest } from '../../test-utils'
+
+test('cancels previous run before starting new one', async () => {
+  const results: Record<string, unknown>[] = []
+
+  const { ctx: vitest, buildTestTree } = await runVitest({
```
AriPerkkio, line 8: Does this bug reproduce with `runVitest`? I'm only able to see it via `createVitest`.

nicojs: I get reproduce it structurally, as long as I use `pool: 'threads'`

AriPerkkio:
The underlying bug is rather in `core.ts` that keeps starting new runs even though it's still cancelling previous ones. I've pushed those fixes in https://github.com/vitest-dev/vitest/pull/8991 to avoid overwriting changes in this PR.

PR #8991, opened 2025-11-10
fix(core): prevent starting new run when cancelling

AriPerkkio:
### Description

Alternative for #8971.

- Fixes https://github.com/vitest-dev/vitest/issues/8852
- Closes https://github.com/vitest-dev/vitest/pull/8971

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/core.ts
+++ b/packages/vitest/src/node/core.ts
@@ -99,6 +99,7 @@ export class Vitest {
   /** @internal */ filenamePattern?: string[]
   /** @internal */ runningPromise?: Promise<TestRunResult>
   /** @internal */ closingPromise?: Promise<void>
+  /** @internal */ cancelPromise?: Promise<void | void[]>
   /** @internal */ isCancelling = false
   /** @internal */ coreWorkspaceProject: TestProject | undefined
   /** @internal */ _browserSessions = new BrowserSessions()
@@ -705,6 +706,7 @@ export class Vitest {
     await this._testRun.start(specs)
 
     // previous run
+    await this.cancelPromise
     await this.runningPromise
     this._onCancelListeners.clear()
     this.isCancelling = false
@@ -806,6 +808,7 @@ export class Vitest {
     this.state.collectPaths(filepaths)
 
     // previous run
+    await this.cancelPromise
     await this.runningPromise
     this._onCancelListeners.clear()
     this.isCancelling = false
@@ -859,7 +862,9 @@ export class Vitest {
    */
   async cancelCurrentRun(reason: CancelReason): Promise<void> {
     this.isCancelling = true
-    await Promise.all([...this._onCancelListeners].map(listener => listener(reason)))
+    this.cancelPromise = Promise.all([...this._onCancelListeners].map(listener => listener(reason)))
+
+    await this.cancelPromise.finally(() => (this.cancelPromise = undefined))
     await this.runningPromise
   }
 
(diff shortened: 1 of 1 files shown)


test/cli/test/bail-race.test.ts
```diff
@@ -0,0 +1,31 @@
...
+      },
+    }],
+  })
+
+  for (let i = 0; i <= 10; i++) {
+    await vitest.start()
+  }
+
+  // No "Error: [vitest-pool]: Cannot run tasks while pool is cancelling" errors should show up
+  expect(errors).toHaveLength(0)
```
AriPerkkio, line 30: Example failure without the fix:

https://github.com/vitest-dev/vitest/actions/runs/19228675773/job/54961833341?pr=8991

packages/vitest/src/node/core.ts
```diff
@@ -858,9 +861,14 @@ export class Vitest {
...
-    this.isCancelling = true
-    await Promise.all(this._onCancelListeners.splice(0).map(listener => listener(reason)))
-    await this.runningPromise
+    const cancel = async () => {
+      this.isCancelling = true
+      await Promise.all(this._onCancelListeners.splice(0).map(listener => listener(reason)))
+      await this.runningPromise
+    }
+
+    this.cancelPromise = new Promise<void>(resolve => cancel().then(resolve))
```
sheremet-va: Can't you just call `cancel` here without creating a promise manually

packages/vitest/src/node/pools/poolRunner.ts
```diff
@@ -63,7 +63,11 @@ export class PoolRunner {
       },
     )
 
-    this.project.vitest.onCancel(reason => this._rpc.onCancel(reason))
+    this.project.vitest.onCancel((reason) => {
+      if (!this._rpc.$closed) {
+        this._rpc.onCancel(reason)
```
sheremet-va: Let's return the `onCancel` promise here like it was previously. I don't think it matters much in our case, but could create an unexpected race condition down the line

packages/vitest/src/node/core.ts
sheremet-va: It should also reset `this.cancelPromise` in `finally`

AriPerkkio: Ah, yes. I had more logic in this part earlier so additional promise was needed.

packages/vitest/src/node/pools/poolRunner.ts
AriPerkkio: Sure. As `onCancel` is defined in `eventNames` it won't be awaiting for rpc response. But `birpc` could still have promise based logic on sender side so better to handle the promise.

AriPerkkio:
> Can't we merge both solutions? 😄

I would rather not have this logic on multiple levels. And it would be good to see pool throwing the error when `vitest/ctx/core` attempts to use it incorrectly. 

But feel free to choose one or both.

nicojs:
I don't mind as long as it's fixed ☺️. StrykerJS users need it solved 😅

nicojs:
Any way we can get this merged?

packages/vitest/src/node/pools/poolRunner.ts
AriPerkkio: This wasn't needed after https://github.com/vitest-dev/vitest/pull/9033 anymore.

sheremet-va (approved):

PR #9010, opened 2025-11-12
chore: enable trust policy

iiio2:
Since we are using pnpm 10.21, it has a very good security feature for enabling trust policy. i think maybe we can use it.
https://pnpm.io/settings#trustpolicy


iiio2:
Thanks @sheremet-va . :blush:

PR #9011, opened 2025-11-12
feat(browser): add page.route intercept support

apple-yagi:
### Description

Adds first-class network interception support to Vitest browser mode. page.route, page.unroute, and page.unrouteAll are now available to tests, piping through new RPC commands and provider hooks. The Playwright backend reuses its native routing while WebdriverIO integrates with browser.mock (BiDi) so both providers can fulfill, continue, or abort requests. Tests cover fulfill/abort flows and cleanup guarantees.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/browser-playwright/src/playwright.ts
+++ b/packages/browser-playwright/src/playwright.ts
@@ -12,6 +12,7 @@ import type {
   LaunchOptions,
   Page,
   CDPSession as PlaywrightCDPSession,
+  Route as PlaywrightRoute,
 } from 'playwright'
 import type { SourceMap } from 'rollup'
 import type { ResolvedConfig } from 'vite'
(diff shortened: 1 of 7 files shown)


apple-yagi:
@sheremet-va Thanks for the feedback. My thinking was that supporting the Playwright Route API would let us avoid bringing in third-party tooling like MSW and improve the developer experience. Could you explain why MSW would be preferable here? I may not fully understand Vitest’s direction, so I’d appreciate a bit more context.

sheremet-va:
> Could you explain why MSW would be preferable here? I may not fully understand Vitest’s direction, so I’d appreciate a bit more context.

I am not saying _MSW_ would be preferable, but _MSW API_, if we ever implement HTTP interception. I see several reasons:
- MSW users can reuse the handlers. The convention is already to have handlers separated from MSW's `createServer`
- MSW API promotes the usage of standards by accepting `Response` and providing `Request`
- We already promote using MSW in the documentation and providing a different API is confusing
- MSW also exposes its primitives, so we can build on top of them

In general, we don't provide all playwright APIs, we provide APIs that we consider to be good. I really like locators, for example. But I think the `route` API is worse than what MSW provides.

apple-yagi:
@sheremet-va  Thank you very much for your explanation! That really helped me understand the situation. I appreciate your help. I'll go ahead and close this pull request.

sheremet-va:
As a note, I think the interception you implemented is really good, I would just prefer if the user-facing API (page.*) would follow MSW interface

apple-yagi:
@sheremet-va Taking the feedback into account, would it be alright if I submit another PR sometime? I’m not sure when yet, but I’d like to give it another shot.

sheremet-va:
> @sheremet-va Taking the feedback into account, would it be alright if I submit another PR sometime? I’m not sure when yet, but I’d like to give it another shot.

Yes, if you want to work on it

PR #9016, opened 2025-11-12
fix(browser): handle dependency stack traces with external source maps. Resolves: #9003

iclectic:
**Description**

- When running browser tests, stack traces from bundled dependencies (like preact.mjs inside node_modules/.vite) were pointing to minified code, which made debugging harder.
- This update lets the browser read external source maps so those stack traces now point to the original source.

**What changed:**

- Updated getBrowserFileSourceMap to read .map files if transformResult.map is missing.
- Added a new helper called retrieveSourceMapURL to find and load the map file.

**Why this matters:**

- Vite sometimes outputs dependency maps as separate files. Without loading them, stack traces stay minified and unreadable. This change makes them clearer and easier to debug.

**How it works:**

- We now parse the sourceMappingURL from transformed code, locate the map file on disk, read it, and return the result. Added the necessary imports to support this.

**Files updated:**

- packages/browser/src/node/rpc.ts

**Testing:**

- Build, type-check, and browser test suites have been run locally, along with the main repo tests — all passed.

**Compatibility:**

- No breaking changes.


iclectic:
## CI Status Update

The 3 failing checks (Header rules, Pages changed, Redirect rules) appear to be deployment-related checks that were canceled, not test failures caused by this PR.

- All build and test checks are passing
- The failing checks are related to vitest-dev deployment infrastructure
- My changes only modify [packages/browser/src/node/rpc.ts] to fix source map resolution
- No documentation or deployment configuration was changed

These failures appear to be unrelated to the browser stack trace fix in this PR. Could a maintainer please advise if any action is needed on my end, or if these can be re-triggered?

Resolves #9003

sheremet-va (approved):

PR #9022, opened 2025-11-12
feat(expect): provide `task` in `MatchState`

macarie:
### Description

Provide access to `task` in `MatchState`. This makes it easier to access context and other test-related data in custom assertions.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/runtime/runners/test.ts
+++ b/packages/vitest/src/runtime/runners/test.ts
@@ -158,7 +158,6 @@ export class VitestTestRunner implements VitestRunner {
   onBeforeTryTask(test: Task): void {
     clearModuleMocks(this.config)
     this.snapshotClient.clearTest(test.file.filepath, test.id)
-
     setState(
       {
         assertionCalls: 0,


packages/vitest/src/integrations/chai/index.ts
```diff
@@ -15,7 +15,7 @@ import { getWorkerState } from '../../runtime/utils'
 import { createExpectPoll } from './poll'
 import './setup'
 
-export function createExpect(test?: TaskPopulated): ExpectStatic {
```
macarie, line 18: This has to be considered a breaking change, right? The only place I could find where this is used with a parameter, it provides a `Test`, which extends `TaskPopulated`.

Could use `TaskPopulated | Test` instead and check if `test.type === 'test'` maybe.

docs/guide/extending-matchers.md
```diff
@@ -112,6 +112,14 @@ This contains a set of utility functions that you can use to display messages.
 
 Full name of the current test (including describe block).
 
+### `task`
```
macarie, line 115: Should I add a `<Version>` here?

sheremet-va (approved):

Issue #8968, opened 2025-11-07
v4: Timeout starting forks runner.


lfantone:
### Describe the bug

After upgrading to Vitest 4 our test started to fail randomly with: 

> Error: [vitest-pool]: Timeout starting forks runner.

<img width="892" height="138" alt="Image" src="https://github.com/user-attachments/assets/095e08c7-fce1-4e67-bb3b-a89128fda828" />

The project is a monorepo with 25 different packages. The test are being orchestrated with `turborepo`

The failures not only happens on my local but also in our CI workflows.

The vite config in use is:

```js
{
  plugins: [
    react(),
    babel({
      filter: /\.[jt]sx?$/,
      babelConfig: {
        presets: ['@babel/preset-typescript'],
        plugins: ['@babel/plugin-syntax-jsx', ['babel-plugin-react-compiler', { target: '19' }]],
      },
    }),
    vanillaExtractPlugin(),
    tsconfigPaths(),
  ],
  test: {
    bail: 3,
    logHeapUsage: true,
    include: [ '**/*.test.*' ],
    exclude: [
      '**/node_modules',
      '**/coverage',
      '**/build',
      '**/dist',
      '**/types'
    ],
    setupFiles: [ './vitest.setup.js' ],
    coverage: {
      provider: 'v8',
      clean: true,
      include: ['**/app/**/*.{js,ts,jsx,tsx}', '**/components/**/*.{js,ts,jsx,tsx}'],
      allowExternal: false,
      exclude: ['**/node_modules', '**/coverage', '**/build', '**/dist', '**/types', '**/*.css'],
    },
    environment: 'happy-dom'
  }
}
```
Everything was running fine in v3.

### Reproduction

https://github.com/lfantone/vitest-threads-error

### System Info

```shell
System:
    OS: macOS 26.0.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 1.44 GB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.21.0 - /Users/lfantone/.volta/tools/image/node/22.21.0/bin/node
    Yarn: 4.10.3 - /Users/lfantone/.volta/tools/image/yarn/3.8.3/bin/yarn
    npm: 10.9.4 - /Users/lfantone/.volta/tools/image/node/22.21.0/bin/npm
  Browsers:
    Chrome: 142.0.7444.60
    Safari: 26.0.1
```

### Used Package Manager

yarn

### Validations

[...]

PR #9027, opened 2025-11-13
fix(forks): increase worker start timeout

AriPerkkio:
### Description

Fixes https://github.com/vitest-dev/vitest/issues/8968.

Increase worker start timeout from 5s to 90s and worker startup ping from 10s to 60s. We didn't have these timeout in v3.

Currently there's also 10s timeout used inside methods that are called within the 5s timer, so 5s is definitely not enough.

https://github.com/vitest-dev/vitest/blob/a415d0375afa4c3726174d8f87db340ca20a428c/packages/vitest/src/node/pools/poolRunner.ts#L105

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/pools/pool.ts
+++ b/packages/vitest/src/node/pools/pool.ts
@@ -8,7 +8,7 @@ import { TypecheckPoolWorker } from './workers/typecheckWorker'
 import { VmForksPoolWorker } from './workers/vmForksWorker'
 import { VmThreadsPoolWorker } from './workers/vmThreadsWorker'
 
-const WORKER_START_TIMEOUT = 5_000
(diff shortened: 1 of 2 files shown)


sheremet-va:
> There's also 10s timeout used inside methods that are called within this timer, so 5s is definitely not enough.

Actually, I am not even sure if 10s is enough now that we have environment resolution there (when timeout was added, it was a simple `return { ready: true }`)

AriPerkkio:
And we've seen some use custom environments to spin up databases (instead of fixtures, not sure why), which definitely can make startup take longer than 5s. 

Let's increase both of these timers to something longer, like 60s and 90s.

sheremet-va (approved):

PR #9033, opened 2025-11-14
fix: remove `onCancel` when worker is terminated

sheremet-va:
### Description

Resolves #issue-number

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/core.ts
+++ b/packages/vitest/src/node/core.ts
@@ -139,7 +139,7 @@ export class Vitest {
   private _onRestartListeners: OnServerRestartHandler[] = []
   private _onClose: (() => Awaitable<void>)[] = []
   private _onSetServer: OnServerRestartHandler[] = []
-  private _onCancelListeners: ((reason: CancelReason) => Awaitable<void>)[] = []
+  private _onCancelListeners = new Set<(reason: CancelReason) => Awaitable<void>>()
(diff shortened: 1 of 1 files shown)


packages/vitest/src/node/core.ts
```diff
@@ -1257,8 +1257,11 @@ export class Vitest {
   /**
    * Register a handler that will be called when the test run is cancelled with `vitest.cancelCurrentRun`.
    */
-  onCancel(fn: (reason: CancelReason) => Awaitable<void>): void {
+  onCancel(fn: (reason: CancelReason) => Awaitable<void>): () => void {
     this._onCancelListeners.push(fn)
+    return () => {
+      this._onCancelListeners.splice(this._onCancelListeners.indexOf(fn))
```
AriPerkkio: I think this needs count too, otherwise it removes everything starting from that index.


```suggestion
      this._onCancelListeners.splice(this._onCancelListeners.indexOf(fn), 1)
```

Also I would rather have a `index !== -1` check here too just to be safe. If the listener was somehow already removed, we may remove the last item from the array accidentally.

```ts
const index = this._onCancelListeners.indexOf(fn)

if(index !== -1) {
  this._onCancelListeners.splice(index, 1)
}
```

sheremet-va: Changed to `Set` instead, much easier to handle

Issue #8853, opened 2025-10-28
toMatchScreenshot cannot find currentTest


djrenren:
### Describe the bug

I'm attempting to use vitest's new `toMatchScreenshot()` matcher to test storybook stories.
After invoking storybook's `composeStory` function, the test context seems to get lost. It's likely that storybook is mucking with something global that's breaking something inside vitest but I couldn't find it on my own since I'm not very familiar with how browser mode works under the hood.

### Reproduction

The minimal repro looks like this:

```
import { expect, test } from 'vitest'
import { render } from 'vitest-browser-react'
import { composeStory } from '@storybook/react-vite'
import meta, { ExampleStory } from './HelloWorld.stories.tsx'

const Example = composeStory(ExampleStory, meta)

test('renders name', async () => {
  const screen = await render(<div data-testid="story"><Example /></div>);
  await expect(screen.getByTestId('story')).toMatchScreenshot()
})
```

and the error looks like:

<img width="699" height="287" alt="Image" src="https://github.com/user-attachments/assets/7c84ec04-30f5-4a02-a071-2d4ef8cef04d" />

but I've made a runnable project here which is a minimal change on top of the `npx vitest init browser`: https://github.com/djrenren/vitestScreenshotTestFailure.

Just run `npm run test:browser`

### System Info

```shell
System:
    OS: macOS 15.6.1
    CPU: (14) arm64 Apple M3 Max
    Memory: 1000.64 MB / 36.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.16.0 - /Users/john/.nvm/versions/node/v22.16.0/bin/node
    npm: 10.9.2 - /Users/john/.nvm/versions/node/v22.16.0/bin/npm
    pnpm: 10.18.3 - /Users/john/.nvm/versions/node/v22.16.0/bin/pnpm
    bun: 1.1.29 - /Users/john/.bun/bin/bun
    Deno: 1.41.3 - /Users/john/.deno/bin/deno
  Browsers:
    Brave Browser: 141.1.83.120
    Chrome: 141.0.7390.123
    Firefox: 144.0
    Firefox Developer Edition: 132.0
    Safari: 18.6
    Safari Technology Preview: 18.4
  npmPackages:
    @vitejs/plugin-react: ^5.1.0 => 5.1.0 
    @vitest/browser-playwright: ^4.0.4 => 4.0.4 
    vitest: ^4.0.4 => 4.0.4 
[...]

PR #9087, opened 2025-11-24
feat(runner): Add full names to tasks

macarie:
### Description

Add `fullName` and `fullTestName` to `TaskBase`.

- `fullName` is the concatenation of any parent suites and the task's name using ` > ` as a separator, it includes the file path.
- `fullTestName` is like `fullName`, excluding the file path. It's `undefined` for `File` tasks.

This should be the last necessary step towards a fix for #8853.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/runner/src/collect.ts
+++ b/packages/runner/src/collect.ts
@@ -42,7 +42,7 @@ export async function collectTests(
 
         runner.onCollectStart?.(file)
 
-        clearCollectorContext(filepath, runner, file)
+        clearCollectorContext(file, runner)
 
         try {
           const setupFiles = toArray(config.setupFiles)
--- a/packages/runner/src/suite.ts
+++ b/packages/runner/src/suite.ts
@@ -214,16 +214,15 @@ function createDefaultSuite(runner: VitestRunner) {
 }
 
 export function clearCollectorContext(
-  filepath: string,
-  currentRunner: VitestRunner,
   file: File,
+  currentRunner: VitestRunner,
 ): void {
   if (!defaultSuite) {
     defaultSuite = createDefaultSuite(currentRunner)
   }
   defaultSuite.file = file
   runner = currentRunner
-  currentTestFilepath = filepath
+  currentTestFilepath = file.filepath
   collectorContext.tasks.length = 0
   defaultSuite.clear()
   collectorContext.currentSuite = defaultSuite


test/core/test/task-names.test.ts
macarie, line 1: This test should most likely never change, so I kept manual assertions for _debuggability_. Could convert to a snapshot or something more automated if necessary.

packages/runner/src/suite.ts
```diff
@@ -495,6 +503,7 @@ function createSuiteCollector(
 
     allChildren.forEach((task) => {
       task.file = file
+      task.fullName = createTaskName([file.fullName, task.fullName])
```
macarie: Not entirely sure about this. `file` seems to be "injected" at a later stage, so to include the file path I had to change all children's `fullName`. Is there a better way of doing this?

packages/runner/src/utils/tasks.ts
```diff
@@ -74,9 +74,13 @@ export function getNames(task: Task): string[] {
 }
 
 export function getFullName(task: Task, separator = ' > '): string {
-  return getNames(task).join(separator)
+  return createTaskName(getNames(task), separator)
 }
 
 export function getTestName(task: Task, separator = ' > '): string {
-  return getNames(task).slice(1).join(separator)
+  return createTaskName(getNames(task).slice(1), separator)
```
sheremet-va: This just adds more loops, let's keep these two as they were

packages/runner/src/suite.ts
sheremet-va: yeah, it seems weird that it works like that, maybe we can add the file sooner?

packages/runner/src/collect.ts
```diff
@@ -42,7 +42,7 @@ export async function collectTests(
 
         runner.onCollectStart?.(file)
 
-        clearCollectorContext(filepath, runner)
+        clearCollectorContext(filepath, runner, file)
```
sheremet-va: We can just pass down file first and get `file.filepath` from it inside the function

packages/runner/src/suite.ts
```diff
@@ -215,9 +217,10 @@ function createDefaultSuite(runner: VitestRunner) {
 export function clearCollectorContext(
   filepath: string,
   currentRunner: VitestRunner,
+  file: File,
 ): void {
   if (!defaultSuite) {
-    defaultSuite = createDefaultSuite(currentRunner)
+    defaultSuite = createDefaultSuite(currentRunner, file)
```
sheremet-va: Maybe it's better to assign it after it's created/read here? (on line 225) If `defaultSuite` already exists in `isolate: false`, I think it will have incorrect `file` at one point

Issue #8583, opened 2025-09-16
Vitest browser tests hang on navigation or form submission


bteng22:
### Describe the bug

When running Vitest in browser mode, tests that trigger navigation or execute the default function of an onSubmit handler will hang indefinitely. This appears to happen because the test browser attempts to navigate away potentially causing the test runner to lose connection with the underlying node process. There doesn't seem to be clear documentation or guidance about how to handle or prevent these hangs in browser-enabled tests.

### Reproduction

External reproduction: https://github.com/bteng22/vitest-browser-hangs-repro

**Steps to reproduce:**

1. Clone the reproduction repository: https://github.com/bteng22/vitest-browser-hangs-repro
2. Follow the instructions in the README to install dependencies and run the browser tests.
3. Observe that when a navigation is triggered (via e.g. a form's default onSubmit), the test hangs and never completes.

**Actual behaviour:**

Vitest browser tests hang forever when navigation or a default onSubmit handler is triggered. The test never completes and requires manual intervention to stop.

**Expected behaviour:**

Vitest browser tests should not hang or lose connection with the node process when navigation or form submission occurs. There should be clear documentation or built-in handling for these situations, or at least guidance on best practices for preventing such hangs in browser-enabled tests? 

### System Info

```shell
System:
    OS: macOS 15.6.1
    CPU: (16) arm64 Apple M3 Max
    Memory: 322.59 MB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 23.7.0 - ~/.nvm/versions/node/v23.7.0/bin/node
    npm: 10.9.2 - ~/.nvm/versions/node/v23.7.0/bin/npm
  Browsers:
    Chrome: 140.0.7339.133
    Edge: 140.0.3485.66
    Safari: 18.6
  npmPackages:
    @vitejs/plugin-react: ^4.7.0 => 4.7.0 
    @vitejs/plugin-react-swc: ^3.11.0 => 3.11.0 
    @vitest/browser: 3.2.4 => 3.2.4 
    @vitest/ui: 3.2.4 => 3.2.4 
    playwright: 1.54.2 => 1.54.2 
    vite: 7.0.6 => 7.0.6 
    vitest: 3.2.4 => 3.2.4 
    vitest-browser-react: 1.0.1 => 1.0.1
```

### Used Package Manager

npm

### Validations

[...]

PR #9095, opened 2025-11-24
doc: A detailed guide on how browser-mode works

goldbergyoni:
**TL&DR:** After a few talks with Vladimir who was kind to explain how browser-mode works, I crafted this guide to help developers make thougful decisions. Happy to improve based on your feedback

### Description

  This PR improves the documentation for Vitest Browser Mode's "How It Works"
   guide to make it clearer and better organized.

  **Changes made:**
  - Clarified that Browser Mode can be used for any kind of webpage testing,
  not just component-level tests
  - Improved explanation of performance characteristics to be more accurate
  - Enhanced description of Browser Mode capabilities (visual page testing +
  unit tests)
  - Reorganized content by blending the "Where Tests Actually Run" section
  into the Key Players section for better flow
  - Expanded explanation of how realistic browser events work (low-level API
  vs JS-space DOM clicks)

  No related issue - this is a documentation improvement to enhance clarity
  and organization.

  ### Please don't delete this checklist! Before submitting the PR, please
  make sure you do the following:
  - [x] It's really useful if your PR references an issue where it is
  discussed ahead of time. If the feature is substantial or introduces
  breaking changes without a discussion, PR might be closed.
    - *N/A - Documentation-only changes to improve clarity*
  - [x] Ideally, include a test that fails without this PR but passes with
  it.
    - *N/A - Documentation-only changes*
  - [x] Please, don't make changes to `pnpm-lock.yaml` unless you introduce a
   new test example.
    - *No changes to pnpm-lock.yaml*
  - [x] Please check [Allow edits by maintainers](https://docs.github.com/en/
  pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-
  changes-to-a-pull-request-branch-created-from-a-fork) to make review
  process faster. Note that this option is not available for repositories
  that are owned by Github organizations.

  ### Tests
    - *N/A - Documentation-only changes, no code changes to test*

  ### Documentation
  - [x] If you introduce new functionality, document it. You can run
  documentation with `pnpm run docs` command.
    - *This PR only modifies existing documentation*

  ### Changesets
  - [x] Changes in changelog are generated from PR name. Please, make sure
  that it explains your changes in an understandable manner. Please, prefix
  changeset messages with `feat:`, `fix:`, `perf:`, `docs:`, or `chore:`.
    - *Suggested PR title: `docs: improve clarity and organization of browser
   mode "how it works" guide`*

### DIFF
--- a/docs/guide/browser/how-it-works.md
+++ b/docs/guide/browser/how-it-works.md
@@ -47,6 +47,11 @@ The attached diagram shows how the main thread's worker pool opens the browser u
 2. The main thread asks the orchestrator to start running a specific test suite
 3. The orchestrator instructs the test runner to execute the file
 4. The test runner processes each test and executes it
+
+::: warning
+Since the test runner is executed within the same page along with the code under test (e.g., a component or a page), should the code tries to navigate away to another page - The text execution will get interrupted and never end with a successful result. Make sure to prevent page refresh and navigation, for example, by stubbing your router navigation functions
+:::
+
 5. If a test renders a component, it appears inside an anchor element placed in the page
 6. Finally, the test reports success or failure back to the main thread via the orchestrator
 


docs/guide/browser/how-it-works.md
```diff
@@ -0,0 +1,152 @@
...
+
+# How Browser Mode Works
+
+## Why This Guide Is Important
+
+This article explains the architecture, internals, and workflow of Vitest's Browser Mode. Whether you're a user wanting to understand how your tests execute or a contributor exploring the codebase, this guide covers both high-level concepts and technical implementation details.
+
+## The Philosophy
+
+Vitest Browser Mode can be used for any kind of webpage testing (e.g., also end-to-end tests), though its unique characteristics truly shine in page-level or component testing. Here, you can achieve both realism with a real browser and exceptional developer experience with performance.
```
sheremet-va:
```suggestion
Vitest Browser Mode can be used for any kind of webpage testing, though its unique characteristics truly shine in page-level or component testing. Here, you can achieve both realism with a real browser and exceptional developer experience with performance.
```

Technically true if you really try, but I don't want to have this mentioned in the documentation to avoid making any promises.

docs/guide/browser/how-it-works.md
```diff
@@ -0,0 +1,152 @@
...
+
+This guide helps you find the sweet spot between visual and realistic tests and performant, developer-friendly testing. Understanding the mechanisms and configuration options explained here will help you tune your setup to meet your testing goals.
+
+## The Key Players {#key-players}
+
+- **The Node.js process** - This is where the CLI is invoked. It performs common setup tasks like filtering test files and orchestrates a pool of test workers where actual execution happens.
+
+- **The provider** - Responsible for instantiating and interacting with a real browser using specific technologies like Playwright, WebDriver I/O, or custom implementations. This is configurable. For simplicity, the rest of this guide assumes Playwright usage, but most concepts apply to every provider.
```
sheremet-va:
```suggestion
- **The provider** - Responsible for instantiating and interacting with a real browser using specific technologies like Playwright, WebdriverIO, or custom implementations. This is configurable. For simplicity, the rest of this guide assumes Playwright usage, but most concepts apply to every provider.
```

docs/guide/browser/how-it-works.md
```diff
@@ -0,0 +1,152 @@
...
+![Browser Mode Initialization Flow](/vitest-browser-initializing-test-suite.png)
+
+### Typical Action Flow {#typical-action-flow}
+
+The diagram illustrates the following flow:
+
+1. When a test needs to perform a user interaction (like a click), the test runner wants to trigger a realistic event using the browser's low-level driver
+2. However, it lacks permission to access the browser's low-level API directly, so it sends a message via **WebSocket** to the orchestrator in the Node.js main thread
+3. The main thread receives the message and sends a command to the provider (in this example, Playwright)
+4. The provider sends a **CDP message** (Chrome DevTools Protocol)—or a WebDriver message if using WebDriver I/O—to the browser
```
sheremet-va:
```suggestion
4. The provider sends a **CDP message** (Chrome DevTools Protocol)—or a WebDriver message if using WebdriverIO—to the browser
```

docs/guide/browser/how-it-works.md
```diff
@@ -0,0 +1,152 @@
...
+
+```typescript
+export default defineConfig({
+  test: {
+    isolate: false, // Share iframe between test files
+  },
+})
+```
+
+By default, test files execute sequentially, one after another, not concurrently. While this reduces performance, it provides better isolation so each test file can assume no other tests are running simultaneously in the same browser.
```
sheremet-va, line 135: This is true only for headed mode. Headless mode runs in parallel by default.

Also worth mentioning that file parallelism works only for playwright provider

docs/guide/browser/how-it-works.md
```diff
@@ -0,0 +1,152 @@
...
+  test: {
+    fileParallelism: true, // Run test files in parallel
+  },
+})
+```
+
+::: warning Interactive APIs and Parallelism
+From the [official Vitest documentation](https://v1.vitest.dev/config/#browser-fileparallelism):
+
+"This makes it impossible to use interactive APIs (like clicking or hovering) because there are several iframes on the screen at the same time, but if your tests don't rely on those APIs, it might be much faster to just run all of them at the same time."
```
sheremet-va, line 150: This is not relevant anymore (the link leads to a v1 implementation that's completely different from v4)

docs/guide/browser/how-it-works.md
goldbergyoni: 100% agree

docs/guide/browser/how-it-works.md
goldbergyoni: I'll clarify. To ensure I understand correctly, with Playwright+headless, the node.js process sends actions CDP to each IFrame, this functionality is now stable. Correct? I assume this is somehow happening by sending the frmae identifier along

docs/guide/browser/how-it-works.md
goldbergyoni: Removing. Just waiting for one small clarification (above), then I'll send an improved version

docs/guide/browser/how-it-works.md
sheremet-va: Yes, it sends actions to each iframe that are opened in [different pages](https://playwright.dev/docs/api/class-browsercontext). The amount of pages is configured with `maxWorkers`

userquin: Uhmmm, you don't get paralelism opening pages in the same browser instance, the browser will reuse the main thread (same origin); we get parallelism when we execute multiple browser instances and send/distribute tests to the corresponding iframe test in the corresponding browser instance (this is why now headless mode is fastest => in previous browser mode versions we didn't distribute the test load between multiple browser instances IIRC, that's why it was fastest running browser mode with te browser opened ); when you say `pages`, did you mean `pages in browser instances`? Maybe I'm wrong 🙏

sheremet-va: > Uhmmm, you don't get paralelism opening pages in the same browser instance, the browser will reuse the main thread (same origin); we get parallelism when we execute multiple browser instances and send/distribute tests to the corresponding iframe test in the corresponding browser instance (this is why now headless mode is fastest => in previous browser mode versions we didn't distribute the test load between multiple browser instances IIRC, that's why it was fastest running browser mode with te browser opened ); when you say `pages`, did you mean `pages in browser instances`? Maybe I'm wrong 🙏

I linked the context page: https://playwright.dev/docs/api/class-browsercontext 

We open different browser contexts and a single page within them and that's how we get parallelism _WITHIN_ a single instance (test files in a single chomium browser run in parallel).

goldbergyoni: Interesting, TIL about renderer reuse within same origin. I'm asking for last clarification as I'd like to ensure to put the right wording, and having this page serve you well while responding to issues that lack a basic understanding of how things are supposed to work (that's a sophisticated machine...) 

1. config.maxWorkers -> How many instances (browsers) will get opened
2. config.fileParallelism -> Parallelize suites execution _within a worker_  which get isolated using a browser context (Playwright only)
3. config.Isolate -> Fresh suite IFrame within the same browser context
4. Parallelize tests -> Not possible by design, also Playwright does support 'fullyParallelized'

So having a.test.ts, b.test.ts, c.test.ts, and maxWorkers=2, fileParallelism=true will result in:

```
                    Vite/Vitest Test Runner
                             |
           -----------------------------------------
           |                                       |
      Browser 1                               Browser 2
    (Worker / page)                         (Worker / page)
           |                                       |
   -------------------                           -------
   |                 |                           |     |
a.test.ts       b.test.ts                    c.test.ts
 (spec)           (spec)                       (spec)
 running        running                       running
 in parallel    in parallel                    alone

```

docs/guide/browser/how-it-works.md
```diff
@@ -0,0 +1,157 @@
...
+### Initializing a Test Flow {#initializing-test-flow}
+
+The attached diagram shows how the main thread's worker pool opens the browser using one of the providers (in this example, Playwright). Here's what happens:
+
+1. The worker pool requests the provider to render the tester page, which contains the test orchestrator and a test runner inside an iframe
+2. The main thread asks the orchestrator to start running a specific test suite
+3. The orchestrator instructs the test runner to execute the file
+4. The test runner processes each test and executes it
+
+::: warning
```
goldbergyoni, line 51: @sheremet-va Note I added here, lesson learned through blood 😆

sheremet-va: This should actually no longer be an issue in Vitest 4 - see https://github.com/vitest-dev/vitest/issues/8583#issuecomment-3312083649

Maybe it's still an issue if another page is open, but iframe is not refreshed? It would be a bug then.

docs/guide/browser/how-it-works.md
sheremet-va: I prefer to explain in playwright terms since it's the only provider that supports parallelisation in Vitest

1. `config.maxWorkers` translates to how many `browser.newContext()` will be opened:
```ts
const playwright = require('playwright')
const chromium = await playwright.chromium.launch()
for (const worker of maxWorkers) {
  const context = await chromium.newContext()
  const page = await context.newPage()
  await page.goto(iframeTester)
}
```
2. `config.fileParallelism` just allows `maxWorkers` to be higher than 1, otherwise it always defaults to 1
3. If `config.isolate` is `false`, then the iframe will be destroyed when the test ends. If not, the iframe will still be opened and worker will just reuse it to start another test. If parallelisation is enabled, there could be more than 1 worker. Vitest waits until one is not busy to start a new test
4. Not sure what you mean. They follow the same rules that any same origin parallelisation would follow. I think timers for example stop execution in other workers until they are over

Graph:
No, that is not how it would work. One worker runs one test at a time. There is one iframe in one page in a single context/worker.

userquin: So, I was right , you can't run tests in parallel on the same browser instance. The only way to achieve parallelism is through workers (browser instances), and it's not related to parallelism within the browser. You were referring to parallelism from the perspective of vitest + playwright, and I was referring to parallelism within the browser, which can't be done because of how browsers work and how they "share" the main thread when we're on the same origin:
- [Out-of-Process iframes (OOPIFs)](https://www.chromium.org/developers/design-documents/oop-iframes/)
- [V8BindingDesign](https://chromium.googlesource.com/chromium/src/+/master/third_party/blink/renderer/bindings/core/v8/V8BindingDesign.md)

docs/guide/browser/how-it-works.md
goldbergyoni: My experience was: page under test tried to navigate away, the test hung forever

Wouldn't the best case would be a test failure ("Found a way to throw an error in this case")? If yes, we should explain this

docs/guide/browser/how-it-works.md
goldbergyoni: @sheremet-va Thanks for your explanations, always patient. See some clarifications below

1. Do we ever start more than 1 browser instance (not context, but rather a whole browser)? To contrast for the sake of understanding, in Playwright each worker process launches its own browser instance
3. When you write "...then the iframe will be destroyed when the test ends...", do you mean test **file**?
4. Playwright can [parallelize tests](https://playwright.dev/docs/api/class-testproject#test-project-fully-parallel) (not files), can we as well? Not that I care what Playwright does, using this as a reference to make my questions clearer

docs/guide/browser/how-it-works.md
sheremet-va: There is an error if iframe's url changes. I need a reproduction, it shouldn't hang.

goldbergyoni: I'll put aside my personal case which might have hit an edge case, let's stick to the core idea: The system under test shouldn't navigaye and expect the test to pass. I'll phrase this nicely. Agree?

sheremet-va: Yes, this sounds reasonable.

docs/guide/browser/how-it-works.md
sheremet-va: 1. There is one browser per [instance](https://vitest.dev/config/browser/instances.html). If you have multiple instances that specify the same `browser`, then will be opened in different "browsers"
2. Yes, I mean when all tests in a file end
3. No, we cannot because Playwright runs tests in Node, and we run tests in the browser. If we did, then we would have to import the _test file_ for every _test_ separately

PR #9132, opened 2025-11-30
fix(browser): run `toMatchScreenshot` only once when used with `expect.element`

### LINKED ISSUE #8959: Infinite loop of screenshot creation on `toMatchScreenshot` failure
### Describe the bug

Vitest starts creating reference screenshots in an infinite loop when my test fails.

Expected vitest to take a single screenshot on `toMatchScreenshot` failure.

### Reproduction
https://github.com/Romej/vitest-browser-infinite-screenshots#

Image

### System Info

```shell
System:
    OS: macOS 15.7
    CPU: (14) arm64 Apple M4 Pro
    Memory: 3.52 GB / 48.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.19.0
    npm: 10.9.3
    pnpm: 10.17.1
```

### Used Package Manager

npm

### Validations

macarie:
### Description

Fixes #8959

### Tests

### Documentation

### Changesets

### DIFF
--- a/test/browser/fixtures/expect-dom/toMatchScreenshot.test.ts
+++ b/test/browser/fixtures/expect-dom/toMatchScreenshot.test.ts
@@ -527,8 +527,8 @@ describe('.toMatchScreenshot', () => {
 
       const elapsed = performance.now() - start
 
-      // Allow some buffer for screenshot capture + diffing
-      expect(elapsed).toBeLessThan(renderDelay + 500)
+      // Elapsed time should be lower than the default `poll`/`element` timeout
+      expect(elapsed).toBeLessThan(30_000)
     },
   )
 })


packages/vitest/src/integrations/chai/poll.ts
```diff
@@ -69,12 +69,36 @@ export function createExpectPoll(expect: ExpectStatic): ExpectStatic['poll'] {
...
+                try {
+                  resolve(await assertionFunction.call(assertion, ...args))
+                }
+                catch (err) {
+                  if (chai.util.flag(assertion, '_poll.assert_once')) {
+                    clearTimeout(intervalId)
+                    clearTimeout(timeoutId)
+
+                    rejectWithCause(err)
+                  }
```
macarie: Not entirely convinced by this solution, it adds _yet-another_ exit point to a function that already has many...

macarie:
Pushed a refactor of the poll assertion logic from the nested timer callbacks to a linear async/await loop.

The control flow was quite hard to follow, there were multiple timeouts to track and exit points across different scopes. This also fixes a potential race condition where `check` could run concurrently when the timeout fired while the previous call was mid-`await`.

The behavior is _essentially_ unchanged, the only difference is the last attempt now waits `interval` before running and the assertion succeeds if it is successful (could keep this behavior, but there were edge cases in which `lastError` could still be `undefined`, thus throwing when attempting to set `cause`).

Happy to revert if there are concerns.

test/browser/fixtures/expect-dom/toMatchScreenshot.test.ts
```diff
@@ -421,4 +421,114 @@ describe('.toMatchScreenshot', () => {
...
+          <actual>
+
+        Diff image:
+          <diff>
+      `)
+
+      const elapsed = performance.now() - start
+
+      // Allow some buffer for screenshot capture + diffing
+      expect(elapsed).toBeLessThan(renderDelay + 500)
```
macarie: Locally, 500ms are enough for all browsers to finish. Might be worth monitoring on CI to see if it's flaky.

test/browser/fixtures/expect-dom/toMatchScreenshot.test.ts
```diff
@@ -421,4 +421,114 @@ describe('.toMatchScreenshot', () => {
...
+
+      let errorMessage: string
+
+      try {
+        await expect.element(locator).toMatchScreenshot()
+      } catch (error) {
+        errorMessage = error.message
+      }
+
+      expect(typeof errorMessage === 'string')
```
sheremet-va: This doesn't do anything, did you mean `expect.assert`? We should maybe throw an error if the methods are not called 😄

macarie: Editing fail on my part, it was an `if` before but forgot to change `===` with `).toBe(` 😅

Good catch, thanks!

test/browser/fixtures/expect-dom/toMatchScreenshot.test.ts
macarie: Looks like it's failing in CI. Maybe I can just check `< 30_000` since that's poll's default timeout.

sheremet-va (approved):

PR #9188, opened 2025-12-05
feat(browser): add `userEvent.wheel` API

### LINKED ISSUE #6936: Implement `userEvent.wheel` in interactivity api

Allow a test to scroll the wheel in browser mode. We want to use wheel within a number input to change the value.

### Suggested solution

Hook up to Playwright's [wheel](https://playwright.dev/docs/api/class-mouse#mouse-wheel)

### Alternative

_No response_

### Additional context

See https://github.com/testing-library/user-event/issues/475 for existing feature request on testing-library/user-event.

### Validations

macarie:
### Description

Resolves #6936

A couple of notes regarding the API design decision.

Playwright and WebDriverIO both use something like `await mouse.wheel(deltaX, deltaY)`. Their approach isn't tied to any element. Playwright recommends hovering over an element first, then calling this method to trigger a scroll event on it → initial attempt was following that approach with the addition of automatic hover, but this feels _very_ low level.

Another thing worth considering is that one call equals to one event. Usually, testing these interactions require more than that and waiting for a full client-server-client loop, depending on the environment, might be slow.

Landed on something that should provide enough flexibility while keeping a high-level API:

```ts
type ScrollOptions = (
  | { direction: 'up' | 'down' | 'left' | 'right' }
  | { delta: { x?: number, y?: number } }
) & { times?: number }

wheel(element: Element, options: ScrollOptions): Promise<void>
wheel(options: ScrollOptions): Promise<void>
```

> [!NOTE]
> The initially planned element-less version (`wheel(options)`) was not implemented due to behavior differences between WebDriverIO and Playwright. I'll open a feature request issue to gauge community interest for this addition.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/browser/context.d.ts
+++ b/packages/browser/context.d.ts
@@ -561,7 +561,7 @@ export interface Locator extends LocatorSelectors {
    */
   tripleClick(options?: UserEventTripleClickOptions): Promise<void>
   /**
-   * Triggers a wheel event on an element.
+   * Triggers a {@link https://developer.mozilla.org/en-US/docs/Web/API/Element/wheel_event|`wheel` event} on an element.
    *
    * @param options - Scroll configuration using `delta` or `direction`.
    * @returns A promise that resolves when all wheel events have been dispatched.


macarie:
Looks like there's a huge difference in running times between Firefox and Chrome in WebdriverIO:
```
 ✓  chrome  wheel.test.ts (9 tests) 11175ms
   ✓ fires wheel event multiple times when `times` option is set  10266ms (repeat x50)
 ✓  firefox  wheel.test.ts (9 tests) 1049ms
   ✓ fires wheel event multiple times when `times` option is set  966ms (repeat x50)
```
In Playwright all timings are roughly aligned, but all are ~4x slower than Firefox in WebdriverIO:
```
 ✓  webkit  wheel.test.ts (9 tests) 4422ms
   ✓ fires wheel event multiple times when `times` option is set  4080ms (repeat x50)
 ✓  firefox  wheel.test.ts (9 tests) 4460ms
   ✓ fires wheel event multiple times when `times` option is set  4101ms (repeat x50)
 ✓  chromium  wheel.test.ts (9 tests) 4992ms
   ✓ fires wheel event multiple times when `times` option is set  4591ms (repeat x50)
```

packages/browser/src/client/tester/context.ts
```diff
@@ -69,6 +109,32 @@ export function createUserEvent(__tl_user_event_base__?: TestingLibraryUserEvent
...
+        const browser = getBrowserState().config.browser.name
+
+        // looks like on Chromium the scroll event gets dispatched a frame later
+        if (browser === 'chromium' || browser === 'chrome') {
+          return new Promise((resolve) => {
+            requestAnimationFrame(() => {
+              resolve()
+            })
+          })
+        }
```
macarie: Not sure if this branching is worth it or if we should just resolve the promise after a frame on all browsers.

packages/browser/src/client/tester/tester-utils.ts
```diff
@@ -217,3 +217,41 @@ const kLocator = Symbol.for('$$vitest:locator')
 export function isLocator(element: unknown): element is Locator {
   return (!!element && typeof element === 'object' && kLocator in element)
 }
+
+const DEFAULT_WHEEL_DELTA = 100
+
+export function resolveUserEventWheelOptions(options: UserEventWheelOptions): UserEventWheelOptionsWithDelta {
```
macarie: Not entirely sure whether this is the best place for this function.

packages/browser/src/client/tester/locators/index.ts
```diff
@@ -86,6 +87,23 @@ export abstract class Locator {
...
+      const browser = getBrowserState().config.browser.name
+
+      // looks like on Chromium the scroll event gets dispatched a frame later
+      if (browser === 'chromium' || browser === 'chrome') {
+        return new Promise((resolve) => {
+          requestAnimationFrame(() => {
+            resolve()
+          })
+        })
+      }
```
macarie, line 103: As said in another comment, I'm not sure if this branching is worth it or if we should just resolve the promise after a frame on all browsers.

The tricky part is that _sometimes_, on resource-constrained machines (~CI), one frame might not be enough. From my testing this happens rarely tho, maybe a couple of times in 50 runs (with `repeats`). A "proper" solution would be waiting for the event to fire with a listener, but we risk not resolving the promise if user-code intercepts and stops the propagation of the event.

packages/browser/context.d.ts
```diff
@@ -344,6 +362,43 @@ export interface UserEventDoubleClickOptions {}
...
+    direction: 'up' | 'down' | 'left' | 'right',
+    /**
+     * Number of wheel events to fire. Defaults to `1`.
+     *
+     * Useful for simulating multiple scroll steps in a single call.
+     */
+    times?: number,
+  },
+  'delta' | 'direction'
+>
```
macarie: I can rewrite this type without using any helpers, but thought they might be useful for other types too.

docs/api/browser/interactivity.md
```diff
@@ -162,6 +162,54 @@ References:
 - [WebdriverIO `browser.action` API](https://webdriver.io/docs/api/browser/action/): implemented via actions api with `move` plus three `down + up + pause` events in a row
 - [testing-library `tripleClick` API](https://testing-library.com/docs/user-event/convenience/#tripleClick)
 
+## userEvent.wheel
```
sheremet-va:
```suggestion
## userEvent.wheel <Version>4.1.0</Version> {#userevent-wheel}
```

docs/api/browser/interactivity.md
```diff
@@ -162,6 +162,54 @@ References:
...
+
+```ts
+function wheel(
+  element: Element | Locator,
+  options: UserEventWheelOptions,
+): Promise<void>
+```
+
+Triggers a wheel event on an element. This is useful for testing any UI that responds to wheel events.
```
sheremet-va:
```suggestion
Triggers a [wheel event](https://developer.mozilla.org/en-US/docs/Web/API/Element/wheel_event) on an element. This is useful for testing any UI that responds to wheel events.
```

docs/api/browser/interactivity.md
```diff
@@ -162,6 +162,54 @@ References:
...
+
+You can specify the scroll amount using either `delta` for precise pixel-based control, or `direction` for simpler directional scrolling (`up`, `down`, `left`, `right`). When you need to trigger multiple wheel events, use the `times` option rather than calling the method multiple times.
```
sheremet-va:
```suggestion
You can specify the scroll amount using either `delta` for precise pixel-based control, or `direction` for simpler directional scrolling (`up`, `down`, `left`, `right`). When you need to trigger multiple wheel events, use the `times` option rather than calling the method multiple times for better performance.
```

docs/api/browser/interactivity.md
```diff
@@ -162,6 +162,54 @@ References:
...
+
+  // Scroll right 5 times
+  await userEvent.wheel(tablist, { direction: 'right', times: 5 })
+
+  // Scroll left once
+  await userEvent.wheel(tablist, { direction: 'left' })
+})
+```
+
+Wheel events can also be triggered directly from locators:
```
sheremet-va:
```suggestion
Wheel events can also be triggered directly from [locators](/api/browser/locator#wheel):
```

docs/api/browser/locators.md
```diff
@@ -648,6 +648,23 @@ await page.getByRole('img', { name: 'Rose' }).tripleClick()
 
 - [See more at `userEvent.tripleClick`](/api/browser/interactivity#userevent-tripleclick)
 
+### wheel
```
sheremet-va:
```suggestion
### wheel <Version>4.1.0</Version>
```

docs/api/browser/locators.md
```diff
@@ -648,6 +648,23 @@ await page.getByRole('img', { name: 'Rose' }).tripleClick()
...
+
+```ts
+function wheel(options: UserEventWheelOptions): Promise<void>
+```
+
+Triggers a wheel event on an element. You can use the options to choose a general scroll `direction` or a precise `delta` value.
```
sheremet-va:
```suggestion
Triggers a [wheel event](https://developer.mozilla.org/en-US/docs/Web/API/Element/wheel_event) on an element. You can use the options to choose a general scroll `direction` or a precise `delta` value.
```

packages/browser/context.d.ts
sheremet-va: Honestly, to me this just looks like overengineering, there are just two unions, it might be easier (for readability) to just define them

docs/api/browser/interactivity.md
macarie: Good suggestion, I'll add the link in the JSDocs too.

I think I should also add to not use this for general scrolling but rely on automatic scrolling other `userEvent` methods have built-in.

macarie:
Applied the suggestions and split the type (used different interfaces as I was extracting a subtype anyway for the server function)

Issue #7354, opened 2025-01-24
How should I configure my custom reporter in Vitest UI?


zhangtao25:
### Clear and concise description of the problem

This is our custom HTML reporter, and I want to use it in the coverage section of Vitest UI.

![Image](https://github.com/user-attachments/assets/be19d353-61c9-4189-96dd-5e06bd18d52d)

### Suggested solution

It is recommended to configure the coverage reporter for coverage reports.

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vitest.dev/guide/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

AriPerkkio:
This is not possible. The `html` reporter is hard-coded here:

https://github.com/vitest-dev/vitest/blob/9e404375825d870ed9d90db854d96f77a7abeaff/packages/ui/node/index.ts#L67-L102

https://github.com/vitest-dev/vitest/blob/9e404375825d870ed9d90db854d96f77a7abeaff/packages/ui/node/index.ts#L30-L52

zhangtao25:
When configuring an HTML reporter, if I also configure my own custom HTML reporter, it can overwrite the previous HTML report.

https://github.com/canyon-project/canyon-test/blob/main/vitest.config.ts#L10

![Image](https://github.com/user-attachments/assets/3ee7c7ff-2584-4657-b824-7a82fadafa8d)

AriPerkkio:
That actually works by accident - the reporters are looped in the order they are defined in configuration. If you are saving the reports in same directory as built-in `html` does, it will just overwrite the results.

https://github.com/vitest-dev/vitest/blob/9e404375825d870ed9d90db854d96f77a7abeaff/packages/coverage-v8/src/provider.ts#L129-L138

I think we should have a better solution for this. Maybe something like

```ts
import { defineConfig } from "vitest/config";

export default defineConfig({
  test: {
    ui: {
      coverageDirectory: "/somepath"
    }
  },
});
```

... and if that's defined, use it instead of trying to resolve the path from coverage options. 
The current resolving logic should be used when `coverageDirectory` is not defined for backwards compatibility.

sheremet-va:
[...]

PR #9282, opened 2025-12-17
feat: conditional judgment for deleting the "html" reporter

travzhang:
### Description

This PR removes the hard dependency on the built-in `html` coverage reporter when resolving the coverage HTML folder exposed by the Vitest API server.

Previously, the coverage UI was only enabled when the `html` reporter was explicitly configured. This assumption prevents custom or third-party coverage reporters from exposing valid HTML coverage output through the Vitest server, even when they generate static HTML reports.

With this change, Vitest determines whether to expose the coverage folder based on coverage configuration and resolved output paths, rather than relying on the reporter name. This makes the behavior more flexible and allows custom HTML-based coverage reporters to integrate with the Vitest UI without needing to masquerade as the built-in `html` reporter.

截屏2025-12-17 19 27 51

---

### Additional Context

This change is backward compatible with the existing `html` reporter behavior and does not alter the default coverage directory structure. It only relaxes the reporter-type constraint to improve extensibility for custom coverage reporters.

---

### Tests

* [x] Run the tests with `pnpm test:ci`.

---

### Checklist

* [ ] This PR references a related issue discussing the behavior change.
* [ ] Tests are added or updated to cover the new behavior.
* [x] No unnecessary changes to `pnpm-lock.yaml`.
* [x] Allow edits by maintainers is enabled.

### DIFF
--- a/packages/ui/node/index.ts
+++ b/packages/ui/node/index.ts
@@ -2,7 +2,6 @@ import type { Plugin } from 'vite'
 import type { Vitest } from 'vitest/node'
 import fs from 'node:fs'
 import { fileURLToPath } from 'node:url'
-import { toArray } from '@vitest/utils/helpers'
 import { basename, resolve } from 'pathe'
 import sirv from 'sirv'
 import c from 'tinyrainbow'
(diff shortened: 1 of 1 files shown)


AriPerkkio (commented):
This closes https://github.com/vitest-dev/vitest/issues/7354, right?

travzhang:
> This closes #7354, right?

Yes, but this closed issue still fails to fix the problem.

PR #9297, opened 2025-12-18
feat: interactive snapshot update

### LINKED ISSUE #2229: Interactive snapshot update/create mode

We are testing media assets, by matching their md5 basically. So we have a suit, that basically calls the same render function but in different contexts (projects, parameters). Basically, the output must be validated before confirming, that this md5 should be a snapshot.
Jest has an interactive mode, but async describe is not a thing anymore, so we tried to switch to Vitest. 

### Suggested solution

1. Maybe do not create snapshots by default (I know, that env CI=1 solves it)
2. Have interactive update/create mode, that does 1 test at a time and asks to validate output

### Alternative

I could implement it myself but have no way to interrupt runner and it just times out. Also, libraries such as `enquirer` do not work during the run.

### Additional context

_No response_

### Validations

mzhubail:
### Description

This adds a cli option, that would go through failed snapshots and display them one by one, prompting user to validate output.

Resolves #2229

image

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/stdin.ts
+++ b/packages/vitest/src/node/stdin.ts
@@ -275,8 +275,6 @@ export function registerConsoleShortcuts(
 
     const specs: TestSpecification[] = []
     for (const test of tests) {
-      ctx.logger.clearFullScreen()
-
       const snapshotErrors = test.result?.errors?.filter(
         e => typeof e?.message === 'string'
           && e.message.match(/Snapshot .* mismatched/),
@@ -325,7 +323,6 @@ export function registerConsoleShortcuts(
       }
     }
     else {
-      ctx.logger.clearFullScreen()
       // TODO: This is a hack to trigger report summary again
       ctx.report('onTestRunEnd', testModules, errors as SerializedError[], state)
     }


mzhubail:
This is still an early PR.  I would like to check two things before I proceed any further.

1. For test running I've used `project.createSpecification` with `ctx.rerunTestSpecifications` to create test specs and run them.  The obvious downside of this is that this would only work if we have `includeTaskLocation: true` in our config.

    I'm not sure how to move forward from this of if there's an API that's a better fit.

2. When selecting snapshots, ~I clear screen between snapshots to separate snapshots~ if user ends up not selecting any snapshot (skips on all or quits `q`), I print out the `onTestRunEnd` to get the user to a view similar to that before the running the interactive snapshot update.

    However, the current implementation doesn't include suite / test summary and feels overall a bit too hacky.  Any suggestions on how to handle this would be much welcomed.

    https://github.com/vitest-dev/vitest/blob/395d2a7fe33cb777ab9c5b5184061b70782eeb0d/packages/vitest/src/node/stdin.ts#L327-L331

PR #9310, opened 2025-12-20
Use `meta.url` as the argument to `createRequire()` when loading a module

penalosa:
### Description

When creating a `require()` function using `createRequire()`, pass in the URL rather than the filename. Passing in the filename can cause bugs on Windows, and error messages along the lines of `TypeError: The argument 'path' The argument must be a file URL object, a file URL string, or an absolute path string.. `

This matches the implementation in `vite-node`: https://github.com/antfu-collective/vite-node/blob/63e2a8111cb674335823746a5df87eae666eee76/src/client.ts#L550

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/runtime/moduleRunner/moduleEvaluator.ts
+++ b/packages/vitest/src/runtime/moduleRunner/moduleEvaluator.ts
@@ -273,7 +273,7 @@ export class VitestModuleEvaluator implements ModuleEvaluator {
       'code.file.path': filename,
     })
 
-    const require = this.createRequire(filename)
+    const require = this.createRequire(meta.url)
 
     const argumentsList = [
       ssrModuleExportsKey,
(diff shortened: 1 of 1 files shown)


sheremet-va:
This seems to break some tests. I don't think the `isAbsolute` check there is correct now

packages/vitest/src/runtime/moduleRunner/moduleEvaluator.ts
```diff
@@ -355,7 +355,7 @@ export class VitestModuleEvaluator implements ModuleEvaluator {
   private createRequire(filename: string) {
     // \x00 is a rollup convention for virtual files,
     // it is not allowed in actual file names
-    if (filename[0] === '\x00' || !isAbsolute(filename)) {
+    if ((filename[0] === '\x00' || !isAbsolute(filename))) {
```
petebacondarwin, line 358: why the extra params here?

PR #9313, opened 2025-12-20
perf: optimise code patterns with modern JavaScript syntax

AlexanderKaran:
### Description

This PR adds `@e18e/eslint-plugin` to modernise the codebase with performance-optimised JavaScript patterns.

I did not open an issue to discuss first as it seemed like such a small change but super happy to close the PR and open an issue. Sorry.

**Changes applied:**
- Use spread syntax (`[...arr]`) instead of `Array.from(arr)` where no mapper is used
- Use `Array.from(arr, fn)` instead of `[...arr].map(fn)` to avoid intermediate arrays
- Use modern array methods (`.at()`, `.toReversed()`)
- Use nullish coalescing operator (`??`) instead of verbose null checks
- Use `Object.hasOwn()` instead of `hasOwnProperty`

**Plugin rules disabled:**
- `e18e/ban-dependencies` - Disabled to avoid suggesting package replacements as it suggests removing Axios
- `e18e/prefer-timer-args` - Disabled due to incorrect auto-fixes. I can fix this up in its own PR if wanted

All tests pass and no functionality changes were introduced - only performance optimisations.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/browser/src/client/vite.config.ts
+++ b/packages/browser/src/client/vite.config.ts
@@ -54,7 +54,10 @@ export default vite.defineConfig({
         const uiEntryPoint = resolve(ui, 'index.html')
         const browser = resolve(root, 'browser/dist/client/__vitest__/')
 
-        const timeout = setTimeout(console.log, 1000, '[copy-ui-plugin] Waiting for UI to be built...')
+        const timeout = setTimeout(
+          () => console.log('[copy-ui-plugin] Waiting for UI to be built...'),
+          1000,
+        )
         await waitFor(() => fs.existsSync(ui) && fs.existsSync(uiEntryPoint))
         clearTimeout(timeout)
 


packages/browser/src/client/tester/expect/toHaveClass.ts
```diff
@@ -28,11 +28,9 @@ export default function toHaveClass(
   const received = splitClassNames(htmlElement.getAttribute('class'))
   const expected = expectedClassNames.reduce(
```
AriPerkkio, line 29: Instead of creating another array with `[...acc, ...]` (or previously used `.concat`), we could just push to `acc`:

```ts
const expected = expectedClassNames.reduce((acc, className) => {
  const item = typeof className === 'string' || !className
    ? splitClassNames(className)
    : [className];

      acc.push(item);
      return acc;
  },
  [] as (string | RegExp)[],
)
```

AlexanderKaran: @AriPerkkio `splitClassNames` returns an array so we could do `acc.push(...item)` for that one?

packages/browser/src/client/vite.config.ts
```diff
@@ -54,10 +54,7 @@ export default vite.defineConfig({
         const uiEntryPoint = resolve(ui, 'index.html')
         const browser = resolve(root, 'browser/dist/client/__vitest__/')
 
-        const timeout = setTimeout(
-          () => console.log('[copy-ui-plugin] Waiting for UI to be built...'),
-          1000,
-        )
+        const timeout = setTimeout(console.log, 1000, '[copy-ui-plugin] Waiting for UI to be built...')
```
sheremet-va: I prefer creating an inline function in these cases, this has 0 effect on performance

packages/browser/src/node/commands/screenshotMatcher/utils.ts
```diff
@@ -262,7 +262,7 @@ export function asyncTimeout(timeout: number): Promise<null> {
       resolve(null)
     }
     else {
-      setTimeout(() => resolve(null), timeout)
+      setTimeout(resolve, timeout, null)
```
sheremet-va: I think this removes typechecking of resolve now?

packages/browser/src/node/rpc.ts
```diff
@@ -257,23 +257,21 @@ export function setupBrowserRpc(globalServer: ParentBrowserProject, defaultMocke
...
-              contextId: sessionId,
-              sessionId,
-              triggerCommand: (name: string, ...args: any[]) => {
-                return project.browser!.triggerCommand(
-                  name as any,
-                  context,
-                  ...args,
-                )
-              },
+          const context = {
```
sheremet-va, line 260: I think this breaks getter access? Or does assign also do that?

packages/browser-playwright/src/commands/keyboard.ts
```diff
@@ -30,7 +30,7 @@ export const keyboard: UserEventCommand<(text: string, state: KeyboardState) =>
   )
 
   return {
-    unreleased: Array.from(pressed),
+    unreleased: [...pressed],
```
sheremet-va, line 33: I would prefer Array.from because it is more readable. I think we even had a rule at one point that enforced it

sheremet-va: I also could not confirm that spreading is faster

packages/runner/src/fixture.ts
```diff
@@ -321,10 +321,7 @@ function resolveDeps(
     }
     if (depSet.has(fixture)) {
       throw new Error(
-        `Circular fixture dependency detected: ${fixture.prop} <- ${[...depSet]
-          .reverse()
-          .map(d => d.prop)
-          .join(' <- ')}`,
+        `Circular fixture dependency detected: ${fixture.prop} <- ${[...depSet].toReversed().map((d: FixtureItem) => d.prop).join(' <- ')}`,
```
sheremet-va: Honestly, I feel uncomfortable using toReversed  yet. It requires a polyfill and can’t be reduced like syntax can, and it costs us nothing to use the older syntax

AlexanderKaran:
> Out of all of the changes, I feel like only concat makes sense, everything else is just subjective changes for the sake of changes

@sheremet-va so the ESLint rule has three sections modernisation, module replacements and perf. We can turn of rules you don't think apply to Vitest or just can the PR entirly if thats better. Should of opened an issue first to discuss sorry.

The idea (taken from the intro in the rules)

> There are a few categories of rules in this plugin:
> 
> Modernization - New syntax and APIs which improve code readability and performance
> Module replacements - Community recommended alternatives to popular libraries, focused on performance and size
> Performance improvements - Patterns that can be optimized for better runtime performance

> Each of these can be enabled individually, or you can use the recommended configuration to enable all rules.

packages/browser/src/node/rpc.ts
AlexanderKaran: Pretty sure they both invoke but do not copy

packages/runner/src/fixture.ts
AlexanderKaran: Personaly I have no issue with pollyfill but I can turn off the rule and undo the change here

packages/browser-playwright/src/commands/keyboard.ts
AlexanderKaran: This is a modernisation rule, not a performance rule - the primary goal is to use modern JavaScript syntax. That said, we obviously want to ensure we're not making things slower.

For the Array.from(iter) case (without a mapper function), spread syntax is either faster or performance-equivalent across modern engines. If you have benchmarks showing otherwise, I'd be happy to document that edge case back in e18e, but in general the perf impact is neutral-to-positive.

Re: readability - I can see both perspectives. `[...pressed]` is the more idiomatic modern syntax and is widely used in contemporary codebases, but I understand `Array.from()` can be more explicit about intent.

The rule handles 4 patterns:

```
arr.concat(arr2) → [...arr, ...arr2]
Object.assign({}, obj1, obj2) → {...obj1, ...obj2}
fn.apply(null, args) → fn(...args)
Array.from(iter) → [...iter] (when no mapper)
```

For the first 3, spread is cleaner and generally more performant. For `Array.from`, it's about consistency with modern JavaScript patterns.

packages/runner/src/fixture.ts
AlexanderKaran: Done

packages/browser-playwright/src/commands/keyboard.ts
sheremet-va: This is all incredibly subjective. This code base uses `Array.from` and I do not find your argument compelling to enforce something otherwise, even if I agree with them. `arr.concat` is a legitimate performance improvement, for example, even from my experience, so I don't see why not adopt it. 

In general, we just follow https://github.com/antfu/eslint-config which also includes performance and modernisation rules. Every time it needs an update, we also run the `eslint --fix` command. So I just don't see a lot of benefit of adding yet another preset

If I had to change anything, I would rather adopt oxc with oxfmt

AlexanderKaran: No problem at all. Will close the PR.

AlexanderKaran:
Closing PR as Vitest already uses https://github.com/antfu/eslint-config which includes its own perf and modernisation rules

PR #9325, opened 2025-12-22
docs(browser): add render libraries examples

### LINKED ISSUE #9019: Documentation: add `vitest-browser-*` packages to the API section

We should concentrate the documentation about browser mode. Instead of keeping it only in packages README, the API should be documented in our docs.

### Suggested solution

See: https://github.com/vitest-dev/vitest/blob/353ee5bbcb9e1fad3147e067cd8ba1b10ba71da2/docs/.vitepress/config.ts#L538

The comment is currently inside the "Config" category, but it fits better in API

### Alternative

_No response_

### Additional context

_No response_

### Validations

sheremet-va:
### Description

Resolves https://github.com/vitest-dev/vitest/issues/9019

### Tests

### Documentation

### Changesets


docs/api/browser/react.md
```diff
@@ -0,0 +1,334 @@
...
+By default, Vitest will create a `div` and append it to the `document.body` and this is where your component will be rendered. If you provide your own `HTMLElement` container via this option, it will not be appended to the `document.body` automatically.
+
+For example, if you are unit testing a `tablebody` element, it cannot be a child of a `div`. In this case, you can specify a `table` as the render container.
+
+```jsx
+const table = document.createElement('table')
+
+const { container } = await render(<TableBody {...props} />, {
+  container: document.body.appendChild(table),
+})
```
macarie, line 66: > If you provide your own `HTMLElement` container via this option, it will not be appended to the `document.body` automatically.

Maybe the example should include `document.body.appendChild(table)`.

Edit for clarity: I mean using it beneath the declaration of `table`, on its own line, so it's easier to spot, then just `container: table`.

docs/api/browser/react.md
```diff
@@ -0,0 +1,334 @@
...
+const screen = render(<Component />) // [!code --]
+const screen = await render(<Component />) // [!code ++]
+```
+:::
+
+### Options
+
+#### container
+
+By default, Vitest will create a `div` and append it to the `document.body` and this is where your component will be rendered. If you provide your own `HTMLElement` container via this option, it will not be appended to the `document.body` automatically.
```
macarie: What about changing the wording slightly and provide a suggestion on what to do?

```suggestion
By default, Vitest will create a `div`, append it to `document.body`, and render your component there. If you provide your own `HTMLElement` container, it will not be appended automatically — you'll need to call `document.body.appendChild(container)` before calling `render`.
```

docs/api/browser/react.md
```diff
@@ -0,0 +1,334 @@
...
+
+For example, if you are unit testing a `tablebody` element, it cannot be a child of a `div`. In this case, you can specify a `table` as the render container.
```
macarie: The HTML element is just `tbody`

docs/api/browser/react.md
```diff
@@ -0,0 +1,334 @@
...
+
+This method is a shortcut for `console.log(prettyDOM(baseElement))`. It will print the DOM content of the container or specified elements to the console.
+
+### rerender
+
+```ts
+export function rerender(ui: React.ReactNode): Promise<void>
+```
+
+It is better if you test the component that's doing the prop updating to ensure that the props are being updated correctly to avoid relying on implementation details in your tests. That said, if you'd prefer to update the props of a rendered component in your test, this function can be used to update props of the rendered component.
```
macarie, line 157: Feels a bit weird to start with what not to do (altho I understand the sentiment), what about wording it like this?

```suggestion
Re-renders your component with new props.

When possible, prefer testing prop changes through the parent component that passes them. This keeps your tests decoupled from implementation details. That said, when you need to test a component's response to prop changes directly, that's what `rerender` is for.
```

macarie (commented):
I like having these in the main docs, they're easier to discover 😄

Just a couple of small suggestions regarding wording. I believe some of them (the container-related ones) also apply to the Vue page.

docs/api/browser/react.md
sheremet-va: I am confused - it does use `document.body.appendChild(table)`

docs/api/browser/react.md
sheremet-va: It is how testing-library phrases it and I actually prefer it that way to discourage its usage

docs/api/browser/react.md
macarie: Yes, sorry, I updated the comment after submitting, I meant like this:

```ts
const table = document.createElement('table')

document.body.appendChild(table)

const { container } = await render(<TableBody {...props} />, {
  container: table,
})
```

Just to _highlight_ that it's being appended manually, otherwise it might be easy to miss.

sheremet-va: I think adding a note like `// note that it appended manually` would be enough, `appendChild` returns the same node

macarie: That could work too, yes. Just something to draw attention to it, otherwise might be easy to miss from a quick read

docs/api/browser/svelte.md
```diff
@@ -0,0 +1,273 @@
...
+
+For example, if you are unit testing a `tablebody` element, it cannot be a child of a `div`. In this case, you can specify a `table` as the render container.
+
+```ts
+const table = document.createElement('table')
+
+const screen = render(TableBody, props, {
+  // ⚠️ appending the element to `body` manually before rendering
+  container: document.body.appendChild(table),
+})
```
mcous: **issue (multiple sites):** Svelte calls this the `target` option and it is part of component options, not setup options

```suggestion
const screen = render(TableBody, {
  props,
  // ⚠️ appending the element to `body` manually before rendering
  target: document.body.appendChild(table),
})
```

docs/api/browser/svelte.md
```diff
@@ -0,0 +1,273 @@
...
+```
+
+#### baseElement
+
+If the `container` is specified, then this defaults to that, otherwise this defaults to `document.body`. This is used as the base element for the queries as well as what is printed when you use `debug()`.
```
mcous: **suggestion:** Minor suggestion due to recent confusion around `baseElement` in `vitest-browser-svelte` repo issues

```suggestion
If the `target` is specified, then this defaults to that, otherwise this defaults to `document.body`. This is used as the base element for the queries as well as what is printed when you use `debug()`. You should rarely, if ever, need to use this option.
```

docs/api/browser/svelte.md
```diff
@@ -0,0 +1,273 @@
...
+  props: { // [!code --]
+    initialCount: 1, // [!code --]
+  }, // [!code --]
+  initialCount: 1, // [!code ++]
+})
+```
+
+### Render Options
+
+#### container
```
mcous: **issue (multiple sites):** Svelte calls this the `target` option and it is part of component options, not setup options

docs/api/browser/svelte.md
```diff
@@ -0,0 +1,273 @@
...
+})
+```
+
+#### baseElement
+
+If the `container` is specified, then this defaults to that, otherwise this defaults to `document.body`. This is used as the base element for the queries as well as what is printed when you use `debug()`.
+
+#### mountOptions
+
+Options to pass down to [`mount`](https://svelte.dev/docs/svelte/imperative-component-api#mount).
```
mcous: **issue (duplicate):** I think this section is redundant with `### Options` - there's no `mountOptions` to pass to the setup options


```suggestion
```

docs/api/browser/svelte.md
```diff
@@ -0,0 +1,273 @@
...
+
+<p>
+  {@render message?.(greeting)}
+</p>
+```
+:::
+
+## See also
+
+- [Svelte Testing Library documentation](https://testing-library.com/docs/svelte-testing-library/intro)
```
mcous, line 280: **suggestion:** should we throw a link to the STL examples, too? I'm still planning on making a similar `examples` directory in `vitest-browser-svelte` when I get a spare moment

```suggestion
- [Svelte Testing Library documentation](https://testing-library.com/docs/svelte-testing-library/intro)
- [Svelte Testing Library examples](https://github.com/testing-library/svelte-testing-library/tree/main/examples)
```

mcous (commented):
Looks great! Got a few minor corrections for the Svelte usage since its options aren't exactly symmetrical with React testing library

docs/api/browser/svelte.md
sheremet-va: I confused `SetupResult` with `SetupOptions` in the types, pretty hard to focus in pure `d.ts` files 😄

docs/api/browser/angular.md
```diff
@@ -0,0 +1,392 @@
...
+@Component({
+  selector: 'app-hello-world',
+  template: '<h1>Hello, {{ name() }}!</h1>',
+})
+export class HelloWorldComponent {
+  name = input.required<string>()
+}
+
+test('renders name', async () => {
+  const { component } = await render(HelloWorldComponent, {
```
shairez: needs updating after https://github.com/vitest-community/vitest-browser-angular/pull/18 is merged

docs/api/browser/angular.md
```diff
@@ -0,0 +1,422 @@
...
+
+  await expect.element(locator).toHaveTextContent('Hello, World!')
+})
+```
+
+::: tip
+`vitest-browser-angular` returns APIs that interact well with built-in [locators](/api/browser/locators), [user events](/api/browser/interactivity) and [assertions](/api/browser/assertions): for example, Vitest will automatically retry the element until the assertion is successful, even if it was rerendered between the assertions.
+:::
+
+The package exposes two entry points: `vitest-browser-angular` and `vitest-browser-angular/pure`. They expose identical API, but the `pure` entry point doesn't add a handler to remove the component before the next test has started.
```
shairez: I'm not sure this is accurate, the package.json currently doesn't export the `/pure`

docs/api/browser/angular.md
```diff
@@ -0,0 +1,422 @@
...
+
+## Setup
+
+Before using `vitest-browser-angular`, you need to set up the test environment using `@analogjs/vitest-angular`:
```
shairez: I think it's best to take the approach that @yjaaidi and I took which is to point to the analog vitest docs (because the setup might change from time to time so you don't want to keep maintaining this). 

So we could point to their docs (like with do on the README) and also give a link to our vitest setup file (for the repo's own tests) to show a practical example

docs/api/browser/angular.md
```diff
@@ -0,0 +1,422 @@
...
+const { locator } = await render(UserComponent, {
+  providers: [
+    { provide: UserService, useValue: mockUserService },
+  ],
+})
+```
+
+#### componentProviders
+
+An array of providers to add at the component level. Use this to override or add providers specific to the component being tested.
```
shairez: "Use this to override or add providers configured on the component's decorator"

docs/api/browser/angular.md
sheremet-va: This is a convention in other packages, do you think it makes sense to expose in angular?

docs/api/browser/angular.md
```diff
@@ -0,0 +1,422 @@
...
+export default defineConfig({
+  test: {
+    setupFiles: ['./vitest.setup.ts'],
+  },
+})
+```
+
+For detailed setup instructions for both Zone.js and Zoneless configurations, please refer to the [Analog Vitest documentation](https://analogjs.org/docs/features/testing/vitest).
+
+## render
```
shairez: do you think it's worth showing the "screen" convention / shortcut for querying? 

because otherwise it's not documented that the RenderResult also contains locators instead of destructuring the `locator`

docs/api/browser/angular.md
sheremet-va: To be fair, as someone who is not familiar with angular, I was very confused by the setup instructions in vitest-browser-angular and had to spend more time than I'd like in unrelated (to my eye) documentation (analog)

sheremet-va: We try to follow this principle in our docs:

> Be Self-Contained - While there are a number of existing tools Vitest pulls in or works like (Chai, Jest, etc…), whenever something is important enough to include in documentation, descriptions and usage examples should be included in the docs instead of linked to an outside source. This will help keep users focused on their task at hand without jumping between multiple tabs to understand a single concept.

https://github.com/vitest-dev/vitest/issues/279

Changing a doc if something changes doesn't seem bad in my opinion

docs/api/browser/angular.md
```diff
@@ -0,0 +1,433 @@
...
+```ts
+import { Component, input } from '@angular/core'
+import { expect, test } from 'vitest'
+import { render } from 'vitest-browser-angular'
+
+@Component({
+  selector: 'app-hello-world',
+  template: '<h1>Hello, {{ name() }}!</h1>',
+})
+export class HelloWorldComponent {
```
yjaaidi: nit: I would remove the `Component` suffix. The convention was removed from the Angular style guide + it aligns more with other frameworks.

docs/api/browser/angular.md
yjaaidi: Oh! I didn't know `pure` was exposed in other packages.
What are the use cases for the `pure` entry point? Why would users opt-out of cleanup?

Also, there are two Angular specificities:

1. whether users are using Angular CLI or Analog, cleanup will be registered between tests _(either through afterEach or beforeEach depending on an Angular's `TestBed` flag)_

2. not cleaning up between tests will often break because the the `TestBed` has two phases, a configuration phase, and an instantiation phase. Once a component is instantiated, calling a configuration API would fail unless there was a cleanup.

Therefore, I still can't see the value of exposing `pure`.

docs/api/browser/angular.md
yjaaidi: The think about Angular is that we currently have two options:
- Angular CLI's `unit-test` builder (the official approach which has limitations as it wraps Vitest instead of providing a plugin)
- Analog approach for those who want more features, more control, and more alignment with Vitest.

We could either document the Angular CLI approach and link to the Analog documentation for those who want to use it.
Or document both in different sections or tabs.

For Angular, enabling browser mode resides in adding `browsers` option in `angular.json`'s test config.
https://angular.dev/guide/testing#running-tests-in-a-browser:~:text=Once%20the%20provider,angular.json

```json
{
  "projects": {
    "your-project-name": {
      "architect": {
        "test": {
          "builder": "@angular/build:unit-test",
          "options": {
            "browsers": ["Chromium", "Firefox", "Safari"] // Add "Headless" suffix to enable headless 🤷
          }
        }
      }
    }
  }
}
```

docs/api/browser/angular.md
```diff
@@ -0,0 +1,433 @@
...
+`vitest-browser-angular` returns APIs that interact well with built-in [locators](/api/browser/locators), [user events](/api/browser/interactivity) and [assertions](/api/browser/assertions): for example, Vitest will automatically retry the element until the assertion is successful, even if it was rerendered between the assertions.
+:::
+
+The package exposes two entry points: `vitest-browser-angular` and `vitest-browser-angular/pure`. They expose identical API, but the `pure` entry point doesn't add a handler to remove the component before the next test has started.
+
+## Setup
+
+Before using `vitest-browser-angular`, you need to set up the test environment using `@analogjs/vitest-angular`:
+
+::: code-group
```
yjaaidi: With the Angular CLI approach, we don't need the configuration below.

docs/api/browser/angular.md
```diff
@@ -0,0 +1,433 @@
...
+```ts [zoneless]
+import '@angular/compiler'
```
yjaaidi: We still have to discuss this with @shairez. `@angular/compiler` should not be necessary, especially if we enable AOT which is the default behavior with the Angular CLI `unit-test` builder.

docs/api/browser/angular.md
```diff
@@ -0,0 +1,433 @@
...
+console.log(container.innerHTML)
+```
+
+:::danger
+If you find yourself using `container` to query for rendered elements then you should reconsider! The [locators](/api/browser/locators) are designed to be more resilient to changes that will be made to the component you're testing. Avoid using `container` to query for elements!
+:::
+
+### componentClassInstance
+
+The actual Angular component class instance. Use this to access component methods and properties directly.
```
yjaaidi: I would add a warning here to discourage this usage.
The component instance should be considered an implementation detail.

docs/api/browser/angular.md
sheremet-va: It is very confusing how to configure `vitest-browser-angular` to work. The readme just redirects to analog, so these examples are from there. If it is not valid, then let's update README and this doc to include all possible setups. I have _no idea_ how to set it up

docs/api/browser/angular.md
sheremet-va: This is an example from analog linked from `vitest-browser-angular` README

docs/api/browser/angular.md
sheremet-va: We need to document all possible approaches. Setting this up is frankly annoying without a proper guide. We don't need to document all config options (like the headless note), just how to get started with testing

docs/api/browser/angular.md
sheremet-va: In other frameworks people can disable automatic cleanup to do it manually. For example, to move it into a different hook, or add an extra action before/after cleanup. If this is not advisable with how angular testing works, we don't have to add this entry point

docs/api/browser/angular.md
shairez: I agree if the approach is to copy their docs every time they change, we need to show both, not just one way

sheremet-va: Sure, let’s show both. We don’t need to copy the whole doc, just the setup with a link to the original doc that expands what we showed. At the moment readme from your package lead me to analog doc that I just copied here. Turns out, it’s not actually correct. Which means the current approach to just simply link is flawed

shairez: well each one has its trade-offs IMO, I don't think one is flawed and the other is not, it's just a matter of how many places you want to keep submitting the fix PR to... :sweat_smile: 

(and I'm not sure the analog docs has a mistake, what younes said is that it's just not relevant if one goes with the native angular cli way,  but AFAIK you do need the compiler if you go with analog )

I originally had all the setup steps as part of the README of the package, but after talking with @yjaaidi , we agreed that it's best to link to analog's setup as it keeps being updated, than to play catch with it every time it changes. 

Angular, unfortunately, has its own way of doing things ( I wish it was just a vite plugin and we could consolidate everything into one thing), and that forces us to instead of showing one way we need to show 2 ways... which is not fun. 

But we can show 2 minimal setups, and we'll just need to keep in mind we'll need to maintain them as things change

yjaaidi: Hey guys, I've just created a PR for `vitest-browser-angular` README.
I tried to document both approaches (Angular CLI & Analog) while embedding instructions that should work out of the box.

Once @shairez reviews it, we could use it as a source of inspiration for the docs.

yjaaidi:
@shairez suggested that we contribute directly to this PR. Would that work for you @sheremet-va?

I just took time to read everything thoroughly and here is a quick summary of the changes that I would make:

1. I would update the first example to be interactive (e.g. a counter just like other frameworks.
2. Update setup instructions as in https://github.com/vitest-community/vitest-browser-angular/pull/21
3. Remove the example from `componentProviders` option as it can be considered an advanced use case and users should pick the `providers` option unless they know what they are doing
4. Do not document the `imports` option. Users should use `providers` combined with [importProvidersFrom](https://angular.dev/api/core/importProvidersFrom)
5. Do not document the `withRouting`, `componentClassInstance`, `router`, and `routerHarness`. This is something that we still have to discuss internally but I'd rather not expose these in the docs until we decide whether we keep them or not.
6. Update `fixture` example by removing the `fixture.detectChanges` call — it's better to discourage users from using it.
7. We could add a link to `https://angular.dev/api/core/testing/ComponentFixture#` in `fixture`
8. Probably remove the examples at the bottom. Having an interactive first example such as a counter is probably enough.

What do you think Shai?
Depending on Shai's feedback we can then:
- suggest changes on this PR
- make a PR against this PR
- remove the `angular.md` file from this PR so that we do not block other frameworks then submit a dedicated Angular PR

Whatever works for you Vlad.

sheremet-va:
> remove the `angular.md` file from this PR so that we do not block other frameworks then submit a dedicated Angular PR

Let's add the angular one in a separate PR to unblock this one.

PR #9332, opened 2025-12-23
feat: return a disposable from doMock()

### LINKED ISSUE #8425: Provide disposable APIs for releasing mocks

It's easy to forget to release mocks. For example, the following code mocking the `node:path` module pollutes subsequent test cases by forgetting to release the mock.

```ts
import { describe, expect, vi, it } from 'vitest';

describe('pollution', () => {
  it('mocks a module', async () => {
    vi.doMock(import('node:path'), async (importActual) => {
      const pathActual = await importActual();
      return {
        ...pathActual,
        default: 'mocked!'
      }
    })

    const path = await import('node:path');

    expect(path.default).toBe('mocked!');
  })

  it('should not be mocked anymore', async () => {
    const path = await import('node:path');
    expect(path.default).not.toBe('mocked!'); // fails; polluted by other test case
  });
})
```

### Suggested solution

It would be highly ergonomic to release mocks by providing a disposable API that releases the mock

```ts
import { describe, expect, vi, it } from 'vitest';

describe('no pollution', () => {
  it('mocks a module', async () => {
    using _temporary_mock = vi.disposableMock(import('node:path'), async (importActual) => {
      const pathActual = await importActual();
      return {
        ...pathActual,
        default: 'mocked!'
      }
    })

    const path = await import('node:path');
  
    expect(path.default).toBe('mocked!');
  })

  it('should not be mocked anymore', async () => {
    const path = await import('node:path');
    expect(path.default).not.toBe('mocked!');
  });
})
```

### Alternative

One can either
- keep track of all mocked modules, and use a `beforeEach`/`afterEach` with `vi.doUnmock('node:path');`
- use `try ... finally` to call `vi.doUnmock('node:path')` after calling `vi.doMock('node:path')`;

These both rely on non-local code to clean up the mock state, whereas `using` would localize all information to the declaration of the mock

### Additional context

`using` is available out of the box in current versions of node.

keywords: explicit resource management

### Validations

kirkwaiblinger:
### Description

This makes `doMock()` return a disposable that calls the corresponding `doUnmock()` upon being disposed.

Resolves #8425 

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/api/vi.md
+++ b/docs/api/vi.md
@@ -230,7 +230,7 @@ test('importing the next module imports mocked one', async () => {
 ```
 
 ::: tip
-In environments that support [Explicit Resource Management](https://github.com/tc39/proposal-explicit-resource-management), you can use `using` instead of `const` to automatically call [`vi.doUnmock()`](#vi-dounmock) on the mocked module when the containing block is exited. This is especially useful when mocking a dynamically imported module for a single test case.
+In environments that support [Explicit Resource Management](https://github.com/tc39/proposal-explicit-resource-management), you can use `using` on the value returned from `vi.doMock()` to automatically call [`vi.doUnmock()`](#vi-dounmock) on the mocked module when the containing block is exited. This is especially useful when mocking a dynamically imported module for a single test case.
 
 ```ts
 it('uses a mocked version of my-module', () => {


test/core/test/do-mock.test.ts
kirkwaiblinger, line 1: I think there is an oddity when it comes to nesting... 

```ts
it('should release 1 mock at a time?', async () => {
  const myModuleNoMock = await import('my-module');
  using _mockDisposable1 = vi.doMock('my-module', ...);
  const myModuleMock1 = await import('my-module');
  { 
    using _mockDisposable2 = vi.doMock('my-module', ...);
    const myModuleMock2 = await import('my-module');
  }
  // What should this be? myModuleNoMock or myModuleMock1?
  const myModuleMock3 = await import('my-module'); 
});
```

I think `myModuleMock3 === myModuleMock1` is probably the _best_ answer, but then the `[Symbol.dispose]()` method must not be literally `vi.doUnmock()` but instead must either track the "stack" of mocks in some way, or capture the preceding resolved value of the mocked module and restore it.

For the moment I've gone with `myModuleMock3 === myModuleNoMock` instead, since the implementation is simpler

Let me know if you'd like changes to this behavior or if it's ok to leave as is.

test/core/test/do-mock.test.ts
```diff
@@ -30,3 +30,35 @@ test('the second doMock can override the first doMock', async () => {
...
+    expect(incrementWith10(1)).toBe(11)
+  }
+
+  {
+    const { increment: incrementWith1 } = await import('./fixtures/increment')
+    expect(incrementWith1(1)).toBe(2)
+  }
+})
+
+test.skipIf(Symbol.dispose)('doMock works with using', async () => {
```
sheremet-va, line 58: This test seems weird, because `Symbol.dispose` in `_incrementMock?.[Symbol.dispose]` is always `undefined` and test always checks for `_incrementMock[undefined]`

Maybe it should instead check `Object.keys(_incrementMock)`

test/core/test/do-mock.test.ts
sheremet-va: I would expect `myModuleMock3 === myModuleNoMock` because `doUnmock` removed the mock, it doesn't restore the previous mock

sheremet-va (changes requested):

test/core/test/do-mock.test.ts
kirkwaiblinger: I had lifted the pattern from here:

https://github.com/vitest-dev/vitest/blob/d3afa601a63cc2eb6e2aa2acc9369a10873e7acb/test/core/test/jest-mock.test.ts#L758-L762

I wasn't sure if `Symbol.dispose` was patched in between evaluating the `skipIf` condition and the test body or something, for example, by something like

https://github.com/vitest-dev/vitest/blob/e922e9266c50c110f797704ed41cb3cc1ec1a58b/test/core/test/esnext.test.ts#L22

---

If `Symbol.dispose` really is `undefined`, then maybe we just skip the test altogether? There are already tests that assert that `doMock()`'s functionality works, so it doesn't really matter what the return value is in a non-`using`-supporting environment 🤷 Or, like you say, we could examine the keys of the object, but it's just a little weird to try to do that when we can't name the key that we're interested in.

sheremet-va: Sure, if there is no reason to test something, then don't test it

Issue #9271, opened 2025-12-16
Vitest hangs if `poolRunner.start` throws an error


sheremet-va:
### Describe the bug

If anything inside `poolRunner.start` throws an error, then Vitest hangs indefinitely. I've noticed that when trying to run the reproduction as a test via `runInlineTests`, it passes, but when running in an actual terminal, it hangs the terminal.

I've noticed that:
- Worker is not stopped
- `Pool.prototype.activeTasks` still has the task

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-byqm9r5n?file=vite.config.ts,env.js&initialPath=__vitest__/

### System Info

```shell
Any
```

### Used Package Manager

pnpm

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vitest.dev/guide/).
- [x] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #9337, opened 2025-12-23
fix(pool): handle worker start failures gracefully

AriPerkkio:
### Description

- Resolves https://github.com/vitest-dev/vitest/issues/9271

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/pools/poolRunner.ts
+++ b/packages/vitest/src/node/pools/poolRunner.ts
@@ -15,6 +15,7 @@ enum RunnerState {
   IDLE = 'idle',
   STARTING = 'starting',
   STARTED = 'started',
+  START_FAILURE = 'start_failure',
   STOPPING = 'stopping',
   STOPPED = 'stopped',
 }
@@ -231,7 +232,7 @@ export class PoolRunner {
       this._state = RunnerState.STARTED
     }
     catch (error: any) {
-      this._state = RunnerState.IDLE
+      this._state = RunnerState.START_FAILURE
       startSpan?.recordException(error)
       throw error
     }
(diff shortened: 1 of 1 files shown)


packages/vitest/src/node/pools/pool.ts
```diff
@@ -117,21 +118,27 @@ export class Pool {
...
+        span = runner.startTracesSpan(`vitest.worker.${method}`)
+
+        // Start running the test in the worker
+        runner.request(method, task.context)
+      }
 
       await resolver.promise
-        .catch((error) => {
-          span.recordException(error)
-          throw error
```
AriPerkkio, line 131: We should not `throw` anything in this scope. `resolver.reject` is intentionally used for that instead - even though it adds ton of complexity.

AriPerkkio:
Repro from https://stackblitz.com/~/edit/vitest-dev-vitest-byqm9r5n exits fine with these changes locally.

packages/vitest/src/node/pools/poolRunner.ts
```diff
@@ -245,7 +245,7 @@ export class PoolRunner {
 
     this._otel?.span.setAttribute('vitest.worker.files', this._otel.files)
 
-    if (this._state === RunnerState.IDLE) {
+    if (this._state === RunnerState.IDLE && !options?.force) {
```
sheremet-va: I don't really like this logic. I think the pool runner should change the state to something like `ERRORED` or `START_ERRORED` instead of idle. idle is the state before `start`

packages/vitest/src/node/pools/pool.ts
sheremet-va: Would be nice to have this as the comment in the code somewhere

AriPerkkio: There is mention of `resolver` requirement at

https://github.com/vitest-dev/vitest/blob/150065459b96a5ed3b9f07ab3001596492032183/packages/vitest/src/node/pools/pool.ts#L56-L60

Is that enough?

sheremet-va (approved):

PR #9339, opened 2025-12-24
fix: replace crypto.randomUUID to allow insecure environments (fix #9…

### LINKED ISSUE #9338: Browser-Mode fails in insecure environment
### Describe the bug

When using vitest with a server configured to not use localhost, but any other domain without https, vitest fails with the exception "TypeError: crypto.randomUUID is not a function" thrown in wrapModule
https://github.com/vitest-dev/vitest/blob/7b10ab4cd7d23d3c520fd1f6956ce99108f70a04/packages/browser/src/client/public/esm-client-injector.js#L12

The function crypto.randomUUID is not available in insecure environments https://developer.mozilla.org/de/docs/Web/API/Crypto/randomUUID
In CI setups with a remote-selenium-hub it is very useful to spin up vitest-server accessable through the machines hostname, without the need to configure tls.

I consider this a regression, since a similar thing got fixed in #6079

### Reproduction

https://github.com/plusgut/vitest-minimal-reproduction-randomuui/commit/02a2398df11d521405b9ce0754a2b9d8369d6684

### System Info

```shell
System:
    OS: Linux 6.18 cpe:/o:nixos:nixos:26.05 26.05 (Yarara)
    CPU: (16) x64 AMD Ryzen 7 PRO 5850U with Radeon Graphics
    Memory: 17.72 GB / 27.23 GB
    Container: Yes
    Shell: 4.2.1 - /run/current-system/sw/bin/fish
  Binaries:
    Node: 22.21.1 - /nix/store/l85fis49agvp5q1ild1rfh4rrgmn92sr-nodejs-22.21.1/bin/node
    npm: 10.9.4 - /nix/store/l85fis49agvp5q1ild1rfh4rrgmn92sr-nodejs-22.21.1/bin/npm
  Browsers:
    Chromium: 143.0.7499.40
    Firefox: 145.0.2
    Firefox Developer Edition: 145.0.2
  npmPackages:
    @vitest/browser-preview: ^4.0.16 => 4.0.16 
    @vitest/ui: latest => 4.0.16 
    vite: latest => 7.3.0 
    vitest: latest => 4.0.16
```

### Used Package Manager

npm

### Validations

plusgut:
fixes #9338

### Description

crypto.randomUUID is not available in the browser in case it gets executed in an insecure environment Math.random() should suffice for mocking

Resolves #issue-number

### Tests

### Documentation

### Changesets

### DIFF
--- a/test/browser/specs/insecure-context.test.ts
+++ b/test/browser/specs/insecure-context.test.ts
@@ -1,0 +1,13 @@
+import os from 'node:os'
+import { expect, test } from 'vitest'
+import { instances, runBrowserTests } from './utils'
+
+// server.host = os.hostname // doesnt work on mac, therefore the test is only run on linux
+test.runIf(os.platform() === 'linux')('server-host check dynamic import at insecure context', async () => {
+  const { stdout, stderr } = await runBrowserTests({
+    root: './fixtures/insecure-context',
+  })
+
+  expect(stderr).toBe('')
+  expect(stdout).toReportSummaryTestFiles({ passed: instances.length })
+})
--- a/packages/browser/src/client/public/esm-client-injector.js
+++ b/packages/browser/src/client/public/esm-client-injector.js
@@ -9,7 +9,7 @@
     }
 
     const { evaluatedModules } = __vitest_worker__
-    const moduleId = crypto.randomUUID()
+    const moduleId = `${Math.random()}`
     const viteModule = evaluatedModules.ensureModule(moduleId, moduleId)
 
     viteModule.evaluated = false
--- a/test/browser/fixtures/insecure-context/src/basic.test.ts
+++ b/test/browser/fixtures/insecure-context/src/basic.test.ts
@@ -1,0 +1,6 @@
+import { expect, test } from "vitest";
+
+test("basic", async () => {
+  expect(window.isSecureContext).toBe(false);
+  expect(1).toBe((await import("./dynamic-import")).default);
+});
--- a/test/browser/fixtures/insecure-context/src/dynamic-import.ts
+++ b/test/browser/fixtures/insecure-context/src/dynamic-import.ts
@@ -1,0 +1,1 @@
+export default 1;
--- a/test/browser/fixtures/insecure-context/vitest.config.ts
+++ b/test/browser/fixtures/insecure-context/vitest.config.ts
@@ -1,0 +1,19 @@
+import { fileURLToPath } from "node:url";
+import os from "node:os";
+import { defineConfig } from "vitest/config";
+import { instances, provider } from "../../settings";
+
+export default defineConfig({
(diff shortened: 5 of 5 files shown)


userquin:
Check #6079 maybe we can just reuse the tested id using replaceAll in the vite plugin

sheremet-va:
> Check #6079 maybe we can just reuse the tested id using replaceAll in the vite plugin

We just need a random string, it has nothing to do with the tester id

plusgut:
> > Check #6079 maybe we can just reuse the tested id using replaceAll in the vite plugin
> 
> We just need a random string, it has nothing to do with the tester id

@sheremet-va So good ol Math.random is fine by you, if it is in a tenplate literal?

I undrafted the pr now, if you want something to be changed let me know. Do you think a test-case for this would be beneficial?

sheremet-va:
> I undrafted the pr now, if you want something to be changed let me know. Do you think a test-case for this would be beneficial?

Feel free to add a test case. I think we only have one instance where we run with http:
https://github.com/vitest-dev/vitest/blob/f9271a0811147d4ca71488ebd4fc442733d014e5/test/browser/specs/server-url.test.ts#L8

plusgut:
@sheremet-va setting it to http is actually no enough for it to be considered insecure. https://www.w3.org/TR/secure-contexts/#is-origin-trustworthy states that 127.0.0.1 and localhost are always considered trustworthy, no matter if http or https.

Thats why i used the hostname of the machine, but sadly that doesnt seem to work with mac-os. I guess mac-os doesnt automatically resolve it.

I now looked up if there is some browser-capability to set it into insecure-context mode, but that doesn`t seem to exist. I only was able to find the opposite in chrome "--unsafely-treat-insecure-origin-as-secure".

I'm not really sure how to proceed from here, what do you think of skipping the tests in case the os.platform is not linux or windows?
Alternatively i could go over os.networkInterfaces().eth0 and see if that works inside the vm`s.

What do you think would be the best approach without being to brittle?

sheremet-va:
I think running the test on a single platform that allows an easy check is fine.

plusgut:
@sheremet-va that sounds very reasonable to me.

I believe everything should be looking good now, let me know if everything is to your liking or if I should change something.

There are a couple of failing thing in CI, but i believe they are unrelated to my changes. Could you verify that?

test/browser/specs/insecure-context.test.ts
```diff
@@ -0,0 +1,14 @@
+import os from 'node:os'
+import { expect, test } from 'vitest'
+import { instances, runBrowserTests } from './utils'
+
+test('server-host check dynamic import at insecure context', async () => {
```
sheremet-va:
```suggestion
test.runIf(os.platform() === 'linux')('server-host check dynamic import at insecure context', async () => {
```

PR #9345, opened 2025-12-26
docs: rebrand

yyx990803:


sheremet-va:
Something is wrong with the example at https://deploy-preview-9345--vitest-dev.netlify.app/api/expect.html - it is probably related to twoslash somehow

Screenshot 2025-12-26 at 13 27 43

sheremet-va:
Landing doesn't have a dark/light theme toggle, but other pages have it

Screenshot 2025-12-26 at 13 58 44

yyx990803:
> Landing doesn't have a dark/light theme toggle, but other pages have it
> 
> Screenshot 2025-12-26 at 13 58 44

This is by design - landing only has dark mode

userquin:
I have some questions:
- Vitest is not licensed under MIT? We don't have the license in the docs (we should add back the license to the docs home footer or change the LICENSE and footer license at README.md => https://github.com/vitest-dev/vitest/pull/9345/files#diff-3def678deb1b1d5a53948eb8491817dde4dc881e032e785fc61cc93d334eefcdR145-R146)
- IIRC Safari not working correctly with svg favicon (.ico has been removed from config.ts), maybe we can use `public/logo-without-border.svg` in the pwa assets config file to generate the pwa icons or split the generation (use public/logo-without-border.svg for favicon generation and public/logo.svg for the web manifest pwa icons)
- why excluding the theme js and css here https://github.com/vitest-dev/vitest/pull/9345/files#diff-0ef4af6ce5dc9a857e08841f8e686ef7ca83d90fa3d9d570cc92f7a0f9a71cb3R71 ? Won't have theme applied when offline.

sheremet-va:
> Vitest is not licensed under MIT? We don't have the license in the docs (we should add back the license to the docs home footer or change the LICENSE and footer license at README.md => [#9345 (files)](https://github.com/vitest-dev/vitest/pull/9345/files#diff-3def678deb1b1d5a53948eb8491817dde4dc881e032e785fc61cc93d334eefcdR145-R146))

I don't really see how these two are connected, the LICENSE file _is_ the licence, not the note in the documentation. Although I am also a bit confused why the note is removed

> Won't have theme applied when offline

If you can help with checking the PWA setup, please do that, you should be able to push to the branch

userquin:
> I don't really see how these two are connected, the LICENSE file _is_ the licence, not the note in the documentation. Although I am also a bit confused why the note is removed

I mean if removing the license entry, we can use the voidzero like the license al readme.md with the mit license as prefix.

> If you can help with checking the PWA setup, please do that, you should be able to push to the branch

Adding those files to the globIgnore will remove the files from sw precaching, if you just go offline in the network tab and refresh the page the theme is gone. We can just remove the globIgnore, running the docs in local the pwa is not enabled, I use always private browsing when checking pr preview and so I don't need to await the pwa update (maybe Evan add it to test theme changes in the pr preview).

I don't have my laptops, I will back home around mid-January.

sheremet-va:
Seems like PWA plugin doesn't reset the cache properly, I always see the broken CSS when opening it after a new commit

TheAlexLichter (approved):

PR #9347, opened 2025-12-26
fix: deprecate several vitest/* entry points

sheremet-va:
### Description

In an effort to reduce the amount of chunks in Vitest's `dist` directory (which should make the startup time faster), I propose removing some of the entry points:

- `vitest/coverage` -> use `vitest/node` instead
- `vitest/reporters` -> use `vitest/node` instead
- `vitest/environments` -> use `vitest/runtime` instead
- `vitest/snapshot` -> use `vitest/runtime` instead
- `vitest/runners` -> use `vitest` instead
- `vitest/suite` -> use static methods on `TestRunner` from `vitest` instead: `import("vitest").TestRunner.getCurrentTest()`
- `vitest/mocker` is removed completely, use `@vitest/mocker` package directly (this was published by accident at one point and never removed)
- `vitest/internal/module-runner` is removed, the code is moved to `__INTERNAL` in `vitest/runtime`

### DIFF
--- a/packages/vitest/src/public/runners.ts
+++ b/packages/vitest/src/public/runners.ts
@@ -2,4 +2,4 @@ export { NodeBenchmarkRunner } from '../runtime/runners/benchmark'
 export { TestRunner as VitestTestRunner } from '../runtime/runners/test'
 export type { VitestRunner } from '@vitest/runner'
 
-console.warn('Importing from "vitest/runners" is deprecated since Vitest 4.1. Please use "vitest/runtime" instead.')
+console.warn('Importing from "vitest/runners" is deprecated since Vitest 4.1. Please use "vitest" instead.')


packages/vitest/src/public/runners.ts
```diff
@@ -1,3 +1,5 @@
 export { NodeBenchmarkRunner } from '../runtime/runners/benchmark'
-export { VitestTestRunner } from '../runtime/runners/test'
+export { TestRunner as VitestTestRunner } from '../runtime/runners/test'
 export type { VitestRunner } from '@vitest/runner'
+
+console.warn('Importing from "vitest/runners" is deprecated since Vitest 4.1. Please use "vitest/runtime" instead.')
```
sheremet-va:
```suggestion
console.warn('Importing from "vitest/runners" is deprecated since Vitest 4.1. Please use "vitest" instead.')
```

sheremet-va:
TODO: update when https://github.com/vitest-dev/vitest/pull/9367#issue-comment-box is merged

PR #9348, opened 2025-12-27
refactor: remove duplicate values from `coverage.exclude`

btea:
### Description
When `vitest --coverage` is executed, `resolveConfig` will be executed twice.

Resolves #issue-number

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/config/coverage.md
+++ b/docs/config/coverage.md
@@ -58,6 +58,27 @@ List of files excluded from coverage as glob patterns.
 
 See [Including and excluding files from coverage report](/guide/coverage.html#including-and-excluding-files-from-coverage-report) for examples.
 
+The internal implementation will by default write some configuration files into it, as listed below:
+```ts
+[
+  'vitest.config.ts',
+  'vitest.config.mts',
+  'vitest.config.cts',
+  'vitest.config.js',
+  'vitest.config.mjs',
+  'vitest.config.cjs',
+  'vite.config.ts',
+  'vite.config.mts',
+  'vite.config.cts',
+  'vite.config.js',
+  'vite.config.mjs',
+  'vite.config.cjs',
+  '**\/virtual:*',
+  '**\/__x00__*',
+  '**/node_modules/**'
+]
+```
+
 ## coverage.clean
 
 - **Type:** `boolean`
--- a/docs/config/coverage.md
+++ b/docs/config/coverage.md
@@ -58,27 +58,6 @@ List of files excluded from coverage as glob patterns.
 
 See [Including and excluding files from coverage report](/guide/coverage.html#including-and-excluding-files-from-coverage-report) for examples.
 
-The internal implementation will by default write some configuration files into it, as listed below:
-```ts
-[
-  'vitest.config.ts',
-  'vitest.config.mts',
-  'vitest.config.cts',
-  'vitest.config.js',
-  'vitest.config.mjs',
-  'vitest.config.cjs',
-  'vite.config.ts',
-  'vite.config.mts',
-  'vite.config.cts',
-  'vite.config.js',
-  'vite.config.mjs',
-  'vite.config.cjs',
-  '**\/virtual:*',
-  '**\/__x00__*',
-  '**/node_modules/**'
-]
-```
-
 ## coverage.clean
 
 - **Type:** `boolean`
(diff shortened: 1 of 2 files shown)


sheremet-va:
> When `vitest --coverage` is executed, `resolveConfig` will be executed twice.

What do you mean? It is called once.

btea:
As shown in the image below, I set a breakpoint when setting the value of `coverage.exclude`. The first time, `resolved.coverage.exclude` was an empty array, and then the default value was set. The second time, since the value had already been set, the same value was set again.

<details>
	<summary>screenshot</summary>
	image
image
</details>

sheremet-va:
> As shown in the image below, I set a breakpoint when setting the value of `coverage.exclude`. The first time, `resolved.coverage.exclude` was an empty array, and then the default value was set. The second time, since the value had already been set, the same value was set again.
> 
> 
> 
> <details>
> 
> 	<summary>screenshot</summary>
> 
> 	image
> 
> image
> 
> </details>
> 
> 
> 
> 

How does your config look like?

btea:
https://github.com/element-plus/element-plus/pull/23223

sheremet-va:
This seems more like a bug with `projects` resolution, the `coverage.exclude` is not the only affected array. Do you really need a `projects` field? Can't you just put a `unit` in the root's `test.name`?

btea:
yeah, you are right, it appears that after removing the `projects` configuration, `resolveConfig` will only execute once.

By the way, I saw in the documentation that `coverage.exclude` defaults to an empty array, but the actual code has a default value configured. Perhaps the documentation should be updated accordingly.

sheremet-va:
> By the way, I saw in the documentation that `coverage.exclude` defaults to an empty array, but the actual code has a default value configured. Perhaps the documentation should be updated accordingly.

Technically speaking, the default is empty. Custom excludes don't override "the default". This is just an internal implementation detail of how the coverage providers treat certain files (like unreadable virtual files), it could've been hardcoded inside the provider, for example, but it should probably be mentioned in the docs in some way.

btea:
> but it should probably be mentioned in the docs in some way.

I have updated the documentation.

By default, `resolved.config` is usually included in `configFiles`, but its result is an absolute path. Should this be treated as an identical value and removed?
https://github.com/vitest-dev/vitest/blob/150065459b96a5ed3b9f07ab3001596492032183/packages/vitest/src/node/config/resolveConfig.ts#L413-L414

docs/config/coverage.md
```diff
@@ -58,6 +58,27 @@ List of files excluded from coverage as glob patterns.
...
+  'vite.config.mts',
+  'vite.config.cts',
+  'vite.config.js',
+  'vite.config.mjs',
+  'vite.config.cjs',
+  '**\/virtual:*',
+  '**\/__x00__*',
+  '**/node_modules/**'
+]
+```
```
AriPerkkio: I don't think we should document these as it makes them part of public API now. These are implementation details of Vitest. There are some things that should never show up in coverage results.

AriPerkkio (changes requested):

docs/config/coverage.md
btea: That makes sense, I've already removed it.

packages/vitest/src/node/config/resolveConfig.ts
```diff
@@ -410,15 +411,15 @@ export function resolveConfig(
     ...resolved.include,
 
     // Configs
-    resolved.config && slash(resolved.config),
+    resolved.config && basename(slash(resolved.config)),
```
AriPerkkio, line 414: Can you show example what cases this fixes? Does the configuration file show up in coverage in some cases?

btea: vitest

As shown in the image above, `resolved.config` contains the full path, but it should be the same file as the second value in the `configFiles` array.

AriPerkkio: Does it cause issues with full path? For example, if you use `vitest --config some-custom-config.ts`, does that show up in coverage report?

btea: No, it doesn't seem to cause any problems at the moment. Actually, this PR wasn't initially intended to address this issue.

Initially, it was here: https://github.com/vitest-dev/vitest/pull/9348#issuecomment-3693886172. My configuration caused the configuration content to be loaded twice, resulting in duplicate content in the second load. So, I wanted to remove duplicates.

Later, with @sheremet-va's guidance https://github.com/vitest-dev/vitest/pull/9348#issuecomment-3693942145, I adjusted the configuration so that Vitest's configuration loading logic worked correctly, loading only once. Then I noticed that this configuration file might also be duplicated, so I tried removing duplicates there as well. Of course, I think it's probably fine not to modify this one.

PR #9349, opened 2025-12-27
fix(pool): validate environment options when reusing the worker

sheremet-va:
### Description

The VM workers can run any environments, but they need to receive correct options. The base workers can only run one environment with the same options.

### DIFF
--- a/packages/vitest/src/node/pools/poolRunner.ts
+++ b/packages/vitest/src/node/pools/poolRunner.ts
@@ -113,8 +113,13 @@ export class PoolRunner {
     this._offCancel = vitest.onCancel(reason => this._rpc.onCancel(reason))
   }
 
+  /**
+   * "reconfigure" can only be called if `environment` is different, since different project always
+   * requires a new PoolRunner instance.
+   */
   public reconfigure(task: PoolTask): void {
     this.environment = task.context.environment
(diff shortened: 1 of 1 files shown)


sheremet-va:
Gonna merge this for now because I am annoyed at flaky CI, but please have a look after the holidays, @AriPerkkio

packages/vitest/src/node/pools/pool.ts
```diff
@@ -289,11 +292,47 @@ function isEqualRunner(runner: PoolRunner, task: PoolTask) {
...
+  }
+  return runner.worker.canReuse(task)
+}
+
+function isEnvironmentEqual(env1: ContextTestEnvironment, env2: ContextTestEnvironment): boolean {
+  if (env1.name !== env2.name) {
+    return false
+  }
+  return deepEqual(env1.options, env2.options)
+}
```
AriPerkkio, line 311: This looks the same as

https://github.com/vitest-dev/vitest/blob/150065459b96a5ed3b9f07ab3001596492032183/packages/vitest/src/node/pool.ts#L364-L385

sheremet-va:
> This should be implemented in

How can we do it there when there is no worker yet that can say if it's reusable? The groups are needed only for threads/forks, not for vmForks/vmThreads

PR #9370, opened 2026-01-01
feat(test-runner): continue enhanced retry options (follow-up to #8812)

### LINKED ISSUE #8482: Feature Request: Enhanced `retry` options

MazenSamehR:
### Description

Follow-up to #8812. Align retry types with serialization boundaries, pass `TestError` directly to retry condition, use `SerializableRetry` for reporter/runtime-facing data, and sanitize function conditions before reporters. Keeps function conditions usable in test files (workers) while ensuring reporter/config data stays serializable.

Resolves #8482

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/config/retry.md
+++ b/docs/config/retry.md
@@ -35,7 +35,7 @@ vitest --retry 3
 vitest --retry.count 3 --retry.delay 500 --retry.condition 'ECONNREFUSED|timeout'
 ```
 
-## Advanced Options
+## Advanced Options <Version>4.1.0</Version> {#advanced-options}
 
 Use an object to configure retry behavior:
 


MazenSamehR:
Hi @AriPerkkio, can you please review this PR?

docs/config/retry.md
```diff
@@ -5,8 +5,127 @@ outline: deep
...
+    },
+  },
+})
+```
+
+### condition
+
+A string pattern or a function to determine if a test should be retried based on the error.
+
+- When a **string**, it's treated as a regular expression pattern to match against the error message
```
sheremet-va: If it's a regexp, why not use regexp

docs/config/retry.md
```diff
@@ -5,8 +5,127 @@ outline: deep
...
+- When a **function**, it receives the error and returns a boolean
+
+**Note:** When defining `condition` as a function, it must be done in a test file directly, not in `vitest.config.ts` (configurations are serialized for worker threads).
```
sheremet-va:
```suggestion
::: warning
When defining `condition` as a function, it must be done in a test file directly, not in `vitest.config.ts` (configurations are serialized for worker threads).
:::
```

packages/vitest/src/node/cli/cli-config.ts
```diff
@@ -528,8 +528,9 @@ export const cliOptionsConfig: VitestCLIOptions = {
   },
   retry: {
     description:
-      'Retry the test specific number of times if it fails (default: `0`)',
+      'Retry the test specific number of times if it fails (default: `0`). For advanced retry options (count, delay, condition), use the config file.',
```
sheremet-va: Why can't we support `--retry.timeout` and `--retry.count`? Or even `--retry.condition=regexp` (we do support `--testNamePattern` already)

packages/runner/src/run.ts
```diff
@@ -266,6 +302,45 @@ async function callCleanupHooks(runner: VitestRunner, cleanups: unknown[]) {
...
+    return regex.test(error.message || '')
+  }
+  else if (typeof condition === 'function') {
+    // Function condition is called with TestError
+    try {
+      return condition(error)
+    }
+    catch (e) {
+      // If condition function throws, treat as no match
+      console.error('retryCondition function threw error:', e)
```
sheremet-va: Let's rethrow an error that condition has an error

packages/vitest/src/node/config/serializeConfig.ts
```diff
@@ -6,6 +6,20 @@ export function serializeConfig(project: TestProject): SerializedConfig {
   const viteConfig = project._vite?.config
   const optimizer = config.deps?.optimizer || {}
 
+  // Handle retry configuration serialization
+  let retry = config.retry
+  if (retry && typeof retry === 'object' && typeof retry.condition === 'function') {
+    console.warn(
```
sheremet-va: We don't use `console`, there is `vitest.logger.warn`, see other examples with it

packages/vitest/src/node/reporters/reported-tasks.ts
```diff
@@ -524,11 +525,28 @@ export interface TaskOptions {
...
   readonly shuffle: boolean | undefined
-  readonly retry: number | undefined
+  readonly retry: SerializableRetry | undefined
   readonly repeats: number | undefined
   readonly mode: 'run' | 'only' | 'skip' | 'todo'
 }
 
+function sanitizeRetryForReporters(retry: unknown): SerializableRetry | undefined {
+  if (retry === undefined || typeof retry === 'number') {
+    return retry as SerializableRetry | undefined
```
sheremet-va: Why do we need to cast?

packages/vitest/src/node/reporters/reported-tasks.ts
```diff
@@ -524,11 +525,28 @@ export interface TaskOptions {
   readonly fails: boolean | undefined
   readonly concurrent: boolean | undefined
   readonly shuffle: boolean | undefined
-  readonly retry: number | undefined
+  readonly retry: SerializableRetry | undefined
   readonly repeats: number | undefined
   readonly mode: 'run' | 'only' | 'skip' | 'todo'
 }
 
+function sanitizeRetryForReporters(retry: unknown): SerializableRetry | undefined {
```
sheremet-va: Maybe if it wasn't `unknown`, we wouldn't need to cast the type

packages/vitest/src/node/types/config.ts
```diff
@@ -783,11 +783,17 @@ export interface InlineConfig {
...
+   * Retry configuration for tests.
+   * - If a number, specifies how many times to retry failed tests
+   * - If an object, allows fine-grained retry control
    *
-   * @default 0
+   * ⚠️ WARNING: Function form is NOT supported in vitest.config.ts
+   * because configurations are serialized when passed to worker threads.
+   * Use the function form only in test files directly.
+   *
+   * @default 0 (retry on all errors)
```
sheremet-va: Doesn't 0 mean "don't retry"?

packages/vitest/src/node/reporters/reported-tasks.ts
sheremet-va: Actually, why do we even need this, we remove the function in packages/vitest/src/runtime/runners/index.ts already

sheremet-va (changes requested):

docs/config/retry.md
MazenSamehR: I used a string because it works seamlessly in JSON configs and CLI arguments where raw RegExp objects don't. So we ensure it works everywhere. We still treat it as a Regex internally (and make it case-insensitive), so it covers most use cases. And if a user needs specific flags or complex logic, they can use the function option instead.

packages/vitest/src/node/cli/cli-config.ts
MazenSamehR: I've updated the CLI config to support these flags. Ready for review!

packages/runner/src/run.ts
MazenSamehR: I removed the try-catch entirely to let errors propagate naturally

packages/vitest/src/node/types/config.ts
MazenSamehR: My mistake! I mixed up the retry default with the default condition logic. It is fixed now.

docs/config/retry.md
sheremet-va: What json config? Vitest doesn't support json config. CLI options have a `transform` method that you can define. 

Once again, we already have regexp in options, just define a regexp, you don't need to serialise it, except in the CLI

MazenSamehR: Ohh sorry my bad, you're right.

Just to confirm, do you mean I should remove the string type and have condition accept only RegExp in addition to functions?

sheremet-va: Yes, accept regexp and a function

MazenSamehR: Can you review now if it has been resolved?

docs/config/retry.md
```diff
@@ -5,8 +5,140 @@ outline: deep
...
+
+### condition
+
+A RegExp pattern or a function to determine if a test should be retried based on the error.
+
+- When a **RegExp**, it's tested against the error message
+- When a **function**, it receives the error and returns a boolean
+
+> [!WARNING]
+When defining `condition` as a function, it must be done in a test file directly, not in `vitest.config.ts` (configurations are serialized for worker threads).
```
sheremet-va:
```suggestion
::: warning
When defining `condition` as a function, it must be done in a test file directly, not in `vitest.config.ts` (configurations are serialized for worker threads).
:::
```

packages/runner/src/run.ts
```diff
@@ -412,9 +479,23 @@ export async function runTest(test: Test, runner: VitestRunner): Promise<void> {
       }
 
       if (retryCount < retry) {
-        // reset state when retry test
+        // Check if we should retry based on the error condition
+        const shouldRetry = shouldRetryTest(test, test.result.errors)
+
+        if (!shouldRetry) {
+          // Error doesn't match retry condition, stop retrying
```
sheremet-va: Can we remove the AI generated comments that don't explain anything? I feel like they just pollute the code (I would remove most of the non jsdoc comments in this PR)

packages/runner/src/run.ts
```diff
@@ -266,6 +302,37 @@ async function callCleanupHooks(runner: VitestRunner, cleanups: unknown[]) {
   }
 }
 
+/**
+ * Determines if a test should be retried based on its retryCondition configuration
+ */
+function shouldRetryTest(test: Test, errors: TestError[] | undefined): boolean {
```
sheremet-va: This seems a little bit weird to me because it's applied _after_ the `retryCount > retry` is checked, so maybe the name should be a bit different? (`passesRetryCondition` or something) On the `!condition` line I thought that we _always_ just retry all tests until the pass

packages/vitest/src/node/config/serializeConfig.ts
```diff
@@ -1,11 +1,27 @@
 import type { TestProject } from '../project'
 import type { SerializedConfig } from '../types/config'
+import c from 'tinyrainbow'
 
 export function serializeConfig(project: TestProject): SerializedConfig {
   const { config, globalConfig } = project
   const viteConfig = project._vite?.config
   const optimizer = config.deps?.optimizer || {}
 
+  // Handle retry configuration serialization
```
sheremet-va: I think this should actually be done in `resolveConfig` once, not for every `serializeConfig` call

There should also be a test for this (see failures tests in test/cli)

MazenSamehR: Since I'm intentionally passing an invalid condition type (a function) in the config fixture, TypeScript is complaining. Is it ok to add `// @ts-expect-error` like this: 
```ts
export default defineConfig({
  test: {
    retry: {
      count: 3,
      // @ts-expect-error
      condition: () => true
    }
  }
})
```

Or did I misunderstand you?

sheremet-va: I don't know what you are referring to, I am saying the code from this file should be moved to `resolveConfig.ts`

MazenSamehR: Yes, I moved the code to resolveConfig.ts as requested. I'm just creating a test to verify this change as requested.

sheremet-va: Ah, sorry. Yes, it is ok to have a `ts-expect-error` in the test

docs/config/retry.md
```diff
@@ -5,8 +5,141 @@ outline: deep
...
+
+### condition
+
+A RegExp pattern or a function to determine if a test should be retried based on the error.
+
+- When a **RegExp**, it's tested against the error message
+- When a **function**, it receives the error and returns a boolean
+
+::: warning
+When defining `condition` as a function, it must be done in a test file directly, not in `vitest.config.ts` (configurations are serialized for worker threads).
```
sheremet-va:
```suggestion
When defining `condition` as a function, it must be done in a test file directly, not in a configuration file (configurations are serialized for worker threads).
```

packages/vitest/src/node/config/resolveConfig.ts
```diff
@@ -142,6 +142,18 @@ export function resolveConfig(
     mode,
   } as any as ResolvedConfig
 
+  if (resolved.retry && typeof resolved.retry === 'object' && typeof resolved.retry.condition === 'function') {
+    logger.console.warn(
+      c.yellow('Warning: retry.condition function cannot be used in vitest.config.ts. '
```
sheremet-va:
```suggestion
      c.yellow('Warning: `retry.condition` function cannot be used inside a config file. '
```

packages/vitest/src/node/types/config.ts
```diff
@@ -783,11 +783,17 @@ export interface InlineConfig {
   bail?: number
 
   /**
-   * Retry the test specific number of times if it fails.
+   * Retry configuration for tests.
+   * - If a number, specifies how many times to retry failed tests
+   * - If an object, allows fine-grained retry control
    *
-   * @default 0
+   * ⚠️ WARNING: Function form is NOT supported in vitest.config.ts
```
sheremet-va:
```suggestion
   * ⚠️ WARNING: Function form is NOT supported in a config file
```

sheremet-va:
Looks good, just some wording changes needed

MazenSamehR:
Done

docs/config/retry.md
```diff
@@ -5,8 +5,141 @@ outline: deep
...
+
+```bash
+# Simple retry count
+vitest --retry 3
+
+# Advanced options using dot notation
+vitest --retry.count 3 --retry.delay 500 --retry.condition 'ECONNREFUSED|timeout'
+```
+
+## Advanced Options
```
sheremet-va:
```suggestion
## Advanced Options <Version>4.1.0</Version> {#advanced-options}
```

sheremet-va (approved):

sheremet-va:
Thank you!

PR #9373, opened 2026-01-02
fix(browser-webdriverio): fall back to WebDriver Classic #9244

### LINKED ISSUE #9244: BUG: @vitest/browser-webdriverio fails with Safari when BiDi commands are used despite enforcing WebDriver Classic
### Describe the bug

Original Report: https://github.com/webdriverio/webdriverio/issues/14927
Reproduction Repo: https://github.com/chrisguttandin/vitest-webdriverio

Root Cause
The @vitest/browser-webdriverio package's setViewport method unconditionally uses a WebDriver BiDi command without checking if BiDi is actually available:

Why This Fails
Safari on GitHub Actions may not support WebDriver BiDi yet OR the BiDi session wasn't established
User sets 'wdio:enforceWebDriverClassic': true to explicitly disable BiDi
WebdriverIO respects this and doesn't create a BiDi session (no webSocketUrl in capabilities)
Vitest's provider ignores this and tries to use BiDi commands anyway
Result: Command fails because no BiDi session exists

we need to check BiDi availability before using BiDi commands, and fall back to WebDriver Classic commands when BiDi isn't available.

### Reproduction

Original Report: https://github.com/webdriverio/webdriverio/issues/14927
Reproduction Repo: https://github.com/chrisguttandin/vitest-webdriverio

### System Info

```shell
Original Report: https://github.com/webdriverio/webdriverio/issues/14927
Reproduction Repo: https://github.com/chrisguttandin/vitest-webdriverio
```

### Used Package Manager

npm

### Validations

JustasMonkev:
### Description

  Check `browser.isBidi` before using BiDi commands in `setViewport`. When BiDi is unavailable (e.g., Safari or when `wdio:enforceWebDriverClassic` is set), fall back to `setWindowSize` instead of failing.

Resolves #9244

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/browser-webdriverio/src/webdriverio.ts
+++ b/packages/browser-webdriverio/src/webdriverio.ts
@@ -117,14 +117,19 @@ export class WebdriverBrowserProvider implements BrowserProvider {
     if (this.topLevelContext == null || !this.browser) {
       throw new Error(`The browser has no open pages.`)
     }
-    await this.browser.send({
-      method: 'browsingContext.setViewport',
-      params: {
-        context: this.topLevelContext,
-        devicePixelRatio: 1,
-        viewport: options,
-      },
-    })
+    if (this.browser.isBidi) {
+      await this.browser.send({
+        method: 'browsingContext.setViewport',
+        params: {
+          context: this.topLevelContext,
+          devicePixelRatio: 1,
+          viewport: options,
+        },
+      })
+    }
+    else {
+      await this.browser.setWindowSize(options.width, options.height)
+    }
   }
 
   getCommandsContext(): {


packages/browser-webdriverio/src/webdriverio.ts
```diff
@@ -117,14 +117,19 @@ export class WebdriverBrowserProvider implements BrowserProvider {
...
+        method: 'browsingContext.setViewport',
+        params: {
+          context: this.topLevelContext,
+          devicePixelRatio: 1,
+          viewport: options,
+        },
+      })
+    }
+    else {
+      await this.browser.setWindowSize(options.width, options.height)
```
sheremet-va, line 131: Why not `browser.setViewport`?

JustasMonkev: `browser.setViewport()` in `WebDriverIO` uses `browsingContextSetViewport()` under the hood, which is a BiDi-only command.

  `setWindowSize` is the WebDriver Classic equivalent for controlling browser dimensions. While not a perfect 1:1 match (it sets window size including browser chrome, not just viewport). 
  
  https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/src/commands/browser/setViewport.ts#L61
  
  https://github.com/webdriverio/webdriverio/blob/5b35e4e2b94e3ac5247ec2bf27f2f5e41ef00372/packages/webdriverio/src/commands/browser/setWindowSize.ts#L26

PR #9381, opened 2026-01-03
fix(browser): ignore channel events in a tester opened as a top-level window (fix #9379)

### LINKED ISSUE #9379: Browser mode: Infinite 'Unknown event: response:response:...' loop when running multiple click tests
## Describe the bug

When running multiple browser tests that involve click events on anchor elements, thousands of "Unknown event: response:response:response:..." errors occur. The `response:` prefix accumulates recursively, suggesting an infinite feedback loop in iframe communication.

## Reproduction

External reproduction repo: https://github.com/simshanith/lit-ui-router/issues/69

Steps:
1. Clone https://github.com/simshanith/lit-ui-router
2. Checkout commit `9d116d4`
3. Run `pnpm install`
4. Run: `pnpm --filter lit-ui-router test ui-sref.spec -t "click modifiers" --browser=chromium`

### Expected behavior
Tests pass with no errors.

### Actual behavior
Tests pass but with thousands of errors:
```
Error: Unknown event: response:response:response:response:response:execute
Error: Unknown event: response:response:response:response:response:cleanup
```

## Analysis

Looking at [`packages/browser/src/client/tester/tester.ts` lines 94-97](https://github.com/vitest-dev/vitest/blob/main/packages/browser/src/client/tester/tester.ts#L94-L97):

```typescript
channel.postMessage({
  event: `response:${data.event}`,
  iframeId: getBrowserState().iframeId!,
})
```

When an unknown event arrives:
1. It logs "Unknown event: {eventName}" via `unhandledError()`
2. It still sends a response with `response:${data.event}`

If responses somehow get echoed back (iframe routing issue with click events?), this creates a feedback loop:
- `execute` → `response:execute` → `response:response:execute` → ...

### Hypothesis

Click events on anchors with `target="_blank"` or navigation-triggering attributes may cause iframe communication that echoes responses back to the handler.

## Proposed Fix

Add a guard to prevent responding to events that already start with `response:`:

```typescript
// Before sending response, check if this is already a response event
if (data.event.startsWith('response:')) {
  // Don't respond to response events - this would create infinite loop
  return
}
```

## Key Details

- Issue only occurs when **multiple tests** run (individual tests pass)
- Tests involve **click events** on anchor elements  
- Anchors may trigger **navigation**
- Some tests use `target="_blank"` or `rel="external"` attributes
- Error count grows exponentially with test count

## Environment

- vitest: 4.0.16
- @vitest/browser-playwright: 4.0.16
- playwright: 1.54.1
- macOS

simshanith:
## Description

Fixes #9379 — the recursive `Unknown event: response:response:...:execute/cleanup` storm in browser mode.

### Root cause (traced)

My earlier explanation (BroadcastChannel echoing to the sender) was wrong, as pointed out in review — thanks. The actual mechanism:

1. The orchestrator creates tester iframes with `src = "/?sessionId=<id>&iframeId=<id>"` (`createTestIframe`, `packages/browser/src/client/orchestrator.ts`).
2. The dev server serves a fully functional tester page to **any** window that requests `/` with a live `sessionId` (`packages/browser/src/node/serverTester.ts`).
3. An anchor's `href=""` resolves to the current URL **including the query string**. So a test that clicks `<a href="" target="_blank">` — or ctrl/meta-clicks a relative link, which routers deliberately don't `preventDefault()` — opens a popup at the tester URL. The popup boots a **second, top-level tester** with the same `sessionId` and the same `iframeId`, joining the same `BroadcastChannel`.

From then on the two same-`iframeId` testers ping-pong:

- The popup posts `ready`. The real tester receives it (the `iframeId` check passes — it's its own id), posts `ack:ready`, falls into the `default:` case (`Unknown event: ready`), and the `finally` block posts `response:ready`.
- The popup receives `ack:ready` and `response:ready` and does the same. Every received message spawns an `ack:` **and** a `response:` in the twin, so growth is exponential: `ack:response:response:ack:...:ready`.
- The storm also starves the real tester, so the run additionally fails with `The iframe ... did not acknowledge the "cleanup" message within 60000ms`.

The orchestrator already ignores `response:`/`ack:` events in its `default:` case ("ignore acknowledgements and responses to events we sent", added in #10656); the tester side had no equivalent protection against a duplicate participant.

This also explains the coverage report in #9379 (grant-progress): anything that opens the tester URL in a second browsing context triggers the same loop.

### Fix

The tester only participates in the iframe channel when it is actually embedded in the orchestrator (`window.self !== window.top`): a top-level copy of the page neither handles channel messages nor announces `ready`. This removes the duplicate participant instead of filtering event names, addressing the review feedback that the previous `startsWith('response:')` guard only hid the symptom.

### Numbers (fixture: one test clicking `<a href="" target="_blank">` + one trivial follow-up file, chromium)

| | before | after |
| --- | --- | --- |
| result | 2 tests pass, exit code 1 | 2 tests pass, exit code 0 |
| unhandled errors | 1,238,507 `Unknown event: ...` | 0 |
| duration | 96.6s (cleanup ack timeout) | 1.8s |

### Regression test

`test/browser/specs/duplicate-tester.test.ts` (same style as `specs/readiness.test.ts`): runs the popup-click scenario inline and asserts no `Unknown event` in stderr, exit code 0, and both tests passing.

Verified failing without the fix and passing with it on chromium, firefox and webkit:

- unfixed: chromium exit 1 (579k `Unknown event` errors in ~42s), webkit exit 1 (446k errors), firefox exit 1 (inner run flooded until the spec's 60s timeout)
- fixed: all three pass in 2–4s; `specs/readiness.test.ts` (3 tests over the same handshake machinery) still passes on all three

## AI disclosure

Root-cause tracing, the fix, and the regression test were developed with assistance from Claude Code; I reviewed all changes and the verification runs above were executed locally.

## Please don't delete this checklist! Before submitting the PR, please make sure you do the following:

## Tests

- [x] Run the tests with `pnpm test:ci`. (`duplicate-tester` + `readiness` specs verified on chromium/firefox/webkit; a full local `test/browser` run hits pre-existing failures that reproduce identically on unmodified main in my environment, so relying on CI for the full matrix)

## Documentation

## Changesets

- [x] Changes in changelog are generated from PR name. Please, make sure that it explains your changes in an understandable way. Please, prefix changeset messages with `feat:`, `fix:`, `perf:`, `docs:`, or `chore:`.


packages/browser/src/client/tester/tester.ts
```diff
@@ -51,6 +51,10 @@ channel.addEventListener('message', async (e) => {
     return
   }
 
+  if (data.event.startsWith('response:')) {
```
sheremet-va: It should not receive this even here at all, this doesn't fix the issue, it hides it

sheremet-va (changes requested):

packages/browser/src/client/tester/tester.ts
simshanith: Thanks for the review @sheremet-va - agreed this is a symptom and not the root cause.

investigating further and adding test case. will share plan

simshanith: I still don't know exactly how iframes of two or more tests end up communicating, potentially an iframeId collison.

I've simulated the behavior with a second broadcast channel that echos with the `response:` prefix unguarded. Additionally added the guard around postMessage. With the browser package from main, the test exhibits the infinite loop that crashed my computer a few times. With the guards, it passes without error.

simshanith:
@sheremet-va this is ready for another look — rewrote it from scratch, and your review was right on both counts:

- BroadcastChannel indeed never echoes to the sender. The actual mechanism: the dev server serves a functional tester page to **any** window with a live `sessionId`, and a test that lets the browser's default action run on `<a href="" target="_blank">` (`href=""` resolves to the tester URL, query string included) opens a popup that boots a **second, top-level tester with the same `iframeId`** — the two then ping-pong `ack:`/`response:` events exponentially. Full trace and a self-contained repro are now in #9379 (issue body updated today).
- The `startsWith('response:')` guard is gone. The fix now removes the duplicate participant instead of filtering event names: the tester only joins the iframe channel when it is actually embedded in the orchestrator (`window.self !== window.top`) — the tester-side counterpart of the orchestrator guard from #10656.

New regression test `test/browser/specs/duplicate-tester.test.ts` (same style as `specs/readiness.test.ts`): on main it fails with ~580k `Unknown event` errors / exit 1; with the fix it passes in ~2s. Verified on chromium, firefox and webkit; `specs/readiness.test.ts` still passes on all three. The one failing browser shard here looks unrelated (`mocking.test.ts` watch-rerun + a firefox `wheel.test.ts` flake) — `duplicate-tester` itself passed on CI.

sheremet-va (approved):

sheremet-va:
Sorry for the long wait, the fix makes sense to me

PR #9396, opened 2026-01-05
docs: Update globals.md with tsconfig.app.json guidance

SaKaNa-Y:
When creating a project with Create Vue, I noticed that adding "types\": [\"vitest/globals\"] to tsconfig.json does not take effect. However, placing this configuration in tsconfig.app.json works as expected. This behavior is not caused by my local setup, but by how Create Vue structures and applies TypeScript configuration files. Therefore, it’s worth adding a small note to clarify this.

If this issue is caused by my own mistake, I would appreciate it if you could help clarify my confusion.

### Description

Resolves #issue-number

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/config/globals.md
+++ b/docs/config/globals.md
@@ -34,6 +34,10 @@ To get TypeScript working with the global APIs, add `vitest/globals` to the `typ
 }
 ```
 
+::: tip
+If you create a project using `create vue`, the `tsconfig.app.json` file should be used to place these configurations; only then will they take effect in the project.
+:::
+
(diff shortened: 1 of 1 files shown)


docs/config/globals.md
```diff
@@ -34,6 +34,10 @@ To get TypeScript working with the global APIs, add `vitest/globals` to the `typ
 }
 ```
 
+::: tip
+If you create a project using `create vue`, the `tsconfig.app.json` file should be used to place these configurations; only then will they take effect in the project.
```
hi-ogawa, line 38: I don't think mentioning specifically `create vue` would be the solution since other templates have different convention about tsconfig files. You need to understand why templates sets up such tsconfig files and that should be technically explained by their template (and then where to put `"types": ["vitest/globals"]` can be understood from the principle).

Issue #9141, opened 2025-12-01
Enhance importBreakdown support


sheremet-va:
### Clear and concise description of the problem

At the moment Vitest can only show information about import breakdown. It would be nice to be able to control it and react to the duration.

Related: https://github.com/vitest-dev/vitest/pull/9105

### Suggested solution

Enhance the import breakdown feature with these options (WIP):

```ts
interface ImportBreakdownOptions {
  /**
   * Always show an import breakdown after the test summary.
   * @default false
   */
  enabled?: boolean
  /**
   * If Vitest finds a file that takes longer than `experimental.importDurationThresholds.warn`,
   * it will show an Import Breakdown.
   *
   * Imports Breakdown is always shown in UI inside the "Module Graph" tab.
   * @default true
   */
  warnInTerminal?: boolean
  /**
   * Fail the test run if there is a module that takes longer than `experimental.importDurationThresholds.danger`.
   */
  // failOnDanger?: boolean
  importDurationThresholds?: {
    /**
     * Warning threshold in milliseconds. Warning are shown in orange/yellow color.
     * @default 100
     */
    warn?: number
    /**
     * Danger threshold in milliseconds. These are shown in red color.
     * @defaul 500
     */
    danger?: number
  },
  // @default 10
  // throw error if less than 0
  // 0 is Infinity
  limit?: number
}
```

Turn self/total into a column

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vitest.dev/guide/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

hi-ogawa:
After experimenting with optimizing output format https://github.com/vitest-dev/vitest/pull/9265, there's an inherent limit on how far we can optimize since the raw data can scale by (number of test files) x (number of source files).

We should likely re-purpose or extend the flag discussed here to also truncate/slim-down reporter file output (blob, html). For example, we could reuse `limit` here to truncate data, so the raw output file get limited by (number of test files) x (limit).

sheremet-va:
> After experimenting with optimizing output format https://github.com/vitest-dev/vitest/pull/9265, there's an inherent limit on how far we can optimize since the raw data can scale by (number of test files) x (number of source files).
[...]

PR #9401, opened 2026-01-07
feat(experimental): add `importDurations: { limit, print }` options

hi-ogawa:
### Description

- Closes https://github.com/vitest-dev/vitest/issues/9216
- Part of https://github.com/vitest-dev/vitest/issues/9141

This PR restructure `experimental.printImportDurations: boolean` options into following.

```ts
experimental: {
  importDurations?: {
    limit?: number  // Collection and display limit (default: 0, or 10 if print/UI enabled)          
    print?: boolean // Print to CLI terminal (default: false)
  }
}
```

This structure is chosen to address following requirements:

| Requirement | How Addressed |
|-------------|---------------|
| **Limit data collection** (#9216) | `limit` explicitly controls collection size. |
| **3rd party reporter access** | Data collected when `limit > 0`, available regardless of `print` setting |
| **UI can display data** | UI has data when `limit > 0`, can toggle visibility independently |
| **CLI output is opt-in** | `print: false` (default) - terminal stays clean |
| **Easy CLI toggle** | `vitest --experimental.importDurations.print` enables terminal output |
| **Broader feature umbrella** (#9141) | `importDurations` namespace can expand to hold other related options |

Regarding data size, reporter side `importDurations` metadata scales by the order`O(test files × source files)` but reporting inherently has limit on how much it can show, so the new option `importDurations.limit` gives a way to control the display and collection limit and it allows capping overall data size to `O(test files x limit)`.

### Tests

### Documentation

### Changesets


packages/vitest/src/runtime/runners/test.ts
```diff
@@ -227,10 +227,25 @@ export class VitestTestRunner implements VitestRunner {
...
+    if (limit === 0) {
+      return {}
+    }
+
+    const entries = [...(this.workerState.moduleExecutionInfo?.entries() || [])]
 
-    for (const [filepath, { duration, selfTime, external, importer }] of entries) {
+    // Trim to top N by duration to reduce IPC payload
+    // Keep enough entries for UI "show more" and aggregation
+    const retention = Math.max(50, limit * 5)
```
sheremet-va: I don't think this makes sense, to be honest. If there is a limit, then there is a limit. Maybe remove "show more" in UI

sheremet-va (changes requested):

packages/vitest/src/node/config/resolveConfig.ts
```diff
@@ -823,6 +823,9 @@ export function resolveConfig(
       resolved.experimental.fsModuleCachePath,
     )
   }
+  resolved.experimental.printImportBreakdown ??= {} as any
+  resolved.experimental.printImportBreakdown.enabled ??= false
+  resolved.experimental.printImportBreakdown.limit ??= 10
```
sheremet-va: Shouldn't it be 0 by default then

I think we can just say "enable this option to see import breakdown" in the current popup

hi-ogawa: Right, that works too. It looks like I was just over-complicating things.

sheremet-va: Yeah, would be nice

packages/vitest/src/runtime/runners/test.ts
```diff
@@ -235,10 +235,21 @@ export class TestRunner implements VitestTestRunner {
   }
 
   getImportDurations(): Record<string, ImportDuration> {
-    const importDurations: Record<string, ImportDuration> = {}
-    const entries = this.workerState.moduleExecutionInfo?.entries() || []
+    const { limit } = this.config.experimental.printImportBreakdown
+    // limit = 0 means opt-out from collection entirely
+    if (limit === 0) {
```
sheremet-va, line 247: I am so incredibly confused by the logic. By default, we _always_ collect it since limit is set to 10 in the config, but never show it? (Including UI)

Why do we need this "set to 0 to disable" note everywhere? just use `.enabled` and throw an error on `limit: 0` to remove any ambiguity

hi-ogawa: > By default, we always collect it since limit is set to 10 in the config, but never show it? (Including UI)

`printImportBreakdown.enabled` controls terminal reporter output and additionally UI default state. UI Hide/show can always toggle. `printImportBreakdown.limit` 10 means we always collect so the data is there, yes.

> just use .enabled and throw an error on limit: 0 to remove any ambiguity

What exact condition to throw? You mean special casing `limit: 0` is useless? (oh well it looks like so).

sheremet-va: > What exact condition to throw? You mean special casing limit: 0 is useless? (oh well it looks like so).

I think we can just throw "invalid positive" error in `resolveConfig` is `limit` is less than 1 or is not an integer. We do similar validation for shards, for example.

sheremet-va: > printImportBreakdown.limit 10 means we always collect so the data is there, yes.

Let's not do that, and just check `enabled`.

sheremet-va: Or introduce two separate flags for collecting/reporting

hi-ogawa: I'm genuinely struggling with coming up with the most intuitive user-facing feature and option structure, so bear with me.

Feature wise, having one flag to control terminal CLI output without influencing UI's ability to show the default limit seems good to have. Such flag is currently called `printImportDurations.enabled`. If this flag affected collection (i.e. enabled=false works as limit=0), then UI display would also require `printImportDurations.enabled`.  For me, `limit` serving as data collection explicitly makes sense to me. I think what could be improved is around the implicitness or naming of `printImportDurations.enabled`.

hi-ogawa: My proposal would be to give explicit flag to control builtin CLI reporting behavior. Also renaming `printImportBreakdown` to something else, but here `importDurations`.

```js
experimental: {
  importDurations?: {
    limit?: number  // Collection limit (default: 10, 0 = disabled)
    print?: boolean // Print to CLI terminal (default: false)
  }
}
```

The further rational is here organized by Claude:

<details>

### How This Addresses Each Requirement

| Requirement | How Addressed |
|-------------|---------------|
| **Limit data collection** (#9216) | `limit` explicitly controls collection size. `limit: 0` disables entirely. |
| **3rd party reporter access** | Data collected when `limit > 0`, available regardless of `print` setting |
| **UI can display data** | UI has data when `limit > 0`, can toggle visibility independently |
| **CLI output is opt-in** | `print: false` (default) - terminal stays clean |
| **Easy CLI toggle** | `vitest --experimental.importDurations.print` enables terminal output |
| **Broader feature umbrella** (#9141) | `importDurations` namespace can expand to hold other related options |

### Explicit Semantics (No Inference)

- `limit` controls **raw data collection** - explicit, `0` = no collection
- `print` controls **CLI terminal output** - independent, just one consumer of the data

No flag infers the other:
- `limit > 0` does NOT imply `print: true`
- `print: true` does NOT imply `limit > 0` (though has no effect if `limit: 0`)

### Default Behavior

```typescript
{ limit: 10, print: false }
```

- **Collects data** for UI and 3rd party reporters (avoids #9216 bloat with limit)
- **Does not print to terminal** (keeps CLI clean by default)
- UI can show/hide via toggle without config changes

### CLI Usage Examples

```bash
# Default: collect for UI, no terminal output
vitest

# Enable terminal output (uses default limit: 10)
vitest --experimental.importDurations.print

# Custom limit + terminal output
vitest --experimental.importDurations.limit=20 --experimental.importDurations.print

# Disable collection entirely (saves memory/bandwidth)
vitest --experimental.importDurations.limit=0
```

### Why Rename from `printImportBreakdown`?

Current name `printImportBreakdown.enabled` is confusing because:
- It sounds like it enables the whole feature
- But it only controls terminal output, not collection

New name `importDurations` is clearer because:
- It's about the `importDurations` data (matches the reporter API field name)
- `limit` = data, `print` = one way to display it
- Namespace can expand for future related options (#9141)

</details>

sheremet-va: I like the proposal with renaming and separating the options

docs/config/experimental.md
```diff
@@ -176,16 +176,32 @@ export default defineConfig({
 It's important that Node can process `sdkPath` content because it is not transformed by Vitest. See [the guide](/guide/open-telemetry) on how to work with OpenTelemetry inside of Vitest.
 :::
 
-## experimental.printImportBreakdown <Version type="experimental">4.0.15</Version> {#experimental-printimportbreakdown}
+## experimental.importDurations <Version type="experimental">4.0.15</Version> {#experimental-importdurations}
```
hi-ogawa:
```suggestion
## experimental.importDurations <Version type="experimental">4.1.0</Version> {#experimental-importdurations}
```

packages/vitest/src/node/config/resolveConfig.ts
```diff
@@ -835,6 +835,9 @@ export function resolveConfig(
       resolved.experimental.fsModuleCachePath,
     )
   }
+  resolved.experimental.importDurations ??= {} as any
+  resolved.experimental.importDurations.limit ??= 10
```
sheremet-va: Shouldn't it be 0 by default? Why are we collecting and not displaying? I thought the consensus after the reporter issue was that it's a waste of space to do that, and we should let the tool configure it.

We can have 10 by default if `print` is enabled, maybe.

hi-ogawa: My take is that `limit = 10` can be a reasonable default, for example, for UI to work with zero config. Otherwise, that use case also need either explicit `limit = 10` (or `print = true` if we infer `print: true -> limit = 10`).

With `limit = 10`, it shouldn't cause huge data size since reporter data size already has O(number of test files) and fixed cap of `limit = 10` per test file won't contribute that much compared to the associated per test file data structure we already hold.

sheremet-va: But what is the point of storing it if user doesn't use UI? Maybe UI plugin should just set the value?

hi-ogawa: For sure, what's collected can be waste, but I think the implicit config dependence can be unintuitive too. I'm probably leaning on the latter, both from our extra logic on config side and also users perspective.

hi-ogawa:
We discussed to make `limit = 0` by default, but additionally we infer based on `print = true --> limit = 10` and `ui = true --> limit = 10`.

PR #9421, opened 2026-01-09
fix(mocker): fix mock transform with class

hi-ogawa:
### Description

The bug is found when investigating Vite 8 https://github.com/vitest-dev/vitest/pull/9415. 

https://github.com/vitest-dev/vitest/blob/f8879a82f95e7640d379e1e4b3ef30dec4ff8f3a/test/core/test/mocking/automocking-class-inheritence.test.ts#L45-L51

The transform scope issue didn't manifest with esbuild ts transform since esbuild was deconflicting by itself by `Bar -> Bar2`. Now this manifested with Oxc since inner `Bar` were kept as is. 

The test case added here is raw js, so it fails without the fix.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/mocker/src/node/esmWalker.ts
+++ b/packages/mocker/src/node/esmWalker.ts
@@ -157,6 +157,17 @@ export function esmWalker(
           identifiers.push([node, parentStack.slice(0)])
         }
       }
+      else if (node.type === 'ClassDeclaration' && node.id) {
+        // A class declaration name could shadow an import, so add its name to the parent scope
+        const parentScope = findParentScope(parentStack)
+        if (parentScope) {
+          setScope(parentScope, node.id.name)
+        }
+      }
+      else if (node.type === 'ClassExpression' && node.id) {
+        // A class expression name could shadow an import, so add its name to the scope
+        setScope(node, node.id.name)
+      }
       else if (isFunctionNode(node)) {
         // If it is a function declaration, it could be shadowing an import
         // Add its name to the scope so it won't get replaced
--- a/test/core/test/injector-mock.test.ts
+++ b/test/core/test/injector-mock.test.ts
@@ -1117,6 +1117,52 @@ const Baz = class extends Foo {}
     `)
   })
 
+  test('classDeclaration shadowing import', () => {
+    const input = `\
+import { Bar } from "./repro.js"
+{
+  class Bar {
+    test() {}
+  }
+  const Zoo = class Zoo extends Bar {}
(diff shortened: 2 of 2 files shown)


test/core/test/injector-mock.test.ts
```diff
@@ -1117,6 +1117,52 @@ const Baz = class extends Foo {}
...
+    expect(hoistSimpleCodeWithoutMocks(input)).toMatchInlineSnapshot(`
+      "vi.mock('faker');
+      const __vi_import_0__ = await import("./foo.js");
+      import {vi} from "vitest";
+
+      const Bar = class Foo {
+        static create() {
+          return new Foo()
+        }
+      }"
```
hi-ogawa, line 1162: Without the fix, this becomes

```js
const Bar = class Foo {
  static create() {
    return new __vi_import_0__.Foo()
  }
}
```

test/core/test/injector-mock.test.ts
```diff
@@ -1117,6 +1117,52 @@ const Baz = class extends Foo {}
...
+    expect(hoistSimpleCodeWithoutMocks(input)).toMatchInlineSnapshot(`
+      "vi.mock('faker');
+      const __vi_import_0__ = await import("./repro.js");
+      import {vi} from "vitest";
+
+      {
+        class Bar {
+          test() {}
+        }
+        const Zoo = class Zoo extends Bar {}
```
hi-ogawa, line 1139: Without fix, this becomes:

```js
class Bar {
  test() {}
}
const Zoo = class Zoo extends __vi_import_0__.Bar {}
```

packages/mocker/src/node/esmWalker.ts
```diff
@@ -157,6 +157,17 @@ export function esmWalker(
...
+        // A class declaration name could shadow an import, so add its name to the parent scope
+        const parentScope = findParentScope(parentStack)
+        if (parentScope) {
+          setScope(parentScope, node.id.name)
+        }
+      }
+      else if (node.type === 'ClassExpression' && node.id) {
+        // A class expression name could shadow an import, so add its name to the scope
+        setScope(node, node.id.name)
+      }
```
hi-ogawa, line 170: Vite SSR transform has this already, so I just copied from there https://github.com/vitejs/vite/blob/964c718a382ff46ec1f906d7d6bc3f135a6dcd3f/packages/vite/src/node/ssr/ssrTransform.ts#L623-L631

PR #9426, opened 2026-01-09
fix: update coverage.include and coverage.exclude

### LINKED ISSUE #9395: Options coverage.include and coverage.exclude do not work as expected
### Describe the bug

Hello! I am migrating from v3 to v4 and found that new coverage include/exclude logic in v4 does not quite match what was working in v3 and seems like a bug.

### Case 1: I want to include all files, but exclude files from top-level

**Coverage options**:
```ts
coverage: {
      include: [
        // all .js,.ts,.tsx files
        '**/*.ts',
      ],
      exclude: [
        // exclude .ts files for top-level
        '*.ts',
      ],
}
```

**Expected behavior**: files inside folders are included in coverage and files on top-level - are not

**Actual behavior**: all files are not included

**Potential fix**: Do not use [`contains` options for `picomatch.isMatch`](https://github.com/vitest-dev/vitest/blob/v4.0.16/packages/vitest/src/node/coverage.ts#L165). When I remove it, I achieve expected behavior.

### Case 2: I want to exclude all dirs with some particular name even if it happens that CWD already includes such name

**Ignore Name**: foo

**CWD**: /Users/iam/code/foo

**Coverage options**:
```ts
coverage: {
      include: [
        // all .js,.ts,.tsx files
        '**/*.ts',
      ],
      exclude: [
        // exclude .ts files for top-level
        '**/foo/**',
      ],
}
```

**Expected behavior**: all .ts files inside CWD are included in coverage, except files that are under folders with name "foo"

**Actual behavior**: all files are not included, because CWD is ignored

**Potential fix**: Use [relative filename with `picomatch.isMatch`](https://github.com/vitest-dev/vitest/blob/v4.0.16/packages/vitest/src/node/coverage.ts#L164). When I use `path.relative(process.cwd(), filename)` instead of `filename`, I achieve expected behavior.

### Reproduction

Case 1: https://stackblitz.com/edit/vitest-dev-vitest-4nte3ynw?file=vitest.config.ts

Case 2: https://stackblitz.com/edit/vitest-dev-vitest-tguz5bfs?file=vitest.config.ts

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    @vitest/coverage-istanbul: 4.0.16 => 4.0.16 
    vite: latest => 7.3.0 
    vitest: 4.0.16 => 4.0.16
```

### Used Package Manager

npm

### Validations

DevJoaoLopes:
### Description

Fixes `coverage.include`/`coverage.exclude` glob matching so it behaves like `test.include`/`test.exclude` (relative to the project root), addressing Vitest v4 regressions reported in #9395.

In v4, matching was done against the full (absolute) filename with `picomatch` using `contains: true` and `ignore`, which made patterns such as `./*.ts` or `**/foo/**` unexpectedly match parts of the absolute path (including the CWD), causing everything to be excluded.

This change:
- Matches include/exclude against the path relative to each project root when the file is inside the project, so top-level-only excludes like `./*.ts` work as expected.
- Avoids excluding the whole project when the CWD contains an excluded segment (e.g. `/.../foo/...`), by using relative paths for in-project files and only using the absolute filename for truly external files.
- Keeps `allowExternal: false` semantics by skipping external files instead of incorrectly failing matches for everything.
- Adjusts the glob cache key to include the root (so multi-project roots don’t share incorrect cached results).

fix #9395

- [ ] Please check [Allow edits by maintainers](https://docs.github.com/en/pull-requests/collaborating-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to make review process faster. Note that this option is not available for repositories that are owned by Github organizations.

### Tests

### Documentation

### Changesets


AriPerkkio (changes requested):
Could you check the failing tests and add new tests to `test/coverage-test` to validate this fix? Or maybe use the `test/coverage-test/test/include-exclude.test.ts`.

DevJoaoLopes:
> Could you check the failing tests and add new tests to `test/coverage-test` to validate this fix? Or maybe use the `test/coverage-test/test/include-exclude.test.ts`.

@AriPerkkio Sure, I apologize for not doing it from the beginning... I'll work on it. 😅

test/coverage-test/test/bundled-sources.test.ts
```diff
@@ -8,7 +8,7 @@ test('bundled code with source maps to originals', async () => {
     include: [normalizeURL(import.meta.url)],
     coverage: {
       reporter: 'json',
-      exclude: ['./utils.ts'],
+      exclude: ['**/utils.ts', '**/test/**'],
```
AriPerkkio:
```suggestion
      exclude: ['./utils.ts', '**/test/**'],
```

test/coverage-test/test/exclude-after-remap.test.ts
```diff
@@ -6,7 +6,7 @@ test('{ excludeAfterRemap: true } should exclude files that come up after remapp
   await runVitest({
     include: [normalizeURL(import.meta.url)],
     coverage: {
-      exclude: ['fixtures/src/pre-bundle/second.ts', './utils.ts'],
+      exclude: ['**/pre-bundle/second.ts', '**/utils.ts', '**/test/**'],
```
AriPerkkio:
```suggestion
      exclude: ['fixtures/src/pre-bundle/second.ts', './utils.ts', '**/test/**'],
```

test/coverage-test/test/exclude-after-remap.test.ts
```diff
@@ -47,7 +47,7 @@ test('{ excludeAfterRemap: true } should exclude uncovered files that come up af
     include: ['fixtures/test/math.test.ts'],
     coverage: {
       include: ['fixtures/src/pre-bundle/**.ts'],
-      exclude: ['fixtures/src/pre-bundle/second.ts'],
+      exclude: ['**/pre-bundle/second.ts'],
```
AriPerkkio:
```suggestion
      exclude: ['fixtures/src/pre-bundle/second.ts'],
```

AriPerkkio (changes requested):

test/coverage-test/test/exclude-after-remap.test.ts
```diff
@@ -26,7 +26,7 @@ test('{ excludeAfterRemap: false } should not exclude files that come up after r
   await runVitest({
     include: [normalizeURL(import.meta.url)],
     coverage: {
-      exclude: ['fixtures/src/pre-bundle/second.ts', './utils.ts'],
+      exclude: ['**/pre-bundle/second.ts', '**/utils.ts', '**/test/**'],
```
AriPerkkio:
```suggestion
      exclude: ['fixtures/src/pre-bundle/second.ts', './utils.ts', '**/test/**'],
```

packages/vitest/src/node/coverage.ts
```diff
@@ -145,31 +145,38 @@ export class BaseCoverageProvider<Options extends ResolvedCoverageOptions<'istan
...
-    // File outside project root with default allowExternal
-    if (this.options.allowExternal === false && roots.every(root => !filename.startsWith(root))) {
-      this.globCache.set(filename, false)
+    const includeGlobs = this.options.include || '**'
+    const excludes = this.options.exclude
 
-      return false
-    }
+    for (const projectRoot of roots) {
+      const relativePath = slash(relative(projectRoot, filename))
```
AriPerkkio, line 163: Doesn't this actually change how defined globs are interpreted when using `--project` filter? 
- When it's not used, globs are relative to `root` ✅ 
-  When it's used, globs are relative to project's `root` ❌ 

```
├── package.json
├── vitest.config.ts
└── packages
    ├── one
    │   ├── src
    │   │   └── math.ts
    │   └── test
    │       └── math.test.ts
    └── two
        ├── src
        │   └── get-user.ts
        └── test
            └── get-user.test.ts
```

```ts

export default defineConfig({
  test: {
    coverage: {
      include: ["./packages/**/*.ts"],
    },

    projects: [
      {
        test: {
          name: "One",
          root: "./packages/one",
        },
      },
      {
        test: {
          name: "Two",
          root: "./packages/two",
        },
      },
    ],
  },
});
```

```
> vitest --run --coverage

 RUN  v4.0.16 /x/examples/packages/vitest-example
      Coverage enabled with v8

 ✓  Two  test/get-user.test.ts (1 test) 1ms
 ✓  One  test/math.test.ts (1 test) 1ms

 Test Files  2 passed (2)
      Tests  2 passed (2)
   Start at  10:12:31
   Duration  115ms (transform 32ms, setup 0ms, import 43ms, tests 2ms, environment 0ms)

 % Coverage report from v8
--------------|---------|----------|---------|---------|-------------------
File          | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
--------------|---------|----------|---------|---------|-------------------
All files     |      40 |      100 |      40 |      40 |                   
 one/src      |      25 |      100 |      25 |      25 |                   
  math.ts     |      25 |      100 |      25 |      25 | 6-14              
 two/src      |     100 |      100 |     100 |     100 |                   
  get-user.ts |     100 |      100 |     100 |     100 |                   
--------------|---------|----------|---------|---------|-------------------
```

This should show coverage for `packages/one`, but it's missing:

```
> vitest --run --coverage --project one

 RUN  v4.0.16 /x/examples/packages/vitest-example
      Coverage enabled with v8

 ✓  One  test/math.test.ts (1 test) 1ms
   ✓ sum 0ms

 Test Files  1 passed (1)
      Tests  1 passed (1)
   Start at  10:12:50
   Duration  99ms (transform 14ms, setup 0ms, import 19ms, tests 1ms, environment 0ms)

 % Coverage report from v8
All files |       0 |        0 |       0 |       0 |                   
```

packages/vitest/src/node/coverage.ts
```diff
@@ -145,31 +145,38 @@ export class BaseCoverageProvider<Options extends ResolvedCoverageOptions<'istan
...
 
-    const included = pm.isMatch(filename, glob, {
-      contains: true,
-      dot: true,
-      ignore: this.options.exclude,
-    })
+      const matchTarget = isExternal ? filename : relativePath
 
-    this.globCache.set(filename, included)
+      if (excludes.length && pm.isMatch(matchTarget, excludes, { dot: true })) {
```
AriPerkkio: `excludes` is always defined array.

```suggestion
      if (pm.isMatch(matchTarget, excludes, { dot: true })) {
```

packages/vitest/src/node/coverage.ts
```diff
@@ -145,31 +145,38 @@ export class BaseCoverageProvider<Options extends ResolvedCoverageOptions<'istan
...
+        continue
+      }
+
+      if (pm.isMatch(matchTarget, includeGlobs, { dot: true, contains: true })) {
```
AriPerkkio, line 183: Why do we need 2x `pm.isMatch` calls, instead of doing just one with passing `ignore: options.exclude`? Any examples that would fail without 2x calls?

test/coverage-test/test/bundled-sources.test.ts
```diff
@@ -8,7 +8,7 @@ test('bundled code with source maps to originals', async () => {
     include: [normalizeURL(import.meta.url)],
     coverage: {
       reporter: 'json',
-      exclude: ['./utils.ts'],
+      exclude: ['./utils.ts', '**/test/**'],
     },
   })
```
AriPerkkio, line 14: Looks like you needed to add `**/test/**` here to avoid `test/coverage-test/test/bundled-sources.test.ts` being showed up here. I think that's actually bug.

Debugging this case further shows that `isIncluded` is falsely returning `true` for `isIncluded('/x/vitest/test/coverage-test/test/bundled-sources.test.ts')` even when `this.options.excludes` contains following:

```ts
[
  './utils.ts',
  '/x/vitest/test/coverage-test/test/bundled-sources.test.ts',
  '/x/vitest/test/coverage-test/fixtures/configs/vitest.config.ts',
  'vitest.config.ts',
  'vitest.config.mts',
  'vitest.config.cts',
  'vitest.config.js',
  'vitest.config.mjs',
  'vitest.config.cjs',
  'vite.config.ts',
  'vite.config.mts',
  'vite.config.cts',
  'vite.config.js',
  'vite.config.mjs',
  'vite.config.cjs',
  '**/virtual:*',
  '**/__x00__*',
  '**/node_modules/**'
]
```


```suggestion
      exclude: ['./utils.ts'],
```

AriPerkkio (changes requested):

packages/vitest/src/node/coverage.ts
DevJoaoLopes: @AriPerkkio 

You’re right that we call pm.isMatch twice, but they serve different purposes and need distinct options:

First call (`excludes`, `{ dot: true }`): early-out for any path matching `coverage.exclude`, regardless of how includes are written. This prevents excluded files from slipping through when an include pattern is broad.
Second call (`includeGlobs`, `{ dot: true, contains: true }`): only after passing exclusion do we check if the file qualifies for `coverage.include`. The `contains: true `is important for patterns like `"src/utils"` that should match nested paths (e.g., `src/utils/math.ts`), which a strict match would miss.
Dropping either call breaks behavior: without the exclude check, excluded files (e.g.,` __mocks__`) get counted; without the include check (with `contains: true`), legitimate files under broad includes are skipped.

test/coverage-test/test/bundled-sources.test.ts
DevJoaoLopes: Following the suggestion below, I'm getting a timeout error: 

```ts
await runVitest({
    // include: [normalizeURL(import.meta.url)],
    // coverage: {
    //   reporter: 'json',
    //   exclude: ['./utils.ts', '**/test/**'],
    // },
    exclude: ['./utils.ts'],
  })
  ```
  
  Error:
  
  
image

packages/vitest/src/node/coverage.ts
DevJoaoLopes: Okay, I will evaluate these cases.

test/coverage-test/test/bundled-sources.test.ts
AriPerkkio: Don't remove `include` and `coverage`, just change the `coverage.exclude` back to `exclude: ['./utils.ts']`.

AriPerkkio (commented):
Thanks for the PR @DevJoaoLopes! We will finalize the fix in https://github.com/vitest-dev/vitest/pull/9818. 🙌

PR #9514, opened 2026-01-23
feat(browser): use BlazeDiff instead of pixelmatch

### LINKED ISSUE #9431: Use a faster comparator for image diff

As a developer using VItest, I want to have the fastest image diff out of the box.

I've been digging into the image diff problem for a very long time. I made [BlazeDiff](https://github.com/teimurjan/blazediff), which is beating all existing OSS image diffs and is already widely adopted. [`@blazediff/core`](https://www.blazediff.dev/docs/core)  (pure JavaScript approach) is ~50% faster than pixelmatch, using the same API and producing the same results. It weighs twice as much, is actively supported, and is type-safe out of the box. You can check the benchmarks here: https://github.com/teimurjan/blazediff/blob/main/BENCHMARKS.md.

### Suggested solution

I want to add a new `blazediff` comparator.

### Alternative

As an alternative, I can replace the existing one (due to 1-1 identity in API & results).

### Additional context

BlazeDiff is battle-tested and already running visuals for:
- https://github.com/ant-design/ant-design
- https://github.com/ant-design/x
- https://github.com/antvis/G
- https://github.com/antvis/GPT-Vis
- https://github.com/Shopify/react-native-skia
- https://github.com/vega/vega
- https://github.com/apexcharts/apexcharts.js

### Validations

macarie:
### Description

Resolves #9431

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/api/browser/assertions.md
+++ b/docs/api/browser/assertions.md
@@ -1152,10 +1152,9 @@ await expect.element(getByTestId('button')).toMatchScreenshot('fancy-button', {
 
 - `comparatorName: "pixelmatch" = "pixelmatch"`
 
-  The name of the algorithm/library used for comparing images.
+  The algorithm/library used for comparing images.
 
-  Currently, [`"pixelmatch"`](https://github.com/mapbox/pixelmatch) is the only
-  supported comparator.
+  `"pixelmatch"` is the only built-in comparator, but you can use custom ones by [registering them in the config file](/config/browser/expect#browser-expect-tomatchscreenshot-comparators).
 
 - `comparatorOptions: object`
 
@@ -1210,7 +1209,7 @@ await expect.element(getByTestId('button')).toMatchScreenshot('fancy-button', {
 
 #### `"pixelmatch"` comparator options
 
-The following options are available when using the `"pixelmatch"` comparator:
+The `"pixelmatch"` comparator uses [`@blazediff/core`](https://blazediff.dev/docs/core) under the hood. The following options are available when using it:
 
 - `allowedMismatchedPixelRatio: number | undefined = undefined`
 
--- a/packages/browser/package.json
+++ b/packages/browser/package.json
@@ -64,10 +64,10 @@
     "vitest": "workspace:*"
   },
   "dependencies": {
+    "@blazediff/core": "1.9.1",
     "@vitest/mocker": "workspace:*",
     "@vitest/utils": "workspace:*",
     "magic-string": "catalog:",
-    "pixelmatch": "7.1.0",
     "pngjs": "^7.0.0",
     "sirv": "catalog:",
     "tinyrainbow": "catalog:",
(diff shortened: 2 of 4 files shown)


macarie:
Not sure if we should add another `"comparatorName"` literal (`'blazediff'` or something) or not. On the one hand, it's not pixelmatch anymore, although compatible, on the other hand BlazeDiff supports multiple algorithms, so it might be confusing.

userquin:
Maybe we can do the same Vlad did with the browser provider and move them to vitest community organization. This way we remove another dependency here. For blazediff we can include some options.

userquin:
Ohhh, browser providers are here, we can just move them to a new package here and so the user should configure one of them (or use a factory) same way we had with the provider names.

https://github.com/vitest-dev/vitest/pull/8629

sheremet-va:
> Ohhh, browser providers are here, we can just move them to a new package here and so the user should configure one of them (or use a factory) same way we had with the provider names.
> 
> #8629

We discussed this during the last team call on Friday and decided to just replace it with BlazeDiff since there is seemingly no downsides.

docs/api/browser/assertions.md
```diff
@@ -1150,12 +1150,13 @@ await expect.element(getByTestId('button')).toMatchScreenshot('fancy-button', {
...
 
-- `comparatorName: "pixelmatch" = "pixelmatch"`
+- `comparatorName: "@blazediff/core" | "pixelmatch" = "@blazediff/core"`
 
-  The name of the algorithm/library used for comparing images.
+  The algorithm/library used for comparing images.
 
-  Currently, [`"pixelmatch"`](https://github.com/mapbox/pixelmatch) is the only
-  supported comparator.
+  [`"@blazediff/core"`](https://blazediff.dev/docs/core) is the only built-in comparator, but you can use custom ones by [registering them in the config file](/config/browser#browser-expect-toMatchScreenshot-comparators).
```
macarie: Should we link the repository or the website? 🤔

docs/api/browser/assertions.md
```diff
@@ -1133,15 +1133,15 @@ await expect.element(getByTestId('button')).toMatchScreenshot('fancy-button')
 
 // with options
 await expect.element(getByTestId('button')).toMatchScreenshot({
-  comparatorName: 'pixelmatch',
+  comparatorName: '@blazediff/core',
```
sheremet-va: Shouldn't it just be called `blazediff`?

sheremet-va (changes requested):

docs/api/browser/assertions.md
macarie: There's also [`@blazediff/ssim`](https://www.blazediff.dev/docs/ssim) and [`@blazediff/gmsd`](https://www.blazediff.dev/docs/gmsd). Thought it would create less confusion using the full library name as these algorithms don't really have a specific name.

Alternatively we could use the "strategy", something like `"pixel-by-pixel"`, keeping the library information hidden from the end users.

sheremet-va: I like using a "strategy" approach

sheremet-va (approved):

AriPerkkio (approved):

PR #9521, opened 2026-01-25
feat(coverage): add `coverage.changed` option to report only changed files

### LINKED ISSUE #8747: Run all tests, but report partial coverage, for CI purposes

This is quite common need at CI pipeline that these two objectives need to be fulfilled:
1) run all tests
2) compute coverage of a pull request

The problem is that the second command reruns the same tests. When we talk about CI pipeline which does these two jobs, this is a lot of wasted time.

### Suggested solution

My request is to add functionality to limit only collecting the coverage to what `--changed` (or similar) flag says, and still run the full test set. This could be e.g. `--changed-coverage` param.

#### First

Currently this can be achieved by two separate commands, but takes up to twice as much time depending on changes:
1) run all tests: `vitest run`
2) compute coverage of a pull request: `vitest run --coverage --changed=origin/main`

#### Second
Compute coverage of all the project. Unfortunately this blurs the result as it includes a lot of files which are unrelated to pull request. This also is a problem at legacy projects which are in the process of increasing test coverage as unrelated files which may have no coverage are mixed with affected files with missing coverage.

### Additional context

This is performance improvement for CI pipelines which use vitest

### Validations

kykim00:
### Description

This PR adds a new `coverage.changed` option that allows running all tests while only computing coverage for changed files. This is useful for CI pipelines that need to:

1. Run the full test suite to ensure all tests pass
2. Compute coverage only for files changed in a pull request

Previously, this required running two separate commands, which doubled the test execution time.

Resolves #8747 

### Usage

```typescript
export default defineConfig({
  test: {
    coverage: {
      changed: 'HEAD', // or any git ref like 'main', 'origin/main'
    },
  },
})
```

The `changed` option accepts:
- `'HEAD'` - Compare against the last commit
- `'main'` - Compare against the main branch
- `'origin/main'` - Compare against remote main branch
- Any valid git reference

### Testing Instructions

1. Navigate to the test directory:
   ```bash
   cd test/coverage-test
   ```

2. Create a demo config file `vitest.demo.config.ts`:
   ```typescript

   export default defineConfig({
     test: {
       include: ['**/file-to-change.test.ts', '**/math.test.ts'],
       coverage: {
         enabled: true,
         provider: 'istanbul',
         reporter: ['text'],
         reportsDirectory: './coverage-demo',
       },
     },
   })
   ```

3. Run tests **without** the `changed` option:
   ```bash
   pnpm vitest run --config=vitest.demo.config.ts
   ```
   
   **Result:** Coverage includes all files
   ```
   File               | % Stmts | % Branch | % Funcs | % Lines |
   -------------------|---------|----------|---------|---------|
   All files          |   33.33 |      100 |   33.33 |   33.33 |
    file-to-change.ts |      50 |      100 |      50 |      50 |
    math.ts           |      25 |      100 |      25 |      25 |
   ```

4. Modify `fixtures/src/file-to-change.ts` (make any change)

5. Update config to add `changed: 'HEAD'`:
   ```typescript
   coverage: {
     enabled: true,
     provider: 'istanbul',
     reporter: ['text'],
     reportsDirectory: './coverage-demo',
     changed: 'HEAD', // ← Add this line
   },
   ```

6. Run tests again:
   ```bash
   pnpm vitest run --config=vitest.demo.config.ts
   ```
   
   **Result:** Coverage includes only changed files
   ```
   File               | % Stmts | % Branch | % Funcs | % Lines |
   -------------------|---------|----------|---------|---------|
   All files          |      50 |      100 |      50 |      50 |
    file-to-change.ts |      50 |      100 |      50 |      50 |
   ```

Note: Both test runs execute **all tests** (`file-to-change.test.ts` and `math.test.ts`), but the second run only reports coverage for the modified file.

### Tests

### Documentation

### Changesets


packages/vitest/src/node/core.ts
```diff
@@ -1361,7 +1367,26 @@ export class Vitest {
...
+    const changedFiles = await vitestGit.findChangedFiles({
+      changedSince: coverageChanged,
+    })
+    if (!changedFiles) {
+      process.exitCode = 1
+      throw new GitNotFoundError()
+    }
+    reportContext.changedFiles = Array.from(new Set(changedFiles))
+    return reportContext
+  }
```
AriPerkkio: I think this logic should be here instead:

https://github.com/vitest-dev/vitest/blob/7ce3417b1be71689a784cc7a717bd0bce18e39a3/packages/vitest/src/node/coverage.ts#L77

packages/vitest/src/node/coverage.ts
```diff
@@ -330,11 +347,15 @@ export class BaseCoverageProvider<Options extends ResolvedCoverageOptions<'istan
...
-      (coverageMap as CoverageMap) || this.createCoverageMap(),
-      allTestsRun,
-    )
+  async reportCoverage(coverageMap: unknown, reportContext: ReportContext): Promise<void> {
+    this.setReportContext(reportContext)
+    const finalCoverageMap = (coverageMap as CoverageMap) || this.createCoverageMap()
+
+    if (this.changedFiles) {
+      finalCoverageMap.filter(filename => this.changedFiles!.has(slash(filename)))
+    }
```
AriPerkkio, line 352: Filtering at this point will be more expensive than filtering before remapping is done, e.g. in `isIncluded`.

AriPerkkio (changes requested):
Initial approach looks good, but I would rather see all this logic being added into `BaseCoverageProvider` so that it doesn't leak into `core.ts` for example.

In the beginning of `reportCoverage`:

https://github.com/vitest-dev/vitest/blob/7ce3417b1be71689a784cc7a717bd0bce18e39a3/packages/vitest/src/node/coverage.ts#L333

... we would do something like

```ts
if(this.options.changed) {
  this.changedFiles = this.getChangedFiles()
}
```

The `getChangedFiles` would use `VitestGit` like you've done now in `core.ts`. Then in `isIncluded` and `getUntestedFilesByRoot` we would be checking the `this.changedFiles`. 

Would that work?

kykim00:
Sorry for the late response! That makes sense, I'll move the logic into `BaseCoverageProvider` and filter earlier in `isIncluded`. Will push the changes shortly.

kykim00:
@AriPerkkio I updated the code to move the logic into `BaseCoverageProvider` and filtering now happens in `isIncluded`. Also fixed query-param/@fs normalization and added tests for remap + query-param cases. Please take another look when you get a chance!

packages/vitest/src/node/types/coverage.ts
```diff
@@ -67,6 +67,8 @@ export interface CoverageProvider {
 export interface ReportContext {
   /** Indicates whether all tests were run. False when only specific tests were run. */
   allTestsRun?: boolean
+  /** Absolute paths for files changed since a given commit/branch. */
+  changedFiles?: string[]
```
```suggestion
```

AriPerkkio (commented):
Sorry haven't had time to review this. Implementation looks a bit too complex atm. I'll try to find time to review and push changes directly there.

test/coverage-test/test/query-param-transforms.test.ts
```diff
@@ -46,3 +48,52 @@ test('query param based transforms are resolved properly', async () => {
     ]
   `)
 })
+
+test('query param transforms respect coverage.changed', async () => {
```
AriPerkkio: Oh good catch! Looks like this is also broken for existing `{ test: { changed: 'HEAD' } }`. I'll extend the test to cover both options. 💯

AriPerkkio (approved):
Looks good to me. Great work @kykim00! 💯 

@sheremet-va / @hi-ogawa any thoughts about the new `onTestRunStart` hook? If not, please approve. 🙏

packages/vitest/src/node/coverage.ts
```diff
@@ -324,6 +329,23 @@ export class BaseCoverageProvider<Options extends ResolvedCoverageOptions<'istan
...
+    if (this.ctx.config.changed) {
+      this.changedFiles = this.ctx.config.related
+    }
+    else if (this.options.changed) {
+      const { VitestGit } = await import('./git')
+      const vitestGit = new VitestGit(this.ctx.config.root)
+      const changedFiles = await vitestGit.findChangedFiles({ changedSince: this.options.changed })
+
+      this.changedFiles = changedFiles ?? undefined
+    }
```
hi-ogawa: Is this if-else order correct? From the test `coverage.changed inherits from test.changed but can be overridden`, it looks like `this.options.changed` should be respected over `this.ctx.config.changed`.

hi-ogawa (commented):
`onTestRunStart` looks good to me 👍

packages/vitest/src/node/coverage.ts
AriPerkkio: Good point! We should add this test to `changed.test.ts` too. Right now we are only testing this case in CLI option overrides.

test/coverage-test/test/changed.test.ts
AriPerkkio, line 1: Add test for https://github.com/vitest-dev/vitest/pull/9521#discussion_r2810335643 and apply fix if needed.

AriPerkkio (changes requested):

packages/vitest/src/node/coverage.ts
AriPerkkio: Fixed in https://github.com/vitest-dev/vitest/pull/9521/changes/aa4c3a2329869b248eac412b38595dc8a81d2b93. The test case failed initially without the fix. ✅

hi-ogawa (approved):

AriPerkkio (approved):

PR #9525, opened 2026-01-26
feat(browser): added `detailsPanelPosition` option and button

shairez:
### Description

This PR adds a new `detailsPanelPosition` configuration option for browser mode that allows users to control the layout of the details panel in the Vitest Browser Mode. 

Users can now choose between a horizontal split (right side, default) or a vertical split (bottom) for the details panel, providing more flexibility for different screen sizes and preferences.

**Key changes:**
- Added `detailsPanelPosition` option to browser config with values `'right'` (default) or `'bottom'`
- Implemented UI support for both horizontal and vertical panel layouts (a "switch layout" button)
- Added CLI option `--browser.detailsPanelPosition` to override config values
- Refactored `DetailsPanel` component to `ResultsPanel` for better naming clarity
- Added tests for the new configuration option
- Added docs for `detailsPanelPosition`
- Updated documentation link in CONTRIBUTING.md to Browser Mode development guide

The feature maintains backward compatibility by defaulting to `'right'` (the existing behavior) and respects user preferences stored in localStorage when the config option is not explicitly set.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/ui/client/composables/navigation.ts
+++ b/packages/ui/client/composables/navigation.ts
@@ -1,7 +1,7 @@
 import type { File, Task } from '@vitest/runner'
 import type { Params } from './params'
 import { useLocalStorage, watchOnce } from '@vueuse/core'
-import { computed, reactive, ref, watch } from 'vue'
+import { computed, nextTick, reactive, ref, watch } from 'vue'
 import { viewport } from './browser'
 import { browserState, client, config, findById } from './client'
 import { testRunState } from './client/state'
@@ -34,6 +34,31 @@ export const detailSizes = useLocalStorage<[left: number, right: number]>(
     67,
   ],
 )
+
+export const detailsPanelVisible = useLocalStorage<boolean>(
+  'vitest-ui_details-panel-visible',
+  true,
+)
+
+export const detailsPosition = ref<'right' | 'bottom'>('right')
+
+nextTick(() => {
+  watch(config, () => {
+    if (config.value?.browser?.detailsPanelPosition) {
+      detailsPosition.value = config.value.browser.detailsPanelPosition
+    }
+  })
+})
+
+export function hideDetailsPanel() {
+  // setTimeout is used to avoid splitpanes throwing a race condition error
+  setTimeout(() => {
+    detailsPanelVisible.value = false
+  }, 0)
+}
+export function showDetailsPanel() {
+  detailsPanelVisible.value = true
+}
 
 // live sizes of panels in percentage
 export const panels = reactive({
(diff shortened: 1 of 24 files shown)


packages/ui/client/composables/navigation.ts
```diff
@@ -35,6 +35,31 @@ export const detailSizes = useLocalStorage<[left: number, right: number]>(
...
+  'right',
+)
+
+// Config overrides localStorage when set
+// If config specifies detailsPanelPosition, it always takes precedence
+// If config doesn't set it, localStorage value (user preference) is used
+// Defer watch setup to avoid accessing config before initialization
+nextTick(() => {
+  watchOnce(
+    () => config.value?.browser?.detailsPanelPosition,
```
sheremet-va: Is it a browser only setting? Or does it affect UI without the browser?

local storage in browser tests will reset between sessions so there is not a lot of reason to use it there

If it's for UI, then the setting shouldn't be in `browser` config

shairez:
> We have UI tests in `test/ui`. They are also failing on CI, can you look into that?

@sheremet-va yes I found the issue and fixed it, seems like there are other tests failing not related to my changes (saw these fail in other PRs as well)

packages/ui/client/composables/navigation.ts
shairez: well, it's just for when the browser is opened and you want to have more space on the right for the component preview. 

I saw that everything else is also using localstorage, that's why I added

Not sure where to put this configuration option if not in "browser"

shairez: I changed the local storage to a plan `ref` (also fixed a bug along the way)

sheremet-va: > I saw that everything else is also using localstorage, that's why I added

That's for UI used separately from the browser mode

> Not sure where to put this configuration option if not in "browser"

If it's only for the browser, then UI should ignore it unless it's in browser mode (I don't think it even makes sense without browser mode?) (I didn't check if that's the case already)

shairez: OK, so it's only for browser mode, so it's from what I understand, it's OK to leave this config option under `browser` config right?

sheremet-va: If it’s only for the browser, then it’s fine to leave it there, but ui in normal mode should ignore it (imagine a config with this option, but disabled browser and enabled ui)

shairez: AFAIK checking if it's in "UI" normal mode is `!browserState` right?

In that case, non of my changes affect the normal UI, they only affect the browser mode (I already removed the local storage saving so this code example isn't relevant anymore)

so I think this could be resolved, unless I'm missing something

sheremet-va: The buttons are rendered and do nothing in `ui` mode for regular Node tests (`pnpm test --ui`)

image

shairez: good catch, I've pushed a fix

sheremet-va:
Thank you! LGTM 🙏

PR #9528, opened 2026-01-26
feat: add `--detect-async-leaks`

### LINKED ISSUE #3119: Detect hanging async operations

AriPerkkio:
### Description

- Resolves https://github.com/vitest-dev/vitest/issues/3119
- Closes https://github.com/vitest-dev/vitest/pull/5395

Highly inspired by
- https://github.com/mafintosh/why-is-node-running
- https://jestjs.io/docs/cli#--detectopenhandles

### Tests

### Documentation

### Changesets


test/cli/test/detect-async-leaks.test.ts
```diff
@@ -0,0 +1,76 @@
...
+     ❯ packages/example/test/example.test.ts:5:9
+      3| 
+      4|       test('leaking timeout', () => {
+      5|         source()
+       |         ^
+      6|       })
+      7|     
+
+    "
+  `)
```
AriPerkkio, line 72: Discussed in team meeting that we show just the last frame:


```suggestion
  expect(stderr).toMatchInlineSnapshot(`
    "
    ⎯⎯⎯⎯⎯⎯⎯ Async leaks 1 ⎯⎯⎯⎯⎯⎯⎯⎯

     Timeout  packages/example/test/example.test.ts
     ❯ source packages/example/src/source.ts:3:9
      1| 
      2|       export default function source() {
      3|         setTimeout(() => {}, 100_000)
       |         ^
      4|       }
      5|     
     ❯ packages/example/test/example.test.ts:5:9
    "
  `)
```

packages/vitest/src/node/reporters/base.ts
```diff
@@ -776,6 +785,45 @@ export abstract class BaseReporter implements Reporter {
...
+
+      this.ctx.logger.error(c.red(`${leak.type} leaking in ${filename}`))
+
+      const stacks = parseStacktrace(leak.stack)
+
+      if (stacks.length === 0) {
+        continue
+      }
+
+      const sourceCode = readFileSync(stacks[0].file, 'utf-8')
```
sheremet-va: I would add try/catch here just in case

sheremet-va (dismissed):

test/cli/test/detect-async-leaks.test.ts
```diff
@@ -0,0 +1,283 @@
...
+      6|         await new Promise(resolve => app.listen({ host: "localhost", port:…
+       |                                          ^
+      7|       })
+      8|
+     ❯ packages/example/test/example.test.ts:6:42
+     ❯ packages/example/test/example.test.ts:6:15
+
+    "
+  `)
+})
```
AriPerkkio: This one seems to produce different errors on Windows. I'll try to debug from CI what the error is 👀

AriPerkkio:
Windows CI passing tests now too ✅

docs/config/detectasyncleaks.md
```diff
@@ -0,0 +1,14 @@
...
+outline: deep
+---
+
+# detectAsyncLeaks
+
+- **Type:** `boolean`
+- **CLI:** `--detectAsyncLeaks`, `--detect-async-leaks`
+- **Default:** `false`
+
+Detect asynchronous resources leaking from the test file.
```
sheremet-va, line 16: Can we elaborate more here of what this option does? Maybe have a link to Node.js docs too.

I think it would also be nice to mention this in "guide/features.md".

packages/vitest/src/node/config/resolveConfig.ts
```diff
@@ -354,6 +354,10 @@ export function resolveConfig(
     throw new Error(`"Istanbul" coverage provider is not compatible with "experimental.viteModuleRunner: false". Please, enable "viteModuleRunner" or switch to "v8" coverage provider.`)
   }
 
+  if (browser.enabled && resolved.detectAsyncLeaks) {
+    logger.console.warn(c.yellow('The option "detectAsyncLeaks" is not supported in browser mode and will be ignored.'))
```
sheremet-va, line 358: So if you do `vitest --detectAsyncLeaks`, have both node and browser tests, then you will always see this log? (just to confirm)

test/config/test/failures.test.ts
```diff
@@ -391,6 +391,16 @@ test('maxConcurrency 0 prints a warning', async () => {
   expect(stderr).toMatch('The option "maxConcurrency" cannot be set to 0. Using default value 5 instead.')
 })
 
+test('detectAsyncLeaks with browser mode prints a warning', async () => {
+  const { stderr } = await runVitest({
+    include: [],
```
sheremet-va, line 396: Since it's a warning, vitest might run tests here because `include` doesn't override config's `include` (I found that out when moving tests). Adding `config: false` here should guarantee it doesn't run tests (it might be that it already picks up a config that has no include, not sure). Could you confirm?

sheremet-va (dismissed):

test/config/test/failures.test.ts
AriPerkkio: Adding `include` here stopped headful browser from being started.

docs/config/detectasyncleaks.md
AriPerkkio: Added more docs in https://github.com/vitest-dev/vitest/pull/9528/changes/6c9c5c4605aac538a4211c58bdedc41275ef0669. Did not add mention to `features.md` though.

packages/vitest/src/node/config/resolveConfig.ts
AriPerkkio: Yep, I think it works correct now:

test/config/test/failures.test.ts
AriPerkkio: Replacing `include: []` with `config: false` here seems to start running tests and opening headful browser. I'll leave the empty include here.

packages/vitest/src/runtime/detect-async-leaks.ts
```diff
@@ -0,0 +1,85 @@
...
+
+interface PossibleLeak extends AsyncLeak {
+  isActive: () => boolean
+}
+
+const IGNORED_TYPES = new Set([
+  'DNSCHANNEL',
+  'ELDHISTOGRAM',
+  'PerformanceObserver',
+  'PROMISE',
```
AriPerkkio: Let's also track unresolved promises:


```suggestion
```

Though this requires
- https://github.com/vitejs/vite/pull/21585

... as our async wrapper in `node:vm` causes negative columns in stacktraces.

AriPerkkio: Tracking promises now too. The bug on module runner is not a blocker as we can do `try catch` to suppress the error.

hi-ogawa:
I tested some random examples locally and I noticed simple `fetch` is detected as leak with `STREAM_END_OF_STREAM`.  Is this  a known limitation? This can be also partially seen in `pnpm -C examples/fastify/ test --detectAsyncLeaks`:

```js
STREAM_END_OF_STREAM leaking in test/app.test.ts
 34|   const port = typeof address === 'string' ? address : address?.port
 35| 
 36|   const response = await fetch(`http://localhost:${port}/users`).then(r => r.json())
   |                                                                  ^
 37| 
 38|   expect(response).toHaveLength(4)
 ❯ test/app.test.ts:36:66
```

Also some cases in `pnpm test --detectAsyncLeaks` looks like it's flagging Vitest's API such as `annotate` and mocking. It looks like this happens only on `test:threads` and not on `test:forks`.

AriPerkkio:
Looking into that `STREAM_END_OF_STREAM` and I have no idea how to track if it was closed 🤷

hi-ogawa:
Does that mean prior arts (jest, why-is-node-running) also flag STREAM_END_OF_STREAM from fetch? I don't think these are blockers but I wonder we may discover more edge cases in the future.

hi-ogawa (dismissed):

AriPerkkio:
I'm not sure how Jest and `why-is-node-running` handle it. 🤔 

I've now pushed a change that makes `STREAM_END_OF_STREAM` ignored so it won't show up as leak. In future we can enable it back if a fix is found.

hi-ogawa (approved):
Ah, didn't know there's already `IGNORED_TYPES` list. That looks good to me.

AriPerkkio:
> I'm not sure how Jest and `why-is-node-running` handle it. 🤔

Okey - with Jest it always prints warning about that code. Though it's `TLSWRAP`. 🤷 






With the current changes in Vitest at least this wouldn't print any errors.

sheremet-va (approved):

comatory:
Is it possible to have `--detect-async-leaks` metadata in external file, if I use `--reporter=json` and `--outFile`? Or any other way to output it as a structured data. Thanks 🙏

PR #9550, opened 2026-01-29
feat: support type inference with a new `test.extend` syntax

### LINKED ISSUE #9305: Support scoped fixtures with dependencies defined in the outer scope

Currently, vitest doesn't appear to support scoped fixtures that depend on fixtures defined in the outer scope.

```ts
import { test as baseTest, describe, expect } from "vitest";

interface Fixtures {
    a: number,
    b: number,
    numbers: number[]
}

const test = baseTest.extend<Fixtures>({
    a: 1,
    b: 2,
    numbers: async ({ a }, use) => use([a]),
});

describe("suite fails", () => {
    test.scoped({
        numbers: async ({ a, b }, use) => use([a, b]),
    });

    test("test fails", async ({ numbers }) => {
        expect(numbers).toStrictEqual([1, 2]); // fails because numbers is [undefined, undefined]
    });
});
```

If I redefine the `a` and `b` fixtures in the narrower scope, then the test passes.

```ts
describe("suite passes", () => {
    test.scoped({
        // setting the dependencies resolves the issue
        a: 1,
        b: 2,
        numbers: async ({ a, b }, use) => use([a, b]),
    });

    test("test passes", async ({ numbers }) => {
        expect(numbers).toStrictEqual([1, 2]);
    });
});
```

While this is a valid workaround, it can be quite tedious to redefine fixtures that have already been defined in the outer scope. This is especially true for large dependency graphs where every upstream fixture would have to be redefined. 

### Suggested solution

Scoped fixtures should be able to use any of the fixtures that were previously defined in the outer scope. (This behaviour is in-line with pytest.)

### Alternative

An alternative would be to have users adopt the following approach:

```ts
import { test as baseTest, describe, expect } from "vitest";

interface Fixtures {
    a: number,
    b: number,
    numbers: number[]
}

const fixtures = {
    a: 1,
    b: 2,
    numbers: async ({ a }, use) => use([a]),
};

const test = baseTest.extend<Fixtures>(fixtures);

describe("suite passes", () => {
    test.scoped({
        ...fixtures,
        numbers: async ({ a, b }, use) => use([a, b]),
    });

    test("test passes", async ({ numbers }) => {
        expect(numbers).toStrictEqual([1, 2]);
    });
});
```

### Additional context

_No response_

### Validations

### LINKED ISSUE #9555: Two independent file-scoped `test.extend` with same key get mixed
### Describe the bug

When two independent `test.extend` calls define file-scoped fixtures with the same key, the second one silently reuses the first's cached value instead of running its own fixture.

```ts
const myTest1 = test.extend<{ value: string }>({
  value: [async ({}, use) => { await use('first-value') }, { scope: 'file' }]
})

const myTest2 = test.extend<{ value: string }>({
  value: [async ({}, use) => { await use('second-value') }, { scope: 'file' }]
})

myTest1('test1', ({ value }) => {
  expect(value).toBe('first-value') 
})

myTest2('test2', ({ value }) => {
  expect(value).toBe('first-value') // should be `second-value`
})
```

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-cmeatr7q?file=test%2Frepro.test.ts

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    vite: latest => 7.3.1 
    vitest: beta => 4.1.0-beta.2
```

### Used Package Manager

npm

### Validations

sheremet-va:
### Description

This PR adds new `test.extend` syntax that supports type inference:

```ts

import { test as baseTest } from 'vitest'

const test = baseTest
  .extend('config', { scope: 'worker' }, async () => {
    return { port: 3000, host: 'localhost' }
  })
  .extend('database', { scope: 'file' }, async ({ config }, { onCleanup }) => {
    const db = new Database(`${config.host}:${config.port}`)
    await db.connect()

    onCleanup(async () => {
      await db.disconnect()
    })

    return db
  })
  .extend('user', async ({ database }) => {
    return await database.createTestUser()
  })

test('user is created in database', async ({ user, database }) => {
  // TypeScript knows:
  // - user is the return type of createTestUser()
  // - database is Database
  const found = await database.findUser(user.id)
  expect(found).toEqual(user)
})
```

Fixes #9305
Fixes #9555

### Tests

### Documentation

### Changesets

### DIFF
--- a/test/cli/test/fails.test.ts
+++ b/test/cli/test/fails.test.ts
@@ -101,7 +101,7 @@ it('prints a warning if the assertion is not awaited', async () => {
         at <rootDir>/base.test.js:9:33",
       "Promise returned by \`expect(actual).resolves.toBe(expected)\` was not awaited. Vitest currently auto-awaits hanging assertions at the end of the test, but this will cause the test to fail in the next Vitest major. Please remember to await the assertion.
         at <rootDir>/base.test.js:14:33",
-      "Promise returned by \`expect(actual).toMatchFileSnapshot(expected)\` was not awaited. Vitest currently auto-awaits hanging assertions at the end of the test, but this will cause the test to fail in Vitest 3. Please remember to await the assertion.
+      "Promise returned by \`expect(actual).toMatchFileSnapshot(expected)\` was not awaited. Vitest currently auto-awaits hanging assertions at the end of the test, but this will cause the test to fail in the next Vitest major. Please remember to await the assertion.
         at <rootDir>/base.test.js:19:17",
     ]
   `)


sheremet-va:
* **#9553** <a href="https://app.graphite.com/github/pr/vitest-dev/vitest/9553?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9550** <a href="https://app.graphite.com/github/pr/vitest-dev/vitest/9550?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vitest-dev/vitest/9550?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

docs/guide/test-context.md
```diff
@@ -309,178 +672,146 @@ export default defineConfig({
...
+  expect(server).toBe('http://localhost:3000')
 })
+```
 
-describe('one type of schema', () => {
-  test.scoped({ schema: 'schema-1' })
-
-  // ... tests
-})
+#### Chaining Multiple Pverrides
```
hi-ogawa:
```suggestion
#### Chaining Multiple Overrides
```

test/cli/test/scoped-fixtures.test.ts
```diff
@@ -690,9 +690,1319 @@ describe('browser tests', () => {
...
+          // runtime overrides, but ts shows an error
+          // we don't enforce it because it's possible to provide objects
+          // which are hard to compare at runtime
+          expect(apiUrl).toBe(false)
+        })
+      },
+    })
+    expect(stderr).toBe('')
+    expect(tests).toMatchInlineSnapshot(`" ✓ basic.test.ts > onCleanup works <time>"`)
+  })
```
hi-ogawa, line 1389: There is an issue with duplicate extend. This looks similar to scoped/override bug https://github.com/vitest-dev/vitest/issues/9305.

```js
  // Current behavior: .extend() with duplicate prop is broken
  // Direct access works, but dependent fixtures still use old value
  describe('chained .extend() duplicate is broken (current behavior)', () => {
    const chainedTest = test
      .extend('a', () => 1)
      .extend('b', ({ a }) => a + 1)
      .extend('a', () => 100) // duplicate extend - broken!

    chainedTest('direct access returns new value', ({ a }) => {
      expect(a).toBe(100) // works
    })

    chainedTest('but dependent fixture still uses old value', ({ b }) => {
      // b uses old 'a' (1), so b = 1 + 1 = 2, not 101
      expect(b).toBe(2) // broken behavior
    })
  })
```

There's also dup extends patten in `test/core/test/fixture-initialization.test.ts`. 

https://github.com/vitest-dev/vitest/blob/20e00ef7808de6d330c5e2fda530f686e08f1c8d/test/core/test/fixture-initialization.test.ts#L12-L38

Fixing both seems complicated and I (and claude) gave up.

docs/guide/test-context.md
```diff
@@ -127,148 +127,511 @@ The [`onTestFinished`](/api/hooks#ontestfailed) hook bound to the current test.
...
+
+#### Setup and Cleanup with `onCleanup`
+
+For fixtures that need setup or cleanup logic, use a function. The `onCleanup` callback registers teardown logic that runs after the fixture's scope ends:
+
+```ts
+import { test as baseTest } from 'vitest'
+
+export const test = baseTest
+  .extend('tempFile', async ({}, onCleanup) => {
```
hi-ogawa: I prefer object api like `{ onCleanup }` just to prepare for any future addition.

test/cli/test/scoped-fixtures.test.ts
hi-ogawa: Also I wonder if we support this `originA` pattern for `test.override` now?

```js
   // override origin a 
   a: async ({ a: originA }, use) => { 
```

sheremet-va: > There is an issue with duplicate extend. This looks similar to scoped/override bug https://github.com/vitest-dev/vitest/issues/9305.

Good catch, I did fix it in `test.override` though. I think this should be fixed for the future `mergeTests` to work properly

> Also I wonder if we support this originA pattern for test.override now?

Did we not support it? It uses the same parser

sheremet-va: > Did we not support it? It uses the same parser

Ah, you don't mean the syntax

sheremet-va: > Also I wonder if we support this originA pattern for test.override now?

I feel like this is cursed and should not even be supported in `extend`, especially with types. Currently I am actually forbidding returning different types

sheremet-va: The `extend.extend` is now fixed though, let's handle overriding the original one separetly

test/core/test/fixture-initialization.test.ts
```diff
@@ -74,7 +73,7 @@ describe('fixture initialization', () => {
       expectTypeOf(b).toEqualTypeOf<string>()
 
       expect(fnA).toBeCalledTimes(1)
-      expect(fnB).toBeCalledTimes(1)
+      expect(fnB).toBeCalledTimes(0)
```
sheremet-va, line 77: This actually seems like a bug in older implementation - why would original `b` be called if we override it and never ask?

test/cli/test/scoped-fixtures.test.ts
```diff
@@ -169,12 +203,29 @@ test('worker fixture can import a static value from test fixture', async () => {
       })
     },
   }, { globals: true })
-  expect(stdout).toContain('basic.test.ts')
-  expect(stderr).toBe('')
+  expect(stderr).toMatchInlineSnapshot(`
```
sheremet-va, line 206: This is a breaking change, but it simplifies mental model A LOT. Scoped fixtures can only use other scoped fixtures with a higher scope, and scope is _explicit_ (but overriding the fixture also inherits all the options by default, like in PW). 

By default, any extended value has a `test` scope. You cannot override the extended value to provide a DIFFERENT `scope`, it now throws an error. This is needed to remove the ambiguity of defining a worker fixture and then overriding its "static" dependency inside the suite:

```ts
// this worked previously
const test = base
  .extend('port', 5300)
  .extend('config', { scope: 'worker' }, ({ port }) => {
    return `localhost:${port}`
  })
```

But if you want to override the port, you need to invalidate or overwrite the WORKER fixture that we already initialized - which makes little sense

```ts
test.override('port', 5000) // what to do here?
```

Using `test.override` with a scope other than `test` in general is now forbidden, unless it's called at the top level of the file (which still causes some unexpected issues with worker fixtures because they will leak to other test files in the same worker - I wonder if we could force a new worker instead).

sheremet-va:
@hi-ogawa I updated the fixtures inheritance (there is now a `TestFixters` class that tracks test's fixtures), please take a look. I am also introducing a small breaking change, not sure if we can put it into 4.1, but the impact should be small, I believe (I could probably implement it with backwards compatibility if we are ok with small bugs)

test/cli/test/around-each.test.ts
```diff
@@ -1156,8 +1156,8 @@ test('aroundEach with fixtures', async () => {
     >> user fixture setup
     >> test running, db available: true
     >> user fixture teardown
-    >> db fixture teardown
-    >> aroundEach teardown"
+    >> aroundEach teardown
+    >> db fixture teardown"
```
sheremet-va, line 1160: This behaviour seems to make more sense than what was here before because you should be able to access `db` after calling `runTest()`. Previously the db was torn down by that point 🤔

test/cli/test/scoped-fixtures.test.ts
sheremet-va: This also makes it so you cannot call `test.extend('', { scope: 'file' })` inside a `suite`, it's only allowed in the top level of the module now

sheremet-va:
Damn it, snapshots fail on rolldown because it processes files differently 😡

docs/guide/test-context.md
```diff
@@ -127,148 +127,537 @@ The [`onTestFinished`](/api/hooks#ontestfailed) hook bound to the current test.
...
+    // TypeScript knows apiKey is string
+    return new ApiClient(apiKey)
+  })
+```
+
+### Fixture Scopes <Version>3.2.0</Version> {#fixture-scopes}
+
+By default, fixtures are initialized for each test. You can change this with the `scope` option to share fixtures across tests.
+
+::: warning
```
sheremet-va, line 492: Added this warning section, @hi-ogawa. Would love to know what you think

packages/runner/src/map.ts
```diff
@@ -17,12 +17,12 @@ export function getFn<Task = Test>(key: Task): () => Awaitable<void> {
...
 export function setTestFixture(
   key: TestContext,
-  fixture: FixtureItem[] | undefined,
+  fixture: TestFixtures,
 ): void {
   testFixtureMap.set(key, fixture)
 }
 
-export function getTestFixture<Context = TestContext>(key: Context): FixtureItem[] {
+export function getTestFixturesManager<Context = TestContext>(key: Context): TestFixtures {
```
hi-ogawa: nit: any reason renaming only `getTestFixture -> getTestFixturesManager` but keeping `setTestFixture`?

packages/runner/src/fixture.ts
```diff
@@ -1,125 +1,235 @@
...
+  constructor(
+    registrations?: Map<string, TestFixtureItem>,
+  ) {
+    this._registrations = registrations ?? new Map()
+    this._suiteContexts = new WeakMap()
+    TestFixtures._definitions.push(this)
+  }
+
+  extend(runner: VitestRunner, userFixtures: Record<string, any>): TestFixtures {
+    const { suite } = getCurrentSuite()
```
hi-ogawa, line 58: nit: `TestFixtures` structure looks great, but `TestFixtures.extend/override` relying on `getCurrentSuite` feels odd in terms of state separation. How about pass it from argument?

test/cli/test/around-each.test.ts
hi-ogawa: Right, this looks like what I suggested to happen in https://github.com/vitest-dev/vitest/pull/9450#discussion_r2723909520 I forgot to double check the test.

test/cli/test/scoped-fixtures.test.ts
hi-ogawa: I wasn't checking these test cases previously and the new extend/override behavior makes sense to me.

hi-ogawa (commented):
`TestFixtures` class looks great.

docs/guide/test-context.md
hi-ogawa: The behavior looks reasonable to me. Note sure "force a new worker" behavior is better.

packages/runner/src/fixture.ts
sheremet-va: I did pass it before, but moved it into the function directly in the last commit. We already call this function inside `withFixtures` anyway. The suite collection is intentionally global and fixture values rely on what suite they are in, so they are inherently tied. If we had a separate Runner class that does the collection, then it would make sense

docs/guide/test-context.md
sheremet-va: I think "force a new worker" is what playwright does, otherwise worker fixtures ovrerride each other in different files

packages/runner/src/map.ts
sheremet-va: That was one of the old attempts, I reverted to `TestFixtures`. There are now `fixtures` (the instance of the class) that have `registrations`

PR #9572, opened 2026-02-02
feat(fixtures): pass down file context to `beforeAll`/`afterAll`

sheremet-va:
### Description

Closes #9553

This PR makes it so every suite hook (beforeAll, afterAll, aroundAll) receives file context as one of the arguments:

```ts
import { test as base } from 'vitest'

const test = base
	.extend('config', { scope: 'file' }, () => loadConfig())
	.extend('db', { scope: 'file' }, ({ config }) => createDatabase(config.port))

test.beforeAll(async ({ db }) => {
  await db.migrateUsers()
})

test.afterAll(async ({ db }) => {
  await db.deleteUsers()
})
```

The fixtures are only available if hook was called on `test`, like `test.beforeAll`. Using a global `beforeAll` won't pass down fixtures because they are scoped to the test (if there are multiple `test.extend`, it's not possible to know which context should be used, for example). Vitest will show a warning if user tries to access fixtures in the global hook. This will be an error in the next major.

After the rewrite this is much easier to implement

### DIFF
--- a/test/cli/test/scoped-fixtures.test.ts
+++ b/test/cli/test/scoped-fixtures.test.ts
@@ -44,8 +44,8 @@ test('test fixture cannot import from file fixture', async () => {
           2|   const extendedTest = it.extend({
            |                           ^
           3|     local: ({}, use) => use("local"),
-          4|     file: [
-     ❯ basic.test.ts:11:1
+          4|     file: [({ local }, use) => use(local), { scope: "file" }]
+     ❯ basic.test.ts:8:1
 
     ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯
 
@@ -679,8 +679,6 @@ test('global beforeAll/afterAll hooks throw error when accessing any fixture', a
   })
 
   expect(fixtures).toMatchInlineSnapshot(`">> fixture | beforeAll | file: undefined"`)
-  // TODO(debug): remove before merge
-  console.log('root', fs.root)
   expect(replaceRoot(stderr, fs.root)).toMatchInlineSnapshot(`
     "stderr | basic.test.ts
     The beforeAll hook uses fixtures "fileValue", but has no access to context. Did you forget to call it as "test.beforeAll()" instead of "beforeAll()"? This will throw an error in a future major. See https://vitest.dev/guide/test-context#suite-level-hooks
--- a/test/test-utils/index.ts
+++ b/test/test-utils/index.ts
@@ -498,21 +498,29 @@ export async function runInlineTests(
   }
 }
 
+const isWindows = process.platform === 'win32'
+
 export function replaceRoot(string: string, root: string) {
   const schemaRoot = root.startsWith('file://') ? root : pathToFileURL(root).toString()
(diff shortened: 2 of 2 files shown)


packages/runner/src/hooks.ts
```diff
@@ -67,17 +72,19 @@ export function getBeforeHookCleanupCallback(hook: Function, result: any, contex
...
-  fn: BeforeAllListener,
+export function beforeAll<ExtraContext = object>(
+  this: unknown,
+  fn: BeforeAllListener<ExtraContext>,
   timeout: number = getDefaultHookTimeout(),
 ): void {
   assertTypes(fn, '"beforeAll" callback', ['function'])
   const stackTraceError = new Error('STACK_TRACE_ERROR')
-  return getCurrentSuite().on(
+  const context = getChainableContext(this)
```
sheremet-va, line 83: Since we put hooks on `testFn`, they have access to the same `this`. If global `beforeAll` (or other suite hooks) are called, `this` is `undefined`, and fixtures always return `{}`

packages/runner/src/suite.ts
```diff
@@ -403,17 +403,16 @@ function createSuiteCollector(
     })
     setTestFixture(context, options.fixtures ?? new TestFixtures())
 
-    // custom can be called from any place, let's assume the limit is 15 stacks
```
sheremet-va, line 406: Removed because we don't support `type: custom` anymore

packages/runner/src/fixture.ts
```diff
@@ -269,19 +269,45 @@ export async function callFixtureCleanupFrom(context: object, fromIndex: number)
...
     if (!context) {
+      if (options?.suiteHook) {
+        validateSuiteHook(fn, options.suiteHook, options.stackTraceError)
+      }
+
       return fn({})
     }
 
-    const fixtures = getTestFixtures(context)
+    const fixtures = options?.fixtures || getTestFixtures(context)
```
sheremet-va, line 310: We could actually just always pass it down without any helpers now 🤔

sheremet-va: Welp, `each` hooks don't have access to it like tests do (only if on `test.beforeEach`, but that would be a breaking change)

docs/guide/test-context.md
```diff
@@ -845,3 +835,73 @@ test.afterEach(({ counter }) => {
...
+  await database.migrate()
+})
+
+test.afterAll(async ({ database }) => {
+  await database.cleanup()
+})
+
+test.aroundAll(async (runSuite, { database }) => {
+  await database.transaction(runSuite)
+})
```
hi-ogawa, line 863: Is this realistic example? `aroundAll` wraps `beforeAll/afterAll`, so transaction is wrapping migration (i.e. schema change) and if I remember correctly, that's not what db normally supports.

Since `aroundAll` runs first, this doc should order `aroundAll / beforeAll / afterAll`, so it's easier to follow (then caveat about hooks ordering can be documented elsewhere likely in hooks.md).

hi-ogawa (commented):
Looks good and a way simpler. I should've waited for this instead of trying to figure out the first one :smile:

docs/guide/test-context.md
sheremet-va: sure, updated to use `createUsers`/`deleteUsers`

PR #9579, opened 2026-02-03
feat(vitest): add run summary in GitHub Actions Reporter

### LINKED ISSUE #9498: add "flaky tests" summary to Github Actions Reporter

Sometimes tests are flaky and people use the `retry` option to avoid failing CI when that happens.
But you still want to know which tests are flaky to try and improve them.

### Suggested solution

Collect flaky tests (reruns > 0) in the reporter and print a warning summary at the end of the run.

The reason for suggesting the warning summary at the end is that this is obviously less important than failures and shouldn't be intermittent with other errors or warnings in the output.

Ideally link to the test by line and if possible also link back to the execution in the log (no idea if the latter is possible)

### Alternative

A custom reporter could be built to provide this

### Additional context

discord discussion https://discord.com/channels/917386801235247114/1462789015895019724

### Validations

macarie:
### Description

Resolves #9498

Screenshot showing Vitest's run summary in a GitHub Action page

### Tests

### Documentation

### Changesets


packages/vitest/src/node/reporters/github-actions.ts
```diff
@@ -165,3 +180,111 @@ function escapeProperty(s: string): string {
...
+            ratio: retriesRatio,
+          },
+          line: test.task.location?.line,
+          testName: test.task.fullTestName,
+        })
+      }
+    }
+
+    if (flakyTests.tests.length > 0) {
+      flakyTests.tests.sort((a, b) => b.retries.ratio - a.retries.ratio)
```
macarie, line 358: Should these be sorted or should we keep them in the order they're returned from the generator?

macarie (commented):
I think we can use this as a starting point and add more information in the future.

I still have to add a test, but other than that the logic should be done.

packages/vitest/src/node/reporters/github-actions.ts
sheremet-va: I like the sorted list

packages/vitest/src/node/reporters/github-actions.ts
```diff
@@ -165,3 +184,117 @@ function escapeProperty(s: string): string {
...
+        summary += `\n- ${mdLink(`\`${flakyTest.testName}\``, fileLinkCreator(flakyTests.path.absolute, flakyTest.line))} (${flakyTest.retries.ratio >= 0.8 ? `**${retriesText}**` : retriesText})`
+      }
+
+      summary += '\n'
+    }
+  }
+
+  if (summary === SUMMARY_HEADER) {
+    return null
+  }
```
macarie: Should it write an "empty summary" when there's nothing to report?

If there will be stats (X passed, Y skipped, etc.) at some point, I don't think this would matter anymore.

test/cli/test/reporters/github-actions.test.ts
```diff
@@ -59,4 +60,44 @@ describe(GithubActionsReporter, () => {
     `)
     expect(stderr).toBe('')
   })
+
+  // summary gets written in $GITHUB_STEP_SUMMARY
+  it.runIf(process.env.GITHUB_STEP_SUMMARY !== undefined)('creates summary', async () => {
```
macarie: I don't particularly like this test as it's not _stateless_. If we log stats at the end, this will fail as content gets appended to `$GITHUB_STEP_SUMMARY`...

Should I add an `outfile` option to the class to overwrite that path? I'm not fond of the idea as it would get used mostly for this test, rather than by users 😕

This test also depends on `GITHUB_REPOSITORY`, `GITHUB_SHA`, and `GITHUB_WORKSPACE`.

hi-ogawa (commented):
Do we need dedicated summary with `GITHUB_STEP_SUMMARY`?
I realized what playwright does is just writing summary as extra annotations at the end, which still shows up in the same place.

Example: https://github.com/vitejs/vite-plugin-react/actions/runs/22048735814

<details><summary>Screenshot</summary>

Image

</details>

macarie:
@hi-ogawa, I've tried but looks like those blocks don't render markdown, everything is just mono-spaced text. Not sure how user-friendly that would be. They also [seem to have significant limitations](https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md#limitations) which would make them unsuitable for large suites or when using multiple projects.

<details>
<summary>Screenshot</summary>

Screenshot showing annotations with text rendered as raw markdown

</details>

I do like having one for each run tho, as opposed to summaries which are just merged together:

<details>
<summary>Multiple summaries preview</summary>

Screenshot of multiple reports merged into one summary

</details>

hi-ogawa:
Good point. Wasn't aware that the limitation. Makes sense to me to use the summary in that case :+1:

test/cli/test/reporters/github-actions.test.ts
sheremet-va: > I'm not fond of the idea as it would get used mostly for this test, rather than by users 😕

In other reporters we do the same for `isTTY` checks to be fair

sheremet-va: You can also set the `process.env.GITHUB_STEP_SUMMARY`

hi-ogawa:
How about adding test files and test counts, something equivalent to this? https://github.com/vitest-dev/vitest/blob/959aea4052344659d039ff9b49443d2b51259897/packages/vitest/src/node/reporters/base.ts#L558-L559

hi-ogawa (dismissed):

packages/vitest/src/node/reporters/github-actions.ts
```diff
@@ -4,24 +4,95 @@ import type { Vitest } from '../core'
...
+       * This value is used to compute relative file paths for the permalink URLs.
+       *
+       * @default process.env.GITHUB_WORKSPACE
+       */
+      workspacePath?: string | undefined
+    }
+  }
+}
+
+type SummaryOptions = NonNullable<GithubActionsReporterOptions['jobSummary']>
```
sheremet-va: Shouldn't we do the opposite? Define `SummaryOptions` and use them in `GithubActionsReporterOptions['jubSummary']`?

They can even be required by default so we don't do so much nesting in `Required<Pick<SummaryOptions, 'enabled'>> & Omit<SummaryOptions, 'enabled'>`

macarie: Done, if I understood correctly what you meant 😄

macarie:
The job summary looks like this, after the change proposed by @hi-ogawa:

GitHub Job Summary

packages/vitest/src/node/reporters/github-actions.ts
```diff
@@ -4,24 +4,92 @@ import type { Vitest } from '../core'
...
+     */
+    commitHash?: string | undefined
+    /**
+     * The absolute path to the root of the repository on disk.
+     *
+     * This value is used to compute relative file paths for the permalink URLs.
+     *
+     * @default process.env.GITHUB_WORKSPACE
+     */
+    workspacePath?: string | undefined
```
macarie, line 67: I think I'm using this the same way `onWritePath` — introduced in #8015 — would, but I don't quite understand why the default just returns the provided path.

Does GitHub pick the correct file to associate an annotation to automatically by disregarding the "cwd" part?

I could drop this option and change the default `onWritePath` to resolve paths relative to `process.env.GITHUB_WORKSPACE` if it's available, otherwise return the input string.

cc (?) @AriPerkkio

packages/vitest/src/node/reporters/github-actions.ts
```diff
@@ -165,3 +246,192 @@ function escapeProperty(s: string): string {
...
+  if (stats.failed > 0) {
+    primaryInfo.push(`❌ **${stats.failed} ${noun(stats.failed, 'failure', 'failures')}**`)
+  }
+
+  if (stats.passed > 0) {
+    primaryInfo.push(`✅ **${stats.passed} ${noun(stats.passed, 'pass', 'passes')}**`)
+  }
+
+  if (stats.expectedFail > 0) {
+    primaryInfo.push(`🔵 **${stats.expectedFail} expected ${noun(stats.expectedFail, 'failure', 'failures')}**`)
```
macarie: Would it make sense for this to be 🟠 (or another color) instead or to keep the blue as it's aligned with the CLI reporters?

packages/vitest/src/node/reporters/github-actions.ts
AriPerkkio: > but I don't quite understand why the default just returns the provided path.

In #8015 a function was added that users can override. The reason why it just returns the provided path is due to this change in https://github.com/vitest-dev/vitest/pull/8015/changes:



_(image since linking in Github is nowadays horrible)_

test/cli/test/reporters/github-actions.test.ts
```diff
@@ -59,4 +62,148 @@ describe(GithubActionsReporter, () => {
...
+
+      const summary = await readFile(outputPath, 'utf8')
+
+      expect(summary).toMatchInlineSnapshot(`
+        "## Vitest Test Report
+
+        ### Summary
+
+        - **Test Results**: ❌ **1 failure** · ✅ **9 passes** · 🔵 **1 expected failure** · 11 total
+        - **Other**: 1 skip · 1 todo · 2 total
```
hi-ogawa, line 188: Can we include test files count too? Something like the first one in default terminal summary.

```
 Test Files  2 passed (2)
      Tests  6 passed (6)
```

packages/vitest/src/node/reporters/github-actions.ts
```diff
@@ -165,3 +246,224 @@ function escapeProperty(s: string): string {
...
+    }>
+  }>
+}
+
+function collectSummaryData(testModules: ReadonlyArray<TestModule>): SummaryData {
+  const summaryData: SummaryData = {
+    fileStats: {
+      failed: 0,
+      passed: 0,
+    },
```
macarie, line 277: Can files have other states?

hi-ogawa: I don't think there is, so this should be fine.
What base reporter does on terminal is reusing the same utility `getStateString` for both file and test stats, so it doesn't tell whether that happens.
https://github.com/vitest-dev/vitest/blob/97dae8a75e8ca72fc42db190f17d818c41515ef6/packages/vitest/src/node/reporters/base.ts#L558-L559

hi-ogawa (approved):
Looks good!

chvanam:
Can this markdown output be retrieved programmatically from the cli, so it can be further used? (For example, posted as a sticky comment on a PR)
Otherwise, I think this is something a potential action/vitest could do!

macarie:
@chvanam the content gets written to `process.env.GITHUB_STEP_SUMMARY` by default.

For example, with [peter-evans/create-or-update-comment](https://github.com/peter-evans/create-or-update-comment) you could use `body-path: ${{ env.GITHUB_STEP_SUMMARY }}` to post it.
Alternatively you could `cat ${{ env.GITHUB_STEP_SUMMARY }}` the contents and save them as a step's output to use later, but keep in mind it's multiline and has dynamic contents so you need to write it using EOF delimiters:
```
EOF=$(openssl rand -hex 16)
echo "summary<<$EOF" >> $GITHUB_OUTPUT
cat ${{ env.GITHUB_STEP_SUMMARY }} >> $GITHUB_OUTPUT
echo "$EOF" >> $GITHUB_OUTPUT
```

Just note that the file has to be read in the same job, otherwise I think it will reset.

chvanam:
@macarie good catch, I completely missed that!

PR #9586, opened 2026-02-04
fix: resolve "vite" relative to the project root instead of using a dependency, if possible

sheremet-va:
### Description

The goal is to avoid any breaking changes even with the most unhinged solutions

Turns out we cannot use TLA because it taints the whole module graph, meaning user dependencies in the config can't be loaded, for example (even `require(vite)`). I assumed that it doesn't affect siblings imports previously.

Testing simple use cases with vite 6, vite 7 and vite 8 seems to work (including Node 20 without `require(esm)` support).

The exports tests should be fixed with https://github.com/antfu/vitest-package-exports/pull/5 because `require(esm)` cannot be imported in `Promise.all()` 🤷


sheremet-va:
Superseded by a much simpler fix at #9587

PR #9591, opened 2026-02-05
fix: trim trailing white spaces in code block

hi-ogawa:
### Description

I have my vscode with auto trimming of trailing white spaces, which causes breaking snapshots with code block generated by `expect(stderr).toMatchInlineSnapshot()`.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/printError.ts
+++ b/packages/vitest/src/node/printError.ts
@@ -469,10 +469,8 @@ export function generateCodeFrame(
           return ''
         }
 
-        res.push(
-          (lineNo(j + 1)
-            + truncateString(lines[j].replace(/\t/g, ' '), columns - 5 - indent)).trimEnd(),
-        )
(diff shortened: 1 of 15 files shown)


AriPerkkio (approved):
Yes please, this is great! I've done similar in https://github.com/AriPerkkio/vitest/blob/0a6588439e72c9f06d6bb7a71fa95efa54c01acc/test/cli/test/detect-async-leaks.test.ts#L310-L315 test case to avoid prettier from trimming the lines.

For end users this will show up as changed snapshots, but I would consider this as bug fix rather than breaking change. 👍

sheremet-va (approved):

sheremet-va:
> For end users this will show up as changed snapshots, but I would consider this as bug fix rather than breaking change. 👍

It would only happen if they test Vitest error output, which we decided is not subject to semver.

PR #9594, opened 2026-02-05
feat: add assertion helper to hide internal stack traces

### LINKED ISSUE #4494: Test line that fails is well hidden if a helper method fails
### Describe the bug

Given a test like
```
    it('should fail at the second expect', () => {
      expect(1).to.equal(1);
      expect(2).to.equal(1);
    });
```

you get the expect output, e.g.
```
 FAIL  my/test/selection.test.ts > selection > single > should fail at the second expect
AssertionError: expected 2 to equal 1

- Expected
+ Received

- 1
+ 2

 ❯ my/test/selection.test.ts:123:19
    121|     it('should fail at the second expect', () => {
    122|       expect(1).to.equal(1);
    123|       expect(2).to.equal(1);
       |                   ^
    124|     });
    125|   });

```

But when `expect` is extracted to a helper like 
```
  function e(a: number, b: number) {
    expect(a).to.equal(b);
  }
```

the output of the test
```
    it('should fail at the second expect', () => {
      e(1, 1);
      e(2, 1);
    });
```

is
```
 FAIL  my/test/selection.test.ts > selection > single > should fail at the second expect
AssertionError: expected 2 to equal 1

- Expected
+ Received

- 1
+ 2

 ❯ e my/test/selection.test.ts:80:17
     78| 
     79|   function e(a: number, b: number) {
     80|     expect(a).to.equal(b);
       |                 ^
     81|   }
     82| 
 ❯ my/test/selection.test.ts:123:6
```

and actually I did not realize before pasting the error here that the line number is actually present for the test case (`my/test/selection.test.ts:123:6`). I thought it was completely missing. Could maybe the line of the actual test always be printed?

### Reproduction

As above

### System Info

```shell
System:
    OS: macOS 14.0
    CPU: (10) arm64 Apple M1 Max
    Memory: 870.36 MB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.9.0 - ~/.nvm/versions/node/v20.9.0/bin/node
    npm: 10.1.0 - ~/.nvm/versions/node/v20.9.0/bin/npm
    bun: 1.0.7 - ~/.bun/bin/bun
  Browsers:
    Brave Browser: 107.1.45.123
    Chrome: 119.0.6045.123
    Chrome Canary: 121.0.6125.0
    Edge: 119.0.2151.58
    Safari: 17.0
  npmPackages:
    @vitejs/plugin-react: 4.1.1 => 4.1.1 
    @vitest/browser: ^0.34.6 => 0.34.6 
    @vitest/ui: ^0.34.6 => 0.34.6 
    vite: 4.5.0 => 4.5.0 
    vitest: ^0.34.6 => 0.34.6
```

### Used Package Manager

npm

### Validations

hi-ogawa:
### Description

Resolves https://github.com/vitest-dev/vitest/issues/4494

This PR implements `vi.defineHelper`, which works like this:

```ts
import { expect, vi } from 'vitest'

const assertPair = vi.defineHelper((a, b) => {
  expect(a).toEqual(b)  // 🙅‍♂️ error code block will NOT point to here
})

test('example', () => {
  assertPair('left', 'right')  // 🙆 but point to here
})
```

Example output:

```js
FAIL  example.test.ts > example
AssertionError: expected 'left' to deeply equal 'right'

Expected: "right"
Received: "left"

 ❯ example.test.ts:8:3
      7| test('example', () => {
      8|   assertPair('left', 'right')
       |   ^
      9| })
```

The implementation relies on error stack error filtering based on special marker `function __VITEST_HELPER__`. This processing happens during `parseStacktrace`, so the filtering also applies to `--printConsoleTrace` (which I just realized and maybe we can tweak the behavior.)

The error stack and `__VITEST_HELPER__` filtering requires the original error to have proper trace. It turns out Webkit is a bit behind other runtimes, so currently "async" helper doesn't work. But, it looks like it's coming soon as Bun has it already https://github.com/WebKit/WebKit/pull/57832 and this is also reached main line webkit https://github.com/oven-sh/bun/pull/22517. We'll likely just need to wait for a few weeks until Playwright webkit update and see what happens.

TODO
- [x] test
  - [x] console
- [x] doc

TBD
- [x] should `printConsoleTrace` also hide helper internal?
- [x] rename to `vi.defineHelper`

### Tests

### Documentation

### Changesets


docs/api/vi.md
```diff
@@ -1335,3 +1335,42 @@ function resetConfig(): void
 ```
 
 If [`vi.setConfig`](#vi-setconfig) was called before, this will reset config to the original state.
+
+### vi.helper <Version>4.1.0</Version> {#vi-helper}
```
sheremet-va: Let's name it `defineHelper`

sheremet-va (changes requested):

hi-ogawa:
The API is renamed to `vi.defineHelper`.

sheremet-va (approved):

sheremet-va:
LGMT!

PR #9647, opened 2026-02-12
feat(reporters): add OSC 8 hyperlinks for stacktrace file paths

### LINKED ISSUE #9565: Add support for OSC 8 hyperlink in output

When running vitest from Ghostty, file path in the output are not CMD+clickable:

Image

This is due to the fact that terminals implement some logic to try and match file paths using a regexp, which is an imperfect process.

Terminal emulators and terminal CLI commands have started converging on the [OSC 8](https://github.com/Alhadis/OSC8-Adoption) "standard" to tag hyperlinks in terminal output, but vitest does not use that standard.

### Suggested solution

Wrap file paths in the terminal output with the required tags to mark them as hyperlinks.

### Alternative

The https://github.com/sentriz/add-osc-8-hyperlink tool can be used to transform a process's output and add clickable links, but it does not seem to work with the output of `vitest run`.

### Additional context

_No response_

### Validations

Ujjwaljain16:
### Description

This adds optional OSC 8 hyperlink support to stacktrace file paths in terminal output. This allows compatible terminals (Ghostty, iTerm, VS Code, etc.) to make file paths clickable even if they wrap or contain special characters.

- Links are enabled only when `stdout` is TTY
- Disabled in CI or when `VITEST_FORCE_TTY=false`
- Uses `pathToFileURL` for cross-platform support (handles Windows paths correctly)
- Reporter-only change (no runtime logic affected)

Resolves #9565

### Tests

### Documentation

### Changesets


packages/vitest/src/node/reporters/terminalLink.ts
```diff
@@ -0,0 +1,23 @@
+import { pathToFileURL } from 'node:url'
+
+export function createTerminalLink(text: string, file: string): string {
+  // Graceful fallback for non-TTY or CI
+  // Also check VITEST_FORCE_TTY to allow explicit disabling
+  if (!process.stdout?.isTTY || process.env.CI || process.env.VITEST_FORCE_TTY === 'false') {
```
sheremet-va: That checks seems incomplete

1. logger can accept a different `stdout` implementation, so using `process.stdout` is not sufficient
2. `CI` is not a sufficient check for a CI environment, we have `isCI` already that you can use
3. I am conflicted on `VITEST_FORCE_TTY: 'false'`, what do you think, @AriPerkkio? Feels like the presence of a ENV should be a check (like `NO_COLOR=anything`)

test/core/test/terminal-link.test.ts
```diff
@@ -0,0 +1,68 @@
+import { afterEach, describe, expect, test, vi } from 'vitest'
+import { createTerminalLink } from '../../../packages/vitest/src/node/reporters/terminalLink'
+
+describe('createTerminalLink', () => {
+  const originalEnv = process.env
+  let originalIsTTY: boolean | undefined
+
+  afterEach(() => {
```
sheremet-va: Instead of all of this mocking, it might be better to turn `createTerminalLink` into a pure function that accepts these as options

sheremet-va (changes requested):

packages/vitest/src/node/reporters/terminalLink.ts
AriPerkkio: Please no new environmental variables for TTY checking.

I think this function should be no-op only when these two indicate so:

https://github.com/vitest-dev/vitest/blob/0583f067eb6e6e5544b0fb603b0469bf50d69ab3/packages/vitest/src/node/logger.ts#L52

`isCI` from:

https://github.com/vitest-dev/vitest/blob/0583f067eb6e6e5544b0fb603b0469bf50d69ab3/packages/vitest/src/utils/env.ts#L15

hi-ogawa: We also have `import c from 'tinyrainbow'; c.isColorSupported`. Didn't it also had nice baseline check?

hi-ogawa:
This would be also more beneficial in import durations reporting because we truncate the file name which does actually break IDE (e.g. vscode's integrated terminal) to auto link the file https://github.com/vitest-dev/vitest/blob/a628859350b51a0f3c16eb942960eb282b666685/packages/vitest/src/node/reporters/base.ts#L714-L715

packages/vitest/src/node/reporters/terminalLink.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  // Ensure file exists and is valid before creating URL
+  if (!file || file.startsWith('node:')) {
+    return text
+  }
+
+  try {
+    const href = pathToFileURL(file).href
+    // OSC 8 escape sequence
+    return `\u001B]8;;${href}\u001B\\${text}\u001B]8;;\u001B\\`
```
AriPerkkio, line 22: Looking at https://github.com/sindresorhus/ansi-escapes/blob/6db12c839dafc906cbceda299026d8e78e501478/base.js#L19-L23, this might need some `tmux` specific handling

AriPerkkio: There's now `clickable-path` that we could probaly use to get all these edge cases handled automatically instead of maintaining the logic ourselves 🤔 (👋  @danielroe) 
- https://github.com/danielroe/clickable-path

PR #9651, opened 2026-02-13
fix: handle module-sync condition in vmThreads/vmForks require (fix #9650)

### LINKED ISSUE #9650: vmThreads pool: custom `require()` matches `module-sync` condition but cannot load ESM
### Describe the bug

When using the `vmThreads` pool, Vitest's host-thread `require()` matches the `module-sync` exports condition but then fails to load the resulting `.mjs` file, throwing `Cannot use import statement outside a module`.

This is related to but distinct from #7692 — that issue covers Vite's ESM resolver not picking up `module-sync`. This issue is about vmThreads' custom `require()` implementation, which has its own condition matching that bypasses Vite's `resolve` config entirely.

### Reproduction

Any package using `module-sync` in its exports map will trigger this. The simplest case is [`async-function`](https://github.com/ljharb/async-function) (v1.0.0), which has:

```json
{
  "exports": {
    ".": [
      {
        "module-sync": "./require.mjs",
        "import": "./index.mjs",
        "default": "./index.js"
      },
      "./index.js"
    ]
  }
}
```

`require.mjs` is a valid ESM file designed for Node 22.12+'s native `require()` (which can load ESM synchronously). vmThreads' custom `require()` resolves to this file via the `module-sync` condition, then tries to parse it as CJS and fails.

**vitest.config.ts:**

```ts

export default defineConfig({
  test: {
    pool: "vmThreads",
  },
});
```

**Error:**

```
SyntaxError: Cannot use import statement outside a module
 ❯ node_modules/async-function/require.mjs:1
```

The `forks` pool does not have this problem — it uses Node's native `require()` which handles `module-sync` correctly.

## What's been tried

- **`resolve.conditions: ["module-sync"]`** — No effect. This only controls Vite's ESM resolver, not vmThreads' host `require()`.
- **`poolOptions.vmThreads.execArgv: ["--experimental-require-module"]`** — No effect. The flag applies to Node's native module loader, but vmThreads builds its own `require()` using V8's `vm` API, bypassing it entirely.
- **Yarn patches converting `require.mjs` to CJS** — Works, but is a workaround.

## Expected behavior

vmThreads' custom `require()` should either:

1. Not advertise the `module-sync` condition if it cannot load ESM, or
2. Handle ESM when `module-sync` resolves to an `.mjs` file

## Affected packages

This affects any package using `module-sync` exports. In practice, the most common are ljharb's ecosystem packages:

- `async-function` (via `is-async-function` → `which-builtin-type` → `reflect.getprototypeof`)
- `generator-function` (via `is-generator-function` → `which-builtin-type` → `reflect.getprototypeof`)
- `async-generator-function` (via `is-async-generator-function` → `which-builtin-type` → `reflect.getprototypeof`)

These are transitive dependencies — they're pulled in by common packages like `es-abstract` and `deep-equal`.

See also: https://github.com/ljharb/async-function/issues/1

### System Info

```shell
System:
  OS: macOS
  Node: v22.x

Package:
  vitest: 4.0.18

Config:
  pool: vmThreads
```

### Used Package Manager

yarn

### Validations

lesleh:
## Summary

Fixes #9650

The `module-sync` exports condition can resolve to ESM files (`.mjs` or `.js` with `"type": "module"`) that the CJS `vm.Script` executor in vmThreads/vmForks cannot handle, causing `SyntaxError: Unexpected token 'export'`.

### Fix

Pass explicit CJS conditions (`node`, `require`, `node-addons`) to `require.resolve` in `CommonjsExecutor.createRequire`, excluding `module-sync`. This works on Node 22.12+ where `_resolveFilename` supports a `conditions` option. On older Node versions the option is silently ignored.

User-specified `--conditions`/`-C` flags (from both `process.execArgv` and `NODE_OPTIONS`) are respected and included in the conditions set, except for `module-sync` which is always filtered out. There is currently no public Node.js API to retrieve the resolved conditions (https://github.com/nodejs/node/issues/55824), so they are parsed from the CLI args directly.

The conditions must be a `Set` (not an array) because Node's internal resolver calls `conditions.has()`.

The hardcoded default conditions match Node's own [`initializeCjsConditions`](https://github.com/nodejs/node/blob/a57893b7993daa111849b161e62a926848e40da3/lib/internal/modules/helpers.js#L78), minus `module-sync`.

### Why the fix only works on Node 22.12+

The `module-sync` condition was added in Node 22.12 and backported to 20.19 when `require(esm)` was unflagged. The bug exists on both versions, but the fix only works on 22.12+:

- **The `conditions` option was added in Node 22.12** -- `_resolveFilename` only supports the `conditions` option starting from 22.12. On Node 20, the option is silently ignored and there is no clean way to exclude `module-sync` from userland.
- **`--no-experimental-require-module` is not viable** -- This removes `module-sync` but also disables `require(esm)` entirely, breaking other packages.
- **Node 20 reaches EOL April 2026.**

## Test plan

- Added integration test with two mock packages using `module-sync` exports condition:
  - `module-sync-mjs`: points to `.mjs` file
  - `module-sync-js`: points to `.js` file with `"type": "module"`
  - Both verify that `require()` resolves to the CJS fallback entry
  - Test is skipped on Node < 22 (fix requires the `conditions` option added in 22.12)
- Added unit tests for `parseCjsConditions` covering all flag forms (`--conditions=`, `--conditions `, `-C=`, `-C `), NODE_OPTIONS parsing, module-sync filtering, and edge cases

### DIFF
--- a/packages/vitest/src/runtime/vm/commonjs-executor.ts
+++ b/packages/vitest/src/runtime/vm/commonjs-executor.ts
@@ -200,10 +200,27 @@ export class CommonjsExecutor {
     m.exports = JSON.parse(code)
   }
 
+  // The "module-sync" exports condition (added in Node 22.12/20.19 when
+  // require(esm) was unflagged) can resolve to ESM files that our CJS
+  // vm.Script executor cannot handle. We exclude it by passing explicit
+  // CJS conditions to require.resolve (Node 22.12+).
+  // Must be a Set because Node's internal resolver calls conditions.has().
+  // If Node introduces new built-in CJS conditions in the future, this
+  // set may need to be extended.
+  private static cjsConditions = new Set(['node', 'require', 'node-addons'])
+
   public createRequire = (filename: string | URL): NodeJS.Require => {
     const _require = createRequire(filename)
+    const resolve = (id: string, options?: { paths?: string[] }) => {
+      return _require.resolve(id, {
+        ...options,
+        // Works on Node 22.12+ where _resolveFilename supports conditions.
+        // Silently ignored on older Node versions.
+        conditions: CommonjsExecutor.cjsConditions,
+      } as any)
+    }
     const require = ((id: string) => {
-      const resolved = _require.resolve(id)
+      const resolved = resolve(id)
       const ext = extname(resolved)
       if (ext === '.node' || isBuiltin(resolved)) {
         return this.requireCoreModule(resolved)
@@ -211,7 +228,8 @@ export class CommonjsExecutor {
       const module = new this.Module(resolved)
       return this.loadCommonJSModule(module, resolved)
     }) as NodeJS.Require
-    require.resolve = _require.resolve
+    require.resolve = resolve as NodeJS.RequireResolve
+    require.resolve.paths = _require.resolve.paths
     Object.defineProperty(require, 'extensions', {
       get: () => this.extensions,
       set: () => {},
--- a/packages/vitest/src/runtime/vm/commonjs-executor.ts
+++ b/packages/vitest/src/runtime/vm/commonjs-executor.ts
@@ -200,35 +200,14 @@ export class CommonjsExecutor {
     m.exports = JSON.parse(code)
   }
 
-  // The "module-sync" exports condition (added in Node 22.12/20.19 when
-  // require(esm) was unflagged) can resolve to ESM files that our CJS
-  // vm.Script executor cannot handle. We exclude it by passing explicit
-  // CJS conditions to require.resolve (Node 22.12+).
-  // Must be a Set because Node's internal resolver calls conditions.has().
-  // User-specified --conditions/-C flags are respected, except module-sync.
   private static cjsConditions: Set<string> | undefined
   private static getCjsConditions(): Set<string> {
-    if (CommonjsExecutor.cjsConditions) {
-      return CommonjsExecutor.cjsConditions
+    if (!CommonjsExecutor.cjsConditions) {
+      CommonjsExecutor.cjsConditions = parseCjsConditions(
+        process.execArgv,
+        process.env.NODE_OPTIONS,
+      )
     }
-    const conditions = ['node', 'require', 'node-addons']
-    const args = [
-      ...process.execArgv,
-      ...(process.env.NODE_OPTIONS?.split(/\s+/) ?? []),
-    ]
-    for (let i = 0; i < args.length; i++) {
-      const arg = args[i]
-      const eqMatch = arg.match(/^(?:--conditions|-C)=(.+)$/)
-      if (eqMatch) {
-        conditions.push(eqMatch[1])
-      }
-      else if ((arg === '--conditions' || arg === '-C') && i + 1 < args.length) {
-        conditions.push(args[++i])
-      }
-    }
-    CommonjsExecutor.cjsConditions = new Set(
-      conditions.filter(c => c !== 'module-sync'),
-    )
     return CommonjsExecutor.cjsConditions
   }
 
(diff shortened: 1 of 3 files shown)


packages/vitest/src/runtime/vm/commonjs-executor.ts
```diff
@@ -200,6 +201,13 @@ export class CommonjsExecutor {
     m.exports = JSON.parse(code)
   }
 
+  // On Node 22.12+, require.resolve() may resolve to .mjs via "module-sync"
+  // exports condition. Native require() can load these; our CJS vm.Script cannot.
```
sheremet-va: > our CJS vm.Script cannot

Yeah, and it MUST be loaded with _OUR_ `vm.Script`. Native `_require` looses the `vm.Context` of vmThreads wrapper.

This "fix" will add a lot more issues

sheremet-va (changes requested):
This is not fixable with current vm limitations.

packages/vitest/src/runtime/vm/commonjs-executor.ts
lesleh: Apologies, I was a little too eager to get in and write a fix. I think I have a better understanding of the issue now. If we can't use the native loader because it loses the `vm.Context`, and we can't use `vm.Script` because it doesn't support esm, then I think the next best thing to do is to prevent `require.resolve` from attempting to load export map entries labelled as `module-sync`. That way, it should only load CommonJS formats. This won't work for modules that are ESM only, of course, but it'd work fine for dual-export packages.

What do you think of this approach instead? I can investigate how feasible it is.

lesleh: @sheremet-va ready for another look when you have a chance, I took a different approach as discussed.

packages/vitest/src/runtime/vm/commonjs-executor.ts
```diff
@@ -200,18 +200,36 @@ export class CommonjsExecutor {
...
   }
 
+  // The "module-sync" exports condition (added in Node 22.12/20.19 when
+  // require(esm) was unflagged) can resolve to ESM files that our CJS
+  // vm.Script executor cannot handle. We exclude it by passing explicit
+  // CJS conditions to require.resolve (Node 22.12+).
+  // Must be a Set because Node's internal resolver calls conditions.has().
+  // If Node introduces new built-in CJS conditions in the future, this
+  // set may need to be extended.
+  private static cjsConditions = new Set(['node', 'require', 'node-addons'])
```
sheremet-va: I am ok with this solution, but it seems like we don't respect `--conditions` here (can be passed down in `execArgv` or inside `NODE_OPTIONS`). Should they be respected?

sheremet-va (changes requested):

packages/vitest/src/runtime/vm/commonjs-executor.ts
lesleh: Yes, I think we should respect `--conditions` here. I've updated the code to parse them from both `process.execArgv` and `process.env.NODE_OPTIONS`, filtering out `module-sync` so it can't be re-added.

Unfortunately there's no public Node API to get the resolved conditions, it's only available via the internal `getOptionValue('--conditions')`. There's a [closed issue for this](https://github.com/nodejs/node/issues/55824), and there is a [related open PR](https://github.com/nodejs/node/pull/57688) that would add `module.customConditions`. The original issue was filed because of Vite's need for exactly this.

sheremet-va (dismissed):
Looks good to me, but lint is failing, can you fix it?

lesleh:
@sheremet-va fixed the linting issues, and I've rebased from main. The browser CI failures on Node 24 (macOS/Windows) are pre-existing on main. Should be good to go 🙂

sheremet-va:
Thanks!

PR #9653, opened 2026-02-13
fix: hooks should respect `maxConcurrency`

### LINKED ISSUE #8367: `maxConcurrency` is not respected in `beforeAll`/`afterAll` hooks
### Describe the bug

More than `maxConcurrency` number of hooks can run in parallel. 

### Reproduction

package.json
```json title="package.json"
{
  "name": "vitest",
  "version": "1.0.0",
  "description": "",
  "license": "ISC",
  "author": "",
  "type": "commonjs",
  "devDependencies": {
    "vitest": "^3.2.4"
  }
}
```

a.test.js
```js
import { beforeAll, describe, it } from 'vitest';

async function sleep(ms=1000) {
    return new Promise(resolve => setTimeout(resolve, ms));
}

describe.concurrent('Test Suite', () => {
    describe.for(["a", "b"])('$0', (name) => {
        beforeAll(async () => {
            console.log(`Running beforeAll for ${name}`);
            await sleep();
        });

        it(`test ${name} 1`, async () => {
        });
    });
});
```

```console
$ npm exec -- vitest run --max-concurrency=1 a.test.js

 RUN  v3.2.4 <snip>

stdout | a.test.js > Test Suite > 'b'
Running beforeAll for a
Running beforeAll for b

 ✓ a.test.js (2 tests) 1008ms
   ✓ Test Suite > 'a' > test a 1 2ms
   ✓ Test Suite > 'b' > test b 1 0ms

 Test Files  1 passed (1)
      Tests  2 passed (2)
   Start at  09:50:21
   Duration  1.26s (transform 17ms, setup 0ms, collect 12ms, tests 1.01s, environment 0ms, prepare 59ms)
```

I would expect the test to run for at least 2 seconds, because the calls to sleep should not be run in parallel with `--max-concurrency=1`.

### System Info

```shell
Not relevant
```

### Used Package Manager

npm

### Validations

hi-ogawa:
### Description

- Closes https://github.com/vitest-dev/vitest/issues/8367

(This includes the same change from https://github.com/vitest-dev/vitest/pull/9673 for `aroundEach/All` error aggregation behavior.)

#### Problems 

Currently we apply promise limit only at `runTest` level, which executes `beforeEach -> test fn -> afterEach > onTestFinished/onTestFailed` routine for single test case. This doesn't limit other parallelism we have, which are:
- suite hooks (`beforeAll`, `afterAll`) between `describe.concurrent` neighbors
- running hooks (`beforeAll`, `afterAll`, `beforeEach`, `afterEach`) within a single group via `sequence.hooks: "parallel"`

#### Solution

This PR moves promise limit from `runTest` level to each "leaf level" function calls, which are user functions registered via `test/beforeEach/afterEach/beforeAll/afterAll/onTestFinished/onTestFailed`. This ensures any user code managed by test runner is bounded by single `maxConcurrency` limit. Since we apply this only at "leaf level", it's free from dead locking itself by nested promise limit.

~I'm not sure how `aroundEach/aroundAll` can work with this model, so they aren't currently bounded by `maxConcurrency`. (AI might figure out somehow making this work but I haven't tried yet).~ (EDIT: apparently it's possible)

TODO
- [x] summary
- [x] docs
- [x] test

### Tests

### Documentation

### Changesets


packages/runner/src/utils/limit-concurrency.ts
hi-ogawa, line 1: As I've put a lot more `limitConcurrency` in runner, I optimized this to reduce microtask as much as possible. This should cost only one microtask in the happy path (i.e. when not reaching limit) and two when limited. The previous implementation always paid three microtasks.

test/test-utils/index.ts
```diff
@@ -423,7 +424,7 @@ export default config
   return `export default ${JSON.stringify(content)}`
 }
 
-export function useFS<T extends TestFsStructure>(root: string, structure: T, ensureConfig = true) {
+export function useFS<T extends TestFsStructure>(root: string, structure: T, ensureConfig = true, context?: TestContext) {
```
sheremet-va: Maybe it's better to pass `task`? It has `task.context` that you can use and it still allows destructuring

sheremet-va (changes requested):

test/test-utils/index.ts
```diff
@@ -554,16 +556,33 @@ export class StableTestFileOrderSorter {
...
+        return result.errors.map(e => e.message)
+      }
+      return result.state
+    },
+    (testSuite, suiteChildren) => {
+      const errors = testSuite.errors().map(error => error.message)
+      if (errors.length > 0) {
+        return {
+          ...suiteChildren,
+          __suite_errors__: errors,
```
hi-ogawa, line 573: Tweaked test util so that `errorTree` includes suite errors.

test/cli/test/expect-task.test.ts
hi-ogawa, line 1: The concurrent tests was broken since it's using global `onTaskFinished`. I adjusted `runInlineTests` helper so it can use `context.onTaskFinished`.

test/cli/fixtures/fails/concurrent-suite-deadlock.test.ts
hi-ogawa, line 1: I moved these "fails" tests to more through integration test in `concurrent.test.ts`

sheremet-va (approved):

PR #9658, opened 2026-02-13
fix: handle Symbol values in format utility

### LINKED ISSUE #9377: [Bug] Browser mode: Cannot convert a Symbol value to a string in console.error
## Describe the bug

In vitest 4.0.16 browser mode (`@vitest/browser` + Playwright runner), when a raw `Symbol` value is passed directly as a `console.log` / `console.error` argument, the `baseFormat` function in `@vitest/utils` throws a TypeError.

The root cause is that `baseFormat` uses `str += \`${x}\`` for non-object trailing arguments. JavaScript does not allow implicit string conversion of Symbols via template literals, so this throws `TypeError: Cannot convert a Symbol value to a string`.

```js
// DOES crash in vitest 4.0.16
console.log('value:', Symbol('test'))
// → TypeError: Cannot convert a Symbol value to a string

// Does NOT crash (Symbol as object key is handled differently)
console.error('Test error:', { status: 400, [Symbol('rtk-query')]: { type: 'error' } })
```

This is a regression from 4.0.15.

## Reproduction

https://github.com/Suto-Michimasa/vitest-symbol-repro

### Steps to reproduce

```bash
git clone https://github.com/Suto-Michimasa/vitest-symbol-repro
cd vitest-symbol-repro
pnpm install
pnpm test:browser
```

### Stack trace

```
TypeError: Cannot convert a Symbol value to a string
    at baseFormat (@vitest/utils@4.0.16/dist/display.js:700:15)
    at browserFormat (@vitest/utils@4.0.16/dist/display.js:711:9)
    at processLog (__vitest_browser__/tester.js:1202:10)
    at console.error (__vitest_browser__/tester.js:1198:23)
```

### Root cause

In `@vitest/utils` `baseFormat` function:

```js
// display.js:700 (simplified)
str += `${x}`  // Throws TypeError when x is a Symbol
```

The fix should use `String(x)` instead of template literal interpolation, as `String(Symbol('test'))` safely returns `"Symbol(test)"`.

## Regression confirmation

Downgrading to 4.0.15 resolves the issue:

```bash
pnpm up vitest@4.0.15 @vitest/browser@4.0.15 @vitest/browser-playwright@4.0.15
pnpm test:browser  # passes
```

## Expected behavior

- Raw Symbol values should be safely handled during console output formatting (e.g., converted via `String(symbol)`)
- Tests should pass without TypeError

## Actual behavior

- Tests fail with `TypeError: Cannot convert a Symbol value to a string`
- The browser runner crashes when formatting raw Symbol values in console arguments

## System Info

- vitest: 4.0.16
- @vitest/browser: 4.0.16
- @vitest/browser-playwright: 4.0.16
- playwright: 1.57.0
- Browser: Chromium (via Playwright)
- Node.js: 20+
- OS: macOS

## Used Package Manager

pnpm

## Validations

- [x] Check that there isn't already an issue that reports the same bug
- [x] The provided reproduction is a minimal reproducible example

nami8824:
### Description

  Handle Symbol values correctly in the format utility (packages/utils/src/display.ts) to match util.format behavior:
  - %d formatter now returns NaN for Symbol input
  - Trailing Symbol arguments are output as Symbol(...) instead of throwing
  - Add Symbol test cases to each formatter group (`%s`, `%d`, `%i`, `%f`, `%o`, `%O`, `%j`, and trailing arguments

Resolves https://github.com/vitest-dev/vitest/issues/9377

### Tests

### Documentation

### Changesets


packages/utils/src/display.ts
```diff
@@ -177,6 +177,9 @@ function baseFormat(args: unknown[], options: FormatOptions = {}): string {
         if (typeof value === 'bigint') {
           return `${value.toString()}n`
         }
+        if (typeof value === 'symbol') {
+          return 'NaN'
+        }
```
nami8824, line 182: The `%d` case is also fixed. It's not in the original issue, but has the same root cause — `Number(Symbol())` throws a TypeError.

sheremet-va (approved):

sheremet-va:
LGTM! Thanks!

PR #9662, opened 2026-02-14
feat(runner): add mergeTests utility to compose TestAPI fixtures

### LINKED ISSUE #9483: Implement a way to merge test contexts

Implement an API for merging multiple test contexts, similar to Playwright’s `mergeTests` feature: [https://playwright.dev/docs/test-fixtures#combine-custom-fixtures-from-multiple-modules](https://playwright.dev/docs/test-fixtures#combine-custom-fixtures-from-multiple-modules).

This would enable more flexible and composable strategies for reusing test contexts across test suites.

### Suggested solution

Introduce a `mergeTests` (name TBD) API that composes multiple test contexts into a single one by combining their fixtures.

The merged context should:

* Preserve fixture scoping and execution order
* Allow partial overrides where explicitly defined
* Be composable, so merged contexts can be merged again

Example:

```typescript
import { mergeTests } from 'vitest';
import { test as dbTest } from './database-test-context';
import { test as a11yTest } from './a11y-test-context';

export const test = mergeTests(dbTest, a11yTest);
```

### Alternative

_No response_

### Additional context

_No response_

### Validations

Ujjwaljain16:
## Description

This adds a new `mergeTests(testA, testB)` utility that allows composing fixtures from multiple `TestAPI` instances.

Example:

```ts
const testA = test.extend({ a: 1 })
const testB = test.extend({ b: 2 })

const merged = mergeTests(testA, testB)

merged('works', ({ a, b }) => {
  // both fixtures available
})
```

### Implementation

The implementation reuses the existing `.extend()` mechanism to preserve fixture graph integrity.

Instead of manually merging internal `TestFixtures` maps, the approach is:

* Extract fixture definitions from `testB` via `resolveFixtures()`
* Call `testA.extend(fixturesFromB)`
* Return a new `TestAPI<A & B>`

This ensures:

* Proper dependency resolution
* Correct scope handling (test / worker)
* Auto fixtures behave naturally
* No mutation of internal state

### Override Semantics

If both tests define the same fixture name, the second argument (`testB`) overrides the first.
This mirrors `.extend()` behavior (last-writer-wins).

### Type Support

The return type is `TestAPI<A & B>`.
Type-level tests verify correct intersection inference and chained merging.

### Tests

Added runtime tests covering:

* Basic merging
* Override behavior
* Nested `describe`
* Shared base extension
* Chained merging

Added type-level tests verifying fixture intersection types.

All tests pass.

---

Resolves #9483 

## Tests

Run:

```
pnpm test:ci
```

---

## Documentation

No public docs changes required. The utility is self-contained and exposed via public API.

---

## Changesets

PR name follows conventional format:

```
feat(runner): add mergeTests utility to compose TestAPI fixtures
```

### DIFF
--- a/test/cli/test/merge-tests.test.ts
+++ b/test/cli/test/merge-tests.test.ts
@@ -197,11 +197,12 @@ test('mergeTests same fixture name with incompatible types still follows last-wi
     })
     return mergeTests(t1, t2)
   }, {
-    'basic.test.ts': ({ extendedTest, expect, expectTypeOf: _expectTypeOf }) => {
+    'basic.test.ts': ({ extendedTest, expect, expectTypeOf }) => {
       extendedTest('last wins regardless of shape', ({ foo }) => {
         // t2's static `123` wins over t1's async function fixture.
         // No stale `deps` or `value` should leak from t1's TestFixtureItem.
         expect(foo).toBe(123)
+        expectTypeOf(foo).toEqualTypeOf<number>()
       })
     },
   })
--- a/test/core/test/merge-tests.test.ts
+++ b/test/core/test/merge-tests.test.ts
@@ -1,4 +1,4 @@
-import { describe, expect, mergeTests, test } from 'vitest'
+import { describe, expect, expectTypeOf, mergeTests, test } from 'vitest'
 
 describe('mergeTests', () => {
   describe('basic merge behavior', () => {
@@ -30,7 +30,7 @@ describe('mergeTests', () => {
       expect(c).toBe(3)
     })
 
-    test('mergeTests(t) returns equivalent test instance', () => {
+    test('mergeTests with single test returns valid TestAPI', () => {
       const t = test.extend({ a: 1 })
       const merged = mergeTests(t)
 
@@ -175,6 +175,7 @@ describe('mergeTests', () => {
 
     mergedTypes('handles type conflicts by following last-wins metadata', ({ value }) => {
       expect(value).toBe(123)
+      expectTypeOf(value).toEqualTypeOf<number>()
     })
 
     const tNulls1 = test.extend({ undefinedValue: undefined, nullValue: null })
@@ -257,13 +258,6 @@ describe('mergeTests', () => {
     })
 
     test('mergeTests preserves built-in fixtures', () => {
-      const t = test.extend({ custom: 1 })
-      const merged = mergeTests(t)
-
-      expect(typeof merged).toBe('function')
-    })
-
-    test('preserves built-in fixtures', () => {
       const t1 = test.extend({ task: 'custom' } as any)
       const merged = mergeTests(t1)
       expect(typeof merged).toBe('function')
@@ -284,32 +278,16 @@ describe('mergeTests', () => {
       expect(typeof merged).toBe('function')
     })
 
-    test('mergeTests does not mutate source test instances', () => {
+    test('mergeTests does not mutate source tests', () => {
       const t1 = test.extend({ a: 1 })
       const t2 = test.extend({ b: 2 })
 
       const merged = mergeTests(t1, t2)
 
-      // original tests should still behave independently
-      const mergedAgain = mergeTests(t1)
+      const t1Extended = t1.extend({ c: 3 })
 
       expect(typeof merged).toBe('function')
-      expect(typeof mergedAgain).toBe('function')
-
-      // ensure original APIs are intact
-      expect(typeof t1.extend).toBe('function')
-      expect(typeof t2.extend).toBe('function')
-    })
-
-    test('source tests remain extendable after merge', () => {
-      const t1 = test.extend({ a: 1 })
-      const t2 = test.extend({ b: 2 })
-
-      mergeTests(t1, t2)
-
-      const extended = t1.extend({ c: 3 })
-
-      expect(typeof extended).toBe('function')
+      expect(typeof t1Extended).toBe('function')
     })
 
     test('subsequent merges work after local failure', () => {


Ujjwaljain16:
@sheremet-va  waiting for the review

packages/runner/src/fixture.ts
```diff
@@ -77,6 +77,25 @@ export class TestFixtures {
     return this._registrations
   }
 
+  /**
+   * @internal
```
sheremet-va: Everything here is internal, so no need to add this

packages/runner/src/fixture.ts
```diff
@@ -77,6 +77,25 @@ export class TestFixtures {
...
+   */
+  resolveFixtures(): UserFixtures {
```
sheremet-va: I don't like this idea, why do we need to convert already converted fixtures into user definitions and then convert them back _again_?

I would do the opposite and add a `merge` function that accepts another `Fixtures`, then we can just iterate over the registrations overriding them

packages/runner/src/suite.ts
```diff
@@ -1092,3 +1092,15 @@ function formatTemplateString(cases: any[], args: any[]): any[] {
   }
   return res
 }
+
+export function mergeTests<A, B>(
+  test: TestAPI<A>,
+  testB: TestAPI<B>,
+): TestAPI<A & B> {
```
sheremet-va: Hm, I assumed it can accept more than 2 (any number in fact). Would it make sense?

packages/runner/src/suite.ts
```diff
@@ -1092,3 +1092,15 @@ function formatTemplateString(cases: any[], args: any[]): any[] {
...
+  const ctx = getChainableContext(testB)
+  if (!ctx) {
+    throw new TypeError('Cannot merge tests: extension is not a valid TestAPI')
```
sheremet-va: This is a bit cryptic, what is a `TestAPI` to the user? Maybe just mention that this function requires an extended `test`

packages/vitest/src/public/index.ts
```diff
@@ -106,6 +106,7 @@ export {
   beforeEach,
   describe,
   it,
+  mergeTests,
```
sheremet-va: This function needs to be documented

packages/runner/src/suite.ts
Ujjwaljain16: Agreed this utility should support composing any number of test APIs.

I updated the signature to:

```ts
mergeTests<T extends readonly unknown[]>(...tests: T)
```

Runtime behavior:

Iteratively extends left  -> right.

Later fixtures override earlier ones (last-writer-wins)

Works for arbitrarily long argument lists.

Type behavior:

`MergeTestContexts<T>`  iterates over the tuple and intersects all extracted contexts.

Variadic inference is preserved even for nested merges.

This is now covered by runtime and type tests (including 3+ merges)

packages/runner/src/suite.ts
Ujjwaljain16: Agreed exposing internal terminology like "TestAPI" in runtime errors isn’t ideal

I updated the error message to:

`
"mergeTests requires extended test instances created via test.extend()"
`

This makes the requirement clearer without leaking internal implementation concepts

Ujjwaljain16:
@sheremet-va  
I’d really appreciate your guidance on a few architectural decisions here, especially around typing and fixture composition.

1) Strict Type Safety & Generic Constraints

While implementing `mergeTests`, I initially tried constraining inputs as:

```ts
T extends readonly TestAPI<any>[]
````

However this fails because `TestAPI` is invariant in its generic parameter due to the `TestCollectorCallable` definition:

```ts
<ExtraContext extends C>(...)
```

This makes:

```ts
TestAPI<{ a: number }>
```

not assignable to:

```ts
TestAPI<any>
```

So constraining to `TestAPI<any>[]` rejects valid usage.

To work around this, I relaxed the constraint to:

```ts
T extends readonly unknown[]
```

and enforced type validity structurally via `ExtractTestContext<T>`.

This preserves full context inference for consumers while avoiding the invariance trap.

My question:

Would you prefer a stricter nominal constraint even if it requires internal casts, or is structural validation acceptable here given the variance characteristics of `TestAPI`?

---

2)  Context Extraction Strategy

I initially attempted:

```ts
T extends TestAPI<infer C> ? C : never
```

But this proved unreliable because `TestAPI` is a complex intersection type with overloaded call signatures. In many cases TypeScript collapsed inference to `unknown` or `never`

Instead, I switched to structural extraction based on `beforeEach`, since it is the canonical API that consumes the contextual type:

```ts
T extends { beforeEach: (fn: infer F, ...args: any) => any }
  ? F extends (context: infer C, ...args: any) => any
    ? C
    : never
  : never
```

This also correctly accounts for the actual listener signature:

```ts
(context, suite)
```

Structurally matching on `beforeEach` felt more stable than relying on `TestAPI<infer C>`.

Would you prefer:

Structural extraction (as implemented), or
A nominal extraction approach even if it’s slightly less inference-friendly?

---

3) Fixture Merging Strategy

I explored implementing a low-level `merge()` directly on `TestFixtures`.

However, that would:

Bypass the `.extend()` validation pipeline
Duplicate dependency/scope validation logic
Risk drifting from canonical fixture behavior

Instead, I:

1. Added `toUserFixtures()` to serialize definition-time registrations
2. Re-applied them using `.extend()`

This keeps all validation inside the existing extension mechanism and preserves immutability.

Would you consider this acceptable alignment with Vitest’s design, or would you prefer fixture merging to happen at the `TestFixtures` level instead?


I’m happy to adjust direction based on your preference

sheremet-va:
> Would you prefer a stricter nominal constraint even if it requires internal casts, or is structural validation acceptable here given the variance characteristics of `TestAPI`?

Casting types internally is fine to me as long as public API is strict

>Structural extraction (as implemented), or A nominal extraction approach even if it’s slightly less inference-friendly?

I don't have a preference. As long as public API works correctly, any trick is good to me

> Bypass the .extend() validation pipeline
> Duplicate dependency/scope validation logic

Why would it do that? Didn't we already validated everything in the original `test.extend` calls? We are just merging them afterwards, no?

Could we abstract it in a better way, so it's more reusable without extra loops?

> Risk drifting from canonical fixture behavior

I don't know what that means

sheremet-va:
> To work around this, I relaxed the constraint to:

I think one of the possible solutions here is to define a lot of overloads of `mergeTests`:

```ts
function mergeTests<A, B, C>(test1: TestAPI<A>, test2: TestAPI<B>, test3: TestAPI<C>): TestAPI<A & B & C>
function mergeTests<A, B>(test1: TestAPI<A>, test2: TestAPI<B>): TestAPI<A & B>
// ... and so on
```

sheremet-va (changes requested):
It's pretty hard to review

The previous implementation is clean and easy to follow, the new one removes all the helpful comments and introduces concepts and code that's just hard to follow

There are `registrations` and `overrides`. `registrations` are scoped to the last `test.extend`, and `overrides` are applied to tests in specified suites, that's it. What this PR adds are some kind of pure registrations, and keeps track of parents and ancestors (why? we have registrations already - overridden suites also include all fixtures already)

Ujjwaljain16:
Appreciate the push to simplify you were right about the overhead. I've refactored the implementation back to the baseline 'flat model' and moved to direct Map-merging in extend()
It's much cleaner now and avoids the conversion loop entirely.
I also restored all original documentation and synchronized the internal types with main to ensure there's no drift. Everything is verified and synced with upstream

packages/runner/src/suite.ts
```diff
@@ -1092,3 +1092,46 @@ function formatTemplateString(cases: any[], args: any[]): any[] {
...
+ */
+export function mergeTests<A>(a: TestAPI<A>): TestAPI<A>
+export function mergeTests<A, B>(a: TestAPI<A>, b: TestAPI<B>): TestAPI<A & B>
+export function mergeTests<A, B, C>(a: TestAPI<A>, b: TestAPI<B>, c: TestAPI<C>): TestAPI<A & B & C>
+export function mergeTests<A, B, C, D>(a: TestAPI<A>, b: TestAPI<B>, c: TestAPI<C>, d: TestAPI<D>): TestAPI<A & B & C & D>
+export function mergeTests<A, B, C, D, E>(a: TestAPI<A>, b: TestAPI<B>, c: TestAPI<C>, d: TestAPI<D>, e: TestAPI<E>): TestAPI<A & B & C & D & E>
+export function mergeTests<A, B, C, D, E, F>(a: TestAPI<A>, b: TestAPI<B>, c: TestAPI<C>, d: TestAPI<D>, e: TestAPI<E>, f: TestAPI<F>): TestAPI<A & B & C & D & E & F>
+export function mergeTests(...tests: TestAPI<any>[]): TestAPI<any> {
+  if (tests.length === 0) {
+    throw new TypeError('mergeTests requires at least one test')
```
sheremet-va, line 1121: This is not tested

packages/runner/src/suite.ts
```diff
@@ -1092,3 +1092,46 @@ function formatTemplateString(cases: any[], args: any[]): any[] {
...
+  }
+
+  // Use the first test as the base
+  let [currentTest, ...rest] = tests
+
+  for (const nextTest of rest) {
+    const nextContext = getChainableContext(nextTest)
```
sheremet-va: This is not tested

packages/runner/src/suite.ts
```diff
@@ -1092,3 +1092,46 @@ function formatTemplateString(cases: any[], args: any[]): any[] {
...
+    if (!nextContext || typeof nextContext.getFixtures !== 'function') {
+      throw new TypeError('mergeTests requires extended test instances created via test.extend()')
+    }
+
+    // Extract fixtures from the next test and extend the current test
+    // This behaves exactly like currentTest.extend(nextFixtures)
+    const currentContext = getChainableContext(currentTest)
+    if (!currentContext) {
```
sheremet-va: This is not tested

packages/runner/src/fixture.ts
```diff
@@ -54,13 +54,26 @@ export class TestFixtures {
...
+        registrations.set(name, item)
+      }
+      return new TestFixtures(registrations)
+    }
+
     const registrations = this.parseUserFixtures(runner, userFixtures, isTopLevel)
     return new TestFixtures(registrations)
   }
 
+  getFixtures(): TestFixtures {
```
sheremet-va: What is this? Why do you need this if you already have access to the instance?

packages/runner/src/utils/chain.ts
```diff
@@ -32,7 +32,7 @@ export function createChainable<T extends string, Args extends any[], R = any>(
     Object.defineProperty(chain, kChainableContext, {
       value: {
         withContext: () => chain.bind(context),
-        getFixtures: () => (context as any).fixtures,
+        get fixtures() { return (context as any).fixtures },
```
sheremet-va: What is going on? Why are you touching this at all? How is this related to the changes in this PR?

You added a function with the same name from the context to `TestFixtures` that you never used, now you rename the function with that name in other cases that are not related to this PR?

test/cli/test/merge-tests.test.ts
```diff
@@ -0,0 +1,914 @@
...
+    'basic.test.ts': ({ extendedTest, expect }) => {
+      extendedTest('uses both function fixtures', ({ a, b }) => {
+        expect(a).toBe('value-a')
+        expect(b).toBe('value-b')
+      })
+    },
+  })
+
+  expect(stderr).toBe('')
+  expect(fixtures).toContain('>> fixture | init a')
```
sheremet-va: Please, read `AGENTS.md` - test files MUST NOT use `toContain` matcher to test things, they don't test stack traces and general output that is much more useful

test/cli/test/merge-tests.test.ts
```diff
@@ -0,0 +1,914 @@
...
+  expect(fixtures).toContain('>> fixture | init b')
+  expect(tests).toContain('✓ basic.test.ts > uses both function fixtures')
```
sheremet-va: Tests is an array, it is much more useful to see the whole array in the assertion. This kind of test will not catch bugs

test/cli/test/merge-tests.test.ts
```diff
@@ -0,0 +1,1008 @@
...
+      extendedTest('file 2 uses merged fixtures', ({ shared, extra }) => {
+        expect(shared).toBe('shared-val')
+        expect(extra).toBe(42)
+      })
+    },
+  })
+
+  expect(stderr).toMatchInlineSnapshot(`""`)
+  // test-scoped fixture inits once per test across both files
+  expect(fixtures.match(/shared setup/g)?.length).toBe(2)
```
sheremet-va: You can just use `toMatchInlineSnapshot`

test/cli/test/merge-tests.test.ts
```diff
@@ -0,0 +1,1008 @@
...
+  }, {
+    'basic.test.ts': ({ extendedTest }) => {
+      extendedTest('test1', ({ a: _a, b: _b }) => {})
+      extendedTest('test2', ({ a: _a, b: _b }) => {})
+    },
+  })
+
+  expect(stderr).toMatchInlineSnapshot(`""`)
+  // Each test gets its own setup/teardown cycle
+  expect(fixtures.match(/a setup/g)?.length).toBe(2)
```
sheremet-va: Just use `toMatchInlineSnapshot` on `fixtures`

test/cli/test/merge-tests.test.ts
```diff
@@ -0,0 +1,1008 @@
...
+      extendedTest('test2', ({ fileFix, testFix }) => {
+        expect(fileFix).toBe('file-value')
+        expect(testFix).toBe('test-value')
+      })
+    },
+  })
+
+  expect(stderr).toMatchInlineSnapshot(`""`)
+  // file-scoped fixture inits once and tears down once
+  expect(fixtures.match(/fileFix setup/g)?.length).toBe(1)
```
sheremet-va: Where did you get this approach? Just use `toMatchInlineSnapshot`, fixtures is a static string

test/cli/test/merge-tests.test.ts
```diff
@@ -0,0 +1,1008 @@
...
+  expect(fixtures.match(/b teardown/g)?.length).toBe(2)
+  expect(tests).toMatchInlineSnapshot(`
+    " ✓ basic.test.ts > test1 <time>
+     ✓ basic.test.ts > test2 <time>"
+  `)
+})
+
+// --- Dependency chains across merged tests ---
+
+test('mergeTests resolves dependency chain from shared base', async () => {
```
sheremet-va: This is the only test with dependencies and they are not even used in `mergeTests`. Please, add tests that could actually catch bugs, most of the current tests are basically the same test with different values

For example, what happens if you merge with fixtures that have the same fixtures? What happens to fixtures that rely have these in dependencies? Can you merge the same fixtures with different types? What happens when the same fixture is defines with a different scope or other settings?

Please, add some care to your work

test/cli/test/merge-tests.test.ts
```diff
@@ -0,0 +1,1205 @@
...
+  const { stderr, tests } = await runMergeFixtureTests(() => {
+    const t1 = it.extend<{ a: string }>({ a: 'from-t1' })
+    const t2 = it.extend<{ b: string }>({ b: 'from-t2' })
+    // Create merged, then use original t1 independently
+    mergeTests(t1, t2)
+    // Return t1 unchanged — it must NOT see t2's fixture `b`
+    return t1
+  }, {
+    'basic.test.ts': ({ extendedTest, expect }) => {
+      extendedTest('original still only has its own fixtures', ({ a }) => {
```
sheremet-va: I am confused by this test, it will always pass, isn't it? `a` exists in both `t1` and in `mergeTests(t1, t2)`, t2 doesn't affect it at all

test/cli/test/merge-tests.test.ts
```diff
@@ -0,0 +1,1205 @@
...
+    'basic.test.ts': ({ extendedTest }) => {
+      extendedTest('should fail', ({ fileFix: _f }) => {})
+    },
+  })
+  expect(stderr).toMatchInlineSnapshot(`
+    "
+    ⎯⎯⎯⎯⎯⎯ Failed Suites 1 ⎯⎯⎯⎯⎯⎯⎯
+
+     FAIL  basic.test.ts [ basic.test.ts ]
+    FixtureDependencyError: The file "fileFix" fixture cannot depend on a test fixture "local".
```
sheremet-va: This doesn't fail during `mergeTests`, this fails in `it.extend` (which is visible in the stack trace)

What you need to check is when another test has the same fixture but with a different scope

It will most likely NOT FAIL because you don't have any validation in your `extend` method

test/cli/test/merge-tests.test.ts
```diff
@@ -0,0 +1,1205 @@
...
+    })
+    return mergeTests(t1, t2)
+  }, {
+    'basic.test.ts': ({ extendedTest }: any) => {
+      extendedTest('cycle triggers error', ({ a: _a }: any) => {})
+    },
+  })
+  // Error is thrown at t1.extend() time — the merge never runs.
+  // A cycle only detectable post-merge is structurally impossible.
+  expect(stderr).toMatchInlineSnapshot(`
```
sheremet-va: The same thing here, this test is pointless because it already exist. You need to only test what happens in `mergeTests` when BOTH tests define the same fixtures

test/cli/test/merge-tests.test.ts
```diff
@@ -0,0 +1,1205 @@
...
+    const t2 = it.extend<{ foo: number }>({
+      foo: 123,
+    })
+    return mergeTests(t1, t2)
+  }, {
+    'basic.test.ts': ({ extendedTest, expect }) => {
+      extendedTest('last wins regardless of shape', ({ foo }) => {
+        // t2's static `123` wins over t1's async function fixture.
+        // No stale `deps` or `value` should leak from t1's TestFixtureItem.
+        expect(foo).toBe(123)
```
sheremet-va, line 204: Add here `expectTypeOf` - I have a suspicion it is `never` instead of `number`

test/cli/test/merge-tests.test.ts
```diff
@@ -0,0 +1,1205 @@
...
+    const tb = it.extend<{ f: string }>({
+      f: async ({}, use) => use('test-val'),
+    })
+    return mergeTests(ta, tb)
+  }, {
+    'basic.test.ts': ({ extendedTest, expect }) => {
+      extendedTest('file→test wins test-val', ({ f }) => expect(f).toBe('test-val'))
+    },
+  })
+  expect(e1).toMatchInlineSnapshot(`""`)
```
sheremet-va: This should actually fail - you cannot override fixture with a different scope, this is how `extend` works already

sheremet-va (changes requested):
I am really tired of reviewing this, please read your tests before asking for a review

Writing tests should find bugs in your code, but this AI instead just documents the wrong behaviour in the test

Ujjwaljain16:
@sheremet-va

Thanks for sticking with this review and sorry for the earlier iterations  i realize the previous versions made this harder to review than it should have been

I've simplified the implementation and addressed the feedback
Would appreciate another look whenever you have time

Ujjwaljain16:
Core tests cover:

merge semantics
nested merges
override propagation
dependency graphs
diamond inheritance
type inference
validation errors
API surface
scaling

CLI tests cover:

lifecycle ordering
async fixture setup/teardown
worker/file/test scopes
dependency remapping
override pruning
circular dependency runtime detection
cross-file behavior
injected fixtures
hooks integration
self-merge
deep inheritance chains
stress tests (50+ fixtures)

Ujjwaljain16:
```const baseFn = (tests[0] as any).fn```

 Reuse the execution function from the first TestAPI instance.
`mergeTests` only merges fixture graphs; it intentionally preserves the runtime behavior of the base test implementation. Subsequent TestAPI instances contribute fixtures only and do not alter the underlying test execution semantics

Ujjwaljain16:
```const builtinNames = ['task', 'signal', 'onTestFailed', 'onTestFinished', 'skip', 'annotate']```


 Built-in fixtures must remain stable across merged tests
They are defined internally by Vitest and are expected to behave identically for all TestAPI nstances
During merge we therefore preserve the first occurrence and avoid overriding them from later tests to prevent accidental mutation of runner behavior

PR #9665, opened 2026-02-14
fix: deprecate `toBe*` spy assertions in favor of `toHaveBeen*` (and `toThrowError`)

sheremet-va:
### Description

These existed primarily for a better compatibility story with Jest, even though they were already deprecated there back then. Let's cleanup and remove them in the next major completely, but they need to be deprecated first.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/expect/src/jest-expect.ts
+++ b/packages/expect/src/jest-expect.ts
@@ -633,7 +633,7 @@ export const JestChaiExpect: ChaiPlugin = (chai, utils) => {
     }
   })
   def(
-    ['toHaveBeenNthCalledWith', 'nthCalledWith'],
+    'toHaveBeenNthCalledWith',
     function (times: number, ...args: any[]) {
       const spy = getSpy(this)
       const spyName = spy.getMockName()
@@ -657,7 +657,7 @@ export const JestChaiExpect: ChaiPlugin = (chai, utils) => {
     },
   )
   def(
-    ['toHaveBeenLastCalledWith', 'lastCalledWith'],
+    'toHaveBeenLastCalledWith',
     function (...args: any[]) {
       const spy = getSpy(this)
       const spyName = spy.getMockName()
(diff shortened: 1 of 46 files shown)


packages/expect/src/types.ts
```diff
@@ -494,16 +497,6 @@ export interface JestAssertion<T = any> extends jest.Matchers<void, T>, CustomMa
...
 
-  /**
-   * Ensure that a mock function is called with specific arguments on an Nth call.
-   *
-   * Alias for `expect.toHaveBeenNthCalledWith`.
-   *
-   * @example
-   * expect(mockFunc).nthCalledWith(2, 'secondArg');
-   */
-  nthCalledWith: <E extends any[]>(nthCall: number, ...args: E) => void
```
sheremet-va, line 505: All of these are actually defined already in a `ChaiMockAssertion` type on L776

https://github.com/vitest-dev/vitest/blob/4ec6cb3054ac0fa0218837f96847632470346358/packages/expect/src/chai-style-assertions.ts#L63

packages/expect/src/jest-expect.ts
```diff
@@ -633,7 +633,7 @@ export const JestChaiExpect: ChaiPlugin = (chai, utils) => {
     }
   })
   def(
-    ['toHaveBeenNthCalledWith', 'nthCalledWith'],
+    'toHaveBeenNthCalledWith',
```
hi-ogawa, line 636: Are these removed because they are already defined by chai and these are kept working?

hi-ogawa (commented):
So there are three variants like `toHaveBeen*`, `toBe*` and `*` and deprecating `toBe*`. The one defined as `*` is from chai, which has been working and keep working. But then why does it need new doc badge ` <Version>4.1.0</Version>`?

packages/expect/src/jest-expect.ts
sheremet-va: yes

sheremet-va:
> But then why does it need new doc badge ` <Version>4.1.0</Version>`?

Most of `*` were added in 4.1, but the older ones do not need the badge (there are 4 of them)

hi-ogawa (dismissed):

PR #9677, opened 2026-02-16
feat(browser): add `findElement` and enable strict mode in webdriverio and preview

### LINKED ISSUE #9578: [browser mode] Calling .click() or another interaction function on a locator that resolves to more than one element should error
### Describe the bug

It's common that locators can resolve to several elements. When calling `element()` on it, it errors accordingly, providing a useful feedback to the user.

However when using an interaction method instead, vitest doesn't error. This can be confusing to the developer, by interacting with an unexpected control.

### Reproduction

See the testcase in https://github.com/julienw/vitest-multiple-results-issue
Run:
```
npm i
npm test
```

Notice that the second test fails. The test checks that .click() errors.

### System Info

```shell
System:
    OS: Linux 6.12 Debian GNU/Linux 13 (trixie) 13 (trixie)
    CPU: (16) x64 11th Gen Intel(R) Core(TM) i9-11900H @ 2.50GHz
    Memory: 37.99 GB / 62.53 GB
    Container: Yes
    Shell: 5.2.37 - /bin/bash
  Binaries:
    Node: 24.12.0 - /home/julien/.nvm/versions/node/v24.12.0/bin/node
    Yarn: 1.22.22 - /usr/bin/yarn
    npm: 11.6.2 - /home/julien/.nvm/versions/node/v24.12.0/bin/npm
    pnpm: 10.9.0 - /home/julien/.local/share/pnpm/pnpm
    bun: 1.0.29 - /home/julien/.bun/bin/bun
    Deno: 2.0.0 - /home/julien/.cargo/bin/deno
  Browsers:
    Chrome: 144.0.7559.109
    Chromium: 144.0.7559.96
    Firefox: 140.7.0esr
    Firefox Developer Edition: 140.7.0esr
  npmPackages:
    @vitejs/plugin-react: ^5.1.3 => 5.1.3 
    @vitest/browser: ^4.0.18 => 4.0.18 
    @vitest/browser-webdriverio: ^4.0.18 => 4.0.18 
    vite: ^7.3.1 => 7.3.1 
    vitest: ^4.0.18 => 4.0.18 
    vitest-browser-react: ^2.0.5 => 2.0.5
```

### Used Package Manager

npm

### Validations

sheremet-va:
### Description

Resolves #9578

TODO
- [x] Tests

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/api/browser/locators.md
+++ b/docs/api/browser/locators.md
@@ -935,6 +935,61 @@ page.getByText('Hello').elements() // ✅ [HTMLElement, HTMLElement]
 page.getByText('Hello USA').elements() // ✅ []
 ```
 
+### waitForElement <Version>4.1.0</Version> {#waitforelement}
+
+```ts
+function waitForElement(options?: SelectorOptions): Promise<HTMLElement | SVGElement>
+```
+
+::: danger WARNING
+This is an escape hatch for library authors and 3d-party APIs that do not support locators directly. If you are interacting with the element, use [builtin methods](#methods) instead.
+:::
+
+This method returns an element matching the locator. Unlike [`.element()`](#element), this method will wait and retry until a matching element appears in the DOM, using increasing intervals (0, 20, 50, 100, 100, 500ms).
+
+If _no element_ is found before the timeout, an error is thrown. By default, the timeout matches the test timeout.
+
+If _multiple elements_ match the selector and `strict` is `true` (the default), an error is thrown immediately without retrying. Set `strict` to `false` to return the first matching element instead.
+
+It accepts options:
+
+- `timeout: number` - How long to wait in milliseconds until a single element is found. By default, this has the same timeout as the test.
+- `strict: boolean` - When `true` (default), throws an error if multiple elements match the locator. When `false`, returns the first matching element.
+
+Consider the following DOM structure:
+
+```html
+<div>Hello <span>World</span></div>
+<div>Hello Germany</div>
+<div>Hello</div>
+```
+
+These locators will resolve successfully:
+
+```ts
+await page.getByText('Hello World').waitForElement() // ✅ HTMLDivElement
+await page.getByText('World').waitForElement() // ✅ HTMLSpanElement
+await page.getByText('Hello Germany').waitForElement() // ✅ HTMLDivElement
+```
+
+These locators will throw an error:
+
+```ts
+// multiple elements match, strict mode rejects
+await page.getByText('Hello').waitForElement() // ❌
+await page.getByText(/^Hello/).waitForElement() // ❌
+
+// no matching element before timeout
+await page.getByText('Hello USA').waitForElement() // ❌
+```
+
+Using `strict: false` to allow multiple matches:
+
+```ts
+// returns the first matching element instead of throwing
+await page.getByText('Hello').waitForElement({ strict: false }) // ✅ HTMLDivElement
+```
+
 ### all
 
 ```ts
--- a/docs/api/browser/locators.md
+++ b/docs/api/browser/locators.md
@@ -938,11 +938,13 @@ page.getByText('Hello USA').elements() // ✅ []
 ### waitForElement <Version>4.1.0</Version> {#waitforelement}
 
 ```ts
-function waitForElement(options?: SelectorOptions): Promise<HTMLElement | SVGElement>
+function waitForElement(
+  options?: SelectorOptions
+): Promise<HTMLElement | SVGElement>
 ```
 
 ::: danger WARNING
-This is an escape hatch for library authors and 3d-party APIs that do not support locators directly. If you are interacting with the element, use [builtin methods](#methods) instead.
+This is an escape hatch for library authors and 3d-party APIs that do not support locators directly. If you are interacting with the element, use other [builtin methods](#methods) instead.
 :::
 
 This method returns an element matching the locator. Unlike [`.element()`](#element), this method will wait and retry until a matching element appears in the DOM, using increasing intervals (0, 20, 50, 100, 100, 500ms).
@@ -953,7 +955,7 @@ If _multiple elements_ match the selector and `strict` is `true` (the default), 
 
 It accepts options:
 
-- `timeout: number` - How long to wait in milliseconds until a single element is found. By default, this has the same timeout as the test.
+- `timeout: number` - How long to wait in milliseconds until at least one element is found. By default, this shares timeout with the test.
 - `strict: boolean` - When `true` (default), throws an error if multiple elements match the locator. When `false`, returns the first matching element.
 
 Consider the following DOM structure:
--- a/docs/api/browser/locators.md
+++ b/docs/api/browser/locators.md
@@ -944,7 +944,7 @@ function waitForElement(
 ```
 
 ::: danger WARNING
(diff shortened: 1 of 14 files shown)


docs/api/browser/locators.md
```diff
@@ -935,6 +935,63 @@ page.getByText('Hello').elements() // ✅ [HTMLElement, HTMLElement]
...
+### waitForElement <Version>4.1.0</Version> {#waitforelement}
+
+```ts
+function waitForElement(
+  options?: SelectorOptions
+): Promise<HTMLElement | SVGElement>
+```
+
+::: danger WARNING
+This is an escape hatch for library authors and 3d-party APIs that do not support locators directly. If you are interacting with the element, use other [builtin methods](#methods) instead.
```
hi-ogawa: What does this mean? I'm not sure what "library", "3rd-party API", nor "not supporting locators" mean.

packages/browser-webdriverio/src/locators.ts
```diff
@@ -42,33 +55,84 @@ class WebdriverIOLocator extends Locator {
...
   public override hover(options?: UserEventHoverOptions): Promise<void> {
-    return super.hover(processHoverOptions(options))
+    return ensureAwaited(async (error) => {
+      const element = await this.waitForElement(options)
+      return this.withElement(element, error).hover(processHoverOptions(options))
+    })
   }
 
   public override dropTo(target: Locator, options?: UserEventDragAndDropOptions): Promise<void> {
+    // playwright doesn't enforce a single element, it selects the first found one
```
hi-ogawa: Does this comment mean that webdriverio doesn't need waitForElement because playwright doesn't enforce single element either? Just looks puzzling at first even with this PR as a context.

test/browser/test/userEvent.test.ts
hi-ogawa, line 1: How about test case for { strict: false }?

Also any test for `waitForElement` API directly?

packages/browser/src/client/tester/tester-utils.ts
hi-ogawa, line 152: `providerOptions.actionTimeout` can now affect non playwright provider behavior, so this needs to return `{ timeout: actionTimeout }` to actually let `actionTimeout` affect in that case?

packages/browser/src/client/tester/locators/index.ts
```diff
@@ -293,12 +313,82 @@ export abstract class Locator {
...
+      const elements = this.elements()
+      if (elements.length === 1) {
+        return elements[0]
+      }
+      if (elements.length > 1) {
+        if (strict) {
+          throw createStrictModeViolationError(this._pwSelector || this.selector, elements)
+        }
+        return elements[0]
+      }
```
hi-ogawa, line 333: Is it intended that strict violation doesn't retry? Does it match with playwright?

packages/browser/src/client/tester/tester-utils.ts
```diff
@@ -209,15 +204,22 @@ export function escapeForTextSelector(text: string | RegExp, exact: boolean): st
   return `${JSON.stringify(text)}${exact ? 's' : 'i'}`
 }
 
-export function convertToSelector(elementOrLocator: Element | Locator): string {
+const provider = getBrowserState().provider
+const kElementLocator = Symbol.for('$$vitest:locator-resolved')
+
+export async function convertToSelector(elementOrLocator: Element | Locator): Promise<string> {
```
hi-ogawa: `convertToSelector` calling `waitForElement` without passing options feels odd to me. I see `userEvent.type` and `screenshot` calls `convertToSelector`, but they should likely support the same `SelectorOptions` and should be passed to around to reach `convertToSelector`.

packages/browser-webdriverio/src/locators.ts
sheremet-va: Yes, it means exactly that, Playwright doesn't enforce a single element for this method specifically, which is also mentioned in their docs

packages/browser/src/client/tester/locators/index.ts
sheremet-va: Yes, it's mentioned in the docs too

docs/api/browser/locators.md
sheremet-va: What is causing confusion? Library is any code that works in the browser, like jquery or any UI library, the same for 3d party APIs, like vue's `createApp`, for example - it accepts an HTML element, but doesn't support a locator directly

This section tries to say that you should not use this API directly, it is exposed for developers that build things on top of Vitest, or if the external API doesn't accept locator at all, like jquery

packages/browser/src/client/tester/tester-utils.ts
sheremet-va: No, `actionTimeout` is set only if provider is `playwright`

https://github.com/sheremet-va/vitest/blob/209dcd1db47b9f4e30cf8c5911bab9062d0411fe/packages/vitest/src/node/config/serializeConfig.ts#L115

hi-ogawa: Doesn't it make sense to support `actionTimeout` for webdriverio now we have this logic on vitest side (as vitest webdriverio provider)?

sheremet-va: Sure, I guess if it does the same thing

sheremet-va: Hm, it doesn't really do the same thing. `actionTimeout` in playwright also includes the time it took to take the action itself, not just to find the element. This would require a large refactor to track how timers interact with each other. Playwright has an abstraction for that, but it's really hard to follow at times

docs/api/browser/locators.md
sheremet-va: I clarified the sentence, does it sound better?

> This is an escape hatch for cases where you need the raw DOM element — for example, to pass it to a third-party library like FormKit that doesn't accept Vitest locators. If you are interacting with the element yourself, use other [builtin methods](#methods) instead.

test/browser/test/userEvent.test.ts
sheremet-va: Added more tests

docs/api/browser/locators.md
hi-ogawa: Ah, I understood better. I wasn't looking at the return value `HTMLElement | SVGElement` at all, but simply thinking as playwright's `toBeAttached` or `locator.waitFor` API (which is void-returning).

hi-ogawa (dismissed):
Looks good!

packages/browser/src/client/tester/tester-utils.ts
hi-ogawa: I was thinking `waitForElement` timeout only would be good enough approximation.
Also realized, for complete playwright like behavior, I think we'd need to check actionability and not just element exists (as retry conditions). I don't know webdriverio action API does anything like that.

sheremet-va: Yes, I intentionally ignore auctionability for now

docs/api/browser/locators.md
sheremet-va: Maybe it should be renamed to `find()` instead then? Or `findElement()` (similar to how testing-library names things)

hi-ogawa: `findElement` sounds good :+1:

PR #9705, opened 2026-02-20
feat(ui): implement filter for slow tests

### LINKED ISSUE #9704: Slow test filter and highlighting in explorer UI

While it's already possible to sort tests by duration, this is based on the whole test file and not individual tests. As a result, it's not possible to quickly navigate to and analyze files that exceed the `slowTestThreshold`.

### Suggested solution

Add a new filter for slow tests. This filter should operate on a test-level. Additionally, the duration of slow tests could use the yellow text color in the explorer UI.

### Alternative

_No response_

### Additional context

_No response_

### Validations

DerYeger:
Also highlight slow tests with yellow color for duration text

### Description

Add a new filter for slow tests. This filter should operate on a test-level. Additionally, the duration of slow tests could use the yellow text color in the explorer UI.

Resolves #9704

### Tests

### Documentation

### Changesets


hi-ogawa:
Can you add demo screenshot for easier review?

DerYeger:
Yes, sorry I forgot them earlier. Just pushed a minor change for better filter columns as well.

image

image

image

image

DerYeger:
We could also add the number of slow tests to the dashboard if you'd agree to that, as in the below screenshot. If you don't think this is a good addition let me know and I'll drop the commit.

image

DerYeger:
Finally, here's one example where I removed the custom `slowTestThreshold: 1000` from the config.

image

userquin:
I'm fixing some colors (low contrast with light theme) and UI PW tests, including `slowTestThreshold` at `Slow` filter.

Excluded slow entry at UI PW tests since the time at CI is non deterministic, we may need to add a custom test to display `N Slow` check, right now excluded via regex.

<details>
<summary>New colors for light theme</summary>

new colors for light theme
</details>

test/ui/test/html-report.spec.ts
```diff
@@ -52,7 +52,7 @@ test.describe('html report', () => {
     await page.goto(pageUrl)
 
     // dashboard
-    await expect(page.locator('[aria-labelledby=tests]')).toContainText('16 Pass 2 Fail 18 Total')
+    await expect(page.locator('[aria-labelledby=tests]')).toContainText(/(16 Pass | 2 Fail | 18 Total)/)
```
userquin: we should split this regex with 3 assertions

test/ui/test/ui.spec.ts
```diff
@@ -71,7 +71,7 @@ test.describe('ui', () => {
     await page.goto(pageUrl)
 
     // dashboard
-    await expect(page.locator('[aria-labelledby=tests]')).toContainText('16 Pass 2 Fail 18 Total')
+    await expect(page.locator('[aria-labelledby=tests]')).toContainText(/(16 Pass | 2 Fail | 18 Total)/)
```
userquin: same here

userquin:
> Finally, here's one example where I removed the custom `slowTestThreshold: 1000` from the config.
> 

What test are you running for `Expected Fail`? I need to fix the cyan color on light theme...

sheremet-va:
> What test are you running for `Expected Fail`? I need to fix the cyan color on light theme...

```js
test.fail('name', () => {
  throw new Error('expected error')
})
```

packages/ui/client/composables/explorer/collector.ts
```diff
@@ -394,11 +401,15 @@ function collectTests(file: File, search: SearchMatcher = () => true, filter?: F
     ignored: 0,
     todo: 0,
     expectedFail: 0,
+    slow: 0,
   } satisfies CollectFilteredTests
 
   for (const t of testsCollector(file)) {
     if (!filter || testMatcher(t, search, filter)) {
       data.total++
+      if (isSlowTestTask(t)) {
```
sheremet-va, line 410: What I don't like with this check is that `total` number is not a sum of all other states anymore.

sheremet-va (changes requested):

userquin:
Looks like PW line number is wrong on my Windows laptop 🤔:

<details>
<summary>Wrong PW test line number</summary>

Wrong PW test line number
</details>

packages/ui/client/composables/explorer/collector.ts
userquin: you're right, checking logic, thx Vlad

userquin: I'm not sure how to solve this, we'll need 2 filters (split `testMatcher/matchState` at `collector.ts` and `matcher` at `filter.ts`) to collect totals and filtering nodes.

Using a new flag (`filterSlow = true`) at  `testMatcher/matchState`, `collectTests` will use `false`:

image

userquin: check https://github.com/vitest-dev/vitest/pull/9705/commits/94b0c07a9ed9971920e7260b56a730a46702df13 not sure if this is ok

userquin: I guess it is fine since we only show files/suites containing filtered tests, since we're filtering the ui with `forceSlow = true` shouldn't be a problem, we're changing the way we collect the totals (`forceSlow = false`; files or individual tests, PASS (15) is the number of files, if we check `Tests Only` will display the tests PASSING, slow or not)

sheremet-va: I am talking about this total:

Screenshot 2026-02-21 at 19 58 50

They don't add up to 2311 (although to be fair, without slow they only add up to 2310 - what happened to the other 1?)

This number should just be displayed differently (not on the same row at least)

userquin: arggg, so it is test/core with `threads` project only, reverting changes; will check the sum/count

sheremet-va: Also 200 files don't add up to Pass 198 + Fail 1 🤔

userquin: arggg, something is wrong also at default reporter:

image

userquin: missing running test fixed here https://github.com/vitest-dev/vitest/commit/cc735970adab31d500a5ad299fdb0e9bd2bed219

<details>
<summary>Missing running test</summary>

running test

running test mode
</details>

<details>
<summary>Missing running test fixed once main merged here</summary>

running test fixed

</details>

userquin:
I guess we should move Slow after Total or after Todo: `total = pass + fail + expected fail + skip + todo`

userquin:
We still need to fix the filter layout, once `slowTestThreshold` added to the Slow filter the layout is broken:

filter layout broken

userquin:
@DerYeger feel free to fix the filter layout, I'm checking  using 3 columns with `< 700` (maybe even more) instead `< 584`, maybe we can use min width with flex instead

DerYeger:
> @DerYeger feel free to fix the filter layout, I'm checking using 3 columns with `< 700` (maybe even more) instead `< 584`, maybe we can use min width with flex instead

Many thanks for cleaning this up! I switched to the suggested flex layout, since the dependency of the slow test filter width on user input will inherently cause issues in some cases if we go with the existing pixel-width based resize observer.

https://github.com/user-attachments/assets/d6beb991-9679-4dcb-b00f-3cc7f0a1cddc

userquin (approved):
LGTM

Thx

packages/ui/client/composables/explorer/collector.ts
DerYeger: Is anything missing? The failing tests is not related to the changes, right?

sheremet-va (approved):
Thanks!

PR #9729, opened 2026-02-24
fix: cancelling should stop current test immediately

AriPerkkio:
### Description

Speeds up test cancellation by doing similar test function rejection as timeout errors do.

For Node API users this is essential, as they cannot do forceful test cancellation without shutting down whole process.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/runner/src/run.ts
+++ b/packages/runner/src/run.ts
@@ -825,7 +825,7 @@ function markPendingTasksAsSkipped(suite: Suite, runner: VitestRunner, note?: st
     if (!t.result || t.result.state === 'run') {
       t.mode = 'skip'
       t.result = { ...t.result, state: 'skip', note }
-      updateTask('test-finished', t, runner)
+      updateTask('test-cancel', t, runner)
     }
 
     if (t.type === 'suite') {
--- a/packages/runner/src/run.ts
+++ b/packages/runner/src/run.ts
@@ -792,6 +792,12 @@ function failTask(result: TaskResult, err: unknown, diffOptions: DiffOptions | u
     return
   }
 
+  if (err instanceof TestRunAbortError) {
+    result.state = 'skip'
+    result.note = err.message
+    return
+  }
+
   result.state = 'fail'
   const errors = Array.isArray(err) ? err : [err]
   for (const e of errors) {
@@ -810,6 +816,20 @@ function markTasksAsSkipped(suite: Suite, runner: VitestRunner) {
     updateTask('test-finished', t, runner)
     if (t.type === 'suite') {
       markTasksAsSkipped(t, runner)
+    }
+  })
+}
+
+function markPendingTasksAsSkipped(suite: Suite, runner: VitestRunner, note?: string) {
+  suite.tasks.forEach((t) => {
+    if (!t.result || t.result.state === 'run') {
+      t.mode = 'skip'
+      t.result = { ...t.result, state: 'skip', note }
+      updateTask('test-finished', t, runner)
+    }
+
+    if (t.type === 'suite') {
+      markPendingTasksAsSkipped(t, runner, note)
     }
   })
 }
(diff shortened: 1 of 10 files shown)


test/cli/test/cancel-run.test.ts
```diff
@@ -1,21 +1,25 @@
...
 import { Readable, Writable } from 'node:stream'
 import { stripVTControlCharacters } from 'node:util'
 import { createDefer } from '@vitest/utils/helpers'
 import { expect, onTestFinished, test, vi } from 'vitest'
 import { createVitest, registerConsoleShortcuts } from 'vitest/node'
 
-test('can force cancel a run', async () => {
+const CTRL_C = '\x03'
+
+test('can force cancel a run via CLI', { timeout: 5_000 }, async () => {
```
AriPerkkio:
```suggestion
test('can force cancel a run via CLI', async () => {
```

packages/runner/src/run.ts
```diff
@@ -814,6 +820,20 @@ function markTasksAsSkipped(suite: Suite, runner: VitestRunner) {
   })
 }
 
+function markPendingTasksAsSkipped(suite: Suite, runner: VitestRunner, note?: string) {
+  suite.tasks.forEach((t) => {
+    if (!t.result || t.result.state === 'run') {
+      t.mode = 'skip'
+      t.result = { ...t.result, state: 'skip', note }
+      updateTask('test-finished', t, runner)
```
sheremet-va: Shouldn't we trigger both `test-prepare` _and_ `test-finished` if the test was skipped? For example, if `suite-finished` is returned and it was skipped, we report every every: https://github.com/vitest-dev/vitest/blob/0bd2bb44fee11fe69a33f201100c318c4e9bf1f2/packages/vitest/src/node/test-run.ts#L299

sheremet-va (changes requested):

packages/runner/src/run.ts
AriPerkkio: Hmmm maybe. This is similar as existing `markTasksAsSkipped`. I needed same logic here but not change finished tests:

https://github.com/vitest-dev/vitest/blob/1d49874985a5b00f0c2ce291b3edd45f607bf534/packages/runner/src/run.ts#L806-L815

So is `markTasksAsSkipped` also buggy?

sheremet-va: Good question, I am not sure. Is there a test for this?

AriPerkkio: > Shouldn't we trigger both test-prepare and test-finished if the test was skipped? 

I think I remember us discussing this during reporter API rework, and deciding that those hooks need to be called for skipped tests. I'll check this and add test cases too.

AriPerkkio: Ok - `updateTask` here actually causes duplicate `onTestCase*` calls on reporter. Fixing 👀

packages/vitest/src/node/test-run.ts
```diff
@@ -232,6 +232,11 @@ export class TestRun {
       return
     }
 
+    if (event === 'test-cancel' && entity.type === 'test') {
+      // This is used to just update state of the task
+      return
+    }
```
AriPerkkio, line 238: This event is used to just update the task state on main thread side. The actual reporting happens automatically in

https://github.com/vitest-dev/vitest/blob/8880c907a4a279ef7144a64a1cc5f84f14a4574e/packages/vitest/src/node/test-run.ts#L213-L221

test/cli/test/cancel-run.test.ts
```diff
@@ -27,17 +31,130 @@ test('can force cancel a run', async () => {
...
+      onTestCaseReady(testCase) {
+        onTestCaseHooks.push(`onTestCaseReady ${testCase.name}`)
+      },
+      onTestCaseResult(testCase) {
+        onTestCaseHooks.push(`onTestCaseResult ${testCase.name}`)
+        onTestCaseHooks.push('') // padding
+      },
+      onTestCaseAnnotate: (_, annotation) => {
+        if (annotation.message === 'Running long test, do the cancelling now!') {
+          onSlowTestRunning.resolve()
```
sheremet-va, line 70: Can't you just cancel here directly? You can get `vitest` instance in `onInit`

I think I wrote a similar test somewhere 🤔

AriPerkkio: Probably yeah, but calling the return value from `createVitest` resembles actual user's use case.

sheremet-va (approved):

PR #9745, opened 2026-02-25
fix(browser)!: iframe scale

### LINKED ISSUE #9124: page.screenshot is extremely low resolution with a large viewport
### Describe the bug

Running browser tests with playwright and `viewport: { width: 1280, height: 800 }`

This is unusable, the page is taller than 800px so everything else gets cropped off:
```
await page.viewport(600, 800)
await page.screenshot()
```

This captures the full page but #6512 scales it down to only 42px wide when the scrollHeight is 9300:
```
await page.viewport(600, document.body.scrollHeight)
await page.screenshot()
```

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-tszdoqkx

### System Info

```shell
System:
    OS: Linux 6.17 Ubuntu 25.10 25.10 (Questing Quokka)
    CPU: (16) x64 AMD Ryzen 7 5700X3D 8-Core Processor
    Memory: 9.52 GB / 31.27 GB
    Container: Yes
    Shell: 4.0.6 - /usr/bin/fish
  Binaries:
    Node: 24.11.1 - /run/user/1000/fnm_multishells/437459_1764264127836/bin/node
    npm: 11.6.2 - /run/user/1000/fnm_multishells/437459_1764264127836/bin/npm
    pnpm: 10.23.0 - /run/user/1000/fnm_multishells/437459_1764264127836/bin/pnpm
  Browsers:
    Chrome: 142.0.7444.175
    Firefox: 145.0.1
    Firefox Developer Edition: 145.0.1
```

### Used Package Manager

pnpm

### Validations

### LINKED ISSUE #9363: Screenshot dimensions

I'm using the Playwright provider and trying to use screenshots for rendering tests with the following `vite.config.js` test config:
```js
{
  browser: {
    ui: true,
    provider: playwright({
      launchOptions: {
        args: ['--enable-gpu', '--force-device-scale-factor=1'],
      },
    }),
    enabled: true,
    instances: [{browser: 'chromium'}],
    viewport: {width: 512, height: 512},
  },
}
```

In my tests, I capture a screenshot with `await expect(page.elementLocator(target)).toMatchScreenshot();` and the `target` element has CSS dimensions of 512x512. The problem is that the vitest iframe with the Browser UI has an arbitrary width depending on the `--tester-transform` scale that is applied, and the screenshot dimensions are based on this scaled element's dimensions instead of dimensions that I control.

I can get tests to pass by running them in headless mode or by setting `browser.ui` to `false`. Is there an existing way to configure things so that screenshots captured using headed tests with `browser.ui` set to `true` will have a fixed or known dimension?

### Suggested solution

I'm not sure what a good solution would be. Maybe another option to disable scaling of the `tester-ui` element?
```js
{
  viewport: {width: 512, height},
  scaleViewport: false, // this would make it so tester-ui doesn't have a scale transform applied
}
```

### Alternative

_No response_

### Additional context

```
# npm ls vitest
├─┬ @vitest/browser-playwright@4.0.16
│ ├─┬ @vitest/browser@4.0.16
│ │ └── vitest@4.0.16 deduped
│ └── vitest@4.0.16 deduped
└── vitest@4.0.16
```

### Validations

macarie:
### Description

Highly WIP and needs a lot of testing, but so far seems to work with Playwright with `chromium`, `firefox`, and `webkit`.

Have to look into WebdriverIO (🫣)

Fixes #9124
Fixes #9363

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/browser-playwright/src/commands/screenshot.ts
+++ b/packages/browser-playwright/src/commands/screenshot.ts
@@ -9,6 +9,16 @@ interface ScreenshotCommandOptions extends Omit<ScreenshotOptions, 'element' | '
   element?: string
   mask?: readonly string[]
 }
+
+const SCREENSHOT_STYLES = /* css */`
+  iframe {
+    position: absolute !important;
+    inset: 0  !important;
+    z-index: ${Number.MAX_SAFE_INTEGER} !important;
+    transform: none !important;
+  }
+`
+
 /**
  * Takes a screenshot using the provided browser context and returns a buffer and the expected screenshot path.
  *
@@ -43,6 +53,11 @@ export async function takeScreenshot(
   }
 
   const mask = options.mask?.map(selector => getDescribedLocator(context, selector))
+  const style = context.project.config.browser.ui
+    ? options.style === undefined
+      ? SCREENSHOT_STYLES
+      : SCREENSHOT_STYLES + options.style
+    : options.style
 
   if (options.element) {
     const { element: selector, ...config } = options
@@ -51,6 +66,7 @@ export async function takeScreenshot(
       ...config,
       mask,
       path: savePath,
+      style,
     })
     return { buffer, path }
   }
@@ -59,6 +75,7 @@ export async function takeScreenshot(
     ...options,
     mask,
     path: savePath,
+    style,
   })
   return { buffer, path }
 }
--- a/packages/browser-playwright/src/commands/screenshot.ts
+++ b/packages/browser-playwright/src/commands/screenshot.ts
@@ -13,7 +13,7 @@ interface ScreenshotCommandOptions extends Omit<ScreenshotOptions, 'element' | '
 const SCREENSHOT_STYLES = /* css */`
   iframe {
     position: absolute !important;
-    inset: 0  !important;
+    inset: 0 !important;
     z-index: ${Number.MAX_SAFE_INTEGER} !important;
     transform: none !important;
   }
--- a/packages/browser-playwright/src/commands/screenshot.ts
+++ b/packages/browser-playwright/src/commands/screenshot.ts
@@ -11,7 +11,7 @@ interface ScreenshotCommandOptions extends Omit<ScreenshotOptions, 'element' | '
 }
 
 const SCREENSHOT_STYLES = /* css */`
-  iframe {
+  iframe[data-vitest] {
     position: absolute !important;
     inset: 0 !important;
     z-index: ${Number.MAX_SAFE_INTEGER} !important;
--- a/packages/browser-playwright/src/commands/screenshot.ts
+++ b/packages/browser-playwright/src/commands/screenshot.ts
@@ -11,7 +11,7 @@ interface ScreenshotCommandOptions extends Omit<ScreenshotOptions, 'element' | '
 }
 
 const SCREENSHOT_STYLES = /* css */`
-  iframe[data-vitest] {
+  iframe[data-vitest="true"] {
     position: absolute !important;
     inset: 0 !important;
     z-index: ${Number.MAX_SAFE_INTEGER} !important;
--- a/packages/browser-playwright/src/commands/viewport.ts
+++ b/packages/browser-playwright/src/commands/viewport.ts
@@ -1,11 +1,10 @@
-import type { PlaywrightBrowserProvider } from '../playwright'
 import type { UserEventCommand } from './utils'
 
 export const viewport: UserEventCommand<(options: {
   width: number
   height: number
 }) => void> = async (context, options) => {
-  const { provider, sessionId } = context
+  const { page } = context
 
-  await (provider as PlaywrightBrowserProvider).getPage(sessionId).setViewportSize(options)
+  await page.setViewportSize(options)
 }
--- a/packages/browser-playwright/src/commands/viewport.ts
+++ b/packages/browser-playwright/src/commands/viewport.ts
@@ -1,0 +1,11 @@
+import type { PlaywrightBrowserProvider } from '../playwright'
+import type { UserEventCommand } from './utils'
+
+export const viewport: UserEventCommand<(options: {
+  width: number
+  height: number
+}) => void> = async (context, options) => {
+  const { provider, sessionId } = context
+
+  await (provider as PlaywrightBrowserProvider).getPage(sessionId).setViewportSize(options)
+}
--- a/packages/browser-playwright/src/commands/viewport.ts
+++ b/packages/browser-playwright/src/commands/viewport.ts
@@ -4,7 +4,5 @@ export const viewport: UserEventCommand<(options: {
   width: number
   height: number
 }) => void> = async (context, options) => {
-  const { page } = context
-
-  await page.setViewportSize(options)
+  await context.page.setViewportSize(options)
 }
--- a/packages/browser-playwright/src/locators.ts
+++ b/packages/browser-playwright/src/locators.ts
@@ -155,8 +155,17 @@ function processClickOptions(options?: UserEventClickOptions) {
   return options
 }
 
+let scale = getIframeScale()
+
+const iframeContainer = window.frameElement?.parentElement
+
+if (iframeContainer) {
+  new ResizeObserver(() => {
+    scale = getIframeScale()
+  }).observe(iframeContainer)
+}
+
 function processPlaywrightPosition(position: { x: number; y: number }) {
-  const scale = getIframeScale()
   if (position.x != null) {
     position.x *= scale
   }
--- a/packages/browser-playwright/src/locators.ts
+++ b/packages/browser-playwright/src/locators.ts
@@ -155,17 +155,8 @@ function processClickOptions(options?: UserEventClickOptions) {
   return options
 }
 
-let scale = getIframeScale()
-
-const iframeContainer = window.frameElement?.parentElement
-
-if (iframeContainer) {
-  new ResizeObserver(() => {
-    scale = getIframeScale()
-  }).observe(iframeContainer)
-}
-
 function processPlaywrightPosition(position: { x: number; y: number }) {
+  const scale = getIframeScale()
   if (position.x != null) {
     position.x *= scale
   }
--- a/packages/browser-webdriverio/src/commands/screenshot.ts
+++ b/packages/browser-webdriverio/src/commands/screenshot.ts
@@ -66,7 +66,7 @@ export async function takeScreenshot(
 
       stylesheet.id = 'vitest:screenshot-stylesheet'
       stylesheet.textContent = /* css */`
-        iframe[data-vitest] {
+        iframe[data-vitest="true"] {
           position: absolute !important;
           inset: 0 !important;
           z-index: ${Number.MAX_SAFE_INTEGER} !important;
--- a/packages/browser-webdriverio/src/commands/screenshot.ts
+++ b/packages/browser-webdriverio/src/commands/screenshot.ts
@@ -66,7 +66,7 @@ export async function takeScreenshot(
 
       stylesheet.id = 'vitest:screenshot-stylesheet'
       stylesheet.textContent = /* css */`
-        iframe {
+        iframe[data-vitest] {
           position: absolute !important;
           inset: 0 !important;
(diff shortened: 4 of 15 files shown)


sheremet-va:
Any news here? 👀

macarie:
From my testing seems to be working fairly well with Playwright, but I've had no time to implement the WebdriverIO version yet. Hopefully this week 🤞🏼

packages/browser/src/client/orchestrator.ts
```diff
@@ -312,16 +310,37 @@ export class IframeOrchestrator {
...
+      // Firefox does not support typed arithmetic (divisions between typed values): https://bugzilla.mozilla.org/show_bug.cgi?id=1264520
+      if (config.browser.name === 'firefox') {
+        // Firefox cannot resolve relative units like `cqw` directly inside `atan2()`
+        // Storing it in a CSS variable first forces Firefox to resolve `100cqw` to an absolute pixel value
+        iframe.style.setProperty('--container-width', '100cqw')
+        // `tan(atan2(a, b))` produces a unit-less `a / b` ratio:
+        //  - `atan2()` accepts two lengths and returns an `<angle>`
+        //  - `tan()` converts it back to a unit-less `<number>`
+        iframe.style.setProperty('transform', 'scale(min(1, tan(atan2(var(--container-width), var(--viewport-width)))))')
+      }
```
macarie: todo: merge in an if/else block to set the properties only once with Firefox

test/browser/specs/to-match-screenshot.test.ts
```diff
@@ -178,39 +194,76 @@ describe.runIf(provider.name === 'playwright')('--watch', () => {
...
+      vitest.resetOutput()
+      await vitest.waitForStdout(`Test Files  ${instances.length} passed`, 20_000)
+
+      if (!process.env.CI) {
+        // switch to non-headless mode
+        fs.editFile('vitest.config.js', content => content.replace('headless: true,', 'headless: false,'))
+
+        vitest.resetOutput()
+        await vitest.waitForStdout(`Test Files  ${instances.length} passed`, 20_000)
+      }
```
macarie: Not sure about this part running locally either, can be kinda annoying losing window focus because a browser instance opens for this test 👀

packages/ui/client/components/BrowserIframe.vue
```diff
@@ -118,40 +76,29 @@ const marginLeft = computed(() => {
       />
       <span class="pointer-events-none" text-sm>
         {{ viewport[0] }}x{{ viewport[1] }}px
-        <span v-if="scale < 1">({{ (scale * 100).toFixed(0) }}%)</span>
```
macarie, line 121: I think I can add this back now if necessary.

macarie:
I moved the scaling part to be CSS-only, which allowed removing some JS. I think resizing the browser window and the panels now feels more fluid and less janky, especially on old(_er_) hardware.

When taking screenshots a bit of CSS is injected to remove the scaling from the iframe and making it _escape_ the pane. This should fix the "low resolution" issues, removing the scaling also leads to not having to deal with different sub pixel sizes and GPU rendering differences.

In headless mode (with no UI), Playwright now uses the set viewport, same as WebdriverIO, this should allow capturing screenshots larger than the default viewport size. In UI mode this is not done to prevent having flashing windows (more than we already have), so it's up to the user to make sure the rendered elements fit in the window.

packages/browser-playwright/src/commands/screenshot.ts
```diff
@@ -9,6 +9,16 @@ interface ScreenshotCommandOptions extends Omit<ScreenshotOptions, 'element' | '
...
   mask?: readonly string[]
 }
+
+const SCREENSHOT_STYLES = /* css */`
+  iframe {
+    position: absolute !important;
+    inset: 0  !important;
+    z-index: ${Number.MAX_SAFE_INTEGER} !important;
+    transform: none !important;
+  }
```
macarie: We might have to `opacity: 0` everything else. When the iframe doesn't have a background, the UI peeks through.

packages/browser-playwright/src/locators.ts
```diff
@@ -155,8 +155,17 @@ function processClickOptions(options?: UserEventClickOptions) {
...
 }
 
+let scale = getIframeScale()
+
+const iframeContainer = window.frameElement?.parentElement
+
+if (iframeContainer) {
+  new ResizeObserver(() => {
+    scale = getIframeScale()
+  }).observe(iframeContainer)
```
macarie: These `ResizeObserver`s might not be necessary as this gets executed every time a test runs, but it might be useful in long-running tests if the user resizes the panel... Maybe it should only be enabled when UI is `true`?

sheremet-va: Why can't we get the scale like we did before? When action needs it, it resolves the scale

macarie: Thought it might be better if it's cached locally rather than getting the computed styles each time, but position clicking should also be rare enough (I'd hope at least) that this might not even matter. Would you like me to keep this as it was previously?

sheremet-va: Yeah, let’s just keep the previous implementation, it takes nanoseconds, there is no reason to cache it

packages/browser-playwright/src/commands/viewport.ts
```diff
@@ -0,0 +1,11 @@
+import type { PlaywrightBrowserProvider } from '../playwright'
+import type { UserEventCommand } from './utils'
+
+export const viewport: UserEventCommand<(options: {
+  width: number
+  height: number
+}) => void> = async (context, options) => {
+  const { provider, sessionId } = context
+
+  await (provider as PlaywrightBrowserProvider).getPage(sessionId).setViewportSize(options)
```
sheremet-va: `page` is already available in the context

packages/browser-playwright/src/commands/screenshot.ts
```diff
@@ -9,6 +9,16 @@ interface ScreenshotCommandOptions extends Omit<ScreenshotOptions, 'element' | '
   element?: string
   mask?: readonly string[]
 }
+
+const SCREENSHOT_STYLES = /* css */`
+  iframe {
```
sheremet-va: Shouldn't this be `[data-vitest=true]`?

packages/ui/client/components/BrowserIframe.vue
sheremet-va: Why would we remove it? The window is still scaled

test/browser/specs/to-match-screenshot.test.ts
```diff
@@ -25,24 +26,23 @@ test('${testName}', async ({ expect }) => {
...
     'vitest.config.js': `
-      import { playwright } from '@vitest/browser-playwright'
+      import { ${provider.name} } from '@vitest/browser-${provider.name}'
       export default {
         test: {
           browser: {
             enabled: true,
             screenshotFailures: false,
-            provider: playwright(),
+            provider: ${provider.name}(),
```
sheremet-va: `runInlineTests` can inject these automatically if you just pass down an object instead of a file content string (so passing down `provider` in the test will correctly specify it in the config)

packages/ui/client/components/BrowserIframe.vue
macarie: I removed it initially when I wasn't sure about what approach to go with. I'll add it back 👍🏼

packages/browser-playwright/src/commands/screenshot.ts
macarie: Yes, should be more specific. Do you prefer `data-vitest` or its `name` (or both)?

packages/browser-playwright/src/commands/screenshot.ts
macarie: This can be fixed by setting a background color. I'm not sure if it should change with the theme or if it should stay white tho 🤔

packages/browser-playwright/src/commands/screenshot.ts
sheremet-va: I think we use data-vitest everywhere in locators), but I’m not sure

test/browser/specs/to-match-screenshot.test.ts
macarie: Removed this part.

packages/browser-playwright/src/commands/screenshot.ts
macarie: Used `data-vitest`, looks like it's the one that's used in some other places too.

packages/browser-webdriverio/src/commands/screenshot.ts
```diff
@@ -59,12 +59,55 @@ export async function takeScreenshot(
   // webdriverio expects the path to contain the extension and only works with PNG files
   const savePathWithExtension = savePath.endsWith('.png') ? savePath : `${savePath}.png`
 
+  if (context.project.config.browser.ui) {
+    // refactor with `using` and `Symbol.asyncDispose` when support for Node <24 is dropped
+    await context.browser.execute(() => {
+      const document = window.frameElement?.ownerDocument
+
+      if (!document) {
+        throw new Error('cannot access parent document')
```
macarie: todo: improve this error message (and the one beneath)

packages/browser/src/client/tester/tester-utils.ts
```diff
@@ -224,15 +224,15 @@ export function processTimeoutOptions<T extends { timeout?: number }>(options_:
...
-  if (Number.isNaN(scale)) {
-    throw new TypeError(`Cannot parse scale value from Tester element (${scaleAttribute}). This is a bug in Vitest. Please, open a new issue with reproduction.`)
+  const iframe = window.frameElement
+
+  if (!iframe) {
+    throw new Error(`Cannot find iframe element. This is a bug in Vitest. Please, open a new issue with reproduction.`)
   }
+
+  // we can safely use `a` as both vertical and horizontal scale are the same
+  const scale = new DOMMatrix(getComputedStyle(iframe).transform).a
```
sheremet-va, line 235: Can you add an explanation (as a comment) of what is going on here? I've never seen this usage before

test/browser/specs/to-match-screenshot.test.ts
```diff
@@ -178,39 +199,68 @@ describe.runIf(provider.name === 'playwright')('--watch', () => {
...
 
         fs.editFile(testFilename, content => content.replace(bgColor, '#000'))
 
         vitest.resetOutput()
-        await vitest.waitForStdout('Test Files  1 passed')
+        await vitest.waitForStdout(`Test Files  ${instances.length} passed`)
 
-        expect(vitest.stdout).toContain('✓ |chromium| basic.test.ts > screenshot-snapshot')
+        for (const instance of instances) {
+          expect(`✓ |${instance.browser}| basic.test.ts > screenshot-snapshot`)
```
sheremet-va: This doesn't test anything?

sheremet-va: I wonder if we should throw an error if expect was called without a matcher

packages/browser/src/client/tester/tester-utils.ts
macarie: I've updated the comment, does this version explain it better? 🤔

test/browser/specs/to-match-screenshot.test.ts
macarie: Fixed it, messed up with the multi-cursor (again) 😅

macarie: > I wonder if we should throw an error if expect was called without a matcher

That would be nice, or maybe a lint issue 👀

sheremet-va (commented):
There are a few advanced features here. Does it change our compatibility table? https://vitest.dev/guide/browser/#browser-compatibility

macarie:
Yeah, looks like typed-arithmetic in `calc` has been supported by all the browsers we support since September 2025 (Chromium 140, Safari 26, Firefox not yet). Using the `tan`/`atan2` version everywhere would take the support date back to 2023 (Chromium 111, Firefox 108, Safari 15.4).

I guess this would be a breaking change anyway then 🤔

sheremet-va (approved):

PR #9748, opened 2026-02-26
fix: fix re-mocking virtual module

### LINKED ISSUE #9742: Virtual module mocking issue in Vitest 4
### Describe the bug

_This discussion is a follow-up to [this discussion](https://github.com/vitest-dev/vitest/discussions/9734) that I initially opened before this issue.__

In a test file, I have different `describe` blocks where in each of them, I mock a virtual module, e.g. using `vi.doMock()` and mock a function from that virtual module making it return different values.

The behavior seems to be different between Vitest 3 and Vitest 4:

- Vitest 3: everything seems to work as expected, and calling some code that uses that virtual module returns the mocked value.
- Vitest 4: the first mocked value seems to be returned for all tests.

### Reproduction

I've setup a minimal repro available here: https://github.com/HiDeoo/vitest-4-mock-repro

1. Clone the repository
1. Run `pnpm install`
1. Run `pnpm test`

Some tests that are checking that the expected returned mocked value is correct will fail.

A few extra notes:

- Running `pnpm add vitest@3` and `pnpm test` will make the test pass
- I tried swapping the order of `vi.resetModules()` and `vi.doMock()` as [suggested](https://github.com/vitest-dev/vitest/discussions/9734#discussioncomment-15917711) but this didn't change the behavior

### System Info

```shell
System:
    OS: macOS 26.3
    CPU: (8) arm64 Apple M1
    Memory: 197.14 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.14.0 - /Users/hideoo/.local/state/fnm_multishells/36543_1772008897917/bin/node
    npm: 10.9.2 - /Users/hideoo/.local/state/fnm_multishells/36543_1772008897917/bin/npm
    pnpm: 9.15.9 - /Users/hideoo/.local/state/fnm_multishells/36543_1772008897917/bin/pnpm
    bun: 1.3.9 - /opt/homebrew/bin/bun
  Browsers:
    Firefox: 147.0.4
    Safari: 26.3
  npmPackages:
    vitest: ^4.0.18 => 4.0.18
```

### Used Package Manager

pnpm

### Validations

hi-ogawa:
### Description

- Closes https://github.com/vitest-dev/vitest/issues/9742

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/runtime/moduleRunner/startVitestModuleRunner.ts
+++ b/packages/vitest/src/runtime/moduleRunner/startVitestModuleRunner.ts
@@ -108,8 +108,8 @@ export function startVitestModuleRunner(options: ContextModuleRunnerOptions): Vi
             return {
               code: '',
               file: null,
-              id,
-              url: id,
+              id: resolvedMock.id,
+              url: resolvedMock.url,
               invalidate: false,
               mockedModule: resolvedMock,
             }
@@ -137,6 +137,9 @@ export function startVitestModuleRunner(options: ContextModuleRunnerOptions): Vi
             options,
             otelCarrier,
           )
+          // if (id.includes('virtual-module-direct')) {
+          //   console.log("[fetchModule]", result)
+          // }
           if ('cached' in result) {
             const code = readFileSync(result.tmp, 'utf-8')
             return { code, ...result }
--- a/packages/vitest/src/runtime/moduleRunner/startVitestModuleRunner.ts
+++ b/packages/vitest/src/runtime/moduleRunner/startVitestModuleRunner.ts
@@ -137,9 +137,6 @@ export function startVitestModuleRunner(options: ContextModuleRunnerOptions): Vi
             options,
             otelCarrier,
           )
-          // if (id.includes('virtual-module-direct')) {
-          //   console.log("[fetchModule]", result)
-          // }
           if ('cached' in result) {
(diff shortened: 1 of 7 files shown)


packages/vitest/src/runtime/moduleRunner/startVitestModuleRunner.ts
```diff
@@ -108,8 +108,8 @@ export function startVitestModuleRunner(options: ContextModuleRunnerOptions): Vi
             return {
               code: '',
               file: null,
-              id,
-              url: id,
+              id: resolvedMock.id,
+              url: resolvedMock.url,
```
hi-ogawa, line 112: For the case of virtual module, here we previously had `id = /@id/virtual-module` (aka "url") as module runner `fetchModule` result, which then leaded to `evaluatedModuleNode.id = mock:/@id/virtual-module` during later mock interception.

However, when invalidating mock for next `doMock` calls, we use `resolveId` of `virtual-module` and it tries to invalidate `mock:virtual-module`. This inconsistency makes invalidation to miss.

What this `resolvedMock` case should do instead is to return `id` as `resolvedId` of mocked module, which is already available as `resolvedMock.id`, so that is the solution here.

sheremet-va (approved):

PR #9752, opened 2026-02-27
fix: issue #9696

### LINKED ISSUE #9696: Concurrency controls for projects

With over 20 projects using browser mode, it seems our CI is running into some [resource exhaustion issues](https://app.circleci.com/pipelines/github/mui/mui-x/119335/workflows/44038cc6-937e-4733-a908-48f5ae7b0c3d/jobs/752068/parallel-runs/0/steps/0-106). It looks like vitest starts a dev server for each project and a browser tab for each project. We already reduce concurrency per project, but it does seem to start all projects simultaneously. Causing a lot more concurrency than we intend. We already optimized by using a single browser instance.

### Suggested solution

Hard to say with my limited knowledge of vitest architecture. But if we could have a single pool across all projects and control concurrency globally, that would be ideal. Otherwise a way to specify concurrency for running the projects. e.g. only 4 projects may run simultaneously.

### Alternative

Wrapping vitest with an orchestrator that runs only X projects simultaneously, but it complicates things around test output, coverage, projects filtering...

### Additional context

_No response_

### Validations

jorgitin02:
Fixes #9696


sheremet-va:
1. Not following the template
2. The usage of AI is not disclosed
3. The solution is completely incorrect, there is nowhere to start reviewing even

PR #9756, opened 2026-02-27
docs(reporters): correct gitHub-actions reporter documentation

literalpie:
### Description

Based on my experience testing this a bit, it seems like this reporter is only included by default when no reporters are set at all, _not when the reporter called "default" is set_. This documentation states otherwise. 

To test, I did `GITHUB_ACTIONS=true pnpm test` in a project with a failing test. When the config has `reporters: ['default']`, the GHA stuff doesn't get logged. When the config is totally empty or the env var check is manually added, the GHA stuff does get logged.

It's possible that this is a bug and the intention is to match the documentation's behavior? 

(Documentation only change)

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/config/reporters.md
+++ b/docs/config/reporters.md
@@ -14,7 +14,7 @@ interface UserConfig {
 type ConfigReporter = string | Reporter | [string, object?]
 ```
 
-- **Default:** [`'default'`](/guide/reporters#default-reporter)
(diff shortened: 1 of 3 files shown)


hi-ogawa (approved):
Good catch!
When writing this, that's what I actually meant, but my English was just broken 🫠 

I pushed a change to mention this default behavior in other places. 

btw the logic is just simple here

https://github.com/vitest-dev/vitest/blob/46ca8c0edb50f1661200c811dd7987e33bb30626/packages/vitest/src/node/config/resolveConfig.ts#L731-L738

PR #9772, opened 2026-03-03
fix: fix `vi.importActual()` for virtual modules

### LINKED ISSUE #9771: `vi.importActual()` issue with virtual modules in Vitest 4
### Describe the bug

In a test file, I mock a virtual module, e.g. using `vi.mock()` and try to import the original module using either [`vi.importActual()`](https://vitest.dev/api/vi.html#vi-importactual) or the [`importOriginal()`](https://vitest.dev/api/vi.html#vi-mock) helper.

```ts
vi.mock("virtual:test", async (importOriginal) => {
  const mod = await vi.importActual<"virtual:test">("virtual:test");
  console.log("🚨 mod with vi.importActual:", mod);

  const actual = await importOriginal();
  console.log("🚨 actual with importOriginal:", actual);

  // Not doing anything here for the sake of the repro, just returning the original module.
  return mod;
});
```

The behavior seems to be different between Vitest 3 and Vitest 4:

**Vitest 3**: everything seems to work as expected, I can import the original module and get the expected value.
**Vitest 4**: both `vi.importActual()` and the `importOriginal()` helper return what seems to be an empty module (e.g. when logging the result, I get `[Object: null prototype] [Module] {}`)

### Reproduction

I've setup a minimal repro available here: https://github.com/HiDeoo/vitest-4-mock-virtual-module-importactual-repro

1. Clone the repository
1. Run pnpm install
1. Run pnpm test

Some tests that are checking the result of calling a function using the virtual module will fail.

A few extra notes:

- Running `pnpm add vitest@3` and `pnpm test` will make the test pass
- I've carefully read the Vitest 4 migration guide, and specifically the section about ["Changes to Mocking"](https://vitest.dev/guide/migration.html#changes-to-mocking), but unless I'm missing something, I don't think any of the listed changes explain this behavior.

### System Info

```shell
System:
    OS: macOS 26.3
    CPU: (8) arm64 Apple M1
    Memory: 96.19 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.14.0 - /Users/hideoo/.local/state/fnm_multishells/42034_1772463011409/bin/node
    npm: 10.9.2 - /Users/hideoo/.local/state/fnm_multishells/42034_1772463011409/bin/npm
    pnpm: 9.15.9 - /Users/hideoo/.local/state/fnm_multishells/42034_1772463011409/bin/pnpm
    bun: 1.3.10 - /opt/homebrew/bin/bun
  Browsers:
    Chrome: 145.0.7632.117
    Firefox: 148.0
    Safari: 26.3
  npmPackages:
    vitest: 4.1.0-beta.5 => 4.1.0-beta.5
```

### Used Package Manager

pnpm

### Validations

hi-ogawa:
### Description

- Closes https://github.com/vitest-dev/vitest/issues/9771

This PR changes `importActual` to use `?_vitest_original` to fetch original module code. There is already a similar technique on browser mode. Additionally for this to actually work for virtual module, it's necessary to strip `?_vitest_original` before it gets into `load` hook of virtual module. For webdriverio though, this is actually not possible because it relies on plugin pipeline `load(originalId)` to implement mocking. For others, this is possible since "original id" handling happens outside of vite plugin pipeline.

### Tests

### Documentation

### Changesets


packages/vitest/src/runtime/moduleRunner/startVitestModuleRunner.ts
```diff
@@ -103,15 +111,17 @@ export function startVitestModuleRunner(options: ContextModuleRunnerOptions): Vi
...
+            const resolvedMock = moduleRunner.mocker.getDependencyMock(rawId)
+            if (resolvedMock?.type === 'manual' || resolvedMock?.type === 'redirect') {
+              return {
+                code: '',
+                file: null,
+                id: resolvedMock.id,
+                url: resolvedMock.url,
+                invalidate: false,
+                mockedModule: resolvedMock,
+              }
```
hi-ogawa, line 124: The issue was that previously `rawId = "virtual:module"` would get short-circuit here for both initial mocked `import "virtual:module"` and following `importOriginal` calls, which means the real code never got read by module runner.

This PR adds `?_vitest_original` query to `importOriginal`'s `fetchModule` so that it can return the real module code.

test/cli/test/mocking.test.ts
```diff
@@ -1,8 +1,16 @@
...
-import { expect, test } from 'vitest'
+import { playwright } from '@vitest/browser-playwright'
+import { webdriverio } from '@vitest/browser-webdriverio'
+import { afterAll, expect, test } from 'vitest'
 import { rolldownVersion } from 'vitest/node'
 import { runInlineTests, runVitest } from '../../test-utils'
 
+// webdriver@9 sets dns.setDefaultResultOrder("ipv4first") on import,
+// which makes Vite resolve localhost to 127.0.0.1 and breaks other tests asserting "localhost"
+afterAll(() => setDefaultResultOrder('verbatim'))
```
hi-ogawa, line 12: This is needed to avoid this silly `localhost` vs `127.0.0.1` errors https://github.com/vitest-dev/vitest/actions/runs/22614377230/job/65523514606#step:8:911

```
   FAIL  test/browser-multiple.test.ts:7:1 > automatically assigns the port
  AssertionError: expected [ 'http://127.0.0.1:63315/', 'http://127.0.0.1:63316/' ] to include 'http://localhost:63315/'
   ❯ test/browser-multiple.test.ts:31:16
       29|   expect(spy).not.toHaveBeenCalled()
       30|   expect(stderr).not.toContain('is in use, trying another one...')
       31|   expect(urls).toContain('http://localhost:63315/')
         |                ^
       32|   expect(urls).toContain('http://localhost:63316/')
       33| })
```

packages/browser/src/node/plugin.ts
```diff
@@ -49,6 +49,20 @@ export default (parentServer: ParentBrowserProject, base = '/'): Plugin[] => {
...
+            req.url?.includes('_vitest_original')
+            && parentServer.project.config.browser.provider?.name === 'playwright'
+          ) {
+            req.url = req.url
+              .replace(/[?&]_vitest_original(?=[&#]|$)/, '')
+              .replace(/[?&]ext\b[^&#]*/, '')
+              .replace(/\?$/, '')
+          }
+          next()
+        })
```
hi-ogawa, line 65: Ideally this should live in playwright provider, but don't think that's possible now, so here it is.

hi-ogawa: Actually instead of middleware level rewriting, it should be also possible to just strip `?_vitest_original` from `resolveId` like

```js
resolveId(source) {
  if (source.endsWith("?_vitest_original")) { 
    return source.slice(...)
  }
}
```

This is because the `source` comes here is already the resolved id + `?_vitest_original` such e.g. `"\0virtual:module?_vitest_original"`.

In fact, I used a similar idea in RSC plugin to "redirect" already resolved id back as something importable https://github.com/vitejs/vite-plugin-react/blob/main/packages/plugin-rsc/src/plugins/resolved-id-proxy.ts.

sheremet-va (approved):

PR #9775, opened 2026-03-03
chore: add blog post for Vitest 4.1

sheremet-va:
https://deploy-preview-9775--vitest-dev.netlify.app/blog/vitest-4-1.html

PR #9779, opened 2026-03-04
feat(reporters): add `agent` reporter to reduce ai agent token usage

cpojer:
### Description

Test run output tends to be "token inefficient". When running Vitest it prints every passed test and their logs. This PR is aimed at reducing token usage during test runs when Vitest is run by a coding agent by omitting passed tests completely from the streamed output. Instead, it only prints failing tests and the final test run summary. This should make running Vitest much more token efficient. 

* `AI_AGENT=1 pnpm vitest` → only prints failing tests and the final summary
* `pnpm vitest` → no output change

### Tests

### Changesets

### DIFF
--- a/packages/vitest/src/node/reporters/agent.ts
+++ b/packages/vitest/src/node/reporters/agent.ts
@@ -1,0 +1,31 @@
+import type { TestSpecification } from '../test-specification'
+import type { DefaultReporterOptions } from './default'
+import type { TestCase, TestModule, TestModuleState } from './reported-tasks'
+import { DefaultReporter } from './default'
+
+export class AgentReporter extends DefaultReporter {
+  renderSucceed = false
+
+  constructor(options: DefaultReporterOptions = {}) {
+    super({ silent: 'passed-only', ...options, summary: false })
+  }
+
+  onTestRunStart(specifications: ReadonlyArray<TestSpecification>): void {
+    super.onTestRunStart(specifications)
+    this.renderSucceed = false
+  }
+
+  protected printTestModule(testModule: TestModule): void {
+    if (testModule.state() !== 'failed') {
+      return
+    }
+    super.printTestModule(testModule)
+  }
+
+  protected printTestCase(moduleState: TestModuleState, test: TestCase): void {
+    const testResult = test.result()
+    if (testResult.state === 'failed') {
+      super.printTestCase(moduleState, test)
+    }
+  }
+}
--- a/packages/vitest/src/utils/env.ts
+++ b/packages/vitest/src/utils/env.ts
@@ -12,4 +12,4 @@ export const isDeno: boolean
 export const isWindows: boolean = (isNode || isDeno) && process.platform === 'win32'
 export const isBrowser: boolean = typeof window !== 'undefined'
 export const isTTY: boolean = ((isNode || isDeno) && process.stdout?.isTTY && !isCI)
-export { isCI, provider as stdProvider } from 'std-env'
+export { isAgent, isCI, provider as stdProvider } from 'std-env'
--- a/docs/config/reporters.md
+++ b/docs/config/reporters.md
@@ -42,6 +42,7 @@ Note that the [coverage](/guide/coverage) feature uses a different [`coverage.re
 - [`tap-flat`](/guide/reporters#tap-flat-reporter)
 - [`hanging-process`](/guide/reporters#hanging-process-reporter)
 - [`github-actions`](/guide/reporters#github-actions-reporter)
+- [`agent`](/guide/reporters#agent-reporter)
 - [`blob`](/guide/reporters#blob-reporter)
 
 ## Example
--- a/docs/guide/browser/visual-regression-testing.md
+++ b/docs/guide/browser/visual-regression-testing.md
@@ -606,7 +606,6 @@ screenshots should use the service.
 
 The cleanest approach is using [Test Projects](/guide/projects):
 
-
 ```ts [vitest.config.ts]
 import { env } from 'node:process'
 import { defineConfig } from 'vitest/config'
@@ -662,7 +661,6 @@ export default defineConfig({
   },
 })
 ```
-
 
 Follow the [official guide to create a Playwright Workspace](https://learn.microsoft.com/en-us/azure/app-testing/playwright-workspaces/quickstart-run-end-to-end-tests?tabs=playwrightcli&pivots=playwright-test-runner#create-a-workspace).
 
--- a/docs/guide/cli-generated.md
+++ b/docs/guide/cli-generated.md
@@ -102,7 +102,7 @@ Hide logs for skipped tests
 - **CLI:** `--reporter <name>`
 - **Config:** [reporters](/config/reporters)
 
-Specify reporters (default, blob, verbose, dot, json, tap, tap-flat, junit, tree, hanging-process, github-actions)
+Specify reporters (default, agent, blob, verbose, dot, json, tap, tap-flat, junit, tree, hanging-process, github-actions)
 
 ### outputFile
 
--- a/docs/guide/cli.md
+++ b/docs/guide/cli.md
@@ -237,4 +237,3 @@ Merges every blob report located in the specified folder (`.vitest-reports` by d
 ```sh
 vitest --merge-reports --reporter=junit
 ```
-
--- a/docs/guide/reporters.md
+++ b/docs/guide/reporters.md
@@ -98,6 +98,10 @@ This example will write separate JSON and XML reports as well as printing a verb
 
 By default (i.e. if no reporter is specified), Vitest will display summary of running tests and their status at the bottom. Once a suite passes, its status will be reported on top of the summary.
 
+::: tip
+When Vitest detects it is running inside an AI coding agent, the [`agent`](#agent-reporter) reporter is used instead to reduce output and minimize token usage. You can override this by explicitly configuring the [`reporters`](/config/reporters) option.
+:::
+
 You can disable the summary by configuring the reporter:
 
 :::code-group
@@ -636,6 +640,26 @@ export default defineConfig({
   },
 })
 ```
+
+### Agent Reporter
+
+Outputs a minimal report optimized for AI coding assistants and LLM-based workflows. Only failed tests and their error messages are displayed. Console logs from passing tests and the summary section are suppressed to reduce token usage.
+
+This reporter is automatically enabled when no `reporters` option is configured and Vitest detects it is running inside an AI coding agent. If you configure custom reporters, you can explicitly add `agent`:
+
+:::code-group
+```bash [CLI]
+npx vitest --reporter=agent
+```
+
+```ts [vitest.config.ts]
+export default defineConfig({
+  test: {
+    reporters: ['agent']
+  },
+})
+```
+:::
 
 ### Blob Reporter
 
--- a/packages/vitest/src/node/config/resolveConfig.ts
+++ b/packages/vitest/src/node/config/resolveConfig.ts
@@ -23,7 +23,7 @@ import {
   defaultPort,
 } from '../../constants'
 import { benchmarkConfigDefaults, configDefaults } from '../../defaults'
-import { isCI, stdProvider } from '../../utils/env'
+import { isAgent, isCI, stdProvider } from '../../utils/env'
 import { getWorkersCountByPercentage } from '../../utils/workers'
 import { BaseSequencer } from '../sequencers/BaseSequencer'
 import { RandomSequencer } from '../sequencers/RandomSequencer'
@@ -729,7 +729,7 @@ export function resolveConfig(
   }
 
   if (!resolved.reporters.length) {
-    resolved.reporters.push(['default', {}])
+    resolved.reporters.push([isAgent ? 'agent' : 'default', {}])
 
     // also enable github-actions reporter as a default
     if (process.env.GITHUB_ACTIONS === 'true') {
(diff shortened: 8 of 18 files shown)


cpojer:
cc @sheremet-va @hi-ogawa

packages/vitest/src/utils/env.ts
```diff
@@ -12,4 +12,25 @@ export const isDeno: boolean
 export const isWindows: boolean = (isNode || isDeno) && process.platform === 'win32'
 export const isBrowser: boolean = typeof window !== 'undefined'
 export const isTTY: boolean = ((isNode || isDeno) && process.stdout?.isTTY && !isCI)
+export function detectAgent(): boolean {
```
fengmk2: should we use https://npmx.dev/package/@vercel/detect-agent 😄 ?

cpojer: I told Claude to do that! It ran into problems during the test run of Vitest's own test suite because it couldn't control and disable it. It decided to inline the package, and it got things to work.

I thought it's just a few lines, so it's ok to inline it, but we can also switch to the package. Just let me know what you prefer.

fengmk2: keep inline it 👍

sheremet-va: `std-env` provides `isAgent`, we can just use that (the package is already used for isCI check) - https://npmx.dev/package/std-env#user-content-agent-detection

cpojer: Switched to `std-env` v4.

packages/vitest/src/node/cli/cac.ts
```diff
@@ -300,6 +300,10 @@ function normalizeCliOptions(cliFilters: string[], argv: CliOptions): CliOptions
 
 async function start(mode: VitestRunMode, cliFilters: string[], options: CliOptions): Promise<void> {
   try {
+    const { isAgent } = await import('../../utils/env')
+    if (isAgent) {
+      options.env = { AI_AGENT: '1', ...options.env }
```
sheremet-va: I don't think this is used anywhere now

packages/vitest/src/node/reporters/utils.ts
```diff
@@ -35,12 +36,14 @@ async function loadCustomReporterModule<C extends Reporter>(
 function createReporters(
   reporterReferences: ResolvedConfig['reporters'],
   ctx: Vitest,
-): Promise<Array<Reporter | DefaultReporter | BlobReporter | DotReporter | JsonReporter | TapReporter | JUnitReporter | HangingProcessReporter | GithubActionsReporter>> {
+): Promise<Array<Reporter | DefaultReporter | AgentReporter | BlobReporter | DotReporter | JsonReporter | TapReporter | JUnitReporter | HangingProcessReporter | GithubActionsReporter>> {
   const runner = ctx.runner
+  const isAgent = !!ctx.config.env?.AI_AGENT
```
sheremet-va: This could just check `isAgent` form std-env

packages/vitest/src/node/reporters/utils.ts
```diff
@@ -35,12 +36,14 @@ async function loadCustomReporterModule<C extends Reporter>(
...
   const promisedReporters = reporterReferences.map(
     async (referenceOrInstance) => {
       if (Array.isArray(referenceOrInstance)) {
-        const [reporterName, reporterOptions] = referenceOrInstance
+        const [rawReporterName, reporterOptions] = referenceOrInstance
+        const reporterName = isAgent && rawReporterName === 'default' ? 'agent' : rawReporterName
```
sheremet-va: I wonder if this code belongs _here_. I think it should be part of the config resolution in `resolveConfig`

I also wonder if it should affect verbose/tree

sheremet-va (changes requested):

packages/vitest/src/node/reporters/utils.ts
hi-ogawa: Deriving default should normally happen in `resolveConfig` utility. I don't think swapping like this internally is intuitive if we are explicitly making "agent" reporter.

https://github.com/vitest-dev/vitest/blob/b9bdb25fbfbe37825a74d942198648f207b55f00/packages/vitest/src/node/config/resolveConfig.ts#L731-L738

cpojer: Should be done now.

packages/vitest/src/node/reporters/utils.ts
cpojer: Done.

packages/vitest/src/node/cli/cac.ts
cpojer: Done.

packages/vitest/src/node/reporters/utils.ts
cpojer: First item is done.

Personally, I'd suggest changing the default mode first. If somebody explicitly selects another reporter, then I assume they have a reason for it.

AriPerkkio (commented):
I like the idea. I've actually been thinking about similar solution via reporter options that would just hide unnecessary noise that fills token usage. Also agents are eager to run Vitest CLI with `| head -100` kind of filters.

Need to check `std-env`'s `isAgent` first before commenting about the actual implementation. 👀 
Separate reporter is probably the best choice.

sheremet-va:
Seems like `isCI` check is different in the std-env v4 version, the tests are failing now :(

test/cli/test/reporters/default.test.ts
```diff
@@ -299,4 +299,81 @@ describe('default reporter', async () => {
     })
     expect(stderr).toMatchSnapshot()
   })
+
+  describe('agent reporter', () => {
```
sheremet-va: and these tests should probably be moved to `reporters/agent.test.ts` instead of being in the default one

test/cli/test/reporters/default.test.ts
```diff
@@ -299,4 +299,81 @@ describe('default reporter', async () => {
...
+        include: ['b1.test.ts', 'b2.test.ts'],
+        root: 'fixtures/reporters/default',
+        reporters: [['agent', {}]],
+        fileParallelism: false,
+        sequence: {
+          sequencer: StableTestFileOrderSorter,
+        },
+      })
+
+      expect(stdout).toContain('Test Files')
```
sheremet-va: We have a note in AGENTS.md to avoid using `toContain` and use `toMatchInlineSnapshot` instead - I wonder why it ignored it

test/cli/test/reporters/default.test.ts
cpojer: Done.

test/cli/test/reporters/default.test.ts
cpojer: Left one or two versions of it in (especially via `.not.toContain`) because it would snapshot too much stuff irrelevant to what we are testing.

sheremet-va: I personally prefer having more robust tests that could catch unexpected bugs. In your tests you have 2 identical `runVitest` calls in two different tests just to have different assertions when we could've just tested the whole output once or had several assertions in a single test:

- hides passed module headers and shows only failed tests
- still prints end summary

And these two:
- suppresses console logs from passed tests
- shows console logs from failed tests

cpojer: Updated.

cpojer:
I fixed the `std-env` CI issue. Not sure if something is wrong with the PR or that's the repo/CI but I still have a few failing CI tasks that mostly seem to be timeouts.

hi-ogawa:
I was testing around and found that success suite name shows up for some case. looks like condition is
- running single test file
- single test file includes multiple suites where one of suite has failing test case

For example:

```js
describe('good suite', () => {
  it('good test', () => {})
})

describe('bad suite', () => {
  it('bad test', () => {
    throw new Error('boom')
  })
})
```

```sh
 pnpm -C examples/basic test --reporter=agent suite.test.ts

> @vitest/example-test@ test /home/hiroshi/code/others/vitest/examples/basic
> vitest --reporter=agent suite.test.ts


 DEV  v4.1.0-beta.5 /home/hiroshi/code/others/vitest/examples/basic

 ❯ test/suite.test.ts (5 tests | 1 failed) 4ms
   ✓ suite name (3)
   ✓ good suite (1)
   ❯ bad suite (1)
     × bad test 1ms
```

hi-ogawa:
Oh, actually it's collapsed with non tty, so this is not an issue in practice.

```js
$ AI_AGENT=true pnpm -C examples/basic test suite.test.ts | cat

> @vitest/example-test@ test /home/hiroshi/code/others/vitest/examples/basic
> vitest suite.test.ts


 DEV  v4.1.0-beta.5 /home/hiroshi/code/others/vitest/examples/basic

 ❯ test/suite.test.ts (5 tests | 1 failed) 4ms
     × bad test 1ms

⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯
```

sheremet-va (dismissed):
This looks good to me, but should be documented. Let me know if you want to do it yourself, or someone should step in. Note to myself: include this in https://github.com/vitest-dev/vitest/pull/9775

cpojer:
@sheremet-va Added docs, let me know if you are happy with them.

hi-ogawa (dismissed):
looks good 👍

cpojer:
(I think one of you needs to merge this, I don't have access 😇)

packages/vitest/src/node/reporters/agent.ts
```diff
@@ -0,0 +1,50 @@
...
+  renderSucceed = false
+
+  constructor(options: DefaultReporterOptions = {}) {
+    super({ ...options, summary: false })
+  }
+
+  onTestRunStart(specifications: ReadonlyArray<TestSpecification>): void {
+    this.renderSucceed = false
+    super.onTestRunStart(specifications)
+  }
```
AriPerkkio, line 16: Setting `renderSucceed` before `super.onTestRunStart` while also setting its default value in class properties doesn't make sense to me.

https://github.com/vitest-dev/vitest/blob/dfb2159d3b758fa14c08c8d0cf9874fa3312954f/packages/vitest/src/node/reporters/default.ts#L34-L43

Should this instead be


```suggestion
  onTestRunStart(specifications: ReadonlyArray<TestSpecification>): void {
    super.onTestRunStart(specifications)
    this.renderSucceed = false
  }
```

packages/vitest/src/node/reporters/agent.ts
```diff
@@ -0,0 +1,50 @@
...
+    }
+    super.printTestModule(testModule)
+  }
+
+  protected printTestCase(_moduleState: TestModuleState, test: TestCase): void {
+    const testResult = test.result()
+    if (testResult.state === 'failed') {
+      const padding = this.getTestIndentation(test.task)
+      const suffix = this.getTestCaseSuffix(test)
+      this.log(c.red(` ${padding}${taskFail} ${this.getTestName(test.task, separator)}`) + suffix)
```
AriPerkkio: Isn't this same as below?


```suggestion
      super.printTestCase(_moduleState, test)
```

packages/vitest/src/node/reporters/agent.ts
```diff
@@ -0,0 +1,50 @@
...
+    }
+  }
+
+  shouldLog(log: UserConsoleLog, taskState?: TestResult['state']): boolean {
+    if (taskState !== 'failed') {
+      return false
+    }
+    return super.shouldLog(log, taskState)
+  }
```
AriPerkkio: This and `logFailedTask` override above are implementing same logic that `silent: 'passed-only'` option does. We could just enable that option when agent reporter it loaded 🤔

AriPerkkio (dismissed):
Overall looks good. Some comments about code duplication in reporter implementation, no blockers.

packages/vitest/src/node/reporters/agent.ts
cpojer: They are not the same. The base class with `renderSucceed = false` still prints slow tests and all passing tests when `moduleState === 'failed'`. Since `printTestModule` only passes through failed modules, super would print every test in the module, not just failures.

packages/vitest/src/node/reporters/agent.ts
cpojer: Isn't the silent option read from the shared context? If so, mutating it would affect all other reporters in the same test run.

packages/vitest/src/node/reporters/agent.ts
AriPerkkio: Can you point those lines in base class? It's `else if`'s, and we are only calling the `super` after checking the `testResult.state`.

https://github.com/vitest-dev/vitest/blob/dfb2159d3b758fa14c08c8d0cf9874fa3312954f/packages/vitest/src/node/reporters/base.ts#L206-L213

packages/vitest/src/node/reporters/agent.ts
AriPerkkio: Yep. Maybe the `config.silent` should be used as base reporter option that could be overriden in extended reporters.

packages/vitest/src/node/reporters/agent.ts
cpojer: Yup, sorry, got confused. Fixed now.

packages/vitest/src/node/reporters/agent.ts
cpojer: Done.

AriPerkkio (approved):
Looks good! 💯

sheremet-va (approved):

PR #9802, opened 2026-03-05
feat: Introduce a new DOM matcher: `toHaveTagName`

porada:
### Description

This PR implements a new DOM matcher for browser testing: `toHaveTagName`.

`toHaveTagName` is a simple but useful matcher for testing polymorphic components and is much more expressive than checking the `tagName` property directly.

Please compare the following:

```ts
const button = getByTestId('submit-button');

expect(button).toBeVisible();
expect(button).toHaveTagName('button');
expect(button).toHaveTextContent('Sign Up');
```

```ts
expect(button).toBeVisible();
expect(button.element().tagName).toBe('BUTTON'); // 👎
expect(button).toHaveTextContent('Sign Up');
```

### Checklist

- [ ] It’s really useful if your PR references an issue where it is discussed ahead of time. If the feature is substantial or introduces breaking changes without a discussion, PR might be closed.
- [x] Please, don’t make changes to `pnpm-lock.yaml` unless you introduce a new test example.
- [x] Please check [Allow edits by maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to make review process faster. Note that this option is not available for repositories that are owned by GitHub organizations.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/browser/jest-dom.d.ts
+++ b/packages/browser/jest-dom.d.ts
@@ -379,6 +379,23 @@ export interface TestingLibraryMatchers<E, R> {
   toHaveStyle(css: string | Partial<CSSStyleDeclaration>): R
   /**
    * @description
+   * Assert whether the given element has a specific tag name.
+   * @example
+  * <button
+  *   type="submit"
+  *   data-testid="submit-button"
+  * >
+  *   Submit
+  * </button>
+   *
+   * const submit = page.getByTestId('submit-button')
+   * await expect.element(submit).toHaveTagName('button')
+   * await expect.element(submit).not.toHaveTagName('a')
+   * @see https://vitest.dev/api/browser/assertions#tohavetagname
+   */
+  toHaveTagName(expectedTagName: Lowercase<string>): R
+  /**
+   * @description
    * Check whether the given element has a text content or not.
    *
    * When a string argument is passed through, it will perform a partial case-sensitive match to the element
--- a/packages/browser/src/client/tester/expect/index.ts
+++ b/packages/browser/src/client/tester/expect/index.ts
@@ -21,6 +21,7 @@ import toHaveFormValues from './toHaveFormValues'
 import toHaveRole from './toHaveRole'
 import toHaveSelection from './toHaveSelection'
 import toHaveStyle from './toHaveStyle'
+import toHaveTagName from './toHaveTagName'
 import toHaveTextContent from './toHaveTextContent'
 import toHaveValue from './toHaveValue'
 import toMatchScreenshot from './toMatchScreenshot'
@@ -45,6 +46,7 @@ export const matchers: MatchersObject = {
   toHaveFocus,
   toHaveFormValues,
   toHaveStyle,
+  toHaveTagName,
   toHaveTextContent,
   toHaveValue,
   toHaveDisplayValue,
--- a/packages/browser/src/client/tester/expect/toHaveTagName.ts
+++ b/packages/browser/src/client/tester/expect/toHaveTagName.ts
@@ -1,0 +1,35 @@
+import type { ExpectationResult, MatcherState } from '@vitest/expect'
+import type { Locator } from '../locators'
+import { getElementFromUserInput, getMessage, getTag } from './utils'
+
+export default function toHaveTagName(
+  this: MatcherState,
+  actual: Element | Locator,
+  expectedTagName: Lowercase<string>,
+): ExpectationResult {
+  const htmlElement = getElementFromUserInput(actual, toHaveTagName, this)
+  const receivedTagName = getTag(htmlElement).toLowerCase()
+
+  if (typeof expectedTagName === 'string') {
+    expectedTagName = expectedTagName.toLowerCase() as Lowercase<string>
+  }
+
+  return {
+    pass: this.equals(receivedTagName, expectedTagName, this.customTesters),
+    message: () => {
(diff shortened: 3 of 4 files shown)


porada:
@sheremet-va — Before I add documentation, could you let me know if this is a direction you’d be open to?

porada:
Note: `pnpm test:ci` seems to be failing for reasons unrelated to this PR.

sheremet-va:
To be honest, seems a bit too niche to me. We generally promote testing-library's principles and avoid testing implementation details, and tag name _is_ the implementation detail. I would advise testing the role instead

AriPerkkio (commented):
I don't think checking tag is something we should encourage users to. They should default to checking role, that is a lot more better than tag name. Checking tag makes sense sometimes, but in that case I would recommend users to do it explicitly without extra matchers.

So 👎 from me.

porada:
I understand the push for testing `role`. However, there are scenarios where that alone is not enough, because some HTML semantics can’t be asserted through roles.

Elements like `<article>`, `<footer>`, `<nav>`, or `<section>` only map to ARIA roles when specific conditions are met. In other contexts they either expose a generic role or none at all. If a project enforces semantic HTML—for document structure, accessibility discipline, or SEO—there needs to be a direct way to assert the actual element being rendered.

Another angle is that multiple HTML elements can resolve to the same role. For example, a `<nav>` and a `<div role="navigation">` are indistinguishable to role-based queries even though they are not equivalent semantically. In those cases the only reliable distinction is the element itself.

> I don’t think checking tag is something we should encourage users to.

I agree that role-based testing should be the default. I imagine proper documentation would clearly describe the distinction and the cases where asserting the tag itself is appropriate.

> I would recommend users to do it explicitly without extra matchers.

Unlike most DOM properties, `tagName` introduces extra friction because it is always uppercase. Tests end up with patterns like `expect(el.element().tagName).toBe('BUTTON')`, which is noisy, easy to get wrong at first, and—frankly—just ugly.

Being able to assert the tag name directly—normalized, without dropping down to a raw DOM property, and in a way that is consistent with the rest of the testing API—is simply better DX.

AriPerkkio:
> Elements like `<article>`, `<footer>`, `<nav>`, or `<section>` only map to ARIA roles when specific conditions are met. In other contexts they either expose a generic role or none at all.

Can you show some HTML snippets of such cases? All of these elements have implicit ARIA roles. Only the `<section>` one requires accessible name to be present for the role. Testing `section` by tag name would make no sense in that case. 


> Another angle is that multiple HTML elements can resolve to the same role. For example, a `<nav>` and a `<div role="navigation">` are indistinguishable to role-based queries even though they are not equivalent semantically. In those cases the only reliable distinction is the element itself.

Yes, this is the point of explicit ARIA roles. When adding role to div, you are making it a navigation element. It's intentional, and should pass your tests too. Interaction testing is needed to check if it was implemented properly.

sheremet-va:
This doesn't align with the best practices and how we want to promote them; the team decided to reject your PR, but thank you for the effort. Thankfully, you can still have it locally or publish it as an external package.

porada:
Assuming that roles always provide a complete picture of semantics goes against best practices.

```html
<section aria-label="Gallery"><!-- role="region" --></section>

<section><!-- not a region --></section>
```

@AriPerkkio — You mentioned this case, but there are more cases where roles depend on context:

```html
<header><!-- role="banner" --></header>

<section>
  <header><!-- not a banner --></header>
</section>
```

```html
<footer><!-- role="contentinfo" --></footer>

<article>
  <footer><!-- not contentinfo --></footer>
</article>
```

The point here is that context matters when it comes to roles, and that context is not always present when testing individual components in isolation.

Another example is icons that can be rendered as `<span>` or `<div>` depending on layout context. The accessible output is identical, but the inline vs. block semantics are intentional and part of the component contract. Element assertions make that easy to verify.

This becomes even more relevant with polymorphic components (`as="tag"` is a common pattern). In those cases, the role intentionally stays the same while the underlying element changes.

The case where I personally end up using `toHaveTagName` most often is verifying navigation vs. action elements. For example, ensuring that links with an `href` render as `<a>` rather than `<button>`, since the role can still appear identical while the underlying browser behavior is different.

Given that this is a small and non-invasive addition that primarily improves DX, the rejection is surprising.

sheremet-va:
> Given that this is a small and non-invasive addition that primarily improves DX, the rejection is surprising.

Adding this would be an endorsement. And we do not endorse testing implementation details. Having a note buried in the docs won't stop people from choosing a shortcut.

> The case where I personally end up using `toHaveTagName` most often is verifying navigation vs. action elements. For example, ensuring that links with an `href` render as `<a>` rather than `<button>`, since the role can still appear identical while the underlying browser behavior is different.

Maybe it's better to have more granular matchers like `toHaveLink` and `toHaveButton` then. You can do anything inside of it.

Public facing assertions (the code in the test) should not rely on implementation detail, `tag` is one. The assertion implementation can do anything as long as it is expressive. Users see links and buttons, not `a` tag with a `href`.

porada:
Would exposing a normalized `tagName` value on the element locator be a reasonable compromise to this?

sheremet-va:
> Would exposing a normalized `tagName` value on the element locator be a reasonable compromise to this?

No, it has the same issues we already discussed here

PR #9822, opened 2026-03-08
fix(config): validate sequence.groupOrder is a finite number

### LINKED ISSUE #9821: [Bug] sequence.groupOrder silently accepts invalid types (NaN coercion) + file-based projects fail with CLI
## Environment

| | |
|---|---|
| **Vitest** | 4.0.18 |
| **Node.js** | 25.8.0 |
| **OS** | macOS Darwin 25.3.0 |
| **Package Manager** | npm 11.x |

---

### Description

The `sequence.groupOrder` config option is typed as `number` but has **no runtime validation**. When a non-number value (e.g., a string array `['fast']`) is passed, it silently coerces to `NaN` via `groups[order]` lookup, which then falls back to `0`. This causes all projects to receive `groupOrder = 0`, triggering the error:

```
Error: Projects "app" and "fast" have different 'maxWorkers' but same 'sequence.groupId'
```

### Reproduction

**`vitest.config.ts`:**

```ts

export default defineConfig({
  test: {
    projects: [
      {
        test: {
          name: 'fast',
          include: ['src/**/*.fast.test.ts'],
          pool: 'threads',
          maxWorkers: 4,
          sequence: {
            // BUG: type says `number`, but no runtime check.
            // Passing a string array silently coerces to NaN → 0
            groupOrder: ['fast'] as any,
          },
        },
      },
      {
        test: {
          name: 'slow',
          include: ['src/**/*.slow.test.ts'],
          pool: 'forks',
          maxWorkers: 1,
          sequence: {
            groupOrder: ['slow'] as any,
          },
        },
      },
    ],
  },
});
```

**`src/example.fast.test.ts`:**

```ts
import { test, expect } from 'vitest';
test('fast test', () => expect(1 + 1).toBe(2));
```

**`src/example.slow.test.ts`:**

```ts
import { test, expect } from 'vitest';
test('slow test', () => expect(2 + 2).toBe(4));
```

**Run:**

```bash
npx vitest run
```

### Expected behavior

Vitest should either:
1. **Throw a clear validation error** at config resolution time: `"sequence.groupOrder must be a number, received: ['fast']"`
2. Or **coerce to a valid default** (e.g., `0`) with a warning

### Actual behavior

- `groupOrder` receives `['fast']` (a string array)
- In the grouping logic (`groupSpecs` in `packages/vitest/src/node/specs.ts`), `groups[order]` is called where `order` is the invalid value
- Array-to-number coercion: `Number(['fast'])` → `NaN`
- `NaN` is used as an object key, which JavaScript normalizes — all projects end up in the same group (key `"NaN"` or fallback to `0`)
- Since projects with different `maxWorkers` / `pool` settings end up in the same group, Vitest throws the "different maxWorkers but same groupId" error
- **No indication** that the root cause is an invalid config value

### Root cause analysis

The issue spans two locations in the source:

1. **Config resolution** (`packages/vitest/src/node/config/resolveConfig.ts`):
   - `sequence.groupOrder` is passed through without any type validation
   - The Zod/schema validation (if any) does not catch array values at runtime because `as any` bypasses TypeScript, and there is no runtime guard

2. **Spec grouping** (`packages/vitest/src/node/specs.ts`, `groupSpecs` function):
   - The function reads `project.config.sequence.groupOrder` and uses it as an index/key
   - When the value is `NaN` (from invalid coercion), all projects map to the same group
   - The subsequent check `if (a.maxWorkers !== b.maxWorkers) throw` fires because heterogeneous projects were incorrectly grouped together

### Suggested fix

```ts
// In resolveConfig.ts, after resolving sequence options:
if (resolved.sequence.groupOrder !== undefined) {
  const order = Number(resolved.sequence.groupOrder);
  if (!Number.isFinite(order)) {
    throw new Error(
      `Invalid "sequence.groupOrder": expected a finite number, got ${JSON.stringify(resolved.sequence.groupOrder)}`
    );
  }
  resolved.sequence.groupOrder = order;
}
```

Alternatively, add validation in `groupSpecs` as a defensive check:

```ts
const order = Number(project.config.sequence?.groupOrder ?? 0);
if (!Number.isFinite(order)) {
  throw new Error(
    `Project "${project.name}" has invalid sequence.groupOrder: ${JSON.stringify(project.config.sequence?.groupOrder)}`
  );
}
```

---

### Description

When using file-based project references (e.g., `projects: ['./vitest.fast.config.ts']`), the Node API `createVitest` + `globTestSpecifications()` correctly discovers and runs all test files. However, running via CLI (`vitest run`) or `startVitest()` reports "No test files found, exiting with code 1" even though the pool actually executes the tests successfully (0 `testModules` reported).

### Reproduction

**Directory structure:**

```
repro/
├── vitest.config.ts
├── vitest.fast.config.ts
├── vitest.slow.config.ts
└── src/
    ├── math.fast.test.ts
    └── io.slow.test.ts
```

**`vitest.config.ts`:**

```ts

export default defineConfig({
  test: {
    projects: [
      './vitest.fast.config.ts',
      './vitest.slow.config.ts',
    ],
  },
});
```

**`vitest.fast.config.ts`:**

```ts

export default defineConfig({
  test: {
    name: 'fast',
    include: ['src/**/*.fast.test.ts'],
  },
});
```

**`vitest.slow.config.ts`:**

```ts

export default defineConfig({
  test: {
    name: 'slow',
    include: ['src/**/*.slow.test.ts'],
  },
});
```

**`src/math.fast.test.ts`:**

```ts
import { test, expect } from 'vitest';
test('math fast', () => expect(1 + 1).toBe(2));
```

**`src/io.slow.test.ts`:**

```ts
import { test, expect } from 'vitest';
test('io slow', () => expect(true).toBe(true));
```

**Run via Node API (WORKS):**

```ts
// run-api.ts
import { createVitest } from 'vitest/node';

const vitest = await createVitest('test', {});
const specs = await vitest.globTestSpecifications();
console.log('Found specs:', specs.length); // → 2 ✓
await vitest.runTestSpecifications(specs);
await vitest.close();
```

```bash
npx tsx run-api.ts
# Output: Found specs: 2, tests pass ✓
```

**Run via CLI (FAILS):**

```bash
npx vitest run
# Output: No test files found, exiting with code 1
```

**Run via startVitest (FAILS):**

```ts
// run-start.ts
import { startVitest } from 'vitest/node';

const vitest = await startVitest('test', []);
// Output: No test files found, exiting with code 1
// vitest is undefined (process.exit called)
```

**Individual project works:**

```bash
npx vitest run --project=fast   # ✓ works
npx vitest run --project=slow   # ✓ works
```

### Expected behavior

`vitest run` and `startVitest()` should discover and run tests from all file-based projects, identical to the `createVitest` + `globTestSpecifications()` path.

### Actual behavior

1. The pool runner **does execute** — workers spin up and tests actually run inside the pool
2. But the pool returns `0` `testModules` to the reporter
3. The reporter's `onTestRunEnd` callback receives an empty test module array
4. The default reporter calls `printNoTestFound` → exits with code 1
5. Individual `--project=<name>` works correctly — only the combined run fails

### Root cause analysis

The issue is in how `startVitest` (in `packages/vitest/src/node/cli/cli-api.ts`) orchestrates the test run differently from `createVitest`:

1. **`startVitest`** calls `vitest.start()` which internally calls `runTests()`
2. In `runTests()` (`packages/vitest/src/node/core.ts`), the spec collection for file-based projects follows a different code path than `globTestSpecifications()`
3. When projects are defined as file paths (strings), the root project's `include` patterns are empty (it delegates to child configs). The CLI path collects specs from the root project context, finds nothing, and short-circuits
4. `globTestSpecifications()` (the Node API path) correctly iterates over resolved child projects and collects their specs independently

The divergence is in `packages/vitest/src/node/core.ts`:
- The `start()` → `runTests()` path uses `getTestSpecifications()` which may not properly resolve file-based project references
- The `globTestSpecifications()` public API method correctly resolves all projects via `this.projects` iteration

In the reporter chain (`packages/vitest/src/node/reporters/`):
- `onTestRunEnd` checks `testModules.length === 0`
- When it's zero, `printNoTestFound()` is called, which outputs "No test files found" and triggers exit code 1
- This masks the fact that the pool DID run — the issue is in spec collection, not execution

### Suggested fix

In `packages/vitest/src/node/core.ts`, the `getTestSpecifications()` method (or equivalent used by `start()`) should use the same project iteration logic as `globTestSpecifications()`:

```ts
// Current (broken): only collects from root project
const specs = await this.project.getTestSpecifications();

// Fix: iterate over all resolved projects (same as globTestSpecifications)
const specs = (await Promise.all(
  this.projects.map(project => project.getTestSpecifications())
)).flat();
```

This would align the CLI path with the Node API path, ensuring file-based projects are correctly resolved in both scenarios.

---

## Summary

| Bug | Severity | Workaround |
|-----|----------|------------|
| #1: `groupOrder` NaN coercion | Medium | Ensure `groupOrder` is always a `number` in config |
| #2: File-based projects + CLI | High | Use `createVitest` + `globTestSpecifications()` instead of `startVitest` / CLI, or use inline project definitions instead of file references |

## Related issues

- [#8855 — sequence.groupId config doesn't exist](https://github.com/vitest-dev/vitest/issues/8855)
- [#8894 — Project order changed with v4](https://github.com/vitest-dev/vitest/issues/8894)
- [#1950 — startVitest exits the process instead of returning a result if no tests found](https://github.com/vitest-dev/vitest/issues/1950)
- [#7076 — reporters return inconsistent results when no tests are found](https://github.com/vitest-dev/vitest/issues/7076)

youredik:
## Description

When `sequence.groupOrder` receives a non-number value (e.g., an array like `['fast']`), it silently coerces to `NaN` via `groups[order]` lookup in `groupSpecs`. All projects then share the same group key, triggering a confusing error:

```
Error: Projects "fast" and "unit" have different 'maxWorkers' but same 'sequence.groupOrder'.
Provide unique 'sequence.groupOrder' for them.
```

There is no indication that the root cause is an invalid config value — the user sees an error about `maxWorkers` grouping, not about the invalid type.

## Root cause

In `resolveConfig.ts`, `sequence.groupOrder` is defaulted with `??= 0` but has no type validation. When a non-number value passes through:

1. `Number(['fast'])` → `NaN`
2. `groups[NaN]` creates a single group for all projects
3. Projects with different `maxWorkers` end up in the same group → error

## Fix

Added runtime validation in `resolveConfig.ts`:
- Checks `Number.isFinite()` after coercion
- Throws a clear error: `Invalid "sequence.groupOrder" value: expected a finite number, received ["fast"]`
- Coerces valid numeric strings (e.g., `"1"`) to numbers gracefully

## Test

Added a test case in `test/cli/test/group-order.test.ts` that passes an array instead of a number and verifies the validation error message.

Fixes #9821

---

| | |
|---|---|
| **Type** | Bug fix |
| **Scope** | `packages/vitest/src/node/config/resolveConfig.ts` |
| **Test** | `test/cli/test/group-order.test.ts` |

### DIFF
--- a/packages/vitest/src/node/config/resolveConfig.ts
+++ b/packages/vitest/src/node/config/resolveConfig.ts
@@ -775,7 +775,7 @@ export function resolveConfig(
   if (resolved.sequence.groupOrder != null) {
     const order = Number(resolved.sequence.groupOrder)
     if (!Number.isFinite(order)) {
-      throw new Error(
+      throw new TypeError(
         `Invalid "sequence.groupOrder" value: expected a finite number, received ${JSON.stringify(resolved.sequence.groupOrder)}`,
       )
     }
--- a/packages/vitest/src/node/config/resolveConfig.ts
+++ b/packages/vitest/src/node/config/resolveConfig.ts
@@ -772,7 +772,18 @@ export function resolveConfig(
       ? RandomSequencer
       : BaseSequencer
   }
-  resolved.sequence.groupOrder ??= 0
+  if (resolved.sequence.groupOrder != null) {
+    const order = Number(resolved.sequence.groupOrder)
+    if (!Number.isFinite(order)) {
+      throw new Error(
+        `Invalid "sequence.groupOrder" value: expected a finite number, received ${JSON.stringify(resolved.sequence.groupOrder)}`,
+      )
+    }
+    resolved.sequence.groupOrder = order
+  }
+  else {
+    resolved.sequence.groupOrder = 0
+  }
   resolved.sequence.hooks ??= 'stack'
   // Set seed if either files or tests are shuffled
   if (resolved.sequence.sequencer === RandomSequencer || resolved.sequence.shuffle) {
--- a/test/cli/test/group-order.test.ts
+++ b/test/cli/test/group-order.test.ts
@@ -1,6 +1,27 @@
 import { expect, test } from 'vitest'
 import { runInlineTests } from '../../test-utils'
 
+test('throws on invalid groupOrder type (array instead of number)', async () => {
+  const { stderr } = await runInlineTests({
+    'example.test.ts': `test('ok', () => {})`,
+  }, {
+    $cliOptions: { globals: true },
+    projects: [
+      {
+        test: {
+          name: 'fast',
+          include: ['./example.test.ts'],
+          sequence: {
+            // @ts-expect-error -- testing runtime validation
+            groupOrder: ['fast'],
+          },
+        },
+      },
+    ],
+  })
+  expect(stderr).toContain('Invalid "sequence.groupOrder" value')
+})
+
 test('tests run according to the group order', async () => {
   const { stdout, stderr } = await runInlineTests({
     'example.1.test.ts': `test('1', () => {})`,


youredik:
## CI Status

After the lint fix (`TypeError` instead of `Error` per `unicorn/prefer-type-error`), **12/14 checks pass** including the `group-order.test.ts` validation test on all platforms.

The 2 remaining failures are **pre-existing flaky tests on macOS**, unrelated to this PR:

| Failed job | Failing test | Related to this PR? |
|---|---|---|
| Browsers: node-24, macos | `hooks-timeout.test.ts` (browser hook timeouts) | ❌ No |
| Rolldown&Test: node-22, macos | `pool.test.ts > threads > can capture worker's stdout/stderr` | ❌ No |

Our `group-order.test.ts` passes on all platforms:
- ✅ `throws on invalid groupOrder type` 
- ✅ `tests run according to the group order`

sheremet-va:
In general, we don't validate input values. We probably (?) should, but that would be a complex refactor.

PR #9826, opened 2026-03-09
fix: resolve fixture overrides from test's suite in `beforeEach` hooks

### LINKED ISSUE #9810: `test.override` not affecting fixtures used by auto fixtures
### Describe the bug

There's no way to override fixtures that are used by auto fixtures. Code below keeps throwing `Config missing`. Note that this condition also requires `setupFile` to be there.

```ts
import { test as baseTest, describe } from 'vitest';

const test = baseTest
  .extend('config', (): undefined | {port: number, host: string} => undefined)
  .extend('server', { auto: true }, ({ config }) => {
    console.log({config})

    if (!config) {
      throw new Error('Missing config');
    }

    return `http://${config.host}:${config.port}`;
  });

describe('production environment', () => {
  test.override('config', { port: 8080, host: 'api.example.com' });

  test('uses production config', () => {});
});
```

```ts
import { beforeEach } from 'vitest';

beforeEach(({ task }) => {
    console.log(task.id)
});
```

Repro uses new builder pattern, but same issue happens with old pattern too.

### Reproduction

https://stackblitz.com/~/edit/vitest-dev-vitest-ildnktb3?file=basic.test.ts&initialPath=/__vitest__/

### System Info

```shell
Stackblitz, `vitest@4.1.0-beta.5`
```

### Used Package Manager

npm

### Validations

hi-ogawa:
### Description

- Closes https://github.com/vitest-dev/vitest/issues/9810

### Tests

### Documentation

### Changesets


test/core/test/test-extend.test.ts
```diff
@@ -640,3 +640,136 @@ describe('builder pattern with non-function values', () => {
...
+
+  describe('with override', () => {
+    myTest.override('base', 'base:override')
+
+    myTest('beforeEach sees overridden fixture', ({ derived }) => {
+      expect(hookValues).toEqual(['base:override'])
+      expect(derived).toBe('derived:base:override')
+    })
+  })
+})
```
hi-ogawa, line 723: The issue wasn't only `auto: true` case. This was also failing with the same reason that `beforeEach` outside of `describe` aren't picking up inner `override`.

AriPerkkio (dismissed):
Looks good, thanks for fixing! 💯

packages/runner/src/fixture.ts
```diff
@@ -314,6 +314,11 @@ export function withFixtures(fn: Function, options?: WithFixturesOptions) {
       return fn(context)
     }
 
+    // For `xxxEach` hooks, use the test's parent suite to pick up fixture
+    // overrides registered in the test's describe block.
+    // For `test`, `collectorSuite` is already same as `context.task.suite`.
+    // For `xxxAll` hooks, there's no `context.task`.
+    const suite = context.task?.suite ?? collectorSuite
```
sheremet-va: I want to avoid using `context.task` anywhere in fixtures resolution, it's an implied value that makes this harder to track, in my opinion. Could we pass suite down as an option instead?

sheremet-va (changes requested):

packages/runner/src/fixture.ts
hi-ogawa: I didn't see it, but that indeed shows the flow better. I think this is what you mean. Updated.

sheremet-va (approved):

PR #9830, opened 2026-03-09
fix: ensure sequential mock/unmock resolution

hi-ogawa:
### Description

Resolves https://discord.com/channels/804011606160703521/1480463846253662218/1480481214170402993

Using `Promise.all` for handling `mock` and `unmock` in parallel can cause race condition when multiple `mock` and `unmock` of same modules are involved. 

The above CI failures is related because it goes through

- "first import"
  - beforeEach -> doMock
  - test -> import
  - afterEach -> doUnmock
- "second import"
  - beforeEach -> doMock
  - test -> import  
    - 👈 At this step, unmock/mock requests are handled in parallel. If `doUnmock` processed slower, then module doesn't get mocked and thus module not found.

https://github.com/vitest-dev/vitest/blob/4107f8ea53cb0e4a077405ba3d5ca44152345744/test/core/test/unmock-import.test.ts#L4-L27

```js
  ⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯
   FAIL   threads  test/unmock-import.test.ts:23:1 > second import should have been re-mocked
  Error: Cannot find module '/data'
   ❯ test/unmock-import.test.ts:25:20
       23| test('second import should have been re-mocked', async () => {
       24|   // @ts-expect-error I know this
       25|   const { data } = await import('/data')
         |                    ^
       26|   expect(data.state).toBe('STARTED')
       27| })
  ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯
```

### Tests

### Documentation

### Changesets


sheremet-va:
Could we still have it in parallel if all of them are of the same type? (mock/unmock) I feel like this can hit performance hard otherwise.

hi-ogawa:
That optimization should work fine. Just Claude pushed weird code, so don't look that yet :upside_down_face:

packages/vitest/src/runtime/moduleRunner/bareModuleMocker.ts
```diff
@@ -152,27 +152,37 @@ export class BareModuleMocker implements TestModuleMocker {
...
-          )
-        }
-      }),
-    )
+      }
+    }
+
+    // resolve in parallel if all pending mocks are of the same type,
+    // otherwise resolve sequentially to preserve mock/unmock ordering
+    const pendingIds = BareModuleMocker.pendingIds
```
sheremet-va: This still seems harsh. Imagine,

```ts
vi.mock() // 20 times in a setup file
vi.unmock() // single time in the test file
```

Now we have to wait for 20 mocks to be loaded just to unmock one of them.

We should await all actions of the same group

```
[mock, mock, mock]
[unmock]
[mock, mock]
[unmock, unmock]
```

Run in parallel inside the group, but run groups in sequence

sheremet-va (changes requested):

sheremet-va (approved):

PR #9847, opened 2026-03-11
fix(browser): take failure screenshot if `toMatchScreenshot` can't capture a stable screenshot

### LINKED ISSUE #9690: [Regression] Browser mode: Failure screenshot not created when unstable error occurs in v4.1.0-beta.4
### Describe the bug

Regression: In v4.1.0-beta.4, the changes from https://github.com/vitest-dev/vitest/pull/9552 (fix: don't take failure screenshot if tests have artifacts created by `toMatchScreenshot`) unintentionally prevent a failure screenshot from being captured and saved in unstable test failures.

**Problem:**
When a test fails with an unstable screenshot error (e.g., `Could not capture a stable screenshot within...`), Vitest now only provides a reference screenshot artifact. This makes it impossible to see what was actually rendered at the point of instability, hampering debugging ability. Previously, an 'internal:failureScreenshot' artifact was attached, but after this change, it is missing for these failures if any `internal:toMatchScreenshot` artifacts exist.

**Expected:**
When a test fails due to instability or any other reason that doesn't generate regular diff/actual artifacts, the failure screenshot should still be saved and attached, probably in `.vitest-attachments` folder though and not next to referenced screenshot. 

### Reproduction

Attaching minimal reproduction case

[unstable-reproduction-test.zip](https://github.com/user-attachments/files/25393333/unstable-reproduction-test.zip)

### System Info

```shell
System:
    OS: macOS 26.2
    CPU: (12) arm64 Apple M3 Pro
    Memory: 384.31 MB / 36.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.15.1 - /Users/smilaite/.nvm/versions/node/v22.15.1/bin/node
    npm: 10.9.2 - /Users/smilaite/.nvm/versions/node/v22.15.1/bin/npm
    pnpm: 9.15.2 - /Users/smilaite/Library/pnpm/pnpm
  Browsers:
    Chrome: 145.0.7632.76
    Chrome Canary: 147.0.7693.0
    Firefox: 146.0
    Safari: 26.2
```

### Used Package Manager

npm

### Validations

macarie:
### Description

Resolves #9690

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/browser/src/client/tester/expect/toMatchScreenshot.ts
+++ b/packages/browser/src/client/tester/expect/toMatchScreenshot.ts
@@ -120,5 +120,9 @@ export default async function toMatchScreenshot(
           ]
             .filter(element => element !== null)
             .join('\n'),
+    meta: {
+      assertion: 'toMatchScreenshot',
+      outcome: result.outcome,
+    },
   }
 }
--- a/packages/browser/src/client/tester/expect/toMatchScreenshot.ts
+++ b/packages/browser/src/client/tester/expect/toMatchScreenshot.ts
@@ -121,7 +121,6 @@ export default async function toMatchScreenshot(
             .filter(element => element !== null)
             .join('\n'),
     meta: {
-      assertion: 'toMatchScreenshot',
       outcome: result.outcome,
     },
   }
--- a/packages/expect/src/jest-extend.ts
+++ b/packages/expect/src/jest-extend.ts
@@ -65,7 +65,7 @@ function getMatcherState(
 }
 
 class JestExtendError extends Error {
-  constructor(message: string, public actual?: any, public expected?: any) {
+  constructor(message: string, public actual?: any, public expected?: any, public meta?: object) {
     super(message)
   }
 }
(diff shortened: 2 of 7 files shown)


packages/browser/src/client/tester/expect/toMatchScreenshot.ts
```diff
@@ -120,5 +120,9 @@ export default async function toMatchScreenshot(
           ]
             .filter(element => element !== null)
             .join('\n'),
+    meta: {
+      assertion: 'toMatchScreenshot',
+      outcome: result.outcome,
+    },
```
macarie, line 125: Not sure if this is the best way to accomplish this, but I didn't find other ways to check source/origin of an error and to attach additional data

hi-ogawa: Looks good to me, but one question. Is `assertion` here and `expectAssertionName` available in `JestExtendPlugin` always same?

https://github.com/vitest-dev/vitest/blob/fea46b3a21d8f68c81de84406d64cb1e4cf2e66a/packages/expect/src/jest-extend.ts#L80

We could always add assertion as error metadata in the plugin level, then move properties like `JestExtendError.contenxt: { assertionName, meta }`.

macarie: Could definitely do that — I wasn't sure if it made sense to add it to all errors, but I guess it might be useful in the future 👍🏼

packages/expect/src/jest-extend.ts
```diff
@@ -65,9 +65,20 @@ function getMatcherState(
...
+
   constructor(message: string, public actual?: any, public expected?: any) {
     super(message)
   }
+
+  setContext(assertionName: string, meta?: object): this {
+    this.context = { assertionName, meta }
+
+    return this
+  }
```
macarie: Should this be the 4th constructor parameter instead?

hi-ogawa: I think so. I was wondering `JestExtendError.constructor` is somehow public, but probably not.

macarie: Yep, it doesn't seem to be public. It's only used in this file

hi-ogawa:
Is this testable?

test/browser/specs/failure-screenshot.test.ts
```diff
@@ -0,0 +1,100 @@
...
+          reporters: ['verbose'],
+          update: 'new',
+        },
+      }`,
+  })
+}
+
+describe('failure screenshots', () => {
+  describe('`toMatchScreenshot`', () => {
+    test('usually does NOT produce a failure screenshot', async () => {
```
macarie, line 35: This should technically control all exit paths of the assertion, but I don't think it's worth the added time in each CI run. It might be easier to just test the cases that expect the opposite (like the one below) as the condition is quite straightforward.

Edit: since the line this comment is talking about is not clear in the UI, the subject is the test with the `usually does NOT produce a failure screenshot` description.

hi-ogawa (approved):

sheremet-va (approved):

PR #9851, opened 2026-03-12
fix: disable colors if agent is detected

sheremet-va:
- Follow-up fix for https://github.com/vitest-dev/vitest/pull/9779
- Should this be configurable?


hi-ogawa (dismissed):
> configurable?

I don't think that's needed now

AriPerkkio (dismissed):
Looks good. There's still some icon colors that are set at module load level but that's fine I think.

https://github.com/vitest-dev/vitest/blob/731b14440b1b5be60bbfe6e899ec5ad101f87693/packages/vitest/src/node/reporters/renderers/utils.ts#L20-L28

> ```sh
>   RUN  v4.1.0-beta.6 /x/vitest/test/config/fixtures/console-color
> 
> stdout | basic.test.ts\x1B[2m > \x1B[22mconsole color
> true
> 
>  \x1B[32m✓\x1B[39m basic.test.ts (1 test) 1ms
> 
>  Test Files  1 passed (1)
>       Tests  1 passed (1)
>    Start at  22:07:30
>    Duration  85ms (transform 11ms, setup 0ms, import 16ms, tests 1ms, environment 0ms)
> ```

sheremet-va:
> Looks good. There's still some icon colors that are set at module load level but that's fine I think.

Moved the check to CLI, so we disable it before reporters are loaded.

AriPerkkio (approved):

robokozo:
Can you try this in the [kiro ide](https://kiro.dev/) (vscode fork)? I'm running the tests in the integrated terminal and the colors stopped rendering. I tried the same in vscode and the colors showed up normally.

hi-ogawa:
@robokozo That's expected since kiro is detected as ai agent https://npmx.dev/package/std-env#user-content-agent-detection, so Vitest should try to reduce irrelevant output to maximize agents efficiency. That said, if kiro (or other editor encountering the same issue) has specific workflow and color being desirable here, we can probably make it configurable at least. Please create an issue / feature request with your workflow explained.

PR #9860, opened 2026-03-13
fix(browser): use resolve instead of join for screenshotDirectory path resolution

### LINKED ISSUE #8604: `toMatchScreenshot` still creates a folder with the test's filename even when specifing otherwise
### Describe the bug

I have specified in the `vitest.config.ts` a different path for the screenshot directory like so:

```ts
resolveScreenshotPath: ({ root, testFileDirectory, screenshotDirectory, arg, browserName, platform, ext }) => 
    `${root}/${testFileDirectory}/${screenshotDirectory}/${arg}-${browserName}-${platform}${ext}`,

resolveDiffPath: ({ root, testFileDirectory, screenshotDirectory, arg, browserName, platform, ext }) => 
    `${root}/${testFileDirectory}/${screenshotDirectory}/diff/${arg}-${browserName}-${platform}${ext}`,
```

and yet whenever i update my screenshot with the command `pnpm vitest run --update`, it creates the folder with the test file's name (in this screenshot the file name is `button.visual.spec.tsx`)

Image

also important to note that as you can see, it doesnt insert any of the files into it, but simply creates it.

### Reproduction

i feel like the issue is pretty straight forward... if a reproduction is absolutely necessary ill make one.

### System Info

```shell
System:
    OS: macOS 15.6.1
    CPU: (8) x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
    Memory: 248.60 MB / 16.00 GB
    Shell: 5.3.3 - /usr/local/bin/bash
  Binaries:
    Node: 22.18.0 - ~/.nvm/versions/node/v22.18.0/bin/node
    npm: 10.9.3 - ~/.nvm/versions/node/v22.18.0/bin/npm
    pnpm: 10.16.1 - ~/.nvm/versions/node/v22.18.0/bin/pnpm
    Watchman: 2025.08.25.00 - /usr/local/bin/watchman
  Browsers:
    Chrome: 140.0.7339.133
    Safari: 18.6
  npmPackages:
    @vitejs/plugin-react: catalog: => 5.0.2
    @vitest/browser: ^4.0.0-beta.11 => 4.0.0-beta.11
    @vitest/coverage-v8: ^4.0.0-beta.11 => 4.0.0-beta.11
    @vitest/ui: ^4.0.0-beta.11 => 4.0.0-beta.11
    playwright: ^1.55.0 => 1.55.0
    vite: catalog: => 6.3.6
    vitest: ^4.0.0-beta.11 => 4.0.0-beta.11
    vitest-browser-react: ^2.0.0-beta.3 => 2.0.0-beta.3
```

### Used Package Manager

pnpm

### Validations

zeakd:
### Description

When `browser.screenshotDirectory` is explicitly set in config, `resolveConfig` pre-resolves it to an absolute path ([resolveConfig.ts#L812-L817](https://github.com/vitest-dev/vitest/blob/main/packages/vitest/src/node/config/resolveConfig.ts#L812-L817)). The `toMatchScreenshot` path resolver in `screenshotMatcher/utils.ts` then calls `join(root, absolutePath)`, which concatenates the paths literally — unlike `resolve`, `join` does not recognize when the second argument is already absolute.

This produces malformed screenshot paths like:
```
src/__tests__/Users/username/project/__screenshots__/test.tsx/screenshot.png
```

The fix replaces `join` with `resolve` on [utils.ts line 139](https://github.com/vitest-dev/vitest/blob/main/packages/browser/src/node/commands/screenshotMatcher/utils.ts#L139), which correctly handles both relative (default fallback) and absolute (pre-resolved config) paths.

The bug only triggers when `screenshotDirectory` is explicitly set. The default fallback (`?? '__screenshots__'`) produces a relative string, so `join` works correctly in that case.

Resolves #8604

### Tests

### Documentation
- N/A (no new functionality introduced)

### Changesets
- fix: use `resolve` instead of `join` for `screenshotDirectory` path resolution in `toMatchScreenshot`

### DIFF
--- a/packages/browser/src/node/commands/screenshotMatcher/utils.ts
+++ b/packages/browser/src/node/commands/screenshotMatcher/utils.ts
@@ -136,7 +136,7 @@ export function resolveOptions(
     root,
     screenshotDirectory: relative(
       root,
-      join(root, context.project.config.browser.screenshotDirectory ?? '__screenshots__'),
+      resolve(root, context.project.config.browser.screenshotDirectory ?? '__screenshots__'),
     ),
     attachmentsDir: relative(root, context.project.config.attachmentsDir),
     testFileDirectory: relative(root, dirname(context.testPath)),
--- a/test/browser/specs/to-match-screenshot.test.ts
+++ b/test/browser/specs/to-match-screenshot.test.ts
@@ -107,7 +107,7 @@ describe.runIf(provider.name === 'playwright')('screenshotDirectory', () => {
       expect(referencePath).not.toContain('custom-screenshots/Users')
       expect(referencePath).not.toContain('custom-screenshots/home')
 
-      expect(fs.existsFile(referencePath)).toBe(true)
+      expect(() => fs.statFile(referencePath)).not.toThrow()
 
       fs.editFile(testFilename, content => `${content}\n`)
 
--- a/test/browser/specs/to-match-screenshot.test.ts
+++ b/test/browser/specs/to-match-screenshot.test.ts
@@ -55,6 +55,70 @@ async function runBrowserTests(
   })
 }
 
+async function runBrowserTestsWithScreenshotDir(
+  structure: TestFsStructure,
+  screenshotDir: string,
+  config: ViteUserConfig['test'] = {},
+) {
+  return runInlineTests({
+    ...structure,
+    'vitest.config.js': `
+      import { playwright } from '@vitest/browser-playwright'
+      export default {
+        test: {
+          browser: {
+            enabled: true,
+            screenshotFailures: false,
+            screenshotDirectory: ${JSON.stringify(screenshotDir)},
+            provider: playwright(),
+            headless: true,
+            instances: [{ browser: ${JSON.stringify(browser)} }],
+          },
+          reporters: ['verbose'],
+          ...${JSON.stringify(config)},
+        },
+      }`,
+  }, {
+    $cliOptions: {
+      watch: true,
+    },
+  })
+}
+
+describe.runIf(provider.name === 'playwright')('screenshotDirectory', () => {
+  test(
+    'stores screenshots in the configured directory when screenshotDirectory is explicitly set',
+    async () => {
+      const { fs, stderr, vitest } = await runBrowserTestsWithScreenshotDir(
+        {
+          [testFilename]: testContent,
+          'utils.ts': utilsContent,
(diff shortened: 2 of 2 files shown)


Zelys-DFKH:
The fix is correct. When `screenshotDirectory` is explicitly set, `resolveConfig` pre-resolves it to an absolute path, so `pathe.join(root, absolutePath)` appends it as a literal segment rather than treating it as the resolved location. Swapping to `resolve` is exactly the right call.

One thing worth flagging on the test: the `not.toContain('custom-screenshots/Users')` and `not.toContain('custom-screenshots/home')` assertions catch the bug on GitHub Actions (runners live under `/home/runner/` or `/Users/runner/`), but miss it on systems where the workspace is under `/tmp/`, `/workspace/`, or similar. The `statFile` check doesn't close that gap either. With the bug, the screenshot gets written at the wrong absolute path, and the path extracted from stderr is that same wrong path, so both sides align and `statFile` succeeds regardless.

The most reliable way to test this: pass an absolute `screenshotDirectory` in the test config and assert that the reference path starts with it.

```typescript
import { tmpdir } from 'node:os'
import { join } from 'pathe'

const screenshotDir = join(tmpdir(), 'vitest-test-screenshots')
// ...
const [referencePath] = extractToMatchScreenshotPaths(stderr, testName)

// fix: resolve(root, absoluteDir) === absoluteDir → starts here
// bug: join(root, absoluteDir) === root + absoluteDir → starts with root instead
expect(referencePath.startsWith(screenshotDir + '/')).toBe(true)
```

(Cleanup of `screenshotDir` after the test is needed since it lives outside the inline test root.)

One more: #8604 is already closed and describes a different issue — custom `resolveScreenshotPath` creating empty folders. A new dedicated issue for this specific `join`/`resolve` collision would give reviewers cleaner context.

hi-ogawa:
superseded by https://github.com/vitest-dev/vitest/pull/8605

Zelys-DFKH:
Just wanted to flag that PR #8605 (cited as superseding this) fixes a different bug entirely: it prevents unnecessary directory creation when `options.save` is false. It doesn't touch `utils.ts` and doesn't address the path duplication this PR was fixing. The root cause is still present in main.

Your fix was correct. I opened PR #10332 with the same approach, using `resolve(root, screenshotDirectory)` instead of `join`.

PR #9861, opened 2026-03-13
fix: detect fixture that returns without calling `use` (#9831)

### LINKED ISSUE #9831: Improve error message when fixture doesn't call `use` before timeout
### Describe the bug

Had difficult time debugging why a test is timing out. Root cause was that I was returning from fixture when specific condition is met, without calling `use`. Let's improve the error message in these cases by saying something like `Fixture returned value but didn't call 'use'. Hook will timeout.`

```ts
const test = base.extend({
  value: undefined as string | undefined,

  setup: [
    async ({ value }, use) => {
      // optional value, no setups if it's missing
      if (!value) {
        return; // Forgot to call use here 🙃 
        // return await use(undefined)
      }

      await prepare(value);
      await use(undefined);
    },
    { auto: true },
  ],
});

describe("working example", () => {
  test.scoped({ value: "something" })
  test('example 1', () => {});
})

test('example 2', () => {});
```

### Reproduction

https://stackblitz.com/~/edit/vitest-dev-vitest-inffnahb?file=basic.test.ts:L1-L35&initialPath=/__vitest__/

### System Info

```shell
Stackblitz, Vitest 4.0.18
```

### Used Package Manager

npm

### Validations

oilater:
### Description

Resolves #9831

When a fixture function returns without calling `use()`, Vitest currently waits for a timeout and shows a generic `Test timed out` error, making it difficult to debug the root cause.

This PR detects the case immediately and rejects with a descriptive error:

```bash
Error: Fixture returned without calling "use". 
Make sure to call "use" in every code path of the fixture function.
```

**Implementation**                                         
- Added `.then().catch()` chain in `resolveFixtureFunction` — `.then()` detects when the fixture resolves without `use()` being called, `.catch()` handles errors thrown inside the fixture (existing behavior). 

스크린샷 2026-03-14 오전 2 23 18

### Tests

### Documentation
- [ ] If you introduce new functionality, document it. You can run documentation with `pnpm run docs` command. No documentation needed — this only improves an error message.

### Changesets


AriPerkkio (dismissed):
Looks good to me, thanks for the fix! 💯

oilater:
> Looks good to me, thanks for the fix! 💯

Thank you so much! 😊

sheremet-va (dismissed):
This looks good to me, but the error doesn't have a proper stack trace. It would be nice to point to the fixture declaration, but I don't know how to calculate the error trace 🤔 Does `Error.captureStackTrace` on the fixture function works?

I would expect something like

```
Error: Fixture returned without calling "use". Make sure to call "use" in every code path of the fixture function.

    async ({ value }, use) => {
    ^
```

oilater:
> This looks good to me, but the error doesn't have a proper stack trace. It would be nice to point to the fixture declaration, but I don't know how to calculate the error trace 🤔 Does `Error.captureStackTrace` on the fixture function works?
> 
> I would expect something like
> 
> ```
> Error: Fixture returned without calling "use". Make sure to call "use" in every code path of the fixture function.
> 
>     async ({ value }, use) => {
>     ^
> ```

Thanks for the feedback!
I implemented the stack trace using the same Symbol-based pattern as `hooks.ts`

Since the fixture function definition isn't a call frame, pointing to the exact line isn't possible at runtime 🤔 
but it can point to the base.extend(...) call site like below.

스크린샷 2026-03-15 오전 5 20 18

How about also including the fixture name in the error message like this?

```ts
Error: Fixture "setup" returned without calling "use". Make sure to call "use" in every code path of the fixture function.
```

Let me know how you'd like to proceed, and I'll update this PR!

packages/runner/src/fixture.ts
```diff
@@ -500,6 +500,15 @@ async function resolveFixtureFunction(
       await fixtureReturn
     })
     await useReturnPromise
+  }).then(() => {
+    // fixture returned without calling use()
+    if (!isUseFnArgResolved) {
+      useFnArgPromise.reject(
+        new Error(
+          'Fixture returned without calling "use". Make sure to call "use" in every code path of the fixture function.',
```
hi-ogawa: Can it surface fixture name? For the test case, it would be `setup`.

EDIT: ah sorry, you've already suggested that https://github.com/vitest-dev/vitest/pull/9861#issuecomment-4061358426.

hi-ogawa (changes requested):

test/cli/test/__snapshots__/fails.test.ts.snap
```diff
@@ -91,6 +91,8 @@ exports[`should fail test-extend/fixture-rest-props.test.ts 1`] = `"FixtureParse
 
 exports[`should fail test-extend/fixture-without-destructuring.test.ts 1`] = `"FixtureParseError: The 1st argument inside a fixture must use object destructuring pattern, e.g. ({ task } => {}). Instead, received "context"."`;
 
+exports[`should fail test-extend/fixture-without-use.test.ts 1`] = `"Error: Fixture returned without calling "use". Make sure to call "use" in every code path of the fixture function."`;
```
hi-ogawa: I think we should favor `runInlineTests + stderr snapshot` pattern, so we can view full error output (even if we don't have stack right now, it might in the future). Probably we can put it somewhere around https://github.com/vitest-dev/vitest/blob/8fc9cff0c7cbb527a0981ddc89f176a70219ace8/test/cli/test/scoped-fixtures.test.ts#L22-L54

oilater:
Updated in https://github.com/vitest-dev/vitest/pull/9861/changes/09b3f1ca5b41c9de4f80413f1c8e127f13c33647 !

- Added stack trace pointing to base.extend(...) call site
- Included fixture name in the error message
- Moved test to runInlineTests + stderr inline snapshot pattern in `scoped-fixtures.test.ts`

스크린샷 2026-03-16 오후 12 56 01

packages/runner/src/fixture.ts
oilater: Sure, Done in 09b3f1c!

hi-ogawa (approved):
lgtm!

sheremet-va (approved):

sheremet-va:
Thanks!

PR #9879, opened 2026-03-16
docs: add Unhandled Promise Rejection section to common errors guide

Srasti527:
Added a new section explaining the "Unhandled Promise Rejection" error
in the Common Errors guide.

This clarifies that the error originates from JavaScript/Node.js when
a rejected Promise does not have a handler attached.

### DIFF
--- a/docs/guide/common-errors.md
+++ b/docs/guide/common-errors.md
@@ -121,29 +121,27 @@ export default defineConfig({
 vitest --pool=forks
 ```
 :::
+
 ## Unhandled Promise Rejection
 
-This error happens when a Promise is rejected but no error handler
-(`.catch` or `await`) is attached before the event loop finishes.
+This error happens when a Promise is rejected but no error handler (`.catch` or `await`) is attached before the event loop finishes.
 
 This behavior comes from JavaScript itself and is not specific to Vitest.
 
 Learn more in the [Node.js documentation](https://nodejs.org/api/process.html#event-unhandledrejection).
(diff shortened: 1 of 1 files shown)


Srasti527:
Hi! I added a small documentation section explaining the
Unhandled Promise Rejection error in the Common Errors guide.

Please let me know if any changes are required.
Thank you!

docs/guide/common-errors.md
```diff
@@ -121,3 +121,11 @@ export default defineConfig({
...
 ```
 :::
+## Unhandled Promise Rejection
+
+This error happens when a Promise is rejected but no error handler
+(`.catch` or `await`) is attached before the event loop finishes.
+
+This behavior comes from JavaScript itself and is not specific to Vitest.
+
+Learn more in the [Node.js documentation](https://nodejs.org/api/process.html#event-unhandledrejection).
```
hi-ogawa: I think It'd be nice to go more verbose with actual example test code and actual errors. Then can also present how `.catch(() => {}` could be used etc.

hi-ogawa (changes requested):

docs/guide/common-errors.md
Srasti527: Thanks for the suggestion! I have added example test code showing unhandled vs handled promise rejections and how .catch() can be used. Please let me know if any further changes are needed.

sheremet-va (approved):

PR #9884, opened 2026-03-17
docs: add example & fix for Unhandled Promise Rejection in common errors guide

Srasti527:
Added a section in the Common Errors guide explaining the Unhandled Promise Rejection error in JavaScript/Node.js.

Includes example test code showing unhandled vs handled promise rejections.

Shows how .catch(() => {}) can be used to handle errors.

Linked to the official [Node.js documentation](https://nodejs.org/api/process.html#event-unhandledrejection)
 for reference.

This helps readers understand why the error occurs and how to fix it in Vitest tests.


sheremet-va:
Merged in https://github.com/vitest-dev/vitest/pull/9879

PR #9894, opened 2026-03-17
feat(expect): warn on toEqual comparing non-plain objects with empty Object.keys

### LINKED ISSUE #9542: Show warning if `toEqual` compares the object which Object.keys returns empty array for

I was using vitest for test for React Router application.
I noticed that `Object.keys(Response)` returns `[]`, thus it doesn't check anything but whether both objects are response.

This means, this test passes:
```ts
test("hoge", () => {
  const r1 = new Response("hoge", { status: 200 });
  const r2 = new Response(null, { status: 404 });
  expect(r1).toEqual(r2);
});
```
Which is clearly a false-negative.

### Suggested solution

It would be difficult to prepare matchers for all objects, but it would be helpful if vitest shows warning if nothing is compared (except plain empty object: `{}`).

Simple PoC:
```ts
import { test, expect } from "vitest";

const isPlainObject = (obj: unknown) => {
  return obj.constructor === Object || Object.getPrototypeOf(obj) === null;
};
expect.addEqualityTesters([
  (a: unknown, b: unknown) => {
    if (
      Object.keys(a).length === 0 &&
      !isPlainObject(a) &&
      Object.keys(b).length === 0 &&
      !isPlainObject(b)
    ) {
      console.log("Warning: Comparing two empty non-plain objects, causing false-positive results.");
    }
    return undefined;
  },
]);

test("hoge", () => {
  const r1 = new Response("hoge", { status: 200 });
  const r2 = new Response(null, { status: 200 });
  expect(r1).toEqual(r2);
});
```

### Alternative

_No response_

### Additional context

related: https://github.com/vitest-dev/vitest/issues/9522, https://github.com/vitest-dev/vitest/issues/7991

### Validations

ViniciusDev26:
### Description
                                                                                                                                        
  Adds `isPlainObject` and `isNonPlainEmptyObject` helpers to `jest-utils` and emits a `console.warn` in `toEqual` when the actual value  
  is a non-plain object with no enumerable keys, preventing silent false-negatives (e.g. `Response`).                                     
                                                                                                                                          
  **Example of the false-negative this addresses:**
  ```ts
  // This test passes silently today, but should not:
  const r1 = new Response("hoge", { status: 200 })
  const r2 = new Response(null, { status: 404 })
  expect(r1).toEqual(r2) // ⚠ no properties compared                                                                                      
   ```
Resolves #9542    

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/api/expect.md
+++ b/docs/api/expect.md
@@ -563,6 +563,10 @@ expect(new Error('hi')).toEqual(new Error('hi', { cause: 'x' }))
 To test if something was thrown, use [`toThrow`](#tothrow) assertion.
 :::
 
+:::warning
+Some built-in objects (e.g. `Response`, `Request`) expose their data through getters instead of enumerable keys. Since `toEqual` relies on `Object.keys()`, these properties are not compared, which may cause false-negatives. Vitest emits a `console.warn` when this is detected. Consider comparing the relevant properties directly in that case.
+:::
+
 ## toStrictEqual
 
 - **Type:** `(received: any) => Awaitable<void>`
--- a/packages/expect/src/jest-expect.ts
+++ b/packages/expect/src/jest-expect.ts
@@ -17,6 +17,7 @@ import {
   generateToBeMessage,
   getObjectSubset,
   isError,
+  isNonPlainEmptyObject,
   iterableEquality,
   equals as jestEquals,
   sparseArrayEquality,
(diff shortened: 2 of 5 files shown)


sheremet-va (changes requested):
The issue has a comment on the board: 

> have a builtin response/request check, check what node does in assert

We want to process Response/Request/Headers inside the `equals` function. And we need to validate how the native `assert` in Node works with those already

We don't want to add the empty check guard

hi-ogawa:
closing as stale

PR #9897, opened 2026-03-17
feat(reporter): add --junitIncludeConsoleOutput CLI flag

### LINKED ISSUE #9812: CLI support for JUnit reporter includeConsoleOutput option

As a developer using Vitest, I want to control the JUnit reporter’s includeConsoleOutput option from the CLI, so that I can enable or disable <system-out> and <system-err> output in CI without changing or branching my Vitest config.

Today, outputFile has dedicated CLI support, but includeConsoleOutput does not. The JUnit reporter supports this option in config, but it cannot be overridden from the CLI the way outputFile can.

This matters in CI because some environments want smaller JUnit XML files and some want full console output for debugging. Right now, the only way to switch this is through config-level reporter options, which is awkward for matrix jobs and one-off CI overrides.

### Suggested solution

Add CLI support for the JUnit reporter’s includeConsoleOutput option.

A couple of possible shapes that would fit current CLI behavior:

A dedicated top-level flag, for example:
`--junitIncludeConsoleOutput=false` or `--includeConsoleOutput=false`

A reporter-scoped flag, if Vitest wants to support more reporter-specific CLI overrides in the future, for example:
`--reporterOption.junit.includeConsoleOutput=false`

My preference would be the smallest change that is consistent with existing conventions. A dedicated boolean flag seems simplest, especially since outputFile already has special CLI handling.

### Alternative

Otherwise, we have to do some ugly config myself. For instance, I'm doing this nonsense:

CLI command:
```sh
npx cross-env CI=true VITEST_JUNIT_ONLY=true vitest run --outputFile=junit-core.xml src/standard
```

`vite.config.ts`:
```ts
const isJunitOnlyRun = process.env.VITEST_JUNIT_ONLY === "true";

export default defineConfig({
  // ...
  test: {
    reporters: isJunitOnlyRun
      ? [["junit", { includeConsoleOutput: false }]]
      : ["default", ["junit", { includeConsoleOutput: false }]],
```

### Additional context

_No response_

### Validations

mixelburg:
## Summary

Closes #9812

The JUnit reporter already supports an `includeConsoleOutput` option in its config, but there was no way to toggle it from the CLI. This PR adds a dedicated `--junitIncludeConsoleOutput` flag, following the same convention as the existing `--outputFile` flag.

## Usage

```sh
# Disable console output capture in the JUnit report:
vitest --reporter=junit --junitIncludeConsoleOutput=false
vitest --reporter=junit --no-junitIncludeConsoleOutput

# Explicitly enable (this is the default):
vitest --reporter=junit --junitIncludeConsoleOutput
```

## Changes

- `packages/vitest/src/node/types/config.ts` — adds `junitIncludeConsoleOutput?: boolean` to `InlineConfig`
- `packages/vitest/src/node/cli/cli-config.ts` — registers the CLI option
- `packages/vitest/src/node/config/resolveConfig.ts` — propagates the flag into the JUnit reporter's options (using `??=` so an explicit reporter-level config still takes priority)
- `docs/guide/cli-generated.md` — adds the documentation entry

### DIFF
--- a/docs/guide/cli-generated.md
+++ b/docs/guide/cli-generated.md
@@ -111,6 +111,13 @@ Specify reporters (default, agent, blob, verbose, dot, json, tap, tap-flat, juni
 
 Write test results to a file when supporter reporter is also specified, use cac's dot notation for individual outputs of multiple reporters (example: `--outputFile.tap=./tap.txt`)
 
+### junitIncludeConsoleOutput
+
+- **CLI:** `--junitIncludeConsoleOutput`
+- **Config:** [junitIncludeConsoleOutput](/config/#junitincludeconsoleoutput)
+
+Include console output for each test in the JUnit reporter
+
 ### coverage.provider
 
 - **CLI:** `--coverage.provider <name>`
--- a/docs/guide/cli-generated.md
+++ b/docs/guide/cli-generated.md
@@ -114,7 +114,7 @@ Write test results to a file when supporter reporter is also specified, use cac'
 ### junitIncludeConsoleOutput
 
 - **CLI:** `--junitIncludeConsoleOutput`
-- **Config:** [junitIncludeConsoleOutput](/config/#junitincludeconsoleoutput)
+- **Config:** [junitIncludeConsoleOutput](/config/junitincludeconsoleoutput)
 
 Include console output for each test in the JUnit reporter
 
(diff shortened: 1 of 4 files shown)


AriPerkkio (commented):
Instead of adding reporter specific options on root level, we should really add support for passing any reporter options via cli.

Something like:

```sh
$ vitest --reporter.junit.includeConsoleOutput
$ vitest --reporter.junit.include-console-output

$ vitest --reporter.some-custom-reporter.some-option some-value
```

PR #9911, opened 2026-03-19
fix: open browser in `--standalone` mode without running tests

### LINKED ISSUE #9901: Allow idle start for Browser Mode

We want to adopt the Vitest Browser Mode, but as far as I can tell it is not possible to start it in some kind of "idle" mode during development. Whenever I start the Browser Mode (not headless; using watch mode) it automatically starts **all** tests.

Regular Vitest tests seem to have the `--standalone` option exactly for this use case. (So maybe this can also be considered as a bug, that browser mode doesn't respect the `--standalone` setting?)

### Suggested solution

Allow using `--standalone` in browser mode. When run non-headless the UI opens, but not test will automatically run - only on manual interaction.

### Alternative

Introduce a dedicated flag like `--browser.idle`, if we don't want to reuse `--standalone`.

### Additional context

Other testing tools that provide a UI (usually E2E testing tools like Cypress and Playwright) have similar behaviour. They open in a view where the developer can choose to run all tests or some tests manually. I'd like to support a similar workflow when I onboard other devs to vitest.

### Validations

sheremet-va:
### Description

Resolves #9901

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/core.ts
+++ b/packages/vitest/src/node/core.ts
@@ -795,10 +795,18 @@ export class Vitest {
   }
 
   /**
+   * @deprecated use `standalone()` instead
+   */
+  init(): Promise<void> {
+    this.logger.deprecate('`vitest.init()` is deprecated. Use `vitest.standalone()` instead.')
+    return this.standalone()
+  }
+
+  /**
    * Initialize reporters and the coverage provider. This method doesn't run any tests.
    * If the `--watch` flag is provided, Vitest will still run changed tests even if this method was not called.
    */
-  async init(): Promise<void> {
+  async standalone(): Promise<void> {
     await this._traces.$('vitest.init', async () => {
       try {
         await this.initCoverageProvider()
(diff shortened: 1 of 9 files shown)


packages/vitest/src/node/core.ts
```diff
@@ -794,11 +794,19 @@ export class Vitest {
...
+    this.logger.deprecate('`vitest.init()` is deprecated. Use `vitest.standalone()` instead.')
+    return this.standalone()
+  }
+
   /**
    * Initialize reporters and the coverage provider. This method doesn't run any tests.
    * If the `--watch` flag is provided, Vitest will still run changed tests even if this method was not called.
    */
-  async init(): Promise<void> {
+  async standalone(): Promise<void> {
```
AriPerkkio, line 809: Reporter's documentation about `onInit` also needs changes https://vitest.dev/api/advanced/reporters.html#oninit

> Internally this method is called inside `vitest.start`, `vitest.init` or `vitest.mergeReports`

AriPerkkio (approved):

hi-ogawa (approved):

PR #9913, opened 2026-03-19
feat(experimental): expose `matchesTagsFilter` to test if the current filter matches tags

sheremet-va:
### Description

Allows to check dynamically if current tags filter is applied:

```ts
import { TestRunner } from 'vitest'

beforeAll(() => {
  if (TestRunner.matchesTagsFilter(['flaky'])) {
    // ...
  }
})
```

### DIFF
--- a/docs/guide/test-tags.md
+++ b/docs/guide/test-tags.md
@@ -309,6 +309,7 @@ You can use `TestRunner.matchesTagsFilter` (since Vitest 4.1.1) to check whether
 import { beforeAll, TestRunner } from 'vitest'
 
 beforeAll(async () => {
+  // Seed database when "vitest --tags-filter db" is used
   if (TestRunner.matchesTagsFilter(['db'])) {
     await seedDatabase()
   }
--- a/docs/guide/test-tags.md
+++ b/docs/guide/test-tags.md
@@ -300,3 +300,19 @@ You can also pass multiple `--tags-filter` flags. They are combined with AND log
 # Run tests that match (unit OR e2e) AND are NOT slow
 vitest --tags-filter="unit || e2e" --tags-filter="!slow"
 ```
+
+### Checking Tags Filter at Runtime
+
+You can use `TestRunner.matchesTagsFilter` (since Vitest 4.1.1) to check whether the current tags filter matches a set of tags. This is useful for conditionally running expensive setup logic only when relevant tests are included:
+
+```ts
+import { beforeAll, TestRunner } from 'vitest'
+
+beforeAll(async () => {
+  if (TestRunner.matchesTagsFilter(['db'])) {
+    await seedDatabase()
+  }
+})
+```
+
+The method accepts an array of tags and returns `true` if the current `--tags-filter` would include a test with those tags. If no tags filter is active, it always returns `true`.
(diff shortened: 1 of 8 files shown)


docs/guide/test-tags.md
```diff
@@ -300,3 +300,19 @@ You can also pass multiple `--tags-filter` flags. They are combined with AND log
 # Run tests that match (unit OR e2e) AND are NOT slow
 vitest --tags-filter="unit || e2e" --tags-filter="!slow"
 ```
+
+### Checking Tags Filter at Runtime
+
+You can use `TestRunner.matchesTagsFilter` (since Vitest 4.1.1) to check whether the current tags filter matches a set of tags. This is useful for conditionally running expensive setup logic only when relevant tests are included:
```
AriPerkkio, line 306: Feature in patch release 🧐

docs/guide/test-tags.md
```diff
@@ -300,3 +300,19 @@ You can also pass multiple `--tags-filter` flags. They are combined with AND log
...
+
+```ts
+import { beforeAll, TestRunner } from 'vitest'
+
+beforeAll(async () => {
+  if (TestRunner.matchesTagsFilter(['db'])) {
+    await seedDatabase()
+  }
+})
```
AriPerkkio, line 316: So `TestRunner.matchesTagsFilter` compares against all tags in the test module? I feel like this should be mentioned. Otherwise someone (me) will try to use it in `beforeEach` to filter test cases.

sheremet-va: What do you mean? It matches against tags that you passed down. What are tags in the test module?

sheremet-va: If you pass down `--tags-filter=filter`, it will match the array that you pass against _that_

> The method accepts an array of tags and returns `true` if the current `--tags-filter` would include a test with those tags.

docs/guide/test-tags.md
sheremet-va: `@experimental`

docs/guide/test-tags.md
AriPerkkio: Ohh I was thinking of something like:

`setup.ts`:

```ts
// For each test file that has db tag, run db inits
beforeAll(async () => {
  if (TestRunner.matchesTagsFilter(['db'])) {
    await seedDatabase()
  }
})
```

`file1.test.ts`

```ts
// Does not init db for this file
test("nope", { tags: ["nope-dont-run"] }, () => {});
```

`file2.test.ts`

```ts
// Does init db for this file
test("yep", { tags: ["db", "yep-runs-here"] }, () => {});
```

AriPerkkio: Makes sense, it does mention `TagsFilter`.

sheremet-va: No, the function only cares about tags that you pass down to that function. This is a way to check if certain tags are running (`vitest --tags-filter=db`). `TestRunner.matchesTagsFilter(['db'])` will check if `--tags-filter=` has `db` (could be `db && flaky` or anything else)

I thought the last sentence is enough to explain it. How would you improve it to make it more clear?

AriPerkkio: The last sentence does indeed make it clear. 

If more attention is wanted, then something like:

```diff
beforeAll(async () => {
+ // Seed database when "vitest --tags-filter db" is used, or when no filter is used at all
  if (TestRunner.matchesTagsFilter(['db'])) {
    await seedDatabase()
  }
```

AriPerkkio (approved):

PR #9919, opened 2026-03-19
fix(mocker): lazily import createFilter to respect optional vite peer dep

### LINKED ISSUE #9915: @vitest/mocker 4.1.0: vite declared as optional peer dependency but imported unconditionally in dist/node.js
### Describe the bug

Starting with `@vitest/mocker` 4.1.0, running vitest with `@vitest/browser-playwright` fails with `ERR_MODULE_NOT_FOUND` if `vite` is not explicitly installed in the project, even though `vite` is declared as an optional peer dependency.

**Expected:** Tests run (or a graceful error if vite is required)

**Actual:**
```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'vite' imported from /path/to/node_modules/@vitest/browser-playwright/node_modules/@vitest/mocker/dist/node.js
```

## Root cause

In `@vitest/mocker/package.json`, vite is declared as optional:

```json
"peerDependencies": {
  "vite": "^6.0.0 || ^7.0.0 || ^8.0.0-0"
},
"peerDependenciesMeta": {
  "vite": {
    "optional": true
  }
}
```

However, `dist/node.js` appears to have an unconditional static import of `vite`, causing Node.js module resolution to fail immediately at load time when vite is not present.

This worked in 4.0.18

## Workaround

Explicitly install `vite` as a devDependency:

```bash
npm install vite --save-dev
```

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-bjtmmvai?file=package.json&initialPath=__vitest__/

1. Create a project with the following devDependencies:
   ```json
   {
     "devDependencies": {
       "@vitest/browser-playwright": "4.1.0",
       "vitest": "4.1.0"
     }
   }
   ```
2. Run `npm install`
3. Run `vitest`

### System Info

```shell
System:
    OS: macOS 26.3.1
    CPU: (12) arm64 Apple M4 Pro
    Memory: 8.55 GB / 48.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.22.1 - /usr/local/bin/node
    npm: 10.9.4 - /usr/local/bin/npm
  Browsers:
    Chrome: 146.0.7680.153
    Safari: 26.3.1
  npmPackages:
    @vitest/browser-playwright: 4.1 => 4.1.0 
    vitest: 4.1 => 4.1.0
```

### Used Package Manager

npm

### Validations

Hassad674:
Fixes #9915

The static `import { createFilter } from 'vite'` in `hoistMocksPlugin.ts` causes `@vitest/mocker/node` to fail with `ERR_MODULE_NOT_FOUND` when vite is not installed, even though vite is declared as an optional peer dependency.

The fix moves the `createFilter` import into a dynamic `await import('vite')` inside the `buildStart` hook. Since `buildStart` always runs before `transform` in both Vite and Rollup, the filter is guaranteed to be initialized before any file is processed. When no custom filter is provided and vite isn't available, the plugin gracefully accepts all files.

All other vite imports in the mocker package are already `import type` (erased at compile time) — this was the only runtime value import.

### DIFF
--- a/packages/mocker/src/node/hoistMocksPlugin.ts
+++ b/packages/mocker/src/node/hoistMocksPlugin.ts
@@ -1,7 +1,6 @@
 import type { SourceMap } from 'magic-string'
 import type { Plugin, Rollup } from 'vite'
 import type { HoistMocksOptions } from './hoistMocks'
-import { createFilter } from 'vite'
 import { cleanUrl } from '../utils'
 import { hoistMocks } from './hoistMocks'
 
@@ -15,7 +14,7 @@ export interface HoistMocksPluginOptions extends Omit<HoistMocksOptions, 'regexp
 }
 
 export function hoistMocksPlugin(options: HoistMocksPluginOptions = {}): Plugin {
-  const filter = options.filter || createFilter(options.include, options.exclude)
+  let filter: ((id: string) => boolean) | undefined = options.filter
 
   const {
     hoistableMockMethodNames = ['mock', 'unmock'],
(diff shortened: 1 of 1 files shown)


sheremet-va:
If someone uses this plugin, then they should have `vite` installed. Vite is optional for _some_ of the entry points, but there are many of them in this package.

PR #9936, opened 2026-03-21
fix: cache transform results in `getTestDependencies` across projects (fix #9855)

### LINKED ISSUE #9855: vitest related with multiple projects corrupts Vue SFC compilation via shared plugin cache
### Describe the bug

When running `vitest related <file>` in a workspace with multiple projects (e.g., a jsdom project and a browser/Playwright project), the `getTestDependencies` function SSR-transforms Vue SFC files for dependency analysis across ALL projects. This SSR transform poisons `@vitejs/plugin-vue`'s module-level `descriptorCache` singleton, causing subsequent browser/client transforms to produce incorrect output.

Specifically, for `<script setup>` components, **template-only component imports are omitted from the setup function's `__returned__` object**, making them `undefined` in the render function's `$setup` proxy. This causes `[Vue warn]: Invalid vnode type when creating vnode: undefined` and components fail to render.

### Reproduction

**Minimal reproduction:** https://github.com/seanogdev/vue-sfc-cache-poisoning-repro

```bash
pnpm install
pnpm test    # runs: vitest related src/Target.vue --run
```

The bug only triggers via `vitest related` (which calls `getTestDependencies` across all projects). Running `vitest run` directly passes because it skips the multi-project dependency resolution that causes the cache poisoning.

**Environment:**
- Vitest workspace with 2 projects:
  - Project A: jsdom environment (app unit tests)
  - Project B: browser/Playwright environment (Storybook component tests via `@storybook/addon-vitest`)
- Both projects share a common Vite config with `@vitejs/plugin-vue`

**Vue component structure:**
```vue
<script setup lang="ts">
import ChildComponent from './ChildComponent.vue'; // Used ONLY in template
const someRef = ref(0); // Used in script
</script>
<template>
  <ChildComponent />
  <span>{{ someRef }}</span>
</template>
```

**Steps:**
1. `vitest related Parent.vue` → **FAILS** - `ChildComponent` is undefined in render
2. `vitest --project my-storybook-project Parent.vue` → **PASSES**
3. `vitest Parent.vue` (name filter, both projects) → **PASSES**

### Root Cause Analysis

The issue is in the interaction between `vitest related`'s dependency analysis and `@vitejs/plugin-vue`'s caching:

1. **`getTestDependencies`** (in `packages/vitest/src/node/specifications.ts`) calls `project.vite.environments.ssr.transformRequest(filepath)` for each project to build module dependency graphs.

2. **`@vitejs/plugin-vue`'s `descriptorCache`** is a module-level singleton `Map<string, SFCDescriptor>` keyed by filename only — no environment/SSR discrimination. When the SSR transform runs for the jsdom project, it creates/mutates a descriptor that's shared with the browser project.

3. **When `@vue/compiler-sfc`'s `compileScript` runs for SSR**, it processes the `<script setup>` block differently. Template-only component imports may be excluded from `__returned__` because SSR rendering handles component resolution differently.

4. **The subsequent client/browser transform** reuses the poisoned descriptor from the shared cache, producing a setup function where `__returned__` omits template-only imports:

   **Failing (vitest related):**
   ```js
   const __returned__ = { props, modelValue, someRef }
   // ChildComponent is MISSING
   ```

   **Passing (single project):**
   ```js
   const __returned__ = { props, modelValue, someRef, ChildComponent }
   ```

5. The compiled render function accesses `$setup["ChildComponent"]` which returns `undefined`, triggering the "Invalid vnode type" warning.

### Evidence

Confirmed by instrumenting the component's render function at runtime:

```
// Failing run ($setup keys via for..in):
["localVar1", "localVar2", "props", "modelValue", ...]
// ChildComponent and other template-only imports ABSENT

// Passing run ($setup keys):  
["localVar1", "localVar2", "props", "modelValue", ..., "ChildComponent"]
// All imports present
```

The raw import at the module level resolves correctly in both cases (`typeof ChildComponent === 'object'`), but the `$setup` proxy used by the compiled render function does not include it.

### Expected behavior

`vitest related` should produce identical test results to running the same tests via `vitest --project <name>` or `vitest <name-filter>`.

### Possible fixes

1. **Vitest**: `getTestDependencies` could avoid using `environments.ssr.transformRequest` for browser-mode projects, or use an isolated transform pipeline for dependency analysis
2. **`@vitejs/plugin-vue`**: The `descriptorCache` should be keyed by `filename + environment` to prevent cross-environment pollution
3. **Vite Environment API**: Plugin state isolation between environments should be enforced

### System Info

```
Vitest: 4.0.18
@vitejs/plugin-vue: 6.0.4
Vue: 3.5.30
Vite: 7.3.1
OS: macOS (Darwin 25.3.0)
```

### Used Package Manager

pnpm

### Validations

- [x] Check that you are using the latest version of Vitest
- [x] Read the [docs](https://vitest.dev/)
- [x] Check that there isn't already an issue that reports the same bug

seanogdev:
## Summary

When running `vitest related` with multiple projects, `getTestDependencies` calls `transformRequest` on shared files through each project's Vite server independently. This triggers a bug in `@vue/compiler-sfc` where module-level LRU caches (`parseCache` and `templateAnalysisCache`) get desynced across servers, causing compiled output to silently drop template-only component imports.

**Reproduction:** https://github.com/seanogdev/vue-sfc-cache-poisoning-repro

## The problem

`@vue/compiler-sfc` has three interacting issues:

1. `compiler.parse()` returns cached mutable descriptor objects via an LRU(500), shared across all Vite servers in the process
2. `compileTemplate()` mutates `descriptor.template.ast` in place, turning v-if elements (type=1) into IfNodes (type=9)
3. `resolveTemplateAnalysisResult()` only walks `node.type === 1`, so when it re-walks a mutated AST after cache eviction it misses components inside v-if/v-for

In a workspace with a jsdom project and a browser/Playwright project sharing `@vitejs/plugin-vue`, the redundant `transformRequest` calls from `getTestDependencies` cause these caches to desync. The result: `<script setup>` components used only in `<template>` get excluded from `__returned__` and become `undefined` at runtime ("Invalid vnode type" errors).

Running `vitest related` with both projects caused browser test failures. Running `vitest related --project <storybook-only>` passed cleanly, confirming the cross-project interaction.

## The potential fix

A shared `Map<string, string[]>` caches each file's resolved import paths across all specs within a single `filterTestsBySource` call. The cache only stores dependency lists (import paths from `deps`/`dynamicDeps`), not transformed code. This means each file is only transformed once during dependency analysis, preventing the redundant `transformRequest` calls that trigger the upstream cache desync.

## Test plan

- Regression test in `test/cli/test/vue-sfc-cache-poisoning.test.ts` verifies the `compiler-sfc` cache poisoning mechanism directly using the compiler API
- Full pipeline test reproduces the exact desync with multiple Vite servers and LRU eviction
- The end-to-end browser failure was verified against the [reproduction repo](https://github.com/seanogdev/vue-sfc-cache-poisoning-repro)

Closes #9855

### DIFF
--- a/packages/vitest/src/node/specifications.ts
+++ b/packages/vitest/src/node/specifications.ts
@@ -150,9 +150,10 @@ export class VitestSpecifications {
       return []
     }
 
+    const transformCache = new Map<string, string[]>()
     const testGraphs = await Promise.all(
       specs.map(async (spec) => {
-        const deps = await this.getTestDependencies(spec)
+        const deps = await this.getTestDependencies(spec, transformCache)
         return [spec, deps] as const
       }),
     )
@@ -169,19 +170,31 @@ export class VitestSpecifications {
     return runningTests
   }
 
-  private async getTestDependencies(spec: TestSpecification, deps = new Set<string>()): Promise<Set<string>> {
+  private async getTestDependencies(
+    spec: TestSpecification,
+    transformCache?: Map<string, string[]>,
+    deps = new Set<string>(),
+  ): Promise<Set<string>> {
     const addImports = async (project: TestProject, filepath: string) => {
       if (deps.has(filepath)) {
         return
       }
       deps.add(filepath)
 
-      const mod = project.vite.environments.ssr.moduleGraph.getModuleById(filepath)
-      const transformed = mod?.transformResult || await project.vite.environments.ssr.transformRequest(filepath)
-      if (!transformed) {
-        return
+      let dependencies: string[]
+      const cached = transformCache?.get(filepath)
+      if (cached) {
+        dependencies = cached
       }
-      const dependencies = [...transformed.deps || [], ...transformed.dynamicDeps || []]
+      else {
+        const mod = project.vite.environments.ssr.moduleGraph.getModuleById(filepath)
+        const transformed = mod?.transformResult || await project.vite.environments.ssr.transformRequest(filepath)
+        if (!transformed) {
+          return
+        }
+        dependencies = [...transformed.deps || [], ...transformed.dynamicDeps || []]
+        transformCache?.set(filepath, dependencies)
+      }
       await Promise.all(dependencies.map(async (dep) => {
         const fsPath = dep.startsWith('/@fs/')
           ? dep.slice(isWindows ? 5 : 4)
--- a/packages/vitest/src/node/specifications.ts
+++ b/packages/vitest/src/node/specifications.ts
@@ -182,8 +182,7 @@ export class VitestSpecifications {
       deps.add(filepath)
 
       let dependencies: string[]
-      const cacheKey = `${project.name}:${filepath}`
-      const cached = transformCache?.get(cacheKey)
+      const cached = transformCache?.get(filepath)
       if (cached) {
         dependencies = cached
       }
@@ -194,7 +193,7 @@ export class VitestSpecifications {
           return
         }
         dependencies = [...transformed.deps || [], ...transformed.dynamicDeps || []]
-        transformCache?.set(cacheKey, dependencies)
+        transformCache?.set(filepath, dependencies)
       }
       await Promise.all(dependencies.map(async (dep) => {
         const fsPath = dep.startsWith('/@fs/')
--- a/packages/vitest/src/node/specifications.ts
+++ b/packages/vitest/src/node/specifications.ts
@@ -182,7 +182,8 @@ export class VitestSpecifications {
       deps.add(filepath)
 
       let dependencies: string[]
-      const cached = transformCache?.get(filepath)
+      const cacheKey = `${project.name}:${filepath}`
+      const cached = transformCache?.get(cacheKey)
       if (cached) {
         dependencies = cached
       }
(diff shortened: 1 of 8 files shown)


packages/vitest/src/node/specifications.ts
```diff
@@ -169,19 +170,31 @@ export class VitestSpecifications {
...
         return
       }
       deps.add(filepath)
 
-      const mod = project.vite.environments.ssr.moduleGraph.getModuleById(filepath)
-      const transformed = mod?.transformResult || await project.vite.environments.ssr.transformRequest(filepath)
-      if (!transformed) {
-        return
+      let dependencies: string[]
+      const cached = transformCache?.get(filepath)
```
AriPerkkio: Doesn't this cause issues if two Vite plugins transform the same file differently? For example if your Node tests dropped some browser specific module imports, the browser tests would get the cached version and see incorrect test dependencies.

seanogdev: The cache just stores import paths, not transformed code, so different plugin output shouldnt matter here.  Issue *is* probably upstream. But I don't see a harm with this either.

sheremet-va: The transformed code _can_ have different dependencies though depending on the plugin

seanogdev:
Closing — the real fix belongs upstream in `@vitejs/plugin-vue` (see vitejs/vite-plugin-vue#764 / vitejs/vite-plugin-vue#765). Caching transform results in Vitest would only paper over the plugin-vue descriptor poisoning.

PR #9940, opened 2026-03-21
fix: coverageConfigDefaults

Arthie:
### Description

Was reading the coverage options code and here are some small typing suggestions:
- CoverageConfigDefaults should not use the resolved config type and values, adjusted shared coverage defaults to use the base coverage shape
- Move deprecated BaseCoverageOptions usage to CoverageOptions

### Tests

### Documentation

### Changesets


AriPerkkio (commented):
Can you resolve conflicts?

Arthie:
@AriPerkkio conflicts are resolved! I didn't see there was a PR already addressing some of the changes so updated the PR title and Description! Moved the deprecated BaseCoverageOptions usage to CoverageOptions. Thanks!

packages/vitest/src/defaults.ts
```diff
@@ -25,7 +24,7 @@ export const benchmarkConfigDefaults: Required<
 }
 
 // These are the generic defaults for coverage. Providers may also set some provider specific defaults.
-export const coverageConfigDefaults: ResolvedCoverageOptions = {
+export const coverageConfigDefaults: Required<Omit<CoverageOptions, 'include' | 'skipFull' | 'thresholds' | 'watermarks' | 'ignoreClassMethods' | 'htmlDir' | 'changed' | 'customProviderModule'>> = {
```
AriPerkkio: Can you reason about this change? Why shouldn't default values match with the shape of resolved ones?

AriPerkkio (commented):
Not sure about changing `coverageConfigDefaults` types. Otherwise looks good.

packages/vitest/src/defaults.ts
Arthie: Since `coverageConfigDefaults` is available externally, I was under the impression that it should represent the raw/default config values and not the fully normalized/resolved runtime config values. If this is the wrong impression let me know.


#### External usage
```
import { coverageConfigDefaults } from 'vitest/config';
// OR
import { coverageConfigDefaults } from "vite-plus";
```

#### We have to typecast it to `as CoverageOptions` on line 119 in `configDefaults`. This skips some typescript checks.
```
coverage: coverageConfigDefaults as CoverageV8Options, // line 119
                    ^--- ResolvedCoverageOptions
```

#### In the docs the default value for `coverage.reporter` doesn't match with the actual value. The expected value is not possible with `ResolvedCoverageOptions` since it's more restrictive.
```
   // expeted value according to the docs
   reporter: [
    'text',
    'html',
    'clover',
    'json',
  ],
  // actual value 
  reporter: [
    ['text', {}],
    ['html', {}],
    ['clover', {}],
    ['json', {}],
  ],
```


####  `coverageConfigDefaults` now aligns with the code style of `benchmarkConfigDefaults` above it.
```
export const benchmarkConfigDefaults: Required<
  Omit<BenchmarkUserOptions, 'outputFile' | 'compare' | 'outputJson'>
>
```

AriPerkkio: Yeah that makes sense. Also by not exposing the resolved types it allows us to use something completely different type internally in future, if ever needed.

I think we should use here `FieldsWithDefaultValues` so that we don't create two separate source of truth that can be out-of-sync:

https://github.com/vitest-dev/vitest/blob/36a6fd8334323e5a883ab5130de243f1a5d0f99b/packages/vitest/src/node/types/coverage.ts#L98-L108

We should also define default values for following properties here:

```
include
ignoreClassMethods
changed
skipFull
thresholds
watermarks
```

Arthie: Ok just tried updating the type with `FieldsWithDefaultValues` and adding the suggested properties.

```ts
/** Fields that have default values. Internally these will always be defined. */
export type FieldsWithDefaultValues // added export
  = | 'provider'
    | 'enabled'
    | 'clean'
    | 'cleanOnRerun'
    | 'reportsDirectory'
    | 'exclude'
    | 'reportOnFailure'
    | 'allowExternal'
    | 'processingConcurrency'
// added fields below
    | 'reporter' // required for no ts errors in packages/vitest/src/node/coverage.ts
    | 'excludeAfterRemap' // required for no ts errors in packages/vitest/src/node/coverage.ts
    | 'include'
    | 'ignoreClassMethods'
    | 'changed'
    | 'skipFull'
    // | 'thresholds' // default value is undefined?
    | 'watermarks'
```

and 

```ts
// These are the generic defaults for coverage. Providers may also set some provider specific defaults.
export const coverageConfigDefaults: Required<Pick<CoverageOptions, FieldsWithDefaultValues>> = {
  provider: 'v8',
  enabled: false,
  clean: true,
  cleanOnRerun: true,
  reportsDirectory: './coverage',
  exclude: [],
  reportOnFailure: false,
  reporter: [
    'text',
    'html',
    'clover',
    'json',
  ],
  allowExternal: false,
  excludeAfterRemap: false,
  processingConcurrency: Math.min(
    20,
    os.availableParallelism?.() ?? os.cpus().length,
  ),
  include: [], // Default: Files that were imported during test run
  ignoreClassMethods: [],
  changed: false,
  skipFull: false,
  // thresholds: undefined,
  watermarks: {
    statements: [50, 80],
    functions: [50, 80],
    branches: [50, 80],
    lines: [50, 80],
  },
}
```

The issues / changes:
- adding export to `FieldsWithDefaultValues`
- `'reporter'` and `'excludeAfterRemap'` need to be also defined otherwise we get tsc errors in packages/vitest/src/node/coverage.ts
- `include` has as default 'Files that were imported during test run' is setting `include: []` correct in coverageConfigDefaults?
- `thresholds` seems to have undefined as default? So it should not be part of `FieldsWithDefaultValues`?

Not sure about `include` and `thresholds` fields

AriPerkkio: > `'reporter'` and `'excludeAfterRemap'` need to be also defined otherwise we get tsc errors in `packages/vitest/src/node/coverage.ts`

Sounds correct.

Looks like `incude` and `thresholds` should actually not have default values. So let's leave them out of `FieldsWithDefaultValues`. These checks should not pass when user does not define values there:

https://github.com/vitest-dev/vitest/blob/faace1fbe09133fa3641164c1d58538b316a38ee/packages/coverage-v8/src/provider.ts#L77-L79

https://github.com/vitest-dev/vitest/blob/faace1fbe09133fa3641164c1d58538b316a38ee/packages/coverage-v8/src/provider.ts#L136-L138

Otherwise looks good.

Arthie: Yes leaving out `include` and `thresholds` is also what I was thinking.

We also have to leave out `changed` since its default value is actually `undefined` and not `false` (Inherits the default value from `test.changed`). I've also fixed the `changed` comment and removed the `@default: false`.

Think this PR is good to go!

AriPerkkio (approved):
Looks good, thanks!

PR #9967, opened 2026-03-25
feat: support merge reports for non-sharded multi-environment runs

### LINKED ISSUE #9954: Support merging reports for non-sharded multi-environment runs

[Draft plan with Claude](https://gist.github.com/hi-ogawa/0d11864e5dbf54d152135bf8ce61ea06)

As a developer using Vitest in CI with a matrix strategy (e.g., linux/macos/windows, or node 20/22), I want to merge test reports from multiple platform runs into a single report, so that I can get a unified view of test results across all environments.

Currently `--merge-reports` only supports merging sharded runs — where test files are split across shards with no overlap. When the same tests are run on multiple platforms (a common CI pattern), merging those blob reports produces duplicate unlabeled entries with no way to tell which result came from which platform.

Playwright supports this via their [`merge-reports`](https://playwright.dev/docs/test-sharding#merging-reports-from-multiple-shards) + [`tag` config](https://playwright.dev/docs/api/class-testconfig#test-config-tag), and it's a frequently needed workflow.

### Suggested solution

> [!NOTE]
> Reworked solution is in https://github.com/vitest-dev/vitest/pull/9967

Use explicit tags to distinguish blobs from different environments. The blob reporter would accept a `tag` option that gets stored in the blob metadata and injected into each test's `tags` array during merge. This builds on Vitest's existing tag infrastructure (filtering, reporter display).

**Blob reporter config:**
```typescript
// vitest.config.ts
export default defineConfig({
  test: {
    reporter: [['blob', {
      tag: process.env.CI_PLATFORM, // e.g., "linux", "windows", "node-22"
    }]],
  },
})
```

**Merge behavior:**
- Each blob carries its tag in metadata
- On merge, the tag is prepended to every test's `tags` array (e.g., `["linux", ...]`)
- Same test file from different blobs becomes distinct entries, each labeled by platform
- Reporters already understand tags, so results are immediately distinguishable

**Example CI workflow:**
```yaml
jobs:
  test:
    strategy:
      matrix:
        os: [ubuntu-latest, windows-latest, macos-latest]
    steps:
      - run: vitest run --reporter=blob
        env:
          CI_PLATFORM: ${{ matrix.os }}
        with:
          name: blob-${{ matrix.os }}
          path: .vitest-reports/

  merge:
    needs: test
    steps:
        with:
          path: .vitest-reports/
          merge-multiple: true
      - run: vitest --merge-reports --reporter=default --reporter=json
```

### Alternative

- Auto-detect platform (`process.platform`, `process.arch`) and inject as tags without explicit config. Simpler for users but less flexible — doesn't cover non-OS dimensions like Node version or custom environments.
- Playwright's approach: automatic ID deconfliction (salt duplicate IDs) + separate explicit `tag` config for labeling. We could start with the simpler combined approach (tag = deconflict + label) and add automatic deconfliction later if needed.

### Additional context

_No response_

### Validations

hi-ogawa:
### Description

- Closes https://github.com/vitest-dev/vitest/issues/9954

### Example

Set `VITEST_BLOB_LABEL` environment variable (or `[["blob", { label }]]` in config) when creating blobs:

```bash
VITEST_BLOB_LABEL=linux vitest run --reporter=blob
VITEST_BLOB_LABEL=macos vitest run --reporter=blob
```

```ts
// Or in config
export default defineConfig({
  test: {
    reporters: [['blob', { label: 'linux' }]],
  },
})
```

The reporter during `--merge-reports` can show the test run results from different blob label as if they are from different projects:

```sh
$ vitest run --merge-reports
...
 ✓  linux  src/basic.test.ts (2 tests)
 ✓  macos  src/basic.test.ts (2 tests)
```

### How it works

The label is a **blob reporter option** — it does not affect the test run itself.

**Write side:** The blob reporter stores the label in the blob payload. The label also determines the default output filename (`blob-linux.json`). Project names are not mutated during the test run.

**Read side:** `readBlobs()` discovers labels from parsed blob content, clones `TestProject` instances per label (same pattern as browser mode), then rewrites `file.projectName`, `File.id`, child task IDs, and `environmentModules` keys before module graph restoration. The merge pipeline downstream sees properly-discriminated projects with no collisions.

TODO
- [x] test
- [x] dog-food
- [ ] update https://github.com/vitest-tests/test-sharding

🎦 Examples from dog-fooding

- `test/cli`

image

- `test/core`

image

### Tests

### Documentation

### Changesets


test/cli/test/workers-option.test.ts
```diff
@@ -1,28 +1,5 @@
 import * as testUtils from '#test-utils'
-import { describe, expect, test, vi } from 'vitest'
-import { getWorkersCountByPercentage } from 'vitest/src/utils/workers.js'
-
-vi.mock(import('node:os'), async importOriginal => ({
```
hi-ogawa, line 5: This unit test is evacuated to `test/core` since `vi.mock` inside `test/cli` is breaking something.

sheremet-va (changes requested):
I am confused by the implementation. Wouldn't it make more sense to change the project's name _when running_ tests so they are collected with new ids instead of patching them after?

hi-ogawa:
> I am confused by the implementation. Wouldn't it make more sense to change the project's name _when running_ tests so they are collected with new ids instead of patching them after?

I basically prototype both but I preferred current version since Blob reporter option influencing test run seems wrong and merge-time patch didn't look too far fetched. 
For example, if user test code checks project name during test run to do something specifically, then that code needs to take into account blob labeling. The same can be true for patching-in project name for merge time reporter though. The choice is just based on my initial trade-off instinct. We should discuss to figure out which is better.

sheremet-va:
> We should discuss to figure out which is better.

In my opinion, it's better to change the name when running tests to reduce the complexity. I would like to avoid regenerating and recalculating hashes.

As an option, we could add a static field that influences the hash (like file.hashKey or something). I would also be useful for typechecking (which injects `__typecheck__` into a file id right now)

hi-ogawa:
Based on the discussion, I'm going to explore the approach to encode "blob label" into `File.id` and `File.meta` instead of differentiating the test files by artificially created "label project". The reporter makes use of `File.meta` to differentiate visually.

PR #9976, opened 2026-03-26
feat(coverage): v8 to track `node:child_process` and `node:worker_threads` contexts

### LINKED ISSUE #7064: coverage not collected from sub-propcesses result from child_process APIs
## Describe the bug

Consider a package which is a CLI tool.
Consider a test suite that does the testing-trophy: it runs e2e with few happy paths, generates coverage, and then write unit-tests to cover only the parts that are not covered by the e2e and component tests.
Now consider that the e2e test cases run the CLI using any of the APIs of `child_process`: `exec`,  `spawn`, `fork`, or their sync equivalents.

The tool fails to collect coverage from the e2e suites.
The reproduction does only `exec`, because it is minimal. however, tried all of them.

I'd be delighted if it's just some setting that I'm failing to pass...

### scenario
1. given a project (full structure below) that uses `child_process` to run the bin of the package in e2e test of CLI tools, and unit tests to cover code-paths that do not appear in the e2e coverage
  
2. Run `vitest run --coverage` 

###  expected

100% coverage. for all files.

###  Found
All tests run successfully, but only coverage collected using unit-tests is observed.
```
 RUN  v2.1.8 /home/user/ws/opensource/reproduction/vitest/child_process-not-covered
      Coverage enabled with v8

 ✓ test/calc.unit.mjs (1)
   ✓ calc (1)
     ✓ sub (1)
       ✓ should substract positive integers
 ✓ test/cli.e2e.mjs (2)
   ✓ fxgen command (2)
     ✓ when called with valid parameters (2)
       ✓ should not fail
       ✓ should return the right answer

 Test Files  2 passed (2)
      Tests  3 passed (3)
   Start at  17:13:40
   Duration  291ms (transform 18ms, setup 0ms, collect 16ms, tests 32ms, environment 1ms, prepare 128ms)

 % Coverage report from v8
-----------|---------|----------|---------|---------|-------------------
File       | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
-----------|---------|----------|---------|---------|-------------------
All files  |   14.28 |    33.33 |      25 |   14.28 |
 bin       |       0 |        0 |       0 |       0 |
  cli.mjs  |       0 |        0 |       0 |       0 | 1-3
 lib       |   18.18 |       50 |   33.33 |   18.18 |
  calc.mjs |     100 |      100 |      50 |     100 |
  cli.mjs  |       0 |        0 |       0 |       0 | 1-9
-----------|---------|----------|---------|---------|-------------------
```

## Project structure:
image

`vitest.config.js`
```
export default defineConfig({
  test: {
    globals: true,
    reporters: ['verbose'],
    include: [
      'test/**/*.unit.mjs',
      'test/**/*.e2e.mjs',
    ],
  },
});
```

`package.json`
```
{
  "name": "child_process-not-covered",
  "version": "1.0.0",
  "bin": {
    "calc": "bin/calc.mjs"
  },
  "scripts": {
    "test": "vitest run --coverage"
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "description": "",
  "devDependencies": {
    "@vitest/coverage-v8": "^2.1.8",
    "vitest": "^2.1.8"
  }
}
```

`bin/cli`
```
#!/usr/bin/env node
import cli from '../lib/cli.mjs';
cli({ process, console });
```

`lib/cli`
```
import * as calc from './calc.mjs';
export default ({ process, console }) => {
  const { argv: [ ,, op, a, b ] } = process;
  const result = calc[op](Number(a), Number(b));
  console.log(result);
}
```

`lib/calc`
```
export const add = (a, b) => a + b;
export const sub = (a, b) => a - b;
```

`test/calc.unit.mjs`
```
import * as calc from '../lib/calc.mjs';

describe('calc', () => {
  describe('sub', () => {
    it('should substract positive integers', () => {
      expect(calc.sub(44,2)).toEqual(42);
    });
  });
});
```

`test/cli.e2e.mjs`
```
import { execSync } from 'node:child_process';

describe('fxgen command', () => {
  describe('when called with two integers', () => {
    const ctx = {};
    beforeAll(() => {
      try { 
        ctx.result = execSync('node bin/cli.mjs add 20 22').toString().trim();
      } catch(err) {
        ctx.err = err;
      }
    });

    it('should not fail', () => expect(ctx.err).toBeFalsy());
    it('should return their sum', () => expect(ctx.result).toEqual('42'));
  });
});
```

## System Info

```shell
$ vitest --version
vitest/2.1.8 linux-x64 node-v20.18.0
```

## Used Package Manager

npm

## Validations

## disclaimer
Feels like de j'avoux.
I encountered this before while trying to migrating a `moch+nyc` project, and that effort turned away from vitest.
If I opened such a ticket before - excuse me, for some reason I cannot find it :( 
I'm now working with a new customer on a new codebase - new chance :) 

Anyway, now I really put the effort to reproduce the but in isolation :)

AriPerkkio:
### Description

- Fixes https://github.com/vitest-dev/vitest/issues/7064
- Fixes https://github.com/vitest-dev/vitest/discussions/3851
- Fixes https://github.com/vitest-dev/vitest/discussions/9842
- Also related to https://github.com/vitest-dev/vitest/issues/4899

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/config/coverage.md
+++ b/docs/config/coverage.md
@@ -459,12 +459,12 @@ Note that setting this option does not change where coverage HTML report is gene
 
 Collect coverage only for files changed since a specified commit or branch. When set to `true`, it uses staged and unstaged changes.
 
-## coverage.autoAttachWorkers <Version>5.0.0</Version> {#coverage-autoattachworkers}
+## coverage.autoAttachSubprocess <Version>5.0.0</Version> {#coverage-autoattachsubprocess}
 
 - **Type:** `boolean`
 - **Default:** `false`
 - **Available for providers:** `'v8'`
-- **CLI:** `--coverage.autoAttachWorkers`
+- **CLI:** `--coverage.autoAttachSubprocess`
 
 Track coverage of the `node:child_process` and `node:worker_threads` spawned during test run.
 
--- a/docs/config/coverage.md
+++ b/docs/config/coverage.md
@@ -459,7 +459,7 @@ Note that setting this option does not change where coverage HTML report is gene
 
 Collect coverage only for files changed since a specified commit or branch. When set to `true`, it uses staged and unstaged changes.
 
-## coverage.autoAttachWorkers <Version type="experimental">4.1.5</Version> {#coverage-autoattachworkers}
+## coverage.autoAttachWorkers <Version>5.0.0</Version> {#coverage-autoattachworkers}
 
 - **Type:** `boolean`
 - **Default:** `false`
--- a/docs/config/coverage.md
+++ b/docs/config/coverage.md
@@ -458,3 +458,14 @@ Note that setting this option does not change where coverage HTML report is gene
 - **CLI:** `--coverage.changed`, `--coverage.changed=<commit/branch>`
 
 Collect coverage only for files changed since a specified commit or branch. When set to `true`, it uses staged and unstaged changes.
+
+## coverage.trackProcessAndWorker
+
+- **Type:** `boolean`
+- **Default:** `false`
+- **Available for providers:** `'v8'`
+- **CLI:** `--coverage.trackProcessAndWorker`
+
+Track coverage of the `node:child_process` and `node:worker_threads` spawned during test run.
+
+Note that this option has some performance overhead as its using [`NODE_V8_COVERAGE`](https://nodejs.org/api/cli.html#node-v8-coveragedir) internally. This triggers Node to write lots of unnecessary files on file system.
--- a/docs/config/coverage.md
+++ b/docs/config/coverage.md
@@ -459,12 +459,12 @@ Note that setting this option does not change where coverage HTML report is gene
 
 Collect coverage only for files changed since a specified commit or branch. When set to `true`, it uses staged and unstaged changes.
 
-## coverage.trackProcessAndWorker
+## coverage.autoAttachWorkers <Version type="experimental">4.1.5</Version> {#coverage-autoattachworkers}
 
 - **Type:** `boolean`
 - **Default:** `false`
 - **Available for providers:** `'v8'`
-- **CLI:** `--coverage.trackProcessAndWorker`
+- **CLI:** `--coverage.autoAttachWorkers`
 
 Track coverage of the `node:child_process` and `node:worker_threads` spawned during test run.
 
--- a/packages/coverage-v8/src/index.ts
+++ b/packages/coverage-v8/src/index.ts
@@ -3,7 +3,7 @@ import type { CoverageProviderModule } from 'vitest/node'
 import type { ScriptCoverageWithOffset, V8CoverageProvider } from './provider'
 import { randomUUID } from 'node:crypto'
 import { existsSync } from 'node:fs'
-import { readdir, readFile } from 'node:fs/promises'
+import { readdir, readFile, rm } from 'node:fs/promises'
 import inspector from 'node:inspector/promises'
 import { resolve } from 'node:path'
 import { fileURLToPath } from 'node:url'
@@ -22,16 +22,16 @@ const mod: CoverageProviderModule & {
   session: null,
 
   async startCoverage({ isolate, autoAttachSubprocess, reportsDirectory }) {
-    if (autoAttachSubprocess) {
-      this.extendedContextCoverageDir = resolve(reportsDirectory, 'tmp', randomUUID())
-      process.env.NODE_V8_COVERAGE = this.extendedContextCoverageDir
-    }
-
     if (isolate === false && enabled) {
       return
     }
 
     enabled = true
+
+    if (autoAttachSubprocess) {
+      this.extendedContextCoverageDir = resolve(reportsDirectory, 'tmp', randomUUID())
+      process.env.NODE_V8_COVERAGE = this.extendedContextCoverageDir
+    }
 
     this.session ||= new inspector.Session()
     const session = this.session as inspector.Session
@@ -69,7 +69,14 @@ const mod: CoverageProviderModule & {
       const contents = await Promise.all(
         filenames
           .filter(filename => filename.endsWith('.json'))
-          .map(filename => readFile(`${this.extendedContextCoverageDir}/${filename}`, 'utf8')),
+          .map(async (filename) => {
+            const path = `${this.extendedContextCoverageDir}/${filename}`
+
+            const content = await readFile(path, 'utf8')
+            await rm(path)
+
+            return content
+          }),
       )
 
       for (const content of contents) {
--- a/packages/coverage-v8/src/index.ts
+++ b/packages/coverage-v8/src/index.ts
@@ -22,16 +22,16 @@ const mod: CoverageProviderModule & {
   session: null,
 
   async startCoverage({ isolate, autoAttachWorkers, reportsDirectory }) {
+    if (autoAttachWorkers) {
+      this.extendedContextCoverageDir = resolve(reportsDirectory, 'tmp', randomUUID())
+      process.env.NODE_V8_COVERAGE = this.extendedContextCoverageDir
+    }
+
     if (isolate === false && enabled) {
       return
     }
 
     enabled = true
-
-    if (autoAttachWorkers) {
-      this.extendedContextCoverageDir = resolve(reportsDirectory, 'tmp', randomUUID())
-      process.env.NODE_V8_COVERAGE = this.extendedContextCoverageDir
-    }
 
     this.session ||= new inspector.Session()
     const session = this.session as inspector.Session
--- a/packages/coverage-v8/src/index.ts
+++ b/packages/coverage-v8/src/index.ts
@@ -20,14 +20,14 @@ const mod: CoverageProviderModule & {
   extendedContextCoverageDir: undefined,
   session: null,
 
-  async startCoverage({ isolate, trackProcessAndWorker, reportsDirectory }) {
+  async startCoverage({ isolate, autoAttachWorkers, reportsDirectory }) {
     if (isolate === false && enabled) {
       return
     }
 
     enabled = true
 
-    if (trackProcessAndWorker) {
+    if (autoAttachWorkers) {
       this.extendedContextCoverageDir = resolve(reportsDirectory, 'tmp', randomUUID())
       process.env.NODE_V8_COVERAGE = this.extendedContextCoverageDir
     }
--- a/packages/coverage-v8/src/index.ts
+++ b/packages/coverage-v8/src/index.ts
@@ -21,8 +21,8 @@ const mod: CoverageProviderModule & {
   extendedContextCoverageDir: undefined,
   session: null,
 
-  async startCoverage({ isolate, autoAttachWorkers, reportsDirectory }) {
-    if (autoAttachWorkers) {
+  async startCoverage({ isolate, autoAttachSubprocess, reportsDirectory }) {
+    if (autoAttachSubprocess) {
       this.extendedContextCoverageDir = resolve(reportsDirectory, 'tmp', randomUUID())
       process.env.NODE_V8_COVERAGE = this.extendedContextCoverageDir
     }
--- a/packages/coverage-v8/src/index.ts
+++ b/packages/coverage-v8/src/index.ts
@@ -2,6 +2,7 @@ import type { Profiler } from 'node:inspector'
 import type { CoverageProviderModule } from 'vitest/node'
 import type { ScriptCoverageWithOffset, V8CoverageProvider } from './provider'
 import { randomUUID } from 'node:crypto'
+import { existsSync } from 'node:fs'
 import { readdir, readFile } from 'node:fs/promises'
 import inspector from 'node:inspector/promises'
 import { resolve } from 'node:path'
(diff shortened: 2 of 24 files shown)


packages/coverage-v8/src/intercept-new-run-context.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    reportedCoverage = true
+
+    const coverage = await provider.takeCoverage?.({
+    // Start offset should be 0 as these run outside of Vite
+      moduleExecutionInfo: undefined,
+    }) as { result: ScriptCoverageWithOffset[] }
+
+    // TODO: Now where do we get source maps
+    ws.send(JSON.stringify(coverage.result))
```
AriPerkkio: Next todo to solve 🙃

AriPerkkio: image

test/coverage-test/test/extended-run-context.v8.test.ts
```diff
@@ -0,0 +1,159 @@
...
+    expect(files).toMatchInlineSnapshot(`
+      [
+        "<process-cwd>/fixtures/src/math.ts",
+        "<process-cwd>/fixtures/src/start-fork-and-thread.ts",
+        "<process-cwd>/fixtures/src/worker-or-process.nested.ts",
+      ]
+    `)
+
+    assertMath(coverageMap)
+  })
```
AriPerkkio: Last failing test. Source maps from nested child_process/worker_thread are off.

packages/coverage-v8/src/index.ts
```diff
@@ -3,21 +3,41 @@ import type { CoverageProviderModule } from 'vitest/node'
...
+      const port = await getPort()
+      this.wss = new WebSocketServer({ port })
+
+      this.wss.on('connection', socket => socket.on('message', (raw) => {
+        const result: ScriptCoverageWithOffset[] = JSON.parse(raw.toString())
+        this.extendedContextCoverage.push(...(result || []))
+      }))
+
+      process.env.NODE_OPTIONS ||= ''
+      process.env.NODE_OPTIONS += ' --import @vitest/coverage-v8/intercept-new-run-context'
```
AriPerkkio: Let's check if `NODE_V8_COVERAGE` could be used here instead. It would be added inside test runner thread/process and read in the same context. Though we will likely need to work-around https://github.com/nodejs/node/issues/46378 now 🤷‍♂️ 

The `NODE_V8_COVERAGE` has the ability to collect coverage even when user calls `process.exit()` as it's handled by lower level Node internals. It's also extremely verbose and can cause +70MB coverage files when used with Vitest - that's why we don't use it by default.

packages/coverage-v8/src/index.ts
```diff
@@ -1,23 +1,34 @@
...
+import { readdir, readFile, rm } from 'node:fs/promises'
 import inspector from 'node:inspector/promises'
+import { resolve } from 'node:path'
 import { fileURLToPath } from 'node:url'
 import { normalize } from 'pathe'
 import { provider } from 'std-env'
 import { loadProvider } from './load-provider'
 
-const session = new inspector.Session()
+let session: inspector.Session | null = null
```
AriPerkkio: This used to create unused `new Session()` on main thread too, oops.

docs/config/coverage.md
```diff
@@ -418,3 +418,14 @@ Note that setting this option does not change where coverage HTML report is gene
 - **CLI:** `--coverage.changed`, `--coverage.changed=<commit/branch>`
 
 Collect coverage only for files changed since a specified commit or branch. When set to `true`, it uses staged and unstaged changes.
+
+## coverage.trackProcessAndWorker
```
AriPerkkio: Naming this option is difficult. Right now it only covers `node:child_process` and `node:worker_threads`, but in future it's likely going to be used for web workers and similar on browser. @hi-ogawa & @sheremet-va - please propose better name here. 🙏 

Internally this feature also uses term `extendedContextCoverage`. 🙃

sheremet-va: I don't think `worker` really fits in here, both child process and a worker thread is a worker in my opinion, at least that's how we defined it for pools, maybe `trackProcessAndThreads` is more accurate.

I would vote for something like `includeSubprocesses`/`instrumentSubprocesses` - subprocess can refer to anything, not just a fork/thread/service worker

sheremet-va: In debugging world there is also `--follow-forks` mode (in C), in vscode it's `autoAttachChildProcesses` (although it also follows workers)

AriPerkkio: How about `autoAttachChildContext`? Or `autoAttachWorkers` and imply `Worker` is processes too. 🤔

docs/config/coverage.md
```diff
@@ -458,3 +458,14 @@ Note that setting this option does not change where coverage HTML report is gene
 - **CLI:** `--coverage.changed`, `--coverage.changed=<commit/branch>`
 
 Collect coverage only for files changed since a specified commit or branch. When set to `true`, it uses staged and unstaged changes.
+
+## coverage.autoAttachWorkers <Version type="experimental">4.1.5</Version> {#coverage-autoattachworkers}
```
AriPerkkio:
```suggestion
## coverage.autoAttachWorkers <Version>5.0.0</Version> {#coverage-autoattachworkers}
```

packages/coverage-v8/src/index.ts
```diff
@@ -28,16 +45,43 @@ const mod: CoverageProviderModule = {
...
-          startOffset: options?.moduleExecutionInfo?.get(normalize(fileURLToPath(entry.url)))?.startOffset || 0,
-        })
+        entry.startOffset = options?.moduleExecutionInfo?.get(normalize(fileURLToPath(entry.url)))?.startOffset || 0
+
+        result.push(entry)
+      }
+    }
+
+    if (this.extendedContextCoverageDir) {
+      const filenames = await readdir(this.extendedContextCoverageDir)
```
hi-ogawa, line 68: This directory doesn't exist when there are no children spawned and this happens:

```sh
$ pnpm -C examples/basic test --coverage --coverage.autoAttachWorkers

> @vitest/example-test@ test /home/hiroshi/code/others/vitest-wt1/examples/basic
> vitest --coverage --coverage.autoAttachWorkers


 DEV  v4.1.4 /home/hiroshi/code/others/vitest-wt1/examples/basic
      Coverage enabled with v8

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Unhandled Errors ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯

Vitest caught 2 unhandled errors during the test run.

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Test Run Error ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
Error: ENOENT: no such file or directory, scandir '/home/hiroshi/code/others/vitest-wt1/examples/basic/coverage/tmp/af48afe8-e4b0-4810-b7b2-e6dbea470389'
 ❯ readdir node:internal/fs/promises:956:18
 ❯ Object.takeCoverage ../../packages/coverage-v8/dist/index.js:50:22
 ❯ TestRunner.testRunner.onAfterRunFiles ../../packages/vitest/dist/chunks/index.DtMvdosS.js:173:20
 ❯ startTests ../../packages/runner/dist/chunk-artifact.js:3281:3
 ❯ ../../packages/vitest/dist/chunks/base.DuCI8aBZ.js:90:5
 ❯ run ../../packages/vitest/dist/chunks/base.DuCI8aBZ.js:81:2
 ❯ runBaseTests ../../packages/vitest/dist/chunks/base.DuCI8aBZ.js:200:2
 ❯ executeTests ../../packages/vitest/dist/chunks/init-forks.UV3ZQGQH.js:29:4
 ❯ execute ../../packages/vitest/dist/chunks/init.D98-gwRW.js:158:3
 ❯ process.onMessage ../../packages/vitest/dist/chunks/init.D98-gwRW.js:287:14
```

hi-ogawa, line 68: Also does this ensure the coverage json emitted by children to be collected only once? Looks like multiple `takeCoverage` calls (maybe isolate false) can read the same json and sent as coverage.

docs/config/coverage.md
sheremet-va: I feel like the common denominator here is `subprocess`, I wouldn't call any `subprocess` a worker, but every worker a `subprocess`

AriPerkkio: Though Worker is not a sub process 🤔

packages/coverage-v8/src/index.ts
AriPerkkio: Good catch, fixed!

AriPerkkio: Good point, we'll need to ensure new `this.extendedContextCoverageDir` is generated each time `startCoverage()` is called. 🤔

docs/config/coverage.md
sheremet-va: It's not a process in a sense of `node:child_process`, but it is a process in a general sense

AriPerkkio: No, it's not a process even in a general sense.

sheremet-va: Well, process is not a worker in a general sense

sheremet-va: Worker is already taken and it would be confusing to reuse the word within the Vitest repository. Subprocess to me sounds like the most general way to describe anything spawned by the user code

AriPerkkio: Worker is not a term that's used in general sense 🙃. But sure let's think in NodeJS context.

Ok, let's use `autoAttachSubprocess`. I don't care that much.

packages/coverage-v8/src/index.ts
hi-ogawa: Is resetting on `startCoverage` correct? The scenario I was thinking is that when a test worker runs multiple files, it goes like:
- startCoverage 
- takeCoverage (after 1st test file)
- takeCoverage (after 2nd test file) 
- ...
- stopCoverage

If children coverage is generated in the first test, that can doubly tracked by each `takeCoverage` unless children coverage is cleaned up in each `takeCoverage` time.

AriPerkkio: > when a test worker runs multiple files, it goes like

It goes like:
- startCoverage
- takeCoverage
- stopCoverage
- startCoverage
- takeCoverage
- stopCoverage

hi-ogawa: Doesn't this code mean running multiple `takeCoverage` between start/stop? `packages/vitest/src/runtime/runBaseTests.ts` 

```js
startCoverageInsideWorker
for each file
  startTests -> onAfterRunFiles -> takeCoverageInsideWorker
stopCoverageInsideWorker
```

hi-ogawa: I reproduced with `examples/basic` where `suite.test.ts` runs first and spawn children. The line counter marker shows `x 2` of child file when running

```js
pnpm -C examples/basic test --coverage --coverage.autoAttachSubprocess --isolate=false --maxWorkers=1
```

AriPerkkio: > Doesn't this code mean running multiple `takeCoverage` between start/stop? `packages/vitest/src/runtime/runBaseTests.ts` 

As far as I remember, we've always passed just a single file to the worker at once. Even when isolation is disabled. Maybe it has changed? 👀 

> I reproduced with examples/basic where suite.test.ts runs first and spawn children. The line counter marker shows x 2 of child file when running

Ok I think we'll need to add new `isolate: false, autoAttachSubprocess: true` test case too. 🤔

AriPerkkio: > As far as I remember, we've always passed just a single file to the worker at once. Even when isolation is disabled. Maybe it has changed? 👀

Oh wow, yes, it has changed. And I was the one who changed it in https://github.com/vitest-dev/vitest/pull/8705 as performance optimization trick. I don't think it was possible before the rewrite. 🙃

https://github.com/vitest-dev/vitest/blob/1ba7338c3c84ee13bc4a57efac002306ce0399a6/packages/vitest/src/node/pool.ts#L347-L348

https://github.com/vitest-dev/vitest/blob/1ba7338c3c84ee13bc4a57efac002306ce0399a6/packages/vitest/src/node/pool.ts#L423-L424

Let's remove each JSON right after reading it to make sure we don't end up reading them twice. Good catch as always @hi-ogawa 💯

hi-ogawa (approved):

sheremet-va (approved):

sheremet-va:
Awesome!

PR #9994, opened 2026-03-27
fix(coverage-v8): silence uncovered parse failures

### LINKED ISSUE #9972: Rolldown PARSE_ERROR on Typescript files importing virtual modules
### Describe the bug

When using `@vitest/coverage-v8`, files that import from virtual modules (e.g. `virtual:__federation__`, `virtual:pwa-register`, etc.) cause rolldown `PARSE_ERROR` during coverage map generation.

The error occurs in `V8CoverageProvider.getCoverageMapForUncoveredFiles`, which uses rolldown's `parseAstAsync` to parse all source files included in coverage. Since virtual modules don't exist on disk and rolldown cannot resolve them, parsing fails with:

```
Error [RolldownError]: Parse failed with 1 error:
Expected `,` or `)` but found `:`
```

The error message `Expected ',' or ')' but found ':'` suggests rolldown is falling back to parsing the `.ts` file as plain JavaScript, so the TypeScript type annotations (`:`) cause the parse failure.

The coverage provider does catch the error and logs `"Failed to parse file:///.../file.ts. Excluding it from coverage."`, but the full error stack trace is still printed to stderr, making test output noisy and causing non-zero exit codes in CI pipelines.

Example file that triggers this:

```ts
// any-file.ts
import {
  __federation_method_getRemote as getRemote,
  __federation_method_setRemote as setRemote,
} from 'virtual:__federation__';

export const loadRemote = async (name: string): Promise<unknown> => {
  const module = await getRemote(name, './index');
  return module;
};
```

This affects **any** virtual module import — not just federation. Any Vite plugin that exposes a `virtual:` module will trigger this if the importing file is included in coverage.

### Workaround

The only workaround is to manually exclude affected files from coverage:

```ts
// vitest.config.ts
coverage: {
  exclude: [
    'src/path/to/file-with-virtual-import.ts',
  ],
},
```

This is not ideal as it requires knowing which files import from virtual modules and maintaining the exclusion list manually.

### Expected behaviour

Files that fail to parse during `getCoverageMapForUncoveredFiles` should be silently excluded from coverage without printing a full error stack trace or causing a non-zero exit code.

Alternatively, rolldown should be configured to parse `.ts` files as TypeScript (not JavaScript) so that type annotations don't cause parse failures, and virtual module imports should be handled gracefully.

### Reproduction

1. Create a Vite project with any plugin that provides a virtual module (e.g. `vite-plugin-federation`, `vite-plugin-pwa`, or a custom plugin)
2. Import from the virtual module in a `.ts` file
3. Configure vitest with `@vitest/coverage-v8`
4. Run tests with coverage enabled

```ts
// vite.config.ts

export default defineConfig({
  plugins: [
    // Any plugin providing virtual modules
  ],
});
```

```ts
// vitest.config.ts

export default defineConfig({
  test: {
    coverage: {
      provider: 'v8',
      enabled: true,
      include: ['src/**/*.ts'],
    },
  },
});
```

```ts
// src/example.ts
import { something } from 'virtual:some-plugin';

export const doWork = (name: string): string => {
  return something(name);
};
```

Running `npx vitest run` will produce:

```
Error [RolldownError]: Parse failed with 1 error:
Expected `,` or `)` but found `:`
    at error (.../node_modules/rolldown/dist/shared/logs-D80CXhvg.mjs:147:24)
    at normalizeParseError (.../node_modules/rolldown/dist/parse-ast-index.mjs:25:9)
    ...
Failed to parse file:///path/to/src/example.ts. Excluding it from coverage.
```

### System Info

```shell
System:
  OS: Linux 6.6 Ubuntu 22.04.5 LTS 22.04.5 LTS (Jammy Jellyfish)
  CPU: (16) x64 AMD Ryzen 7 PRO 7840U w/ Radeon 780M Graphics
  Memory: 25.06 GB / 50.98 GB
  Shell: 5.1.16 - /bin/bash
Binaries:
  Node: 25.8.1
  npm: 11.11.0
npmPackages:
  @vitest/coverage-v8: ^4.1.1 => 4.1.1
  vitest: ^4.1.1 => 4.1.1
  vite: ^8.0.2 => 8.0.2
```

### Used Package Manager

npm

### Validations

raashish1601:
## Summary
- stop writing caught uncovered-file parse failures to stderr in `@vitest/coverage-v8`
- route the parse call through a small internal helper so the catch path can be regression-tested directly
- add a unit regression for the parse-failure exclusion path and keep the nearby virtual-files coverage test green

Fixes #9972

## Testing
- `corepack pnpm --dir packages/coverage-v8 run build`
- `corepack pnpm --dir test/coverage-test exec vitest run test/parse-failures.unit.test.ts test/mixed-versions-warning.unit.test.ts --project unit`
- `corepack pnpm --dir test/coverage-test exec vitest run test/virtual-files.test.ts --project v8`

### DIFF
--- a/packages/coverage-v8/src/provider.ts
+++ b/packages/coverage-v8/src/provider.ts
@@ -28,6 +28,12 @@ const FILE_PROTOCOL = 'file://'
 
 const debug = createDebug('vitest:coverage')
 
+export const providerInternals = {
+  parseCoverageAst(code: string): ReturnType<typeof parseAstAsync> {
+    return parseAstAsync(code)
+  },
+}
+
 export class V8CoverageProvider extends BaseCoverageProvider implements CoverageProvider {
   name = 'v8' as const
   version: string = version
@@ -200,10 +206,12 @@ export class V8CoverageProvider extends BaseCoverageProvider implements Coverage
     let ast
 
     try {
-      ast = await parseAstAsync(result.code)
+      ast = await providerInternals.parseCoverageAst(result.code)
     }
     catch (error) {
-      this.ctx.logger.error(`Failed to parse ${filename}. Excluding it from coverage.\n`, error)
+      if (debug.enabled) {
+        debug('Failed to parse %s. Excluding it from coverage. %O', filename, error)
+      }
       return {}
     }
 
--- a/test/coverage-test/test/parse-failures.unit.test.ts
+++ b/test/coverage-test/test/parse-failures.unit.test.ts
@@ -1,0 +1,29 @@
+import { configDefaults } from 'vitest/config'
+import { afterEach, expect, test, vi } from 'vitest'
+import { providerInternals, V8CoverageProvider } from '../../../packages/coverage-v8/src/provider.ts'
+
+afterEach(() => {
+  vi.restoreAllMocks()
+})
+
+test('v8 provider excludes parse failures without logging an error', async () => {
+  const parseCoverageAst = vi.spyOn(providerInternals, 'parseCoverageAst').mockRejectedValue(new Error('Parse failed with 1 error'))
+  const provider = new V8CoverageProvider() as any
+  const error = vi.fn()
+
+  provider.ctx = {
+    logger: { error },
+  }
+  provider.options = configDefaults.coverage
+
+  const result = await provider.remapCoverage(
+    'file:///fixtures/src/example.ts',
+    0,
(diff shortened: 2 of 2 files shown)


AriPerkkio (changes requested):
Thank you for your tokens but this behaviour is 100% intentional. We want users to see this error instead of silently hiding it.

PR #9996, opened 2026-03-27
fix(snapshot): limit properties mismatch diff to asserted keys

### LINKED ISSUE #9985: Incomplete diff of `toMatchSnapshot(properties)` mismatch error
### Describe the bug

(not sure how to describe, so 🤖's best shot. repro is mine though.)

When `toMatchSnapshot(properties)` fails, the diff misleadingly shows fields like `x` as additions, as if they shouldn't be there. But `x` is a valid part of the object — the only actual problem is that `y` changed from `5` to `10`. The diff should focus on the mismatched values, not imply that extra fields are wrong.

- repro.test.ts

```js

test('example', () => {
  // initially snapshot created with
  // { x: 'hello', y: 5 }
  expect({ x: 'world', y: 10 }).toMatchSnapshot({ y: 5 });
});
```

Test run:

```sh
 FAIL  test/repro.test.ts > example
Error: Snapshot properties mismatched

- Expected
+ Received

  {
-   "y": 5,
+   "x": "world",
+   "y": 10,
  }

 ❯ test/repro.test.ts:6:33
      4|   // initially snapshot created with
      5|   // { x: 'hello', y: 5 }
      6|   expect({ x: 'world', y: 10 }).toMatchSnapshot({ y: 5 });
       |                                 ^
      7| });
      8|
 ❯ new Promise ../../../blitz.4c73681d.js:31:30606
```

<details>
<summary>🤖 Root cause and suggestions</summary>

In `packages/snapshot/src/client.ts`, the properties mismatch error passes `received` and `properties` as `actual`/`expected`:

```ts
if (!propertiesPass) {
  return {
    pass: false,
    message: () => 'Snapshot properties mismatched',
    actual: received,      // full object
    expected: properties,  // subset — missing fields show as additions
  }
}
```

The diff is generated downstream by `processError` (`packages/utils/src/error.ts`), which just calls `printDiffOrStringify(err.actual, err.expected)`. It has no knowledge of snapshot semantics — it faithfully diffs the two values it's given. So the fix needs to happen at the source, not in the diff renderer.

This is inherently tricky because the properties check is a **subset equality** — the user intentionally provides fewer keys than the full object. There's no "correct full expected" available at this point (the previous snapshot isn't loaded yet since `snapshotState.match()` hasn't been called).

Possible approaches:
- **Filter `actual` to only the keys in `properties`** — diff becomes `{ length: 11 }` vs `{ length: 6 }`. Clean and focused on what the user cares about, but loses context of the full object.
- **Load the previous snapshot first** to construct a full expected value — more accurate but adds complexity and couples the properties check to snapshot state.

</details>

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-xnjrwzha?file=test%2Frepro.test.ts

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    @vitest/ui: latest => 4.1.2 
    vite: latest => 8.0.3 
    vitest: latest => 4.1.2
```

### Used Package Manager

npm

### Validations

raashish1601:
### Description

Limit snapshot-properties mismatch diffs to the asserted subset.

Resolves #9985

When `toMatchSnapshot(properties)` fails, Vitest currently diffs the full received object against the properties subset, which makes unrelated keys look like incorrect additions. This trims the received payload to the keys described by the properties matcher before building the mismatch error, so the diff stays focused on the failing subset while preserving nested object and array shapes.

### Tests
- [x] Run the focused snapshot regression: `pnpm --filter @vitest/test-snapshots test test/properties.test.ts`.
- [x] Rebuild the touched package: `pnpm --filter @vitest/snapshot build`.
- [x] Lint the touched files with `pnpm exec eslint packages/snapshot/src/client.ts packages/snapshot/src/port/utils.ts test/snapshots/test/properties.test.ts`.

### Documentation

No documentation change is needed because this fixes the mismatch diff for an existing matcher.

### Changesets

### DIFF
--- a/packages/snapshot/src/port/utils.ts
+++ b/packages/snapshot/src/port/utils.ts
@@ -237,24 +237,56 @@ export function deepMergeSnapshot(target: any, source: any): any {
 }
 
 export function getSnapshotSubset(target: any, source: any): any {
+  const seenReferences = new WeakMap<object, any>()
+
+  return getSnapshotSubsetWithContext(target, source, seenReferences)
+}
+
+function getSnapshotSubsetWithContext(
+  target: any,
+  source: any,
+  seenReferences: WeakMap<object, any>,
+): any {
   if (Array.isArray(target) && Array.isArray(source)) {
+    if (seenReferences.has(target)) {
+      return seenReferences.get(target)
+    }
+
     const subset: any[] = []
+    seenReferences.set(target, subset)
+
     Object.keys(source).forEach((key) => {
       if (key in target) {
         const index = Number(key)
-        subset[index] = getSnapshotSubset(target[index], source[index])
+        subset[index] = getSnapshotSubsetWithContext(
+          target[index],
+          source[index],
+          seenReferences,
+        )
       }
     })
+
     return subset
   }
 
   if (isObject(target) && isObject(source) && !source.$$typeof) {
+    if (seenReferences.has(target)) {
+      return seenReferences.get(target)
+    }
+
     const subset = Object.create(Object.getPrototypeOf(target))
+    seenReferences.set(target, subset)
+
     Object.keys(source).forEach((key) => {
       if (key in target) {
-        subset[key] = getSnapshotSubset(target[key], source[key])
+        subset[key] = getSnapshotSubsetWithContext(
+          target[key],
+          source[key],
+          seenReferences,
+        )
       }
     })
+
     return subset
   }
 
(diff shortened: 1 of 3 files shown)


sheremet-va:
This account has repeatedly abused our [AI policy](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md#ai-contributions). All their issues are closed to reduce the team's maintenance burden.

PR #10000, opened 2026-03-27
fix(runner): guard getNames against suite cycles

raashish1601:
## Summary
- guard `getNames()` against cyclic `.suite` links so malformed task graphs stop hanging Vitest error reporting
- add a focused regression that matches the public repro shape and locks the helper output for `getNames()`, `getFullName()`, and `getTestName()`

## Testing
- `corepack pnpm --filter @vitest/test-core exec vitest run test/utils.spec.ts -t "stops when a suite chain becomes cyclic"`
- `corepack pnpm exec eslint packages/runner/src/utils/tasks.ts test/core/test/utils.spec.ts`
- `git diff --check`


sheremet-va:
Hallucination

PR #10001, opened 2026-03-27
fix(browser-playwright): clear duplicate module mock routes

### LINKED ISSUE #9957: Browser Mode leaks a manual mock across spec files when the same module is mocked via two ids
### Describe the bug

This repro is pure TypeScript. Vue components are not required.

`src/probe.spec.ts` registers two manual mocks for the same logical module using two different raw ids:

- `vi.mock('~/modal', ...)`
- `vi.mock('./modal', ...)`

`src/target.spec.ts` does not register any mock. It only imports another module that touches `~/modal` again.

Both spec files pass when run individually, but when they run in the same Browser Mode process, the second file fails after the first file passes:

```text
[vitest] There was an error when mocking a module
[birpc] rpc is closed, cannot call "resolveManualMock"
```

This looks like a Browser Mode mock lifecycle bug. During the second file import, Vitest appears to still try resolving the manual mock registered in the previous file, even though the second file did not register that mock.

### Reproduction

Repository repro:

[A-kirami/vitest-browser-manual-mock](https://github.com/A-kirami/vitest-browser-manual-mock)

Commands:

```bash
bun vitest run --config vite.config.ts --no-cache src/probe.spec.ts
bun vitest run --config vite.config.ts --no-cache src/target.spec.ts
bun vitest run --config vite.config.ts --no-cache --reporter verbose src/probe.spec.ts src/target.spec.ts
```

Notes:

- `--no-cache` is important here because Vitest's sequencer uses cached file stats and durations to sort files.
- With cache enabled, the non-mocking file can run first and hide the problem.
- With `--no-cache`, the repro behaves like a fresh run and the failing order stays stable.

Expected behavior:

- `src/probe.spec.ts` passes alone
- `src/target.spec.ts` passes alone
- both specs also pass when run together in the same Browser Mode process

Actual behavior:

- the single-file runs pass
- the combined run fails after `src/probe.spec.ts` passes

Additional observation:

On a fresh run, `vitest list` also fails:

```bash
bun vitest list --config vite.config.ts --no-cache src/probe.spec.ts src/target.spec.ts
```

with:

```text
Mock /src/modal.ts wasn't registered. This is probably a Vitest error. Please, open a new issue with reproduction.
```

### System Info

```shell
System:
  OS: Windows 11 10.0.26200
  CPU: (20) x64 12th Gen Intel(R) Core(TM) i7-12700H
  Memory: 6.99 GB / 31.63 GB
Binaries:
  Node: 23.5.0
  npm: 10.9.2
  bun: 1.3.10
  Deno: 2.1.4
Browsers:
  Internet Explorer: 11.0.26100.7309
npmPackages:
  @vitest/browser-playwright: 4.1.0
  playwright:1.58.2
  vite: 7.3.1
  vitest: 4.1.0
```

### Used Package Manager

bun

### Validations

raashish1601:
Fixes #9957.

## Summary
- unroute any existing Playwright mock route before re-registering the same resolved module URL
- dedupe stored session route ids so clearing a session removes each route exactly once
- add browser mocking coverage for dual-id manual mocks that should not leak into a later test file

## Testing
- `PROVIDER=playwright BROWSER=chromium corepack pnpm -C test/browser exec vitest run --config vitest.config.unit.mts specs/mocking.test.ts -t "mocking works correctly - isolated true"`

### DIFF
--- a/packages/browser/src/client/tester/runner.ts
+++ b/packages/browser/src/client/tester/runner.ts
@@ -240,7 +240,13 @@ export function createBrowserRunner(
         }
         catch {}
       }
-      return rpc().onCollected(this.method, files)
+      const result = await rpc().onCollected(this.method, files)
+
+      if (this.method === 'collect') {
+        await mocker.invalidate()
+      }
+
+      return result
     }
 
     onTestAnnotate = (test: Test, annotation: TestAnnotation): Promise<TestAnnotation> => {
--- a/test/browser/specs/mocking.test.ts
+++ b/test/browser/specs/mocking.test.ts
@@ -1,5 +1,5 @@
 import { expect, onTestFailed, onTestFinished, test } from 'vitest'
-import { editFile, runVitest } from '../../test-utils'
+import { editFile, runVitest, runVitestCli } from '../../test-utils'
 import { instances } from '../settings'
 
 // TODO: investigate `isolate: false` tests.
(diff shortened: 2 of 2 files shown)


sheremet-va:
This account has repeatedly abused our [AI policy](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md#ai-contributions). All their issues are closed to reduce the team's maintenance burden.

Issue #9064, opened 2025-11-20
[4.0.11] @opentelemetry/api type error with new release


AviVahl:
### Describe the bug

Saw this is a downstream repository:
```
node_modules/vitest/dist/chunks/traces.d.B8ukBJqA.d.ts:5:44 - error TS2307: Cannot find module '@opentelemetry/api' or its corresponding type declarations.

5 import { Context, Span, SpanOptions } from '@opentelemetry/api';
                                             ~~~~~~~~~~~~~~~~~~~~

node_modules/vitest/dist/index.d.ts:26:8 - error TS2307: Cannot find module '@opentelemetry/api' or its corresponding type declarations.

26 import '@opentelemetry/api';
          ~~~~~~~~~~~~~~~~~~~~


Found 2 errors in 2 files.

Errors  Files
     1  node_modules/vitest/dist/chunks/traces.d.B8ukBJqA.d.ts:5
     1  node_modules/vitest/dist/index.d.ts:26
```

I saw in another type-related issue you guys are suggesting to turn `skipLibCheck` on. We cannot do that. It would skip type checking **all** .d.ts files in our project (not just `node_modules`), hiding potential real issues. Already got bit by that.

### Reproduction

```sh
mkdir vitest-type-issue
cd vitest-type-issue
npm init -y
npm pkg set type=module
npm i vitest typescript
npx tsc --init --skipLibCheck false
echo "import { expect } from 'vitest'; console.log(expect);" > main.ts
npx tsc --noEmit
```

### System Info

```shell
Fedora 43
Node 24.11.1
npm 11.6.3
```

### Used Package Manager

npm

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vitest.dev/guide/).
- [x] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

[...]

PR #10003, opened 2026-03-28
fix(build): align dist dependencies and add depcruise checks

ikeyan:
### Description

This PR adds dependency-cruiser checks for built `dist` outputs and aligns Rollup externals plus package metadata with what the generated files actually reference.

#### Build helpers and Rollup externals

- add shared build helpers to `scripts/build-utils.js`
- in each `rollup.config.js`, use those helpers to:
  - derive Rollup externals from `package.json` metadata instead of manually maintaining package-specific lists
  - include both bare and `node:`-prefixed builtin modules
  - allow package-specific self-import allowlists where needed
- remove the old Vite types resolve workaround from the dts build helper
- set `inputBase: 'src'` for `@vitest/browser` dts generation so temporary declaration files stay under `.types`

#### Package metadata changes  in `@vitest/ws-client`
- `@vitest/ws-client`: add optional peer dependency on `vitest`
- `ws`: removed because it was completely unused
- `flatted`: moved to `devDependencies` because it is bundled

#### depcruise

- rules: `.dependency-cruiser.cjs`
- baseline: `.dependency-cruiser-known-violations.json`
- scripts:
  - `pnpm depcruise` to check built `dist` dependencies while ignoring the current baseline
  - `pnpm depcruise:baseline` to regenerate the current baseline
- this PR adds local configuration and scripts only
- it does not add a CI check; wiring depcruise into CI can be done in a follow-up PR

The depcruise rules currently check built `dist` files for:

- unresolved dependencies
- undeclared or dev-only npm dependencies
- sharp imports
- ancestor npm dependencies
- circular dependencies
- orphaned (unused) dist files
- deprecated npm packages
- undetermined dependencies

depcruise helps catch regressions like:
  - missing @opentelemetry/api types in published vitest declarations (#9064)
  - vite type surface mismatches leaking through published vitest declarations (#9055)

#### UI virtual module import

- change the UI global setup import from `uno.css` to `virtual:uno.css`
- this makes the dependency explicit for depcruise, since `uno.css` is a Vite virtual module and would otherwise be treated as an unknown dependency

#### Bundled `ws` optional peers

- in the `vitest` bundle, define both `process.env.WS_NO_UTF_8_VALIDATE` and `process.env.WS_NO_BUFFER_UTIL`
- this removes runtime references from bundled `ws` code to its optional native peers `utf-8-validate` and `bufferutil`

#### Dist changes

- in the published `@vitest/browser` package, the following temporary declaration files disappear:
  - `dist/shared/screenshotMatcher/types.d.ts`
  - `dist/types.d.ts`
- the published `vitest` dist no longer references `utf-8-validate` or `bufferutil`
- other generated `js` / `d.ts` changes are the expected result of the externals cleanup above

  - Added `dependency-cruiser` to the root `package.json`, so `pnpm-lock.yaml` is changed.
  - `pnpm depcruise` can be seen as a new test.
- [x] Please check [Allow edits by maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to make review process faster. Note that this option is not available for repositories that are owned by Github organizations.

### Tests
  - Executed locally.
  - Current failures also reproduce on `main` after `pnpm build`, for example:
    - `pnpm -C test/cli test test/list.test.ts -t 'browser.enabled'`
    - `pnpm -C test/cli test test/mocking.test.ts -t 'playwright'`

### Documentation
  - No new feature.

### Changesets

#### Notes
- AI assistance disclosure: this PR was prepared with help from Codex.


sheremet-va:
Thanks for finding peer dependency issues. I made to PRs to resolve them, since they are not supposed to be in the bundle:
- https://github.com/vitest-dev/vitest/pull/10004
- https://github.com/vitest-dev/vitest/pull/10005

The `ws-client` is deprecated, so I am not fixing it there.

.dependency-cruiser-known-violations.json
```diff
@@ -0,0 +1,74 @@
+[
+  {
+    "type": "dependency",
+    "from": "packages/vitest/dist/chunks/cli-api.CdrCPbCl.js",
```
sheremet-va: These chunks will change the hash all the time

.dependency-cruiser.cjs
```diff
@@ -0,0 +1,374 @@
...
+        pathNot: [
+          ...orphanPathIgnorePatterns,
+          '^packages/browser/dist/state[.]js$', // packages/browser/src/node/projectParent.ts
+          '^packages/browser/dist/client/esm-client-injector[.]js$', // packages/browser/src/node/projectParent.ts
+        ],
+      },
+      to: {},
+    },
+    {
+      name: 'no-deprecated-core-in-dist',
```
sheremet-va: I don't think this one makes sense. If node deprecated a module, it will print a warning. We always catch unexpected warnings in tests.

It's also a _manual_ list - who will support it?

sheremet-va (changes requested):
I am actually working on another PR where I removed `tsconfig.paths` in favor of exports field. It should address at least that, but it's good to have a `dist` analyser because otherwise no one imports from there.

.dependency-cruiser.cjs
ikeyan: Dropped.

.dependency-cruiser-known-violations.json
ikeyan: @sheremet-va Resolved these violations instead by defining `process.env.WS_NO_BUFFER_UTIL` and `process.env.WS_NO_UTF_8_VALIDATE`.

sheremet-va:
The team decided that we don't want to use `depcruise`

PR #10016, opened 2026-03-28
feat: add @vitest/coverage-v8 and @vitest/coverage-istanbul as optional peer dependencies

### LINKED ISSUE #10015: vitest should declare @vitest/coverage-v8 and @vitest/coverage-istanbul as optional peer dependencies
`vitest@4.1.1` dynamically imports `@vitest/coverage-v8` when `coverage.provider` is set to `"v8"`:

```
Error: Cannot find package '@vitest/coverage-v8' imported from .../node_modules/vitest/dist/chunks/coverage.CTzCuANN.js
```

This works in hoisted `node_modules` layouts (npm, yarn) because `@vitest/coverage-v8` is hoisted from the consuming project. In pnpm's strict isolation (default in pnpm 10), vitest can only resolve packages declared in its own dependency tree — and since `@vitest/coverage-v8` isn't listed as a peer, pnpm doesn't link it.

`vitest` already declares optional peers for `@vitest/browser-*` and `@vitest/ui`. Adding coverage providers would be consistent with that pattern.

**Temp workaround:**
```yaml
# pnpm-workspace.yaml
packageExtensions:
  "vitest@4":
    peerDependencies:
      "@vitest/coverage-v8": "*"
    peerDependenciesMeta:
      "@vitest/coverage-v8":
        optional: true
```

**Suggested proper fix:** 

Add `@vitest/coverage-v8` and `@vitest/coverage-istanbul` as optional peer dependencies in `vitest`'s `package.json`, matching the existing pattern for `@vitest/browser-*` and `@vitest/ui`.

**Why is it important?**

The import **failure happens in runtime**, so it can silently break things for people who adopting strict isolation (no hoisting).

mixelburg:
Fixes #10015

## Problem

In pnpm's strict isolation mode (default in pnpm 10), packages can only resolve modules listed in their own dependency tree. Since `@vitest/coverage-v8` and `@vitest/coverage-istanbul` are not declared as peer dependencies of `vitest`, pnpm doesn't link them — even when the consuming project has them installed — causing a runtime `Cannot find package '@vitest/coverage-v8'` error when `coverage.provider` is configured.

This works in npm/yarn due to hoisting, but breaks under strict isolation.

## Fix

Add both coverage providers as optional peer dependencies, following the same pattern already used for `@vitest/browser-*` and `@vitest/ui`.

```json
"@vitest/coverage-v8": "workspace:*",
"@vitest/coverage-istanbul": "workspace:*",
```

With optional in `peerDependenciesMeta`:

```json
"@vitest/coverage-v8": { "optional": true },
"@vitest/coverage-istanbul": { "optional": true },
```

Users no longer need the `packageExtensions` workaround in `pnpm-workspace.yaml`.

### DIFF
--- a/packages/vitest/package.json
+++ b/packages/vitest/package.json
@@ -139,6 +139,8 @@
     "@vitest/browser-playwright": "workspace:*",
     "@vitest/browser-preview": "workspace:*",
     "@vitest/browser-webdriverio": "workspace:*",
+    "@vitest/coverage-istanbul": "workspace:*",
+    "@vitest/coverage-v8": "workspace:*",
     "@vitest/ui": "workspace:*",
     "happy-dom": "*",
     "jsdom": "*",
@@ -161,6 +163,12 @@
       "optional": true
     },
     "@vitest/browser-webdriverio": {
+      "optional": true
+    },
+    "@vitest/coverage-istanbul": {
+      "optional": true
+    },
+    "@vitest/coverage-v8": {
       "optional": true
     },
     "@vitest/ui": {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1017,6 +1017,12 @@ importers:
       '@vitest/browser-webdriverio':
         specifier: workspace:*
         version: link:../browser-webdriverio
+      '@vitest/coverage-istanbul':
+        specifier: workspace:*
+        version: link:../coverage-istanbul
+      '@vitest/coverage-v8':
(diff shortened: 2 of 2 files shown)


AriPerkkio (changes requested):
How about custom coverage providers? Should we list every possible provider that the community has built? Same for reporters and other integrations?

We'll need minimal reproduction in the reported issue to see if this is real issue.

PR #10023, opened 2026-03-30
docs: add tip for using HTML reporter on CI

### LINKED ISSUE #9971: feat: hosted viewer for HTML report artifacts (zero-install, browser-only)

Vitest's `--reporter=html` produces a self-contained static site. In CI, this gets uploaded as an artifact zip. Viewing it requires: download zip -> unzip -> `npx serve html/` -> open browser. This applies to any CI workflow producing HTML reports — whether from `--merge-reports` or a regular run.

Note: GitLab CI can [preview HTML artifacts directly in the browser](https://docs.gitlab.com/ee/ci/jobs/job_artifacts.html) (powered by GitLab Pages). But GitHub Actions — the most common CI for Vitest users — does not support this, so the friction is real for most users.

### Suggested solution

(🤖 This could be a slop. The idea to be verified.)

Host a static viewer page (e.g. `vitest.dev/report`) that lets users open an HTML report artifact zip entirely in the browser — no server, no install.

Implementation: a service worker intercepts fetch requests and serves files from the unzipped artifact. Since the HTML report is already a self-contained static site (SPA + data), the viewer is just a thin "zip unpacker + virtual file server" shell. The actual rendering comes from the artifact's own `index.html`, so there are no version compatibility concerns between the hosted viewer and the report format.

Prior art:
- Playwright's trace viewer (`trace.playwright.dev`) uses this exact pattern — a service worker + `@zip.js/zip.js` to serve trace data from a zip URL
- [`client-side-zip-server`](https://github.com/gkjohnson/client-side-zip-server) — generic SW-based zip serving

The core SW implementation is small (~100 lines). UX would primarily be drag-and-drop. A `?url=` param is technically feasible — both GitHub API and the Azure blob storage redirect target set `Access-Control-Allow-Origin: *` — but the GitHub artifact download API requires authentication. The redirect target (a pre-signed Azure blob URL, expires in ~1 min) is publicly accessible, so `?url=` could work if a CI step posts a comment with that URL. Drag-and-drop is the simpler, more universal approach.

This concept might be actually generic to any zipped static site, not Vitest-specific. If implementation turns out to be fully agnostic, it could live as a standalone tool that Vitest simply documents/recommends.

### Alternative

_No response_

### Additional context

_No response_

### Validations

hi-ogawa:
### Description

- Closes https://github.com/vitest-dev/vitest/issues/9971

~I added a general CI usage of HTML reports as example, then did a small mention of https://zipview.hiro18181.workers.dev/ We could tweak zipview and host under vitest.dev domain, but not sure if it can improve the experience in a meaningful way.~

EDIT: Now the code lives in https://github.com/vitest-dev/viewer and available in https://viewer.vitest.dev/

### Tests

### Documentation

### Changesets


docs/guide/ui.md
```diff
@@ -53,6 +53,19 @@ npx vite preview --outDir ./html
...
+To view the HTML report from CI, for example in GitHub Actions, upload the output directory as an artifact:
+
+```yaml
+- uses: actions/upload-artifact@v4
+  with:
+    name: vitest-report
+    path: html/
+```
+
+Download and extract it, then run `vite preview` locally as above — or open the zip directly in [Zipview](https://zipview.hiro18181.workers.dev/) without any local setup.
```
AriPerkkio: I didn't (yet) test how this works, but would it be possible to make the URLs stateful? There could be query param that points to the zip file download, that is then automatically downloaded in the browser/worker. Something like `https://zipview.hiro18181.workers.dev?s=https://github.com/vitest-tests/test-sharding/actions/runs/23730331118/artifacts/6171948355`? Or even use [`lz-string`](https://github.com/AriPerkkio/html-preview/tree/main/src/lib/lz-string) or similar compression.

This would allow auto-generating the URLs and make them shareable inside teams.

Would auto-downloading be risky though? If yes, it could be prompted first: `Download and view report from <url>? Yes/No`.

hi-ogawa: If the zip is CORS accessible, then it's already supported and that's the demo example of https://zipview.hiro18181.workers.dev/?url=https://cdn.jsdelivr.net/gh/hi-ogawa/zipview@main/docs/assets/vitest-html-reporter.zip

The issue is mostly CORS on Github assets. Github's download links are all protected (with a redirect to an actual azure object storage url with timestamp and signature based on github auth coockie status), so it's likely difficult to support just accepting the link and make it viewable.

I think your suggestion makes sense and I can probably check client fetch error of `?url=....` and prompt users to download through browser (which should activate cookie), then users can drop it from browser download box. Will try this later.

Regarding lz-string compression, I'm not sure it's practical since the above demo is from `examples/basic` and this is already taking 457KB https://github.com/hi-ogawa/zipview/blob/main/docs/assets/vitest-html-reporter.zip. I don't this would fit in browser URL in any compression.

hi-ogawa: > I think your suggestion makes sense and I can probably check client fetch error of ?url=.... and prompt users to download through browser (which should activate cookie), then users can drop it from browser download box. Will try this later.

Okay, this works now. https://zipview.hiro18181.workers.dev/?url=https://github.com/vitest-dev/vitest/actions/runs/23575176696/artifacts/6115674019

This artifact is from https://github.com/vitest-dev/vitest/pull/9967 and it will prompt like:

image

hi-ogawa: Here is an example of github summary with zipview link https://github.com/vitest-dev/vitest/actions/runs/23734731289?pr=10023#summary-69136434148

docs/guide/ui.md
```diff
@@ -53,6 +53,23 @@ npx vite preview --outDir ./html
...
+
+```yaml
+- uses: actions/upload-artifact@v4
+  id: upload-report
+  with:
+    name: vitest-report
+    path: html/
+
+- name: Report summary
+  run: echo "[View HTML report on Zipview](https://zipview.hiro18181.workers.dev/?url=${{ steps.upload-report.outputs.artifact-url }})" >> $GITHUB_STEP_SUMMARY
```
sheremet-va: Shouldn't we host it in our docs? `zipview.vitest.dev`

hi-ogawa: It maybe simpler to use https://vitest.dev/zipview. Let me test a bit.

sheremet-va: No, I think it should be _published_ and _hosted_ by us, not the external website. Just redirecting it doesn't solve the issue

It should be under our domain is the point I am making

hi-ogawa: https://github.com/vitest-dev/vitest/pull/10023/commits/cf79c500aa14a61b8ce7428b20852a905e9631e5 This isn't 30x redirection though. It's served directly on https://deploy-preview-10023--vitest-dev.netlify.app/zipview/ 

Anyways, I can move it to vitest netlify account and also repo https://github.com/hi-ogawa/zipview to under https://github.com/vitest-dev 

What do we need to do for these?

sheremet-va (approved):

AriPerkkio (approved):

PR #10024, opened 2026-03-30
fix(vm): fix external module resolve error with deps optimizer query

### LINKED ISSUE #9899: vmThreads/vmForks pool crashes with ERR_MODULE_NOT_FOUND when deps.optimizer.client.enabled is true
## Description

When `pool: 'vmThreads'` (or `pool: 'vmForks'`) is combined with `deps.optimizer.client.enabled: true`, every test worker crashes immediately with:

```
Error: Cannot find module '/path/to/node_modules/vitest/dist/spy.js?v=d0e9b0ac'
Serialized Error: { code: 'ERR_MODULE_NOT_FOUND' }
```

The `?v=<hash>` suffix is appended by the Vite optimizer as a cache-buster for pre-bundled modules. Node's VM module executor (used by both VM pool types) resolves modules via the filesystem and does not strip query strings from paths, so the file is never found.

The two features are therefore mutually exclusive in their current form — enabling the optimizer to speed up cold-cache CI runs prevents using VM pools to speed up local/warm runs.

---

## Reproduction

```ts
// vitest.config.ts

export default defineConfig({
  test: {
    environment: 'happy-dom',
    pool: 'vmThreads', // also fails with 'vmForks'
    deps: {
      optimizer: {
        client: {
          enabled: true,
          include: ['react', 'react-dom'],
        },
      },
    },
  },
});
```

```ts
// example.test.ts
test('basic', () => { expect(1 + 1).toBe(2); });
```

Run:
```bash
vitest run
```

---

## Error output

```
Error: Cannot find module '/absolute/path/node_modules/vitest/dist/spy.js?v=d0e9b0ac'
  at ExternalModulesExecutor.createModule (vitest/dist/chunks/vm.D3epNOPZ.js:621:34)
  at ExternalModulesExecutor.import (vitest/dist/chunks/vm.D3epNOPZ.js:553:29)
Serialized Error: { code: 'ERR_MODULE_NOT_FOUND' }
```

The hash suffix (`?v=d0e9b0ac`) is non-deterministic — it changes per run.

---

## Root cause analysis

The Vite optimizer appends `?v=<hash>` to pre-bundled module specifiers as a cache-invalidation mechanism (see Vite source: `packages/vite/src/node/optimizer/index.ts`). When vitest's VM executor (`ExternalModulesExecutor`) tries to load vitest's own internal modules (e.g. `vitest/dist/spy.js`) that were caught up in the optimizer's module graph, it receives the path with the query string appended. Node's `vm.Module` / `createRequire` resolves strictly against the filesystem — `spy.js?v=d0e9b0ac` does not exist as a file, so resolution fails.

The `forks` pool is unaffected because it uses standard Node.js `require`/`import` in a subprocess, which goes through Vite's dev server transform pipeline where the `?v=` suffix is handled correctly.

---

## Expected behaviour

Either:
1. The VM executor strips query strings before filesystem resolution (matching how the Vite dev server handles them), or
2. Vitest's internal modules are excluded from the optimizer's module graph so they are never given a `?v=` suffix, or
3. The combination is detected at startup and a clear error is thrown explaining the incompatibility

---

## Environment

| | |
|---|---|
| vitest | 4.0.18 |
| vite | 7.3.1 |
| node | 24.12.0 |
| OS | macOS |
| pool | vmThreads / vmForks (both affected) |
| environment | happy-dom (also reproducible with jsdom) |

---

## Workaround

Use `pool: 'forks'` (the default) when `deps.optimizer.client.enabled: true`. The two features cannot be used together until this is resolved.

hi-ogawa:
### Description

- closes https://github.com/vitest-dev/vitest/issues/9899

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/runtime/external-executor.ts
+++ b/packages/vitest/src/runtime/external-executor.ts
@@ -138,10 +138,9 @@ export class ExternalModulesExecutor {
       return { type: 'network', url: identifier, path: identifier }
     }
 
-    const fileUrl = identifier.startsWith('file://')
-      ? identifier
-      : (pathToFileURL(file) + postfix)
-    const pathUrl = fileURLToPath(fileUrl) // this trims off `?...`
+    const isFileUrl = identifier.startsWith('file://')
+    const fileUrl = isFileUrl ? identifier : `${pathToFileURL(file)}${postfix}`
+    const pathUrl = isFileUrl ? fileURLToPath(file) : file
 
     let type: 'module' | 'commonjs' | 'vite' | 'wasm'
     if (this.vite.canResolve(fileUrl)) {
--- a/packages/vitest/src/runtime/external-executor.ts
+++ b/packages/vitest/src/runtime/external-executor.ts
@@ -5,7 +5,7 @@ import type { VMModule } from './vm/types'
 import fs from 'node:fs'
 import { isBuiltin } from 'node:module'
 import { fileURLToPath, pathToFileURL } from 'node:url'
-import { cleanUrl, isBareImport } from '@vitest/utils/helpers'
+import { isBareImport, splitFileAndPostfix } from '@vitest/utils/helpers'
 import { findNearestPackageData } from '@vitest/utils/resolver'
 import { extname, normalize } from 'pathe'
 import { CommonjsExecutor } from './vm/commonjs-executor'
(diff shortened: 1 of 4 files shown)


packages/vitest/src/runtime/external-executor.ts
```diff
@@ -125,7 +125,8 @@ export class ExternalModulesExecutor {
       return { type: 'data', url: identifier, path: identifier }
     }
 
-    const extension = extname(identifier)
+    const { file, postfix } = splitFileAndPostfix(identifier)
```
hi-ogawa, line 128: Added `splitFileAndPostfix` here, so later `pathUrl` won't include `?v=...`, which was the cause of `existsSync` not finding the file and causing module not found.

sheremet-va (approved):

marcebdev:
Any ideas when this may release as a patch?

PR #10031, opened 2026-03-31
feat: support merge reports for non-sharded multi-environment runs (take 2)

### LINKED ISSUE #9954: Support merging reports for non-sharded multi-environment runs

hi-ogawa:
### Description

- Closes https://github.com/vitest-dev/vitest/issues/9954
- Closes https://github.com/vitest-dev/vitest/pull/9967 (supersedes)

Alternative take for https://github.com/vitest-dev/vitest/pull/9967#issuecomment-4141442709

This PR adds new blob reporter option `label` so `--merge-reports` can distinguish results from the same test files across different environments, such as linux/macos/windows CI jobs.

### Usage example

Set `VITEST_BLOB_LABEL` when creating blobs:

```bash
VITEST_BLOB_LABEL=linux vitest run --reporter=blob
VITEST_BLOB_LABEL=macos vitest run --reporter=blob
```

Or pass the label via the blob reporter option:

```ts
// vitest.config.ts
export default defineConfig({
  test: {
    reporters: [
      ['blob', { label: process.platform }],
    ],
  },
})
```

The label is included in the default blob filename (`blob-linux.json`) to avoid collisions. When merging, each labeled run is shown separately:

```sh
$ vitest --merge-reports
 ✓  linux  src/basic.test.ts (2 tests)
 ✓  macos  src/basic.test.ts (2 tests)
```

### How it works

The label is a blob reporter option, with `VITEST_BLOB_LABEL` as an environment-variable shortcut for CI usage. During config resolution, Vitest extracts the label from `['blob', { label }]` or `process.env.VITEST_BLOB_LABEL` and stores it as internal core config (`mergeReportsLabel`).

During the original test run, this label is not exposed as a normal built-in reporter label. This keeps terminal output unchanged for runs that create blobs with another reporter enabled, for example `reporters: ['default', ['blob', { label: 'linux' }]]`.

The label is stored in blob task metadata and salts the file identity used for merged reports, so state deduplication can keep same-filepath files from different environments separate. The default blob output filename also uses the label as `blob-(label)-(shard).json` when no explicit output file is configured.

Reporters read the label from merged blob metadata and display it when replaying `--merge-reports`.

### The difference from #9967

| Concern | #9967 (label as project.name) | This PR (label as blob metadata) |
|---------|------------------------|---------------------|
| Mechanism | Clone `TestProject` per label at read time | Store label in blob task metadata and salt file identity |
| `readBlobs` changes | Significant — label discovery, ID rewrite, env module rewrite | No project cloning or read-time project rewrite |
| `file.projectName` | Mutated to include label | projectName unchanged |
| API | `[["blob", { label }]]` reporter option or `VITEST_BLOB_LABEL` | `[["blob", { label }]]` reporter option or `VITEST_BLOB_LABEL` (same) |
| Original run output | Label could appear as a project-style label | Built-in reporters do not show the blob label during the original run |

### Demo

I setup CI to upload html reports for `test/cli` and `test/core`. It's viewable on newly introduced viewer https://github.com/vitest-dev/vitest/pull/10023.

https://viewer.vitest.dev/?url=https://github.com/vitest-dev/vitest/actions/runs/24275938517/artifacts/6383832383

image

### Tests

### Documentation

### Changesets


packages/vitest/src/node/reporters/blob.ts
```diff
@@ -50,9 +50,9 @@ export class BlobReporter implements Reporter {
       = this.options.outputFile ?? getOutputFile(this.ctx.config, 'blob')
     if (!outputFile) {
       const shard = this.ctx.config.shard
-      outputFile = shard
-        ? `.vitest-reports/blob-${shard.index}-${shard.count}.json`
-        : '.vitest-reports/blob.json'
+      const label = this.ctx.config.label
+      const suffix = [label, shard && `${shard.index}-${shard.count}`].filter(Boolean).join('-')
+      outputFile = suffix ? `.vitest-reports/blob-${suffix}.json` : '.vitest-reports/blob.json'
     }
```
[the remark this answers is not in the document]
hi-ogawa: @copilot apply changes based on [this feedback](https://github.com/vitest-dev/vitest/pull/10031#discussion_r3014907460)

sheremet-va:
I like the approach, but I feel like `meta.label` might be too generic. Typecheck is a static one where users can't even modify meta, and in benchmark you don't have access to `meta`.

Should we make it more internal to avoid clashes with user meta? It could also be removed from json.meta: https://github.com/vitest-dev/vitest/pull/10078

hi-ogawa:
Agree `meta.label` can definitely crash. I'll rename it to sound like something vitest owned.

hi-ogawa:
Hash was aligned so pre-parse mode was broken. I consolidated ast-collect with the same `createFileTask` from runner package.

hi-ogawa:
We discussed that
- receive the label via blob reporter option `['blob', { label: ... }]`
- stash this as core level internal config by extracting it during `resolveConfig`
- during the original test run, this label shouldn't show up as label for builtin reporters.

hi-ogawa:
The PR is updated with blob reporter level option. I've also revived `VITEST_BLOB_LABEL` environment support so it can one shot from CLI `VITEST_BLOB_LABEL=linux vitest --reporter=blob`.

sheremet-va:
LGTM, but I inlined `generateFileId` in the orchestrator which is not completely correct now:

https://github.com/vitest-dev/vitest/blob/95dc6e3f16ce20a75df21a05fa6d2181d9dd9b1a/packages/browser/src/client/orchestrator.ts#L464

Importing from utils also imports the runner for some reason and the size balloons. I am removing the runner package in the next PR anyway

docs/guide/reporters.md
```diff
@@ -695,13 +695,32 @@ By default, stores all results in `.vitest-reports` folder, but can be overridde
...
 npx vitest --merge-reports=reports --reporter=json --reporter=default
 ```
 
+When running the same tests across multiple environments, use the `VITEST_BLOB_LABEL` environment variable to distinguish each environment's blob. Vitest reads labels at merge time and displays results separately:
+
+```bash
+VITEST_BLOB_LABEL=linux vitest run --reporter=blob
+```
+
+You can also provide the label via the blob reporter option:
```
AriPerkkio: Based on `resolveConfig()` I think we should state:

```suggestion
You can also provide the label via the blob reporter option. This has higher priority than `VITEST_BLOB_LABEL`.
```

packages/vitest/src/node/project.ts
```diff
@@ -551,6 +555,7 @@ export class TestProject {
       server.config,
     )
     this._config.api.token = this.vitest.config.api.token
+    this._config.mergeReportsLabel = this.vitest.config.mergeReportsLabel
```
AriPerkkio, line 558: Just to verify - this sets the root config's `mergeReportsLabel` to each Vitest project?

sheremet-va (dismissed):

test/unit/vite.config.ts
```diff
@@ -125,9 +125,14 @@ export default defineConfig({
       },
     },
     includeTaskLocation: true,
-    reporters: process.env.GITHUB_ACTIONS
-      ? ['default', ['github-actions', { displayAnnotations: false }]]
-      : [['default', { summary: true }], 'hanging-process'],
+    reporters: [
+      ...(process.env.GITHUB_ACTIONS
+        ? [['default'], ['github-actions', { displayAnnotations: false }] as any]
```
AriPerkkio, line 130: Btw we should probably add `displayAnnotations` to test utils where we spin up Vitest inside Vitest. Right now all those nested runs are spamming their results to Github's env, so Actions tab looks really verbose.

packages/vitest/src/node/project.ts
hi-ogawa: Yes, this is intended to force `mergeReportsLabel` to work like "root only" config. `api.token` was also something like that if I remember correctly.

test/unit/vite.config.ts
sheremet-va: Maybe it should just be disabled by default overall

packages/ui/client/components/views/ViewTestReport.vue
```diff
@@ -25,6 +25,7 @@ const failed = computed(() => {
 const kWellKnownMeta = new Set([
   'benchmark',
   'typecheck',
+  'label',
```
sheremet-va: Wasn't it renamed?

hi-ogawa: Good catch. Just replaced.

Btw, this does't actually reach though because ~both typecheck and~ `__vitest_label__` are file level only meta. (Ah typecheck does appear each suite and test).

sheremet-va: I think tests merge parent's meta, now? At least that was the behaviour I intended in 4.1

hi-ogawa: This is probably an inconsistency because our runtime collect does odd thing with "default root suite" (like fixing up hash after collect), so file is missing as parent suite.

sheremet-va: Why does `@module-tag` work then 🤔

hi-ogawa: I didn't  actually verify the raw data. I didn't see any meta in the html report from the commit before the fix, so I was inferring.

hi-ogawa: Okay, so conveniently there's a divergence between `tags` and `meta` inheritance

https://github.com/vitest-dev/vitest/blob/273933440292e2e669ee7d415bf24a467f816617/packages/runner/src/suite.ts#L317-L320

https://github.com/vitest-dev/vitest/blob/273933440292e2e669ee7d415bf24a467f816617/packages/runner/src/suite.ts#L350

At the top level, we have `currentSuite = undefined` but `parentTask = file`, so `file.tags` can get inherited but `file.meta` doesn't. 

Unless there's a concrete use cases or bugs, we can probably leave this divergence as is.

sheremet-va:
Lint is failing 😞

sheremet-va (approved):

PR #10033, opened 2026-03-31
fix(typecheck): log experimental feature warning once per run

### LINKED ISSUE #9823: Option to disable "Testing types with tsc and vue-tsc is an experimental feature." logs

Right now, when you run with typecheck enabled, this warning is emitted:

https://github.com/vitest-dev/vitest/blob/b0b83e21424a9e9330c07b5f1d5ed82e01aa3f29/packages/vitest/src/node/config/resolveConfig.ts#L791-L795

```plaintext
Testing types with tsc and vue-tsc is an experimental feature.
Breaking changes might not follow SemVer, please pin Vitest's version when using it.
```

That's good to know when setting up the typecheck feature. But if a project is already using the feature, it can become kind of a lot. Especially in monorepos where multiple instances of Vitest get run.

### Suggested solution

Could we have an option to say _"yes, I know this is experimental, don't log"_? Or at least only log once per system, determined by writing to a cache?

### Alternative

...or stabilize the typecheck feature?

### Additional context

Related:

* https://github.com/vitest-dev/vitest/issues/7773: filed last year, still awaiting triage
* https://github.com/flint-fyi/flint/issues/2398: we're in a monorepo with projects such that 33 of these logs are written upon every root-level `pnpm test` (😭)

### Validations

bwyard:
When using Vitest with multiple projects that have typecheck.enabled: true, the experimental feature warning is logged once per project config resolved. In a large monorepo this can mean 30+ identical warnings per run. Fixes #9823.

Changed the deduplication key from a string in a module-level Set to the Logger instance in a WeakSet. The logger is constructed per Vitest instance, so the warning fires once per run regardless of how many project configs are resolved - and correctly resets between separate runs.

Added a test: two projects with typecheck.enabled: true in a single run produce exactly one occurrence of the warning.

Noticed the same pattern in the benchmark() function in cac.ts. Left it out, out of scope.

AI assisted: Claude

### DIFF
--- a/packages/vitest/src/node/config/resolveConfig.ts
+++ b/packages/vitest/src/node/config/resolveConfig.ts
@@ -28,6 +28,8 @@ import { getWorkersCountByPercentage } from '../../utils/workers'
 import { BaseSequencer } from '../sequencers/BaseSequencer'
 import { RandomSequencer } from '../sequencers/RandomSequencer'
 
+const loggedExperimentalWarnings = new Set<string>()
+
 function resolvePath(path: string, root: string) {
   // local-pkg (mlly)'s resolveModule("./file", { paths: ["/some/root"] }) tries
   // /some/file
@@ -796,7 +798,8 @@ export function resolveConfig(
   resolved.typecheck ??= {} as any
   resolved.typecheck.enabled ??= false
 
-  if (resolved.typecheck.enabled) {
+  if (resolved.typecheck.enabled && !loggedExperimentalWarnings.has('typecheck')) {
+    loggedExperimentalWarnings.add('typecheck')
     logger.console.warn(
       c.yellow(
         'Testing types with tsc and vue-tsc is an experimental feature.\nBreaking changes might not follow SemVer, please pin Vitest\'s version when using it.',
--- a/packages/vitest/src/node/config/resolveConfig.ts
+++ b/packages/vitest/src/node/config/resolveConfig.ts
@@ -28,7 +28,7 @@ import { getWorkersCountByPercentage } from '../../utils/workers'
 import { BaseSequencer } from '../sequencers/BaseSequencer'
 import { RandomSequencer } from '../sequencers/RandomSequencer'
 
-const loggedExperimentalWarnings = new Set<string>()
+const loggedExperimentalWarnings = new WeakSet<Logger>()
 
 function resolvePath(path: string, root: string) {
   // local-pkg (mlly)'s resolveModule("./file", { paths: ["/some/root"] }) tries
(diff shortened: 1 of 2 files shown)


bwyard:
The Rolldown&Test: node-22, macos-latest failure looks like a pre-existing flaky UI test (test/ui/test/ui.spec.ts locator timeout). Same job is also failing on main and on other unrelated PRs (e.g. #7100). Flagging in case it's worth a look before reviewing. Will appreciate any guidance on how to proceed.

travisbreaks (commented):
Clean fix for monorepo noise. The WeakSet approach is interesting. A few observations:

1. **WeakSet<Logger> vs Set<string>**: The WeakSet scopes the dedup to the Logger instance's lifetime, which naturally aligns with the run lifecycle. If a Logger is GC'd between runs (which it should be), the next run gets a fresh WeakSet entry. This is better than a module-level Set<string> which would persist across runs in watch mode. Good call.

2. **GC timing**: In practice, the Logger instance should be held by the test runner for the duration of the run, so premature GC shouldn't be a concern. But worth confirming that the Logger reference isn't dropped and recreated mid-run in any monorepo configuration.

3. **Regression test**: Does the test actually exercise the multi-project scenario (multiple Vitest projects logging the same experimental warning)? That's the real reproduction case for 30+ duplicate warnings per run.

Looks correct to me.

bwyard:
Thanks I actually started with set as a fix and off the top of my head can't remember but I ran into issue or tests or something that caused set to not work how I washed to so swapped to weak set. Not at computer now but the regression and the gc are good observations, when I'm at my computer I'll sit down and double check things based on your feedback as well just to be sure I covered everything.

bwyard:
Thanks. Checked the Logger lifecycle: `logger` is declared `public readonly` on the Vitest class, assigned once in the constructor, and accessed in `resolveConfig` as `const logger = vitest.logger`. Should be good on GC.

On the test: I believe the original did cover it, but replaced it with a version that makes the intent clearer. `PROJECT_COUNT = 5` projects, each triggering a separate `resolveConfig` call via `TestProject._configureServer`. Without the fix that would be 5 warnings. Assertion is still `toBe(1)`. The const makes it easy to push the number higher to verify.

packages/vitest/src/node/config/resolveConfig.ts
```diff
@@ -796,7 +798,8 @@ export function resolveConfig(
   resolved.typecheck ??= {} as any
   resolved.typecheck.enabled ??= false
 
-  if (resolved.typecheck.enabled) {
+  if (resolved.typecheck.enabled && !loggedExperimentalWarnings.has(logger)) {
+    loggedExperimentalWarnings.add(logger)
     logger.console.warn(
```
sheremet-va, line 803: I think we should define `logger.warnOnce` like vite does. And maybe we should use it for all experimental warning

sheremet-va (changes requested):

packages/vitest/src/node/config/resolveConfig.ts
bwyard: I'm definitely not against that,  and id prefer that as well was just trying to keep the pr small and to scope since it's first contribution. But I'll go ahead and work on doing it that's way it's probably cleaner.

bwyard: Got real sick I'll look at this again this weekend

PR #10034, opened 2026-03-31
docs(reporters): add example usage for .vitest output directory

Srasti527:
Adds a small example demonstrating how to configure reporter output to use the `.vitest` directory.

This helps make the reporter output convention more practical and easier to adopt for users.


sheremet-va:
> @sheremet-va Thanks for reviewing! Could you please share any feedback on why this was closed? I’d love to improve and contribute better.

If you were a human, you would've known.

PR #10045, opened 2026-04-02
docs: document browser mode auto cleanup behavior

### LINKED ISSUE #35: Cursors position not reset between tests?
In this test:

```
import React from 'react';
import { render } from 'vitest-browser-react';
import './aa.css';

const TestComponent = (props: {children: any }) => {
    return <div className='aa' role='button' style={{width: 100, height: 50}}>{props.children}</div>
}

describe('Test', () => {
	test('default', async () => {
		const { container } = await render(<TestComponent>default</TestComponent>);
		await expect.element(container).toMatchScreenshot('default');
	});
	test('default hower', async () => {
        const screen = await render(<TestComponent>default</TestComponent>);
		await screen.getByRole('button').click();
		await expect.element(screen.container).toMatchScreenshot('default-expanded');
	});
	test('default again', async () => {
		const { container } = await render(<TestComponent>default</TestComponent>);
		await expect.element(container).toMatchScreenshot('default again');
	});
});
```

aa.css
```
.aa {
    background-color: blue;
    &:hover {
        background-color: yellow;
    }
}
```

The screenshot of `default again` will have the element with yellow background even if there was no interaction with that element in that test.

Image

When `default again` is tagged with `.only` the test creates the expected screenshot with blue bg.

Image

To me it seems it might be because cursor position is leaking between the tests, and since the previous test clicked an element in that position, next test is affected.

hi-ogawa:
### Description

- Closes https://github.com/vitest-community/vitest-browser-react/issues/35

### Tests

### Documentation

### Changesets


test/browser/fixtures/user-event-hover/vitest.config.ts
```diff
@@ -0,0 +1,19 @@
...
+import { provider, instances } from '../../settings'
+
+export default defineConfig({
+  cacheDir: fileURLToPath(new URL("./node_modules/.vite", import.meta.url)),
+  test: {
+    // TODO:
+    // playwright/firefox doesn't seem to have mouse state isolation per frames.
+    // for example, the following can fail by race conditions:
+    // BROWSER=firefox pnpm -C test/browser/ test-fixtures --root fixtures/user-event-hover --browser.headless --fileParallelism
+    fileParallelism: false,
```
hi-ogawa, line 12: Not sure if this is expected. Is it the issue on vitest or playwright? Should we track the issue?

sheremet-va (approved):
I wonder if we should track and reset only if needed, but I'm not sure

test/browser/fixtures/user-event-hover/vitest.config.ts
sheremet-va: Interesting, I would expect it to spawn different contexts with different positions, although since it's not chromium maybe it works differently

hi-ogawa:
Also another concern is that actually resetting cursor position seems a bit tricky. `unhover` just move cursors to the body. On playwright, I tried `{ position: { x: -1, y: -1 } }` and webkit got broken. With `{ x: 0, y : 0 }`, not sure if this can still hover the element at the top left. 

Even if we want, I'm not sure it's possible to absolutely reset the cursor.

PR #10049, opened 2026-04-02
fix(browser): spread user server options into browser Vite server in project

GoldStrikeArch:
### Description

`createBrowserServer` passes `project.options` into `createViteServer()`, but the nested `server` block was replaced with a fresh object containing only `hmr` and `watch`. This caused user settings like `server.https`, `server.host`, and `server.cors` to be silently dropped.

Spread `project.options?.server` before the overrides so that user-level server configuration is preserved while Vitest still controls HMR and file watching.

### Tests

### Documentation

### Changesets


hi-ogawa:
Can you provide reproduction to show what's not working for your use case? Alternatively you can add new test cases to show what gets fixed by this PR.

GoldStrikeArch:
> Can you provide reproduction to show what's not working for your use case

This is not working:

```ts
// vitest.config.ts
export default {
    test: {
      projects: [{
        server: { host: '0.0.0.0', cors: true },
        test: { browser: { enabled: true, ... } }
      }]
    }
  }
```

I can add the tests for it, sure

hi-ogawa:
Right, the behavior looks specific to `projects`, so it's probably a bug. For root config, we do have merging behavior as it's tested in https://github.com/vitest-dev/vitest/blob/f6b2322b16e18669db24ade23eefe66aba171435/test/browser/test/server-headers.test.ts#L3-L7

hi-ogawa (approved):

sheremet-va (approved):

PR #10051, opened 2026-04-03
fix(cli): use createColors instead of disableDefaultColors in agent environments

### LINKED ISSUE #10046: Agent detection disables colors in user code via shared tinyrainbow singleton
# Describe the bug

Vitest 4.1.2 calls `disableDefaultColors()` from tinyrainbow 3.1.0 when `std-env`'s `isAgent` is true (e.g. running inside Claude Code, Cursor, etc.):

```js
// vitest/dist/chunks/cac.DRKYQDPl.js
import { isAgent } from 'std-env';
import { disableDefaultColors } from 'tinyrainbow';

if (isAgent) disableDefaultColors();
```

`disableDefaultColors()` mutates tinyrainbow's shared default export singleton via `Object.assign()`, turning all color functions into identity functions. This affects not just Vitest's own output, but also **user code under test** that imports tinyrainbow — because both Vitest and user code resolve to the same singleton instance.

This means any test that asserts on ANSI-colored output (e.g. testing a CLI formatter that uses tinyrainbow) silently breaks when run inside an AI agent environment.

# Reproduction

1. Create a module that uses tinyrainbow:
```ts
import color from "tinyrainbow";
export const warn = (msg: string) => `${color.yellow("⚠")} ${msg}`;
```

2. Write a test asserting colored output:
```ts
it("should produce yellow warning symbol", () => {
  expect(warn("oops")).toContain("\x1b[33m⚠\x1b[39m");
});
```

3. Run with `CLAUDECODE=1` (or any env that makes `std-env` detect an agent):
```bash
CLAUDECODE=1 vitest run
```

The test fails because `color.yellow` becomes an identity function.

**Expected behavior:** Vitest's agent detection should only affect Vitest's own reporter output, not mutate shared state that user code depends on.

**Possible solutions:**
- Use `createColors()` for Vitest's internal rendering instead of mutating the default singleton
- Scope the disable to Vitest's reporter layer rather than the global default

# System Info

```shell
  System:
    OS: Windows 11 10.0.26100
    CPU: (32) x64 AMD Ryzen 9 9950X3D 16-Core Processor
    Memory: 37.30 GB / 93.64 GB
  Binaries:
    Node: 25.8.2 - C:\Users\Christopher\.local\share\mise\installs\node\25.8.2\node.EXE
    npm: 11.11.1 - C:\Users\Christopher\.local\share\mise\installs\node\25.8.2\npm.CMD
    pnpm: 10.33.0 - C:\Users\Christopher\.local\share\mise\installs\npm-corepack\0.34.6\pnpm.CMD
    bun: 1.3.11 - C:\Users\Christopher\.local\share\mise\installs\bun\1.3.11\bin\bun.EXE
    Deno: 2.7.10 - C:\Users\Christopher\.local\share\mise\installs\deno\2.7.10\bin\deno.EXE
  Browsers:
    Firefox: 149.0 - C:\Program Files\Mozilla Firefox\firefox.exe
  npmPackages:
    @vitest/coverage-v8: catalog:test => 4.1.2
    @vitest/eslint-plugin: catalog:lint => 1.6.13
    vitest: catalog:test => 4.1.2
```

# Used Package Manager

pnpm

# Validations

---

*Generated with [Claude Code](https://claude.ai/claude-code)*
Verified manually by me!

wiliyam:
## Summary

Fixes #10046

### Problem

When Vitest detects an agent environment (`isAgent=true`, e.g. Claude Code, Cursor, Copilot), `createCLI()` calls `disableDefaultColors()` from tinyrainbow. This mutates tinyrainbow's **shared default-export singleton** via `Object.assign()`, turning all color functions into identity functions.

Since both Vitest and user code under test resolve to the same singleton instance, any test asserting on ANSI-colored output from tinyrainbow silently fails in agent environments:

```ts
// user test
it('should produce yellow warning', () => {
  expect(warn('oops')).toContain('\x1b[33m⚠\x1b[39m') // fails — color.yellow is now a no-op
})
```

### Fix

Replace `disableDefaultColors()` with `createColors({ force: !isAgent })` to produce an **isolated** color instance for Vitest's own CLI output. The default singleton is no longer mutated, so user code's tinyrainbow imports are unaffected.

`createColors()` has been part of tinyrainbow's public API since v1 and is already in the lockfile.

### Before / After

| Environment | Before | After |
|---|---|---|
| Normal TTY | colors enabled (singleton) | colors enabled (isolated) |
| Agent (Claude Code, etc.) | `disableDefaultColors()` nukes singleton | isolated instance has colors disabled, singleton intact |
| User code tinyrainbow | silently broken in agent env | always works |

### DIFF
--- a/packages/vitest/src/node/cli/cac.ts
+++ b/packages/vitest/src/node/cli/cac.ts
@@ -5,11 +5,18 @@ import type { CLIOption, CLIOptions as CLIOptionsConfig } from './cli-config'
 import { toArray } from '@vitest/utils/helpers'
 import cac from 'cac'
 import { normalize } from 'pathe'
-import c, { disableDefaultColors } from 'tinyrainbow'
+import { createColors } from 'tinyrainbow'
 import { version } from '../../../package.json' with { type: 'json' }
 import { isAgent } from '../../utils/env'
 import { benchCliOptionsConfig, cliOptionsConfig, collectCliOptionsConfig } from './cli-config'
 import { setupTabCompletions } from './completions'
+
+// Use an isolated color instance for Vitest's own CLI output when running
+// inside an agent environment (Claude Code, Cursor, etc.). This avoids
+// mutating the shared tinyrainbow singleton via disableDefaultColors(), which
+// would also disable colors in user code under test that imports tinyrainbow.
+// See: https://github.com/vitest-dev/vitest/issues/10046
+const c = createColors({ force: !isAgent })
 
 function addCommand(cli: CAC | Command, name: string, option: CLIOption<any>) {
   const commandName = option.alias || name
@@ -75,10 +82,6 @@ function addCliOptions(cli: CAC | Command, options: CLIOptionsConfig<any>) {
 }
 
 export function createCLI(options: CliParseOptions = {}): CAC {
-  if (isAgent) {
-    disableDefaultColors()
-  }
-
   const cli = cac('vitest')
 
   cli.version(version)
--- a/packages/vitest/src/node/cli/cac.ts
+++ b/packages/vitest/src/node/cli/cac.ts
@@ -11,12 +11,14 @@ import { isAgent } from '../../utils/env'
 import { benchCliOptionsConfig, cliOptionsConfig, collectCliOptionsConfig } from './cli-config'
 import { setupTabCompletions } from './completions'
 
-// Use an isolated color instance for Vitest's own CLI output when running
-// inside an agent environment (Claude Code, Cursor, etc.). This avoids
-// mutating the shared tinyrainbow singleton via disableDefaultColors(), which
-// would also disable colors in user code under test that imports tinyrainbow.
+// Use an isolated color instance for Vitest's own CLI output.
+// In agent environments (Claude Code, Cursor, etc.) we explicitly disable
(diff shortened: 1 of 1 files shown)


wiliyam:
Fixed the CI failure — the test `agent` expects Vitest's own reporter to suppress ANSI colors in agent mode. Updated the fix to use `createColors({ force: false })` explicitly in agent environments, and `createColors()` (auto-detect) in normal environments. This keeps Vitest's reporter color-free in agent mode while not touching the shared singleton.

wiliyam:
Closing this PR — the fix is more complex than a single-file change.

The root cause is that Vitest's reporter layer (renderers/utils.ts and ~24 other files) all import the shared tinyrainbow singleton. A proper fix would require either:
1. Passing an isolated `createColors()` instance to all reporter/renderer files (~24 files), or
2. Running user test code in a separate module scope so it gets its own tinyrainbow instance

Both are non-trivial refactors that deserve a dedicated issue/design discussion. Our approach of patching only `cac.ts` doesn't affect the reporter (which is what produces the colored output), so the test keeps failing.

Thanks for the interesting problem — the root cause analysis in #10046 is correct, the fix scope is just larger than one file.

Issue #9762, opened 2026-03-01
Error: [vitest-pool]: Worker forks emitted error.


scottysseus:
### Describe the bug

I am attempting to run `vitest` with `bunx`.

```
bunx vitest
```

Before the tests even start, however, I get the following error:

```
bunx vitest

 DEV  v4.0.18 /home/.../...

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Unhandled Errors ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯

Vitest caught 1 unhandled error during the test run.
This might cause false positive tests. Resolve unhandled errors to make sure your tests are not affected.

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Unhandled Error ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
Error: [vitest-pool]: Worker forks emitted error.
 ❯ EventEmitter.<anonymous> ../node_modules/vitest/dist/chunks/cli-api.B7PN_QUv.js:8043:22
 ❯ EventEmitter.emit ../node:events:508:28
 ❯ ChildProcess.emitUnexpectedExit ../node_modules/vitest/dist/chunks/cli-api.B7PN_QUv.js:7610:22
 ❯ ChildProcess.emit ../node:events:508:28
 ❯ Process.ChildProcess._handle.onexit ../node:internal/child_process:294:12

Caused by: Error: Worker exited unexpectedly
[...]

PR #10057, opened 2026-04-03
[Fix]: Remove worker error listener early during forks pool shutdown

Morabbin:
## Description

During `PoolRunner.stop()`, the `exit` listener is removed early (line 271) to avoid spurious "Worker exited unexpectedly" errors during graceful shutdown. However, the `error` listener (`emitWorkerError`) remains active throughout the entire shutdown sequence -- stop message, wait for response, SIGTERM, wait for exit.

If the worker process encounters IPC errors during this window (e.g., EPIPE when the parent sends a stop message to a process that already exited, or broken IPC channel during SIGTERM), the error listener fires and propagates the error to the pool's error handler, which rejects the task resolver. This causes spurious test failures even when all tests passed.

This is especially common on Windows, where process termination during IPC communication is more likely to produce EPIPE/ERR_IPC_CHANNEL_CLOSED errors.

## Fix

Remove the `error` listener at the same point as the `exit` listener, early in the `stop()` sequence. The `message` listener is intentionally kept active because it is needed to receive the `stopped` response from the worker.

```diff
  this.worker.off('exit', this.emitUnexpectedExit)
+ this.worker.off('error', this.emitWorkerError)
```

## Real-world impact

This bug causes intermittent CI failures across multiple public repositories using vitest 4.x with the default forks pool. The error surfaces as:

```
Error: [vitest-pool]: Worker forks emitted error.
Caused by: Error: Worker exited unexpectedly
```

### kubestellar/console (vitest ^4.1.2, Linux CI)

Four separate nightly CI failures from this exact error, recurring roughly weekly:

- [kubestellar/console#4215](https://github.com/kubestellar/console/issues/4215) (Apr 2, 2026) -- [CI run](https://github.com/kubestellar/console/actions/runs/23887254784)
- [kubestellar/console#4088](https://github.com/kubestellar/console/issues/4088) (Apr 1, 2026) -- [CI run](https://github.com/kubestellar/console/actions/runs/23835458264)
- [kubestellar/console#2188](https://github.com/kubestellar/console/issues/2188) (Mar 11, 2026) -- [CI run](https://github.com/kubestellar/console/actions/runs/22939609692)
- [kubestellar/console#2057](https://github.com/kubestellar/console/issues/2057) (Mar 8, 2026) -- [CI run](https://github.com/kubestellar/console/actions/runs/22810006340)

All show the identical stack trace through `emitUnexpectedExit` in the forks pool runner, with tests passing but the run marked as failed.

### Other affected repos

- [openclaw/openclaw#2077](https://github.com/openclaw/openclaw/pull/2077) -- macOS vitest worker crash, references #8564
- [sanity-io/pkg-utils#2494](https://github.com/sanity-io/pkg-utils/pull/2494) -- had to disable file parallelism when upgrading to vitest 4.x

## Related issues

- #9762 -- same `Worker forks emitted error` / `Worker exited unexpectedly` stack trace
- #8564 -- `p5-urgent` regression introduced by the tinypool-removal pool rewrite in #8705

## Reproduction

**Repo:** [Morabbin/vitest-forks-shutdown-race-repro](https://github.com/Morabbin/vitest-forks-shutdown-race-repro)

### Deterministic mode

The repo includes `inject-bug.sh`, which patches vitest to check whether `emitWorkerError` is still registered on the ChildProcess at the point where the stop message is about to be sent. If it is, the bug is present and any IPC error during this window would propagate as a spurious failure.

```bash
git clone https://github.com/Morabbin/vitest-forks-shutdown-race-repro
cd vitest-forks-shutdown-race-repro
npm ci

# Unpatched: FAIL (exit 1) -- emitWorkerError still attached during stop()
bash inject-bug.sh
bash stress-test.sh deterministic

# Patched: PASS (exit 0) -- emitWorkerError removed before stop message
npm ci
bash apply-fix.sh
bash inject-bug.sh
bash stress-test.sh deterministic
```

The [GitHub Actions workflow](https://github.com/Morabbin/vitest-forks-shutdown-race-repro/actions/runs/23977807653) confirms this on both Windows and Linux:

| Job | Result |
|-----|--------|
| Unpatched (windows-latest) | `[BUG] emitWorkerError still attached (5/5 workers)` -- FAIL |
| Unpatched (ubuntu-latest)  | `[BUG] emitWorkerError still attached (5/5 workers)` -- FAIL |
| Patched (windows-latest)   | `[FIX] emitWorkerError removed (5/5 workers)` -- PASS |
| Patched (ubuntu-latest)    | `[FIX] emitWorkerError removed (5/5 workers)` -- PASS |

### Stress test mode

The repo also supports a stress mode that runs vitest repeatedly to try to trigger the race naturally:

```bash
npm ci
bash stress-test.sh stress 50
```

This is timing-dependent and requires high-core runners (e.g., 16-core Windows) to trigger reliably. On 2-core `windows-latest` runners, 50 runs of 500 test files did not trigger the race, which is consistent with the bug manifesting primarily on larger CI runners under load.

## Regression test

A unit test is included in this PR (`test/cli/test/forks-shutdown-race.test.ts`) that deterministically verifies the fix. It creates a mock `PoolWorker` that tracks listener registrations on its `EventEmitter` and records the error listener count at the exact moment the `stop` message arrives -- this is the window where IPC errors would fire on a real `ChildProcess`.

### Running the test (with the fix)

```bash
pnpm install && pnpm run build
pnpm -C test/cli test forks-shutdown-race
```

Expected output: **PASS** -- every worker reports `stop-received:error-listeners=0`, confirming the error listener was removed before the shutdown window.

### Verifying it catches the bug (without the fix)

To confirm the test would fail on the current `main` branch, temporarily revert the one-line fix:

```bash
# Revert the fix (remove the added line)
sed -i'' -e "/this\.worker\.off('error', this\.emitWorkerError)/d" \
  packages/vitest/src/node/pools/poolRunner.ts

pnpm run build
pnpm -C test/cli test forks-shutdown-race
```

Expected output: **FAIL** -- workers report `stop-received:error-listeners=1`, meaning the error listener is still active when the stop message is sent. Any IPC error during this window propagates as a spurious "Worker exited unexpectedly" failure.

```
# Restore the fix
git checkout packages/vitest/src/node/pools/poolRunner.ts
```


AriPerkkio (commented):
Do you have reproduction repository where we could see the error, and see that this fixes it?

Morabbin:
> Do you have reproduction repository where we could see the error, and see that this fixes it?

We see it in CI in internal github repos, but I see if I can make a repro (or find an example in our public repos) and show that this fixes it. It also needs a robust test or two. Lastly, this is so far totally copilot-generated, and I haven't had a chance to properly review it, nor have any of my colleagues. Given all that, I'd like to keep it in draft for now.

Morabbin:
> Do you have reproduction repository where we could see the error, and see that this fixes it?

See the updated PR description for some examples in the wild, and details on a test repro repo. The failure only happens about 15% of the time, so it's tricky to repro.

**Root cause**: In `PoolRunner.stop()`, the exit listener is removed early (line 271) to avoid spurious "Worker exited unexpectedly" errors during graceful shutdown, but the error listener (`emitWorkerError`) stays active until `removeAllListeners()` much later (line 308/313). During this window -- stop message, wait for response, SIGTERM, wait for exit -- any IPC/EPIPE error from the dying ChildProcess propagates through `emitWorkerError` to the pool's error handler, which rejects an already-resolved task resolver. This is especially common on Windows.

**Fix**: One line -- also remove the error listener early alongside the exit listener. The message listener is intentionally kept because it's needed to receive the stopped response.

**Instances**:
* **Real-world**: [kubestellar/console](https://github.com/kubestellar/console) has 4 separate nightly CI failures from this exact bug on vitest ^4.1.2 (#4215, #4088, #2188, #2057), recurring roughly weekly. Also affects openclaw/openclaw and sanity-io/pkg-utils.
* **Related issues**: #9762 (same stack trace), #8564 (`p5-urgent` regression from tinypool-removal rewrite in #8705).

**Reproduction**: Public repo at [Morabbin/vitest-forks-shutdown-race-repro](https://github.com/Morabbin/vitest-forks-shutdown-race-repro) with two modes:
  * **Deterministic**: `inject-bug.sh` patches vitest to check whether emitWorkerError is still on the ChildProcess when the stop message is sent, simulating the error conditions. Unpatched = [BUG], patched = [FIX]. [GitHub Actions confirms](https://github.com/Morabbin/vitest-forks-shutdown-race-repro/actions/runs/23977807653) on both Windows and Linux.
  * **Stress test**: Runs `vitest` repeatedly to trigger the race naturally (timing-dependent, needs high-core runners).
Regression test: `test/cli/test/forks-shutdown-race.test.ts` uses a mock PoolWorker that tracks listener registrations and asserts error listener count is 0 when the stop message arrives. Fails without the fix (error-listeners=1), passes with it (error-listeners=0).

sheremet-va:
> * **Real-world**

Linked issues don't seem to have anything to do with this? Many of them are dated before we rewrote the pool, and one is about a `todo` mode in _benchmarking_ from _2022_. Does it link the wrong repo?

AriPerkkio (commented):
This is quite confusing. But it's understandable as this is complex part of Vitest's codebase. Couple of questions below.

1. There are two different cases mentioned here: `Worker exited unexpectedly` and `ERR_IPC_CHANNEL_CLOSED`. Which one is it? One of these happens on `exit`, another one on the `error`. Are you 100% sure both of these errors are related in this bug, or did you pick one of them from some other unrelated Github issue?

2. Is `ERR_IPC_CHANNEL_CLOSED` on main Vitest process/thread or on the worker side? Is this code not working as expected?

https://github.com/vitest-dev/vitest/blob/691d341fd3370336ac5150428cb8f589c9697001/packages/vitest/src/runtime/workers/init-forks.ts#L60-L66

Maybe we should not `off` it? Can you try just removing the listener unsubscribing and see if it fixes anything?

https://github.com/vitest-dev/vitest/blob/691d341fd3370336ac5150428cb8f589c9697001/packages/vitest/src/runtime/workers/init-forks.ts#L41-L44

It's important to understand which side of the IPC is trying to call the closed one. If it's the main process, there is something more serious broken. If it's on the worker side, we are just clearing the error handler too early.

3. If we had a reproduction we should check what message is causing the `ERR_IPC_CHANNEL_CLOSED`. Fixing this blindly might work for now but it's not proper fix in long term.


Thanks for all the debugging so far.

Morabbin:
> > * **Real-world**
> 
> Linked issues don't seem to have anything to do with this? Many of them are dated before we rewrote the pool, and one is about a `todo` mode in _benchmarking_ from _2022_. Does it link the wrong repo?

The links to kubestellar/console are PRs from March/April 2026 in that repo which seem to exhibit the problem (same error message). The "related issue" links could well be bogus though; copilot found them and I haven't checked them diligently.

Morabbin:
Thanks for the thorough review! I think you're right, that the situation is more complex than I originally thought.

##### 1. `Worker exited unexpectedly` vs `ERR_IPC_CHANNEL_CLOSED`

My thinking is that the worker throws the error, which ends up looking like this:
```
Error: [vitest-pool]: Worker forks emitted error.
Caused by: Error: Worker exited unexpectedly
```
So it's an unexpected exit, caused by an error emission, which ends up being caught by the exit listener. I had thought it was the error listener, but those look different.

##### 2. `ERR_IPC_CHANNEL_CLOSED` coming from worker or main process?

I think this involves work-side crashes, but you're spot on: we need to know which side is actually causing the errors.

##### 3. Repro!

The parent-side fix in this PR is an improvement (the error listener shouldn't be active during the stop window), but yeah, until we know what's actually happening, this is likely premature. The flow in point 1 above isn't fixed by this, anyway.

I agree we need a proper repro. Running the test repo stress test on 16 core Windows machines replicates our internal environment; hopefully, that will be more likely to repro. It could be due to something outside of vitest ofc.

Putting this PR back in draft for now; I'll take another run at making a good repro. I'll do that internally for now, as that's the easiest way for me to access beefy windows boxes.

Morabbin:
Closing this for now, as I'm about to go on holiday and haven't managed a proper repro.

PR #10059, opened 2026-04-04
chore: add agent instruction for pr creation

hi-ogawa:
### Description

- Inspired by https://github.com/vercel/next.js/pull/92235
- Demo PR https://github.com/vitest-dev/vitest/pull/10060

The demo PR successfully included "VITEST_AUTOMATED_PR" though the task might've been too easy (especially I forced "no test" since otherwise it takes a while).

<details>
<summary>Claude Code session for demo</summary>

```sh
 ▐▛███▜▌   Claude Code v2.1.92
▝▜█████▛▘  Sonnet 4.6 with medium effort · Claude Max
  ▘▘ ▝▝    ~/code/others/vitest

❯ let's tackle this issue                                              
https://github.com/vitest-dev/vitest/issues/10053                      
looks like what's suggested in https://github.com/vitest-dev/vitest/is 
sues/10053#issuecomment-4182371693 is a reasonable path.               
implement a fix and then create pr. no test needed.                    

● Explore(Fetch GitHub issue and comments)
  ⎿  Done (1 tool use · 15.7k tokens · 13s)
  (ctrl+o to expand)

  Searched for 1 pattern, read 2 files (ctrl+o to expand)

● The fix is clear: add a hasTTY check so colors aren't disabled when
  there's an actual TTY (like in Kiro IDE's integrated terminal).

● Update(packages/vitest/src/node/cli/cac.ts)
  ⎿  Added 1 line, removed 1 line
      75  }
      76  
      77  export function createCLI(options: CliParseOptions = {}): CAC {
      78 -  if (isAgent) {                                                    
      78 +  if (isAgent && !process.stdout?.isTTY) {                          
      79      disableDefaultColors()
      80    }
      81  


● Now let me check if there are other places where isAgent triggers
  color disabling:

  Searched for 1 pattern, read 1 file (ctrl+o to expand)

● This one also needs the TTY check:

  Read 1 file (ctrl+o to expand)

● Update(packages/vitest/src/runtime/workers/init.ts)
  ⎿  Added 1 line, removed 1 line
      50          process.env.VITEST_WORKER_ID = String(message.workerId)
      51          reportMemory = message.options.reportMemory
      52  
      53 -        if (message.context.config.isAgent) {                       
      53 +        if (message.context.config.isAgent && !process.stdout?.isTTY
         +) {                                                                 
      54            disableDefaultColors()
      55          }
      56  

● Now let me run lint:

● Bash(pnpm lint:fix 2>&1 | tail -20)
[...]

PR #10060, opened 2026-04-04
fix: skip disabling colors when stdout is a TTY even if running as an agent

### LINKED ISSUE #10053: Colors are being disabled improperly within the kiro ide integrated terminal
### Describe the bug

When I updated to the latest version of vitest, I noticed my intergrated terminal stopped showing test result output in color.

I think the underlying problem is there doesn't seem to be a reliable way of distinguishing between kiro ide and the kiro cli.

See: 
- https://github.com/vitest-dev/vitest/pull/9851 (PR that introduced this bug)
- https://github.com/unjs/std-env/issues/185#issuecomment-4180728160 (lib that powers vitest's agent detection

I think the new agent based de-coloring should be opt in, or at least provide an escape hatch to modify this behavior.

### Reproduction

Using the [kiro ide](https://kiro.dev/) run your vitest tests in the integrated terminal.

### System Info

```shell
vitest 4.1.2
```

### Used Package Manager

npm

### Validations

hi-ogawa:
## Summary

- When `isAgent` is detected but `process.stdout.isTTY` is truthy (e.g. Kiro IDE integrated terminal), colors are no longer disabled
- Applied the same TTY guard in both the CLI entry point (`cac.ts`) and worker initialization (`init.ts`)

Fixes #10053

### DIFF
--- a/.github/workflows/issue-labeled.yml
+++ b/.github/workflows/issue-labeled.yml
@@ -11,7 +11,6 @@ permissions:
   issues: write
   pull-requests: write
 
-
 jobs:
   reply-labeled:
     runs-on: ubuntu-latest
--- a/.github/workflows/pr-labeled-automated.yml
+++ b/.github/workflows/pr-labeled-automated.yml
@@ -1,0 +1,21 @@
+name: Label Automated PR
+
+on:
+  pull_request:
+    types: [opened, edited]
+
+permissions:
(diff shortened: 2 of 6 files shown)


hi-ogawa:
This was a demo PR to test agent instruction from https://github.com/vitest-dev/vitest/pull/10059

PR #10062, opened 2026-04-04
fix: avoid mutating tinyrainbow default colors in agent mode

### LINKED ISSUE #10046: Agent detection disables colors in user code via shared tinyrainbow singleton

teee32:
Problem

In agent environments, Vitest disabled colors by mutating `tinyrainbow`'s shared default export. That leaked into user code running under test, so code importing `tinyrainbow` could unexpectedly lose ANSI output.

Reproduction / Observation

Issue #10046 reports that agent detection turns `tinyrainbow` color functions into identity functions for user code. This change adds a regression that runs Vitest in agent mode and verifies user code can still observe `tinyrainbow` color output while Vitest reporter output remains uncolored.

Root cause

Vitest called `disableDefaultColors()` on the shared `tinyrainbow` singleton during CLI and worker startup. Because user code and Vitest resolve the same default export, that mutation escaped Vitest internals.

Change

- remove the global `disableDefaultColors()` calls from CLI and worker startup
- introduce a Vitest-local colors wrapper in `packages/vitest/src/utils/colors.ts`
- switch Vitest internal color consumers in `packages/vitest/src/**` to use that local wrapper
- add an agent-mode regression fixture/test covering user-imported `tinyrainbow`

Why this fix

This preserves the existing agent-mode intent for Vitest’s own output, but stops mutating shared library state that user code depends on. It is a narrow internal fix with a direct regression test.

Risk

Low to moderate. The behavior change is limited to how Vitest sources internal colors in agent mode. No public API changes are involved.

Validation

- `cd vitest/test/config && pnpm exec vitest run test/console-color.test.ts -t 'agent'`
- `cd vitest && pnpm --filter vitest build`
- `cd vitest && pnpm exec eslint packages/vitest/src/utils/colors.ts packages/vitest/src/node/cli/cac.ts packages/vitest/src/runtime/workers/init.ts test/config/test/console-color.test.ts test/config/fixtures/console-color-agent-user-code/basic.test.ts test/config/fixtures/console-color-agent-user-code/vitest.config.ts --no-ignore`
- `cd vitest && git diff --check`

Closes #10046


sheremet-va (changes requested):
This is AI slop. The issue is bigger - we use tinyrainbow in multiple packages, including expect, pretty-format and utils

teee32:
Thanks — you were right that the first pass was too narrow. I pushed a follow-up that routes agent-mode color handling through detached wrappers across the shared internal paths called out in review, including utils / expect / pretty-format, plus the remaining internal tinyrainbow consumers that participate in agent-sensitive output. I also added a regression covering failed matcher output in agent mode while keeping user-imported tinyrainbow colors enabled.

PR #10090, opened 2026-04-07
fix(snapshot): fail test when snapshot assertion is used with `test.fails`

### LINKED ISSUE #9495: Dont update snapshot inside `it.fails`

I think this would be a feature request instead of a bug report. But IMO snapshots should not update when used inside a test that is marked with `.fails`. 

### Suggested solution

Even if `vitest run -u` is used, all snapshots inside the `.fails` directory should be checked but not updated.

### Alternative

_No response_

### Additional context

https://github.com/nuxt/nuxt/pull/34125

### Validations

hi-ogawa:
### Description

- Closes https://github.com/vitest-dev/vitest/issues/9495

`expect.soft` handling is a bit annoying :thinking: I ended up with using `error.__vitest_test_syntax_error__` to allow communicate between expect/runner/vitest packages basically.

This is likely a breaking change since users can do really weird thing with snapshot.

### Tests

### Documentation

### Changesets

### DIFF
--- a/.github/workflows/issue-labeled.yml
+++ b/.github/workflows/issue-labeled.yml
@@ -9,6 +9,7 @@ on:
 # for actions-cool/issues-helper to update issues
 permissions:
   issues: write
+  pull-requests: write
 
 jobs:
   reply-labeled:
--- a/.github/workflows/pr-labeled-automated.yml
+++ b/.github/workflows/pr-labeled-automated.yml
@@ -6,6 +6,7 @@ on:
 
 permissions:
   issues: write
+  pull-requests: write
 
 jobs:
   label:
--- a/docs/config/experimental.md
+++ b/docs/config/experimental.md
@@ -479,3 +479,36 @@ export default {
 If module runner is disabled, Vitest uses a native [Node.js module loader](https://nodejs.org/api/module.html#customization-hooks) to transform files to support `import.meta.vitest`, `vi.mock` and `vi.hoisted`.
 
 If you don't use these features, you can disable this to improve performance.
+
+## experimental.preParse <Version type="experimental">4.1.3</Version> {#experimental-preparse}
+
+- **Type:** `boolean`
+- **Default:** `false`
+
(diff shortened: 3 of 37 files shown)


sheremet-va (approved):

sheremet-va:
> This is likely a breaking change since users can do really weird thing with snapshot.

I'm not so sure it's a breaking change. What kind of use cases does this break?

hi-ogawa:
> > This is likely a breaking change since users can do really weird thing with snapshot.
> 
> I'm not so sure it's a breaking change. What kind of use cases does this break?

I removed one test from `test/core/test/snapshot.test.ts` and that's the one.

sheremet-va:
> > > This is likely a breaking change since users can do really weird thing with snapshot.
> > 
> > 
> > I'm not so sure it's a breaking change. What kind of use cases does this break?
> 
> I removed one test from `test/core/test/snapshot.test.ts` and that's the one.

Oh, I don't think this PR fixes the issue then. I think`toMatchSnapshot` should still work, it just shouldn't update the snapshot

hi-ogawa:
> > > > This is likely a breaking change since users can do really weird thing with snapshot.
> > > 
> > > 
> > > I'm not so sure it's a breaking change. What kind of use cases does this break?
> > 
> > 
> > I removed one test from `test/core/test/snapshot.test.ts` and that's the one.
> 
> Oh, I don't think this PR fixes the issue then. I think`toMatchSnapshot` should still work, it just shouldn't update the snapshot

`test.fails` outright rejecting snapshot matchers makes sense to me. This is basically same as how we reject `expect(...).not.toMatchSnapshot()`. Also that's at least what `TestSyntaxError` should entail. It shouldn't dependent on runtime surrounding.

sheremet-va:
> `test.fails` outright rejecting snapshot matchers makes sense to me.

Why? The test can have other assertions that fail, and snapshot to pass:

```ts
test.fails('test', () => {
  expect(shape).toMatchInlineSnapshot(`{ name: 'Alice' }`)
  expect(otherShape).toEqual({ name: 'Bob' }) // fails! 
})
```

It should just ignore the snapshot error when `vitest` tries to _update_ it - at least that's what the issue is.

```ts
test.fails('test', () => {
  expect(shape).toMatchInlineSnapshot(`{ name: 'Bob' }`) // fails
})
vitest run -u // should ignore the error in `test.fails`, treat as --update=none in that test
```

hi-ogawa:
The reason to prefer reject everything is because supporting the case by case can be confusing when users end up reaching such invalid state during the development without directly modifying the test itself to introduce forbidden combination of `test.fails + snapshot`.

sheremet-va (approved):

PR #10092, opened 2026-04-07
docs: add "Learn" section

sheremet-va:
This PR adds a new "Learn" section meant to provide more context to new developers or those coming from a different test runner. Most of them could be useful in isolation too.

I have also cleaned up "Getting Started" by removing too advanced examples or moving them to other parts

- Writing Tests
- Using Matchers
- Testing Async Code
- Setup and Teardown
- Mock Functions
- Snapshot Testing
- Testing in Practice
- Debugging Tests
- Writing Tests with AI

This was originally generated with Claude and reviewed and improved by me. The documentation structure is based on Jest

### DIFF
--- a/docs/guide/index.md
+++ b/docs/guide/index.md
@@ -115,10 +115,18 @@ Learn more about [IDE Integrations](/guide/ide)
 |---|---|---|
 | `basic` | [GitHub](https://github.com/vitest-dev/vitest/tree/main/examples/basic) | [Play Online](https://stackblitz.com/fork/github/vitest-dev/vitest/tree/main/examples/basic?initialPath=__vitest__/) |
 | `fastify` | [GitHub](https://github.com/vitest-dev/vitest/tree/main/examples/fastify) | [Play Online](https://stackblitz.com/fork/github/vitest-dev/vitest/tree/main/examples/fastify?initialPath=__vitest__/) |
+| `in-source-test` | [GitHub](https://github.com/vitest-dev/vitest/tree/main/examples/in-source-test) | [Play Online](https://stackblitz.com/fork/github/vitest-dev/vitest/tree/main/examples/in-source-test?initialPath=__vitest__/) |
+| `lit` | [GitHub](https://github.com/vitest-dev/vitest/tree/main/examples/lit) | [Play Online](https://stackblitz.com/fork/github/vitest-dev/vitest/tree/main/examples/lit?initialPath=__vitest__/) |
 | `vue` | [GitHub](https://github.com/vitest-tests/browser-examples/tree/main/examples/vue) | [Play Online](https://stackblitz.com/fork/github/vitest-tests/browser-examples/tree/main/examples/vue?initialPath=__vitest__/) |
+| `marko` | [GitHub](https://github.com/vitest-tests/browser-examples/tree/main/examples/marko) | [Play Online](https://stackblitz.com/fork/github/vitest-tests/browser-examples/tree/main/examples/marko?initialPath=__vitest__/) |
+| `preact` | [GitHub](https://github.com/vitest-tests/browser-examples/tree/main/examples/preact) | [Play Online](https://stackblitz.com/fork/github/vitest-tests/browser-examples/tree/main/examples/preact?initialPath=__vitest__/) |
+| `qwik` | [GitHub](https://github.com/vitest-tests/browser-examples/tree/main/examples/qwik) | [Play Online](https://stackblitz.com/fork/github/vitest-tests/browser-examples/tree/main/examples/qwik?initialPath=__vitest__/) |
 | `react` | [GitHub](https://github.com/vitest-tests/browser-examples/tree/main/examples/react) | [Play Online](https://stackblitz.com/fork/github/vitest-tests/browser-examples/tree/main/examples/react?initialPath=__vitest__/) |
+| `solid` | [GitHub](https://github.com/vitest-tests/browser-examples/tree/main/examples/solid) | [Play Online](https://stackblitz.com/fork/github/vitest-tests/browser-examples/tree/main/examples/solid?initialPath=__vitest__/) |
 | `svelte` | [GitHub](https://github.com/vitest-tests/browser-examples/tree/main/examples/svelte) | [Play Online](https://stackblitz.com/fork/github/vitest-tests/browser-examples/tree/main/examples/svelte?initialPath=__vitest__/) |
+| `profiling` | [GitHub](https://github.com/vitest-dev/vitest/tree/main/examples/profiling) | Not Available |
 | `typecheck` | [GitHub](https://github.com/vitest-dev/vitest/tree/main/examples/typecheck) | [Play Online](https://stackblitz.com/fork/github/vitest-dev/vitest/tree/main/examples/typecheck?initialPath=__vitest__/) |
+| `projects` | [GitHub](https://github.com/vitest-dev/vitest/tree/main/examples/projects) | [Play Online](https://stackblitz.com/fork/github/vitest-dev/vitest/tree/main/examples/projects?initialPath=__vitest__/) |
 
 ## Community
 
--- a/docs/guide/index.md
+++ b/docs/guide/index.md
@@ -101,7 +101,7 @@ To run tests once without watching for file changes, use `vitest run`. You can a
 
 ## Configuring Vitest
 
-Vitest reads your `vite.config.ts` by default, so your existing Vite plugins and configuration work out-of-the-box. You can also create a dedicated `vitest.config.ts` for test-specific settings. See the [Config Reference](/config/) for details.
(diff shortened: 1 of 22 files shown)


docs/guide/index.md
```diff
@@ -237,65 +115,10 @@ Learn more about [IDE Integrations](/guide/ide)
...
-- [tdesign-vue-next](https://github.com/Tencent/tdesign-vue-next)
-- [cz-git](https://github.com/Zhengqbbb/cz-git)
-
-<!--
-For contributors:
-We no longer accept new entries to this list a this moment.
-Thanks for choosing Vitest!
--->
-
-## Using Unreleased Commits
```
sheremet-va, line 283: This is moved to CONTRIBUTING

docs/guide/learn/writing-tests.md
```diff
@@ -0,0 +1,203 @@
...
+
+## Grouping Tests with `describe`
+
+As your test files grow, you'll want to organize related tests together. [`describe`](/api/describe) creates a test suite, which is a named group of tests:
+
+```js
+import { describe, expect, it } from 'vitest'
+
+describe('Math.sqrt', () => {
+  it('returns the square root of perfect squares', () => {
```
hi-ogawa: Though it's mentioned that they are alias, test vs it should be made consistent across learn section.

docs/guide/learn/writing-tests.md
```diff
@@ -0,0 +1,203 @@
...
+```txt
+✓ src/utils.test.js (3 tests) 5ms
+  ✓ Math.sqrt 4ms
+    ✓ returns the square root of perfect squares 2ms
+    ✓ returns NaN for negative numbers 1ms
+    ✓ returns 0 for 0 1ms
+
+Test Files  1 passed (1)
+     Tests  3 passed (3)
+```
```
hi-ogawa: We can make it colorful by including ansi snippet. Probably it's worth the effort at least this learn section too.

For example, I used a script like this on slidev before:

```sh
script -q -c "pnpm -C examples/basic test --run" reporter-output.ansi
```

then vitepress can include it with

```
<<< @/snippet/reporter-output.ansi
```

Example: https://github.com/vuejs/vitepress/blob/19357f9d337472572a500b8d2af8ef97932bfdda/docs/en/guide/getting-started.md?plain=1#L74

docs/guide/learn/writing-tests.md
```diff
@@ -0,0 +1,203 @@
...
+
+```js
+import { expect, it } from 'vitest'
+
+it('should compute square roots', () => {
+  expect(Math.sqrt(4)).toBe(2)
+})
+```
+
+Use whichever you prefer. Both work the same way, and you can mix them freely in a project. If you want to enforce a consistent choice across your codebase, the [`consistent-test-it`](https://github.com/vitest-dev/eslint-plugin-vitest/blob/main/docs/rules/consistent-test-it.md) lint rule (also available in [oxlint](https://oxc.rs/docs/guide/usage/linter/rules/jest/consistent-test-it.html)) can help with that.
```
hi-ogawa: Can this alias mention put as folded section, so it can be naturally skipped without breaking flow? This feels a bit too verbose to have from the get-go.

hi-ogawa (commented):
The structure looks good to me.
`writing-tests-with-ai.html` looks quite new, but can this be separate from the PR? I feel this would need a separate aspect of review.

docs/guide/learn/async.md
```diff
@@ -0,0 +1,136 @@
...
+
+By default, each test has a 5-second timeout. If a test takes longer than that (perhaps because a promise never resolves, or a network request hangs), it will fail with a timeout error. This prevents your test suite from getting stuck indefinitely.
+
+You can set a [custom timeout](/api/test#timeout) as the third argument to `test`, which is useful for tests that legitimately need more time:
+
+```js
+test('long-running operation', async () => {
+  await someSlowOperation()
+}, 10_000) // 10 seconds
+```
```
AriPerkkio, line 81: Should we rather recommend using object with `timeout` instead?


```suggestion
```js
test('long-running operation', { timeout: 10_000 }, async () => {
  await someSlowOperation()
})
```
```

File: docs/guide/learn/writing-tests.md (hunk 3 above)
Comment (sheremet-va) [reply docs/guide/learn/writing-tests.md]: Didn't know it's supported!

File: docs/guide/learn/writing-tests.md (hunk 4 above)
Comment (sheremet-va) [reply docs/guide/learn/writing-tests.md]: do you mean the eslint part of the whole "use whichever you prefer" part?

File: docs/guide/learn/async.md (hunk 5 above)
Comment (sheremet-va) [reply docs/guide/learn/async.md:81]: Why? The timeouts as a third argument is not deprecated and supported widely in formatters

Comment (AriPerkkio) [reply docs/guide/learn/async.md:81]: I remember you specifically talking about enforcing the object format instead of using the third argument as timeout long time ago. If that's no longer the case, let's leave it like this.

File: docs/guide/learn/writing-tests.md (hunk 4 above)
Comment (hi-ogawa) [reply docs/guide/learn/writing-tests.md]: To me, whole `test vs it` part can be folded. Something like this 

<details><summary>Use <code>test</code> or <code>it</code>?</summary>

You might also see... 

</details>

File: docs/guide/learn/async.md (hunk 5 above)
Comment (sheremet-va) [reply docs/guide/learn/async.md:81]: I personally like it as the last option (familiar API). If we want to deprecate it, I think we can talk about it on weekly

File: docs/guide/learn/writing-tests.md (hunk 3 above)
Comment (sheremet-va) [reply docs/guide/learn/writing-tests.md]: Added snippets

File: docs/guide/learn/async.md [hunk 6]
```diff
@@ -0,0 +1,136 @@
...
+  await expect(fetchUser(1)).resolves.toMatchObject({ name: 'Alice' })
+})
+
+test('rejects with an error', async () => {
+  await expect(fetchInvalidUser()).rejects.toThrow('User not found')
+})
+```
+
+::: warning
+Don't forget the `await` before `expect`. Vitest will detect unawaited assertions and print a warning at the end of the test, but it's best to always include `await` explicitly. Vitest will also wait for all pending promises in `Promise.all` before starting the next test, but relying on this behavior makes tests harder to understand.
```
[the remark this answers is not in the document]
sheremet-va: They are awaited by `recordAsyncExpect`

docs/guide/learn/snippets/test-output-single.ansi
```diff
@@ -0,0 +1,8 @@
+ [32m✓[39m src/utils.test.js [90m(3 tests)[39m [32m5ms[39m
+   [32m✓[39m Math.sqrt [32m4ms[39m
+     [32m✓[39m returns the square root of perfect squares [32m2ms[39m
+     [32m✓[39m returns NaN for negative numbers [32m1ms[39m
+     [32m✓[39m returns 0 for 0 [32m1ms[39m
+
+ [1mTest Files[22m  [32m1 passed[39m (1)
+ [1m     Tests[22m  [32m3 passed[39m (3)
```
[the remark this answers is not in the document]
sheremet-va: Lol

docs/guide/learn/writing-tests-with-ai.md
```diff
@@ -0,0 +1,145 @@
...
+
+### Are there real edge cases?
+
+AI tools tend to generate happy-path tests and skip the hard cases. After reviewing the generated tests, ask yourself: what happens with empty input? What about `null` or `undefined`? What if the network request fails? What if the list is empty?
+
+If these scenarios aren't covered, ask the AI to add them, or write them yourself.
+
+### Are snapshot tests overused?
+
+AI loves snapshot tests because they're easy to generate: just call `toMatchSnapshot()` on everything. But snapshots should be used deliberately for structured output (HTML, error messages, serialized data), not as a lazy substitute for specific assertions.
```
hi-ogawa: Isn't this opposite? They prefer weaker test without snapshot and they avoid snapshot because they often feels they need to write the snapshot value "by hand" instead of generating by `--update`.

docs/guide/learn/testing-in-practice.md
hi-ogawa, line 1: Not blocking, but this looks like a good place to also mention bug fix scenario where it describes three steps of 
- write reproduction test case
- fix 
- confirm test passes

This is something agents should also do to help avoiding hallcinated fix.

docs/guide/learn/writing-tests-with-ai.md
sheremet-va: Yeah, in my experience it prefers to use `toContain` over anything else

sheremet-va: I removed the section. the weaker test is covered by "Do the tests actually assert something meaningful?" in my opinion

docs/guide/learn/testing-in-practice.md
sheremet-va: added the section

hi-ogawa (approved):
Looks good!

Issue #303, opened 2021-12-25
Bad code transformation for class fields if an import has the same name


Monkatraz:
### Describe the bug

Vitest seems to be generating a bad code transformation when a class field has the same name as an imported variable. This causes the script to fail.

### Reproduction

Sorry for the lack of precision in the reproduction here, but I have successfully logged the bad code.
Original (some bits snipped for clarity):
```ts
import { clearTimeout, Timeout, timeout } from "@wikijump/util"
import { Gesture, GestureObserver, type Direction } from "./gesture"

export interface SwipeOpts {
   // - snip -
}

const SWIPE_DEFAULT_OPTS: SwipeOpts = {
  // - snip -
}

export class SwipeObserver {
  /** Swipe recognition configuration. */
  private declare opts: SwipeOpts

  /** List of valid swipe directions. */
  private declare directions: Direction[]

  /** Internal {@link GestureObserver} used to get gesture information. */
  private declare observer: GestureObserver

  /** The current {@link Gesture}. */
  private gesture: Gesture | null = null

  /** The currently running {@link Timeout}. */
  private timeout: Timeout | null = null

  /** True if the swipe recognition is running. */
  private running = false

  /** True if the current swipe gesture is being ignored. */
  private ended = false

  /** The element being tracked for gestures. */
  declare target: HTMLElement
  
  // - snip -
  
}
```

Output that causes a script error: 
```js
const __vite_ssr_import_0__ = await __vite_ssr_import__("/modules/util/src/index.ts");

const __vite_ssr_import_1__ = await __vite_ssr_import__("/modules/dom/src/gesture.ts");

const SWIPE_DEFAULT_OPTS = {
  // -snip-
};
class SwipeObserver {
  gesture = null;
  __vite_ssr_import_0__.timeout = null;
  running = false;
  ended = false;
  // - snip -
}
```

Pretty sure this:
```js
__vite_ssr_import_0__.timeout = null;
```
is the problem.

Let me know if you need more details than this. I believe this is related to the import name, but if it turns out to be more subtle than this I can get a proper reproduction going.

### System Info

```shell
System:
    OS: Linux 5.15 Pop!_OS 21.10
    CPU: (12) x64 AMD Ryzen 5 5500U with Radeon Graphics
    Memory: 729.74 MB / 15.05 GB
    Container: Yes
    Shell: 5.8 - /usr/bin/zsh
  Binaries:
[...]

PR #10094, opened 2026-04-07
fix: Do not hoist imports whose names match class properties (fix #10093).

### LINKED ISSUE #10093: ReferenceError / crash when a class property inside a mocked module matches the name of an import (TypeScript + useDefineForClassFields)
### Describe the bug

When a mock defines a class with a property key that matches that of an import, vitest incorrectly hoists the imported reference above that mock, while leaving the instantiation below it.

The bug is triggered with the following conditions:
- Using typescript 6.0.2
- With "useDefineForClassFields" turned on in the tsconfig
- Mocking a class with a class property
- In a file that imports a function whose name matches the class property

The `useDefineForClassFields` setting is a requirement for this bug.  Without this setting, typescript writes the property set inside the constructor, where presumably vitest knows it is a property name and not a reference.  With this setting on, typescript emits code more-or-less as written.

I tried to reproduce this with pure javascript, but after a few tries I wasn't able to.  It seems to be typescript specific.

### Reproduction

Full reproduction: https://github.com/SunsetFi/vitest-class-method-name-collision/tree/main
Failing test in CI: https://github.com/SunsetFi/vitest-class-method-name-collision/actions/runs/24091793731/job/70279852955

Sample test:
```ts
const { myMethodMock } = vi.hoisted(() => ({
  myMethodMock: vi.fn(),
}));

vi.mock("./MyClass", () => ({
  MyClass: class {
    // This line triggers the crash
    myMethod = myMethodMock;
    // This workaround fixes it:
    // ["myMethod"] = myMethodMock;
  },
}));

// This incorrectly gets hoisted above the MyClass mock,
// presumably because vitest sees the property name and thinks its accessing
// this function.
import { myMethod } from "./my-method";

describe("myMethod", () => {
  it("should call MyClass.myMethod", () => {
    myMethodMock.mockReturnValue("Mocked Hello, World!");
    const result = myMethod();
    expect(myMethodMock).toHaveBeenCalled();
    expect(result).toBe("Mocked Hello, World!");
  });
});
```

This produces a ReferenceError, Cannot access `<import>` before initialization.  However, the line and character number are nonsense as the error occurs in generated code.
```
ReferenceError: Cannot access '__vi_import_0__' before initialization
 ❯ src/my-method.spec.ts:4:18
      2|
      3| const { myMethodMock } = vi.hoisted(() => ({
      4|   myMethodMock: vi.fn(),
       |                  ^
      5| }));
```

The bug seems agnostic to how the class itself is defined.  All of these trigger the bug:
```ts
vi.mock("./MyClass", () => ({
  MyClass: class {
    myMethod = myMethodMock;
  },
}));
```
```ts
vi.mock("./MyClass", () => ({
  MyClass: class MyClass {
    myMethod = myMethodMock;
  },
}));
```
```ts
vi.mock("./MyClass", () => {
  class MyClass {
    myMethod = myMethodMock;
  }
  return { MyClass };
});
```

Method declarations, however, do not trigger it.  The following works without issue:
```ts
vi.mock("./MyClass", () => ({
  MyClass: class {
    myMethod() {
      return myMethodMock();
    }
  },
}));
```
### System Info

```shell
System:
    OS: Linux 6.19 Arch Linux
    CPU: (16) x64 AMD Ryzen 7 7840HS w/ Radeon 780M Graphics
    Memory: 15.06 GB / 30.65 GB
    Container: Yes
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.14.1 - /home/sunset/.vite-plus/js_runtime/node/24.14.1/bin/node
    npm: 11.11.0 - /home/sunset/.vite-plus/js_runtime/node/24.14.1/bin/npm
    pnpm: 10.17.1 - /home/sunset/.local/share/pnpm/pnpm
    Deno: 2.7.11 - /usr/bin/deno
  Browsers:
    Firefox: 149.0
    Firefox Developer Edition: 149.0
  npmPackages:
    vitest: ^4.1.3 => 4.1.3
```

### Used Package Manager

pnpm

### Validations

SunsetFi:
fix: Do not hoist imports whose names match class properties (fix #10093).

### Description

Fixes vitest hoisting imports when a class property shares the same name.

Resolves #10093

[Full reproduction](https://github.com/SunsetFi/vitest-class-method-name-collision/tree/main)

When vitest 4.1.3 encounters a class property that matches the name of an import, it attempts to hoist the import above the class property, often resulting in reference errors:

```ts
const { myMethodMock } = vi.hoisted(() => ({
  myMethodMock: vi.fn(),
}));

vi.mock("./MyClass", () => ({
  MyClass: class {
    // This line triggers the crash
    myMethod = myMethodMock;
    // This workaround fixes it:
    // ["myMethod"] = myMethodMock;
  },
}));

// This incorrectly gets hoisted above the MyClass mock,
// presumably because vitest sees the property name and thinks its accessing
// this function.
import { myMethod } from "./my-method";

describe("myMethod", () => {
  it("should call MyClass.myMethod", () => {
    myMethodMock.mockReturnValue("Mocked Hello, World!");
    const result = myMethod();
    expect(myMethodMock).toHaveBeenCalled();
    expect(result).toBe("Mocked Hello, World!");
  });
});
```

Since the property name is not a reference to the import, the hoisting is unnecessary, and hoisting itself causes a ReferenceError as it accesses the vite import of the module before the actual import is performed.

I am familiar with javascript parsing, but not too familiar with estree or vitest's internals.  This PR is mostly a hunch, and I am not sure of the full ramifications of the change.  However, it fixes the issue in my bug example repo.

There is a test at test/core/test/injector-mock.test.ts that seems to explicitly test for the faulty behavior.  That is, it looks for the `add` and `remove` variables being hoisted above the class even though the class does not reference them.  Fixing this bug caused this test to fail, as now there is no need to hoist the variables at all.  I tweaked the test to reference the imports, and checked that the imports are still referenced correctly.  However, I am not sure why this test existed in its original form, and I don't know if I'm missing some reason as to why this original behavior is actually the desired one.

### Tests
- [x] Introduce a new test to check for this behavior

### Changesets

### DIFF
--- a/packages/mocker/src/node/esmWalker.ts
+++ b/packages/mocker/src/node/esmWalker.ts
@@ -311,6 +311,11 @@ function isRefIdentifier(id: Identifier, parent: _Node, parentStack: _Node[]) {
     return false
   }
 
+  // class property
+  if (parent.type === 'PropertyDefinition' && !parent.computed) {
+    return false
+  }
+
   // property key
   if (isStaticPropertyKey(id, parent)) {
     return false
--- a/packages/mocker/src/node/esmWalker.ts
+++ b/packages/mocker/src/node/esmWalker.ts
@@ -313,7 +313,8 @@ function isRefIdentifier(id: Identifier, parent: _Node, parentStack: _Node[]) {
 
   // class property
   if (parent.type === 'PropertyDefinition' && !parent.computed) {
-    return false
+    // Default values can still reference identifiers
+    return id === parent.value
   }
 
   // property key
--- a/packages/mocker/src/node/esmWalker.ts
+++ b/packages/mocker/src/node/esmWalker.ts
@@ -255,17 +255,13 @@ export function esmWalker(
   identifiers.forEach(([node, stack]) => {
     if (!isInScope(node.name, stack)) {
       const parent = stack[0]
-      const grandparent = stack[1]
       const hasBindingShortcut
         = isStaticProperty(parent)
           && parent.shorthand
           && (!isNodeInPattern(parent)
             || isInDestructuringAssignment(parent, parentStack))
 
-      const classDeclaration
-        = (parent.type === 'PropertyDefinition'
-          && grandparent?.type === 'ClassBody')
-        || (parent.type === 'ClassDeclaration' && node === parent.superClass)
+      const classDeclaration = (parent.type === 'ClassDeclaration' && node === parent.superClass)
 
       const classExpression
         = parent.type === 'ClassExpression' && node === parent.id
--- a/test/core/test/injector-mock.test.ts
+++ b/test/core/test/injector-mock.test.ts
@@ -775,7 +775,6 @@ add(4);
         update = update
         del = () => __vi_import_0__.del()
         call = __vi_import_0__.call(4)
-        timeout = null
       }
       
       __vi_import_0__.remove(2);
--- a/test/core/test/injector-mock.test.ts
+++ b/test/core/test/injector-mock.test.ts
@@ -754,6 +754,7 @@ class A {
   update = update
   del = () => del()
   call = call(4)
+  timeout = null
 }
 
 remove(2);
@@ -774,6 +775,7 @@ add(4);
         update = update
         del = () => __vi_import_0__.del()
         call = __vi_import_0__.call(4)
+        timeout = null
       }
       
       __vi_import_0__.remove(2);
--- a/test/core/test/injector-mock.test.ts
+++ b/test/core/test/injector-mock.test.ts
@@ -746,11 +746,12 @@ const obj = {
     expect(
       hoistSimpleCodeWithoutMocks(
         `
-import { remove, add } from 'vue'
+import { remove, add, update } from 'vue'
 
 class A {
   remove = 1
   add = null
+  update = update
 }
 
 remove(2);
@@ -763,10 +764,12 @@ add(4);
       import {vi} from "vitest";
 
 
-
+      
+      const update = __vi_import_0__.update;
       class A {
         remove = 1
         add = null
+        update = update
       }
       
       __vi_import_0__.remove(2);
--- a/test/core/test/injector-mock.test.ts
+++ b/test/core/test/injector-mock.test.ts
@@ -746,12 +746,14 @@ const obj = {
     expect(
       hoistSimpleCodeWithoutMocks(
         `
-import { remove, add, update } from 'vue'
+import { remove, add, update, del, call } from 'vue'
 
 class A {
   remove = 1
   add = null
   update = update
+  del = () => del()
+  call = call(4)
 }
 
 remove(2);
@@ -764,12 +766,14 @@ add(4);
       import {vi} from "vitest";
 
 
-      
+
       const update = __vi_import_0__.update;
       class A {
         remove = 1
         add = null
         update = update
+        del = () => __vi_import_0__.del()
+        call = __vi_import_0__.call(4)
       }
       
       __vi_import_0__.remove(2);
--- a/test/core/test/injector-mock.test.ts
+++ b/test/core/test/injector-mock.test.ts
@@ -767,11 +767,10 @@ add(4);
 
 
 
-      const update = __vi_import_0__.update;
       class A {
         remove = 1
         add = null
-        update = update
+        update = __vi_import_0__.update
         del = () => __vi_import_0__.del()
         call = __vi_import_0__.call(4)
       }
--- a/test/core/test/injector-mock.test.ts
+++ b/test/core/test/injector-mock.test.ts
@@ -754,7 +754,6 @@ class A {
   update = update
   del = () => del()
   call = call(4)
-  timeout = null
 }
 
 remove(2);
--- a/test/core/test/injector-mock.test.ts
+++ b/test/core/test/injector-mock.test.ts
@@ -752,6 +752,9 @@ class A {
   remove = 1
   add = null
 }
+
+remove(2);
+add(4);
 `,
       ),
     ).toMatchInlineSnapshot(`
@@ -761,12 +764,13 @@ class A {
 
 
 
-      const add = __vi_import_0__.add;
-      const remove = __vi_import_0__.remove;
       class A {
         remove = 1
         add = null
-      }"
+      }
+      
+      __vi_import_0__.remove(2);
+      __vi_import_0__.add(4);"
     `)
   })
 


sheremet-va:
> There is a test at test/core/test/injector-mock.test.ts that seems to explicitly test for the faulty behavior. That is, it looks for the `add` and `remove` variables being hoisted above the class even though the class does not reference them. Fixing this bug caused this test to fail, as now there is no need to hoist the variables at all. I tweaked the test to reference the imports, and checked that the imports are still referenced correctly. However, I am not sure why this test existed in its original form, and I don't know if I'm missing some reason as to why this original behavior is actually the desired one.

I think it is just being cautious. In your PR this is processed incorrectly:

```ts
hoistSimpleCodeWithoutMocks(
        `
import { remove, add, update } from 'vue'

class A {
  remove = 1
  add = null
  update = update // notice update is not transformed
}

remove(2);
add(4);
`,
      )
```

SunsetFi:
Thank you, I pushed a fix for that case and added it to the test.

SunsetFi:
I expanded the test with more access types:
```js
import { remove, add, update, del, call } from 'vue'

class A {
  // No hoisting - good
  remove = 1
  // No hoisting - good
  add = null
  // Hoisting - good
  update = update
  // No hoisting, fine as the function captures the scope
  // del = () => __vi_import_0__.del()
  del = () => del()
  // No hoisting - Is this correct?
  // call = __vi_import_0__.call(4)
  call = call(4)
}
```

Most cases look good, but I'm not sure if we want to hoist the invocation of `call` or not. I would have expected the tree walk to enter the property value node and handle this appropriately, but I'm not sure if this behavior is what it 'should' be.

Debugging the test, the CallExpression => 'call' does return true for a reference, and is added to the identifiers array, but the hoisting never occurs...

SunsetFi:
Well, that seems to have made the CI pass.

I'm trying to wrap my head around why the hoisting is needed at all.  I think the reason for it is to ensure the name is preserved for [NamedEvaluation](https://tc39.es/ecma262/multipage/syntax-directed-operations.html#sec-runtime-semantics-namedevaluation) in the spec; so that the name of the assignment inherited from the variable name is the same as the original code.  I'm not confident on that determination though.

But if that's true, I believe a NamedEvaluation isn't relevant for the property when its set to the result of a call expression, so I don't think hoisting `call` here is required.

If i'm wrong, though, the bug lies elsewhere in vitest, as esmWalker still calls onIdentifier for it, albeit with false for all 3 options.

sheremet-va:
It seems like the current behavior was added to fix this issue - https://github.com/vitest-dev/vitest/issues/303

SunsetFi:
Looking at the issue, I believe the fix introduced by #303 is preserved; the property 'add', which shares a name with an import, is not transformed as was happening prior to that fix.  The test still covers this case.

Interestingly, the hoisting seems almost incidental to the bug fix, which was that class property keys were being transformed as imports.  Unfortunately, that  still leaves me with no clear idea why the hoisting itself was required.

sheremet-va:
> Interestingly, the hoisting seems almost incidental to the bug fix, which was that class property keys were being transformed as imports. Unfortunately, that still leaves me with no clear idea why the hoisting itself was required.

I think hoisting is a side effect of the way the transform works - https://github.com/vitejs/vite/pull/6261 I'm not sure it was intended to be hoisted. The if check was initially applied only to `class` declarations

SunsetFi:
Ya, I think I agree this might be unintended.  All the code surrounding the classDeclaration usage does seem to only want class declarations, even beyond the name of the variable.

Reverting its computation to:
```js
const classDeclaration = (parent.type === 'ClassDeclaration' && node === parent.superClass)
```
doesn't seem to have broken anything, and the bug in #303 still remains fixed.

The condition itself makes much more sense for hoisting, as we want to pull up a superclass that the parent class is dependent on.  

The `test:ci` tests seem to mostly pass locally, but I am having difficulty getting playwright to run on my arch install.  I'll push it up experimentally to see if the GHA runner passes it.

SunsetFi:
There seems to be a test for the hoisting behavior [here](https://github.com/vitest-dev/vitest/blob/2fdcab875585a4ed8405aaf66926a1741d71e3e6/test/core/test/injector-mock.test.ts#L453):

```ts
  // #2221
  test('should declare variable for imported super class', () => {
    expect(
      hoistSimpleCodeWithoutMocks(
        `import { Foo } from './dependency';` + `class A extends Foo {}`,
      ),
    ).toMatchInlineSnapshot(`
      "vi.mock('faker');
      const __vi_import_0__ = await import("./dependency");
      import {vi} from "vitest";
      const Foo = __vi_import_0__.Foo;
      class A extends Foo {}"
    `)

    // exported classes: should prepend the declaration at root level, before the
    // first class that uses the binding
    expect(
      hoistSimpleCodeWithoutMocks(
        `import { Foo } from './dependency';`
        + `export default class A extends Foo {}\n`
        + `export class B extends Foo {}`,
      ),
    ).toMatchInlineSnapshot(`
      "vi.mock('faker');
      const __vi_import_0__ = await import("./dependency");
      import {vi} from "vitest";
      const Foo = __vi_import_0__.Foo;
      export default class A extends Foo {}
      export class B extends Foo {}"
    `)
  })
```

It's marked #2221, but I can't find that issue or PR in the vitest repository.

sheremet-va:
> It's marked #2221, but I can't find that issue or PR in the vitest repository.

This transform is based on Vite's implementation so most of the # reference vite's repo: https://github.com/vitejs/vite/pull/2221

SunsetFi:
Okay, so the purpose isn't actually hoisting, but just getting the appropriate reference.  That makes a lot more sense.

I definitely think aliasing the properties was an unintended bug by #303, and we can probably remove it.

So in this PR, esmWalker now correctly outputs identifiers used in the assignment portion of class properties, but not for the properties themselves.  It also now no longer indicates that such references are class declarations, resulting in the references no longer getting aliased.

I also think I was off base with my assumption that it was a named reference thing.  Thinking about it more, I believe the name for the named reference is always set by the assignment target, not the reference identifier.  So I don't think removing the aliasing is going to result in observable differences to tests, although I might be wrong there.

sheremet-va:
Can you also port the fix to Vite repo? It's here: https://github.com/vitejs/vite/blob/05803be47e88a0a0f2b792f0d9b819770bd23dbf/packages/vite/src/node/ssr/ssrTransform.ts#L367

SunsetFi:
> Can you also port the fix to Vite repo? It's here: https://github.com/vitejs/vite/blob/05803be47e88a0a0f2b792f0d9b819770bd23dbf/packages/vite/src/node/ssr/ssrTransform.ts#L367

Certainly, I will get to this in the next day.

---

I'm not sure what's causing tests to fail, only on macos, and only for a splattering of completely unrelated tests.

For the rolldown tests, the vite-ui tests are failing.  Its looking for a button, found it, but then waits too long to click it and it disappears?
```
Error: locator.click: Test timeout of 30000ms exceeded.
      Call log:
        - waiting for getByRole('button', { name: 'Run current file' })
          - locator resolved to <button rounded="" opacity="70" role="button" text-green-700="" data-v-7f38aded="" dark:text-green-500="" hover="bg-active op100" data-testid="btn-run-test" aria-label="Run current file" class="w-1.4em h-1.4em flex v-popper--has-tooltip">…</button>
        - attempting click action
          - waiting for element to be visible, enabled and stable
          - element is not stable
        - retrying click action
          - waiting for element to be visible, enabled and stable
          - element is not visible
        - retrying click action
          - waiting 20ms
          2 × waiting for element to be visible, enabled and stable
            - element is not visible
          - retrying click action
            - waiting 100ms
          50 × waiting for element to be visible, enabled and stable
             - element is not visible
           - retrying click action
             - waiting 500ms
```

And for Build&Test, it seems to be missing some stdout content:
```
test/pool.test.ts > threads > can capture worker's stdout and stderr 191ms
     → expected 'Worker writing to stderr' to contain 'MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 3 message listeners added to [TestFixturesCustomEmitter]'
```

Neither error reproduces locally for me, and neither is happening for the other OSes in the pipeline.

These tests seemed to have failed earlier, but all of them passed in at least one commit.  I want to blame test flakiness, especially because [b08d4b8](https://github.com/vitest-dev/vitest/pull/10094/commits/b08d4b8b82e06516185b51717692c14d95c0cc03) and [b937428](https://github.com/vitest-dev/vitest/pull/10094/commits/b93742830e8daca6864bfbc52d3af4c448698854) should be identical content.

travisbreaks (commented):
Clean fix for a subtle AST walker bug. The core insight is correct: in `PropertyDefinition`, only the `value` position can be a genuine identifier reference, not the `key`.

### The fix is sound

The two-part approach is right:
1. Removing `PropertyDefinition` from the `classDeclaration` variable stops the key from being treated as a class-level declaration reference
2. The new `isRefIdentifier` guard (`id === parent.value` when `!parent.computed`) correctly distinguishes key vs. value positions

### Computed property handling is correct

The `!parent.computed` guard ensures `[update] = 1` (computed property key using an import) is still treated as a reference, since computed keys are evaluated at runtime and genuinely reference the identifier. Good edge case handling.

### The test cases are well-chosen

The five patterns (`remove = 1`, `add = null`, `update = update`, `del = () => del()`, `call = call(4)`) cover: literal value, null value, same-name reference, closure reference, and invocation. The `update = update` case where key and value share the same identifier name is the trickiest, and it was caught during review (per sheremet-va's feedback).

### One edge case to consider

What about method definitions vs. property definitions? In:

```javascript
class A {
  remove() { return remove }
}
```

`MethodDefinition` has the same key/value structure as `PropertyDefinition`. The fix only targets `PropertyDefinition`. Worth checking whether `MethodDefinition` keys have the same false-positive issue, or if they are already handled correctly via a different code path.

### Upstream

+1 on sheremet-va's suggestion to port this to Vite as well since the walker is based on vitejs/vite#6261. The fix would benefit both ecosystems.

sheremet-va (approved):

PR #10095, opened 2026-04-07
feat(experimental): expose `assertion` as a public field

sheremet-va:
### Description

I think this might be useful as a public field


hi-ogawa (approved):
👍

PR #10113, opened 2026-04-09
feat(benchmark)!: rewrite the public API

sheremet-va:
Related https://github.com/vitest-dev/vitest/discussions/7850

New API as part of the test instead of _being_ a test:

```ts

test('compare JSON libraries', async ({ bench }) => {
  const input = '{"key":"value","nested":{"a":1}}'

  const result = await bench.compare(
    bench('JSON.parse', () => {
      JSON.parse(input)
    }),
    bench('custom parser', () => {
      customParse(input)
    }),
  )
})
```


sheremet-va:
The docs: https://deploy-preview-10113--vitest-dev.netlify.app/guide/benchmarking

This part is not yet implemented:

> In regular tests this overhead is negligible, but in benchmarks where a function is called millions of times, the getter call itself can dominate the measurement.


Regarding this:

> Executes the benchmark and stores the result to a __benchmarks__/ directory next to the test file

Maybe we should use the new convention instead 🤔 But it should not be in gitignore 🤔

docs/guide/benchmarking.md
```diff
@@ -0,0 +1,354 @@
...
+  await bench('parse', () => {
+    const result = JSON.parse(input)
+    doSomething(result)
+  }).run()
+})
+```
+
+This applies to all engines (V8, JavaScriptCore, SpiderMonkey) but is especially aggressive in V8's TurboFan and JavaScriptCore's FTL compiler tiers.
+
+### Module Runner Overhead
```
hi-ogawa, line 364: We can also suggest pre-built library as external when comparing own library against others. Also for pre-built one, users can disable module runner potentially. Related https://github.com/vitest-dev/vitest/issues/6543

test/config/test/mode.test.ts
hi-ogawa, line 1: The fixture can be also removed `test/config/fixtures/mode`

docs/guide/benchmarking.md
hi-ogawa, line 1: The content looks good, but I was puzzled by how I can actually get started. Probably first section or new section should mention the activation switch. Codex told me something like this:

---

| Command/config | Regular test project | Benchmark project |
|---|---:|---:|
| `vitest` / `vitest run`, no `benchmark.enabled` | runs | not created |
| `vitest` / `vitest run`, with `benchmark.enabled: true` | runs | runs too |
| `vitest bench` | filtered out | runs only |

A benchmark project is a cloned project using `benchmark.include` instead of `test.include`, defaulting to `*.bench.ts` / `*.benchmark.ts`.

So practically:

- `vitest run` runs normal tests.
- `vitest run` + `benchmark.enabled: true` runs normal tests and benchmark files.
- `vitest bench` runs benchmark files only, no config needed.

And `{ bench }` itself does not decide collection. It only works once the file is already in the benchmark project.

packages/vitest/src/node/projects/resolveProjects.ts
```diff
@@ -197,10 +197,72 @@ export async function resolveProjects(
     names.add(name)
   }
 
-  return resolveBrowserProjects(vitest, names, resolvedProjects)
+  return resolveDefaultProjects(vitest, names, resolvedProjects)
 }
 
-export async function resolveBrowserProjects(
+export async function resolveDefaultProjects(
```
hi-ogawa, line 203: nit. struggled to follow the logic around `benchmark.enabled` being mutated. Maybe this model can be simplified by
> `// TODO: mark if benchmark project?`

packages/vitest/src/node/reporters/reported-tasks.ts
```diff
@@ -205,6 +206,15 @@ export class TestCase extends ReportedTaskImplementation {
     return [...this.task.artifacts]
   }
 
+  /**
+   * @exorimental
```
hi-ogawa:
```suggestion
   * @experimental
```

docs/guide/test-context.md
```diff
@@ -117,6 +117,33 @@ it('stop request when test times out', async ({ signal }) => {
...
+
+  expect(result.get('JSON.parse')).toBeFasterThan(result.get('custom parser'))
+})
+```
+
+See the [Benchmarks guide](/guide/benchmarking) for full documentation on comparisons, baselines, and assertion matchers.
+
+::: warning
+When the first argument is a function, Vitest uses it only to derive the benchmark's name. It is not executed. The benchmark body must be passed as the second argument.
+:::
```
hi-ogawa: This looks out of place without example. Also, since `bench` requires multiple arguments anyways, this doesn't look like so pitfall.

packages/vitest/src/node/reporters/base.ts
```diff
@@ -860,34 +896,77 @@ export abstract class BaseReporter implements Reporter {
     return leakWithStacks.size
   }
 
-  reportBenchmarkSummary(files: File[]): void {
-    const benches = getTests(files)
-    const topBenches = benches.filter(i => i.result?.benchmark?.rank === 1)
+  protected printPerProjectBenchmarks(): void {
```
hi-ogawa, line 899: This shows with single project run too. Perhaps we can hide this in this case?

```sh
$ BROWSER=chromium pnpm -C test/browser/ test-fixtures bench --browser.headless

> @vitest/test-browser@ test-fixtures /home/hiroshi/code/others/vitest-wt2/test/browser
> vitest bench --browser.headless

Testing types with tsc and vue-tsc is an experimental feature.
Breaking changes might not follow SemVer, please pin Vitest's version when using it.

 DEV  v5.0.0-beta.1 /home/hiroshi/code/others/vitest-wt2/test/browser

 ✓  chromium (bench)  test/browser.bench.ts (3 tests) 6ms
   ✓ perProject registrations flow through the browser RPC (onTestBenchmark) 3ms
     name       hz     min     max    mean     p75     p99    p995    p999     rme  samples
     1 + 1  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  ±0.00%        2

     name       hz     min     max    mean     p75     p99    p995    p999     rme  samples
     1 + 2  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  ±0.00%        2
   ✓ bench.compare resolves a BenchStorage in the browser 1ms
     name      hz     min     max    mean     p75     p99    p995    p999     rme  samples
     a     0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  ±0.00%        2   fastest
     b     0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  ±0.00%        2
   ✓ bench.withBaseline exercises the readBenchmarkBaseline RPC round-trip 1ms
     name               hz     min     max    mean     p75     p99    p995    p999     rme  samples
     with-baseline  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  ±0.00%        2

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Cross-Project Benchmark Comparison ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯

  test/browser.bench.ts > perProject registrations flow through the browser RPC (onTestBenchmark) > 1 + 1
   project               hz     min     max    mean     p75     p99    p995    p999     rme  samples
   chromium (bench)  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  ±0.00%        2

  test/browser.bench.ts > perProject registrations flow through the browser RPC (onTestBenchmark) > 1 + 2
   project               hz     min     max    mean     p75     p99    p995    p999     rme  samples
   chromium (bench)  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  ±0.00%        2


 Test Files  1 passed (1)
      Tests  3 passed (3)
   Start at  16:11:53
   Duration  578ms (transform 0ms, setup 0ms, import 6ms, tests 6ms, environment 0ms)

 PASS  Waiting for file changes...
```

hi-ogawa (commented):
The new architecture makes sense. I've read through main logic and looks good to me. Will skim through test cases later.

skill-t.md
hi-ogawa, line 1: Nice skill 😄

docs/guide/benchmarking.md
```diff
@@ -0,0 +1,354 @@
...
+test('no performance regression', async ({ bench }) => {
+  const result = await bench.withBaseline('parse', () => {
+    parse(largeInput)
+  }).run()
+})
+```
+
+- **First run**: Executes the benchmark and stores the result to a `__benchmarks__/` directory next to the test file.
+- **Subsequent runs**: Executes the benchmark and compares against the stored result, reporting any regressions.
+- **Updating baselines**: Use the `--update-baselines` flag to overwrite stored baselines with fresh results.
```
hi-ogawa: I got confused / mislead by `Subsequent runs` claim. It reads like `bench.withBaseline + run` alone somehow surfaces the baseline comparison, but it's meant to combined with general `bench.compare` pattern to show comparison table in reporter.

I know the API is servicing as designed but the specific baseline workflow probably requires more comprehensive explainer like maybe with step by step example.

---

Also another perspective is that, how about providing ergonomic wrapper API to combine `compare + withBaseline (+ toBeFasterThan)`?

docs/guide/benchmarking.md
```diff
@@ -0,0 +1,354 @@
...
+  const result = await bench.compare(
+    bench('lib1', () => { lib1() }),
+    bench('lib2', () => { lib2() }),
+  )
+
+  expect(result.get('lib1')).toBeFasterThan(result.get('lib2'))
+})
+```
+
+## Baselines
```
hi-ogawa: I'm not sure new baseline API is convenient for common workflow. 

For example, previous approach with throw-away output json + compare allowed switching around branch to get the comparison like https://main.vitest.dev/config/benchmark.html#benchmark-outputJson

```sh
# save main branch's result
git checkout main
vitest bench --outputJson main.json

# change a branch and compare against main
git checkout feature
vitest bench --compare main.json
```

How does the new API (with `__benchmark__` encouraged to be committed) allow similar workflow?

sheremet-va: You can define both:

```ts
await bench.compare(
  bench('always run', () => { code() })
  bench.withBaseline('stored bench', () => { code() })
)
```

Instead of doing it in CLI level, you define it in test

docs/guide/benchmarking.md
```diff
@@ -0,0 +1,354 @@
...
+- **Updating baselines**: Use the `--update-baselines` flag to overwrite stored baselines with fresh results.
+
+Baselines work inside `bench.compare()` too. You can mix regular and baseline benchmarks:
+
+```ts
+test('compare against baseline', async ({ bench }) => {
+  const result = await bench.compare(
+    bench('current implementation', () => { current() }),
+    bench.withBaseline('previous implementation', () => { previous() }),
+  )
```
hi-ogawa, line 269: Is this intended to be both `current`?

```suggestion
  const result = await bench.compare(
    bench('current implementation', () => { current() }),
    bench.withBaseline('previous implementation', () => { current() }),
  )
```

docs/guide/test-context.md
sheremet-va: Sure, let's just remove it

skill-t.md
sheremet-va: yeah, this is supposed to be released separately and better formatted :D

docs/guide/benchmarking.md
sheremet-va: I'm not sure about adding more and more wrappers, I'd rather we design an API that's easy to combine together. I think `withBaseline` is flexible enough, but I am willing to see the exact proposal you are thinking of (how the code looks like?)

docs/config/benchmark.md
```diff
@@ -32,39 +39,17 @@ Include globs for in-source benchmark test files. This option is similar to [`in
...
-git checkout main
-vitest bench --outputJson main.json
+- **Type:** `boolean`
+- **Default:** `false`
 
-# change a branch and compare against main
-git checkout feature
-vitest bench --compare main.json
-```
+Include the `samples` array of per-iteration timings on every benchmark result. Off by default to reduce memory usage; enable when a custom reporter or API consumer needs the raw samples.
```
AriPerkkio:
```suggestion
Include the `samples` array of per-iteration timings on every benchmark result. Disabled by default to reduce memory usage; enable when a custom reporter or API consumer needs the raw samples.
```

docs/guide/benchmarking.md
```diff
@@ -0,0 +1,404 @@
...
+
+Use the `bench` fixture to define a benchmark. Call `.run()` to execute it:
+
+```ts
+import { expect, test } from 'vitest'
+
+test('parsing performance', async ({ bench }) => {
+  const result = await bench('parse', () => {
+    JSON.parse('{"key":"value"}')
+  }).run()
```
AriPerkkio, line 19: Below it states following:

> The `bench()` function registers a benchmark without executing it. Calling `.run()` runs the benchmark and returns the result.

To make that even more clear in this example, I would suggest following:


```suggestion
  const jsonParseBenchmark = bench('parse', () => {
    JSON.parse('{"key":"value"}')
  })
  const result = await jsonParseBenchmark.run()
```

docs/guide/benchmarking.md
```diff
@@ -0,0 +1,404 @@
...
+    },
+  },
+})
+```
+
+With this config, `vitest` runs your regular tests first, then runs the benchmarks in a separate isolated group (so benchmark execution never overlaps with test execution and adds noise to results). Useful in CI when you want a single command to validate correctness and performance.
+
+### `vitest bench`
+
+The `bench` subcommand runs benchmarks only and skips regular tests:
```
AriPerkkio: Nit:


```suggestion
The `bench` subcommand runs only benchmarks and skips regular tests:
```

docs/guide/benchmarking.md
```diff
@@ -0,0 +1,404 @@
...
+test('compare implementations across browsers', async ({ bench }) => {
+  await bench.compare(
+    bench.perProject('JSON.parse', () => {
+      JSON.parse('{"key":"value"}')
+    }),
+    bench('custom parser', () => {
+      customParse('{"key":"value"}')
+    }),
+  )
+})
```
AriPerkkio, line 164: Did you consider having `perProject` as configuration property instead? I find it a bit weird as property in `bench` 🤔 


```suggestion
test('compare implementations across browsers', async ({ bench }) => {
  await bench.compare(
    bench('JSON.parse', { perProject: true }, () => {
      JSON.parse('{"key":"value"}')
    }),
    bench('custom parser', () => {
      customParse('{"key":"value"}')
    }),
  )
})
```

test/e2e/fixtures/reporters/function-as-name.bench.ts
```diff
@@ -10,6 +10,12 @@ const options = {
 function foo() {}
 class Bar {}
 
-bench(foo, () => {}, options)
-bench(Bar, () => {}, options)
-bench(() => {}, () => {}, options)
+test('benches', async ({ bench, annotate }) => {
```
AriPerkkio:
```suggestion
test('benches', async ({ bench }) => {
```

docs/guide/benchmarking.md
```diff
@@ -0,0 +1,404 @@
...
+  ```
+
+#### JavaScriptCore (Bun, Safari)
+
+- **Different optimization thresholds**: JSC uses its own JIT tiers (LLInt → Baseline → DFG → FTL) with different inlining and optimization heuristics. A benchmark that is fast on V8 may behave very differently on JSC.
+- **Async benchmarks**: Bun's event loop implementation differs from Node.js. If your benchmark involves async operations or timers, results may not be directly comparable across runtimes.
+
+#### Browser
+
+- **Timer resolution**: Browsers may reduce `performance.now()` precision (e.g., to 100μs or even 1ms) as a Spectre mitigation. This makes very fast operations difficult to measure accurately, so increase iterations to compensate:
```
AriPerkkio: I'd rather not mention specific vulnerabilities, and keep this generic:


```suggestion
- **Timer resolution**: Browsers may reduce `performance.now()` precision (e.g., to 100μs or even 1ms) as a security mechanism. This makes very fast operations difficult to measure accurately, so increase iterations to compensate:
```

docs/guide/benchmarking.md
```diff
@@ -0,0 +1,404 @@
...
+
+```bash
+# only benchmarks in files matching "parser"
+vitest bench parser
+
+# only benchmarks whose test name matches "JSON"
+vitest bench -t JSON
+```
+
+## Comparing Benchmarks
```
AriPerkkio, line 75: Haven't yet read through the whole PR but: Is it possible to save benchmark results with specific name, and then compare against those without ever updating them?

```ts
import { test } from 'vitest'
import { customParser } from '../dist';

test('compare against v1', async ({ bench }) => {
  const v1Results = await bench.load("v1-results.json") // or something?

  const result = await bench.compare(
    bench('current v2', () => {
      customParser(input)
    }),
    v1Results,
  )
})
```

packages/browser/src/node/rpc.ts
```diff
@@ -181,6 +181,14 @@ export function setupBrowserRpc(globalServer: ParentBrowserProject, defaultMocke
 
           return vitest._testRun.recordArtifact(id, artifact)
         },
+        async onTestBenchmark(testId, benchmark) {
+          return vitest._testRun.recordBenchmark(testId, benchmark)
+        },
+        async readBenchmarkBaseline(testFilepath, key) {
+          // TODO: test
```
AriPerkkio: 👀

packages/vitest/src/node/cli/cac.ts
```diff
@@ -263,24 +262,24 @@ export function parseCLI(argv: string | string[], config: CliParseOptions = {}):
...
-  await start('test', cliFilters, options)
+  await start(cliFilters, options)
 }
 
 async function benchmark(cliFilters: string[], options: CliOptions): Promise<void> {
-  console.warn(c.yellow('Benchmarking is an experimental feature.\nBreaking changes might not follow SemVer, please pin Vitest\'s version when using it.'))
-  await start('benchmark', cliFilters, options)
+  options.benchmarkOnly = true
+  await start(cliFilters, options)
 }
```
AriPerkkio, line 282: This should probably disable `options.coverage.enabled` too.

And when running `vitest run` while having benchmarking enabled via configuration file, we should maybe programatically disable coverage and clear module cache to avoid running instrumented files. Didn't yet read whole PR so not sure if that's already done.

AriPerkkio (commented):
Review in progress (`31 / 128 viewed`), below first findings.

docs/guide/benchmarking.md
hi-ogawa: I think the primitive is right, so not a blocker. The documentation wise though, it should imply this primitive can be used together with `compare` API or manual run result comparison.

---

For API wise (or higher level integration suggestion from doc), I spend sometime but struggling how this is meant to be used, so it was just an open question from me.

Exploring some concrete use case to target a tangible goal and here is one story. I was thinking about how to setup CI workflow to compare bench result from main against each PR. One integration example I was thinking would be  something like:

- 1. have ci job like this

```yaml
jobs:
  # either generate fresh main checkout as baseline in each pr job
  # or assume main branch job generated such artifact
  benchmark-baseline:
    steps:
      - uses: actions/checkout@v4
        with:
          ref: ${{ github.base_ref || 'main' }}
      - run: pnpm vitest bench --update-baselines
        with:
          name: benchmark-baselines
          path: '**/__benchmarks__/*.json'  # side note. this could be simpler if we employ .vitest/benchmark/*

  # run pr checkout bench against main baseline artifact
  benchmark-pr:
    needs: benchmark-baseline
    steps:
      ...
        with:
          name: benchmark-baselines
      - run: pnpm vitest bench
```

- 2. the latter bench run wants to surface obvious large regression (assuming github actions environment aren't that stable for precise benchmarking, but maybe good enough to catch bad regression)

Some characteristics for this specific use case are:
- benchmark artifact is a throw-away artifact, so no need to be commited. It could be even simplified if it's consolidated to root `.vitest/benchmark/*` instead of `__benchmark__`
- this use case only requires the simplest combination of `bench.withBaseline + bench.compare` such as:

```js
const result = await bench.compare(
  bench('current', () => current()),
  bench.withBaseline('baseline', () => current()),
)
```

- The pattern of how to present compare `result` can be also somewhat standardized. Being able to define threshold inside code itself (i.e. per bench case) instead of global config is likely still important. But, it feels like it should end up as reporter-accessible metadata so that reporter can eventually surface it specifically for the common pattern.

So given that assumption, I could see API like this seems more ergonomic to fit the story:

```js
// conceptually a sugar of compare + withBaseline + toBeFasterThan
// but additionally extra metadata (e.g. bench name, threshold) are available through
// the assertion error property or maybe as annotation so builtin (or 3rd party) reporter can associate
// with raw benchmark result.
await bench.compareBaseline("my-code", { threshold: etc... }, () => someFn())
```

packages/vitest/src/node/reporters/base.ts
```diff
@@ -860,34 +896,77 @@ export abstract class BaseReporter implements Reporter {
...
+    this.log('')
+    this.log(divider(c.bold(c.bgBlue(` Cross-Project Benchmark Comparison `)), null, null, c.blue))
 
-    for (const bench of topBenches) {
-      const group = bench.suite || bench.file
+    for (const [benchName, projectMap] of this._perProjectBenchmarks) {
+      const tasks = [...projectMap.entries()]
+        .map(([projectName, task]) => ({ ...task, name: projectName }))
+        .sort((a, b) => a.latency.mean - b.latency.mean)
+        .map((task, idx) => ({ ...task, rank: idx + 1 }))
```
AriPerkkio:
```suggestion
      const tasks = [...projectMap.entries()]
        .sort((a, b) => a[1].latency.mean - b[1].latency.mean)
        .map(([projectName, task], index) => ({ ...task, name: projectName, rank: index + 1 }))
```

packages/vitest/src/node/reporters/base.ts
```diff
@@ -1001,6 +1080,71 @@ function sum<T>(items: T[], cb: (_next: T) => number | undefined) {
...
 }
 
+function formatBenchNumber(number: number): string {
+  const res = String(number.toFixed(number < 100 ? 4 : 2)).split('.')
+  return res[0].replace(/(?=(?:\d{3})+$)\B/g, ',') + (res[1] ? `.${res[1]}` : '')
+}
+
+// Plain-text rendering of the benchmark table (no ANSI colors, no indent).
+// Used by the junit reporter to embed benchmark data in <system-out>.
+export function renderBenchmarkTableText(
```
AriPerkkio: Would this fit in `vitest/src/node/reporters/renderers/` instead, maybe as `benchmark-table.ts`?

packages/vitest/src/node/reporters/summary.ts
AriPerkkio, line 1: These changes look like unrelated refactoring. Are they related to the feature somehow?

AriPerkkio (commented):
Second part, `70 / 128 viewed`.

AriPerkkio (commented):
Finished review ✅ 

Overall looks great. This seems much better compared to the previous architecture where benchmark was kind of "addition" rather than built-in. Also removing the `mode` argument is great.

docs/guide/benchmarking.md
sheremet-va: Yes, I went with an option because bench options are just tinybench options (simpler mental image, in my opinion). I am open to reconsider

docs/guide/benchmarking.md
sheremet-va: What would `load` do if there is nothing to load? `bench.withBaseline` has a fallback to run and generate the result

packages/vitest/src/node/reporters/summary.ts
sheremet-va: Yes, the idea was to show the benchmark running (like we show hooks), but I decided to have only a single hook (benchmark end), but I was too lazy to revert the whole file

sheremet-va: If we ever implement `test.step`, we would probably show it here too

docs/guide/benchmarking.md
AriPerkkio: It would be same as `withBaseline`, but it also would make sure that on file system there is a baseline that is never updated, even when `--update-baselines` is used. If result matching the name (`v1-results.json` in example) is not found, it would just error out.

sheremet-va: How would you generate it then?

AriPerkkio: Needs a way to write benchmarks to file system. 🤔

sheremet-va: The API in this PR is clear - you use results for comparison with `expect`, and you use `bench` to show the comparison table (`.compare`, `.run`), I don't understand what `.load` adds that `withBaseline` doesn't cover. What does it return? How do you generate that file if it only reads it? If another function generates it, why should this one return it? _When_ is it generated? How do you use it? Do you show the table? 

`.compare` defines the order and other options of benchmarks (`tinybench.add().add().add().run(options)`) - we don't do `await bench, await bench, await bench` for that exact reason

`.load` just doesn't fit in anywhere

AriPerkkio: Is there a way for me to use this API and see how my project's performance has changed over time, in multiple versions? So if I was using this feature in my project over three major versions, could I somehow see how my project's performance was in v1, v2 and currently for example?

Or is it just possible to see how it's performing right now, and in previous baseline?

sheremet-va: Sure it’s possible, withBaseline reads the file from the disk. If it was generated in v1, it will show that. How do you think withBaseline works? Are docs confusing?

AriPerkkio: Can you write an example of test case that would compare `import { customParser } from '../dist';` function against its performance in `v1`, `v2` and current? How does `--update-baselines` ignore updating those?

sheremet-va: Why would you run update-baselines if you don’t want to update it? Just don’t update baselines

You run it once, it’s generated and it stays there. It would look like 3 bench.withBaseline calls inside bench.compare that were added there gradually 

If you want to be strict, just throw an error in the baseline function

sheremet-va:
```ts
await bench.compare(
  bench.withBaseline('v1', impl)
)
// run, generates v1
await bench.compare(
  bench.withBaseline('v1',() => { throw new Error() })
  bench.withBaseline('v2', impl)
)
// run, generates v2
```

docs/guide/benchmarking.md
sheremet-va: Proposal:

- step by step example of generating baselines
- workflow that commits updated baselines (when/before pr is merged?)

```ts
await bench('fn', () => current()).run({ writeResult: './stored-artifact.json' })

await bench.compare(
  // bench('fn', { writeResult: './stored-artifact.json' }, () => current()),
  bench.from('./stored-artifact.json'),
  bench.from(() => fetch('./stored-artifact.json').json()),
)
```

docs/guide/benchmarking.md
sheremet-va: I prefer the current example because that is how it's going to be used and spiting it into two variables is not really useful

docs/guide/benchmarking.md
hi-ogawa: The minimal example I have in mind for baseline use case demo:

```js
test("demo", ({ bench }) => {
  if (import.meta.env.WRITE_BENCH) {
    const baseline = bench("baseline", { writeResults: "./my-bench.json" }, () => fn())
    await baseline.run()
  }  else {
    const baseline = bench.from("baseline", "./my-bench.json")
    await bench.compare(bench(name, () => fn()), baseline)
  }
})
```

Or extracting it as simple wrapper utility:

```js
test("add", ({ bench }) => {
  await benchCompareBaseline(bench, "add", () => 2 + 2, "./add.json")
})

test("mul", ({ bench }) => {
  await benchCompareBaseline(bench, "mul", () => 2 * 2, "./mul.json")
})

function benchCompareBaseline(bench, name, fn, filepath) {
  if (import.meta.env.WRITE_BENCH) {
    const baseline = bench("baseline", { writeResults: filepath }, () => fn())
    await baseline.run()
  }  else {
    const baseline = bench.from("baseline", { filepath })
    await bench.compare(bench(name, () => fn()), baseline)
  }
}
```

docs/guide/benchmarking.md
```diff
@@ -0,0 +1,480 @@
...
+    await baseline.run()
+  }
+  else {
+    const baseline = bench.from('baseline', './my-bench.json')
+    await bench.compare(bench('current', () => fn()), baseline)
+  }
+})
+```
+
+Run `WRITE_BENCH=1 vitest bench` to refresh the stored result, and `vitest bench` to compare the current implementation against it.
```
hi-ogawa: `VITE_WRITE_BENCH` and `WRITE_BENCH` are inconsistent. Probably Browser mode only picks up `VITE_...` prefix, so it should be `VITEST_WRITE_BENCH=1 vitest bench`.

packages/vitest/src/node/benchmark.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export class BenchmarkManager {
+  constructor(private vitest: Vitest) {}
+
+  // Resolve a user-supplied path against the project root. Reject paths that
+  // escape the project root: `bench.from()` accepts arbitrary input, and we
+  // never want a benchmark file to be able to read or clobber files outside
+  // the workspace.
+  public resolve(relativePath: string): string {
+    const root = this.vitest.config.root
```
hi-ogawa: Do we call this "project root"? Not sure if we have used better word, but technically "root project root".

test/e2e/vitest.config.ts
```diff
@@ -45,8 +45,9 @@ export default defineConfig({
           typecheck: {
             enabled: true,
             include: [
-              './test/config-types.test-d.ts',
               './test/reporters/configuration-options.test-d.ts',
+              './test/benchmarking.test-d.ts',
```
hi-ogawa, line 49: This file doesn't exist. Was looking for type tests for compare `result.get` and maybe that's supposed to be here.

docs/guide/benchmarking.md
```diff
@@ -0,0 +1,480 @@
...
+})
+```
+
+This applies to all engines (V8, JavaScriptCore, SpiderMonkey) but is especially aggressive in V8's TurboFan and JavaScriptCore's FTL compiler tiers.
+
+### Module Runner Overhead
+
+By default, Vitest runs tests in Node.js using Vite's module runner (configured by [`experimental.viteModuleRunner`](/config/experimental#experimental-vitemodulerunner)). This transforms all module exports into getters, so every access to an imported binding goes through something like `__vite_ssr_module__.value`. In regular tests this overhead is negligible, but in benchmarks where a function is called millions of times, the getter call itself can dominate the measurement.
+
+Vitest will print a warning if it detects excessive getter calls, but you should be aware of this when benchmarking imported functions:
```
hi-ogawa: There's no mention of `suppressExportGetterWarnings` in the doc. Probably worth mentioning here and `docs/config/benchmark.md`.

packages/vitest/src/node/config/serializeConfig.ts
```diff
@@ -134,8 +133,11 @@ export function serializeConfig(project: TestProject): SerializedConfig {
     standalone: config.standalone,
     printConsoleTrace:
       config.printConsoleTrace ?? globalConfig.printConsoleTrace,
-    benchmark: config.benchmark && {
-      includeSamples: config.benchmark.includeSamples,
+    benchmark: {
+      enabled: config.benchmark?.enabled ?? false,
+      retainSamples: config.benchmark?.retainSamples ?? false,
+      suppressExportGetterWarnings: config.benchmark?.suppressExportGetterWarnings ?? false,
+      projectName: config.benchmark?.projectName ?? '',
```
hi-ogawa: The optional chain and fallback are not required according to the type `ResolvedConfig.benchmark: Required<BenchmarkUserOptions>`

hi-ogawa (commented):
did surface level check. the api looks good.
will read new part of implementation nexy later.

test/e2e/vitest.config.ts
sheremet-va: Hm, maybe it was removed during the merge by accident 😢

packages/vitest/src/node/benchmark.ts
sheremet-va: Well, I think it should actually be relative to the project, this line of code is wrong

test/e2e/test/benchmarking.test-d.ts
hi-ogawa, line 1: `bench.from("a", ...)` type isn't tested.

docs/guide/benchmarking.md
```diff
@@ -0,0 +1,480 @@
+---
+title: Benchmarking | Guide
+---
+
+# Benchmarking
+
+Vitest lets you write benchmarks alongside your tests using the `bench` fixture from the [test context](/guide/test-context). It is not a top-level import from `vitest`. Benchmarks are powered by [Tinybench](https://github.com/tinylibs/tinybench) and are defined inside regular `test()` calls, giving you access to the full power of Vitest's test runner: retries, lifecycle hooks, filtering, and assertions.
```
hi-ogawa: nit, repeated mention of "not a top-level import from `vitest`" reads a bit off. If this is for migration intent, then I'd either leave it migration.md or the migration intent should be phrased cleary in the small warning block.

docs/guide/benchmarking.md
```diff
@@ -0,0 +1,480 @@
...
+import { expect, test } from 'vitest'
+
+test('parsing performance', async ({ bench }) => {
+  const result = await bench('parse', () => {
+    JSON.parse('{"key":"value"}')
+  }).run()
+})
+```
+
+The `bench()` function registers a benchmark without executing it. Calling `.run()` runs the benchmark and returns the result. Vitest will print the benchmark output (ops/sec, mean time, etc.) after the test completes.
```
hi-ogawa: I had an impression that bare `bench + run` won't print anything as there wasn't reporter output example other than `compare` and `perProject`. Can we add the bare example output here?

hi-ogawa (commented):
lgtm! left minor reviews.

docs/guide/benchmarking.md
sheremet-va: I feel like there are already enough examples, it will be printed the same way as `compare`, just with a single element with it - I can add this as a note to compare result

hi-ogawa (approved):

PR #10117, opened 2026-04-09
fix(coverage): descriptive error message when reports directory is removed during test run

DaveT1991:
Fixes #10111

`onAfterSuiteRun` writes coverage JSON files to `.tmp` without first ensuring the directory exists. Under certain parallel/race conditions the directory may not yet exist, causing `writeFile` to throw even when all tests pass.

## Fix
Chain `fs.mkdir(this.coverageFilesDirectory, { recursive: true })` before `fs.writeFile`. Using `recursive: true` makes this idempotent — it is a no-op when the directory already exists, so there is no performance or correctness impact on normal runs.

### DIFF
--- a/packages/vitest/src/node/coverage.ts
+++ b/packages/vitest/src/node/coverage.ts
@@ -277,7 +277,8 @@ export class BaseCoverageProvider {
     // If there's a result from previous run, overwrite it
     entry[environment][testFilenames] = filename
 
-    const promise = fs.writeFile(filename, JSON.stringify(coverage), 'utf-8')
+    const promise = fs.mkdir(this.coverageFilesDirectory, { recursive: true })
+      .then(() => fs.writeFile(filename, JSON.stringify(coverage), 'utf-8'))
     this.pendingPromises.push(promise)
   }
 
--- a/packages/vitest/src/node/coverage.ts
+++ b/packages/vitest/src/node/coverage.ts
@@ -255,7 +255,7 @@ export class BaseCoverageProvider {
     this.pendingPromises = []
   }
 
-  private normalizeCoverageFileError(error: unknown): Error {
+  private normalizeCoverageFileError(error: unknown): unknown {
     if (
       error instanceof Error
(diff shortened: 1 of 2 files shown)


DaveT1991:
Fixes #10111

When writing or reading temp coverage files, if the coverage temp directory has been removed, Vitest now throws a clearer error instead of silently recreating the directory.

This keeps the existing fail-fast behavior for conflicting concurrent runs, but makes the failure more actionable by pointing users at shared `coverage.reportsDirectory` usage.

packages/vitest/src/node/coverage.ts
```diff
@@ -255,6 +255,22 @@ export class BaseCoverageProvider {
...
+      && error.code === 'ENOENT'
+      && !existsSync(this.coverageFilesDirectory)
+    ) {
+      return new Error(
+        `Something removed the coverage directory "${this.coverageFilesDirectory}" Vitest created earlier. Make sure you are not running multiple Vitests with the same "coverage.reportsDirectory" at the same time.`,
+        { cause: error },
+      )
+    }
+
+    return error instanceof Error ? error : new Error(String(error))
```
AriPerkkio: We can throw the original `error` here as is too.


```suggestion
    return error
```

AriPerkkio (changes requested):

packages/vitest/src/node/coverage.ts
```diff
@@ -255,6 +255,22 @@ export class BaseCoverageProvider {
     this.pendingPromises = []
   }
 
+  private normalizeCoverageFileError(error: unknown): Error {
```
AriPerkkio:
```suggestion
  private normalizeCoverageFileError(error: unknown): unknown {
```

AriPerkkio (approved):
Looks good, thanks! 💯

DaveT1991:
You're welcome AriPerkkio

CarAuto34:
Ty

PR #10122, opened 2026-04-10
feat(coverage): allow perFile to accept an object for separate per-fi…

### LINKED ISSUE #10114: Support combining global coverage thresholds with minimum per-file thresholds

Currently, Vitest forces users to choose between global coverage thresholds and per-file thresholds (`perFile: true`). It is not possible to enforce both simultaneously in a single test run. 

In many real-world projects, teams want to maintain a high overall project coverage (e.g., 80% global average) while simultaneously ensuring that no individual file drops below a certain absolute minimum baseline (e.g., 50% per-file). 

Right now, if we use global thresholds, a single file can have 0% coverage as long as the overall average is high enough. If we use `perFile: true` with 50%, we completely lose the 80% global project requirement. Relying solely on glob patterns grouping doesn't solve this, as it evaluates the grouped files as a whole, not iteratively per file alongside a global check.

### Suggested solution

Introduce a way to define global thresholds and per-file minimums at the same time within the `coverage.thresholds` options. 

A clean approach could be allowing the `perFile` property to accept an object of thresholds instead of just a boolean, or introducing a dedicated `perFileThresholds` key.

Example of how the API could look:

```typescript
export default defineConfig({
  test: {
    coverage: {
      thresholds: {
        // Global project thresholds (Must be met on average)
        lines: 80,
        functions: 80,
        branches: 80,
        statements: 80,
        
        // Absolute minimums (No individual file is allowed to drop below this)
        perFile: { 
          lines: 50,
          functions: 50,
          branches: 50,
          statements: 50
        }
      }
    }
  }
})
```

### Alternative

_No response_

### Additional context

It is worth noting that this dual-threshold behavior was fully supported in **Karma**. With the Angular ecosystem currently shifting away from Karma and moving towards Vitest as the recommended testing framework, many teams are migrating their test setups. During this migration, developers are losing this specific quality gate. Supporting this feature in Vitest would bring important feature parity and significantly ease the transition for the Angular community.

### Validations

DaveT1991:
Closes #10114

perFile: true currently re-uses the same global threshold values for per-file checks, forcing users to choose between a project-wide average and per-file minimums — they can't enforce both at once.

This PR allows perFile to accept a threshold object. When it does, both run simultaneously: the top-level thresholds apply to the global aggregate, and the perFile object thresholds apply to each individual file.

// Require 80% lines globally, but each file must have at least 50%
coverage: {
  thresholds: {
    lines: 80,
    perFile: { lines: 50 },
  },
}
The existing perFile: true behaviour is unchanged.

Changes
packages/vitest/src/node/types/coverage.ts — perFile type widened to boolean | { lines?, branches?, functions?, statements? } with JSDoc example
packages/vitest/src/node/coverage.ts — checkThresholds() builds summaries for both the global aggregate and per-file entries when perFile is an object, each checked against their respective thresholds
test/coverage-test/test/threshold-per-file-object.test.ts — four tests covering: per-file fails while global passes, both fail, both pass, and only specified keys are checked per file

### DIFF
--- a/docs/config/coverage.md
+++ b/docs/config/coverage.md
@@ -233,12 +233,48 @@ Global threshold for statements.
 
 ### coverage.thresholds.perFile
 
-- **Type:** `boolean`
+- **Type:** `boolean | { lines?: number; functions?: number; branches?: number; statements?: number }`
 - **Default:** `false`
 - **Available for providers:** `'v8' | 'istanbul'`
 - **CLI:** `--coverage.thresholds.perFile`, `--coverage.thresholds.perFile=false`
 
 Check thresholds per file.
+
+When set to `true`, the global threshold values are reused for every file.
+
+When set to an object, the top-level threshold values still apply to the global coverage summary, and the nested `perFile` values apply to each individual file.
+
+### coverage.thresholds.perFile.lines
+
+- **Type:** `number`
+- **Available for providers:** `'v8' | 'istanbul'`
+- **CLI:** `--coverage.thresholds.perFile.lines=<number>`
+
+Per-file threshold for lines.
+
+### coverage.thresholds.perFile.functions
+
+- **Type:** `number`
+- **Available for providers:** `'v8' | 'istanbul'`
+- **CLI:** `--coverage.thresholds.perFile.functions=<number>`
+
+Per-file threshold for functions.
+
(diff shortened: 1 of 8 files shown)


DaveT1991:
I think it's ready now after a few fixes. Please check and let me know.

PR #10127, opened 2026-04-12
fix: preserve this context in defineHelper

lawsonjl:
### Description

The `vi.defineHelper` function drops the `this` context, even though the type signature of `defineHelper` (`<F>(fn: F) => F`) suggests that the call signature is preserved exactly. This change simply calls the wrapped function with the same `this` that the wrapper receives.

I intended to add a test, but I wasn't able to understand the existing tests well enough to be sure I was doing it right. At the very least, no existing tests suddenly started failing.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/integrations/vi.ts
+++ b/packages/vitest/src/integrations/vi.ts
@@ -612,8 +612,8 @@ function createVitest(): VitestUtils {
     waitFor,
     waitUntil,
     defineHelper: (fn) => {
-      return function __VITEST_HELPER__(...args: any[]): any {
-        const result = fn(...args)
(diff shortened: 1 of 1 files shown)


hi-ogawa:
Good catch. Not sure about the use case, but makes sense to fix.
For this one,  we can add one more `const myHelperWithThis = vi.defineHelper(function (this) { ... })` in `test/cli/fixtures/assertion-helper/basic.test.ts` and do some assertion one passing and one failing. You update snapshot by `pnpm -C test/cli test assertion-helper.test.ts -u`.

hi-ogawa:
closing in favor of https://github.com/vitest-dev/vitest/pull/10163

PR #10129, opened 2026-04-12
fix(ast-collect): recognize __vi_import_ prefix in static test discovery

### LINKED ISSUE #10128: Does not recognize __vi_import_ prefix in static test discovery
## Description

Test discovery via the static AST parser silently drops `it`/`test` calls when a file contains both:
1. A `vi.mock()` statement (even **commented out**)
2. An extended test function imported from a separate module (e.g. `import { it } from "./utils/test-extend.js"`)

### Reproduction

Given this test file:
```ts
import { describe } from 'vitest'
import { it } from "./Utils/test-extend.js"

// vi.mock('@/composables/test.js', async (importOriginal) => { });

describe('my suite', () => {
  it('should work', () => {})
})
```

Where `test-extend.js` exports an extended test:
```js
import { test as baseTest } from 'vitest'
export const it = baseTest.extend({ /* fixtures */ })
```

**Expected:** Both `describe` and `it` are discovered by the static parser (gutter icons appear).
**Actual:** Only `describe` is discovered; `it` calls are silently skipped.

Removing either the `vi.mock` line **or** the extended import makes it work — the combination of both is required to trigger the bug.

### Reproduction

# StackBlitz

https://stackblitz.com/edit/vitest-dev-vitest-kdxvkvnr

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    @vitest/ui: latest => 4.1.4 
    vite: latest => 8.0.8 
    vitest: latest => 4.1.4
```

### Used Package Manager

npm

### Validations

Yejneshwar:
## Root Cause

The issue is in `astParseFile`'s `getName()` function, which identifies test function calls from SSR-transformed code.

### How `vi.mock` changes the SSR transform

When Vitest's module mocker Vite plugin detects any reference to `vi.mock` in the source — **including inside comments** — it changes how non-vitest imports are transformed:

**Without `vi.mock`** — standard static SSR import:
```js
const __vite_ssr_import_1__ = await __vite_ssr_import__("/src/Utils/test-extend.js", {"importedNames":["it"]});
//                                                      ↓ uses (0, ...) wrapper
(0, __vite_ssr_import_1__.it)("should work", () => {});
```

**With `vi.mock`** — dynamic import via module mocker:
```js
const __vi_import_0__ = await __vite_ssr_dynamic_import__("/src/Utils/test-extend.js");
//                                                        ↓ direct member access, no wrapper
__vi_import_0__.it("should work", () => {});
```

The key difference: the import variable changes from `__vite_ssr_import_N__` to `__vi_import_N__`.

### Why `getName()` fails

The `getName()` function in `astParseFile` handles `MemberExpression` nodes by checking for SSR import prefixes:

```js
if (callee.object?.name?.startsWith("__vite_ssr_") || /* Vitest namespace check */) {
    return getName(callee.property);  // → returns "it" ✅
}
// fallthrough:
return getName(callee.object);  // → returns "__vi_import_0__" ❌
```

`__vi_import_0__` does **not** start with `"__vite_ssr_"`, so the parser falls through to the generic `MemberExpression` handler, which returns the object name `"__vi_import_0__"` instead of the property name `"it"`. Since `"__vi_import_0__"` is not a recognized Vitest function name (`isVitestFunctionName()` returns false), the entire `it()` call is silently skipped.

## Fix

Add `__vi_import_` as a recognized SSR import prefix alongside `__vite_ssr_`:

```diff
  if (
    callee.object?.name?.startsWith("__vite_ssr_") ||
+   callee.object?.name?.startsWith("__vi_import_") ||
    (callee.object?.object?.name?.startsWith("__vite_ssr_") &&
      callee.object?.property?.name === "Vitest")
  ) {
    return getName(callee.property);
  }
```

This ensures that when the module mocker transforms an import to use `__vi_import_N__`, the parser still correctly extracts the property name (e.g. `it`, `test`, `describe`) and recognizes it as a test function.

> [!NOTE]
> The `message` cleanup regex on the line below already handles `__vi_import_N__` in test name strings:
> ```js
> .replace(/__vi_import_\d+__\./g, "")
> ```
> So the prefix was already *partially* accounted for in the message extraction — it was just missing from the `getName()` function that determines whether a call is a test function or not.

## Verification

Tested all reported scenarios after the fix:

| Scenario | Before | After |
|---|---|---|
| `// vi.mock(...)` + extended `it` | ❌ `it` not found | ✅ Found |
| `vi.mock(...)` + extended `it` | ❌ `it` not found | ✅ Found |
| No `vi.mock` + extended `it` | ✅ | ✅ |
| No `vi.mock` + standard `it` | ✅ | ✅ |
| `// vi.mock(...)` + standard `it` (from vitest) | ✅ | ✅ |

Resolves #10128


Yejneshwar:
The PR text was generated with Claude. The file changes were by me

sheremet-va (changes requested):
This needs tests. See static-collect.test.ts

Yejneshwar:
> This needs tests. See static-collect.test.ts

I have added two test cases to ```static-collect.test.ts```

1.  collects tests imported from another file while a vi.mock line is present
2.  collects tests imported from another file while a vi.mock line is present and commented

Test case ```2``` passes with Vite version ```7.1.5``` ( this is the version the repo uses in its current state ), but fails with  Vite ```8.0.8``` without the fix in this PR. 

# Reason

The output of ```ssr.transformRequest``` is different between Vite ```7.1.5``` and ```8.0.8```

## Vite ```7.1.5```

```
const __vite_ssr_import_0__ = await __vite_ssr_import__("/node_modules/.pnpm/vitest@4.1.4_vite@7.1.5_tsx@4.21.0_/node_modules/vitest/dist/index.js", {"importedNames":["describe"]});
const __vite_ssr_import_1__ = await __vite_ssr_import__("./Utils/test-extend.ts", {"importedNames":["it"]});


(0,__vite_ssr_import_0__.describe)("should included", () => {
  (0,__vite_ssr_import_1__.it)("is included because of workspace plugin setting", () => {
  });
});
(0,__vite_ssr_import_0__.describe)("should included", () => {
  (0,__vite_ssr_import_0__.describe)("nested", () => {
    (0,__vite_ssr_import_1__.it)("is included because of workspace plugin setting", ({ server }) => {
    });
  });
});
```

## Vite ```8.0.8```

```
const __vite_ssr_import_0__ = await __vite_ssr_import__("/node_modules/.pnpm/vitest@4.1.4_vite@8.0.8_esbuild@0.27.7_tsx@4.21.0_/node_modules/vitest/dist/index.js", {"importedNames":["describe"]});
const __vi_import_0__ = await __vite_ssr_dynamic_import__("./Utils/test-extend.ts");


// vi.mock('@/composables/test.js', async (importOriginal) => { });
(0,__vite_ssr_import_0__.describe)("should included", () => {
        __vi_import_0__.it("is included because of workspace plugin setting", () => {});
});
(0,__vite_ssr_import_0__.describe)("should included", () => {
        (0,__vite_ssr_import_0__.describe)("nested", () => {
                __vi_import_0__.it("is included because of workspace plugin setting", ({ server }) => {});
        });
});
```

### The difference
```diff
// Vite 7.1.5 
+ const __vite_ssr_import_1__ = await __vite_ssr_import__("./Utils/test-extend.ts", {"importedNames":["it"]});

// Vite 8.0.8
-  const __vi_import_0__ = await __vite_ssr_dynamic_import__("./Utils/test-extend.ts");
```

Is this expected behavior between the two versions??

I have added a ```static-collect-vite-matrix.test.ts``` file that tests parsing behaviour between the two versions; Vite 8.0.8 will fail the ```run parse``` test

test/cli/test/static-collect-vite-matrix.test.ts
```diff
@@ -0,0 +1,261 @@
+/**
```
sheremet-va: This file is not required, we only test for the vite version installed in the repo

test/cli/test/static-collect.test.ts
```diff
@@ -427,6 +427,92 @@ test('collects tests imported from another file', async () => {
...
+    import { describe } from 'vitest';
+    import { it } from './Utils/test-extend.ts';
+      
+    // vi.mock('@/composables/test.js', async (importOriginal) => { });
+      
+    describe('should included', () => {
+      it('is included because of workspace plugin setting', () => {});
+    });
+
+    describe('should included', () => {
```
sheremet-va: Give them different names, they are merged into one which is confusing

test/cli/test/static-collect.test.ts
```diff
@@ -427,6 +427,92 @@ test('collects tests imported from another file', async () => {
...
+    import { describe } from 'vitest';
+    import { it } from './Utils/test-extend.ts';
+      
+    vi.mock('@/composables/test.js', async (importOriginal) => { });
+      
+    describe('should included', () => {
+      it('is included because of workspace plugin setting', () => {});
+    });
+
+    describe('should included', () => {
```
sheremet-va: GIve them different names

sheremet-va (changes requested):

sheremet-va (approved):

PR #10131, opened 2026-04-12
fix: --project negation excludes browser instances

### LINKED ISSUE #9977: Negated project filtering (`!`) broken on browser projects
### Describe the bug

When running vitest against a config with multiple vitest projects, filtering out browser projects (with the negation/`!` operator) does not work. The expected behaviour and current behaviour over various filtering scenarios is laid out in the following table:

| filter               | expected # of tests | actual # of tests | result |
| -------------------- | ------------------- | ----------------- | ------ |
| *none*               | 2                   | 2                 | 🟢     |
| `--project node`     | 1                   | 1                 | 🟢     |
| `--project browser`  | 1                   | 1                 | 🟢     |
| `--project !node`    | 1                   | 1                 | 🟢     |
| `--project !browser` | 1                   | 2                 | 🔴     |

These are the above results executed on the attached reproduction:

Image

### Reproduction

https://github.com/james-camilleri/vitest-project-filtering-reproduction

### System Info

```shell
System:
    OS: Windows 11 10.0.26200
    CPU: (16) x64 AMD Ryzen AI 7 PRO 350 w/ Radeon 860M
    Memory: 3.13 GB / 23.65 GB
  Binaries:
    Node: 24.14.1 - C:\Users\JamesCamilleri\scoop\apps\nodejs-lts\current\node.EXE
    Yarn: 1.22.22 - C:\Users\JamesCamilleri\scoop\apps\nodejs-lts\current\bin\yarn.CMD
    npm: 11.11.0 - C:\Users\JamesCamilleri\scoop\apps\nodejs-lts\current\npm.CMD
    pnpm: 10.33.0 - C:\Users\JamesCamilleri\scoop\shims\pnpm.EXE
    Deno: 2.7.7 - C:\Users\JamesCamilleri\scoop\shims\deno.EXE
  Browsers:
    Edge: Chromium (145.0.3800.82)
  npmPackages:
    @vitest/browser-playwright: ^4.1.2 => 4.1.2
    vite: ^8.0.1 => 8.0.3
    vitest: ^4.1.2 => 4.1.2
```

### Used Package Manager

pnpm

### Validations

### LINKED ISSUE #10130: --project='!name' doesn't exclude browser project instances
## Describe the bug

`--project='!name'` negation patterns do not exclude browser-based projects because browser instances get a derived name with a suffix (e.g. `project-b (chromium)`), which doesn't match the negation regex.

Positive selection works correctly. Negation works for non-browser projects.

## Reproduction

https://github.com/felamaslen/vitest-project-exclude-repro

```bash
git clone https://github.com/felamaslen/vitest-project-exclude-repro.git
cd vitest-project-exclude-repro
npm install
npx playwright install chromium
```

Positive selection works:

```bash
npx vitest run --project=project-a
# Test Files  1 passed (1)
# Tests       1 passed (1)
```

Negation does not exclude the browser project:

```bash
npx vitest run '--project=!project-b'
# ✓ |project-a| a.test.ts > project-a test
# ✓ |project-b (chromium)| b.test.ts > project-b test
# Test Files  2 passed (2)
# Tests       2 passed (2)
```

**Expected:** 1 test file (only `project-a`).
**Actual:** 2 test files (both run).

## Cause

Browser projects create instances with derived names like `project-b (chromium)`. The negation regex `^(?!project\-b$)` anchors to end-of-string, so it doesn't match `project-b (chromium)`, and the instance passes through the filter.

## Workaround

Adding a trailing wildcard works:

```bash
npx vitest run '--project=!project-b*'
# Test Files  1 passed (1)
# Tests       1 passed (1)
```

## System Info

- vitest: 4.1.4
- @vitest/browser-playwright: 4.1.4
- OS: macOS (arm64)
- Node: v22

felamaslen:
## Description

`--project='!name'` did not exclude browser project instances because they have derived names like `name (chromium)`. The instance-level filtering would re-check each derived name against the negation regex, which didn't match, so the instances passed through.

When the original project name is explicitly excluded by a negation filter, all its browser instances are now excluded without falling through to per-instance filtering.

Fixes #10130

## Reproduction

https://github.com/felamaslen/vitest-project-exclude-repro

## Changes

- **`core.ts`**: Added `isExcludedByProjectFilter()` which checks if a name is explicitly matched by a negation pattern (e.g. `--project='!name'`)
- **`resolveProjects.ts`**: When the original project name is excluded by negation, return empty instances instead of falling through to per-instance filtering

### DIFF
--- a/packages/vitest/src/node/core.ts
+++ b/packages/vitest/src/node/core.ts
@@ -1643,6 +1643,24 @@ export class Vitest {
       return regexp.test(name)
     })
   }
+
+  /**
+   * Check if the project with a given name is explicitly excluded
+   * by a negation pattern (e.g. `--project='!name'`).
+   */
+  isExcludedByProjectFilter(name: string): boolean {
+    const projects = this._config?.project || this._cliOptions?.project
+    if (!projects || !projects.length) {
+      return false
+    }
+    return toArray(projects).some((project) => {
+      if (!project.startsWith('!')) {
+        return false
+      }
+      const positivePattern = project.slice(1)
+      return wildcardPatternToRegExp(positivePattern).test(name)
+    })
+  }
 }
 
 function assert(condition: unknown, property: string, name: string = property): asserts condition {
--- a/packages/vitest/src/node/core.ts
+++ b/packages/vitest/src/node/core.ts
@@ -1644,10 +1644,7 @@ export class Vitest {
     })
   }
 
-  /**
-   * Check if the project with a given name is explicitly excluded
-   * by a negation pattern (e.g. `--project='!name'`).
-   */
+  /** @internal */
   isExcludedByProjectFilter(name: string): boolean {
     const projects = this._config?.project || this._cliOptions?.project
     if (!projects || !projects.length) {
--- a/packages/vitest/src/node/projects/resolveProjects.ts
+++ b/packages/vitest/src/node/projects/resolveProjects.ts
@@ -211,23 +211,20 @@ export async function resolveBrowserProjects(
     if (!project.config.browser.enabled) {
       return
     }
+    const originalName = project.config.name
     const instances = project.config.browser.instances || []
-    if (instances.length === 0) {
+    if (instances.length === 0 || vitest.isExcludedByProjectFilter(originalName)) {
       removeProjects.add(project)
       return
     }
-    const originalName = project.config.name
-    // if original name is explicitly excluded by a negation filter, exclude all instances
     // if original name matches a positive filter, keep all instances
     // otherwise, filter instances individually (user may target a specific instance name)
-    const filteredInstances = vitest.isExcludedByProjectFilter(originalName)
-      ? []
-      : vitest.matchesProjectFilter(originalName)
-        ? instances
-        : instances.filter((instance) => {
-            const newName = instance.name! // name is set in "workspace" plugin
-            return vitest.matchesProjectFilter(newName)
-          })
+    const filteredInstances = vitest.matchesProjectFilter(originalName)
+      ? instances
+      : instances.filter((instance) => {
+          const newName = instance.name! // name is set in "workspace" plugin
+          return vitest.matchesProjectFilter(newName)
+        })
 
     // every project was filtered out
     if (!filteredInstances.length) {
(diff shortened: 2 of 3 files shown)


packages/vitest/src/node/core.ts
```diff
@@ -1643,6 +1643,24 @@ export class Vitest {
       return regexp.test(name)
     })
   }
+
+  /**
+   * Check if the project with a given name is explicitly excluded
+   * by a negation pattern (e.g. `--project='!name'`).
+   */
+  isExcludedByProjectFilter(name: string): boolean {
```
sheremet-va, line 1648: This should not be a public method. See how internal methods are handled in the file

packages/vitest/src/node/projects/resolveProjects.ts
```diff
@@ -217,13 +217,17 @@ export async function resolveBrowserProjects(
...
-    const filteredInstances = vitest.matchesProjectFilter(originalName)
-      ? instances
-      : instances.filter((instance) => {
-          const newName = instance.name! // name is set in "workspace" plugin
-          return vitest.matchesProjectFilter(newName)
-        })
+    // if original name is explicitly excluded by a negation filter, exclude all instances
+    // if original name matches a positive filter, keep all instances
+    // otherwise, filter instances individually (user may target a specific instance name)
+    const filteredInstances = vitest.isExcludedByProjectFilter(originalName)
```
sheremet-va: This should be done earlier, on L215

sheremet-va (changes requested):
This needs tests. See test/cli/browser-configs for examples

packages/vitest/src/node/projects/resolveProjects.ts
felamaslen: https://github.com/vitest-dev/vitest/pull/10131/commits/132358c76d31addd368373aee1e38849bf0e02d1

packages/vitest/src/node/core.ts
felamaslen: https://github.com/vitest-dev/vitest/pull/10131/commits/132358c76d31addd368373aee1e38849bf0e02d1

felamaslen:
Thanks for feedback - addressed

sheremet-va (approved):

AriPerkkio (approved):

Issue #3344, opened 2023-05-10
filter should exclude `cwd`


benmccann:
### Describe the bug

I did `cd ~/src` and then did a `git clone` of a Vitest project which has the command `vitest --run src`. Because my project was held in a directory named `src`, the `src` filter matched all files rendering it useless. The filter should work the same regardless of where the user checks out the project because we don't know what the user's directory structure might be and we would otherwise have to list all directory names that they should avoid putting the project into

### Reproduction

```
mkdir src
cd src
git clone git@github.com:sveltejs/kit.git
cd kit/packages/kit
pnpm install
```
Now edit `package.json` and change `"test:unit": "vitest run",` to `"test:unit": "vitest run src",` and do `rm vitest.config.js`. When you run `test:unit` all tests should pass, but it will end up running some that live in the `test` directory because the `src` directory is ignored, so some will fail

### System Info

```shell
System:
    OS: Linux 5.19 Ubuntu 22.10 22.10 (Kinetic Kudu)
  Binaries:
    Node: 18.16.0 - ~/.nvm/versions/node/v18.16.0/bin/node
    npm: 9.5.1 - ~/.nvm/versions/node/v18.16.0/bin/npm
  Browsers:
    Chrome: 113.0.5672.92
  npmPackages:
    vite: ^4.3.0 => 4.3.0
```

### Used Package Manager

pnpm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
- [X] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

sheremet-va:
The filter is just `filename.includes(filter)`. If you have a file named `something-src-something` outside of src directory it will run it.

I recommend just adding a slash to it.

sheremet-va:
[...]

PR #10139, opened 2026-04-13
feat(dev): Match filters against absolute paths of test files

florian-lackner365:
Hey 🌞 This is my first contribution. Please let me know what I need to fix to land it (or if this change is not wanted).

`jest` has this feature and since we've changed to `vitest` it's the only thing I keep missing from `jest`.

### Description

This simplifies running specific tests in monorepo setups. Imagine you have
a repo with several components like this:

components/backend/..
components/frontend/..

Each of the components contains a vitest config. If you cd into components/backend
and run `vitest components/backend/some/test/file.spec.ts`, it won't work
with relative paths, because the `component/backend/` part of the path is not recognized.

This PR matches the filters against the absolute path of a file too, so the example
works as expected.

> Why would you add the path in the first place? 

My editor offers a `copy relative path` feature that I'm using to feed the test
to `vitest`.

I've tested this PR locally with my companies project and it works 👍 

### Tests

### Documentation

*Note from the author: Do we need to document this?*

### Changesets

### DIFF
--- a/packages/vitest/src/node/project.ts
+++ b/packages/vitest/src/node/project.ts
@@ -443,7 +443,10 @@ export class TestProject {
     if (filters.length) {
       return testFiles.filter((t) => {
         const testFile = relative(dir, t).toLocaleLowerCase()
+        const absoluteTestFile = slash(t).toLocaleLowerCase()
         return filters.some((f) => {
+          const normalizedFilter = slash(f).toLocaleLowerCase()
+
           // if filter is a full file path, we should include it if it's in the same folder
           if (isAbsolute(f) && t.startsWith(f)) {
             return true
@@ -453,8 +456,9 @@ export class TestProject {
             ? join(relative(dir, f), '/')
             : relative(dir, f)
           return (
-            testFile.includes(f.toLocaleLowerCase())
+            testFile.includes(normalizedFilter)
             || testFile.includes(relativePath.toLocaleLowerCase())
+            || absoluteTestFile.includes(normalizedFilter)
           )
         })
       })
--- a/test/cli/test/reporters/default.test.ts
+++ b/test/cli/test/reporters/default.test.ts
@@ -110,7 +110,7 @@ describe('default reporter', async () => {
     // one file
     vitest.write('p')
     await vitest.waitForStdout('Input filename pattern')
-    vitest.write('a')
+    vitest.write('a.te')
     await vitest.waitForStdout('a.test.ts')
     vitest.write('\n')
     await vitest.waitForStdout('Filename pattern: a')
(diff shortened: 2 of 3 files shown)


packages/vitest/src/node/project.ts
```diff
@@ -443,7 +443,10 @@ export class TestProject {
     if (filters.length) {
       return testFiles.filter((t) => {
         const testFile = relative(dir, t).toLocaleLowerCase()
+        const absoluteTestFile = slash(t).toLocaleLowerCase()
```
florian-lackner365, line 446: I think `slash` is needed on windows because we're comparing absolute paths. I'm not on windows though so this is hard to check.

AriPerkkio (commented):
> If you cd into components/backend and run `vitest components/backend/some/test/file.spec.ts`, it won't work with relative paths, because the `component/backend/` part of the path is not recognized.


This fix doesn't sound right.

Imagine following case:

```
<path-from-fs-root>
└── components
    ├── backend
    │   ├── some
    │   │   └── test
    │   │       └── file.spec.ts
    │   │
    │   └── components
    │       └── backend
    │           └── some
    │               └── test
    │                   └── file.spec.ts
    └── frontend
```

> If you cd into components/backend

```
backend
├── some
│   └── test
│       └── file.spec.ts # File 1
│
└── components
    └── backend
        └── some
            └── test
                └── file.spec.ts # File 2
```

> and run `vitest components/backend/some/test/file.spec.ts`

With your changes, this will run `File 1`. There is no way to run `File 2`.

florian-lackner365:
> With your changes, this will run File 1. There is no way to run File 2.

Maybe I misunderstood, but with my changes vitest would run both files. 

I've tested it with my code to be sure ✅

test/cli/test/reporters/default.test.ts
```diff
@@ -110,7 +110,7 @@ describe('default reporter', async () => {
     // one file
     vitest.write('p')
     await vitest.waitForStdout('Input filename pattern')
-    vitest.write('a')
+    vitest.write('a.te')
```
florian-lackner365, line 113: just matching on `a` made the runner run other tests that led to a fail. The intention here was a partial file name of the `a.test.ts` file. The new pattern is a bit more specific and makes the test succeed.

florian-lackner365:
I just stumbled across https://github.com/vitest-dev/vitest/issues/3344. This change would re-break it. @AriPerkkio what do you think? should we abandon it? I'm starting to have mixed feelings about its usefulness. No hard feelings. I don't want to regress other peoples workflows for my own convenience. Imo being "strict" is usually better than being lenient in software that runs in automated scripts. `vitest` is such software.

AriPerkkio:
I don't think we can break #3344. Also the logic this adds feels unexpected to me.

florian-lackner365:
> I don't think we can break https://github.com/vitest-dev/vitest/issues/3344. Also the logic this adds feels unexpected to me.

I agree. Thank you for the review 🫶

PR #10153, opened 2026-04-16
fix(reporter): pass project label colors to HTML reporter

### LINKED ISSUE #10141: HTML reporter doesn't load project label color config
### Describe the bug

Just spotted this is empty on html reporter case https://github.com/vitest-dev/vitest/blob/d5a947a72976e18794787714fd9b9bba793be88e/packages/ui/client/composables/client/static.ts#L55-L57

This causes project color from config to be not loaded, e.g.

```js
export default defineConfig({
  test: {
    projects: [
      {
        test: {
          name: {
            label: 'foo',
            color: 'black',
          },
        },
      },
      {
        test: {
          name: {
            label: 'bar',
            color: 'red',
          },
        },
      },
    ],
  },
});
```

- vitest --ui

Image

- vitest --reporter=html

Image

I think we should rework this better so that UI gets more than just serialized root level config. I have a similar needs of wanting to pass per-project `browser.traceView` to be available uniformly both on UI and HTML reporter in https://github.com/vitest-dev/vitest/pull/10102

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-6yspi7fu?file=vite.config.ts

### System Info

```shell
Stackblitz
```

### Used Package Manager

npm

### Validations

elio-Wang:
The HTML reporter serialized only project names, discarding the color config. The static client then returned an empty array for getResolvedProjectLabels. Now both the metadata and the static client carry { name, color } objects, matching the live UI behavior.

Closes #10141

### Description

The HTML reporter only serialized project names as plain strings, discarding the `color` config. The static client also returned an empty array for `getResolvedProjectLabels`. This caused custom project label colors (defined in vitest config) to not render in the HTML report, even though the live UI displayed them correctly.

This PR changes `projects` in the HTML metadata from `string[]` to `{ name, color }[]`, and wires up the static client to return the actual label data.

Resolves #10141

- [x] Please check [Allow edits by maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to make review process faster. Note that this option is not available for repositories that are owned by GitHub organizations.

### Tests

### DIFF
--- a/packages/ui/client/composables/client/static.ts
+++ b/packages/ui/client/composables/client/static.ts
@@ -16,7 +16,7 @@ interface HTMLReportMetadata {
   paths: string[]
   files: RunnerTestFile[]
   config: SerializedConfig
-  projects: string[]
+  projects: { name: string; color?: string }[]
   moduleGraph: Record<string, Record<string, ModuleGraphData>>
   unhandledErrors: unknown[]
   // filename -> source
@@ -50,10 +50,10 @@ export function createStaticClient(): VitestClient {
       return metadata.config
     },
     getResolvedProjectNames: () => {
-      return metadata.projects
+      return metadata.projects.map(p => p.name)
     },
     getResolvedProjectLabels: () => {
-      return []
+      return metadata.projects
     },
     getModuleGraph: async (projectName, id) => {
       return metadata.moduleGraph[projectName]?.[id]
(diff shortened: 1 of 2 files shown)


sheremet-va:
https://github.com/vitest-dev/vitest/pull/10142 fixes it

Issue #10158, opened 2026-04-17
`toMatchAriaInlineSnapshot` cannot match empty aria trees


mrginglymus:
### Describe the bug

Thanks for adding this into core vitest! In the process of switching from my own implementation I've found that vitest's is unable to assert on an empty aria document:

```html
<button aria-hidden="true">Hidden</button>
```

as 

```ts
expect(document.body).toMatchAriaInlineSnapshot('');
```

is intepreted as being a 'fill this in later' call. I think I also encountered an issue with the non-inline variation where an empty string was not considered a valid yaml document.

### Reproduction

https://github.com/vitest-dev/vitest/compare/main...mrginglymus:vitest:empty-aria?expand=1

### System Info

```shell
System:
    OS: Linux 6.6 Ubuntu 22.04.5 LTS 22.04.5 LTS (Jammy Jellyfish)
    CPU: (24) x64 Intel(R) Core(TM) i9-10920X CPU @ 3.50GHz
    Memory: 30.00 GB / 31.24 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 22.15.1 - /home/bill/.nvm/versions/node/v22.15.1/bin/node
    npm: 10.9.2 - /home/bill/.nvm/versions/node/v22.15.1/bin/npm
    pnpm: 10.31.0 - /home/bill/.nvm/versions/node/v22.15.1/bin/pnpm
  npmPackages:
    @vitejs/plugin-basic-ssl: ^2.1.4 => 2.1.4
    @vitest/browser: workspace:* => 4.1.4
    @vitest/browser-playwright: workspace:* => 4.1.4
    @vitest/browser-preview: workspace:* => 4.1.4
    @vitest/browser-webdriverio: workspace:* => 4.1.4
    @vitest/bundled-lib: link:./bundled-lib => undefined
    @vitest/cjs-lib: link:./cjs-lib => undefined
    playwright: catalog: => 1.59.0
    vitest: workspace:* => 4.1.4
    vitest-browser-react: ^2.0.5 => 2.0.5
```

### Used Package Manager

pnpm

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vitest.dev/guide/).
[...]

PR #10171, opened 2026-04-22
feat(browser): export aria tree utils

hi-ogawa:
### Description

- (marginally) related https://github.com/vitest-dev/vitest/issues/10158

This PR exposes `ivya/aria` exports  via `import { utils } from "vitest/browser"`. This can be useful as a standalone primitive of "node to aria tree" serialization similar to `prettyDOM` does for "ndoe to html like" serialization.

```js
import { utils } from 'vitest/browser'

document.body.innerHTML = `
  <h1>Hello, World!</h1>
  <button aria-hidden="true">Hidden</button>
  <button>Visible</button>
`
const tree = utils.aria.generateAriaTree(document.body)
const yaml = utils.aria.renderAriaNode(tree)
console.log(yaml)
// - heading "Hello, World!" [level=1]
// - button "Visible""
```

The wiring is a bit mess like `__INTERNAL` and `@vitest/browser/internal/vendor-types`, but seems necessary to avoid duplicating modules and also hand-rolling same type definition. For example, I noticed `type AriaRole` is currently manually written in `packages/browser/aria-role.d.ts` but this can also come from `ivya` with similar tricks.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/browser/context.d.ts
+++ b/packages/browser/context.d.ts
@@ -2,7 +2,7 @@ import { SerializedConfig } from 'vitest'
 import { StringifyOptions, CDPSession, BrowserCommands } from 'vitest/internal/browser'
 import { ARIARole } from './aria-role.js'
 import {} from './matchers.js'
-import { __vendorIvyaAriaTypes } from '@vitest/browser/internal/vendor-types'
+import { __ivyaAriaTypes } from '@vitest/browser/internal/vendor-types'
 
 export type BufferEncoding =
   | 'ascii'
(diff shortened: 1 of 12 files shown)


packages/browser/context.d.ts
```diff
@@ -934,6 +935,23 @@ export const utils: {
...
+    generateAriaTree: typeof __ivyaAriaTypes.generateAriaTree
+    /** Renders a captured ARIA tree to the textual snapshot format. */
+    renderAriaTree: typeof __ivyaAriaTypes.renderAriaTree
+    /** Renders an ARIA template back to text. */
+    renderAriaTemplate: typeof __ivyaAriaTypes.renderAriaTemplate
+    /** Parses textual ARIA snapshot syntax into a template tree. */
+    parseAriaTemplate: typeof __ivyaAriaTypes.parseAriaTemplate
+    /** Matches a captured ARIA tree against a parsed template. */
+    matchAriaTree: typeof __ivyaAriaTypes.matchAriaTree
+  }
```
hi-ogawa, line 954: Typescript couldn't infer jsdoc from the original `ivya/aria` exports (likely due to dts bundling), so I wrote up it manually by partially duplicating the ones in ivya.

packages/browser/src/client/tester/context.ts
```diff
@@ -556,4 +556,7 @@ export const utils = {
   debug,
   getElementLocatorSelectors,
   configurePrettyDOM,
+  get aria() {
+    return __INTERNAL._aria
+  },
```
AriPerkkio, line 561: What exactly does this work-around fix? Why not `import { aria } from "./aria"` and return that here?

AriPerkkio (dismissed):
Looks good, this will allow users to use a11y tree utils without having to install another ivya next to the bundled one.

packages/browser/src/client/tester/context.ts
hi-ogawa: There's some weird `@vitest/browser` client bundle split and we have `tester/context.ts` and `tester/expect-element.ts` independently built without sharing chunks. Previously `ivya/aria` lived only in `expect-element` bundle, but this one now needs to be available from `context.ts`, so it's passed through `__INTERNAL` global.

There's other globals that are passed around like `__vitest_browser_runner__`, but haven't fully understood what is for what.

JReinhold:
This is pretty awesome! How does it compare to Playwright's `mode: 'ai' | 'default'` option? https://playwright.dev/docs/api/class-page#page-aria-snapshot

Where the AI mode also captures iframe's contents and adds unique references to each element.

hi-ogawa:
> This is pretty awesome! How does it compare to Playwright's `mode: 'ai' | 'default'` option? https://playwright.dev/docs/api/class-page#page-aria-snapshot
> 
> Where the AI mode also captures iframe's contents and adds unique references to each element.

Our implementation is stripped down port and AI part isn't included. https://github.com/vitest-dev/ivya/blob/7d84867d3576ac9c7342299e552d3c78a60a272b/src/aria/folk/injected/ariaSnapshot.ts#L17-L20

It's expected to exactly match with their `default` mode though haven't verified. I haven't looked into beyond what's needed for Vitest's `toMatchAriaSnapshot`. If it's useful as standalone or relevant to snapshot feature, we can probably try supporting that too.

AriPerkkio (approved):

PR #10174, opened 2026-04-22
feat(vitest): create `vi.when()`

### LINKED ISSUE #756: [Feature Request] Extend `vi.fn`

We previously had a disscussion that it would be usefull to have a method on `vi.fn` that will defined behaviour when the function is called with this arguments.

Also does anyone else think that `vi.fn` API is too verbose? Maybe we can tighten it a bit. Would love to see suggestions and thoughts.

### Suggested solution

 For example:

Instead of this
```ts
vi.fn((arg1) => {
  if(arg1 === '3') return true
  if(arg1 === '4') return false
  return null
})
```

We can wright something like this (API is not finalized):

```ts
vi.fn()
  .mockReturn(null)
  .on('3').mockReturn(true)
  .on('4').mockReturn(false)
```

### Alternative

_No response_

### Additional context

_No response_

### Validations

macarie:
### Description

Resolves #756

Todo:

- [x] Check types and code
- [ ] Solve `@todo`s
- [x] Write documentation
- [x] Add more tests

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/api/vi.md
+++ b/docs/api/vi.md
@@ -797,6 +797,134 @@ globalThis.IntersectionObserver === undefined
 IntersectionObserver === undefined
 ```
 
+### vi.when <Version type="experimental">5.0</Version> <Experimental /> {#vi-when}
+
+```ts
+interface WhenOptions {
+  onUnmatched?: 'throw' | 'passthrough' | ((...args: unknown[]) => unknown)
+}
+
+interface BehaviorOptions {
+  times?: number
+}
+
+function when(spy: Mock, options?: WhenOptions): When
+```
+
+Defines per-argument behaviors on a spy, replacing its implementation for the duration of the `when` chain.
+
+Call `.calledWith(...args)` on the returned object to specify which call arguments to match, then chain one or more `then*` methods to declare what the spy should return, throw, or resolve when invoked with those arguments. Arguments are matched with deep equality and support asymmetric matchers such as `expect.any()`.
+
+```ts
+const spy = vi.fn()
+
+vi.when(spy)
+  .calledWith(1)
+  .thenReturn('one')
+  .calledWith(2)
+  .thenReturn('two')
+
+expect(spy(1)).toBe('one')
+expect(spy(2)).toBe('two')
+```
+
+Available `then*` methods:
+
+| Method | Description |
+|--------|-------------|
+| `thenReturn(value, options?)` | Returns `value`. |
+| `thenReturnOnce(value)` | Returns `value` once, then falls back. |
+| `thenThrow(error, options?)` | Throws `error`. |
+| `thenThrowOnce(error)` | Throws `error` once, then falls back. |
+| `thenResolve(value, options?)` | Returns a resolved `Promise` with `value`. |
+| `thenResolveOnce(value)` | Resolves once, then falls back. |
+| `thenReject(error, options?)` | Returns a rejected `Promise` with `error`. |
+| `thenRejectOnce(error)` | Rejects once, then falls back. |
+
+The optional `times` option limits how many times a behavior applies before being exhausted. Behaviors registered for the same arguments are consumed last-in-first-out: the most recently registered behavior is tried first, and once exhausted, earlier ones act as fallbacks.
+
+```ts
+const spy = vi.fn<(key: string) => string>()
+
+vi.when(spy)
+  .calledWith('theme')
+  .thenReturn('light') // fallback, applies indefinitely
+  .thenReturn('dark', { times: 2 }) // applied first for the next 2 calls
+
+expect(spy('theme')).toBe('dark')
+expect(spy('theme')).toBe('dark')
+expect(spy('theme')).toBe('light') // falls back
+```
+
+When called with arguments that match no registered behavior, the spy falls through to its original implementation by default. Use the `onUnmatched` option to change this:
+
+- `'passthrough'` (**default**): delegates to the spy's original implementation
+- `'throw'`: throws an error listing the unmatched arguments
+- a function: called with the unmatched arguments; its return value is used
+
+```ts
+const spy = vi.fn<(id: number) => string>()
+
+vi.when(spy, { onUnmatched: 'throw' })
+  .calledWith(1)
+  .thenReturn('Alice')
+
+expect(spy(1)).toBe('Alice')
+expect(() => spy(99)).toThrow() // no behavior defined for 99
+```
+
+The `When` object returned by `vi.when` supports the [`toHaveBeenExhausted` assertion](/api/expect#tohavebeenexhausted), which passes once every registered behavior has been consumed.
+
+```ts
+const spy = vi.fn()
+const w = vi.when(spy)
+  .calledWith(1)
+  .thenReturnOnce('once')
+  .calledWith(2)
+  .thenReturn('always')
+
+expect(w).not.toHaveBeenExhausted()
+
+spy(1) // consumes the `thenReturnOnce` behavior
+spy(2) // satisfies `thenReturn` (called at least once)
+
+expect(w).toHaveBeenExhausted()
+```
+
+::: tip
+In environments that support [Explicit Resource Management](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Resource_management), you can use `using` instead of `const` to automatically restore the spy's original implementation when the containing block exits:
+
+```ts
+const spy = vi.fn(() => 'original')
+
+{
+  using w = vi.when(spy)
+    .calledWith('hello')
+    .thenReturn('mocked')
+
+  expect(spy('hello')).toBe('mocked')
+} // ← spy's original implementation is restored here
+
+expect(spy('hello')).toBe('original')
+```
+:::
+
+### vi.isWhenChain <Version type="experimental">5.0</Version> <Experimental /> {#vi-iswhenchain}
+
+```ts
+function isWhenChain(input: object): input is When
+```
+
+Returns `true` if the given value is a `When` chain created by [`vi.when`](#vi-when). If you are using TypeScript, it will also narrow down its type.
+
+```ts
+const spy = vi.fn()
+const w = vi.when(spy).calledWith(1).thenReturn(0)
+
+expect(vi.isWhenChain(w)).toBe(true)
+expect(vi.isWhenChain(spy)).toBe(false)
+```
+
 ## Fake Timers
 
 This sections describes how to work with [fake timers](/guide/mocking/timers).
--- a/docs/api/vi.md
+++ b/docs/api/vi.md
@@ -797,7 +797,7 @@ globalThis.IntersectionObserver === undefined
 IntersectionObserver === undefined
 ```
 
-### vi.when <Version>5.0</Version> {#vi-when}
+### vi.when <Version>5.0.0</Version> {#vi-when}
 
 ```ts
 interface WhenOptions {
@@ -909,7 +909,7 @@ expect(spy('hello')).toBe('original')
 ```
 :::
 
-### vi.isWhenChain <Version>5.0</Version> {#vi-iswhenchain}
+### vi.isWhenChain <Version>5.0.0</Version> {#vi-iswhenchain}
 
 ```ts
 function isWhenChain(input: object): input is When
--- a/docs/api/vi.md
+++ b/docs/api/vi.md
@@ -797,7 +797,7 @@ globalThis.IntersectionObserver === undefined
 IntersectionObserver === undefined
 ```
 
-### vi.when <Version type="experimental">5.0</Version> <Experimental /> {#vi-when}
+### vi.when <Version>5.0</Version> {#vi-when}
 
 ```ts
 interface WhenOptions {
@@ -909,7 +909,7 @@ expect(spy('hello')).toBe('original')
 ```
 :::
 
-### vi.isWhenChain <Version type="experimental">5.0</Version> <Experimental /> {#vi-iswhenchain}
+### vi.isWhenChain <Version>5.0</Version> {#vi-iswhenchain}
 
 ```ts
 function isWhenChain(input: object): input is When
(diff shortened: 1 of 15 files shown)


packages/vitest/src/integrations/mock/when.ts
```diff
@@ -0,0 +1,365 @@
...
+ * @example
+ * using w = vi.when(spy)
+ *   .calledWith('hello')
+ *   .thenReturn('HELLO')
+ */
+interface When<Fn extends Procedure> extends Disposable {
+  /**
+   * Defines behavior for a specific set of arguments.
+   *
+   * Multiple behaviors can be stacked for the same arguments. They are matched last-registered first (LIFO stack, last in first out), so earlier entries act as fallbacks once later ones are exhausted.
```
macarie, line 159: Should this be LIFO or FIFO? 👀

packages/vitest/src/integrations/mock/when.ts
```diff
@@ -0,0 +1,365 @@
...
+  (value: ReturnType, options?: OnceBehaviorOptions | undefined) => CalledWithInstance<ReturnType, Fn>
+> & Record<
+  | 'thenThrow'
+  | 'thenReject',
+  (value: unknown, options?: BehaviorOptions | undefined) => CalledWithInstance<ReturnType, Fn>
+> & Record<
+  | 'thenThrowOnce'
+  | 'thenRejectOnce',
+  (value: unknown, options?: OnceBehaviorOptions | undefined) => CalledWithInstance<ReturnType, Fn>
+>
```
macarie: todo: write these manually and add JSDocs

packages/expect/src/jest-expect.ts
```diff
@@ -1058,6 +1058,16 @@ export const JestChaiExpect: ChaiPlugin = (chai, utils) => {
     })
   })
 
+  def('toHaveBeenExhausted', function () {
+    const when = this._obj
+
+    this.assert(
+      when.isExhausted(),
+      'expected when chain to have been exhausted', // @todo improve message
```
macarie: This needs a method that returns all the unused behaviors (and proper formatting), otherwise it becomes difficult to debug.

It would also be awesome to highlight un-accessible behaviors (e.g., `calledWith(...).thenReturn('a').thenReturn('b')`) for a better DX.

packages/vitest/src/integrations/mock/when.ts
```diff
@@ -0,0 +1,531 @@
...
+          value,
+          times,
+          remaining: times,
+          called: false,
+        })
+      }
+
+      const calledWithInstance: CalledWithInstance<ScopedReturn, Fn> = ({
+        ...output,
+        thenThrow: (value, options) => {
```
macarie, line 396: todo: validate options (`times >= 0`)

packages/expect/src/index.ts
```diff
@@ -20,7 +20,7 @@ export {
 } from './jest-asymmetric-matchers'
 export { JestChaiExpect } from './jest-expect'
 export { JestExtend } from './jest-extend'
-export { addCustomEqualityTesters } from './jest-matcher-utils'
+export { addCustomEqualityTesters, getCustomEqualityTesters /* @todo */ } from './jest-matcher-utils'
```
macarie: Is there a function that already retrieves the custom matchers and uses it?

macarie (commented):
I have a small recursive package issue... I think `when` should live in `packages/spy`, but it needs `equals` from `pacakges/expect`, which depends on `packages/spy`...

Where would it make the most sense to have `when`? Because right now `expect` can't use the helper function as it lives in `pacakges/vitest` 🙃

packages/expect/src/index.ts
sheremet-va: I think we can just move it to utils/equals

docs/api/vi.md
```diff
@@ -797,6 +797,134 @@ globalThis.IntersectionObserver === undefined
 IntersectionObserver === undefined
 ```
 
+### vi.when <Version type="experimental">5.0</Version> <Experimental /> {#vi-when}
```
sheremet-va: I think it doesn't need to be experimental in v5

packages/vitest/src/integrations/mock/when.ts
```diff
@@ -0,0 +1,541 @@
...
+   *
+   * const { isExhausted, pendingBehaviors } = w.collectPendingBehaviors()
+   *
+   * expect(isExhausted).toBe(false)
+   *
+   * console.log(pendingBehaviors)
+   * // calledWith(2)
+   * //   ✗ thenReturn(42)    never called
+   */
+  '~getDiagnostics': () => {
```
sheremet-va: I think we can use `_`. The reason why `~` is usually chosen is because it's the last in the list of public APIs, but we have `@internal` flag which removes this method from the bundle, its types are only available in this repo

packages/vitest/src/integrations/mock/when.ts
```diff
@@ -0,0 +1,541 @@
...
+    }
+    default: {
+      (type satisfies never)
+
+      throw new Error(`vi.when: "${type}" is not a known method`)
+    }
+  }
+}
+
+function isExhausted(action: Behavior<unknown[], unknown>['actions'][number]): boolean {
```
sheremet-va: Can we just have a separate type instead of doing archeology in `Behavior<unknown[], unknown>['actions'][number]`?

test/unit/test/mocking/vi-when.test.ts
```diff
@@ -0,0 +1,632 @@
...
+        .thenReturn(value, { times })
+        .thenThrow(error, { times })
+        .thenResolve(value, { times })
+        .thenReject(error, { times })
+
+      const d = w['~getDiagnostics']()
+
+      expect(d.pendingBehaviors).toMatchInlineSnapshot(`
+        "calledWith("a", 0)
+          ✗ thenReturn(97, { times: 2 })                                         2 remaining (of 2)
```
sheremet-va: Shouldn't it be `2 remaining (out of 2)`?

sheremet-va (dismissed):
LGTM! Just some nitpicks

docs/api/vi.md
```diff
@@ -797,6 +797,134 @@ globalThis.IntersectionObserver === undefined
...
+interface BehaviorOptions {
+  times?: number
+}
+
+function when(spy: Mock, options?: WhenOptions): When
+```
+
+Defines per-argument behaviors on a spy, replacing its implementation for the duration of the `when` chain.
+
+Call `.calledWith(...args)` on the returned object to specify which call arguments to match, then chain one or more `then*` methods to declare what the spy should return, throw, or resolve when invoked with those arguments. Arguments are matched with deep equality and support asymmetric matchers such as `expect.any()`.
```
hi-ogawa, line 816: What's the behavior of hanging `calledWith` without `thenXxx`? For example, I'm wondering if it's expected use case to only use `calledWith` to combine with `toHaveBeenExhausted` e.g.

```js
const spy = vi.fn()
const w = vi.when(spy).calledWith(1)
expect(w).not.toHaveBeenExhausted()
spy(1)
expect(w).toHaveBeenExhausted()
```

packages/vitest/src/integrations/mock/when.ts
hi-ogawa: The doc example doesn't read well to me, so I want FIFO.

```js
   * @example
   * // Stack behaviors: first call returns `false`, subsequent calls return `true`
   * vi.when(spy)
   *   .calledWith('darkMode')
   *   .thenReturn(true)
   *   .thenReturnOnce(false)
```

docs/api/vi.md
macarie: It will use the provided fallback action/function. By default it would call the original `spy` implementation (which in this case is a noop).

There's indeed a bug with this, as the assertion would work even if it shouldn't. I wonder if it still makes sense for `toHaveBeenExhausted` to throw when there are no registered actions, as the check would be a bit more complex (checking that at least one behavior has an action) 🤔

packages/vitest/src/integrations/mock/when.ts
macarie: So here you would consider `thenReturnOnce` to be unreachable?

You'd write it like this?

```ts
vi.when(spy)
  .calledWith('darkMode')
  .thenReturnOnce(false)
  .thenReturn(true)
```

I wonder what others think of this, as the fluent/builder APIs I've used were all LIFO-style, but the other mock APIs do seem to be FIFO.

docs/api/vi.md
sheremet-va: I think throwing if there are no registered behaviours makes sense

packages/vitest/src/integrations/mock/chai.ts
```diff
@@ -0,0 +1,28 @@
...
+      if (!isWhenChain(w)) {
+        throw new TypeError(
+          `${utils.inspect(w)} is not a \`vi.when\` instance`,
+        )
+      }
+
+      const diagnostics = w._getDiagnostics()
+
+      if (diagnostics.isExhausted === isNot) {
+        throw new chai.AssertionError(isNot
```
sheremet-va: I don't think you even need to care about `negate` if you use `this.assert` here instead of throwing an error

macarie:
I think I fixed all the review comments. I fixed another edge-case I found with Hiroshi's incomplete chain example and added a couple of test cases, so hopefully the suite has enough coverage and will be able to catch regressions in the future.

packages/vitest/src/integrations/mock/when.ts
hi-ogawa: Not sure, but from my limited english interpretation, the API suggest it should naturally read like:

> when spy is called with x, then return y once, then return z (forever).

```js
vi.when(spy)
  .calledWith(x)
  .thenReturnOnce(y)
  .thenReturn(z)
```

so current order seems inverted.

hi-ogawa: This is to be decided either by internal or public pole.

sheremet-va:
It would be nice to have a new recipe now after the https://github.com/vitest-dev/vitest/pull/10226 was merged. See https://github.com/vitest-dev/vitest/issues/10381 for follow up, and https://main.vitest.dev/guide/recipes/db-transaction.html for example

Just a page explaining the use case of this feature

docs/guide/recipes/conditional-mocking.md
macarie, line 1: I have to triple check this tomorrow (examples were generated, have to check for API hallucinations).

I'm not even sure if it's a recipe or just a guide at this point, but I like the structure and that it's less abstract than what I've added previously.

docs/guide/recipes/conditional-mocking.md
```diff
@@ -0,0 +1,266 @@
...
+  await expect(getUserById(db, 2)).resolves.toEqual({ name: 'Gracie' })
+})
+```
+
+The same approach works across all mock outcome types. Here is the full set of actions and their equivalents:
+
+| Action | Returns | Equivalent to |
+|---|---|---|
+| `thenReturn(value)` | `value` | `mockReturnValue(value)` |
+| `thenThrow(error)` | throws `error` | `mockImplementation(() => { throw error })` |
```
macarie: todo: `mockThrow` is a thing.

docs/guide/recipes/conditional-mocking.md
```diff
@@ -0,0 +1,266 @@
...
+```
+
+This is particularly handy when you need to override a shared spy for just one branch of a test without manually calling cleanup code afterwards.
+
+## See also
+
+- [`vi.when`](/api/vi#vi-when)
+- [`toHaveBeenExhausted`](/api/expect#tohavebeenexhausted)
+- [`vi.isWhenChain`](/api/vi#vi-iswhenchain)
+- [Auto-Cleanup with `using`](/guide/recipes/explicit-resources) — `vi.when` also supports `using` for automatic spy restoration
```
macarie: todo: remove this, there's a section about it now.

docs/api/vi.md
```diff
@@ -797,6 +797,134 @@ globalThis.IntersectionObserver === undefined
 IntersectionObserver === undefined
 ```
 
+### vi.when <Version>5.0</Version> {#vi-when}
```
sheremet-va:
```suggestion
### vi.when <Version>5.0.0</Version> {#vi-when}
```

docs/guide/recipes/conditional-mocking.md
```diff
@@ -0,0 +1,274 @@
+---
+title: Conditional Mocking with vi.when | Recipes
+---
+
+# Conditional Mocking with `vi.when`
+
+<Version>5.0.0</Version>
```
sheremet-va:
```suggestion
```

I don't think we need the version here

docs/guide/recipes/conditional-mocking.md
```diff
@@ -0,0 +1,274 @@
...
+
+  if (id === 2) {
+    return Promise.resolve({ id: 2, name: 'Gracie' })
+  }
+
+  return Promise.resolve(undefined)
+})
+```
+
+This works, but it becomes tedious because you have to write the argument-matching logic yourself. This is something that Vitest can handle for you when using the [`vi.when`](/api/vi#vi-when) API.
```
sheremet-va:
```suggestion
This works, but it becomes tedious because you have to write the argument-matching logic yourself. This is something that Vitest can handle for you when using the [`vi.when`](/api/vi#vi-when) <Version>5.0.0</Version> API.
```

docs/guide/recipes/conditional-mocking.md
```diff
@@ -0,0 +1,274 @@
+---
+title: Conditional Mocking with vi.when | Recipes
+---
+
+# Conditional Mocking with `vi.when`
```
sheremet-va, line 5: Would be nice to also mention `vi.when` in "mocking functions" and just link here in the Learn section too (don't explain the API there for the sake of simplicity, just have a link at the bottom, I think - or a small mention in `::: tip`)

sheremet-va (dismissed):
LGTM

sheremet-va (approved):

PR #10175, opened 2026-04-22
fix(blob): skip import edges to fileless modules to prevent merge-reports crash

### LINKED ISSUE #10173: merge-reports crashes with 'Cannot read properties of undefined (reading importers)' when blob graph contains edges to file-less modules
### Describe the bug

`vitest run --merge-reports` crashes when the serialized module graph in a blob report contains an edge whose target was filtered out of serialization.

```
TypeError: Cannot read properties of undefined (reading 'importers')
 ❯ .../vitest/dist/chunks/index.DXMFO5MJ.js:3325:17
 ❯ deserializeEnvironmentModuleGraph .../vitest/dist/chunks/index.DXMFO5MJ.js:3318:21
 ❯ .../vitest/dist/chunks/index.DXMFO5MJ.js:3256:5
 ❯ .../vitest/dist/chunks/index.DXMFO5MJ.js:3254:50
 ❯ .../vitest/dist/chunks/index.DXMFO5MJ.js:3248:43
```

### Root cause

In `packages/vitest/src/node/reporters/blob.ts`:

```ts
// serialize — skips modules without `file`…
for (const [id, mod] of environment.moduleGraph.idToModuleMap.entries()) {
  if (!mod.file) continue
  const importedIds: number[] = []
  for (const importedNode of mod.importedModules) {
    if (importedNode.id) {
      importedIds.push(getIdIndex(importedNode.id))   // …but still records edges TO them
    }
  }
  modules.push([getIdIndex(id), getIdIndex(mod.file), getIdIndex(mod.url), importedIds])
}

// deserialize
serialized.modules.forEach(([id, _file, _url, importedIds]) => {
  const moduleId = serialized.idTable[id]
  const moduleNode = nodesById.get(moduleId)!
  importedIds.forEach((importedIdIndex) => {
    const importedId = serialized.idTable[importedIdIndex]
    const importedNode = nodesById.get(importedId)!   // undefined when target had no `file`
    moduleNode.importedModules.add(importedNode)
    importedNode.importers.add(moduleNode)            // 💥 TypeError
  })
})
```

`serializeEnvironmentModuleGraph` skips modules without `file`, but still records import edges whose target has an `id` even if that target has no `file`. On restore, `nodesById.get(importedId)` is `undefined` for those targets and the subsequent `.importers.add(...)` throws.

The non-null assertions on `nodesById.get(importedId)!` (and `nodesById.get(moduleId)!`) hide the invariant mismatch between serialize and deserialize.

### Introduced by

PR #9740 / commit 843554bf04aa325c9e28fc71d412660d170e87a5 (first shipped in 4.1.0-beta.6, still present on `main`).

### Trigger

Reproduces reliably with Vitest 4.1.0 + Vite 8 when sharded blob reports contain cross-module import edges that target virtual / file-less modules. Vite 8 exposes more modules that have an `id` but no `file` than Vite 6 does, which is likely why this latent bug now fires for us (our catalog bump from Vite 6 to Vite 8 is what turned green CI red).

The crash does not depend on browser mode, coverage, or UI — a plain sharded run followed by `vitest run --merge-reports=./.vitest-reports --reporter=default` is enough.

### Suggested fix

Guard against the missing node during restore:

```diff
         importedIds.forEach((importedIdIndex) => {
             const importedId = serialized.idTable[importedIdIndex]
-            const importedNode = nodesById.get(importedId)!
+            const importedNode = nodesById.get(importedId)
+            if (!importedNode) return
             moduleNode.importedModules.add(importedNode)
             importedNode.importers.add(moduleNode)
         })
```

Alternative (strictly more correct): filter `importedIds` at serialization time so only edges whose target was actually serialized are kept, then the deserializer invariant holds without a runtime guard.

### System Info

```
vitest: 4.1.0
vite:   8.0.5
node:   v22
pnpm:   10.30.1
```

### Related

- #10032 — also merge-reports + Vite 8 + Vitest 4.1, but a different symptom (v8 coverage native crash)
- #9685 — earlier merge-reports / module graph issue, closed

### Validations

- [x] Follow the Code of Conduct
- [x] Read the Contributing Guidelines
- [x] Read the docs
- [x] Checked that there isn't already an issue that reports the same bug
- [x] This is a concrete bug
- [x] Minimal reproducible example (not yet extracted from our monorepo; happy to build one if the suggested fix/analysis isn't enough to act on)

hitenkalda:
Fixes #10173

## Problem
`serializeEnvironmentModuleGraph` skips modules without `file` but still
records import edges whose target has an `id` but no `file`. On
deserialize, `nodesById.get(importedId)` returns `undefined` for those
targets and `.importers.add()` throws:

TypeError: Cannot read properties of undefined (reading 'importers')

This became visible with Vite 8 which exposes more fileless modules
(virtual/id-only) than Vite 6.

## Fix
Two-part fix:

1. **Serializer** — only emit edges whose target also has a `file`,
   keeping serializer and deserializer in sync:
```ts
   if (importedNode.id && importedNode.file) {
```

2. **Deserializer** — null-guard as safety net for blobs produced by
   older Vitest versions before this fix:
```ts
   const importedNode = nodesById.get(importedId)
   if (!importedNode) return
```

### DIFF
--- a/packages/vitest/src/node/reporters/blob.ts
+++ b/packages/vitest/src/node/reporters/blob.ts
@@ -274,7 +274,7 @@ function serializeEnvironmentModuleGraph(
 
     const importedIds: number[] = []
     for (const importedNode of mod.importedModules) {
-      if (importedNode.id) {
+      if (importedNode.id && importedNode.file) {
         importedIds.push(getIdIndex(importedNode.id))
       }
     }
@@ -320,7 +320,10 @@ function deserializeEnvironmentModuleGraph(
     const moduleNode = nodesById.get(moduleId)!
     importedIds.forEach((importedIdIndex) => {
       const importedId = serialized.idTable[importedIdIndex]
-      const importedNode = nodesById.get(importedId)!
+      const importedNode = nodesById.get(importedId)
+      if (!importedNode) {
+        return
+      }
       moduleNode.importedModules.add(importedNode)
       importedNode.importers.add(moduleNode)
     })
--- a/test/e2e/fixtures/reporters/merge-reports/fileless-edge.test.ts
+++ b/test/e2e/fixtures/reporters/merge-reports/fileless-edge.test.ts
@@ -1,0 +1,11 @@
+import { describe, expect, it, vi } from 'vitest'
+
+vi.mock('#mocks/not-on-disk.ts', () => ({ someExport: { mocked: true } }))
+// @ts-expect-error — file does not exist, mock factory satisfies the import
+import { someExport } from '#mocks/not-on-disk.ts'
+
+describe('fileless edge repro', () => {
+  it('consumes the mock', () => {
+    expect(someExport).toEqual({ mocked: true })
+  })
+})
--- a/test/e2e/fixtures/reporters/merge-reports/package.json
+++ b/test/e2e/fixtures/reporters/merge-reports/package.json
@@ -1,0 +1,6 @@
+{
+  "type": "module",
+  "imports": {
+    "#mocks/*": "./src/mocks/*"
+  }
+}
--- a/test/e2e/test/reporters/merge-reports.test.ts
+++ b/test/e2e/test/reporters/merge-reports.test.ts
@@ -424,6 +424,23 @@ test.for([
   `)
 })
 
+test('merge reports does not crash with fileless module edges', async () => {
+  await runVitest({
+    root: './fixtures/reporters/merge-reports',
+    include: ['fileless-edge.test.ts'],
+    reporters: [['blob', { outputFile: './.vitest-reports/fileless-run.json' }]],
+  })
+
+  const { stderr, exitCode } = await runVitest({
+    root: './fixtures/reporters/merge-reports',
+    mergeReports: reportsDir,
+    reporters: [['default', { isTTY: false }]],
+  })
+
+  expect(stderr).not.toContain('Cannot read properties of undefined')
+  expect(exitCode).toBe(0)
+})
+
 async function getSerializedModuleGraph(ctx: Vitest) {
   const files = ctx.state.getFiles().slice().sort((a, b) => a.filepath.localeCompare(b.filepath))
   const moduleGraphs = Object.fromEntries(


hitenkalda:
Hey! This is my first contribution to open source, and I'm a real person.

I used Claude to help me figure out the codebase and find the bug, but I:
- Read and understood what caused the problem in blob.ts
- I went through both serializeEnvironmentModuleGraph and deserializeEnvironmentModuleGraph myself.
- Checked that the two changes make sense
- Repeated the original crash report from issue #10173

The fix is simple: the serializer was saving edges to modules without files, but the deserializer thought that all referenced nodes were there. I get what every line of this PR says.

I'm happy to answer any questions you have about the code or make changes if you need them.

jltrv:
Hi @hitenkalda, thanks for your help on this issue, do you have a small reproduction you could share? I tried yesterday but failed to reproduce the bug triggered inside our complex monorepo setup

hitenkalda:
Hey, @jltrv! 

I don't have a small repro either.

The fix is based only on the invariant mismatch you found in your problem. It's clear from the code analysis that

- The serializer skips modules that don't have a file, but it still records edges to them using `importedNode.id`.
- The deserializer uses the `!` assertion to check that all nodes are in nodesById.

The fix makes these two functions work together.

This PR adds the stricter serializer fix to the 1-line null-guard that you said worked in your production CI (from your issue comment).

jltrv:
Hey @hitenkalda, I finally found a repro, see https://github.com/vitest-dev/vitest/issues/10173#issuecomment-4313840721

hitenkalda:
Hey, @jltrv found a repro! 

The trigger is a `#mocks/*` subpath alias pointing to a file that 
doesn't exist + `vi.mock` on the same path. Vite registers the module 
with `mod.file === ""` so the serializer skips it but still keeps the 
edge — which crashes on merge.

The two changes in this PR fix both sides of that.

hi-ogawa (changes requested):
Thanks for the repro. Let's add it as test case.

hitenkalda:
Added regression test based on @jltrv's repro.

The one failing check is a Playwright timeout in ui.spec.ts — 
unrelated to this PR.

PR #10178, opened 2026-04-23
feat!: require Node.js 22 and Vite 6.4

sheremet-va:
Node 20's end of life is April 30th: https://github.com/nodejs/release#release-schedule

This PR is in preparation for it


sheremet-va:
v22.12.0 also allows us to use the native `WebSocket` instead of `ws` package 👀

PR #10187, opened 2026-04-24
fix(runner): propagate chainable flags in describe.for

### LINKED ISSUE #10181: `describe.concurrent.for` doesn't enable `concurrent`
### Describe the bug

Just noticed `test.concurrent.for` enables concurrent but `describe.concurrent.for` doesn't. 

For example, the following is expected to run inner `beforeAll` of siblings in parallel, also inner `test x` and `test y` in parallel, but it doesn't.

```js
import { describe, beforeAll, afterAll, test } from 'vitest';

const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));

describe.concurrent.for(['a', 'b', 'c'])('%s', () => {
  beforeAll(async ({}, suite) => {
    console.log('!> beforeAll', suite.name, suite.concurrent);
    await sleep(10);
  });

  afterAll(async ({}, suite) => {
    console.log('!> afterAll', suite.name);
  });

  test('test x', async ({ task }) => {
    console.log('!> test x', task.suite!.name, task.concurrent);
    await sleep(10);
    console.log('!> test x (done)', task.suite!.name);
  });

  test('test y', async ({ task }) => {
    console.log('!> test y', task.suite!.name);
    await sleep(10);
    console.log('!> test y (done)', task.suite!.name);
  });
});
```

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-dmkudkvr?file=test%2Frepro.test.ts

### System Info

```shell
Stackblitz
```

### Used Package Manager

npm

### Validations

DoriAlagem:
Closes #10181

## Problem

`describe.concurrent.for` (and other chainable modifiers like `.only`, `.skip`) were not being propagated to the generated suites.

`suiteFn.each` correctly calls `getChainableContext(this).withContext()` to capture any chained flags before iterating. `suiteFn.for` was missing this — it referenced the outer `suite` variable directly, so modifiers like `concurrent` were silently dropped.

## Fix

Apply the same pattern used in `suiteFn.each`:

```ts
const context = getChainableContext(this)
const suite = context.withContext()
```

This ensures that `describe.concurrent.for`, `describe.only.for`, `describe.skip.for`, etc. all behave consistently.

## Test

Added a regression test to `test/core/test/test-for-suite.test.ts` that verifies `task.suite.concurrent` is `true` when using `describe.concurrent.for`.

### DIFF
--- a/packages/runner/src/suite.ts
+++ b/packages/runner/src/suite.ts
@@ -707,13 +707,13 @@ function createSuite() {
   }
 
   suiteFn.for = function <T>(
-    this: {
-      withContext: () => SuiteAPI
-      setContext: (key: string, value: boolean | undefined) => SuiteAPI
-    },
+    this: SuiteAPI,
     cases: ReadonlyArray<T>,
     ...args: any[]
   ) {
+    const context = getChainableContext(this)
+    const suite = context.withContext()
+
     if (Array.isArray(cases) && args.length) {
       cases = formatTemplateString(cases, args)
     }
--- a/test/core/test/test-for-suite.test.ts
+++ b/test/core/test/test-for-suite.test.ts
@@ -36,9 +36,10 @@ describe.for([
 
 // Regression test: describe.concurrent.for should propagate the concurrent flag
 // to the generated suites, matching how describe.for and test.concurrent.for behave.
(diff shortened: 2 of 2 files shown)


hi-ogawa (dismissed):
Thanks!

hi-ogawa (approved):

PR #10188, opened 2026-04-24
fix(snapshot): treat empty string as valid snapshot

mayrang:
### Description

- Related https://github.com/vitest-dev/vitest/issues/10158

`toMatchAriaInlineSnapshot('')` treats an empty string as "no snapshot provided"
because `''` is falsy in JavaScript. This makes it impossible to assert that an
element has an empty aria tree (e.g. all content is `aria-hidden="true"`).

The fix replaces 5 truthiness checks (`if (x)`, `!!x`, `x ?`, `x &&`) with
explicit `!== undefined` checks in `chai.ts`, `client.ts`, and `state.ts`.
This aligns the domain snapshot path with the non-domain path, which already
uses `expected !== undefined` at `state.ts:516`.

### Testing

Verified using the KV domain snapshot fixture (non-aria):
- Added empty string handling to KV adapter `parseExpected`
- Added `empty snapshot` test case with `toMatchKvInlineSnapshot(\`\`)`
- Updated integration test assertions to include the new case
- All 24 snapshot test files / 45 tests pass

A browser test for the empty aria tree case will be added as a follow-up once
ivya supports empty templates (vitest-dev/ivya#17).

### Impact

- No breaking changes — only changes how `""` (empty string) is distinguished
  from `undefined` (no snapshot)
- Non-domain snapshots (`toMatchInlineSnapshot`, `toMatchSnapshot`) are
  unaffected; they already use the correct `!== undefined` pattern
- Existing aria snapshot tests remain unchanged and pass as before

- [x] It's really useful if your PR references an issue where it is discussed ahead of time.
- [x] Read the Contributing Guidelines.
- [x] PR title follows conventional commits format.


mayrang:
Note: This PR intentionally doesn't include a browser test for the empty aria snapshot case. The test requires `parseAriaTemplate('')` to return an empty fragment instead of throwing, which is currently an ivya-side limitation.

I've submitted a fix for that at vitest-dev/ivya#17. Once it lands and the dependency is bumped, I'll follow up with a PR adding the test coverage.

hi-ogawa (changes requested):
We have test fixture for non-aria domain snapshot, so we should be able to test the change without waiting for https://github.com/vitest-dev/ivya/pull/17.

mayrang:
Thanks for the suggestion! Added a test using the KV domain snapshot fixture — `toMatchKvInlineSnapshot(\`\`)` with an empty object `{}`. I also added empty string handling to the KV adapter's `parseExpected` to return `{}` for empty input.

hi-ogawa (commented):
Can you do same for `test/snapshots/test/fixtures/domain/basic.test.ts`?

mayrang:
Done! Added `test('empty snapshot', () => { expect({}).toMatchKvSnapshot() })` to the file-based fixture as well.

hi-ogawa (dismissed):
Thanks. I updated kv adapter to actually test empty string, which then revealed `pollMatchDomain` case wasn't fixed fully, then fixed that. I also added test case for non-domain snapshot and confirmed that's also fixed.

hi-ogawa (approved):

PR #10190, opened 2026-04-24
feat(coverage)!: allow thresholds.perFile to accept an object

### LINKED ISSUE #10114: Support combining global coverage thresholds with minimum per-file thresholds

vladlenskiy:
### Description

`perFile: true` currently reuses the global threshold values for per-file checks — so there's no way to have e.g. `lines: 80` globally and a 50% floor per file. You either drop the global to your worst file, or set up a glob entry for each file. I ran into this while trying to stop coverage from silently rotting in one module while the overall number still looked fine.

This PR lets `perFile` also accept an object. When `perFile` is an object, top-level thresholds apply to the aggregate coverage, and the object is checked against each file separately. The existing `perFile: true` / `false` behavior is unchanged.

```ts
coverage: {
  thresholds: {
    lines: 80,
    functions: 80,
    perFile: {
      lines: 50,
      functions: 50,
    },
  },
}
```

The same threshold shortcuts are supported inside the object: `{ 100: true }` and negative values for uncovered counts.

Resolves #10114.

This supersedes the direction explored in #10122. This version also supports `{ 100: true }` and negative thresholds inside `perFile`.

CLI accepts only the boolean form (`--coverage.thresholds.perFile`); the object form is config-file only, matching how glob thresholds already work.

### Tests

### Documentation

### Changesets


packages/vitest/src/node/cli/cli-config.ts
```diff
@@ -213,10 +213,14 @@ export const cliOptionsConfig: VitestCLIOptions = {
...
+          // `Thresholds.perFile` is `boolean | object`. The object form is
+          // config-file only; on the CLI it is a plain boolean toggle. Cast
+          // to the boolean shape so the generated CLI docs do not show an
+          // empty argument placeholder.
           perFile: {
             description:
-              'Check thresholds per file. See `--coverage.thresholds.lines`, `--coverage.thresholds.functions`, `--coverage.thresholds.branches` and `--coverage.thresholds.statements` for the actual thresholds (default: `false`)',
-          },
+              'Check thresholds per file. See `--coverage.thresholds.lines`, `--coverage.thresholds.functions`, `--coverage.thresholds.branches` and `--coverage.thresholds.statements` for the actual thresholds (default: `false`). Object form is available in config files only.',
+          } as CLIOption<boolean> as never,
```
AriPerkkio:
```suggestion
              'Check thresholds per file. See `--coverage.thresholds.lines`, `--coverage.thresholds.functions`, `--coverage.thresholds.branches` and `--coverage.thresholds.statements` for the actual thresholds (default: `false`). Object form is available in config files only.',
            subcommands: null,
            argument: '<boolean>',
          },
```

packages/vitest/src/node/cli/cli-config.ts
```diff
@@ -213,10 +213,14 @@ export const cliOptionsConfig: VitestCLIOptions = {
         description: null,
         argument: '', // no displayed
         subcommands: {
+          // `Thresholds.perFile` is `boolean | object`. The object form is
+          // config-file only; on the CLI it is a plain boolean toggle. Cast
+          // to the boolean shape so the generated CLI docs do not show an
+          // empty argument placeholder.
```
AriPerkkio:
```suggestion
```

packages/vitest/src/node/types/coverage.ts
```diff
@@ -322,8 +322,14 @@ interface Thresholds {
...
-  /** Check thresholds per file. */
-  perFile?: boolean
+  /**
+   * Check thresholds per file. When set to an object, the top-level thresholds
+   * still apply to the aggregate and every file must additionally meet these
+   * per-file minimums.
+   *
+   * @default false
+   */
+  perFile?: boolean | PerFileThresholds
```
```suggestion
  perFile?: boolean | Pick<Thresholds, 100 | 'statements' | 'functions' | 'branches' | 'lines'>
```

packages/vitest/src/node/types/coverage.ts
```diff
@@ -346,6 +352,23 @@ interface Thresholds {
...
+  /** Per-file minimum for functions */
+  functions?: number
+
+  /** Per-file minimum for branches */
+  branches?: number
+
+  /** Per-file minimum for lines */
+  lines?: number
+}
+
```
```suggestion
```

packages/vitest/src/node/coverage.ts
```diff
@@ -505,69 +510,82 @@ export class BaseCoverageProvider {
       }
 
       // Construct list of coverage summaries where thresholds are compared against
-      const summaries = this.options.thresholds?.perFile
-        ? coverageMap.files().map((file: string) => ({
+      const summaries = perFileOption === true
+        ? [...coverageMap.files()].sort().map((file: string) => ({
```
AriPerkkio: What's the new Array spread and `sort()` for?

docs/config/coverage.md
```diff
@@ -233,12 +233,48 @@ Global threshold for statements.
...
+      perFile: {
+        lines: 50,
+        functions: 50,
+        branches: 50,
+        statements: 50,
+      },
+    }
+  }
+}
+```
```
AriPerkkio, line 261: `perFile` cannot be set for a glob pattern? Could that be supported too?

test/coverage-test/test/threshold-per-file.test.ts
```diff
@@ -0,0 +1,255 @@
...
+import { branch } from '../fixtures/src/branch'
+import { isEven, isOdd } from '../fixtures/src/even'
+import { sum } from '../fixtures/src/math'
+import { coverageTest, normalizeURL, runVitest, test } from '../utils'
+
+// Each test spawns a child Vitest via `runVitest`. With several outer
+// coverage projects (v8/istanbul/native) running in parallel, those child
+// processes can take a few seconds each. The default 5s test timeout is
+// enough on CI (which runs the projects sequentially) but tight locally.
+vi.setConfig({ testTimeout: 15_000 })
```
AriPerkkio: This isn't needed in other coverage tests 👀

test/coverage-test/test/threshold-per-file.test.ts
```diff
@@ -0,0 +1,255 @@
...
+
+// math.ts: 1/4 functions covered (25%). even.ts: 2/2 (100%). branch.ts: only the
+// true branch of the `if` is taken, so branches are 1/2 (50%).
+
+// Each test also gets its own `reportsDirectory` so that the inner Vitest
+// runs do not race on a shared `coverage/.tmp` directory.
+function uniqueReportsDirectory(): string {
+  return mkdtempSync(join(tmpdir(), 'vitest-threshold-per-file-'))
+}
```
AriPerkkio: This should not be needed. These tests don't run parallel.

AriPerkkio (changes requested):

vladlenskiy:
Reverted the sort on the boolean path. Kept it on the object-form check so the multi-file snapshot stays ordered.

On glob: ```perFile: true``` already works per-glob, only the object form is global-only. Could scope it per-glob too, but that's a bigger change. Fine as a follow-up, or I'll fold it in here if you'd rather.

packages/vitest/src/node/coverage.ts
```diff
@@ -495,6 +495,11 @@ export class BaseCoverageProvider {
    * Check collected coverage against configured thresholds. Sets exit code to 1 when thresholds not reached.
    */
   private checkThresholds(allThresholds: ResolvedThreshold[]) {
```
AriPerkkio, line 504: I'm starting to think we should make `this.resolveThresholds` fill everything ready at this point. The `interface ResolvedThreshold` should probably have `perFile: boolean` that's already set in `this.resolveThresholds`. Then here we could just loop through all the data, without resolving any files again.

AriPerkkio (changes requested):
I think we need a breaking change for this feature. It would not make sense for `perFile: boolean` to apply for both global thresholds and glob thresholds, while `perFile: object` applied only for non-glob files.

Right now in configuration below, `perFile` also applies to `src/utils/**`.

```ts

export default defineConfig({
  test: {
    coverage: {
      thresholds: {
        perFile: true,
        lines: 1,

        "src/utils/**": { lines: 2 },
      },
    },
  },
});
```

There's no way to set `perFile` off globally, while having it enabled in a glob only. Or enable it globally while having glob pattern without it.

```ts
coverage: {
  thresholds: {
    "src/utils/**": { lines: 2, perFile: true },
                             // ^^^^^^ ❌ Type error, not allowed at the moment.
  },
},
```

We should allow setting `coverage.thresholds.<glob-pattern>.perFile: boolean | object`. Breaking change would be that `coverage.thresholds.<glob-pattern>` would not inherit `coverage.thresholds.perFile` at all.

packages/vitest/src/node/coverage.ts
vladlenskiy: Both done, and the branch is up to date with main.

**Glob `perFile`:** `coverage.thresholds.<glob>` now takes its own `perFile` (`boolean | object`) and no longer inherits the top-level one — so you can turn it off globally and on per glob, or on globally with a glob that opts out. Marked breaking in the commit + migration guide.

**Refactor:** `resolveThresholds` precomputes `perFile` (and the resolved object form) onto each `ResolvedThreshold`, so `checkThresholds` is a single loop with no file re-resolving. `updateThresholds` uses the same per-entry `perFile`, and skips globs that matched no files (otherwise `Math.min(...[])` wrote `Infinity`).

Tests cover the glob `boolean` / `object` / `{ 100: true }` cases, the no-inheritance break, and the global-vs-glob split.

docs/config/coverage.md
```diff
@@ -233,12 +233,67 @@ Global threshold for statements.
...
+      lines: 80,
+      perFile: {
+        100: true,
+      },
+    }
+  }
+}
+```
+
+`perFile` can also be set on an individual glob-pattern threshold (documented below). Glob patterns do **not** inherit the top-level `perFile`; set it on each glob explicitly.
```
AriPerkkio:
```suggestion
`perFile` can also be set on an individual [glob-pattern threshold](/config/coverage#coverage-thresholds-glob-pattern). Glob patterns do **not** inherit the top-level `perFile`; set it on each glob explicitly.
```

packages/vitest/src/node/coverage.ts
```diff
@@ -767,6 +788,21 @@ export class BaseCoverageProvider {
...
+    return { perFile: true, perFileThresholds: null }
+  }
+
+  if (value && typeof value === 'object') {
+    return { perFile: false, perFileThresholds: resolveGlobThresholds(value) }
+  }
+
+  return { perFile: false, perFileThresholds: null }
+}
+
```
AriPerkkio: Let's make this start resolving from one level higher, so that we don't need type assertions when passing glob entries:

```diff
- ...resolvePerFile((globEntry as { perFile?: unknown }).perFile),
+ ...resolvePerFile(globEntry),
```

So `value: unknown` is object that may contain `perFile`.

```ts
if (!value || !("perFile" in value) {
 ...
 etc
```

packages/vitest/src/node/coverage.ts
```diff
@@ -495,80 +502,88 @@ export class BaseCoverageProvider {
...
-        && thresholds.lines === undefined
-        && thresholds.statements === undefined
+        thresholds.branches !== undefined
+        || thresholds.functions !== undefined
+        || thresholds.lines !== undefined
+        || thresholds.statements !== undefined
       ) {
+        // Sort so that per-file error output stays stable across providers and runs.
+        const summaries = perFile
+          ? [...coverageMap.files()].sort().map((file: string) => ({
```
AriPerkkio: Can you explain the comment? How does the output look without sorting?

Also no need for new arrays. `.files()` returns new array, so we can do just `coverageMap.files().sort().map(...`
- https://github.com/istanbuljs/istanbuljs/blob/28ffdbc314596bdcb3007e85d30a62372602b262/packages/istanbul-lib-coverage/lib/coverage-map.js#L81-L87

packages/vitest/src/node/coverage.ts
```diff
@@ -495,80 +502,88 @@ export class BaseCoverageProvider {
...
+
+        for (const { summary, file } of summaries) {
+          this.reportThresholdViolations(thresholds, summary, file, label)
+        }
+      }
+
+      if (
+        perFileThresholds
+        && THRESHOLD_KEYS.some(key => perFileThresholds[key] !== undefined)
+      ) {
```
AriPerkkio: Can we keep similar style as global thresholds have above?


```suggestion
      if (
        perFileThresholds?.branches !== undefined
        || perFileThresholds?.functions !== undefined
        || perFileThresholds?.lines !== undefined
        || perFileThresholds?.statements !== undefined
      ) {
```

packages/vitest/src/node/coverage.ts
AriPerkkio: Would this happen if we gathered all info into an array before calling `reportThresholdViolations`? Did some refactoring and ended-up with following:

```ts
  /**
   * Check collected coverage against configured thresholds. Sets exit code to 1 when thresholds not reached.
   */
  private checkThresholds(allThresholds: ResolvedThreshold[]) {
    for (const { coverageMap, thresholds, perFile, perFileThresholds, name } of allThresholds) {
      const groups = []

      // Aggregate summary
      if (!perFile) {
        groups.push({
          file: null,
          thresholds,
          summary: coverageMap.getCoverageSummary(),
          name: name === GLOBAL_THRESHOLDS_KEY ? name : `"${name}"`,
        })
      }

      // Per-file summmary against global thresholds
      if (perFile) {
        for (const file of coverageMap.files().sort()) {
          groups.push({
            file,
            thresholds,
            summary: coverageMap.fileCoverageFor(file).toSummary(),
            name: name === GLOBAL_THRESHOLDS_KEY ? name : `"${name}"`,
          })
        }
      }

      // Per-file summmary with its own thresholds
      if (perFileThresholds) {
        for (const file of coverageMap.files().sort()) {
          groups.push({
            file,
            thresholds: perFileThresholds,
            summary: coverageMap.fileCoverageFor(file).toSummary(),
            name: 'per-file',
          })
        }
      }

      for (const group of groups) {
        if (
          group.thresholds.branches === undefined
          && group.thresholds.functions === undefined
          && group.thresholds.lines === undefined
          && group.thresholds.statements === undefined
        ) {
          continue
        }

        this.reportThresholdViolations(group.thresholds, group.summary, group.file, group.name)
      }
    }
  }
```

test/coverage-test/test/threshold-per-file.test.ts
```diff
@@ -0,0 +1,277 @@
...
+      },
+    },
+  }, { throwOnError: false })
+
+  expect(exitCode).toBe(1)
+  expect(stderr).toMatchInlineSnapshot(`
+    "ERROR: Coverage for functions (25%) does not meet per-file threshold (30%) for fixtures/src/math.ts
+    "
+  `)
+})
```
AriPerkkio: We can remove this test,  as it's identical to `test('per-file object thresholds fail while global thresholds pass')`.

test/coverage-test/test/threshold-per-file.test.ts
```diff
@@ -0,0 +1,277 @@
...
+      },
+    },
+  }, { throwOnError: false })
+
+  expect(exitCode).toBe(1)
+  expect(stderr).toMatchInlineSnapshot(`
+    "ERROR: Coverage for branches (50%) does not meet per-file threshold (80%) for fixtures/src/branch.ts
+    "
+  `)
+})
```
AriPerkkio: Let's remove this test and add `src/branch.ts` as part of `test('per-file object thresholds with { 100: true }')` test.

AriPerkkio: The `test('per-file object thresholds report violations for every failing file')` below can also be removed after this change.

AriPerkkio (changes requested):
This is starting to look good! Some more comments below. These might be the last change requests.

test/coverage-test/test/threshold-per-file.test.ts
vladlenskiy: Applied the `groups` refactor: threshold checks are collected first, then reported from a single loop at the end. The threshold-key guard is folded into that loop, so the separate `perFileThresholds` block is gone.

Tests: removed the redundant per-file cases and moved `branch.ts` into the `{ 100: true }` test.

Two small notes: I kept the explicit type on `groups`, and the `"perFile" in thresholds` check needs the `typeof thresholds === 'object'` guard to type-check.

vladlenskiy:
Hi @AriPerkkio — addressed all the latest comments and resolved the threads, merged in latest main. Ready for another look — thanks!

AriPerkkio (approved):
Looks good! Thanks for addressing all feedback 💯

PR #10192, opened 2026-04-25
fix(mocker)!: deserialize automock as automock

### LINKED ISSUE #9706: `vi.mock()` does not replace the modules implementation in browser mode
### Describe the bug

`vi.mock('some-module')` calls do not replace the modules implementation when running in browser mode. Instead, they appear to behave as if `vi.mock('some-module', {spy: true})` was used and wrap the original implementation in spys.

As a workaround, you can setup empty mocks using a factory function or by importing the mocked functions and giving them a new `mockImplementation`

### Reproduction

https://github.com/bimbiltu/vitest-browser-mock-repro

This repo has 3 simple files:
```ts
// src/answer.ts
export const answer = () => 42;
```

```ts
// src/getTheAnswer.ts
import { answer } from './answer';
export const getTheAnswer = () => answer();
```

```ts
import { it, expect, vi } from 'vitest';
import { getTheAnswer } from './getTheAnswer';
import { answer } from './answer';

vi.mock('./answer')

// fails but should pass
it('should not return 42 when the answer module is mocked', () => {
  expect(getTheAnswer()).not.toBe(42);
});

it('should replace answer with a mock', () => {
  getTheAnswer();
  expect(answer).toHaveBeenCalled();
});
```

### System Info

```shell
System:
    OS: macOS 26.3
    CPU: (10) arm64 Apple M1 Pro
    Memory: 1.58 GB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.13.1 - /Users/USER/.asdf/installs/nodejs/24.13.1/bin/node
    Yarn: 1.22.22 - /Users/USER/.asdf/installs/nodejs/24.13.1/bin/yarn
    npm: 11.8.0 - /Users/USER/.asdf/plugins/nodejs/shims/npm
    pnpm: 8.9.2 - /opt/homebrew/bin/pnpm
    Watchman: 2025.08.04.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 145.0.7632.77
    Firefox: 140.7.0
    Safari: 26.3
  npmPackages:
    @vitest/browser-playwright: ^4.0.18 => 4.0.18
    vitest: ^4.0.18 => 4.0.18
```

### Used Package Manager

npm

### Validations

nami8824:
### Description

Resolves #9706

### Tests

### Documentation

### Changesets

### DIFF
--- a/test/cli/fixtures/mocker/autospy/index.js
+++ b/test/cli/fixtures/mocker/autospy/index.js
@@ -1,7 +1,7 @@
 import { mocker } from 'virtual:mocker'
 import { calculate } from './test'
 
-mocker.customMock(import('./test'))
+mocker.customMock(import('./test'), { spy: true })
 
 document.querySelector('#mocked').textContent = calculate(1, 2)
 
--- a/packages/mocker/src/registry.ts
+++ b/packages/mocker/src/registry.ts
@@ -174,8 +174,8 @@ export class AutomockedModule {
     public url: string,
   ) {}
 
-  static fromJSON(data: AutomockedModuleSerialized): AutospiedModule {
-    return new AutospiedModule(data.raw, data.id, data.url)
+  static fromJSON(data: AutomockedModuleSerialized): AutomockedModule {
+    return new AutomockedModule(data.raw, data.id, data.url)
   }
 
   toJSON(): AutomockedModuleSerialized {
--- a/test/browser/fixtures/mocking/automocked-default-return.test.ts
+++ b/test/browser/fixtures/mocking/automocked-default-return.test.ts
@@ -1,0 +1,8 @@
+import { expect, test, vi } from 'vitest'
+import { calculator } from './src/calculator'
+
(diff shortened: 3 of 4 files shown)


test/cli/fixtures/mocker/autospy/index.js
```diff
@@ -1,7 +1,7 @@
 import { mocker } from 'virtual:mocker'
 import { calculate } from './test'
 
-mocker.customMock(import('./test'))
+mocker.customMock(import('./test'), { spy: true })
```
nami8824, line 4: This fix caused the `"autospy works correctly"` test case in `test/cli/test/public-mocker.test.ts` to fail, because the automock was no longer being deserialized as a spy. To fix it, I added `{ spy: true }` to the `mocker.customMock()` call in the autospy fixture. I think passing `{ spy: true }` explicitly is the correct behavior here.

sheremet-va (approved):

PR #10202, opened 2026-04-26
fix(reporter): guard against non-finite slowTestThreshold in summary reporter

### LINKED ISSUE #10155: `slowTestThreshold: Infinity` triggers Node's `TimeoutOverflowWarning`
### Describe the bug

Noticed a user using `slowTestThreshold` with `Infinity`, probably to avoid seeing any reporter warnings.

https://github.com/vitest-dev/vitest/blob/6abd557b7219156893dd13a1dbe86501d5542d2e/packages/vitest/src/node/reporters/summary.ts#L149-L151

```
(node:96948) TimeoutOverflowWarning: Infinity does not fit into a 32-bit signed integer.
Timeout duration was set to 1.
    at new Timeout (node:internal/timers:179:17)
    at setTimeout (node:timers:138:19)
    at SummaryReporter.onHookStart (file:///x/node_modules/vitest/dist/chunks/index.DWDW6mLz.js:1134:19)
    at TreeReporter.onHookStart (file:///x/node_modules/vitest/dist/chunks/index.DWDW6mLz.js:1323:17)
    at file:///x/node_modules/vitest/dist/chunks/cli-api.CdZ6wo9-.js:12544:56
    at Array.map (<anonymous>)
    at Vitest.report (file:///x/node_modules/vitest/dist/chunks/cli-api.CdZ6wo9-.js:12544:36)
    at TestRun.reportEvent (file:///x/node_modules/vitest/dist/chunks/cli-api.CdZ6wo9-.js:11475:52)
    at TestRun.updated (file:///x/node_modules/vitest/dist/chunks/cli-api.CdZ6wo9-.js:11403:54)
    at processTicksAndRejections (node:internal/process/task_queues:105:5)
    at Proxy.onTaskUpdate (file:///x/node_modules/vitest/dist/chunks/cli-api.CdZ6wo9-.js:7076:9)
    at EventEmitter.onMessage (file:///x/node_modules/vitest/dist/chunks/index.0kCJoeWi.js:157:20)
```

### Reproduction

https://stackblitz.com/~/edit/vitest-dev-vitest-uggyc9t9?file=vite.config.ts:L11&initialPath=/__vitest__/

### System Info

```shell
Stackblitz
```

### Used Package Manager

npm

### Validations

OfekDanny:
## Summary

Closes #10155

When `slowTestThreshold` is set to `Infinity` (the default), passing it directly to `setTimeout` causes Node.js to emit:

```
TimeoutOverflowWarning: Infinity does not fit into a 32-bit signed integer.
```

This happens in two places inside `SummaryReporter`:
- `onHookStart` — schedules a timeout to make a slow hook visible
- `onTestCaseReady` — schedules a timeout to make a slow test visible

**Fix:** Add a `Number.isFinite()` guard before scheduling the timeout. When the threshold is non-finite, no timeout is created, which is the correct semantic — tests should never be marked as "slow" when there is no finite threshold.

## Test plan

- [x] Added a regression test in `test/cli/test/reporters/default.test.ts` that runs a test suite with `slowTestThreshold: Infinity` and asserts that `stderr` contains no `TimeoutOverflowWarning`
- [x] Existing reporter tests continue to pass

### DIFF
--- a/packages/vitest/src/node/reporters/summary.ts
+++ b/packages/vitest/src/node/reporters/summary.ts
@@ -146,6 +146,10 @@ export class SummaryReporter implements Reporter {
     stats.hook?.onFinish?.()
     stats.hook = hook
 
+    if (!Number.isFinite(this.ctx.config.slowTestThreshold)) {
+      return
+    }
+
     const timeout = setTimeout(() => {
       hook.visible = true
     }, this.ctx.config.slowTestThreshold).unref()
@@ -183,9 +187,11 @@ export class SummaryReporter implements Reporter {
       onFinish: () => {},
     }
 
-    const timeout = setTimeout(() => {
-      slowTest.visible = true
-    }, this.ctx.config.slowTestThreshold).unref()
+    const timeout = Number.isFinite(this.ctx.config.slowTestThreshold)
+      ? setTimeout(() => {
+          slowTest.visible = true
+        }, this.ctx.config.slowTestThreshold).unref()
+      : undefined
 
     slowTest.onFinish = () => {
       slowTest.hook?.onFinish()
--- a/test/e2e/test/reporters/default.test.ts
+++ b/test/e2e/test/reporters/default.test.ts
@@ -395,4 +395,13 @@ describe('default reporter', async () => {
     })
     expect(stderr).toMatchSnapshot()
   })
+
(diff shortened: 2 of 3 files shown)


AriPerkkio (approved):
Looks good, thanks.

PR #10203, opened 2026-04-26
fix(runner): tag options should override inherited suite options

### LINKED ISSUE #10199: `test.tags` options doesn't overwrite inherited suite options
### Describe the bug

Found yet another bug that indirectly causes `test.sequential` vs `test(..., { concurrent: false })` behavior divergence.

- vite.config.ts

```js

export default defineConfig({
  test: {
    tags: [
      {
        name: 'my-tag',
        timeout: 5000,
        repeats: 2,
        concurrent: false,
      },
    ],
  },
});
```

- repro.test.ts

```js
import { describe, test } from 'vitest';

describe(
  'with suite options',
  { timeout: 1000, repeats: 1, concurrent: true },
  () => {
    test('bad', { tags: ['my-tag'] }, ({ task }) => {
      console.log({
        timeout: task.timeout,
        repeats: task.repeats,
        concurrent: task.concurrent,
      });
    });
  }
);

describe('without suite options', () => {
  test('good', { tags: ['my-tag'] }, ({ task }) => {
    console.log({
      timeout: task.timeout,
      repeats: task.repeats,
      concurrent: task.concurrent,
    });
  });
});
```

The test run outputs:

```js
stdout | test/repro.test.ts > with suite options > bad
{ timeout: 1000, repeats: 1, concurrent: true }

stdout | test/repro.test.ts > with suite options > bad
{ timeout: 1000, repeats: 1, concurrent: true }

stdout | test/repro.test.ts > without suite options > good
{ timeout: 5000, repeats: 2, concurrent: undefined }

stdout | test/repro.test.ts > without suite options > good
{ timeout: 5000, repeats: 2, concurrent: undefined }

stdout | test/repro.test.ts > without suite options > good
{ timeout: 5000, repeats: 2, concurrent: undefined }

 ✓ test/repro.test.ts (2 tests) 6ms
   ✓ with suite options (1)
     ✓ bad 3ms (repeat x1)
   ✓ without suite options (1)
     ✓ good 2ms (repeat x2)
```

I expect the inner `tags` options should override suite options.

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-3uvkkhlm?file=vite.config.ts

### System Info

```shell
stackblitz
```

### Used Package Manager

npm

### Validations

OfekDanny:
## Summary

Closes #10199

When a test is inside a `describe` with options (e.g. `timeout`, `repeats`, `concurrent`) **and** has a tag that defines those same options, the tag options were silently ignored. The old code used:

```ts
options = Object.assign({}, suiteOptions, options)
```

This gave `suiteOptions` higher priority than the tag options that had already been merged into `options`, so a tag could never override a suite setting.

**Expected priority (lowest → highest):**

```
suite options < tag options < explicit test options
```

### Changes

**`packages/runner/src/suite.ts`**

1. The internal `task()` function now accepts an optional `inheritedSuiteOptions` argument and spreads it *first* in the merge chain — before tag options, before test-own options — so the three levels are correctly ordered.

2. The outer `test` factory no longer pre-merges `suiteOptions` into `options` before calling `task()`. Instead it passes `suiteOptions` as the new third argument. The `concurrent`/`sequential` resolution is updated to only inject the resolved value into `testOwnOptions` when the chainable or the test itself explicitly sets it (so a tag's `concurrent` value can still override a purely-inherited suite value).

## Test plan

- [x] Added integration test in `test/cli/test/test-tags.test.ts` covering:
  - Tag `timeout`/`repeats` wins over suite `timeout`/`repeats`
  - Explicit test `timeout` still beats the tag
  - Tags apply normally when there is no parent suite
- [x] All existing tag tests continue to pass


packages/runner/src/suite.ts
```diff
@@ -443,29 +444,39 @@ function createSuiteCollector(
...
+        || testOwnOptions.sequential != null
+
+    if (hasExplicitConcurrency) {
+      if (concurrent != null) {
+        testOwnOptions = { ...testOwnOptions, concurrent }
+      }
+      if (sequential != null) {
+        testOwnOptions = { ...testOwnOptions, sequential }
+      }
     }
```
hi-ogawa: We just removed `sequential` https://github.com/vitest-dev/vitest/pull/10198 so this monster can likely go away.  Can you merge main and update this logic?

hi-ogawa (changes requested):

OfekDanny:
Good call — merged main and simplified. The  block and all the `sequential` references are gone now. The logic in `createSuiteCollector`'s test function is now just:

```ts
const concurrent = this.concurrent ?? testOwnOptions?.concurrent
if (concurrent != null) {
  testOwnOptions = { ...testOwnOptions, concurrent }
}
```

The original fix still holds: when neither the chainable nor the test's own options set `concurrent`, we leave `testOwnOptions` without a `concurrent` key, so the tag's value propagates correctly through `task()`'s merge order.

hi-ogawa:
Please fix CI. There's a new tests added by https://github.com/vitest-dev/vitest/pull/10198

OfekDanny:
Done — merged main and simplified the logic. Since `sequential` is gone after #10198, the `hasExplicitConcurrency` block is no longer needed. The current code just does:

```ts
const concurrent = this.concurrent ?? testOwnOptions?.concurrent
if (concurrent != null) {
  testOwnOptions = { ...testOwnOptions, concurrent }
}
```

This keeps the correct priority (suite < tag < test-own/chainable) without any of the old complexity. Also updated the snapshot that #10198 added.

PR #10205, opened 2026-04-26
fix(expect): preserve _poll.assert_once flag across assertion failures

### LINKED ISSUE #10204: Bug: toMatchScreenshot retry loop never stops until timeout due to _poll.assert_once flag not being set correctly
## Description

When using `expect.element().toMatchScreenshot()` with a custom comparator that returns `{ pass: false }`, the test doesn't fail immediately. Instead, it keeps retrying until the test timeout (default 30s), causing the test to timeout rather than failing fast with a clear error message.

## Root Cause Analysis

After investigating the code, I found the issue is in how `_poll.assert_once` flag is set in the browser expect implementation.

### The Poll Mechanism

Vitest's chai poll mechanism in `chunks/test.DNmyFkvJ.js`:

```javascript
while (true) {
  const isLastAttempt = hasTimedOut;
  if (isLastAttempt) chai.util.flag(assertion, "_isLastPollAttempt", true);
  try {
    executionPhase = "fn";
    const obj = await fn();  // <-- fn() is called without binding 'this'
    chai.util.flag(assertion, "object", obj);
    executionPhase = "assertion";
    const output = await assertionFunction.call(assertion, ...args);
    return output;  // Success: exit loop
  } catch (err) {
    if (isLastAttempt || executionPhase === "assertion" && chai.util.flag(assertion, "_poll.assert_once")) {
      throwWithCause(err, STACK_TRACE_ERROR);  // Should fail immediately
    }
    await delay(interval, setTimeout);  // Otherwise retry
  }
}
```

### The Problem

In `@vitest/browser/dist/expect-element.js`, the `bn` function (which handles `expect.element()`):

```javascript
n.poll(function() {
  // ...
  if(n===`toMatchScreenshot`&&!r.util.flag(this,`_poll.assert_once`)&&r.util.flag(this,`_poll.assert_once`,!0), ...)
  // ...
}, he(t))
```

The `_poll.assert_once` flag is set using `r.util.flag(this, '_poll.assert_once', true)` inside the `fn()` callback.

**However**, `fn()` is called as `await fn()` without binding `this` to the `assertion` object. This means `this` inside the callback is NOT the assertion object, so the flag is set on the wrong object.

Later, when checking the condition:
```javascript
chai.util.flag(assertion, "_poll.assert_once")  // Returns undefined!
```

The condition `executionPhase === "assertion" && _poll.assert_once` evaluates to `false`, causing vitest to retry instead of failing immediately.

## Reproduction

```typescript
// vitest.config.ts
import { goldenComparator } from './e2e/utils/golden-comparator'

declare module '@vitest/browser/context' {
  interface ScreenshotComparatorRegistry {
    golden: { fixtureName: string }
  }
}

export default defineConfig({
  test: {
    projects: [{
      test: {
        name: 'e2e',
        browser: {
          enabled: true,
          expect: {
            toMatchScreenshot: {
              comparators: { golden: goldenComparator }
            }
          }
        }
      }
    }]
  }
})

// Comparator that returns { pass: false }
export function goldenComparator(reference, _actual, options) {
  return { pass: false, message: 'Size mismatch: 150x100 vs 1x1' }
}

// Test
test('screenshot comparison', async () => {
  await expect.element(element).toMatchScreenshot('test', {
    comparatorName: 'golden',
    comparatorOptions: { fixtureName: 'test' }
  })
})
```

Expected: Test fails immediately with "Size mismatch" error
Actual: Test times out after 30 seconds

## Suggested Fix

Two possible approaches:

1. **In vitest core**: Bind `this` when calling `fn()`:
   ```javascript
   const obj = await fn.call(assertion);
   ```

2. **In @vitest/browser**: Don't rely on `this` inside the poll callback. Instead, set the flag after `fn()` returns:
   ```javascript
   const obj = await fn();
   chai.util.flag(assertion, "_poll.assert_once", true);
   ```

## Environment

- vitest: 4.1.5
- @vitest/browser: 4.1.5
- Node.js: 22.x
- OS: macOS

yingzhi0808:
## Summary

This PR fixes two issues with `expect.element().toMatchScreenshot()`:

### Issue 1: `_poll.assert_once` flag being cleared by chai

When a custom comparator returns `{ pass: false }`, the test would timeout instead of failing immediately.

**Root Cause**: The `_poll.assert_once` flag is used to indicate that the assertion should only run once after the element is resolved. However, chai may reset flags when an assertion fails, causing the flag check to return `undefined` on subsequent retry attempts. This results in vitest retrying indefinitely until timeout instead of failing immediately.

**Fix**: 
- Capture the `_poll.assert_once` flag value before entering the assertion phase and store it in a local variable that chai cannot modify
- Use this stored value in the catch block to determine whether to fail immediately
- Re-set the flag before retry to ensure it persists
- Move flag setting before early returns in expect-element.ts

### Issue 2: `matched-immediately` outcome bypassing final comparison

When `retries === 0` (page stable on first screenshot), vitest would return `matched-immediately` without calling the comparator with `createDiff=true`.

**Root Cause**: The stability check uses `createDiff=false` (pixel-level comparison), while the final comparison uses `createDiff=true` (SSIM or custom metrics). Stability does NOT guarantee match with reference - a page can be stable (two consecutive screenshots identical) but still differ from the reference.

**Fix**: Remove the `matched-immediately` outcome and always call the comparator with `createDiff=true` for the final comparison, ensuring consistent behavior regardless of how quickly the page stabilized.

## Changes

1. `packages/vitest/src/integrations/chai/poll.ts`:
   - Added `shouldAssertOnce` local variable to capture flag before assertion phase
   - Modified catch block to use stored value instead of chai's flag
   - Re-set the flag after catching error to ensure persistence

2. `packages/browser/src/client/tester/expect-element.ts`:
   - Moved `_poll.assert_once` flag setting before early returns

3. `packages/browser/src/node/commands/screenshotMatcher/index.ts`:
   - Removed `matched-immediately` outcome type
   - Removed early return when `retries === 0`
   - Updated `buildOutput` function
   - Updated documentation comments

## Test Plan

1. Create a custom screenshot comparator that returns `{ pass: false }`
2. Use `expect.element().toMatchScreenshot()` with this comparator
3. Verify the test fails immediately with the comparator's error message instead of timing out
4. Verify that all fixtures generate `actual.png` regardless of stability speed

Fixes #10204

### DIFF
--- a/packages/browser/src/client/tester/expect-element.ts
+++ b/packages/browser/src/client/tester/expect-element.ts
@@ -17,12 +17,18 @@ function element<T extends HTMLElement | SVGElement | null | Locator>(elementOrL
   }
 
   const expectElement = expect.poll<HTMLElement | SVGElement | null>(function element(this: object) {
+    // Capture the matcher name early to set assert_once flag before any early returns
+    // This is critical for screenshot comparisons which should only run once
+    const name = chai.util.flag(this, '_name') as string
+    if (name === 'toMatchScreenshot' && !chai.util.flag(this, '_poll.assert_once')) {
+      chai.util.flag(this, '_poll.assert_once', true)
+    }
+
     if (elementOrLocator instanceof Element || elementOrLocator == null) {
       return elementOrLocator
     }
 
     const isNot = chai.util.flag(this, 'negate') as boolean
-    const name = chai.util.flag(this, '_name') as string
     // special case for `toBeInTheDocument` matcher
     if (isNot && name === 'toBeInTheDocument') {
       return elementOrLocator.query()
@@ -31,11 +37,6 @@ function element<T extends HTMLElement | SVGElement | null | Locator>(elementOrL
       // we know that `toHaveLength` requires multiple elements,
       // but types generally expect a single one
       return elementOrLocator.elements() as unknown as HTMLElement
-    }
-
-    if (name === 'toMatchScreenshot' && !chai.util.flag(this, '_poll.assert_once')) {
-      // `toMatchScreenshot` should only run once after the element resolves
-      chai.util.flag(this, '_poll.assert_once', true)
     }
 
     // element selector uses prettyDOM under the hood, which is an expensive call
--- a/packages/browser/src/node/commands/screenshotMatcher/index.ts
+++ b/packages/browser/src/node/commands/screenshotMatcher/index.ts
@@ -27,8 +27,7 @@ interface ScreenshotData {
  * - `unstable-screenshot`: page never stabilized within timeout
  * - `missing-reference`: no baseline exists to compare against
  * - `update-reference`: snapshot update was requested (run with `--update`)
- * - `matched-immediately`: screenshot matched reference on first capture (no retries)
- * - `matched-after-comparison`: screenshot matched after another comparison
+ * - `matched-after-comparison`: screenshot matched reference after comparison
  * - `mismatch`: screenshot differs from reference
  */
 type MatchOutcome
@@ -45,7 +44,6 @@ type MatchOutcome
     type: 'update-reference'
     reference: ScreenshotData
   }
-  | { type: 'matched-immediately' }
   | { type: 'matched-after-comparison' }
   | {
     type: 'mismatch'
@@ -175,11 +173,6 @@ async function determineOutcome(
     }
   }
 
-  // first capture matched reference (used as baseline) - no further comparison needed
-  if (retries === 0) {
-    return { type: 'matched-immediately' }
-  }
-
   const comparisonResult = await comparator(
     reference,
     screenshot,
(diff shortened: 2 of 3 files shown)


yingzhi0808:
I am a real person behind this contribution.

Please reopen this PR — other people have encountered the same bug, and this fix resolves it. The bug has been bothering me for months. I patched this PR locally and verified that it solves both my issue and the issues others are facing.

I confirm that I have personally reviewed the code and stand behind its content. The PR was created with the help of Claude Code, which I used as a tool. Initially, the fix was implemented as a pnpm patch locally. After confirming it worked correctly, I asked Claude Code to convert it into a formal PR to ensure the solution can help others who hit the same problem.

yingzhi0808:
related issues @#10211 @#10204

PR #10206, opened 2026-04-26
feat(browser): stream trace entries to server instead of accumulating client-side

### LINKED ISSUE #10197: Trace view - Stream trace entries incrementally in watch mode so replay can look live/in-progress
Part of https://github.com/vitest-dev/vitest/issues/10156

This looks like a critical improvement to make current default headed live view obsolete. I'm not so convinced but Codex made a draft plan to start with:

---

<details><summary>Plan 🤖</summary>

> [!NOTE]
> The code reference links are invalid unless you view this document locally on Vitest repo with `.dev-notes/dist/*.md

# Trace View Live Replay Plan

Context: follow-up to PR [#10102](https://github.com/vitest-dev/vitest/pull/10102) and issue [#10156](https://github.com/vitest-dev/vitest/issues/10156).

Goal: let the trace viewer replay snapshots while the test is still running in watch mode, instead of waiting for the attempt-complete `internal:browserTrace` artifact.

## Summary

This looks feasible without inventing a brand-new transport. The key observation is that Vitest already supports mid-run artifact delivery to live UI clients, but the current trace-view pipeline only emits one final artifact at attempt end.

Recommended direction:

- Keep the current final `internal:browserTrace` artifact as the canonical completed artifact for HTML/static/reporting and backwards compatibility.
- Add a new append-only internal artifact for in-progress trace chunks in watch mode.
- Let the UI maintain a separate ephemeral live-trace buffer keyed by attempt id and prefer that buffer while the test is running.
- Once the attempt finishes, swap back to the final canonical artifact and clear the ephemeral buffer.

This gives us a low-risk first implementation. If chunk traffic later proves too expensive, we can add a subscription-based optimization path without throwing away the model.

## Current Flow

Current collection and delivery path:

1. Browser-side trace entries are collected in memory per attempt in [packages/browser/src/client/tester/trace.ts#L80](../../packages/browser/src/client/tester/trace.ts#L80).
2. The active attempt is created in `onBeforeTryTask` and only flushed on `onAfterRetryTask` in [packages/browser/src/client/tester/runner.ts#L80](../../packages/browser/src/client/tester/runner.ts#L80).
3. The node-side browser command resolves stack traces to source locations and records a single `internal:browserTrace` artifact in [packages/browser/src/node/commands/trace.ts#L62](../../packages/browser/src/node/commands/trace.ts#L62).
4. `recordArtifact` already supports artifact delivery before the task is finished in [packages/runner/src/artifact.ts#L41](../../packages/runner/src/artifact.ts#L41).
5. `TestRun.recordArtifact()` appends the artifact immediately and forwards it to reporters/clients in [packages/vitest/src/node/test-run.ts#L58](../../packages/vitest/src/node/test-run.ts#L58).
6. The API layer forwards `onTestArtifactRecord` to connected UI clients in [packages/vitest/src/api/setup.ts#L257](../../packages/vitest/src/api/setup.ts#L257).
7. The UI already stores incoming artifacts on the test node in [packages/ui/client/composables/client/index.ts#L39](../../packages/ui/client/composables/client/index.ts#L39) and [packages/ui/client/composables/explorer/collector.ts#L71](../../packages/ui/client/composables/explorer/collector.ts#L71).
8. The actual trace viewer already knows how to render any `BrowserTraceData` object step-by-step in [packages/ui/client/components/trace/TraceView.vue#L18](../../packages/ui/client/components/trace/TraceView.vue#L18).

The missing piece is that trace-view state currently only looks for the final `internal:browserTrace` artifact and only refreshes around run completion in [packages/ui/client/composables/trace-view.ts#L34](../../packages/ui/client/composables/trace-view.ts#L34).

## Important Constraints

- Artifacts are append-only in the current model. There is no built-in “replace previous artifact” or “mutate artifact payload in place” path.
- Because of that, trying to keep a single `internal:browserTrace` artifact “live” would fight the current task/artifact model.
- The viewer itself does not need event sourcing. Each step already contains a full snapshot, so partial replay can be built from an append-only list of entries.
- The browser-side collector is the only place that currently knows the exact snapshot timing. We should keep snapshot capture there.
- Source-map-backed `location` resolution is still best kept on the node side, as it is today.

### 1. Introduce a live chunk artifact

Add a new internal artifact type, tentatively:

- `internal:browserTraceChunk`

Suggested shape:

```ts
interface BrowserTraceChunkArtifact extends TestArtifactBase {
  type: 'internal:browserTraceChunk'
  data: {
    attemptId: string
    retry: number
    repeats: number
    recordCanvas: boolean
    seq: number
    entries: BrowserTraceEntry[]
    done?: boolean
  }
}
```

Notes:

- `attemptId` is required. `testId + retry + repeats` is not enough because the same tuple can reappear on later reruns in watch mode.
- `seq` makes ordering explicit and gives us a way to detect missing/out-of-order chunks.
- `done` is optional but useful if we want the last chunk to explicitly mark completion before the final artifact arrives.

### 2. Keep the final artifact unchanged

Do not replace `internal:browserTrace` as the canonical completed representation yet.

Reason:

- HTML reporter and static consumption already understand “one final full trace artifact”.
- The current viewer and artifact launcher logic are already built around that shape.
- This lets the live replay feature land without simultaneously redesigning trace persistence.

### 3. Change browser-side trace state from “only final buffer” to “full buffer + pending chunk queue”

Evolve the state around [packages/browser/src/client/tester/trace.ts#L64](../../packages/browser/src/client/tester/trace.ts#L64) from:

- one accumulated `entries` array per attempt

to:

- one full `entries` array for the final artifact
- one `pendingEntries` array for the next live chunk flush
- `attemptId`
- `seq`
- optional flush scheduling metadata

Conceptually:

```ts
interface BrowserTraceStreamState {
  attemptId: string
  retry: number
  repeats: number
  recordCanvas: boolean
  entries: BrowserTraceEntry[]
  pendingEntries: BrowserTraceEntry[]
  seq: number
}
```

### 4. Flush live chunks incrementally

When `recordBrowserTraceEntry()` records a step, it should:

- append the entry to `entries`
- append the entry to `pendingEntries`
- schedule a flush for `pendingEntries`

Recommended first batching policy:

- flush on the next microtask or animation frame
- also force flush on lifecycle boundaries such as `onAfterRetryTask`

Why not flush every single entry synchronously:

- each flush goes through RPC, stack parsing, artifact recording, and UI updates
- batching a few steps together should reduce churn without losing the “live” feel

### 5. Add a node-side chunk command parallel to the final command

Add a new browser command next to `_recordBrowserTrace` in [packages/browser/src/node/commands/trace.ts#L62](../../packages/browser/src/node/commands/trace.ts#L62), tentatively:

- `__vitest_recordBrowserTraceChunk`

Responsibilities:

- resolve `stack -> location` for chunk entries exactly like the final path does today
- record `internal:browserTraceChunk` immediately via `project.vitest._testRun.recordArtifact(...)`

This keeps live chunk transport inside the same browser-command boundary as the final trace artifact and reuses the same source-location policy.

### 6. Add UI-side ephemeral live trace state

Do not rely only on `test.artifacts` for the active trace pane state.

Instead, add a dedicated client-side map, roughly:

```ts
Map<testId, {
  attemptId: string
  retry: number
  repeats: number
  recordCanvas: boolean
  entries: BrowserTraceEntry[]
  done: boolean
}>
```

This state should be fed from `onTestArtifactRecord` events for `internal:browserTraceChunk`.

Why keep it separate instead of reading raw chunk artifacts back out of `test.artifacts`:

- the viewer wants one merged ordered trace, not a list of chunk artifacts
- we probably do not want chunk artifacts to show up in the general Artifacts panel as user-facing items
- ephemeral state is easier to clean up when a rerun starts or a final artifact arrives

### 7. Teach trace-view selection logic to prefer the live attempt

Evolve [packages/ui/client/composables/trace-view.ts#L10](../../packages/ui/client/composables/trace-view.ts#L10) from “active trace = selected test + final artifact” to:

- if a selected test has an active live attempt, open/use that
- otherwise fall back to the latest final `internal:browserTrace`

Suggested model:

```ts
type ActiveTraceSource =
  | { kind: 'live', test: RunnerTestCase, trace: BrowserTraceData, attemptId: string }
  | { kind: 'final', test: RunnerTestCase, trace: BrowserTraceArtifactWithData }
```

Behavior:

- auto-open when the first chunk for the selected test arrives
- auto-follow the newest step while the user has not manually pinned an older step
- if the user manually selects an older step, stop auto-follow until they jump back to “latest”
- when the final artifact for the same attempt arrives, swap `kind: 'live'` to `kind: 'final'`

### 8. Leave static/reporter behavior alone at first

For the first slice:

- Browser UI / Vitest UI consume live chunk artifacts
- HTML reporter ignores them
- Artifacts list hides them, or trace-specific UI filters them out

This keeps the change focused on watch-mode live replay rather than mixing it with trace persistence redesign.

### Why not mutate `internal:browserTrace` in place?

Because the current artifact model is append-only. Mid-run artifacts are easy; artifact replacement is not. Forcing “live trace = one mutating artifact” would add protocol complexity everywhere.

### Why not invent a custom websocket event first?

We already have an end-to-end, mid-run, per-test delivery path through artifacts. A new bespoke event channel is possible, but it expands scope across the API, client, and task store for questionable gain in the first iteration.

### Why not stream only the currently selected test?

That may eventually be the right optimization, but it requires a full subscription story:

- UI tells node which test/attempt it wants
- node tells the running browser worker/provider what to stream
- browser collector gates chunk emission accordingly

That is feasible, but it is a larger coordination problem than we need for the first prototype.

## Main Risk

The main risk is transport volume, not conceptual correctness.

Each live chunk contains full rrweb snapshots for one or more steps. For long tests or large DOMs, streaming every step for every traced test in watch mode could create noticeable UI and websocket pressure.

That is why I would ship this in two layers:

1. First prototype: unconditional chunk streaming with batching, only in interactive/watch flows.
2. If needed: add subscription-based streaming so only the actively viewed test emits live chunks.

### Slice 1: prove the UX with minimal architecture

- Add `attemptId` and `seq`.
- Add `internal:browserTraceChunk`.
- Batch-flush chunk artifacts during the attempt.
- Keep the existing final artifact unchanged.
- Add UI ephemeral live-trace state and prefer it for the selected test.

Expected result: the trace pane visibly advances while the test runs.

### Slice 2: harden the viewer behavior

- Add auto-follow vs pinned-step behavior.
- Hide chunk artifacts from generic artifact surfaces.
- Handle reruns, retries, repeats, and canceled runs cleanly.

### Slice 3: optimize if payload is too heavy

- Stream only while a live client is connected.
- Optionally stream only the selected/subscribed test.
- Consider moving final trace assembly from browser memory to node-side aggregation if we want to reduce browser memory pressure at the same time.

### Attempt identity

Use a real `attemptId`, not just `test.id/retry/repeats`. It should survive watch reruns where the same test is re-executed with the same retry/repeat counters.

Simple option:

- `attemptId = ${test.id}:${repeats}:${retry}:${attemptStartTime}`

### Completion handshake

Do we need `done` on the last chunk if the final artifact will arrive anyway?

My current answer: yes, it is still useful. It lets the UI stop treating the attempt as open even if the final artifact is delayed a little behind the last live flush.

### Run mode

I would keep live chunk streaming off for plain `vitest run` and HTML/static paths in the first iteration. The killer feature is the interactive watch-mode experience.

### Artifact visibility

Chunk artifacts should probably be considered transport details, not user-facing artifacts. The trace view can consume them, but the generic artifacts list should hide them.

## Testing Strategy

The most valuable test is an end-to-end UI test that proves the trace pane updates before the run finishes.

Likely shape:

- add a browser fixture with a deliberate pause between two traceable actions
- open the trace view during the paused run
- assert that the first step is already visible before the test finishes
- then assert that the final step appears after completion

The existing trace UI/component test coverage points to reuse are:

- [packages/ui/client/components/trace/TraceView.vue#L42](../../packages/ui/client/components/trace/TraceView.vue#L42)
- [packages/ui/client/composables/trace-view.ts#L34](../../packages/ui/client/composables/trace-view.ts#L34)

The existing artifact delivery path to reuse is:

- [packages/vitest/src/api/setup.ts#L257](../../packages/vitest/src/api/setup.ts#L257)
- [packages/ui/client/composables/client/index.ts#L39](../../packages/ui/client/composables/client/index.ts#L39)

## Recommendation

I would prototype this as “append-only live chunk artifacts plus unchanged final artifact”.

That is not the most theoretically elegant end state, but it is the best architecture for fast iteration:

- it reuses the existing mid-run artifact transport
- it preserves the existing final artifact contract
- it isolates new complexity mostly to trace-specific codepaths
- it gives us a straightforward escape hatch to subscription-based streaming later if traffic becomes the problem

</details>

Thavamani13:
- closes https://github.com/vitest-dev/vitest/issues/10197

## Summary

- Resolves the long-standing TODO in `trace.ts`: *"should we avoid accumulating? send and immediately clear each entry to save memory?"*
- Each browser trace entry is now streamed to the node server immediately via a new `streamBrowserTraceEntry` RPC call, instead of being batched in `browserTraceState` until the retry completes
- Particularly benefits watch mode where long test runs previously held all trace snapshots in browser memory until the end of each retry

## Changes

- **`src/types.ts`** — Move `BrowserTraceEntry` and related types here (away from `client/tester/trace.ts`) to avoid a circular import; add `streamBrowserTraceEntry` to `WebSocketBrowserHandlers`
- **`src/client/tester/trace.ts`** — Remove local accumulation (`getBrowserTraceState`, `getBrowserTrace`, `BrowserTraceState`); `recordBrowserTraceEntry` now calls `rpc().streamBrowserTraceEntry(...)` immediately
- **`src/client/utils.ts`** — Remove `browserTraceState` from `BrowserRunnerState`
- **`src/node/state.ts`** — Add `streamedTraceEntries: Map<string, StreamedTraceBuffer>` to buffer entries per `testId:repeats:retry` on the server
- **`src/node/rpc.ts`** — Add `streamBrowserTraceEntry` handler that pushes entries into the server-side buffer
- **`src/node/commands/trace.ts`** — `_recordBrowserTrace` reads from server buffer and finalizes; payload changes from `{testId, data}` → `{testId, retry, repeats}`
- **`src/client/tester/runner.ts`** — `onAfterRetryTask` streams lifecycle entry then sends `__vitest_recordBrowserTrace` as a finalize signal only

## How it works

WebSocket message ordering guarantees all `streamBrowserTraceEntry` messages arrive at the node before the `__vitest_recordBrowserTrace` finalize signal, so fire-and-forget streaming is safe. Client memory is freed immediately after each entry instead of accumulating until retry completion.

## Test plan

- [ ] Run existing trace tests: `cd test/browser && PROVIDER=playwright pnpm test:unit specs/trace.test.ts`
- [ ] Run playwright trace tests: `PROVIDER=playwright pnpm test:unit specs/playwright-trace.test.ts`


Thavamani13:
Hi, I'm a real person behind this contribution. I used Claude Code (AI-assisted) to help implement this feature, but I personally:       
                                                                                                                                           
  - Reviewed the problem and designed the approach (streaming trace entries from the browser client to the Node server instead of
  accumulating them client-side)                                                                                                           
  - Reviewed all generated code before committing
  - Stand behind the correctness and intent of the changes                                                                                 
                              
  AI tool used: Claude Code (Anthropic)

hi-ogawa (changes requested):
Please add e2e

Thavamani13:
> Please add e2e

image


> Please add e2e

@hi-ogawa  Done added e2e

hi-ogawa:
`test/browser` is not e2e. `test/ui` is. We should probably create `test/ui/fixtures-trace-live/slow.test.ts` and `test/ui/test/trace-live.spec.ts` where the fixture test case has slow test steps and `trace-live.spec.ts` should assert see trace view before test finishes.

Thavamani13:
> `test/browser` is not e2e. `test/ui` is. We should probably create `test/ui/fixtures-trace-live/slow.test.ts` and `test/ui/test/trace-live.spec.ts` where the fixture test case has slow test steps and `trace-live.spec.ts` should assert see trace view before test finishes.

image


Thanks for the clarification . I have added the test here .

PR #10212, opened 2026-04-27
fix!: represent locator as an object instead of a string

sheremet-va:
Currently `asLocator` is re-exported from the Node entry point which bundles `ivya` twice (in the browser and node). To fix that, we now include the `locator` string in the JSON message. This is a breaking change since commands now receive an object instead of a string.


packages/browser/src/client/tester/trace.ts
```diff
@@ -22,7 +23,7 @@ export interface BrowserTraceEntry {
   stack?: string
   // resolved server-side from stack in __vitest_recordBrowserTrace command
   location?: { file: string; line: number; column: number }
-  selector?: string
+  element?: SerializedLocator
```
hi-ogawa, line 26: This looks good. The change should be already verifiable by updating snapshot of `test/browser/specs/trace.test.ts`. UI side wiring is just `TraceView.vue` and just manual testing is enough.

PR #10225, opened 2026-04-28
fix: loosen vi.mock(import(...)) factory types to accept vi.fn()

### LINKED ISSUE #10224: vi.mock(import(...)) typing too strict - won't accept vi.fn()
## What / Why

I'm trying to use `vi.mock(import('./file.js'), () => {...})` for mocking. The main reason for this is so that the file being mocked is visible to TypeScript - namely if renaming or moving the module being mocked the mock reference is updated too. Too many times I re-org my code, and afterwards find during test run that a number of mocks are broken because the string literal points to an old reference.

## Explanation

I noticed that with `vi.mock(import('./file.js'), ...)`, the `vi.mock` function is overly strict with typing and won't allow `vi.fn()`.

Example:

```ts
// Works fine but the reference to the module is just a string literal
vi.mock('./logger.js', () => ({
    default: { info: vi.fn(), warn: vi.fn(), error: vi.fn() }
}));
```

```ts
// Fails typecheck because Mock<Procedure> is not assignable to Logger['info']
vi.mock(import('./logger.js'), () => ({
    default: { info: vi.fn(), warn: vi.fn(), error: vi.fn() }
}));
```

Could `vi.mock` be extended to accept `Mock` or `vi.fn()` for function properties?

I'm happy to have a go a providing a PR for this - just wanted to check it was direction the library wants to take prior.

Repository for recreating the issue:
https://github.com/robertbeal/vitest-mock-type-repro

robertbeal:
### Description

Be able to mock a module where the path is stricter than a string literal - meaning when renaming or moving modules, the mock path references are updated too (and not missed!)

```
// logger.js
class Logger {
  info(msg: string): void {
    console.log(msg);
  }
}

// The module path is just a string, invisible to refactoring tools
vi.mock("./logger.js", () => ({
  default: { info: vi.fn() },
}));

// Ideal, but... fails typecheck as Mock<Procedure> is not assignable to Logger['info']
vi.mock(import("./logger.js"), () => ({
  default: { info: vi.fn() },
}));
```

Resolves #10224 

### Tests
(I get 6 failures even from a clean checkout, but the same 6 failures after my change is applied)

### Documentation
(Not sure if this type change warrants a docs update)

### Changesets

### LLM Usage

I used OpenCode with Opus 4.6. Mainly when playing with few ways that `Mockable` can be written and how it reads - it's not the most readable code :/ the best I could get it.

### DIFF
--- a/packages/mocker/src/types.ts
+++ b/packages/mocker/src/types.ts
@@ -2,9 +2,15 @@
 
 type Awaitable<T> = T | PromiseLike<T>
 
+type Mockable<T> = T extends (...args: any[]) => any
+  ? T | ((...args: any[]) => any)
+  : T extends object
+    ? { [K in keyof T]?: Mockable<T[K]> }
+    : T
+
 export type ModuleMockFactoryWithHelper<M = unknown> = (
   importOriginal: <T extends M = M>() => Promise<T>,
-) => Awaitable<Partial<M>>
+) => Awaitable<Mockable<M>>
 export type ModuleMockFactory = () => any
 export interface ModuleMockOptions {
   spy?: boolean
--- a/test/unit/src/mockedClassInstance.ts
+++ b/test/unit/src/mockedClassInstance.ts
@@ -1,0 +1,7 @@
+export class Logger {
+  #config = {}
+  info(msg: string): void { void msg }
+  warn(msg: string): void { void msg }
+}
+
+export default new Logger()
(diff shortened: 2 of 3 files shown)


robertbeal:
Closing after discussion on the related issue - this isn't a solution that'll work.

The underlying vi.fn() isn't a problem, it's the wider type shape that causes an issue with this attempted solution. And while loosening the typing may arguably be the way forward - it doesn't feel or look right.

PR #10226, opened 2026-04-28
docs: update guide sidebar, add more recipes

sheremet-va:
Over time I am noticing it is hard for users to discover new features that we release. They usually get buried in blog or mentioned off-hand in the api/guide section.

This PR tries to change that by adding new recipes for the features that we added in the last several majors. The content was initially generated by claude, then re-prompted and reviewed by me

Additionally, I restructured the "Guides" sidebar because it became hard to distinguish between entries, it became a huge list of articles. Hopefully grouping them distinguishes them more


docs/guide/recipes/domain-locators.md
```diff
@@ -0,0 +1,116 @@
+---
+title: Domain Locators | Recipes
```
hi-ogawa: Can this be just "custom locators"?

---

Codex and my pick is "Extending browser locators"

docs/guide/recipes/cancellable.md
```diff
@@ -0,0 +1,56 @@
+---
+title: Cancellable Test Resources | Recipes
```
hi-ogawa, line 2: This title seems unclear. Got some candidate from Codex and probably my pick is "Cancelling Long-Running Operations Gracefully":

```
  - Cancelling Async Test Work Gracefully
  - Cancelling Long-Running Test Work Gracefully
  - Cancelling Async Test Operations Gracefully
  - Cancelling Long-Running Test Operations
  - Stopping Async Test Work Gracefully
```

hi-ogawa (commented):
lgtm

AriPerkkio (commented):
I think we should have a "landing page" on the Guides tab. It should describe what sections there are on the sidebar. It would be duplicate content, but it would be more verbose than the actual sidebar. Right now we are expecting users to check sidebar for all sections when they land on guides.

docs/guide/recipes/parallel-sequential.md
```diff
@@ -0,0 +1,89 @@
+---
+title: Parallel and Sequential Test Files | Recipes
+---
+
+# Parallel and Sequential Test Files
+
+Most test files are independent and run faster in parallel. The exception is the handful that share an exclusive resource, like a fixed port, a writable temp directory, or a database without per-test isolation. Those files flake when other tests run concurrently with them.
+
+Disabling parallelism globally would slow down every test in the suite. Splitting the suite into two [`projects`](/guide/projects), one parallel and one sequential, lets only the affected files pay the cost.
```
AriPerkkio, line 9: Would be nice to have an illustration picture here showing what parallel and sequential means.

hi-ogawa:
Astro's recipes look like this also with landing page https://docs.astro.build/en/recipes/

docs/guide/recipes/wait-for.md
```diff
@@ -0,0 +1,99 @@
...
+  const server = createServer()
+
+  await expect.poll(() => server.isReady, {
+    timeout: 500,
+    interval: 20
+  }).toBe(true)
+})
+```
+
+The failure message is the standard `expect` diff, with no manual `throw new Error('Server not started')` to maintain. This is the right tool for most "wait for X to become Y" cases.
```
AriPerkkio, line 29: Reading this made me realize I'm often not using this when I should. For example, doing async stuff before asserting anything can often lead to test timeouts which don't point to the specific line in the test. It's difficult to understand why test timed out. Maybe in `guide/common-errors` we should have a mention of this. 🤔

docs/guide/recipes/domain-locators.md
```diff
@@ -0,0 +1,116 @@
...
+---
+
+# Domain Locators
+
+Built-in [locators](/api/browser/locators) like `getByRole` and `getByText` cover queries that map onto accessibility attributes. They run out when an app has shapes that don't fit ARIA, like a "comment with N replies" or a row in a custom table component.
+
+The fallback is to use raw CSS. That works, but the result is a plain query rather than a locator, so you lose auto-retry and strict-mode protection.
```
AriPerkkio: CSS? Should it be query selector?

docs/guide/recipes/disable-isolation.md
```diff
@@ -0,0 +1,71 @@
...
+```sh
+vitest --shuffle --run --project='Unit tests'
+vitest --shuffle --run --project='Unit tests'
+```
+
+If the second run produces different results, you have order-dependent tests. Either fix the offender or leave isolation enabled for that file.
+
+## Per-pool isolation
+
+`isolate` only governs the [`threads`](/config/pool) and [`forks`](/config/pool) pools. The `vmThreads` and `vmForks` pools always run isolated regardless of the flag, since they trade startup cost for stronger guarantees. If you want maximum reuse, prefer `threads` or `forks`.
```
AriPerkkio: > If you want maximum reuse, prefer `threads` or `forks`

What does this mean?

docs/guide/recipes/parallel-sequential.md
```diff
@@ -0,0 +1,89 @@
...
+          fileParallelism: false,
+          sequence: { groupOrder: 1 },
+        },
+      },
+    ],
+  },
+})
+```
+
+The parallel batch finishes, *then* the sequential batch starts. Total wall clock stays close to the parallel time plus the longest sequential chain.
```
AriPerkkio: What does `longest sequential chain` mean? Maybe instead:


```suggestion
The parallel batch finishes, *then* the sequential batch starts. Total wall clock stays close to the parallel time plus sum of the sequential test run time.
```

AriPerkkio (commented):
Overall looks good! Recipes are great 💯

docs/guide/recipes/wait-for.md
sheremet-va: I don't think we should duplicate suggestions, otherwise they are hard to maintain

docs/guide/recipes/disable-isolation.md
sheremet-va: I don't think it makes sense, removed

docs/guide/recipes/parallel-sequential.md
sheremet-va: I'm gonna merge this and add a comment to create a follow up to have this image, otherwise we will never merge this 😭

PR #10230, opened 2026-04-29
docs: clarify order of `before*` hooks cleanup functions

AriPerkkio:
### Description

Callbacks returned from `before*` hooks execute always after `after*` hooks. They are not equivalent as documentation currently states.

https://stackblitz.com/~/edit/vitest-dev-vitest-xdy4r52x?file=example.test.ts&file=setupFile.ts&file=vite.config.ts&view=editor

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/guide/lifecycle.md
+++ b/docs/guide/lifecycle.md
@@ -137,10 +137,12 @@ The execution follows this order:
    - `beforeEach` hooks execute (in order defined, or based on [`sequence.hooks`](/config/sequence#sequence-hooks))
    - Test function executes
    - `afterEach` hooks execute (reverse order by default with `sequence.hooks: 'stack'`)
+   - Cleanup functions returned from `beforeEach` hooks execute (reverse order by default with `sequence.hooks: 'stack'`)
    - [`onTestFinished`](/api/hooks#ontestfinished) callbacks run (always in reverse order)
    - If test failed: [`onTestFailed`](/api/hooks#ontestfailed) callbacks run
    - Note: if `repeats` or `retry` are set, all of these steps are executed again
 6. **[`afterAll`](/api/hooks#afterall) hooks:** Run once after all tests in the suite complete
+7. **Cleanup functions returned from `beforeAll` hooks:** Run once after all tests in the suite complete
 
 **Example execution flow:**
 
--- a/docs/guide/lifecycle.md
+++ b/docs/guide/lifecycle.md
@@ -164,6 +164,11 @@ describe('User API', () => {
   beforeAll(() => {
     // Runs once before all tests in this suite
     console.log('beforeAll')
+
+    return function beforeAllCleanup() {
+      // Runs once afterAll hooks have run
+      console.log('beforeAllCleanup')
+    }
   })
 
   aroundEach(async (runTest) => {
@@ -176,6 +181,11 @@ describe('User API', () => {
   beforeEach(() => {
     // Runs before each test
     console.log('beforeEach')
+
+    return function beforeEachCleanup() {
+      // Runs after afterEach hooks have run
+      console.log('beforeEachCleanup')
+    }
   })
 
   test('creates user', () => {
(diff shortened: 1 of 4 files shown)


hi-ogawa:
What's the current behavior? I think this is meant to work according to the code here:

https://github.com/vitest-dev/vitest/blob/b01afd26cbd951fa22b8d652565473319fb28518/packages/runner/src/run.ts#L531-L556

---

Ah, just looked at the repro, but the doc isn't about `afterEach(() => {})` vs `beforeEach(() => () => {})` right?

AriPerkkio:
> Ah, just looked at the repro, but the doc isn't about `afterEach(() => {})` vs `beforeEach(() => () => {})` right?

Not sure if I'm following, but the mention about `onTestFinish`'s trade-off applies to callbacks returned from `beforeEach(callback: () => AfterEachFn)` too. `sequence.hooks` has no effect in that `AfterEachFn`. Is this rather a bug? I thought it's intentional since `onTestFinished` also isn't affected.

hi-ogawa:
`onTestFinished` tip is probably about this hard-coded `stack` (so it's not about the ordering of `onTestFinished` vs  `afterEach`)

https://github.com/vitest-dev/vitest/blob/a9d1a84f58c6892f768a08a524f938bbbf0f7000/packages/runner/src/run.ts#L691-L702

(And just noticed `onFailed` (from `onTestFailed`) is respecting `config.sequence.hooks`, which seems like a bug.)

---

Context
- https://github.com/vitest-dev/vitest/pull/5598
- https://github.com/vitest-dev/vitest/issues/5529

AriPerkkio:
@hi-ogawa I added example in https://github.com/vitest-dev/vitest/pull/10230/changes/309c756151921ebb6b4d44dd89a25d88af5cbe68. The `test/hooks-stack.test.ts` fails as `afterEach #3` run last, not first. 


I think this part shows why it's happening. Maybe this is a bug? 🤔 

https://github.com/vitest-dev/vitest/blob/bc73dbd568110050490e9e2ea2fecf95b6786024/packages/runner/src/run.ts#L676-L682

hi-ogawa:
> The `test/hooks-stack.test.ts` fails as `afterEach #3` run last, not first.

`afterEach #3` running last is true (and not sure it's by design or bugs), but what I'm pointing out is that, that's different aspect of the behavior than what `sequence.hooks: stack | list | parallel` controls. 

`sequence.hooks` is about how it runs multiple hooks of the same types. Not sure it's clarified enough in doc but that's how I understood.

AriPerkkio:
I see what you mean now - this `TIP` in docs about `onTestFinish` is about the fact that it's hardcoded to be `'stack'` for some reason. It's not about `onTestFinish` being equivalent to `afterEach`.

The callbacks returned by `beforeEach` are mentioned to be equivalent to `afterEach` here:

https://github.com/vitest-dev/vitest/blob/4a07c25db51ab1caf8df07bf04f319eadecd9180/docs/api/hooks.md?plain=1#L37

We'll need to change that, and mention these cleanups in:

https://github.com/vitest-dev/vitest/blob/4a07c25db51ab1caf8df07bf04f319eadecd9180/docs/guide/lifecycle.md?plain=1#L135-L142

Maybe these should be called "`beforeEach` cleanups"?

AriPerkkio:
Ok - this PR is now about documenting the execution order of `before*` hooks cleanup functions. I actually have real-world use case for hooks that execute after all other `afterEach` hooks, so this is perfect. 😄 

Thanks for discussing this thoroughly @hi-ogawa 💯

docs/guide/lifecycle.md
```diff
@@ -137,10 +137,12 @@ The execution follows this order:
    - `beforeEach` hooks execute (in order defined, or based on [`sequence.hooks`](/config/sequence#sequence-hooks))
    - Test function executes
    - `afterEach` hooks execute (reverse order by default with `sequence.hooks: 'stack'`)
+   - Cleanup functions returned from `beforeEach` hooks execute (reverse order by default with `sequence.hooks: 'stack'`)
```
hi-ogawa, line 140: Can we add beforeEach/All cleanups to `Example execution flow` below?

AriPerkkio: Added.

hi-ogawa (approved):
lgtm!

PR #10231, opened 2026-04-29
fix(browser): guard against toJSON in createSafeRpc proxy

### LINKED ISSUE #10214: birpc proxy in browser mode doesn't guard against toJSON, causing unhandled errors
### Describe the bug

When using Vitest browser mode, the birpc RPC proxy created in `@vitest/browser/dist/client.js` intercepts all property accesses and converts them to RPC calls. The proxy has a special case for `"then"` (to avoid Promise-like behavior), but it does **not** guard against `"toJSON"`.

When any code calls `JSON.stringify()` on an object that references `globalThis.__vitest_worker__` (or any object in the prototype chain that includes the birpc proxy), the `toJSON` trap fires, sending an RPC request to the server. Since `"toJSON"` is not a registered RPC function, this produces:

```
[birpc] function "toJSON" not found
```

This manifests as unhandled errors in the test output, even though the tests themselves pass.

### Reproduction

1. Create a vitest browser mode test
2. In the test, access any object that transitively references `globalThis.__vitest_worker__.rpc`
3. Call `JSON.stringify()` on that object (or let a framework's change detection do it)
4. Observe `[birpc] function "toJSON" not found` error

### Expected behavior

The birpc proxy should handle `"toJSON"` the same way it handles `"then"` — by returning `undefined` to indicate the object is not JSON-serializable via a custom `toJSON` method.

### Suggested fix

In the birpc proxy handler (or `createBirpc` / `createSafeRpc`), add `"toJSON"` to the list of trapped methods that return `undefined`:

```js
// Current
if (method === 'then') return undefined;

// Suggested
if (method === 'then' || method === 'toJSON') return undefined;
```

### Environment

- Vitest: 4.0.18
- @vitest/browser: 4.0.18
- Browser: Chromium (via Playwright)
- OS: macOS / Linux

DoriAlagem:
## Problem

When `JSON.stringify()` is called on an object that holds a reference to the `createSafeRpc` proxy (e.g. inside test utilities or framework integrations), JavaScript looks up the `toJSON` property on the proxy. The `get` trap intercepts this and forwards it to the birpc runtime as a remote procedure call. Since no `toJSON` function is registered on the remote, this produces an unhandled rejection:

```
[birpc] function "toJSON" not found
```

This is the same class of issue as the existing `'then'` guard, which prevents the proxy from being mistaken for a thenable.

## Fix

Extend the early-return guard to also cover `'toJSON'`:

```ts
if (p === 'then' || p === 'toJSON') {
  return
}
```

Returning `undefined` for `toJSON` tells `JSON.stringify` to use the default serialization path and avoids the spurious RPC call.

Fixes #10214

### DIFF
--- a/packages/browser/src/client/tester/rpc.ts
+++ b/packages/browser/src/client/tester/rpc.ts
@@ -37,7 +37,7 @@ export function createSafeRpc(
 ): VitestBrowserClient['rpc'] {
   return new Proxy(client.rpc, {
     get(target, p, handler) {
-      if (p === 'then') {
+      if (p === 'then' || p === 'toJSON') {
         return
       }
       const sendCall = get(target, p, handler)


DORI2001:
Accidentally closed — reopening. The fix is still valid.

Issue #9763, opened 2026-03-02
Optimize html reporter module graph data


hi-ogawa:
### Describe the bug

I noticed while doing https://github.com/vitest-dev/vitest/pull/9740. Currently the module graph is constructed and saved per _entry test file_. However, this involves quite a lot of duplicate information since this means each test file brings the subgraph starting from itself.

In https://github.com/vitest-dev/vitest/pull/9740, this is optimized to save the full graph structure per _environment_ and this avoids such duplicates. Html reporter should also preserve module graph data in that form, then only trim it down (traverse from per test file basis) at the view level.

### Reproduction

na

### System Info

```shell
na
```

### Used Package Manager

npm

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vitest.dev/guide/).
- [x] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #10235, opened 2026-04-30
feat(reporter): support html reporter single file output

### LINKED ISSUE #6425: Option to save report in a single combined HTML file

hi-ogawa:
### Description

- Closes https://github.com/vitest-dev/vitest/issues/6425
- Closes https://github.com/vitest-dev/vitest/pull/8256 (supersedes)

TODO
- [x] dogfood
  - three links generated from "Merge Reports" job https://github.com/vitest-dev/vitest/actions/runs/25790395858/job/75757151633?pr=10235
  - packages/ui: https://github.com/vitest-dev/vitest/actions/runs/25790395858/artifacts/6966733358
  - test/e2e: https://github.com/vitest-dev/vitest/actions/runs/25790395858/artifacts/6966733699
  - test/unit: https://github.com/vitest-dev/vitest/actions/runs/25790395858/artifacts/6966733525
- [x] singleFile + merge reports https://github.com/vitest-dev/vitest/pull/10031
- [x] test
- [x] docs
- [x] follow-up issue for embedding coverage https://github.com/vitest-dev/vitest/issues/10238

### Tests

### Documentation

### Changesets


hi-ogawa:
@codex review

sheremet-va (changes requested):
I asked this in the older PR, but what is the benefit of having BOTH options? Why do we need to support a single file and a directory? Shouldn't we just have one?

hi-ogawa:
> I asked this in the older PR, but what is the benefit of having BOTH options? Why do we need to support a single file and a directory? Shouldn't we just have one?

In my opinion, this should be opt-in. Inlining only known js/css assets would be fine, but there are other varying size of large assets:
- report metadata `./html.data.json.gz`: this scales by the size of project and tests and our `test/unit` is already ~3MB or so. We have an opportunity to squeeze down some (https://github.com/vitest-dev/vitest/issues/9763) but there's inherent limit. 
- file-based attachments: this includes screenshot png and playwright trace zip. The attachments are used as external assets and only needed lazily on UI, so embedding these in initial `index.html` is unnecessary slowdown to view the base report.
- `./coverage`: this can be reasonably out-of-scope but ideally the feature parity should be preserved with single file. This looks doable but mildly more complex and haven't addressed yet https://github.com/vitest-dev/vitest/issues/10238

The main factor is that having too large `index.html` with raw data embedded can make initial page loading unbearably slow compared to when the same data is fetched separately. I haven't tested the boundary but our `test/unit` report with 3MB of index.html feels stuck for a few seconds already. Codex told me it should be avoided if it exceeds 10MB or so.

sheremet-va:
These seem like huge limitations to me, but they are not addresses in the docs. If someone comes in with a “bug” that their inline html report doesn’t work because of the size, how do we fix that?

hi-ogawa:
The trade off seems reasonable to me. It covers many enough cases. Eventually GitHub might support direct multi file view https://github.com/actions/upload-artifact/issues/787 and that can reduce the need of this option.

This looks like only a matter of doc and see what will happen in practice.

sheremet-va:
> The trade off seems reasonable to me. It covers many enough cases. Eventually GitHub might support direct multi file view [actions/upload-artifact#787](https://github.com/actions/upload-artifact/issues/787) and that can reduce the need of this option.
> 
> This looks like only a matter of doc and see what will happen in practice.

Yes, I think we should clarify this in the docs so we can link it in issues that people open.

test/e2e/test/reporters/html.test.ts
```diff
@@ -1,104 +1,46 @@
...
-    const task = file.tasks[0]
-    task.id = 0
-    task.result.duration = 0
-    task.result.startTime = 0
-
-    expect(task.result.errors).not.toBeDefined()
-    expect(task.result.logs).not.toBeDefined()
-
-    expect(resultJson).toMatchSnapshot(`tests are passing`)
-    expect(indexHtml).toMatch('window.METADATA_PATH="html.meta.json.gz"')
```
hi-ogawa: Retiring these internal probing tests in favor of actual behavioral test in `test/ui`.

hi-ogawa:
@sheremet-va This is hanging for a while. I'd appreciate review and merge when you have time.

sheremet-va (approved):

PR #10249, opened 2026-05-02
fix(vitest): rerun only failed tests with the `f` watch shortcut

### LINKED ISSUE #10247: Watch mode "f" command does not rerun failed tests
### Describe the bug

1. Given some test suites with more than one test exists
2. When running tests in watch mode
3. Given test `A` in suite `α` has failed
4. When pressing the `f` shortcut
5. Expects only test `A` to be rerun, instead all tests within `α` are run

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-mqt9hpxp?file=test%2Fsuite.test.ts

### System Info

```shell
System:
    OS: macOS 26.3
    CPU: (10) arm64 Apple M5
    Memory: 151.45 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.14.1 - /Users/sebastianmalton/.nvm/versions/node/v24.14.1/bin/node
    npm: 11.11.0 - /Users/sebastianmalton/.nvm/versions/node/v24.14.1/bin/npm
    pnpm: 10.33.2 - /Users/sebastianmalton/.nvm/versions/node/v24.14.1/bin/pnpm
  Browsers:
    Firefox: 150.0
    Safari: 26.3
  npmPackages:
    @vitest/coverage-v8: 4.1.5 => 4.1.5
    vitest: ^4.1.5 => 4.1.5
```

### Used Package Manager

npm

### Validations

dokson:
### Description

- Closes https://github.com/vitest-dev/vitest/issues/10247

In watch mode, pressing the `f` shortcut (which calls `Vitest#rerunFailed`) reran every test inside files that had at least one failure, instead of rerunning only the failed tests.

This change makes `rerunFailed` build a `TestSpecification` per failed file with `testIds` populated from the tasks whose `result.state === 'fail'`. The runner already supports filtering by `testIds` in `interpretTaskModes`, so non-failed tests in the same file are now skipped — the same pattern already used by `TestCase#rerun` and `TestSuite#rerun` in `reported-tasks.ts`.

If a file failed during collection (no test-level failures available), the original whole-file rerun behaviour is preserved as a fallback.

Task IDs are deterministic (hash of relative path + project name + position), so they remain stable across reruns of unchanged files.

TODO
- [x] fix `rerunFailed` to filter by failed test ids
- [x] preserve fallback to whole-file rerun on collection failure
- [x] test (`test/e2e/test/rerun-failed.test.ts`) — verified locally that it passes with the fix and fails without it (`expected 'run' to be 'skip'`)
- [x] docs — no public API change; aligns `rerunFailed` with the documented "rerun only failed tests" shortcut behaviour

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/core.ts
+++ b/packages/vitest/src/node/core.ts
@@ -1291,30 +1291,51 @@ export class Vitest {
 
   /** @internal */
   async rerunFailed(): Promise<void> {
-    const failedFiles = this.state.getFailedFilepaths()
+    let failedFiles = this.state.getFailedFilepaths()
     if (!failedFiles.length) {
       await this.rerunFiles(failedFiles, 'rerun failed', false)
       return
     }
 
+    if (this.filenamePattern) {
+      const filteredFiles = await this.globTestSpecifications(this.filenamePattern)
+      const allowed = new Set(filteredFiles.map(f => f.moduleId))
+      failedFiles = failedFiles.filter(file => allowed.has(file))
+      if (!failedFiles.length) {
+        await this.rerunFiles(failedFiles, 'rerun failed', false)
+        return
+      }
+    }
+
     const trigger = 'rerun failed'
-    const specifications = failedFiles.flatMap((filepath) => {
-      const fileTasks = this.state.getFiles([filepath])
-      const failedTestIds = fileTasks.flatMap(file =>
-        getTasks(file)
+    const specifications: TestSpecification[] = []
+    for (const filepath of failedFiles) {
+      const projectSpecs = this.getModuleSpecifications(filepath)
+      for (const file of this.state.getFiles([filepath])) {
+        if (file.result?.state !== 'fail') {
+          continue
+        }
+        const projectName = file.projectName || ''
+        const spec = projectSpecs.find(s => s.project.name === projectName)
+        if (!spec) {
+          continue
+        }
+        const failedTestIds = getTasks(file)
           .filter(task => task.type === 'test' && task.result?.state === 'fail')
-          .map(task => task.id),
-      )
-      return this.getModuleSpecifications(filepath).map((spec) => {
+          .map(task => task.id)
         // If the file failed to collect, there are no test ids: rerun the whole file.
         if (!failedTestIds.length) {
-          return spec
+          specifications.push(spec)
         }
-        return new TestSpecification(spec.project, spec.moduleId, spec.pool, {
-          testIds: failedTestIds,
-        })
-      })
-    })
+        else {
+          specifications.push(
+            new TestSpecification(spec.project, spec.moduleId, spec.pool, {
+              testIds: failedTestIds,
+            }),
+          )
+        }
+      }
+    }
 
     await Promise.all([
       this.report('onWatcherRerun', failedFiles, trigger),
--- a/packages/vitest/src/node/core.ts
+++ b/packages/vitest/src/node/core.ts
@@ -11,7 +11,6 @@ import type { VitestFetchFunction } from './environments/fetchModule'
 import type { ProcessPool } from './pool'
 import type { Report } from './reporters/report'
 import type { TestModule } from './reporters/reported-tasks'
-import type { TestSpecification } from './test-specification'
 import type { ResolvedConfig, TestProjectConfiguration, UserConfig, VitestRunMode } from './types/config'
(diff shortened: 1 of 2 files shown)


dokson:
Confirming ownership: I authored and reviewed this change myself. The marker is included per AGENTS.md guidelines for AI-assisted contributions, but the implementation and reasoning have been verified by me.

truffle-dev:
One regression worth flagging: this version of `rerunFailed` no longer respects the active `filenamePattern`.

The existing `rerunFiles` filters failed files against the active filename pattern (`packages/vitest/src/node/core.ts:1204-1207` on `main`):

```ts
if (this.filenamePattern) {
  const filteredFiles = await this.globTestSpecifications(this.filenamePattern)
  files = files.filter(file => filteredFiles.some(f => f.moduleId === file))
}
```

So if a user pressed `p` to scope to a subset and then pressed `f`, the rerun stayed within that subset. By bypassing `rerunFiles`, the new code path skips that filter and reruns every failed file regardless of the active filename pattern.

Re-applying the same filter inline before iterating `failedFiles` (a small `globTestSpecifications(this.filenamePattern)` + `Set.has` intersection) preserves the existing watch UX.

There's also a smaller workspace-mode subtlety: in a multi-project workspace where the same `filepath` belongs to several projects, `state.getFiles([filepath])` returns one `File` per project, and the flatMap unions failed test ids across them. Test ids include the project name in their hash (`generateFileHash` in `test-specification.ts`), so the project-A spec ends up filtered against the union of A+B ids — which still works for A's tests but means project B's spec is constructed with ids it can never match (everything skipped). Iterating per failed `File` and looking up the project via `getProjectByName(file.projectName || '')` keeps the test ids scoped to the right project.

dokson:
Thanks @truffle-dev — both points fixed in 300c9612b.

1. **filenamePattern**: now filtered before iterating `failedFiles` (`globTestSpecifications(this.filenamePattern)` + `Set.has`), matching `rerunFiles` behaviour. Added an e2e test that asserts pressing `f` after `p` stays scoped.
2. **Workspace scoping**: now iterating per `File`, looking up the spec via `getProjectByName(file.projectName || '')` so test ids stay scoped to the right project. Also skip Files whose `result.state !== 'fail'` to avoid rerunning a passing project's whole file when it shares a path with a failing one. Added an e2e test with two projects sharing `shared.test.ts` (one failing, one passing) that verifies only the failing project is rerun.

truffle-dev:
Both fixes look right.

1. The `globTestSpecifications` + `Set.has` intersection mirrors the gate at `core.ts:1204-1207`, and the early-return when the intersection is empty matches the existing no-failed-files behaviour. The `pressing-f-after-p` e2e test is the right shape to lock that in.
2. Iterating per-`File` and resolving the spec via `projectName` resolves the workspace ambiguity cleanly. The `state !== 'fail'` skip inside the loop also drops the file/project pairs that incidentally pass when the same path runs in multiple projects, which is what the second e2e test is asserting.

CI is green across all build/test matrix jobs (the `Browsers: macos-latest` failure looks unrelated to this change).

dokson:
Disclosure per the AI contribution policy:

- I personally reviewed and stand behind the content of this PR. Both review rounds with @truffle-dev (filenamePattern regression and workspace-mode scoping) were addressed with verified fixes and accompanying e2e tests.
- It follows the contribution guidelines and the GitHub PR template, references the related issue, and includes tests that fail without the fix and pass with it.
- AI tool used: Claude (Claude Code). It assisted with code drafting and test scaffolding; the design decisions and the verification work are mine.

truffle-dev:
Both review points addressed cleanly in 300c9612.

1. `filenamePattern` scope (`core.ts:1300-1308`): `globTestSpecifications(this.filenamePattern)` builds the allowed-moduleId set and `failedFiles` is filtered, matching `rerunFiles`. New e2e `rerunFailed respects active filenamePattern` covers the `f` after `p` regression.

2. Project scoping (`core.ts:1311-1336`): per-File loop resolves the spec via `getModuleSpecifications(filepath).find(s => s.project.name === projectName)`, and the `result?.state !== 'fail'` skip prevents re-running a passing sibling File. New e2e `rerunFailed scopes test ids per project in workspace mode` covers the cross-project leak with `shared.test.ts` failing in `p1` and passing in `p2`.

CI is green across the full Build&Test matrix.

PR #10253, opened 2026-05-03
feat(node): export capturePrintError

samchungy:
### Description

I'm trying to implement a custom reporter which uses GitHub check runs annotations based on the [github actions reporter ](https://github.com/vitest-dev/vitest/blob/main/packages/vitest/src/node/reporters/github-actions.ts) but I notice it has access to things which aren't publicly accessible.

The `task` type in `TestModule` is accessed by every other reporter so I'm not entirely sure why it's marked as `internal` as it seems strange that I shouldn't access it if I wanted to implement a custom reporter.

The `capturePrintError` is used by the junit and actions reporters so it would be useful for me to be able to use it too.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/public/node.ts
+++ b/packages/vitest/src/public/node.ts
@@ -40,6 +40,7 @@ export { ThreadsPoolWorker } from '../node/pools/workers/threadsWorker'
 export { TypecheckPoolWorker } from '../node/pools/workers/typecheckWorker'
 export { VmForksPoolWorker } from '../node/pools/workers/vmForksWorker'
 export { VmThreadsPoolWorker } from '../node/pools/workers/vmThreadsWorker'
+export { capturePrintError } from '../node/printError'
 export type { SerializedTestProject, TestProject } from '../node/project'
 
 export {
--- a/test/unit/test/exports.test.ts
+++ b/test/unit/test/exports.test.ts
@@ -107,6 +107,7 @@ it('exports snapshot', async ({ skip, task }) => {
         "VitestPlugin": "function",
         "VmForksPoolWorker": "function",
         "VmThreadsPoolWorker": "function",
+        "capturePrintError": "function",
         "createDebugger": "function",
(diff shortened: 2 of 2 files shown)


hi-ogawa:
> The `task` type in `TestModule` is accessed by every other reporter so I'm not entirely sure why it's marked as `internal` as it seems strange that I shouldn't access it if I wanted to implement a custom reporter.

I think `RunnerTestFile` being internal is an intended architecture. `TestModule` is newer user-facing reporter API and `RunnerXxx` is old (and still raw internal) representation. Builtin reporter should technically operate on newer API too, but they aren't just migrated yet. 

And that also means `capturePrintError(..., { task })` shouldn't be a public API as is though the request sounds reasonable. We probably need to refactor something there, so exploring that aspect would be more helpful.

samchungy:
> > The `task` type in `TestModule` is accessed by every other reporter so I'm not entirely sure why it's marked as `internal` as it seems strange that I shouldn't access it if I wanted to implement a custom reporter.
> 
> I think `RunnerTestFile` being internal is an intended architecture. `TestModule` is newer user-facing reporter API and `RunnerXxx` is old (and still raw internal) representation. Builtin reporter should technically operate on newer API too, but they aren't just migrated yet.
> 
> And that also means `capturePrintError(..., { task })` shouldn't be a public API as is though the request sounds reasonable. We probably need to refactor something there, so exploring that aspect would be more helpful.

Oooo I see, would you be okay if I update the PR for just exporting `capturePrintError`?

hi-ogawa:
> Oooo I see, would you be okay if I update the PR for just exporting `capturePrintError`?

We don't want to leak `{ task }` as options, so need to refactor it first. It appears that's already redundant so testing removing it https://github.com/vitest-dev/vitest/pull/10259.

hi-ogawa:
Would this shape work for you?
- https://github.com/vitest-dev/vitest/pull/10268

PR #10265, opened 2026-05-04
fix(runner): remove AbortSignal listener leak in withCancel

tomohiro86:
## Summary

`withCancel` adds an `'abort'` event listener to the provided `AbortSignal` but never removes it when the wrapped function completes normally (resolve or reject). The listener stays attached until the signal itself is garbage-collected, holding a closure reference to `reject` for longer than necessary.

Since each test gets a fresh `AbortController`, listeners don't accumulate across tests and `MaxListenersExceededWarning` is not a real risk here — the previous description overstated that.

The actual fix is straightforward hygiene: call `removeEventListener` on every exit path so the listener is released as soon as the promise settles.

## Changes

- Extract the abort handler into a named function `onAbort` so it can be removed.
- Add `{ once: true }` to `addEventListener` as a safety net for the abort path.
- Call `removeEventListener` via a `cleanup` helper on every exit path (resolve, reject, and synchronous throw).

```ts
// before
signal.addEventListener('abort', () => reject(signal.reason))

// after
const onAbort = () => reject(signal.reason)
signal.addEventListener('abort', onAbort, { once: true })
const cleanup = () => signal.removeEventListener('abort', onAbort)
// cleanup() called on every resolve/reject/throw path
```


AriPerkkio (changes requested):
This sounds like 100% made up problem. Could you demonstrate a test case that triggers `MaxListenersExceededWarning` from this code path?

tomohiro86:
You're right — `MaxListenersExceededWarning` isn't actually reachable here since each test gets a fresh `AbortController`, so listeners never accumulate on any single signal. I've updated the PR description to drop that claim and reframe the change as straightforward cleanup hygiene: the listener is held longer than necessary when the promise settles normally. Sorry for the overstatement.

AriPerkkio (approved):
I guess this can show up in non-isolated test runs when running thousands of test cases. 🤷‍♂️ 
Too bad `AbortSignal` doesn't have methods to clear all listeners at once.

AriPerkkio (approved):
Thanks!

PR #10267, opened 2026-05-04
fix(browser): remove orphaned Playwright route when same module is mocked via multiple ids (fix #9957)

### LINKED ISSUE #9957: Browser Mode leaks a manual mock across spec files when the same module is mocked via two ids

Zelys-DFKH:
Fixes #9957.

### Description

If you call `vi.mock('~/modal')` and `vi.mock('./modal')` in the same test file, and both aliases resolve to the same URL, `createMocker()` in `playwright.ts` registers two Playwright routes. The second call overwrites the predicate in `idPredicates` without removing the first route. When `clear()` runs at the end of the file, it only finds the second predicate -- the first route stays attached to the browser context.

That orphaned route fires when the next test file loads the same module. Its handler calls `module.resolve()` through a birpc channel that belongs to the previous session, which is already closed. That's where `[birpc] rpc is closed` comes from.

**`playwright.ts`**: before registering a route for a session+URL, check `idPredicates` for an existing predicate and `unroute()` it first. Switched `sessionIds` from `string[]` to `Set<string>` so URL entries can't duplicate and `clear()` removes each route exactly once.

### Testing

Added `test/browser/fixtures/mocking/src/dual-id/` with a source module, a consumer that imports it via the alias, and two test files. The probe registers the module through both `~/dual-id/modal` and `./dual-id/modal`. The target (same browser session, runs after) asserts the real implementation is visible. Without this fix, the target crashes with `[birpc] rpc is closed`; with it, the mock is cleaned up between files.

```
PROVIDER=playwright BROWSER=chromium corepack pnpm -C test/browser exec vitest run --config vitest.config.unit.mts specs/mocking.test.ts -t 'manual mocks do not leak'
```

### DIFF
--- a/packages/browser/src/client/tester/runner.ts
+++ b/packages/browser/src/client/tester/runner.ts
@@ -273,7 +273,11 @@ export function createBrowserRunner(
         }
         catch {}
       }
-      return rpc().onCollected(this.method, files)
+      const result = await rpc().onCollected(this.method, files)
+      if (this.method === 'collect') {
+        await mocker.invalidate()
+      }
+      return result
     }
 
     onTestAnnotate = (test: Test, annotation: TestAnnotation): Promise<TestAnnotation> => {
--- a/packages/browser/src/client/tester/runner.ts
+++ b/packages/browser/src/client/tester/runner.ts
@@ -273,11 +273,7 @@ export function createBrowserRunner(
         }
         catch {}
       }
-      const result = await rpc().onCollected(this.method, files)
-      if (this.method === 'collect') {
-        await mocker.invalidate()
-      }
-      return result
+      return rpc().onCollected(this.method, files)
     }
 
     onTestAnnotate = (test: Test, annotation: TestAnnotation): Promise<TestAnnotation> => {
--- a/packages/browser-playwright/src/playwright.ts
+++ b/packages/browser-playwright/src/playwright.ts
@@ -261,9 +261,9 @@ export class PlaywrightBrowserProvider implements BrowserProvider {
 
   private createMocker(): BrowserModuleMocker {
     const idPredicates = new Map<string, (url: URL) => boolean>()
-    const sessionIds = new Map<string, string[]>()
+    const sessionIds = new Map<string, Set<string>>()
 
-    function createPredicate(sessionId: string, url: string) {
+    function createPredicate(url: string) {
       const moduleUrl = new URL(url, 'http://localhost')
       const predicate = (url: URL) => {
         if (url.searchParams.has('_vitest_original')) {
@@ -293,11 +293,7 @@ export class PlaywrightBrowserProvider implements BrowserProvider {
 
         return true
       }
-      const ids = sessionIds.get(sessionId) || []
-      ids.push(moduleUrl.href)
-      sessionIds.set(sessionId, ids)
-      idPredicates.set(predicateKey(sessionId, moduleUrl.href), predicate)
-      return predicate
+      return { url: moduleUrl.href, predicate }
     }
 
     function predicateKey(sessionId: string, url: string) {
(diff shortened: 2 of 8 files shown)


packages/browser/src/client/tester/runner.ts
```diff
@@ -273,7 +273,11 @@ export function createBrowserRunner(
         }
         catch {}
       }
-      return rpc().onCollected(this.method, files)
+      const result = await rpc().onCollected(this.method, files)
+      if (this.method === 'collect') {
```
sheremet-va: I don't think there are any tests that cover test collection with mocking (this is `collect` only in `vitest list` command)

sheremet-va (changes requested):

Zelys-DFKH:
Good catch. Removed the `runner.ts` change. It was addressing a separate scenario (routes not invalidated after `vitest list`) with no test coverage, and it's out of scope for #9957. The core fix lives entirely in `playwright.ts`. I can file a separate issue for the collect path if that'd be useful.

sheremet-va (approved):

sheremet-va:
Thanks!

Zelys-DFKH:
And you! 🙏

PR #10272, opened 2026-05-05
fix(cli): respect `FORCE_COLOR` over agent detection

dokson:
### Description

Vitest's CLI calls `disableDefaultColors()` whenever `std-env`'s `isAgent` returns true, with no way for the user to opt back in. This strips ANSI from all reporter output even when the user has explicitly requested colours via `FORCE_COLOR=1` — every other CLI that follows the `FORCE_COLOR` / `NO_COLOR` convention treats those as opt-in overrides, so vitest should too.

### Fix

- `packages/vitest/src/node/cli/cac.ts`: skip the auto-disable when `FORCE_COLOR` is set. This matches the convention used by chalk, picocolors, supports-color, etc.
- `test/config/test/console-color.test.ts`: add a regression case `agent with FORCE_COLOR=1` next to the existing `agent` test. Asserts that the ` RUN ` banner is rendered with ANSI when `AI_AGENT` and `FORCE_COLOR` are both set, fails on `main` (no ANSI emitted) and passes with this fix.

### Test plan

- [x] New `agent with FORCE_COLOR=1` fails on `main`, passes with this fix.
- [x] Existing `agent` still passes (no-colour assertion is preserved when `FORCE_COLOR` is unset).

### Tests

### Documentation
- [x] No public API change. The new behaviour follows the standard `FORCE_COLOR` / `NO_COLOR` convention, so no docs change is needed.

### Changesets
- [x] Changes in changelog are generated from PR name. Please, make sure that it explains your changes in an understandable manner. Please, prefix changeset messages with \`feat:\`, \`fix:\`, \`perf:\`, \`docs:\`, or \`chore:\`.

### DIFF
--- a/packages/vitest/src/node/cli/cac.ts
+++ b/packages/vitest/src/node/cli/cac.ts
@@ -7,7 +7,7 @@ import cac from 'cac'
 import { normalize } from 'pathe'
 import c, { disableDefaultColors } from 'tinyrainbow'
 import { version } from '../../../package.json' with { type: 'json' }
-import { isAgent } from '../../utils/env'
+import { isAgent, isForceColor } from '../../utils/env'
 import { benchCliOptionsConfig, cliOptionsConfig, collectCliOptionsConfig } from './cli-config'
 import { setupTabCompletions } from './completions'
 
@@ -75,7 +75,7 @@ function addCliOptions(cli: CAC | Command, options: CLIOptionsConfig<any>) {
 }
 
 export function createCLI(options: CliParseOptions = {}): CAC {
-  if (isAgent) {
+  if (isAgent && !isForceColor()) {
     disableDefaultColors()
   }
 
--- a/test/config/test/console-color.test.ts
+++ b/test/config/test/console-color.test.ts
@@ -44,6 +44,17 @@ test('agent', async () => {
   expect.soft(stdout).not.toContain('\x1B[46m RUN')
 })
 
+test('agent with FORCE_COLOR=1', async () => {
+  // FORCE_COLOR opts back into colours even when isAgent is detected, both in the CLI process and in workers
+  const { stdout } = await runVitestCli({
+    preserveAnsi: true,
+    nodeOptions: { env: { AI_AGENT: 'copilot', FORCE_COLOR: '1' } },
+  }, '--root', 'fixtures/console-color', '--reporter', 'default')
+
+  expect.soft(stdout).toContain('\x1B[33mtrue\x1B[39m\n')
+  expect.soft(stdout).toContain('\x1B[46m RUN')
+})
+
 test.skipIf(process.platform === 'win32')('without color, forks pool in non-TTY parent', async () => {
   const { stdout } = await runVitest({
     root: 'fixtures/console-color',
--- a/packages/vitest/src/node/config/serializeConfig.ts
+++ b/packages/vitest/src/node/config/serializeConfig.ts
@@ -3,7 +3,7 @@ import type { TestProject } from '../project'
 import type { ApiConfig } from '../types/config'
 import { resolve } from 'node:path'
 import { configDefaults } from '../../defaults'
-import { isAgent } from '../../utils/env'
+import { isAgent, isForceColor } from '../../utils/env'
 
 export function serializeConfig(project: TestProject): SerializedConfig {
   const { config, globalConfig } = project
@@ -155,6 +155,6 @@ export function serializeConfig(project: TestProject): SerializedConfig {
       config.slowTestThreshold
       ?? globalConfig.slowTestThreshold
       ?? configDefaults.slowTestThreshold,
-    isAgent,
+    disableColors: isAgent && !isForceColor(),
   }
 }
--- a/packages/vitest/src/runtime/config.ts
+++ b/packages/vitest/src/runtime/config.ts
@@ -156,7 +156,7 @@ export interface SerializedConfig {
   strictTags: boolean
   mergeReportsLabel: string | undefined
   slowTestThreshold: number | undefined
-  isAgent: boolean
+  disableColors: boolean
 }
 
 export interface SerializedCoverageConfig {
--- a/packages/vitest/src/runtime/workers/init.ts
+++ b/packages/vitest/src/runtime/workers/init.ts
@@ -50,7 +50,7 @@ export function init(worker: Options): void {
         process.env.VITEST_WORKER_ID = String(message.workerId)
         reportMemory = message.options.reportMemory
 
-        if (message.context.config.isAgent) {
+        if (message.context.config.disableColors) {
           disableDefaultColors()
         }
 
--- a/packages/vitest/src/utils/env.ts
+++ b/packages/vitest/src/utils/env.ts
@@ -12,4 +12,5 @@ export const isDeno: boolean
 export const isWindows: boolean = (isNode || isDeno) && process.platform === 'win32'
 export const isBrowser: boolean = typeof window !== 'undefined'
 export const isTTY: boolean = ((isNode || isDeno) && process.stdout?.isTTY && !isCI)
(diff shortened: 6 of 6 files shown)


dokson:
Disclosure per the AI contribution policy:

- I personally reviewed and stand behind the content of this PR.
- It follows the contribution guidelines and the GitHub PR template.
- AI tool used: Claude (Claude Code).

packages/vitest/src/node/cli/cac.ts
```diff
@@ -75,7 +75,7 @@ function addCliOptions(cli: CAC | Command, options: CLIOptionsConfig<any>) {
 }
 
 export function createCLI(options: CliParseOptions = {}): CAC {
-  if (isAgent) {
+  if (isAgent && !process.env.FORCE_COLOR && !process.argv.includes('--color')) {
```
AriPerkkio: Respecting `FORCE_COLOR` makes sense. But I think your AI is hallucinating `--color`. What's that?

dokson: Yep, agreed: working on it

dokson: @AriPerkkio you reviewed my only draft PR, lol....thanks btw: I appreciate it 🥲

test/e2e/test/reporters/default.test.ts
```diff
@@ -363,9 +363,10 @@ describe('default reporter', async () => {
 
   test('project name color', async () => {
     const { stdout } = await runVitestCli(
-      { preserveAnsi: true },
+      { preserveAnsi: true, nodeOptions: { env: { FORCE_COLOR: '1' } } },
       '--root',
       'fixtures/reporters/project-name',
+      '--reporter=default',
     )
```
AriPerkkio, line 369: This test should not be changed. Better place would be in

https://github.com/vitest-dev/vitest/blob/d10909848087adeb62a6c1e09140c35fce73d6ca/test/config/test/console-color.test.ts#L1-L79

dokson: makes sense...

dokson: Done in 888b374fb — moved the regression test to `test/config/test/console-color.test.ts` next to the existing `agent` case, and reverted `default.test.ts`.

AriPerkkio (approved):

hi-ogawa:
What about the runtime side? https://github.com/vitest-dev/vitest/blob/95905a872f4432406a8da910fe046159133578db/packages/vitest/src/runtime/workers/init.ts#L53-L55

(Also https://github.com/vitest-dev/vitest/issues/10046 is messy because we use tinyrainbow in many places)

hi-ogawa (changes requested):
see comment above

test/config/test/console-color.test.ts
```diff
@@ -44,6 +44,16 @@ test('agent', async () => {
...
 })
 
+test('agent with FORCE_COLOR=1', async () => {
+  // FORCE_COLOR opts back into colours even when isAgent is detected
+  const { stdout } = await runVitestCli({
+    preserveAnsi: true,
+    nodeOptions: { env: { AI_AGENT: 'copilot', FORCE_COLOR: '1' } },
+  }, '--root', 'fixtures/console-color', '--reporter', 'default')
+
+  expect.soft(stdout).toContain('\x1B[46m RUN')
```
AriPerkkio, line 55: Assert also the `console.log` that the test case does, to make sure https://github.com/vitest-dev/vitest/pull/10272#issuecomment-4384217746 is met.

AriPerkkio (changes requested):

test/config/test/console-color.test.ts
dokson: Done in a918be444 — the test now asserts both the worker-side `\x1B[33mtrue\x1B[39m` (from `console.log(true)` in the fixture) and the CLI-side `\x1B[46m RUN` banner. Also extracted the `'FORCE_COLOR' in process.env` check into `isForceColor` in `utils/env.ts` so `cac.ts` and `workers/init.ts` share the predicate.

dokson:
@hi-ogawa  #10046 is broader: every place vitest pulls in tinyrainbow ends up sharing the mutated singleton, so any user code that imports tinyrainbow at test time inherits the no-colour state. This PR doesn't fix that root cause (the singleton mutation), it just gives users an explicit override. 

Happy to follow up separately on the singleton issue if it makes sense. 🥲

AriPerkkio (dismissed):

hi-ogawa (dismissed):

sheremet-va (changes requested):
How does the detection in browser mode works? The whole reason why isAgent exists on the config is to not use the utils/env because it’s tied to server environments

dokson:
Browser mode isn't affected: `disableDefaultColors()` is only called in `cac.ts` (Node CLI) and `runtime/workers/init.ts` (Node worker pools). Browser-side tinyrainbow uses its own auto-detect in the browser env, so there's no agent-disable path there.

dokson:
@sheremet-va moved the `FORCE_COLOR` check into the server-side computation in `serializeConfig.ts` so `config.isAgent` is the single source of truth for the runtime. `utils/env` is no longer imported by runtime code; `cac.ts` keeps the inline `isForceColor()` check since it runs server-side. `config.isAgent` is now "running under an agent **and** the user hasn't opted into colours via `FORCE_COLOR`", which is exactly the predicate `workers/init.ts` (its only consumer) needs. Browser mode is unaffected: `disableDefaultColors()` is only called from those two Node-side paths, never browser-side.

sheremet-va:
Maybe we should rename it to disableColors or something then

dokson:
@sheremet-va adding a `disableColors` field next to `isAgent` would also work, but `SerializedRootConfig` is public surface (reporters/plugins read it via `ctx.serializedRootConfig`), so a new field is a permanent API addition. Doing it now means committing to it for a single internal consumer (`workers/init.ts`) which feels premature unless we expect a second reader of the strict `isAgent` semantics in the near future.

sheremet-va:
I am not suggesting adding it next to isAgent, but replacing it, so your analysis is meaningless

dokson:
@sheremet-va just to gently flag, your tone has come across as consistently aggressive towards me across our interactions. To clarify: I hadn't proposed the rename earlier because renaming `isAgent` to `disableColors` is technically a breaking change for any external consumer reading `config.isAgent` via `ctx.serializedRootConfig`. The field has been in stable releases since v4.1.2 (added in #9851), so it's not just beta surface, just worth flagging for the changelog.

sheremet-va:
> To clarify: I hadn't proposed the rename earlier because renaming `isAgent` to `disableColors` is technically a breaking change for any external consumer reading `config.isAgent` via `ctx.serializedRootConfig`. The field has been in stable releases since v4.1.2 (added in #9851), so it's not just beta surface, just worth flagging for the changelog.

Flag it in the changelog then. I don't understand the issue here. Your PR changes the meaning of `isAgent`, keeping the name is objectively wrong. If you want to avoid breaking changes, add it alongside the `isAgent` flag. I don't see how "it's only used internally" could be a reason to reject this idea. `serializedRootConfig` is also only used internally, it's not even documented. These APIs are all so insignificant that they doesn't even deserve their own discussion.

dokson:
The field was added in #9851 to propagate the `if (isAgent) disableDefaultColors()` decision from `cac.ts` to the worker, and `workers/init.ts` has been its only consumer ever since. The name `isAgent` reflected the source of the boolean (std-env), not its purpose (the colour decision); renaming to `disableColors` aligns the field with what it has always been used for. I'll push the rename plus a `BREAKING CHANGE:` footer in the commit so changelogithub picks it up.

sheremet-va (approved):

sheremet-va:
Thank you

PR #10276, opened 2026-05-05
perf: stringify diff objects only once

### LINKED ISSUE #781: False Positive Annoyance
### Describe the bug

To be clear, this bug doesn't break functionality, but it does get really annoying sometimes. :)

-----

Vitest extension shows lots of false positive `expect(...).toEqual(...)` mismatches when using partial schema checks.

For instance,

My `expect` statement makes use of `expect.objectContaining`.

Image

&nbsp;

But, unless the entire schema matches perfectly, the Vitest extension assumes these don't match:

The issue here is that the extension assumes that `ObjectContaining != Object` even when there is no actual issue with the equality test, making me think `expect.objectContaining` doesn't work here, or that there's actually a comparison issue (e.g., the object *doesn't* contain something required).

Image

&nbsp;

-----

To be clear, the `vitest` package / commandline has no problem exclusively highlighting where the actual issue is:

Image

&nbsp;

Note that **none** of the `ObjectContaining` lines are marked in red/green because nothing is actually wrong with any of them.

### Expected behavior

Vitest extension should only highlight in red the actual issues, to match the behavior of the commandline.

I'm guessing this occurs because Vitest extension is just using VSCode's built-in text-diffing system. I really think in the future it should be switched to something with better clarity (unless we can control the diff-editor diff-ing behavior :-?). At this point, it makes it really hard sometimes to see where the actual issues are.

### Reproduction

Running this test with the Vitest extension easily reproduces the issue:

```ts
import { describe, expect, it } from "vitest";

describe("diffing issue example", () => {
  it("should not show red for ObjectContaining", () => {
    // arrange
    const actual = [
      { x: 1, y: 2, z: 3 },
      { x: 3, y: 1, z: 2 },
      { x: 2, y: 3, z: 1 },
      "wrong",
    ];

    // act / assert
    expect(actual).toEqual([
      expect.objectContaining({ x: 1 }),
      expect.objectContaining({ z: 2 }),
      expect.objectContaining({ y: 3 }),
    ]);
  });
});
```

Image

### Output

```shell
[4:32:53 PM] [API] Filtering projects: foo
[INFO 4:32:53 PM] [API] Running Vitest v3.0.7 (project/vitest.workspace.ts) with "D:\nodejs\node.EXE C:/Users/Me/.vscode/extensions/vitest.explorer-1.50.2/dist/worker.js"
[4:32:53 PM] Enqueuing "should not show red for ObjectContaining"
[INFO 4:32:53 PM] Running 1 file(s) with name pattern: ^\s?diffing issue example
[INFO 4:32:54 PM] [DECOR] Reset all decorations.
[4:32:54 PM] No task result for "test.test.ts [foo]", ignoring
[4:32:54 PM] No task result for "diffing issue example", ignoring
[4:32:54 PM] Enqueuing "should not show red for ObjectContaining"
[4:32:54 PM] No errors found for "test.test.ts [foo]"
[4:32:54 PM] No errors found for "diffing issue example"
[4:32:54 PM] Could not find a valid stack for should not show red for ObjectContaining [
  {
    "method": "",
    "file": "D:/project/foo/test.test.ts",
    "line": 14,
    "column": 20
  }
]
[4:32:54 PM] Marking "should not show red for ObjectContaining" as failed with 1 errors
[4:32:54 PM] Disposing test runner
[4:32:54 PM] Ending test run foo/test.test.ts
[4:32:54 PM] [API] Vitest WebSocket connection closed, cannot call RPC anymore.
```

### Extension Version

vitest.explorer v1.50.2

### Vitest Version

vitest@3.0.7

### Validations

- [x] Check that you are using the latest version of the extension
- [x] Check that there isn't [already an issue](https://github.com/vitest-dev/vscode/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vscode/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).

sheremet-va:
### Description

Resolves https://github.com/vitest-dev/vscode/issues/781

This PR does two things:

- reuses the formatted string of expected/actual instead of creating a new one
  - this also makes it so the diff in other tools looks the same way because we do a special handling of async matchers
  - we follow `diffOptions` in those properties now (this causes the snapshot in tests to be big)


packages/utils/src/diff/index.ts
```diff
@@ -62,6 +62,17 @@ const FALLBACK_FORMAT_OPTIONS = {
   plugins: PLUGINS,
 } satisfies PrettyFormatOptions
 
+export interface StringifiedMemory {
```
sheremet-va, line 65: I initially just wanted to use `WeakMap`, but we are cloning and reorganizing (set/map) objects, so it's easier to just have an object that sets it

hi-ogawa (commented):
The idea makes sense to me, but is it important to stash the result through `memorize`?
Naively thinking, can we extend `diff` to return something like `{ result: string, actual?: string, expected?: string }`? I'd get stashing could be simpler if `diff` is recursing beast, but doesn't look like so. 
It could be just minor nit, but just asking.

packages/utils/src/diff/index.ts
```diff
@@ -62,6 +62,17 @@ const FALLBACK_FORMAT_OPTIONS = {
...
+  expected?: string
+  actual?: string
+}
+
+interface Memorize {
+  (pointer: 'expected' | 'actual', stringifiedValue: string): string
+}
```
hi-ogawa, line 72: Ah, I think what I felt odd is that `memorize` interface isn't even explicitly "stashing the result as argument" API.
So my question is, can this be `diff(..., memory?: StringifiedMemory)` directly instead of stashing through calling `memorize: Memorize`?

sheremet-va: It could, but it’s just less code with a util. It’s ugly to have val = string after every stringify call 😬

sheremet-va:
> It could be just minor nit, but just asking.

We could return different result, but I think the approach in pr is just less code to implement, but it does rely a lot on mutating state which, I admit, is hard to follow

packages/utils/src/diff/index.ts
hi-ogawa: I get the compactness but it feels an unnecessary indirection for my brain. not blocking though.

hi-ogawa (approved):

PR #10293, opened 2026-05-07
fix!: don't emit localStorage warnings on Node 26, fail gracefully when worker fails to start

sheremet-va:
### Description

Resolves #issue-number

### Tests

### Documentation

### Changesets


packages/vitest/src/node/pools/poolRunner.ts
hi-ogawa, line 1: Is this a separate independent fix?

packages/vitest/src/integrations/env/jsdom-keys.ts
```diff
@@ -209,6 +209,7 @@ const OTHER_KEYS = [
   'innerHeight',
   'innerWidth',
   'length',
+  'localStorage',
```
hi-ogawa, line 212: I forgot what this means. Does this mean previously we didn't expose `local/sessionStorage` as jsdom/happy-dom global?

packages/vitest/src/node/pools/poolRunner.ts
sheremet-va: Ah, yes. This PR was initially just to add support for Node 26, I could split it into separate PRs if you want. It's small enough that I was too lazy and just kept it here :P

packages/vitest/src/integrations/env/jsdom-keys.ts
sheremet-va: Previously we didn't override new window's keys if the value exists on `globalThis` already. This list basically says "override anyway"

packages/vitest/src/node/pools/poolRunner.ts
hi-ogawa: I'd like separate PR, but also fine if it's included in pr title and description properly.

sheremet-va: Ok, included in title 😄

packages/vitest/src/integrations/env/jsdom-keys.ts
hi-ogawa: Now I see. `getWindowKeys` loop is mind-bending and took a while to comprehend 😓

hi-ogawa (approved):

Issue #9437, opened 2026-01-13
Browser mode: Chromium disk usage accumulates per test file, exhausting disk on long runs


MariusVatasoiu:
### Describe the bug

When running vitest with browser mode on Azure DevOps it fails after a few tests. The error is:

```
⎯⎯⎯⎯⎯⎯ Unhandled Error ⎯⎯⎯⎯⎯⎯⎯
Error: Failed to run the test /home/vsts/work/1/s/src/foo/bar.spec.js.
 ❯ ../node_modules/vitest/dist/chunks/cli-api.Cx2DW4Bc.js:7255:17
 ❯ processTicksAndRejections ../node:internal/process/task_queues:103:5

Caused by: Error: Cannot connect to the iframe. Did you change the location or submitted a form? If so, don't forget to call `event.preventDefault()` to avoid reloading the page.

Received URL: unknown
Expected: https://localhost:63315/?sessionId=77b75b8a-78f3-4b0f-87ec-89b3c8d3650b&iframeId=/home/vsts/work/1/s/src/foo/bar.spec.js
 ❯ iframe.onload ../../../../../../__vitest_browser__/orchestrator-S_3e_uzt.js:169:43
```

It fails after more than around 1000 tests. When it fails, most of the time I get a warning from Azure:

```
##[warning]Free disk space on / is lower than 5%; Currently used: 97.31%
```

The agent runs on Ubuntu. I checked the disk usage:

```
Filesystem      Size  Used Avail Use% Mounted on
/dev/root        72G   60G   13G  83% /
tmpfs           3.9G   84K  3.9G   1% /dev/shm
tmpfs           1.6G  1.1M  1.6G   1% /run
tmpfs           5.0M     0  5.0M   0% /run/lock
/dev/sda16      881M   62M  758M   8% /boot
/dev/sda15      105M  6.2M   99M   6% /boot/efi
tmpfs           795M   12K  795M   1% /run/user/1001
```

I've enabled `system.log: true` on my pipeline and I can see some logs about the disk varying quite a lot during the tests.

```
NODE_ENV=production vitest --run --no-color --silent --retry 3

Vite command: serve, mode: test

 RUN  v4.0.17 /home/vsts/work/1/s/src

Vite command: serve, mode: test
[vite-plugin-static-copy] Collected 2 items.
[...]

PR #10300, opened 2026-05-08
fix(browser): fix iframe leak in the orchestrator

mourner:
Fix a renderer-process leak in `IframeOrchestrator` that retained detached iframes and ultimately crashed Chromium during long browser-mode runs.

Background — I'm one of the core developers of [Mapbox GL JS](https://github.com/mapbox/mapbox-gl-js/), which runs an extensive suite of Vitest-based tests. [`unit-test` part](https://github.com/mapbox/mapbox-gl-js/blob/main/vitest.config.unit.ts) of them (~2900 tests across ~190 files) has been increasingly flaky over at least a year, getting worse with increased test coverage, to the point that recently the CI job failed half of the time.

After multiple days of debugging, tracing various leads and eliminating theories, we've finally tracked it down to an iframe leak in the Vitest orchestrator. Disclosing that the fix was made using Claude Opus 4.7, but it's minimal enough to be easily reviewed by a human. ~~We've verified that the fix makes Vitest stop accumulating resources (documents, frames & listeners) during that test run.~~ **Edit** Apologies, converting to draft while we verify correctly that this 100% fixes the issue.

I haven't added a test for this because it's a hard thing to test — it only manifests on long-running browser tests in a slow CI environment, and a more targeted unit test would require rewriting the orchestrator more substantially to hook into it from the test suite. Perhaps the change is minimal enough so that it's fine to just make sure the existing suite passes? Let me know how to make this easier for you to merge.

### Description

When an iframe is force-removed mid-flight (next-batch reset, isolated retry), any in-flight `sendEventToIframe` Promise never settles — the `response:<event>` postMessage from the now-removed iframe never arrives. The async function awaiting it stays suspended forever, and its closure pins the detached iframe, its Document, and all attached event listeners. JS heap stays small (~150 MB), but Chromium-internal frame-graph state grows unbounded across files and eventually crosses a renderer-internal limit, killing the renderer with no JS event of any kind firing — surfacing as the long-standing flake `Browser connection was closed while running tests. Was the page closed unexpectedly?`.

**The fix.** Track per-iframe abort callbacks in a `WeakMap` and reject pending `sendEventToIframe` promises when the iframe is removed, releasing the closures and unwinding the detached frame state.

- New `pendingAborts: WeakMap<HTMLIFrameElement, Set<abort>>` — replaces the prior `iframeEvents` map (1:1 with pending promises, so the two are merged into one).
- New `removeIframe(iframe)` helper drains the pending aborts before calling `iframe.remove()`. All three direct `iframe.remove()` call sites in `IframeOrchestrator` go through it.
- `sendEventToIframe` registers an `abort` callback; `onReceived`, `onError`, and `abort` all funnel through one `cleanup()` that removes the channel/eventTarget listeners and the pending entry.

**Evidence.** Diagnosed via CDP `Performance.getMetrics` polling on a flaky CI run: at the death edge, `Documents` climbed 13 → 51, `Frames` 8 → 27, and `JSEventListeners` ran away from ~200 to **1744 in the final second** before the renderer was killed — while JS heap stayed at 144 MB / 4 GB (3.5%). A `HeapProfiler.collectGarbage` workaround every 5 s held all those counters at baseline and the run passed, confirming the retainers were GC-eligible if pressure was relieved. A heap snapshot then pinned the retainer to a single suspended `prepareIframe` async generator awaiting `sendEventToIframe({event: 'prepare'})`, with its closure capturing `{iframe: <detached>, iframeId, otelContext, startTime}` and a never-settled Promise as a GC root.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/browser/src/client/orchestrator.ts
+++ b/packages/browser/src/client/orchestrator.ts
@@ -80,7 +80,7 @@ export class IframeOrchestrator {
       return
     }
 
-    this.iframes.forEach(iframe => iframe.remove())
+    this.iframes.forEach(iframe => this.removeIframe(iframe))
     this.iframes.clear()
 
     for (let i = 0; i < options.files.length; i++) {
@@ -140,7 +140,7 @@ export class IframeOrchestrator {
       // because we called "cleanup" in the previous run
       // the iframe is not removed immediately to let the user see the last test
       this.recreateNonIsolatedIframe = false
-      this.iframes.get(ID_ALL)!.remove()
+      this.removeIframe(this.iframes.get(ID_ALL)!)
       this.iframes.delete(ID_ALL)
       debug('recreate non-isolated iframe')
     }
@@ -180,7 +180,7 @@ export class IframeOrchestrator {
     const file = spec.filepath
 
     if (this.iframes.has(file)) {
-      this.iframes.get(file)!.remove()
+      this.removeIframe(this.iframes.get(file)!)
       this.iframes.delete(file)
     }
 
@@ -234,8 +234,7 @@ export class IframeOrchestrator {
           )))
         }
         else if (this.iframes.has(iframeId)) {
-          const events = this.iframeEvents.get(iframe)
-          if (events?.size) {
+          if (this.pendingAborts.get(iframe)?.size) {
             this.dispatchIframeError(new Error(this.createWarningMessage(iframeId, 'during a test')))
           }
           else {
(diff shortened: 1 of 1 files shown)


mourner:
Looks like the two failing CI jobs are flaky tests — unrelated to the change.

stepankuzmin:
This is related to the https://github.com/vitest-dev/vitest/discussions/7981

mourner:
Closing because it turned out this doesn't fully fix the issue — only a force-GC every 5s hack works, but any fix on the JS side doesn't, so it seems like this might be a bug or a race condition on the browser level. Here's the hack:

```js
function orchestratorGCReporter(): Reporter {
    let interval: NodeJS.Timeout | null = null;
    return {
        onInit(vitest: {projects: TestProject[]}) {
            let cdp: CDP | null = null;
            const tick = async () => {
                if (!cdp) {
                    const project = vitest.projects.find((p) => (p.browser as unknown as {provider?: Provider})?.provider);
                    const provider = project && (project.browser as unknown as {provider: Provider}).provider;
                    const sessionId = provider?.pages.keys().next().value;
                    if (!provider || !sessionId) return;
                    cdp = await provider.getCDPSession(sessionId);
                }
                await cdp.send('HeapProfiler.collectGarbage');
            };
            interval = setInterval(() => { tick().catch(() => { cdp = null; }); }, 5000);
            interval.unref?.();
        },
        onTestRunEnd() {
            if (interval) clearInterval(interval);
            interval = null;
        },
    };
}
// then you'd do e.g.:
// reporters: [['github-actions'], orchestratorGCReporter()]
```

I'll see if there's a way to add it naturally upstream in Vitest (perhaps via an option?) rather than a reporter hack.

AriPerkkio:
Probably Chromium bug related root cause: https://github.com/vitest-dev/vitest/issues/9437#issuecomment-5016763492

PR #10302, opened 2026-05-08
feat(browser): support custom `kind` in `page.mark`

AriPerkkio:
### Description

Allow users to defined custom `kind` for their markers. For example custom actions may want to show up as actions:

image

### Tests

### Documentation

### Changesets

### DIFF
--- a/docs/api/browser/context.md
+++ b/docs/api/browser/context.md
@@ -82,11 +82,11 @@ export const page: {
   /**
    * Add a trace marker when browser tracing is enabled.
    */
-  mark(name: string, options?: { stack?: string; kind?: MarkKind }): Promise<void>
+  mark(name: string, options?: { stack?: string; kind?: BrowserTraceEntryKind }): Promise<void>
   /**
    * Group multiple operations under a trace marker when browser tracing is enabled.
    */
-  mark<T>(name: string, body: () => T | Promise<T>, options?: { stack?: string; kind?: MarkKind }): Promise<T>
+  mark<T>(name: string, body: () => T | Promise<T>, options?: { stack?: string; kind?: BrowserTraceEntryKind }): Promise<T>
   /**
    * Extend default `page` object with custom methods.
    */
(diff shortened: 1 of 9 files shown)


sheremet-va:
Unrelated, but I wonder if it makes sense to expose mark on the server (when running a command) too so there is no extra calls if you already execute a command

hi-ogawa:
Is it important that your `mark` to show up exactly like current action? I'm thinking `kind` is mostly (randomly-chosen) initial color labeling, which may not be a proper API, so it maybe be better to let mark caller pick the color.

hi-ogawa:
> Unrelated, but I wonder if it makes sense to expose mark on the server (when running a command) too so there is no extra calls if you already execute a command

@sheremet-va I'm not following what this means. Can you explain the idea more detail?

AriPerkkio:
> Is it important that your `mark` to show up exactly like current action? I'm thinking `kind` is mostly (randomly-chosen) initial color labeling, which may not be a proper API, so it maybe be better to let mark caller pick the color.

It's not 100% necessary, but I think it makes sense for users to be able to categorize their custom marks with `kind`. I would expect users to have their custom `kind: 'expect'`, `kind: 'lifecycle'` and `kind: 'action'`. Having `kind: 'mark'` is good default.

> > Unrelated, but I wonder if it makes sense to expose mark on the server (when running a command) too so there is no extra calls if you already execute a command
>
> @sheremet-va I'm not following what this means. Can you explain the idea more detail?

I think that means being able to call `page.mark` on server side, in custom command? That would be very useful. 

```ts
export default defineConfig({
  test: {
    browser: {
      provider: playwright(),
      commands: {
        example(context) {
          context.page.mark("example")
        }
      }
    }
  },
});
```

hi-ogawa:
> It's not 100% necessary, but I think it makes sense for users to be able to categorize their custom marks with `kind`. I would expect users to have their custom `kind: 'expect'`, `kind: 'lifecycle'` and `kind: 'action'`. Having `kind: 'mark'` is good default.

Okay, I think I understood `kind` can be exposed as a known set of presentational metadata. I said it's internal because the `kind` is also partially used for internal step name rewriting. Probably we can just drop `kind` check from here and treat `vitest:` prefix to be owned by Vitest regardless of `kind`.

https://github.com/vitest-dev/vitest/blob/b996f2ebf2c03731f32b23745196224412f04350/packages/ui/client/components/trace/TraceView.vue#L135-L143

sheremet-va:
> That would be very useful.

Yes, that’s what I meant. It could be exposed alongside other properties on command context (context.mark instead of context.page.mark). Page one is playwright specific.

hi-ogawa (dismissed):

AriPerkkio:
Ok ready now. Had to refactor typings to another file to avoid duplication.

hi-ogawa (approved):

sheremet-va (approved):

PR #10309, opened 2026-05-09
fix: export UserConsoleLog type from vitest/node

### LINKED ISSUE #10305: `UserConsoleLog` type is not exported from vitest

The `Reporter.onUserConsoleLog(log)` callback receives a `UserConsoleLog`, but the type isn't exported from `vitest`, `vitest/node`, or `vitest/reporters`. Custom reporters written in TypeScript have to redeclare the interface locally.

For example, that's what we do: https://github.com/flakiness/vitest/blob/39ef4c65e8111c8e2d8612d460332bab44627f9a/src/reporter.ts#L13-L20
```ts
interface UserConsoleLog {
  content: string;
  origin?: string;
  browser?: boolean;
  type: "stdout" | "stderr";
  taskId?: string;
  time: number;
  size: number;
}
```

This is brittle: if the upstream shape changes, the reporter silently diverges.

### Suggested solution

It would be useful to export `UserConsoleLog` (and any other reporter-callback types in the same situation) from `vitest/reporters`.

### Alternative

_No response_

### Additional context

_No response_

### Validations

xxiaoxiong:
## Summary

Fixes #10305

This PR exports the `UserConsoleLog` type from `vitest/node` so custom reporters can import it instead of redeclaring it locally.

## Problem

The `Reporter.onUserConsoleLog(log)` callback receives a `UserConsoleLog`, but the type isn't exported from `vitest/node`. Custom reporters written in TypeScript have to redeclare the interface locally:

```ts
interface UserConsoleLog {
  content: string;
  origin?: string;
  browser?: boolean;
  type: "stdout" | "stderr";
  taskId?: string;
  time: number;
  size: number;
}
```

This is brittle: if the upstream shape changes, the reporter silently diverges.

## Solution

Export `UserConsoleLog` from `vitest/node` alongside other reporter types like `Reporter`, `BaseReporter`, and `TestRunEndReason`.

**After this PR:**
```ts
import type { UserConsoleLog } from 'vitest/node'
```

## Changes

- `packages/vitest/src/node/reporters/index.ts`
  - Import `UserConsoleLog` from `../types/general`
  - Add `UserConsoleLog` to type exports
- `packages/vitest/src/public/node.ts`
  - Add `UserConsoleLog` to the re-exported types from `../node/reporters`

## Testing

The type is already defined in `packages/vitest/src/types/general.ts` and used internally. This PR only adds the export path.

### DIFF
--- a/packages/vitest/src/node/reporters/index.ts
+++ b/packages/vitest/src/node/reporters/index.ts
@@ -1,5 +1,5 @@
 import type { Reporter, TestRunEndReason } from '../types/reporter'
-import type { UserConsoleLog } from '../types/general'
+import type { UserConsoleLog } from '../../types/general'
 import type { BaseOptions, BaseReporter } from './base'
 import type { BlobOptions } from './blob'
 import type { DefaultReporterOptions } from './default'
--- a/packages/vitest/src/node/reporters/index.ts
+++ b/packages/vitest/src/node/reporters/index.ts
@@ -1,4 +1,5 @@
 import type { Reporter, TestRunEndReason } from '../types/reporter'
+import type { UserConsoleLog } from '../types/general'
 import type { BaseOptions, BaseReporter } from './base'
 import type { BlobOptions } from './blob'
 import type { DefaultReporterOptions } from './default'
@@ -33,7 +34,7 @@ export {
   TreeReporter,
   VerboseReporter,
 }
-export type { BaseReporter, Reporter, TestRunEndReason }
+export type { BaseReporter, Reporter, TestRunEndReason, UserConsoleLog }
 
 export type { BenchmarkBuiltinReporters } from './benchmark'
 export {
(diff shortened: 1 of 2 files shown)


packages/vitest/src/node/reporters/index.ts
```diff
@@ -1,4 +1,5 @@
 import type { Reporter, TestRunEndReason } from '../types/reporter'
+import type { UserConsoleLog } from '../types/general'
```
AriPerkkio:
```suggestion
import type { UserConsoleLog } from '../../types/general'
```

AriPerkkio (changes requested):
Already exported here

https://github.com/vitest-dev/vitest/blob/ab09822a4fef07ddecfdce702bd21fcba7881198/packages/vitest/src/public/index.ts#L67-L75

xxiaoxiong:
You're absolutely right! I apologize for the noise. `UserConsoleLog` is already exported in `packages/vitest/src/public/index.ts` at line 74.

I should have checked the existing exports more carefully before submitting this PR. Closing this as it's not needed.

Thank you for the quick review!

PR #10315, opened 2026-05-10
chore(browser): diagnostic logging for playwright trace flake (DO NOT MERGE)

dokson:
## Purpose

This is a **draft** to gather diagnostic data on intermittent failures observed in the browser-mode CI matrix. Not intended for merge.

## What we're after

- Extra `webkit-repeated-retried-tests-0-2.trace.zip` under `on-first-retry` mode (gate should evaluate `false` for `retry === 2`, but the file appears anyway).
- 360 s timeouts on `playwright-trace.test.ts`, `runner.test.ts > stack trace points to correct file ...`, `benchmark.test.ts` ending with `[vitest-pool]: Timeout terminating threads worker ...`. The shape suggests a browser session that does not release on teardown.

The flakes do not reproduce locally on Windows nor in a Docker Linux container with CPU pressure, which is consistent with environment-specific timing.

## Approach

Earlier revisions of this draft instrumented the same surfaces with `console.log` and `console.error`, both of which got captured into stdout / stderr snapshots of unrelated e2e tests (`merge-reports.test.ts`, `list.test.ts`, `runner.test.ts > timeout hooks`) and broke them. This revision routes everything to a file via a no-op-by-default helper, so the diagnostic can never pollute another test.

`packages/browser-playwright/src/dbg.ts` (new):
```ts
export function traceDbg(msg: string): void {
  const f = process.env.VITEST_TRACE_DBG_FILE
  if (!f) return
  appendFileSync(f, `${Date.now()} ${msg}\n`)
}
```

`packages/browser-playwright/src/commands/dbg.ts` (new): registers `__vitest_logTraceDbg`, so the in-browser runner can ship entries to the parent process through the existing RPC channel without ever touching browser-side `console`.

The runner emits `BEFORE`, `AFTER`, `STOP` entries around `onBeforeTryTask`/`onAfterRetryTask`, wrapped in `.catch(() => {})` so the diagnostic cannot fail a test. The provider emits `PROVIDER start/stop`, `SIGTERM`, `CLOSE start/done` from `commands/trace.ts` and `playwright.ts`. Each entry is timestamped.

`.github/workflows/ci.yml` sets `VITEST_TRACE_DBG_FILE: ${{ runner.temp }}/trace-dbg.log` for the `test-browser` job and uploads the file as `trace-dbg-{os}-node-{ver}` artifact (with `if-no-files-found: ignore` so unrelated paths do not break).

## Plan

1. Run CI on this branch.
2. Download the `trace-dbg-*` artifact for the failing job.
3. Walk the timeline around `webkit-repeated-retried-tests-0-2` (or around the first hang for the 360 s timeout) to identify the offending state transition.
4. Open a separate, properly scoped fix PR.
5. Close this draft.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -214,6 +214,16 @@ jobs:
 
       - name: Test Browser (playwright)
         run: pnpm run test:browser:playwright
+        env:
+          VITEST_TRACE_DBG_FILE: ${{ runner.temp }}/trace-dbg.log
+
+      - name: Upload trace-dbg log
+        if: always()
+        uses: actions/upload-artifact@v7
+        with:
+          name: trace-dbg-${{ matrix.os }}-node-${{ matrix.node_version }}
+          path: ${{ runner.temp }}/trace-dbg.log
+          if-no-files-found: ignore
 
       - name: Test Browser (webdriverio)
         run: pnpm run test:browser:webdriverio
--- a/packages/browser-playwright/src/commands/dbg.ts
+++ b/packages/browser-playwright/src/commands/dbg.ts
@@ -1,0 +1,6 @@
+import type { BrowserCommand } from 'vitest/node'
+import { traceDbg } from '../dbg'
+
+export const logTraceDbg: BrowserCommand<[{ msg: string }]> = (_ctx, { msg }) => {
+  traceDbg(msg)
+}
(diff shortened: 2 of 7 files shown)


dokson:
Superseded by #10316 which combines the fix (beforeEach trace cleanup) with this same diagnostic infrastructure plus a no-op webdriverio command stub to avoid the unknown-command-RPC regression that was visible in this PR's runs.

PR #10263, opened 2026-05-04
fix(reporters): include source code in blob to survive cross-OS merge (fix #10260)

dokson:
### Description

- Closes https://github.com/vitest-dev/vitest/issues/10260

When sharded CI matrices produce blob reports on multiple OSes (e.g. macOS + Windows + Ubuntu) and they are then combined with \`--merge-reports\` on a single runner (typically Linux), the HTML reporter cannot read the test sources for paths that live on a different OS. The merger tries \`fs.readFile('/Users/runner/...')\` / \`fs.readFile('C:\...')\` on its own filesystem, the read silently fails (caught by the existing \`try/catch\`), and the resulting \`html.meta.json\` ends up with no \`sources\` entries for those platforms. The "Code" tab in the merged HTML report is empty for any non-Linux runner.

I confirmed this on the artifact attached to #10260: out of the e2e merged report, **336 Linux files had source, 112 macOS and 112 Windows files had no source**. Same pattern as the screenshots in the issue.

### Fix

- Read the test sources on the **producer** side (\`BlobReporter.onTestRunEnd\`) — that runner is the one that has the files locally — and embed them in the blob (a 7th tuple element \`sources: Record<string, string>\`).
- \`readBlobs\` merges the per-blob \`sources\` maps into the returned \`MergedBlobs\` and \`mergeReports\` propagates them to \`state.blobs.sources\`.
- \`HTMLReporter.onTestRunEnd\` now prefers \`state.blobs.sources[file.filepath]\` when present, falling back to the local filesystem read (so non-merge runs are unchanged).

### Test plan

- [x] New e2e test in \`test/e2e/test/reporters/merge-reports.test.ts\` (\`html reporter retains source for files unreachable from the merger filesystem\`) that simulates the cross-OS scenario by **renaming the fixture sources** between the blob run and the merge run, then verifies the merged \`html.meta.json.gz\` still contains the source content for both files. Test fails without this patch (sources are empty maps) and passes with it.
- [x] All 8 existing tests in \`merge-reports.test.ts\` still pass.

### Compatibility

The 7th tuple element is optional (\`sources?\`) and \`readBlobs\` defaults to \`{}\` when missing, so blobs produced by older versions can still be consumed (the version-equality check in \`readBlobs\` already prevents cross-version mixing in practice, but the optional field keeps the type permissive).

### Please don't delete this checklist! Before submitting the PR, please make sure you do the following:
[...]

PR #10323, opened 2026-05-11
fix(reporters): include source code in blob to survive cross-OS merge (fix #10260)

### LINKED ISSUE #10260: HTML reporter source code view is missing in some platform
This is the merged report from current main run and it's missing source code for macos and windows. Not sure it's related to new blob label, merge or just has been missing in html.
https://viewer.vitest.dev/?url=https://github.com/vitest-dev/vitest/actions/runs/25306824033/artifacts/6778859721 

<details><summary>macos</summary>

Image

</details>

<details><summary>ubuntu</summary>

Image

</details>

dokson:
### Description

- Closes #10260
- Reopens the work of #10263 (auto-closed by the AI-detection bot before a maintainer could weigh in; rebased onto current `main` with a couple of TypeScript-pro audit refinements).

When sharded CI matrices produce blob reports on multiple OSes (e.g. macOS + Windows + Ubuntu) and they are then combined with `--merge-reports` on a single runner (typically Linux), the HTML reporter cannot read the test sources for paths that live on a different OS. The merger tries `fs.readFile('/Users/runner/...')` / `fs.readFile('C:\...')` on its own filesystem, the read silently fails (caught by the existing `try/catch`), and the resulting `html.meta.json` ends up with no `sources` entries for those platforms. The "Code" tab in the merged HTML report is empty for any non-Linux runner.

I confirmed this on the artifact attached to #10260: out of the e2e merged report, **336 Linux files had source, 112 macOS and 112 Windows files had no source**. Same pattern as the screenshots in the issue.

### Fix

- Read the test sources on the **producer** side (`BlobReporter.onTestRunEnd`) — that runner is the one that has the files locally — and embed them in the blob (a 7th tuple element `sources?: Record<string, string>`).
- `readBlobs` merges the per-blob `sources` maps into the returned `MergedBlobs` and `mergeReports` propagates them to `state.blobs.sources`.
- `HTMLReporter.onTestRunEnd` now prefers `state.blobs.sources[file.filepath]` when present, falling back to the local filesystem read (so non-merge runs are unchanged).

### Test plan

- New e2e test in `test/e2e/test/reporters/merge-reports.test.ts` (`html reporter retains source for files unreachable from the merger filesystem`) that simulates the cross-OS scenario by **renaming the fixture sources** between the blob run and the merge run, then verifies the merged `html.meta.json.gz` still contains the source content for both files. The test fails without this patch (sources are empty maps) and passes with it.
- All 8 existing tests in `merge-reports.test.ts` still pass.

### Compatibility

The 7th tuple element is optional (`sources?`) and `readBlobs` defaults to `{}` when missing, so blobs produced by older versions can still be consumed (the version-equality check in `readBlobs` already prevents cross-version mixing in practice, but the optional field keeps the type permissive).

### Refinements over the previous attempt

- `MergedBlobs.sources` is required (not optional) so consumers never write a redundant optional chain — `readBlobs` always returns a value, defaulting to `{}` for older blobs.
- The dedup guard on the producer uses `in` instead of truthiness, so empty source files don't get re-read.
- Brief TSDoc on `MergedBlobs.sources` documenting the producer/merger contract.


dokson:
cc @hi-ogawa — you were the assignee on the original #10263 before the AI-detection bot auto-closed it. Same fix, rebased onto current main with a couple of TS audit refinements. Happy to walk through any concern.

PR #10331, opened 2026-05-12
chore!: remove vitest/import-meta reexport

lishaduck:
### Description

Deletes vitest/import-meta, as proposed in https://github.com/vitest-dev/vitest/discussions/10330.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/package.json
+++ b/packages/vitest/package.json
@@ -61,9 +61,6 @@
     "./importMeta": {
       "types": "./importMeta.d.ts"
     },
-    "./import-meta": {
-      "types": "./import-meta.d.ts"
-    },
     "./node": {
       "types": "./dist/node.d.ts",
       "default": "./dist/node.js"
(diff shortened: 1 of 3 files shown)


lishaduck:
Looks like flakes maybe?

packages/vitest/package.json
```diff
@@ -61,9 +61,6 @@
     "./importMeta": {
       "types": "./importMeta.d.ts"
     },
-    "./import-meta": {
-      "types": "./import-meta.d.ts"
```
sheremet-va, line 65: I think we should still keep `./import-meta` - just use `importMeta.d.ts` types for it

sheremet-va (approved):

packages/vitest/package.json
lishaduck: I'm mildly confused-why? The point of this PR is that the import-meta export isn't necessary anymore.
If we want to keep import-meta, shouldn't we close this PR as wontfix?

sheremet-va: I think this is more of a preference issue. Using `import-meta` or `importMeta` should just be a decision by the user, it could be non-breaking. If we need to remove anything, it should be the comment

We also generally use `-` for exports (see `optional-types` and `optional-runtime-types` above)

PR #10332, opened 2026-05-12
fix(browser): use resolve instead of join for screenshotDirectory path resolution

### LINKED ISSUE #9636: Setting `browser.screenshotDirectory` leads to two folders being created
### Describe the bug

If I set `browser.screenshotDirectory` and run my tests, I get:

```
 FAIL   visual (chromium)  tests/routes/(home)/visual.test.ts > Landing page (visuals) > ..looks like before
Error: expect(element).toMatchScreenshot()

No existing reference screenshot found; a new one was created. Review it before running tests again.

Reference screenshot:
  /home/ubuntu/website.1/tests/routes/(home)/home/ubuntu/website.1/__shots__/visual.test.ts/home-chromium-linux.png


Failure screenshot:
  - __shots__/tests/routes/(home)/visual.test.ts/Landing-page--visuals----looks-like-before-1.png

```

While the "Failure screenshot" path is as expected, the "Reference screenshot" path is strange. 

Expectation:
- only one folder to be named (`__shots__/tests/routes/[...]`)

Actual:
- a second, strange path gets created

vitest 4.0.17, 4.0.18

### Reproduction

Sorry no MVR at this hour.. Going on a trip and wanted at least to share this, in hopes a fix might land by 4.1.

### System Info

```shell
System:
    OS: Linux 6.8 Ubuntu 24.04.3 LTS 24.04.3 LTS (Noble Numbat)
    CPU: (2) x64 06/9e
    Memory: 2.53 GB / 3.82 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  Binaries:
    Node: 24.13.0 - /usr/bin/node
    npm: 11.6.2 - /usr/bin/npm
  npmPackages:
    @vitest/browser-playwright: ^4.0.18 => 4.0.18 
    @vitest/ui: ^4.0.15 => 4.0.18 
    playwright: ^1.58.0 => 1.58.2 
    vite: 8.0.0-beta.13 => 8.0.0-beta.13 
    vitest: ^4.0.18 => 4.0.18 
    vitest-browser-svelte: ^2.0.2 => 2.0.2
```

### Used Package Manager

npm

### Validations

Zelys-DFKH:
#### What

When `browser.screenshotDirectory` is set to an absolute path, `resolveConfig` pre-resolves it to an absolute path before it reaches `utils.ts`. The call to `join(root, absolutePath)` then concatenates root and the absolute path as literal segments, producing broken paths like `/home/user/project/home/user/project/__shots__/...`.

The fix is one character: swap `join` for `resolve`. When the second argument is already absolute, `pathe.resolve` returns it as-is rather than appending it.

#### Why `join` vs `resolve` matters here

```ts
// resolveConfig pre-resolves screenshotDirectory when set:
resolved.browser.screenshotDirectory = resolve(resolved.root, resolved.browser.screenshotDirectory)

// So by the time utils.ts runs, it's already absolute.
join('/project', '/home/user/__shots__')    // → '/project/home/user/__shots__'  ✗
resolve('/project', '/home/user/__shots__') // → '/home/user/__shots__'           ✓
```

#### Why not PR #9656?

There's an existing PR (#9656) for the same issue. It uses `resolve('/', relative(root, path))` as the fix, but that approach breaks on Windows because `resolve('/')` anchors to the drive root (`C:\`) rather than a meaningful cross-platform root. That PR has a failing `Build&Test: node-24, windows-latest` CI job and has been stale since February. This PR uses the simpler, cross-platform approach.

#### Testing

This is exactly the scenario covered by the existing `to-match-screenshot` test suite. The default `__screenshots__` fallback is a relative path, so the `join`/`resolve` difference doesn't matter there — the change only affects users who set `screenshotDirectory` explicitly to an absolute path, which is the reported bug.

Fixes #9636.

### DIFF
--- a/packages/browser/src/node/commands/screenshotMatcher/utils.ts
+++ b/packages/browser/src/node/commands/screenshotMatcher/utils.ts
@@ -137,7 +137,7 @@ export function resolveOptions(
     root,
     screenshotDirectory: relative(
       root,
-      join(root, context.project.config.browser.screenshotDirectory ?? '__screenshots__'),
+      resolve(root, context.project.config.browser.screenshotDirectory ?? '__screenshots__'),
     ),
     attachmentsDir: relative(root, context.project.config.attachmentsDir),
     testFileDirectory: relative(root, dirname(context.testPath)),


Zelys-DFKH:
CI failures are all pre-existing. None are caused by the `screenshotDirectory` path change.

**`Browsers: node-24, macos-latest`** and **`Test: vite@7, node-24, macos-latest`**: the `runner.test.ts > timeout hooks` test runs the `timeout-hooks` fixture, which is intentionally designed to produce timeout errors (every test clicks a non-existent element with a 150ms limit). On slow macOS CI runners, the time to run all those intentional timeouts across three browsers exceeds `runBrowserTests`'s own wait limit, so it never receives `"Test Files  3 passed"` and gives up. The same failures appear in #10318 (merged yesterday) and are currently failing on `main`.

**`Cache&Test: node-24, windows-latest`**: `junit.test.ts > many errors without warning` spawns a `runVitestCli` child process and waits up to 60s for it to finish. On this Windows runner the process didn't complete in time. The same test passed in 227ms on May 11.

Zelys-DFKH:
Real person here, or so I'm told. I do use Claude for coding assistance, of course.

The bug is real: `path.join` silently drops absolute paths for `screenshotDirectory`. `path.resolve` handles both. I traced it, reviewed the fix, and stand behind it.

CI failures are all pre-existing, documented in the comment above.

@petergaal91, appreciate you checking in, as I had missed the original ping.

macarie:
We've discussed the issue internally and decided to solve it differently. The fix will be released as part of v5 (as it could be a potentially breaking change).

Thank you for the contribution anyway!

Zelys-DFKH:
Really appreciate you following up with context rather than just closing. Did some digging on the breaking concern: relative paths produce the same result with both calls, so the behavioral difference only surfaces for absolute `screenshotDirectory` values. Narrow edge case, but I get why you'd rather ship that deliberately in v5 with a migration path than catch anyone off guard in a patch. Excited to see what you have planned for v5.

PR #10340, opened 2026-05-13
fix(reporters): `summary` to intercept logger's streams even when they are not `process.std*` streams

AriPerkkio:
### Description

While adding test cases for summary reporter, I noticed bug around window rendered. It's really edge-case bug that happens only when using Node API and passing custom `stdout`/`stderr` while using summary reporter.

So this is more of a `test(reporter): summary reporter tests` than a `fix:`, but fix has more priority.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/node/reporters/renderers/windowedRenderer.ts
+++ b/packages/vitest/src/node/reporters/renderers/windowedRenderer.ts
@@ -14,7 +14,7 @@ const MOVE_CURSOR_ONE_ROW_UP = `${ESC}1A`
 const SYNC_START = `${ESC}?2026h`
 const SYNC_END = `${ESC}?2026l`
 
-interface Options {
+export interface Options {
   logger: Vitest['logger']
   interval?: number
   threshold?: number
--- a/packages/vitest/src/node/reporters/renderers/windowedRenderer.ts
+++ b/packages/vitest/src/node/reporters/renderers/windowedRenderer.ts
@@ -2,6 +2,10 @@ import type { Writable } from 'node:stream'
 import type { Vitest } from '../../core'
 import { stripVTControlCharacters } from 'node:util'
 
+/** Minimum time between two renders, no matter how many scheduled renderes were called */
+const DEFAULT_RENDER_THRESHOLD_MS = 100
+
+/** Interval between automatic renders. If no test state changes happened, this will increase just duration field */
 const DEFAULT_RENDER_INTERVAL_MS = 1_000
 
 const ESC = '\x1B['
@@ -13,6 +17,7 @@ const SYNC_END = `${ESC}?2026l`
 interface Options {
   logger: Vitest['logger']
   interval?: number
+  threshold?: number
   getWindow: () => string[]
 }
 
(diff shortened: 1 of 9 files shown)


packages/vitest/src/node/reporters/renderers/windowedRenderer.ts
```diff
@@ -2,7 +2,7 @@ import type { Writable } from 'node:stream'
 import type { Vitest } from '../../core'
 import { stripVTControlCharacters } from 'node:util'
 
-const DEFAULT_RENDER_INTERVAL_MS = 1_000
+const DEFAULT_RENDER_INTERVAL_MS = 100
```
AriPerkkio: This value was never used 🙃 

https://github.com/vitest-dev/vitest/blob/fab1b6020bb1118b52308bdf02f4fef0567f347a/packages/vitest/src/node/reporters/renderers/windowedRenderer.ts#L89-L91

packages/vitest/src/node/reporters/renderers/windowedRenderer.ts
```diff
@@ -48,6 +49,8 @@ export class WindowRenderer {
     this.cleanups.push(
       this.interceptStream(process.stdout, 'output'),
       this.interceptStream(process.stderr, 'error'),
+      this.interceptStream(options.logger.outputStream, 'output'),
+      this.interceptStream(options.logger.errorStream, 'error'),
```
AriPerkkio: This part fixes a bug where virtual buffer would not capture `logger.log` calls that were made by other reporters when custom `VitestOptions.stdout` was used.

packages/vitest/src/node/reporters/renderers/windowedRenderer.ts
AriPerkkio: Oh actually it is used here, for a bit different purpose.

https://github.com/vitest-dev/vitest/blob/fab1b6020bb1118b52308bdf02f4fef0567f347a/packages/vitest/src/node/reporters/renderers/windowedRenderer.ts#L63

Need to add another (internal) option.

packages/vitest/src/node/reporters/summary.ts
```diff
@@ -12,6 +12,12 @@ const FINISHED_TEST_CLEANUP_TIME_MS = 1_000
 
 interface Options {
   verbose?: boolean
+
+  /** @internal */
+  interval?: ConstructorParameters<typeof WindowRenderer>[0]['interval']
+
+  /** @internal */
+  threshold?: ConstructorParameters<typeof WindowRenderer>[0]['threshold']
```
hi-ogawa: nit. this looks mildly wild. Why not export something like `WindowRendererOptions`?

AriPerkkio: I would not expose name `Window` from summary reporter's API. It's supposed to be implementation detail. Any other naming suggestions?

AriPerkkio: Oh this is about just exporting the type, not about options structure. Now using `WindowRendererOptions` here.

hi-ogawa (approved):

PR #10362, opened 2026-05-15
feat(reporters): add fileTemplate option to JUnit reporter

### LINKED ISSUE #10341: Support fileTemplate/filePrefix alongside junit classnameTemplate

When working in pnpm monorepos with lots of packages (each with their own vitest config extending a shared base config), the default generated file attribute and classname in output junits is relative to the individual package, not the project repo. This interferes with my ability to associate GitHub CODEOWNERS files with the test file, as well as other test analytics projects.

With the following config, I can add a file attribute to my vitest junit outputs, and get the proper classname, but the file attribute is missing the appropriate prefix.

```ts

export default defineConfig({
  test: {
    reporters: [
      ['junit', {
        outputFile: './junit.xml',
        addFileAttribute: true,
        // override the classname to reflect the relative path from the root of the repo
        classnameTemplate: ({ filepath: abs }) => abs.replace(..., ""),
      }],
    ],
  },
});
```

results in a junit like:
```xml
<?xml version="1.0" encoding="UTF-8" ?>
<testsuites name="vitest tests" tests="19" failures="0" errors="0" time="0.049715176">
    <testsuite name="__tests__/my-test.vitest.ts" timestamp="2026-05-13T14:31:55.096Z" hostname="hostname" tests="1" failures="0" errors="0" skipped="0" time="0.00511053">
        <testcase classname="ts/packages/my-package/__tests__/my-test.vitest.ts" file="__tests__/my-test.vitest.ts" name="A useful test" time="0.002876578">
        </testcase>
    </testsuite>
</testsuites>
```

I would like it to be `file="ts/packages/my-package/__tests__/my-test.vitest.ts"`. Also relevant but not required would be `testsuite name="ts/packages/my-package/__tests__/my-test.vitest.ts"`

### Suggested solution

Support a `fileTemplate` or `filePrefix` to configure this file attribute, analogous to the `classnameTemplate` field.

### Alternative

Some other attribute (independent of junit reporters) indicating the location of package. [Projects](https://vitest.dev/guide/projects.html) do not support my use case because I have multiple vitest configs. Overriding `root` also does not seem to fix this (no tests are detected).

### Additional context

The Jest junit reporter supports [`filePathPrefix`](https://github.com/jest-community/jest-junit#configuration) for this issue.

I'm happy to help put up a PR for this, but I would appreciate some more guidance first.

### Validations

mixelburg:
Fixes #10341

Adds a `fileTemplate` option to the JUnit reporter, analogous to the existing `classnameTemplate`. This lets users control the value of the `file` attribute on `<testcase>` elements and the `name` attribute on `<testsuite>` elements — useful in pnpm monorepos where the default relative path may not include enough prefix for test analytics or CODEOWNERS matching.

### Usage

```ts
// string template with {filename} and {filepath} placeholders
reporters: [['junit', { fileTemplate: 'packages/{filename}/src', addFileAttribute: true }]]

// function template
reporters: [['junit', { fileTemplate: ({ filepath }) => filepath.replace(/.*fixtures\//, ''), addFileAttribute: true }]]
```

Unlike `classnameTemplate`, which only affects the `classname` attribute, `fileTemplate` affects both `file` and `testsuite name` while leaving `classname` unchanged (unless `classnameTemplate` is also set).

### DIFF
--- a/packages/vitest/src/node/reporters/junit.ts
+++ b/packages/vitest/src/node/reporters/junit.ts
@@ -17,6 +17,11 @@ interface ClassnameTemplateVariables {
   filepath: string
 }
 
+interface FileTemplateVariables {
+  filename: string
+  filepath: string
+}
+
 export interface JUnitOptions {
   outputFile?: string
 
@@ -39,6 +44,10 @@ export interface JUnitOptions {
    * Hostname to use in the report. By default, it uses os.hostname()
    */
   hostname?: string
+  /**
+   * Template for the file attribute and testsuite name. Can be either a string or a function. The string can contain placeholders {filename} and {filepath}.
+   */
+  fileTemplate?: string | ((fileVariables: FileTemplateVariables) => string)
 }
 
 function flattenTasks(task: Task, baseName = ''): Task[] {
@@ -207,9 +216,9 @@ export class JUnitReporter implements Reporter {
     })
   }
 
-  async writeTasks(tasks: Task[], filename: string): Promise<void> {
+  async writeTasks(tasks: Task[], filename: string, defaultClassname: string): Promise<void> {
     for (const task of tasks) {
-      let classname = filename
+      let classname = defaultClassname
 
       const templateVars: ClassnameTemplateVariables = {
         filename: task.file.name,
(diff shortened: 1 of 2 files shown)


sheremet-va (approved):

sheremet-va:
Seems like your branch is outdated, clanker

733998b2-6b5b-442c-864f-eda83827cdd3_text

TylerJang27:
@AriPerkkio I think this would work for me. Can you look into getting this landed?

sheremet-va:
Closing this as this AI clanker account has been known to be inactive in comments and ignores any reviews. Their account is also blocked

PR #10363, opened 2026-05-16
fix: apply cjs interop for truthy `__esModule`

### LINKED ISSUE #10359: Bun runner makes import { z } from zod undefined
### Describe the bug

When running Vitest with Bun via `bun --bun vitest`, a workspace package that imports Zod like this fails during module evaluation:

```ts
import { z } from 'zod'

export const schema = z.object({
  name: z.string(),
})
```

The error is:

```txt
TypeError: undefined is not an object (evaluating 'z.object')
```

The same Zod import works when executed directly with Bun:

```sh
bun --cwd tests/e2e -e "import { z } from 'zod'; console.log('direct z.object:', typeof z.object)"
# direct z.object: function
```

So the issue appears specific to Vitest's Bun runner/module evaluation path when importing a TypeScript workspace package.

### Reproduction

https://github.com/Mnigos/vitest-bun-zod-symlink-repro

Steps:

```sh
bun install
bun run direct
bun run test:e2e
```

Observed output:

```txt
$ bun run direct
$ bun --cwd tests/e2e -e "import { z } from 'zod'; console.log('direct z.object:', typeof z.object)"
direct z.object: function

$ bun run test:e2e
$ bun --cwd tests/e2e test
$ bun --bun vitest run --config vitest.config.ts

 RUN  v4.1.1 /path/to/vitest-bun-zod-symlink-repro/tests/e2e

 ❯ contracts.e2e.spec.ts (0 test)

 FAIL  contracts.e2e.spec.ts [ contracts.e2e.spec.ts ]
TypeError: undefined is not an object (evaluating 'z.object')
 ❯ node_modules/@repro/contracts/index.ts:3:23
      1| import { z } from 'zod'
      2|
      3| export const schema = z.object({
       |                       ^
      4|   name: z.string(),
      5| })
```

Minimal structure:

```txt
package.json
packages/contracts/package.json
packages/contracts/index.ts
tests/e2e/package.json
tests/e2e/vitest.config.ts
tests/e2e/contracts.e2e.spec.ts
```

`packages/contracts/index.ts`:

```ts
import { z } from 'zod'

export const schema = z.object({
  name: z.string(),
})

export function parseName(input: unknown) {
  return schema.parse(input).name
}
```

`tests/e2e/contracts.e2e.spec.ts`:

```ts
import { describe, expect, test } from 'vitest'
import { parseName } from '@repro/contracts'

describe('workspace contract import', () => {
  test('uses zod through a symlinked workspace package', () => {
    expect(parseName({ name: 'Ada' })).toBe('Ada')
  })
})
```

### System Info

```txt
System:
  OS: macOS 15.7.5
  CPU: (14) arm64 Apple M4 Pro
  Memory: 965.09 MB / 48.00 GB
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 24.14.1 - /opt/homebrew/bin/node
  npm: 11.11.0 - /opt/homebrew/bin/npm
  bun: 1.3.12 - /Users/mnigos/.bun/bin/bun
npmPackages:
  vitest: 4.1.1
  zod: ^4.1.13
```

### Used Package Manager

bun

### Validations

- [x] Follow our Code of Conduct
- [x] Read the Contributing Guidelines
- [x] Read the docs
- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate
- [x] Make sure this is a Vitest issue and not a framework-specific issue
- [x] Check that this is a concrete bug

hi-ogawa:
### Description

- Closes https://github.com/vitest-dev/vitest/issues/10359

As written in https://github.com/vitest-dev/vitest/issues/10359#issuecomment-4465834973, Bun defines `__esModule` export for genuine esm in a way `'__esModule' in mod === true` but `mod.__esModule === undefined`. This behavior is Bun specific quirk, but Vitest applying interop based on `__esModule` property existence seems less conventional and can be naturally mitigated on Vitest side by detecting truthy `__esModule`. That's what PR changes.

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/runtime/moduleRunner/moduleEvaluator.ts
+++ b/packages/vitest/src/runtime/moduleRunner/moduleEvaluator.ts
@@ -531,7 +531,7 @@ function interopModule(mod: any) {
 
   let defaultExport = 'default' in mod ? mod.default : mod
 
-  if (!isPrimitive(defaultExport) && '__esModule' in defaultExport) {
+  if (!isPrimitive(defaultExport) && defaultExport.__esModule) {
     mod = defaultExport
     if ('default' in defaultExport) {
       defaultExport = defaultExport.default
--- a/test/unit/deps/dep-cjs/esmodule-false.js
+++ b/test/unit/deps/dep-cjs/esmodule-false.js
@@ -1,0 +1,2 @@
+exports.__esModule = false
+exports.test = "hello"
--- a/test/unit/deps/dep-cjs/esmodule-false.js
+++ b/test/unit/deps/dep-cjs/esmodule-false.js
@@ -1,2 +1,2 @@
 exports.__esModule = false
-exports.test = "hello"
+exports.test = 'hello'
--- a/test/unit/deps/dep-cjs/esmodule.js
+++ b/test/unit/deps/dep-cjs/esmodule.js
@@ -1,2 +1,2 @@
 exports.__esModule = true
-exports.test = "hello"
+exports.test = 'hello'
--- a/test/unit/deps/dep-cjs/esmodule.js
+++ b/test/unit/deps/dep-cjs/esmodule.js
@@ -1,0 +1,2 @@
+exports.__esModule = true
+exports.test = "hello"
--- a/test/unit/deps/dep-cjs/package.json
+++ b/test/unit/deps/dep-cjs/package.json
@@ -2,6 +2,8 @@
   "name": "@vitest/test-dep-cjs",
   "type": "commonjs",
   "exports": {
+    "./esmodule": "./esmodule.js",
+    "./esmodule-false": "./esmodule-false.js",
     "./esm-comment": "./esm-comment.js",
     "./esm-string": "./esm-string.js"
   }
--- a/test/unit/test/interop.test.ts
+++ b/test/unit/test/interop.test.ts
@@ -4,6 +4,8 @@ import * as esModule from '@vitest/test-dep-cjs/esmodule'
 import * as esModuleFalse from '@vitest/test-dep-cjs/esmodule-false'
 import { expect, test } from 'vitest'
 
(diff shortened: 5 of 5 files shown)


sheremet-va:
I don’t think we should change the behavior for bun. The interop is applied because Node cannot handle it specifically. deps.interopDefault is not required for bun at all, I think. It might be better to just disable it in that runtime

hi-ogawa:
> I don’t think we should change the behavior for bun. The interop is applied because Node cannot handle it specifically. deps.interopDefault is not required for bun at all, I think. It might be better to just disable it in that runtime

Testing `'__esModule' in value` instead of `value.__esModule` doesn't look specifically intended though. I wouldn't be actually invested on Bun specific behavior to verify `interopDefault` is necessary or not, so if you are not merging this, I'll close this.

sheremet-va:
I’m not stopping the merge, I just want to discuss this first

> Testing `'__esModule' in value` instead of `value.__esModule` doesn't look specifically intended though

Why isn’t it intended? One use case I see is proxies throwing errors (I think this is why it’s “in” right now)

hi-ogawa:
Checking `m.__esModule` truthiness is the tradition as far as I understand since what babel and tsc do is something like `Object.defineProperty(exports, "__esModule", { value: true })` https://www.typescriptlang.org/docs/handbook/2/modules.html#commonjs

Like this code https://github.com/vitest-dev/vitest/blob/05a05a07b631ae9ca9b20cc9faf28ef540359b8a/packages/mocker/src/browser/mocker.ts#L80 and also some snippet in Vite also has `m.__esModule ? ... ` style.

sheremet-va:
Ok, I don't have a strong case against it, although I still believe for bun we should just change the default since it's not required in that environment.

sheremet-va (approved):

PR #10366, opened 2026-05-16
docs: fix the file extension glob pattern

ghazi-git:
Changed the invalid glob pattern `tsx?` to a valid one `ts(x)?`

### DIFF
--- a/docs/config/include.md
+++ b/docs/config/include.md
@@ -25,7 +25,7 @@ export default defineConfig({
   test: {
     include: [
       './test',
-      './**/*.{test,spec}.tsx?',
+      './**/*.{test,spec}.ts(x)?',
     ],
   },
 })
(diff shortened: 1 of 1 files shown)


sheremet-va (changes requested):
`tsx?` is a valid syntax

ghazi-git:
my bad @sheremet-va, I should have asked first, what is the intention behind the glob pattern `'./**/*.{test,spec}.tsx?'`?

if it's to match `ts` or `tsx` test files, then the glob pattern does not do that. You can see that in this [stackblitz](https://stackblitz.com/edit/vitest-dev-vitest-gewhunwa?file=vite.config.ts). Despite the fact there are 2 test files under `/test`, none of them are found by vitest.

The stackblitz was created from the [vitest.new](https://vitest.new/) template, then setting `include: ['./**/*.{test,spec}.tsx?']` in `vite.config.ts` and renaming `test/suite.test.ts` to `test/suite.test.tsx`

sheremet-va:
Ah, you are right. It is supposed to be `.ts(x)?`

sheremet-va (approved):

PR #10373, opened 2026-05-18
fix!: allow mutating happy-dom/jsdom window object

### LINKED ISSUE #10343: Setting `window.innerWidth` doesn't actually set `innerWidth` on happy-dom's `window` object
### Describe the bug

I was having unexpected test failures in some of my tests ran in happy-dom environment related to `window.matchMedia()`, after digging a bit and adding a few `console.log` statements in node_modules/happy-dom/ I discovered that the modifications to `window.innerWidth` in my tests were not reflected on the window object the window methods like `matchMedia()` were bound to. Here is the minimal reproduction to demonstrate the issue:

```typescript
test('window.matchMedia()', () => {
  window.innerWidth = 100;
  expect(window.matchMedia('(max-width: 100px)').matches).toBe(true);
});
```

adding the following at the top of this test file fixes the issue:
```typescript
import { Window } from 'happy-dom';

const window = new Window();
```
But obviously it would be desirable to not have to do this in every test file and not have to explain that it's only done due to a bug in comments.

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-xs8z4w5n?file=test%2Fbasic.test.ts

### System Info

```shell
System:
    OS: Linux 7.0 EndeavourOS
    Container: Yes
    Shell: 5.3.9 - /bin/bash
  Binaries:
    Node: 26.1.0 - /run/user/1000/fnm_multishells/135812_1778693298144/bin/node
    npm: 11.13.0 - /run/user/1000/fnm_multishells/135812_1778693298144/bin/npm
  Browsers:
    Chromium: 148.0.7778.167
    Firefox: 150.0.2
    Firefox Developer Edition: 150.0.2
  npmPackages:
    vitest: ^4.1.6 => 4.1.6
```

### Used Package Manager

npm

### Validations

hi-ogawa:
### Description

- Closes https://github.com/vitest-dev/vitest/issues/10343

Like in the issue above, dom implementation supports mutating some global window property, but they weren't propagated through jsdom/happy-dom global wrapper of Vitest. This changes so it writes out to the dom window object.

### Tests

### Documentation

### Changesets


sheremet-va:
Cannot find many issues with this approach, to be honest

packages/vitest/src/integrations/env/utils.ts
```diff
@@ -78,6 +78,15 @@ export function populateGlobal(
...
         overrideObject.set(key, v)
+        // propagate changes to underlying window implementation,
+        // which can affect other window API behavior internally, e.g.
+        // updating `innerWidth` affects `matchMedia("(max-width: *)")` on happy-dom.
+        try {
+          win[key] = v
+        }
+        catch {
+          // haven't found a case it throws but just in case
+        }
```
hi-ogawa: Discussed that we can just remove `try/catch` to surface if there's any setter error.

test/unit/test/environments/happy-dom.spec.ts
```diff
@@ -39,3 +39,28 @@ test('can pass down a simple form data', async () => {
...
+    }).not.toThrow()
+    return
+  }
+
+  expect(() => {
+    Object.assign(window, { navigator: {} })
+  }).toThrowErrorMatchingInlineSnapshot(
+    `[TypeError: Cannot set property navigator of #<GlobalWindow> which has only a getter]`,
+  )
+})
```
hi-ogawa, line 66: There's vm annoyance. Jsdom seems to mitigate this as they provide dedicated API for vm use `JSDOM.getInternalVMContext`, but happy-dom doesn't have the same. 

This minor difference sounds fine as just an edge case. Note that, `window.innerWidth = 50` behavior still works fine.

sheremet-va (changes requested):
I think this might also need a mention in the migration guide

sheremet-va (approved):

PR #10376, opened 2026-05-18
fix(browser-webdriverio): allow gpu in headless chrome

### LINKED ISSUE #10351: browser-webdriverio makes it impossible to set --enable-gpu when using headless Chrome
### Describe the bug

Right now enabling headless mode in Chrome also sets the `--disable-gpu` flag. There is no way to configure Chrome like this:

```js
{
    browser: 'chrome',
    headless: true,
    name: 'Chrome',
    provider: webdriverio({ capabilities: { 'goog:chromeOptions': { args: ['--enable-gpu'] } } })
}
```

Since the `--disable-gpu` will be added automatically when setting `headless` to `true` it will override the `--enable-gpu` flag.

https://github.com/vitest-dev/vitest/blame/d77e93659d1703f9d96b58373b38738bf190289e/packages/browser-webdriverio/src/webdriverio.ts#L185

https://github.com/vitest-dev/vitest/blame/d77e93659d1703f9d96b58373b38738bf190289e/packages/browser-webdriverio/src/webdriverio.ts#L195

It is of course possible to achieve the desired result when doing all the config via `'goog:chromeOptions'`. The following works:

```js
{
    browser: 'chrome',
    name: 'Chrome',
    provider: webdriverio({ capabilities: { 'goog:chromeOptions': { args: ['--headless', '--enable-gpu'] } } })
}
```

However I think this is not really obvious without knowing the source code. I can't find it anymore but I believe using the `--headless` flag directly was also discouraged somewhere in the docs.

Do you remember why `--disable-gpu` was hard-coded in the first place? Maybe it's possible to just drop it.

### Reproduction

As mentioned above configuring Chrome like this is impossible:

```js
{
    browser: 'chrome',
    headless: true,
    name: 'Chrome',
    provider: webdriverio({ capabilities: { 'goog:chromeOptions': { args: ['--enable-gpu'] } } })
}
```

It will not enable GPU rendering.

### System Info

```shell
npmPackages:
    @vitest/browser-webdriverio: ^4.1.5 => 4.1.5
    vitest: ^4.1.5 => 4.1.5
```

### Used Package Manager

npm

### Validations

rotempasharel1:
Resolves #10351

### Description

This PR allows users to explicitly enable GPU rendering for headless Chrome when using `@vitest/browser-webdriverio`.

By default, Vitest still adds `disable-gpu` for headless Chrome to preserve the existing behavior.

However, when the user explicitly provides `--enable-gpu` or `enable-gpu` in `goog:chromeOptions.args`, Vitest no longer adds `disable-gpu` on top of it.

### Tests

No test was added because the direct `buildCapabilities` unit test was considered too low-level during review.

The change is intentionally conservative: `disable-gpu` remains part of the default headless Chrome arguments, but it is skipped when the user explicitly provides `--enable-gpu`.

### DIFF
--- a/packages/browser-webdriverio/src/webdriverio.ts
+++ b/packages/browser-webdriverio/src/webdriverio.ts
@@ -192,7 +192,19 @@ export class WebdriverBrowserProvider implements BrowserProvider {
     if (browser !== 'safari' && options.headless) {
       const [key, args] = headlessMap[browser]
       const currentValues = (this.options?.capabilities as any)?.[key] || {}
-      const newArgs = [...(currentValues.args || []), ...args]
+      const currentArgs = currentValues.args || []
+
+      const hasEnableGpu = browser === 'chrome' && currentArgs.some((arg: string) =>
+        arg === '--enable-gpu' || arg === 'enable-gpu',
+      )
+
+      const newArgs = [
+        ...currentArgs,
+        ...args.filter(
+          arg => !(browser === 'chrome' && hasEnableGpu && arg === 'disable-gpu'),
+        ),
+      ]
+
       capabilities[key] = { ...currentValues, args: newArgs as any }
     }
 
--- a/packages/browser-webdriverio/src/webdriverio.ts
+++ b/packages/browser-webdriverio/src/webdriverio.ts
@@ -194,14 +194,11 @@ export class WebdriverBrowserProvider implements BrowserProvider {
       const currentValues = (this.options?.capabilities as any)?.[key] || {}
       const currentArgs: string[] = currentValues.args || []
 
-      let argsToAdd = [...defaultArgs]
+      const hasEnableGpu = browser === 'chrome' && (currentArgs.includes('--enable-gpu') || currentArgs.includes('enable-gpu'))
 
-      if (browser === 'chrome') {
-        const hasEnableGpu = currentArgs.includes('--enable-gpu') || currentArgs.includes('enable-gpu')
-        if (hasEnableGpu) {
-          argsToAdd = argsToAdd.filter(arg => arg !== 'disable-gpu')
-        }
-      }
+      const argsToAdd = hasEnableGpu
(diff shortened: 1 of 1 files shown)


test/unit/test/webdriverio.test.ts
```diff
@@ -0,0 +1,63 @@
+import type { TestProject } from 'vitest/node'
```
sheremet-va: This test is a bit too low level, not having a test seems better

OR preferably it should be a test somewhere in `e2e/browser-config`

sheremet-va (changes requested):

rotempasharel1:
Thanks! I removed the low-level unit test and updated the implementation to be more conservative.

`disable-gpu` remains in the default headless Chrome args, but it is skipped when the user explicitly provides `--enable-gpu` or `enable-gpu`.

CI is passing now.

packages/browser-webdriverio/src/webdriverio.ts
```diff
@@ -192,7 +192,19 @@ export class WebdriverBrowserProvider implements BrowserProvider {
...
+      const currentArgs = currentValues.args || []
+
+      const hasEnableGpu = browser === 'chrome' && currentArgs.some((arg: string) =>
+        arg === '--enable-gpu' || arg === 'enable-gpu',
+      )
+
+      const newArgs = [
+        ...currentArgs,
+        ...args.filter(
+          arg => !(browser === 'chrome' && hasEnableGpu && arg === 'disable-gpu'),
```
sheremet-va: The code works fine, but it is very hard to follow. Can you refactor it to be more readable?

sheremet-va (changes requested):

rotempasharel1:
I've sorted it out, I'd appreciate your approval.

sheremet-va (approved):

sheremet-va:
Thanks!

PR #10386, opened 2026-05-19
fix(ui): fix module graph in browser mode with --ui

hi-ogawa:
### Description

Currently we rely on caller to pass `browser: boolean` to indicate which module graph to pick, but this is redundant since `project.config.browser.enabled` is already available on server side. 

Additionally UI uses `browserState` global (i.e. browser orchestrator mode) for `browser: boolean` flag switch but that's not correct and causes missing module graph on browser mode with pure `--ui` mode such as:

```sh
pnpm -C packages/ui test:ui --ui
```

### Tests

### Documentation

### Changesets

### DIFF
--- a/test/ui/playwright.config.ts
+++ b/test/ui/playwright.config.ts
@@ -6,7 +6,7 @@ export default defineConfig({
     {
       name: 'chromium',
       // increase viewport height so virtual scroller renders all explorer items
-      use: { ...devices['Desktop Chrome'], viewport: { width: 1280, height: 900 } },
+      use: { ...devices['Desktop Chrome'], viewport: { width: 900, height: 1300 } },
     },
   ],
   use: {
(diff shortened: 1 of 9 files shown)


test/ui/playwright.config.ts
```diff
@@ -6,7 +6,7 @@ export default defineConfig({
     {
       name: 'chromium',
       // increase viewport height so virtual scroller renders all explorer items
-      use: { ...devices['Desktop Chrome'], viewport: { width: 1280, height: 900 } },
+      use: { ...devices['Desktop Chrome'], viewport: { width: 800, height: 1300 } },
```
hi-ogawa, line 9: Made this taller again, so the explore item assertion becomes simpler by default. Added the specific test case to exercise virtual scroll + filtering in `filter reveals initially invisible explorer item`.

sheremet-va (approved):
Nice catch. This was required pre v4 because you could have both node and browser tests, but now it's simplified a lot.

PR #10397, opened 2026-05-19
fix(browser): wait for orchestrator readiness before resolving browser sessions

soconnor-seeq:
### Description

I diagnosed an intermittent failure in our CI runs (about one in five) to a timing issue in vitest. The tests would not fail — vitest would hang until the CI job was killed after a while.

Resolves (do you need an issue for a bug fix?)

 I figured out (with the help of Codex) hanging behaviour was caused when the WS upgrade happened (so the `connected()` event was triggered on the session), but the `orchestrator` wasn't created and assigned to `getBrowserState().orchestrator` yet.

The fix was to introduce a "ready" event in the `IframeOrchestrator`'s `constructor` to signal that the websocket connection is ready. In the session only after both the ready + connected events are seen does the promise returned from `createSession` resolve.

### Tests

### Documentation

### Changesets

AI assisted: Codex

### DIFF
--- a/packages/browser/src/client/orchestrator.ts
+++ b/packages/browser/src/client/orchestrator.ts
@@ -36,6 +36,13 @@ export class IframeOrchestrator {
       'message',
       e => this.onGlobalChannelEvent(e),
     )
+
+    // Notify the server once the websocket is ready without blocking orchestrator creation.
+    void client.waitForConnection()
+      .then(() => client.rpc.onOrchestratorReady())
+      .catch((error) => {
+        debug('failed to notify orchestrator readiness', error)
+      })
   }
 
   public async createTesters(options: BrowserTesterOptions): Promise<void> {
--- a/packages/browser/src/node/rpc.ts
+++ b/packages/browser/src/node/rpc.ts
@@ -69,7 +69,8 @@ export function setupBrowserRpc(globalServer: ParentBrowserProject, defaultMocke
 
     if (type === 'orchestrator') {
       const session = sessions.getSession(sessionId)
-      // it's possible the session was already resolved by the preview provider
+      // it's possible the session was already resolved by the preview provider,
+      // but we still mark the websocket connection when the page reconnects
       session?.connected()
     }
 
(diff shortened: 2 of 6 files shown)


soconnor-seeq:
Notes:
>  Ideally, include a test that fails without this PR but passes with it.

I can't (or couldn't think of how to) test this bug as it is a fairly specific timing issue. It would require somehow mocking Playwright, etc so that the connected event fired before / after the `orchestrator` being created + assigned to `getBrowserState().orchestrator`. I added unit test that simulate the behaviour in the session code.

> It's really useful if your PR references an issue where it is discussed ahead of time. If the feature is substantial or introduces breaking changes without a discussion, PR might be closed.

Because this is just a bug fix is an issue needed?

---

I was also able to reproduce this (intermittently) on my local machine by using Linux's `ulimit -n 302 && pnpm test` (that 302 is probably very repo + OS specific mind you). No other resource pressure seemed to work to re-create it: `nice`, `ionice`, `tasket`, 10 concurrent runs, cgroup with CPU and memory limits, etc.

sheremet-va:
Did you try the fixed version? Does it work for you?

soconnor-seeq:
> Did you try the fixed version? Does it work for you?

If by "try the fixed version" you mean the latest beta release of `v5.0.0-beta.3` then yes I did finally get around to testing that beta version.

I tried upgrading to `v5.0.0-beta.3` but it still hangs. It looks like the source of the hang is different then what this fix is solving though. It's likely caused by this sequence of events:

1. `orchestrator.ts` sends `prepare` as soon as `iframe.onload` fires
2. `tester.ts` only starts listening for `BroadcastChannel` messages after its module finishes evaluating
3. `iframe.onload` can happen before that listener exists. So the first `prepare` message can be lost then the orchestrator waits forever for `response:prepare`.

I'm going to create a new PR for the above race condition = hanging `vitest` bug I found. But I still think there this PR's fix for a different race condition could still surface again w/o this it also being fixed.

sheremet-va:
I meant try the fix that you implemented here. It is obviously not merged, so testing a beta version is meaningless for your use case. You can use the link from the pkg.pr.new or manually applying a fix

soconnor-seeq:
> I meant try the fix that you implemented here.

Ah gotcha — now I see what you are saying. I didn't know that there were built versions per-PR for testing.

Originally I noticed this bug (CI hanging w/ `vitest`) using v4.1.2. I then figured out where the bug was by altering the v4.1.2 version locally by adding logging and re-running CI till a hanging-failure surfaced again.

After figuring out where in code the hanging was coming from I made these changes on the v5 branch — built it locally — and tested that fixed-with-these-changes `vitest` in CI and it passed†

† but sometimes it did fail but I then narrowed it down to [another hanging scenario I detailed in my comment above](https://github.com/vitest-dev/vitest/pull/10397#issuecomment-4558968521).

sheremet-va (approved):

PR #10398, opened 2026-05-19
docs(browser): remove stale TODO in trace-view

casper-lindh:
The HTML comment in `docs/guide/browser/trace-view.md` (lines 62-71) suggests adding a `browser.ui` / `browser.headless` configuration table, but [the table directly below](https://github.com/vitest-dev/vitest/blob/main/docs/guide/browser/trace-view.md#L75-L80) (the "Common Setups" section) already covers those mode combinations using the actual CLI flags.

The note is invisible in the rendered page, so removing it.

### DIFF
--- a/docs/guide/browser/trace-view.md
+++ b/docs/guide/browser/trace-view.md
@@ -59,17 +59,6 @@ Selecting a step also opens its source location in the Editor tab when that loca
 
 ## Common Setups
 
-<!--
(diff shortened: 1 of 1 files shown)


hi-ogawa:
No, this TODO is a different aspect and implying that this "gotcha" belongs to somewhere general places on browser mode doc.

casper-lindh:
thanks for the explanation

PR #10408, opened 2026-05-20
fix(runner): propagate describe.for chainable context

### LINKED ISSUE #10407: describe.skipIf(true).for(cases) does not skip — chained modes lost on `.for`
### Describe the bug

`describe.skipIf(true).for(cases)` runs the parametrized suite instead of skipping it. The same is true for any chained mode on `describe.for` (`.skip.for`, `.only.for`, `.concurrent.for`, `.sequential.for`, `.shuffle.for`, `.todo.for`, `.runIf(...).for`): the chained context is dropped and the inner suites are registered as plain `run`-mode suites.

`describe.skipIf(true).each(cases)` and plain `describe.skipIf(true)(...)` both work as expected — the bug is specific to `.for`.

### Expected

All three suites are skipped — `4 skipped`, `0 failed`.

### Actual (vitest 4.1.6)

```
 × describe.skipIf(true).for — case 1 > should be skipped but runs
 × describe.skipIf(true).for — case 2 > should be skipped but runs
 × describe.skipIf(true).for — case 3 > should be skipped but runs
 ↓ describe.skipIf(true).each — case 1 > correctly skipped
 ↓ describe.skipIf(true).each — case 2 > correctly skipped
 ↓ describe.skipIf(true).each — case 3 > correctly skipped
 ↓ describe.skipIf(true) — plain > correctly skipped

 Test Files  1 failed (1)
      Tests  3 failed | 4 skipped (7)
```

### Root cause

In `packages/runner/src/suite.ts` → `createSuite()`, `suiteFn.for` captures the outer `suite` (the unmodified module-level chainable) instead of resolving it from `this`. That means the `{ skip: true }` (or `only`, `concurrent`, …) context that `.skipIf` / the chain getters set on the chainable is silently discarded by the time `.for` registers the suites.

```js
// dist/chunk-artifact.js, createSuite()
suiteFn.for = function(cases, ...args) {
    // (no getChainableContext(this) here)
    return (name, optionsOrFn, fnOrOptions) => {
        // ...
        cases.forEach((item, idx) => {
            suite(formatTitle(name_, toArray(item), idx), options, ...);
            //  ↑ closure-bound module-level `suite`, NOT the chained one
        });
    };
};
suiteFn.skipIf = (condition) => condition ? suite.skip : suite;
```

`suiteFn.each`, just above, already does the right thing:

```js
suiteFn.each = function(cases, ...args) {
    const context = getChainableContext(this);
    const suite = context.withContext();   // chained suite
    // ...
    suite(formatTitle(...), ...);          // respects skip/only/concurrent/etc.
};
```

### Suggested fix

Mirror `.each`’s pattern in `.for`:

```diff
 suiteFn.for = function(cases, ...args) {
+    const context = getChainableContext(this);
+    const suite = context.withContext();
     if (Array.isArray(cases) && args.length) {
         cases = formatTemplateString(cases, args);
     }
     return (name, optionsOrFn, fnOrOptions) => {
         const name_ = formatName(name);
         const { options, handler } = parseArguments(optionsOrFn, fnOrOptions);
         cases.forEach((item, idx) => {
             suite(formatTitle(name_, toArray(item), idx), options, handler ? () => handler(item) : undefined);
         });
     };
 };
```

Happy to send a PR if useful.

### Reproduction

Minimal repro: https://github.com/rhwinter/vite-skipIfError/

```ts

const SKIP = true;
const cases = [1, 2, 3];

describe.skipIf(SKIP).for(cases)('describe.skipIf(true).for — case %s', (n) => {
  test('should be skipped but runs', () => {
    expect(n).toBe(-1);
  });
});

describe.skipIf(SKIP).each(cases)('describe.skipIf(true).each — case %s', (n) => {
  test('correctly skipped', () => {
    expect(n).toBe(-1);
  });
});

describe.skipIf(SKIP)('describe.skipIf(true) — plain', () => {
  test('correctly skipped', () => {
    expect(1).toBe(-1);
  });
});
```

```
npm install
npm test
```

### System Info

```shell
System:
    OS: macOS 14.4
    CPU: (11) arm64 Apple M3 Pro
    Memory: 9.12 GB / 36.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.8.0 - /Users/robertowinter/.local/state/fnm_multishells/5987_1779203637780/bin/node
    Yarn: 1.22.22 - /Users/robertowinter/.local/state/fnm_multishells/5987_1779203637780/bin/yarn
    npm: 11.6.0 - /Users/robertowinter/.local/state/fnm_multishells/5987_1779203637780/bin/npm
    pnpm: 11.1.3 - /Users/robertowinter/.local/state/fnm_multishells/5987_1779203637780/bin/pnpm
    bun: 1.3.10 - /Users/robertowinter/.bun/bin/bun
  Browsers:
    Brave Browser: 147.1.89.132
    Chrome: 148.0.7778.179
    Chrome Canary: 150.0.7846.0
    Firefox Developer Edition: 137.0
    Safari: 17.4
    Safari Technology Preview: 17.4
```

### Used Package Manager

npm

### Validations

dicnunz:
## Summary
- Backports the chainable context propagation used on main for `describe.for` to the `v4.1` branch
- Preserves `skipIf` / `runIf` flags when `describe` is chained with `.for(cases)`
- Adds regression coverage for skipped `describe.for` and false `runIf(...).for` suites

Fixes #10407.
Backport of the runner-side behavior from #10187 for `v4.1`.

## Tests
- `pnpm exec eslint packages/runner/src/suite.ts test/core/test/test-for-suite.test.ts`
- from `test/core`: `pnpm vitest --project threads test/test-for-suite.test.ts`

### DIFF
--- a/packages/runner/src/suite.ts
+++ b/packages/runner/src/suite.ts
@@ -706,13 +706,13 @@ function createSuite() {
   }
 
   suiteFn.for = function <T>(
-    this: {
-      withContext: () => SuiteAPI
-      setContext: (key: string, value: boolean | undefined) => SuiteAPI
-    },
+    this: SuiteAPI,
     cases: ReadonlyArray<T>,
     ...args: any[]
   ) {
+    const context = getChainableContext(this)
+    const suite = context.withContext()
+
     if (Array.isArray(cases) && args.length) {
       cases = formatTemplateString(cases, args)
     }
--- a/test/core/test/test-for-suite.test.ts
+++ b/test/core/test/test-for-suite.test.ts
@@ -33,3 +33,15 @@ describe.for([
     expect(a + b).matchSnapshot()
   })
 })
+
+describe.skipIf(true).for(['skipped'])('%s describe.for suite', () => {
+  test('does not run skipped describe.for cases', () => {
+    throw new Error('skipped describe.for case should not run')
+  })
+})
+
+describe.runIf(false).for(['skipped'])('%s runIf describe.for suite', () => {
+  test('does not run false runIf describe.for cases', () => {
+    throw new Error('runIf(false) describe.for case should not run')
+  })
+})


dicnunz:
Local verification is clean on this branch:

- `pnpm exec eslint packages/runner/src/suite.ts test/core/test/test-for-suite.test.ts`
- from `test/core`: `pnpm vitest --project threads test/test-for-suite.test.ts`

The current red CI jobs appear to fail before checkout due repository workflow policy, not this patch: GitHub reports `actions/checkout@v6` is not allowed because actions must be pinned to a full-length SHA.

PR #10430, opened 2026-05-22
feat(browser)!: enable `locators.exact` by default

sheremet-va:
Related https://github.com/vitest-dev/vitest/discussions/9664#discussioncomment-16122733

### DIFF
--- a/docs/config/browser/locators.md
+++ b/docs/config/browser/locators.md
@@ -14,16 +14,16 @@ Options for built-in [browser locators](/api/browser/locators).
 
 Attribute used to find elements with `getByTestId` locator.
 
-## browser.locators.exact <Version type="experimental">4.1.3</Version> {#browser-locators-exact}
+## browser.locators.exact
 
 - **Type:** `boolean`
-- **Default:** `false`
+- **Default:** `true`
 
-When set to `true`, [locators](/api/browser/locators) will match text exactly by default, requiring a full, case-sensitive match. Individual locator calls can override this default via their own `exact` option.
+When set to `true`, [locators](/api/browser/locators) match text exactly by default, requiring a full, case-sensitive match. Individual locator calls can override this default via their own `exact` option.
 
 ```ts
-// With exact: false (default), this matches "Hello, World!", "Say Hello, World", etc.
-// With exact: true, this only matches the string "Hello, World" exactly.
+// With exact: true (default), this only matches the string "Hello, World" exactly.
+// With exact: false, this matches "Hello, World!", "Say Hello, World", etc.
 const locator = page.getByText('Hello, World', { exact: true })
 await locator.click()
 ```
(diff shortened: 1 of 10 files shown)


test/browser/specs/trace.test.ts
```diff
@@ -1489,8 +1489,8 @@ test('trace view artifacts', async () => {
                 "entries": [
                   {
                     "element": {
-                      "_pwSelector": "internal:label="Focused pseudo state"i",
-                      "locator": "getByLabel('Focused pseudo state')",
+                      "_pwSelector": "internal:role=textbox[name="Focused pseudo state"s]",
```
sheremet-va, line 1492: This is because `_pwSelector` in webdriverio is now inherited correctly (a separate bugfix)

test/browser/specs/trace.test.ts
```diff
@@ -275,8 +275,8 @@ test('trace view artifacts', async () => {
                 "entries": [
                   {
                     "element": {
-                      "_pwSelector": "internal:role=button[name="Hello"i]",
-                      "locator": "getByRole('button', { name: 'Hello' })",
```
macarie, line 279: Is it expected that the name has been dropped here and below?

docs/config/browser/locators.md
```diff
@@ -14,16 +14,16 @@ Options for built-in [browser locators](/api/browser/locators).
 
 Attribute used to find elements with `getByTestId` locator.
 
-## browser.locators.exact <Version type="experimental">4.1.3</Version> {#browser-locators-exact}
```
macarie, line 17: Are we not keeping the version tags around when removing the experimental flag?

test/browser/specs/trace.test.ts
sheremet-va: Yes, I explained it here - https://github.com/vitest-dev/vitest/pull/10430#discussion_r3288874454 If you look at the source, it just has `getByRole('button')`

docs/config/browser/locators.md
sheremet-va: Generally speaking we remove version tags with the new major because there are v4 and v3, so on docs available, it's not about the experimental status

macarie (dismissed):

packages/browser-webdriverio/src/locators.ts
```diff
@@ -32,8 +32,11 @@ class WebdriverIOLocator extends Locator {
   }
 
   // This exists to avoid calling `this.elements` in `this.selector`'s getter in interactive actions
-  private withElement(element: Element, error: Error | undefined) {
-    const pwSelector = selectorEngine.generateSelectorSimple(element)
+  private withElement(
+    element: Element,
+    error: Error | undefined,
+    pwSelector = selectorEngine.generateSelectorSimple(element),
```
hi-ogawa: The 3rd argument is always passed now. What's the default for?

hi-ogawa (approved):

PR #10439, opened 2026-05-23
fix(ui): preserve about:blank URL in trace iframe

### LINKED ISSUE #10248: Trace view iframe sandbox script warnings on chrome
- Part of https://github.com/vitest-dev/vitest/issues/10156

I just dug in and the issue seems more inherent in rrweb than our integration level quirks. Here is my tentative summary:

## Problem

Opening the browser trace view can produce console warnings like:

```text
Blocked script execution in 'about:blank' because the document's frame is sandboxed and the 'allow-scripts' permission is not set.
Blocked script execution in 'http://127.0.0.1:51204/__vitest__/' because the document's frame is sandboxed and the 'allow-scripts' permission is not set.
```

This is visible when running the UI for browser tests, for example:

```bash
pnpm -C packages/ui test:ui --browser.headless --ui
```

The warnings are confusing for users because the trace iframe appears to be trying to execute scripts from both `about:blank` and the Vitest UI URL (`/__vitest__/`). The second warning especially looks like the iframe navigated to the UI app itself.

rrweb's own public replay demo appears to show the same class of warning. Opening https://www.rrweb.io/replay logs sandboxed script execution warnings for both `about:blank` and the replay page URL. That suggests Vitest is inheriting an rrweb replay-model issue rather than introducing a completely unique behavior, but Vitest trace view can still provide a cleaner replay container.

## Current Behavior

Trace view renders the replay snapshot into a sandboxed iframe without a `src`:

**Location:** [../../packages/ui/client/components/trace/TraceView.vue#L187](../../packages/ui/client/components/trace/TraceView.vue#L187)

```html
<iframe
  v-if="selectedStep"
  ref="iframeEl"
  :key="iframeSandbox"
  :sandbox="iframeSandbox"
/>
```

The sandbox normally includes `allow-same-origin` but not `allow-scripts`:

**Location:** [../../packages/ui/client/components/trace/TraceView.vue#L28](../../packages/ui/client/components/trace/TraceView.vue#L28)

TraceView currently also resets the iframe document before rrweb rebuild:

**Location:** [../../packages/ui/client/components/trace/TraceView.vue#L49](../../packages/ui/client/components/trace/TraceView.vue#L49)

```ts
const doc = iframe.contentDocument!
doc.open()
doc.close()
const mirror = createMirror()
rebuild(serialized, {
  doc,
  cache: createCache(),
  mirror,
})
```

However, rrweb-snapshot itself performs the same reset when rebuilding a serialized `Document`:

```ts
if (n.type === NodeType.Document) {
  doc.close()
  doc.open()
  // maybe doc.write(...) for BackCompat
  node = doc
}
```

Since Vitest snapshots are captured with `snapshot(document, ...)`, the serialized root is expected to be a `Document`.

## Current Understanding

An iframe without `src` starts with an `about:blank` document. That is expected.

The misleading `/__vitest__/` document URL is likely caused by `document.open()`. When called from the parent Vitest UI script on the iframe document, browsers can update the iframe document URL to the entry document URL. That makes DevTools show the iframe document as `http://127.0.0.1:51204/__vitest__/`, even though this is not necessarily a normal network navigation to the UI page.

rrweb's own demo appears to exhibit the same behavior: sandboxed script warnings are logged for both `about:blank` and the embedding replay page URL. So this is not unique to Vitest, but Vitest can still do better because trace view controls the replay container.

## Why Not Just Add `allow-scripts`

Adding `allow-scripts` would silence the warnings, but it changes the security model. The replayed page DOM is user/app HTML captured from the browser test. Running scripts from replayed snapshots is not required for normal trace view and should remain disabled by default.

Vitest currently only enables `allow-scripts` when `recordCanvas` is enabled, because rrweb's canvas replay path relies on script execution for image loading/drawing.

## High-Level Fix Direction

Vitest should avoid relying on `document.open()` for trace view replay.

The better model is:

1. Vitest prepares a clean, same-origin iframe document explicitly.
2. rrweb-snapshot rebuild is told that the target document is already prepared.
3. The serialized `Document` node is mapped to the existing target `doc` without calling `doc.open()`.
4. Rebuild appends the serialized document children into that clean document.

In other words, for prepared documents, rrweb-snapshot's `Document` branch should behave closer to:

```ts
if (n.type === NodeType.Document) {
  node = doc
}
```

instead of:

```ts
doc.close()
doc.open()
node = doc
```

This probably needs a small rrweb-snapshot patch or upstream option, for example:

```ts
rebuild(serialized, {
  doc,
  cache,
  mirror,
  resetDocument: false,
})
```

or:

```ts
rebuild(serialized, {
  doc,
  cache,
  mirror,
  assumeCleanDocument: true,
})
```

## Preparing a Clean Document

The main open question is how Vitest should create a clean document with the correct compat mode.

Possible direction:

- Point the iframe at a Vitest-controlled, same-origin empty page instead of relying on the browser's implicit `about:blank`.
- Use a standards-mode empty page for normal snapshots.
- Use a quirks-mode empty page when the serialized snapshot has `compatMode === 'BackCompat'`.
- Recreate or renavigate the iframe for each selected snapshot so the document starts clean before rebuild.

This would also make DevTools show a purposeful trace-frame URL instead of `about:blank` or `/__vitest__/`.

Potential endpoint names:

```text
/__vitest__/__trace_frame__/standards.html
/__vitest__/__trace_frame__/quirks.html
```

or a single endpoint with a mode query:

```text
/__vitest__/__trace_frame__?compat=css1
/__vitest__/__trace_frame__?compat=back
```

## Immediate Cleanup

Vitest's local `doc.open(); doc.close()` before `rebuild()` is redundant with rrweb-snapshot's internal reset and should be removed as part of this work.

That removal alone is not the full fix. It only removes one extra reset. The remaining warning and URL behavior will continue as long as rrweb-snapshot itself calls `doc.open()`.

## Non-Goals

- Do not add `allow-scripts` just to suppress console warnings.
- Do not accept rrweb's current behavior as the desired final state.
- Do not treat this as a quick console-noise cleanup only; the goal is a clearer and safer trace replay document model.

## Open Questions

- Can rrweb-snapshot rebuild append a serialized `DocumentType` and `html` node into an already clean iframe document without browser-specific issues?
- What is the most reliable way to create a clean quirks-mode iframe document without using `document.open()` from the parent?
- Should trace view recreate the iframe per selected step, or can it safely reuse one iframe by navigating it to the controlled blank endpoint before each rebuild?
- Do nested iframes in captured snapshots need the same prepared-document strategy?
- Should this be solved as a local `patches/rrweb-snapshot.patch` change first, then proposed upstream?

ShaharAviram1:
### Description

Fixes #10248.

The trace iframe is supposed to stay on about:blank, but rebuilding the rrweb document was causing the iframe URL to switch to the parent /__vitest__/ URL instead.

This removes the extra doc.open() / doc.close() call in TraceView.vue and updates the rrweb snapshot patch to avoid using doc.open() for standards-mode documents, while keeping the existing BackCompat behavior the same.

Also adds a regression test for the iframe behavior.

- [x] Please, don't make changes to pnpm-lock.yaml unless you introduce a new test example.
- [x] Please check Allow edits by maintainers to make review process faster. Note that this option is not available for repositories that are owned by Github organizations.

### Tests
- [x] Relevant trace tests were run locally
- [ ] Run the tests with pnpm test:ci.

### Documentation
- [x] No documentation changes needed

### Changesets
- [x] PR title uses fix:

### DIFF
--- a/packages/ui/client/components/trace/TraceView.spec.ts
+++ b/packages/ui/client/components/trace/TraceView.spec.ts
@@ -1,0 +1,53 @@
+import { afterEach, beforeEach, describe, expect, it } from 'vitest'
+import { createCache, createMirror, rebuild, snapshot } from 'rrweb-snapshot'
+
+// Verifies that rebuilding an rrweb snapshot into a sandboxed iframe does not
+// change the iframe document's URL. The bug: document.open() — called either
+// from TraceView.vue directly or from inside rrweb-snapshot's buildNodeWithSN —
+// makes Chrome update the iframe document URL to the parent page's URL, which
+// produces confusing "Blocked script execution in '/__vitest__/'" sandbox warnings.
+
+describe('TraceView iframe sandbox', () => {
+  let targetIframe: HTMLIFrameElement
+  let sourceIframe: HTMLIFrameElement
+
+  beforeEach(() => {
+    // Use a standards-mode document so rrweb does not serialize
+    // compatMode: 'BackCompat', which intentionally still uses doc.open().
+    sourceIframe = document.createElement('iframe')
+    document.body.appendChild(sourceIframe)
+    const srcDoc = sourceIframe.contentDocument!
+    srcDoc.open()
+    srcDoc.write('<!DOCTYPE html><html><head></head><body></body></html>')
+    srcDoc.close()
+
+    // Target: a sandboxed iframe that mirrors TraceView's setup (allow-same-origin, no allow-scripts).
+    targetIframe = document.createElement('iframe')
+    targetIframe.setAttribute('sandbox', 'allow-same-origin')
+    document.body.appendChild(targetIframe)
+  })
+
+  afterEach(() => {
+    sourceIframe.remove()
+    targetIframe.remove()
+  })
+
+  it('does not change contentDocument.URL after rebuild', () => {
+    const serialized = snapshot(sourceIframe.contentDocument!)
+    expect(serialized, 'snapshot() must succeed on a blank document').not.toBeNull()
+
+    const doc = targetIframe.contentDocument!
+    expect(doc.URL).toBe('about:blank')
+
+    rebuild(serialized!, {
+      doc,
+      cache: createCache(),
+      mirror: createMirror(),
+    })
+
+    // If doc.open() was called (either locally or inside rrweb-snapshot),
+    // Chrome changes the iframe document URL to the parent page URL.
+    // The correct value after rebuild is still 'about:blank'.
+    expect(targetIframe.contentDocument!.URL).toBe('about:blank')
+  })
+})
--- a/packages/ui/client/components/trace/TraceView.spec.ts
+++ b/packages/ui/client/components/trace/TraceView.spec.ts
@@ -1,5 +1,5 @@
-import { afterEach, beforeEach, describe, expect, it } from 'vitest'
 import { createCache, createMirror, rebuild, snapshot } from 'rrweb-snapshot'
+import { afterEach, beforeEach, describe, expect, it } from 'vitest'
 
 // Verifies that rebuilding an rrweb snapshot into a sandboxed iframe does not
 // change the iframe document's URL. The bug: document.open() — called either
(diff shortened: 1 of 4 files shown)


ShaharAviram1:
I did use Claude a bit while investigating the issue and tracing the iframe URL change through the rrweb rebuild flow, but I went through the implementation and testing myself before opening the PR

hi-ogawa:
> The trace iframe is supposed to stay on about:blank, but rebuilding the rrweb document was causing the iframe URL to switch to the parent /**vitest**/ URL instead.

That's not the issue. The issue is iframe sandbox script warning. Please help us on confirming the behavior on upstream as indicated in https://github.com/vitest-dev/vitest/issues/10248#issuecomment-4367602203

PR #10458, opened 2026-05-26
ci: use minimal reporter in this repo in CI

sheremet-va:
It is annoying to scroll down for half a minute when an error happens.


AriPerkkio (dismissed):
🤩

PR #10466, opened 2026-05-27
fix(coverage): fail fast when `coverage.reportsDirectory` conflicts between concurrent runs 

jgamaraalv:
## Description

Two or more `vitest run --coverage` processes pointed at the same working directory (the default `coverage.reportsDirectory`) delete each other's reports. Both `clean()` (start of a run) and `cleanAfterRun()` (end of a run) operate on the shared `coverage/` directory, so this surfaced intermittently as a raw `ENOENT: lstat coverage/.tmp` crash — and even when it didn't crash, the runs silently clobbered each other's final reports.

### What changed

The first approach in this PR isolated each run's temp directory (`.tmp-<nanoid>`). As raised in review, that stopped the crash but was **not enough**: the final reports still share `coverage/`, so concurrent same-directory runs keep destroying each other's results.

So instead of isolating, this PR makes concurrent same-directory runs **fail fast with an actionable error**:

- `BaseCoverageProvider` now acquires a **cross-process lock** on `coverage.reportsDirectory` for the duration of the run. The lock lives in the OS temp directory, keyed by the resolved reports directory, so it is not wiped by the directory cleanup it guards.
- If a second live process targets the same reports directory, it fails immediately with an error pointing the user to a unique `--coverage.reportsDirectory`, instead of crashing or losing reports.
- The lock is **re-entrant** for the owning process (watch-mode reruns).
- **Stale locks** left by processes that no longer exist are reclaimed automatically (PID liveness check).
- The temp directory name is restored to the deterministic `.tmp[-shard]`.

The fix lives entirely in `BaseCoverageProvider`, so both the `v8` and `istanbul` providers inherit it.

Running coverage for several processes in parallel is still done the supported way — giving each run its own `coverage.reportsDirectory` — which is exactly what the new error message tells users to do.

Related to #10111.

---

## Please don't delete this checklist!

Before submitting the PR, please make sure you do the following:

- [x] It's really useful if your PR references an issue where it is discussed ahead of time.
  - References #10111 (related)

  - Unit/regression tests in `test/coverage-test/test/temporary-files.unit.test.ts`

  - Not touched

- [x] Please check **Allow edits by maintainers**

---

## Tests

- [x] Run the tests with `pnpm test:ci`
- [x] Unit tests cover: lock acquire/release, re-entrancy (watch reruns), fail-fast when another **live** process holds the lock, and automatic reclaim of a **stale** lock.

---

## Documentation

- [x] N/A — bug fix. The new error message is self-documenting and points to `coverage.reportsDirectory`.

---

## Changesets

  - fix(coverage): fail fast when concurrent runs share coverage.reportsDirectory


jgamaraalv (commented):
Confirmo que sou um humano e executei todos os passos requisitados, inclusive a issue em questão.

Na verdade, eu mesmo enfrentei o problema corrigido e quando fui verificar já existia uma issue sobre isso relacionada ao problema que enfrentei.

jgamaraalv:
My problem before saw issue #10111: 

# v8 coverage: concurrent runs in the same project dir crash with ENOENT lstat coverage/.tmp

- Environment: vitest 4.0.18, @vitest/coverage-v8 4.0.18, Node 22.14.0, pnpm 10.30.2, macOS (Darwin 25.3.0)
- What happens: Running two or more `vitest run --coverage processes concurrently` in the same working directory intermittently fails with: `Error: ENOENT: no such file or directory, lstat '<cwd>/coverage/.tmp'
{ errno: -2, code: 'ENOENT', syscall: 'lstat', path: '<cwd>/coverage/.tmp' }`
- One process's coverage cleanup/merge empties coverage/.tmp while another process is still reading it. A single run is always clean.

## Repro: 
In any project with coverage.provider: 'v8' and default reportsDirectory:
`rm -rf coverage`
`for i in 1 2 3; do (vitest run --coverage &) ; done; wait   # ENOENT on coverage/.tmp`
vs. sequential: 
`vitest run --coverage  → always exit 0`

## Expected: 
concurrent same-dir runs either isolate their tmp dir per process or fail with a clear "coverage dir already in use" message, instead of an unhandled `ENOENT. Workaround: give each run a unique --coverage.reportsDirectory (e.g. coverage-$PPID) or serialize the runs.`

jgamaraalv:
> Hello @jgamaraalv. Your PR has been labeled `maybe automated` because it appears to have been fully generated by AI with no human involvement.
> 
> 
> 
> 

- Confirmo que sou um humano.
- Li, revisei, testei e retestei tudo, da issue originária até a solução final.
- Li toda a issue após enfrentar o mesmo problema e ver que já existia alguém que passou por algo semelhante.
- Segui todas as diretrizes de contribuição e o template.
- Utilizei o Claude Code para desenvolvimento assistido por IA.

AriPerkkio (changes requested):
> Two or more `vitest run --coverage` processes running in the same working directory

Each of them will also remove `./coverage` during start, removing each others possible results. This fix is not enough.

Proper fix is from user side to use `vitest run --coverage --coverage.reportsDirectory=some-unique-path`.

jgamaraalv:
> > Two or more `vitest run --coverage` processes running in the same working directory
> 
> Each of them will also remove `./coverage` during start, removing each others possible results. This fix is not enough.
> 
> Proper fix is from user side to use `vitest run --coverage --coverage.reportsDirectory=some-unique-path`.


@AriPerkkio  You're right — isolating only the temp dir wasn't enough. Even with separate .tmp dirs, both runs still write their final reports into the same coverage/  directory and clean() wipes the other run's results, so the data loss stayed; I had only hidden the ENOENT crash.

I pushed a different approach. Instead of isolating, Vitest now takes a lock on coverage.reportsDirectory (stored in the OS temp dir, keyed by the resolved path) for the duration of the run. 

If a second process targets the same directory while another is still running, it fails right away with an actionable error telling the user to give each run its own --coverage.reportsDirectory, which is exactly the fix you described, just surfaced by Vitest instead of an intermittent  crash. 

The lock is re-entrant for the same process (watch reruns) and stale locks from dead processes are reclaimed automatically. The temp dir name is back to the deterministic .tmp[-shard].

So this doesn't try to make same-directory concurrent runs "work" — it makes them fail fast with a clear message and points to the unique-directory fix.

If you'd prefer not to add this and just document the unique reportsDirectory requirement instead let me know.

AriPerkkio:
> [...] Vitest now takes a lock on coverage.reportsDirectory (stored in the OS temp dir, keyed by the resolved path) for the duration of the run.
> 
> If a second process targets the same directory while another is still running, it fails right away with an actionable error telling the user to give each run its own --coverage.reportsDirectory, which is exactly the fix you described, just surfaced by Vitest instead of an intermittent crash.

I like this approach, but need some time to think through it before commenting about the implementation. 👍 

Maybe the lockfiles could be stored in [`.vitest/lockfiles`](https://main.vitest.dev/guide/advanced/reporters.html#storing-artifacts-on-file-system) instead of in `tmpdir()`. We might want similar solution for other reporters too, not sure yet.

packages/vitest/src/node/coverage.ts
```diff
@@ -74,6 +76,16 @@ export async function getCoverageProvider(
   return null
 }
 
+function isProcessAlive(pid: number): boolean {
+  try {
+    process.kill(pid, 0)
```
AriPerkkio: This looks so weird that we need some comment above it:


```suggestion
    // Sending signal 0 checks if process exists without actually killing it: https://nodejs.org/api/process.html#processkillpid-signal
    process.kill(pid, 0)
```

packages/vitest/src/node/coverage.ts
```diff
@@ -262,6 +276,84 @@ export class BaseCoverageProvider {
...
+
+  private async acquireReportsDirectoryLock(): Promise<void> {
+    const lockFile = this.reportsDirectoryLockFile
+    const payload = JSON.stringify({
+      pid: process.pid,
+      reportsDirectory: resolve(this.options.reportsDirectory),
+      timestamp: Date.now(),
+    })
+
+    for (let attempt = 0; attempt < 10; attempt++) {
```
AriPerkkio: Why do we need to loop this 10 times here? Isn't once enough to know it's in use?

test/coverage-test/test/temporary-files.unit.test.ts
```diff
@@ -1,7 +1,34 @@
...
+import { afterEach, expect, test } from 'vitest'
 import { BaseCoverageProvider } from 'vitest/node'
 
+const cleanups: Array<() => void> = []
+
+afterEach(() => {
+  while (cleanups.length) {
+    cleanups.pop()!()
+  }
+})
```
AriPerkkio: Let's use `import { onTestFinished } from 'vitest'` instead of manual `cleanups` array.

packages/vitest/src/node/coverage.ts
```diff
@@ -262,6 +276,84 @@ export class BaseCoverageProvider {
...
+        + `in the same directory at the same time is not supported, because they would delete each `
+        + `other's reports.\nGive each run its own "coverage.reportsDirectory" `
+        + `(e.g. --coverage.reportsDirectory=coverage-${process.pid}) or run them sequentially.`,
+      )
+    }
+
+    throw new Error(
+      `Could not acquire the coverage report directory lock for "${this.options.reportsDirectory}". `
+      + `Give each run its own "coverage.reportsDirectory" or run them sequentially.`,
+    )
```
AriPerkkio: In practice, what case could trigger this?

AriPerkkio (changes requested):
Overall looks good. This will prevent two simultaneous Vitest runs from deleting each other's results. Some minor comments below.

packages/vitest/src/node/coverage.ts
jgamaraalv: Multiple agents working on and testing parallel tasks in the same worktree

packages/vitest/src/node/coverage.ts
jgamaraalv: I will add some extra comments and you can decide if we keep them

packages/vitest/src/node/coverage.ts
jgamaraalv: I tested it again.. in practice almost nothing reached it. With the old 10-iteration loop that final throw was only hit if a stale lock kept being reclaimed yet re-created by a competing process on every iteration, which is effectively unreachable. I've simplified acquire to a single attempt plus one stale-lock reclaim retry, so that dead-code fallback is gone, the only remaining "in use" throw is the explicit live-owner one.

test/coverage-test/test/temporary-files.unit.test.ts
```diff
@@ -17,3 +36,57 @@ test('missing coverage temp directory throws an actionable error', async () => {
...
+
+  await provider.cleanAfterRun()
+})
+
+test('clean() throws an actionable error when another live process holds the lock', async () => {
+  const { provider, reportsDirectory, lockFile } = createProvider()
+
+  const child = spawn(process.execPath, ['-e', 'setInterval(() => {}, 1000)'], { stdio: 'ignore' })
+  await new Promise(resolve => child.once('spawn', resolve))
+  onTestFinished(() => child.kill())
```
AriPerkkio:
```suggestion
  onTestFinished(() => {
    child.kill()
  })
```

AriPerkkio (dismissed):
Looks good to me. @hi-ogawa & @sheremet-va any concerns?

packages/vitest/src/node/coverage.ts
```diff
@@ -262,6 +277,105 @@ export class BaseCoverageProvider {
     this.pendingPromises = []
   }
 
+  private get reportsDirectoryLockFile(): string {
```
hi-ogawa: nit. not sure why but this private getter doesn't feel called for. can this be in initialize? if not, prefer `reportsDirectoryLockFile()`.

packages/vitest/src/node/coverage.ts
```diff
@@ -262,6 +277,105 @@ export class BaseCoverageProvider {
...
+    const hash = createHash('sha256')
+      .update(resolve(this.options.reportsDirectory))
+      .digest('hex')
+      .slice(0, 16)
+
+    return resolve(tmpdir(), `vitest-coverage-${hash}.lock`)
+  }
+
+  private async acquireReportsDirectoryLock(): Promise<void> {
```
hi-ogawa: nit, I feel putting all implementation under `BaseCoverageProvider` feels make lock implementation cluttered. Didn't fully follow the logic yet but would it be more self-explanatory if we split this into standalone-ish class/instance like this?

```js
class MyLockThing {
  constructor(maybeoption) {}
  async acquire() {}
  async release() {}
}

class BasCoverageProvder {
   lock!: MyLockThing // initi somewhere
}
```

jgamaraalv:
@hi-ogawa @AriPerkkio @sheremet-va 

**Last two review comments (resolved in e5199e2)**

Two nits are addressed by e5199e2 (`refactor(coverage): extract reportsDirectory lock into a dedicated class`):

- The `reportsDirectoryLockFile` getter that "didn't feel called for" is gone. The lock logic now lives in a dedicated `ReportsDirectoryLock` class that owns its own `lockFile` (computed in its constructor) and exposes `acquire()` / `release()`.
- The lock implementation no longer clutters `BaseCoverageProvider` — the provider just holds a `ReportsDirectoryLock` instance, which also resolves the "split this into a standalone class" suggestion.

**About the follow-up commit 4d81d3a**

Heads up that I pushed one more commit after the approval — `fix(coverage): harden reportsDirectory lock against races and corrupt locks` (4d81d3a). While revisiting the extracted lock I ran it through a deeper correctness/concurrency pass and tightened a few things on the lock primitive itself (not the public behaviour):

- atomic lock publish (write temp + `fs.link`) so the lock file is never observable empty/partial and can't be stolen mid-write;
- release the lock on process exit so a throwing report path / `reportOnFailure` / watch / graceful close no longer leaks a stale lock into the OS temp dir;
- race-safe stale-lock reclaim (rename-aside + re-read + fail-closed, never destroying a displaced live owner's copy);
- full owner-identity check (pid + reportsDirectory + timestamp + per-acquire nonce) to defeat PID reuse and hash-collision false matches;
- only `ESRCH` counts as a dead process; unreadable-lock I/O errors propagate instead of silently stealing the lock;
- tolerate `EPERM`/`EBUSY`/`EACCES` on the reclaim rename (Windows).

This sits on top of an already-reviewed/approved feature, so **if you'd rather keep the PR minimal and scoped to just the requested fixes, I'm happy to drop 4d81d3a and leave only the review-comment resolution** — just let me know your preference. Equally happy to keep it if the extra hardening is welcome.

jgamaraalv:
@AriPerkkio any news on this issue?

packages/vitest/src/node/coverage.ts
```diff
@@ -891,3 +908,319 @@ function resolveMergeConfig(mod: any): any {
...
+      }
+
+      // The lock is stale: its owner exited (its own exit handler should have
+      // removed it; a leaked lock with a since-reused pid is a SIGKILL-only
+      // edge), the file is unreadable, or it belongs to a different directory.
+      // Reclaim it via an atomic rename onto a private path, then re-read that
+      // moved inode race-free: only discard-and-reclaim if it is still the
+      // stale/garbage lock we intended to remove. A concurrent reclaimer may
+      // have already published a live lock at this path between our read and our
+      // rename (CWE-367); if we displaced a live/newer owner we restore it and
```
AriPerkkio: What's CWE-367? How is it relevant here?

packages/vitest/src/node/coverage.ts
```diff
@@ -891,3 +908,319 @@ function resolveMergeConfig(mod: any): any {
...
+        ) {
+          rmSync(this.lockFile, { force: true })
+        }
+      }
+      catch {
+        // Exit handlers must never throw.
+      }
+    }
+
+    process.once('exit', this.exitHandler)
```
AriPerkkio: Why is exit handler needed? Doesn't the next run see that process has exited and just clear the lockfile?

AriPerkkio (changes requested):
This PR started as small improvement but now the LLM behind this code is getting quite verbose. I'm starting to consider whether we should just discard this PR and let human rewrite it without LLMs wall-of-text comments and _never-going-to-happen-edge-case_ -handlings. In the end we will be the ones maintaining this for years. 😬 

Maybe this lockfile stuff and all possible edge cases should be extracted into its own npm package even.

packages/vitest/src/node/coverage.ts
jgamaraalv: That was just TOCTOU (time-of-check-to-time-of-use) jargon I'd stuck on the rename-aside reclaim, and yeah it had no business being in here. That whole path is gone in 888edb5d, so the reference went with it.

packages/vitest/src/node/coverage.ts
jgamaraalv: Exactly, it was redundant. The next run reads the lock, sees the owner pid is dead and clears it before reclaiming, so the exit handler wasn't really buying anything. Dropped it in 888edb5d, a lock leaked by a hard-killed process just gets cleaned up on the next run now.

jgamaraalv:
Yeah, fair, and you're right. That last hardening commit got away from me.. I went down a rabbit hole guarding race windows that don't really happen (the 3-process SIGKILL reclaim especially) and ended up with comments longer than the code they were explaining.

Pushed 888edb5d that rips all of it back out. What's left is about as small as this lock gets:

- write the lockfile with `wx` (create-or-fail), that's the actual mutual exclusion
- already there + owner pid alive -> fail fast with the actionable error
- already there + owner pid dead -> remove it, try once more
- release in `cleanAfterRun`

No more nonce/timestamp identity, no write+link publish, no rename-aside/restore dance, no exit handler. coverage.ts is ~280 lines lighter and the tests dropped the cases that only existed to prove the machinery I just deleted.

Honestly I'd rather keep it this small than spin it into its own package, there isn't much left to extract now. If you still want a human to rewrite it from scratch I won't argue, but I think this version is one you'd actually be ok maintaining for years. Sorry for the noise on the way here.

jgamaraalv:
> Yeah, fair, and you're right. That last hardening commit got away from me.. I went down a rabbit hole guarding race windows that don't really happen (the 3-process SIGKILL reclaim especially) and ended up with comments longer than the code they were explaining.
> 
> Pushed [888edb5](https://github.com/vitest-dev/vitest/commit/888edb5d6e5292f46dfe63855b6585d1b54aeee8) that rips all of it back out. What's left is about as small as this lock gets:
> 
> * write the lockfile with `wx` (create-or-fail), that's the actual mutual exclusion
> * already there + owner pid alive -> fail fast with the actionable error
> * already there + owner pid dead -> remove it, try once more
> * release in `cleanAfterRun`
> 
> No more nonce/timestamp identity, no write+link publish, no rename-aside/restore dance, no exit handler. coverage.ts is ~280 lines lighter and the tests dropped the cases that only existed to prove the machinery I just deleted.
> 
> Honestly I'd rather keep it this small than spin it into its own package, there isn't much left to extract now. If you still want a human to rewrite it from scratch I won't argue, but I think this version is one you'd actually be ok maintaining for years. Sorry for the noise on the way here.

A propósito, quero frisar que nenhuma alteração feita com o auxilio de LLM foi feita sem a minha revisão e testes de que as implementações realmente funcionavam. Os comentários extras foram adicionados depois de surgirem algumas dúvidas sobre algumas implementações, conforme  respondi em um dos seus request changes e mais uma vez peço desculpas por isso. 

Novamente, acho as correções desse PR 100% válidas, não descartaria, mas fica ao seu critério e a critério do time, que respeito muito o trabalho.

Encerro as minhas contribuições nesse PR por aqui.
Abraços! 🇧🇷

AriPerkkio (dismissed):
Looks good to me now. Thanks for addressing the feedback and minimizing the changes again.

packages/vitest/src/node/coverage.ts
```diff
@@ -92,6 +94,14 @@ export class BaseCoverageProvider {
   roots: string[] = []
   changedFiles?: string[]
 
+  private _reportsDirectoryLock?: ReportsDirectoryLock
+
+  private get reportsDirectoryLock(): ReportsDirectoryLock {
+    return (this._reportsDirectoryLock ??= new ReportsDirectoryLock(
+      resolve(this.options.reportsDirectory),
+    ))
+  }
```
hi-ogawa: nit. any reason why this isn't instantiated just once at `_initialize`?

AriPerkkio (approved):

PR #10468, opened 2026-05-27
fix: remove browser packages from peer dependencies

sheremet-va:
We don't directly import them so there is no reason to define them in peer dependencies. They are imported by the user in the config file


hi-ogawa (dismissed):

hi-ogawa:
lockfile needs update.

sheremet-va:
Ohhh, we do actually need them in peer dependencies for _types_

PR #10469, opened 2026-05-27
fix: fix `importOriginal` with optimizer and query import

### LINKED ISSUE #9887: vi.mock importOriginal causes ERR_MODULE_NOT_FOUND with &v=<hash> suffix appended to path in 4.1.0 (regression from 4.0.x)
## Bug Report

**Vitest version:** 4.1.0
**Regression from:** 4.0.18 (works correctly)
**Environment:** Node 22, happy-dom

### Description

In Vitest 4.1.0, calling `vi.mock` with `importOriginal` (or `vi.importActual`) on a package listed in `test.server.deps.inline` causes `ERR_MODULE_NOT_FOUND`. The module evaluator appends `&v=<hash>` directly to the resolved file path, producing a path that does not exist on disk.

The same error occurs when the package is in `test.deps.optimizer.client.include` — the pre-bundled chunk path also gets the suffix appended.

This is a regression — identical configuration works correctly on 4.0.18.

### Minimal reproduction

```ts
// vitest.config.ts

export default defineConfig({
  test: {
    environment: 'happy-dom',
    server: {
      deps: {
        inline: ['react-redux'],
      },
    },
  },
});
```

```ts
// some.test.ts
vi.mock('react-redux', async (importOriginal) => ({
  ...(await importOriginal()),
  useSelector: vi.fn(),
}));

test('example', () => {});
```

### Error output

```
Error: [vitest] There was an error when mocking a module.
Caused by: Error: Cannot find module '/node_modules/react-redux/dist/react-redux.mjs&v=c1c7e391'
  { code: 'ERR_MODULE_NOT_FOUND' }
```

The `&v=<hash>` suffix is appended to the literal resolved file path by the module evaluator, making it unresolvable on disk.

When the package is instead added to `test.deps.optimizer.client.include`, the pre-bundled chunk gets the same treatment:

```
Error: Cannot find module '/node_modules/.vite/vitest/.../deps/react-redux.js&v=8f02ed9a'
  { code: 'ERR_MODULE_NOT_FOUND' }
```

### Expected behavior

`importOriginal` / `vi.importActual` should resolve the module correctly regardless of whether it is listed in `server.deps.inline` or `optimizer.client.include`, as it did in 4.0.x.

### Workaround

Pinning to `~4.0.0` via `overrides` in `package.json` resolves the issue.

davidxharris:
### Description

- closes https://github.com/vitest-dev/vitest/issues/9887

<details><summary>original ocmment</summary> 

Fxes a bug where `removeQuery` incorrectly strips out the leading `?` parameter boundary symbol when purging internal runtime flags (like `_vitest_original`), leaving an orphaned `&` boundary behind. When a URL contains multiple query parameters (e.g., `file.js?_vitest_original&v=12345`), removing the first parameter results in an invalid string format (`file.js&v=12345`). The module evaluator subsequently treats the query parameter as a literal extension on disk, throwing an `ERR_MODULE_NOT_FOUND` runtime error. 

This fix ensures that if a leading `?` is stripped but subsequent `&` parameters remain, the first remaining ampersand is properly converted back into a `?` query parameter delimiter.

</details>

### Tests

### Documentation

### Changesets


hi-ogawa (changes requested):
Please add integration test.

davidxharris:
Hi @hi-ogawa, I'd be happy to add one

I couldn't find an existing test suite that directly covers this scenario or usage. Could you point me toward the preferred location for this integration test, along with maybe an example of a similar test you're looking for please? Thank you

hi-ogawa:
For example, there's an integration test with `optimzeDeps` where we run fixture like this 

https://github.com/vitest-dev/vitest/blob/46b8277add7978095392c2b9b03a90d02d8ed3e8/test/e2e/test/optimize-deps.test.ts#L7-L10

https://github.com/vitest-dev/vitest/blob/46b8277add7978095392c2b9b03a90d02d8ed3e8/test/e2e/fixtures/optimize-deps/vitest.config.ts#L3-L6

davidxharris:
Added integration test. The test fails on the old code with the exact issue reported `ERR_MODULE_NOT_FOUND` and now passes with the fix

hi-ogawa:
Can you test optmizeDeps scneario that reproduced the issue previously?

davidxharris:
I apologize i misunderstood. I updated the optimizeDeps test suite to include a new test. This new test uses the same `vitest.config` as the others which includes using `optimizeDeps.include` and `test.deps.optimizer.client` which is the scenario thats specified in the original ticket. 

The test uses `vi.mock` along with `importOriginal` which is also the scenario specified in the ticket. I confirmed the test does fail with the old code and pass with my changes.

I removed the other test i had added but i can add it back if you'd like though i think only one is necessary

packages/vitest/src/runtime/moduleRunner/utils.ts
```diff
@@ -16,6 +19,6 @@ export function injectQuery(url: string, queryToInject: string): string {
 
 export function removeQuery(url: string, queryToRemove: string): string {
   return url
-    .replace(new RegExp(`[?&]${queryToRemove}(?=[&#]|$)`), '')
-    .replace(/\?$/, '')
+    .replace(new RegExp(`([?&])${queryToRemove}(?:&|$)`), '$1')
+    .replace(trailingSeparatorRE, '')
```
hi-ogawa, line 23: It looks like Claude and Me hallucinated bad `removeQuery` in https://github.com/vitest-dev/vitest/pull/9772. Now I ported the one from Vite that does handles the multiple queries properly.

hi-ogawa (approved):

tomasz-megaport:
Thanks for fixing this! Would it be possible to backport to the `4.1.x` line? This still reproduces on the latest published releases - `vitest@4.1.8` and `vitest@5.0.0-beta.4` (with `vite@8.0.16`), and it's pool-independent (threads/forks/vmThreads).

It blocks `deps.optimizer` + `vi.mock(mod, async (importOriginal) => ({ ...await importOriginal() }))` for any optimized dep (a common pattern for partially mocking a component library), with `ERR_MODULE_NOT_FOUND` on a malformed `deps/<dep>.js&v=<hash>` URL.

Minimal config that triggers it:
```ts
// vitest.config.ts
export default defineConfig({
  optimizeDeps: { include: ['@scope/ui'] },
  test: { deps: { optimizer: { client: { enabled: true, include: ['@scope/ui'] } } } },
})
// a spec doing: vi.mock('@scope/ui', async (io) => ({ ...await io(), x: vi.fn() })) -> ERR_MODULE_NOT_FOUND
```

PR #10486, opened 2026-05-31
docs: clarify assert usage in in-source tests

### LINKED ISSUE #10471: `assert` doesn't work in in-source testing
### Describe the bug

In in-source testing, calling `assert` (or any function typed with asserts expression) produces a TypeScript error:

```
TS2775: Assertions require every name in the call target to be declared with an explicit type annotation.
```

The same code works without error in a regular `.test.ts` file.

If this is a TypeScript language limitation that cannot be fixed on Vitest's side, the in-source testing or `assert` documentation should mention it with a note.

### Reproduction

```ts
const add = (a: number, b: number): number => a + b

if (import.meta.vitest) {
  const { assert, test } = import.meta.vitest

  test('add', () => {
    // @ts-expect-error -- TS2775: Assertions require every name in the call target to be declared with an explicit type annotation.
    assert(add(1, 2) === 3)
  })
}
```

### System Info

```shell
vitest: 4.1.6
typescript: 5.8.3
node.js: 22.20.0
```

### Used Package Manager

pnpm

### Validations

rohanadepu:
### Description

Documents the TypeScript assertion-function limitation for `assert` in in-source tests.

The issue is that destructuring `assert` from `import.meta.vitest` can trigger TS2775 unless the assertion function has an explicit type annotation. This adds the maintainer-suggested workaround to the in-source testing guide:

- use `import.meta.vitest!.assert(...)` directly, or
- annotate the destructured assertion as `Chai.Assert`.

Closes #10471

### Validation

- `pnpm exec eslint docs/guide/in-source.md`
- `pnpm run docs:build` attempted, but failed in existing Twoslash examples before this changed page because the local docs build could not resolve existing `vitest` / `vitest/node` imports without built package artifacts.

### DIFF
--- a/docs/guide/in-source.md
+++ b/docs/guide/in-source.md
@@ -152,6 +152,22 @@ To get TypeScript support for `import.meta.vitest`, add `vitest/importMeta` to y
 
 Reference to [`examples/in-source-test`](https://github.com/vitest-dev/vitest/tree/main/examples/in-source-test) for the full example.
 
+When using assertion functions like `assert`, TypeScript requires the call target to have an explicit type annotation. If you destructure `assert` from `import.meta.vitest`, annotate it before use:
+
+```ts [src/index.ts]
+if (import.meta.vitest) {
+  const { test } = import.meta.vitest
+  const assert: Chai.Assert = import.meta.vitest.assert
+
+  test('add', () => {
+    const value = add(1, 2)
+
+    import.meta.vitest!.assert(value === 3)
+    assert(value === 3)
+  })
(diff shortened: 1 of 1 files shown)


rohanadepu:
Hi, Yes I am a real person and this is my GitHub account. 

I made this pr for issue #10471. I have read and reviewd the pr and stand by it. I reviewed the final diff myself and stand behind the change. 

I used Codex to help find the issue and prepare the initial documentation change, then I reviewed the content and validation before opening the PR.

PR #10495, opened 2026-06-01
feat(coverage): pass previous threshold as second argument to autoUpdate function

### LINKED ISSUE #10459: coverage.thresholds.autoUpdate: pass previous threshold as second argument

As a developer using Vitest I want `coverage.thresholds.autoUpdate` to receive
the previous threshold as a second argument so that I can tell apart V8 variance
from a real regression.

The current callback signature is `(newValue: number) => number`. Without the
old value there is no way to write logic like "only accept a drop of max X%"

### Suggested solution

Pass the current (pre-update) threshold as a second argument to the `autoUpdate`
callback. The value is already in scope inside `updateThresholds` as `thresholds[key]` — it just isn't forwarded to the formatter. The change is a single line and fully backwards-compatible (existing single-arg callbacks continue to work).

```ts
// packages/vitest/src/node/coverage.ts — updateThresholds()
// before
const formattedValue = thresholdFormatter(actual)
// after
const formattedValue = thresholdFormatter(actual, threshold ?? 100)
```

This enables a drift-tolerance pattern:
```ts
// vitest.config.ts
coverage: {
  thresholds: {
    autoUpdate: (newValue, oldValue) => {
      if ((oldValue - newValue) > X) {
        throw new Error(`Coverage dropped ${(oldValue - newValue).toFixed(1)}% — intentional?`)
      }
      return Math.max(0, Math.floor(newValue) - X)
    }
  }
}
```

### Alternative

_No response_

### Additional context

In my experience V8's native coverage can produce slightly different numbers between runs, which is why a fixed buffer feels necessary. Is this something others run into too? We would love to hear what buffer values people are using in practice.

I looked at the source and the change appears small enough that we would be happy to open a PR if the maintainers agree with the direction. The old threshold is already in scope when `thresholdsToUpdate` is built — it just needs to be
carried through to the formatter call, and the `autoUpdate` type signature updated to `(newValue: number, oldValue: number) => number`. 
Is that the right approach?

### Validations

wouterkroes:
### Description

When `coverage.thresholds.autoUpdate` is set to a function, the previous threshold value is now passed as the second argument. This allows users to log or track changes, e.g.:

```ts
autoUpdate: (newThreshold, previousThreshold) => {
  console.log(`Updated threshold from ${previousThreshold} to ${newThreshold}`)
  return Math.floor(newThreshold)
}
```

Resolves #10459

Please don't delete this checklist! Before submitting the PR, please make sure you do the following:

- [x] Please, don't make changes to pnpm-lock.yaml unless you introduce a new test example.
- [x] Please check Allow edits by maintainers to make review process faster. Note that this option is not available for repositories that are owned by Github organizations.

Tests

- [x] Run the tests with pnpm test:ci.

Documentation

- [x] If you introduce new functionality, document it. You can run documentation with pnpm run docs command.

Changesets

- [x] Changes in changelog are generated from PR name. Please, make sure that it explains your changes in an understandable manner. Please, prefix changeset messages with feat:, fix:, perf:, docs:, or chore:.

### DIFF
--- a/docs/config/coverage.md
+++ b/docs/config/coverage.md
@@ -250,7 +250,7 @@ Check thresholds per file.
 Update all threshold values `lines`, `functions`, `branches` and `statements` to configuration file when current coverage is better than the configured thresholds.
 This option helps to maintain thresholds when coverage is improved.
 
-You can also pass a function for formatting the updated threshold values:
+You can also pass a function for formatting the updated threshold values. The function receives the new threshold as the first argument and the previous threshold as the second:
 
 <!-- eslint-skip -->
 ```ts
@@ -259,6 +259,12 @@ You can also pass a function for formatting the updated threshold values:
     thresholds: {
       // Update thresholds without decimals
       autoUpdate: (newThreshold) => Math.floor(newThreshold),
+
+      // Log the change and update without decimals
+      autoUpdate: (newThreshold, previousThreshold) => {
+        console.log(`Updated threshold from ${previousThreshold} to ${newThreshold}`)
+        return Math.floor(newThreshold)
+      },
 
       // 95.85 -> 95
       functions: 95,
--- a/packages/vitest/src/node/coverage.ts
+++ b/packages/vitest/src/node/coverage.ts
@@ -595,7 +595,7 @@ export class BaseCoverageProvider {
             )
         : [coverageMap.getCoverageSummary()]
 
-      const thresholdsToUpdate: [Threshold, number][] = []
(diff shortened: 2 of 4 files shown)


AriPerkkio (approved):
Looks good, thanks!

wouterkroes:
Hi @AriPerkkio, I just noticed that not all tests are passing. Is this a known issue, or did I do something wrong on my end?

image

AriPerkkio:
That's just flakiness. Our test suite is quite extensive and resource heavy.

PR #10497, opened 2026-06-01
fix(browser): wait for iframe tester readiness before preparing

soconnor-seeq:
### Description

During the investigation for `vitest`-randomly-hanging-in-CI I (with the help of Codex) initially [diagnosed a race condition that was ultimately fixed](https://github.com/vitest-dev/vitest/pull/10397) but I still was getting CI failures / hangs (even after that first fix) — exposing another race condition.

It's a browser-runner readiness race where the orchestrator could send the initial `prepare` event before the tester `<iframe>` was actually ready to receive channel messages.

Resolves (do you need an issue for a bug fix?)

This PR adds an explicit tester-ready handshake between the browser tester and orchestrator.

#### Changes

- Add a new `<iframe>` channel event: `ready`
- Have the tester post `ready` only after its channel listener is registered
- Make the orchestrator wait for the current `<iframe>` to report `ready` before sending `prepare`

### Tests

### Documentation

### Changesets

AI assisted: Codex

### DIFF
--- a/packages/browser/src/client/orchestrator.ts
+++ b/packages/browser/src/client/orchestrator.ts
@@ -1,5 +1,5 @@
 import type { Context as OTELContext } from '@opentelemetry/api'
-import type { GlobalChannelIncomingEvent, IframeChannelIncomingEvent, IframeChannelOutgoingEvent, IframeViewportDoneEvent, IframeViewportFailEvent } from '@vitest/browser/client'
+import type { GlobalChannelIncomingEvent, IframeChannelEvent, IframeChannelOutgoingEvent, IframeViewportDoneEvent, IframeViewportFailEvent } from '@vitest/browser/client'
 import type { BrowserTesterOptions, SerializedConfig } from 'vitest'
 import type { FileSpecification } from 'vitest/internal/browser'
 import { channel, client, globalChannel } from '@vitest/browser/client'
@@ -16,6 +16,8 @@ export class IframeOrchestrator {
   private cancelled = false
   private recreateNonIsolatedIframe = false
   private iframes = new Map<string, HTMLIFrameElement>()
+  private readyIframes = new Set<string>()
+  private readyWaiters = new Map<string, () => void>()
 
   public eventTarget: EventTarget = new EventTarget()
 
@@ -91,6 +93,8 @@ export class IframeOrchestrator {
 
     this.iframes.forEach(iframe => iframe.remove())
     this.iframes.clear()
+    this.readyIframes.clear()
+    this.readyWaiters.clear()
 
     for (let i = 0; i < options.files.length; i++) {
       if (this.cancelled) {
@@ -149,8 +153,7 @@ export class IframeOrchestrator {
       // because we called "cleanup" in the previous run
       // the iframe is not removed immediately to let the user see the last test
       this.recreateNonIsolatedIframe = false
-      this.iframes.get(ID_ALL)!.remove()
-      this.iframes.delete(ID_ALL)
+      this.removeIframe(ID_ALL)
       debug('recreate non-isolated iframe')
     }
 
@@ -189,8 +192,7 @@ export class IframeOrchestrator {
     const file = spec.filepath
 
     if (this.iframes.has(file)) {
-      this.iframes.get(file)!.remove()
-      this.iframes.delete(file)
+      this.removeIframe(file)
     }
 
     await this.prepareIframe(
@@ -253,12 +255,14 @@ export class IframeOrchestrator {
         }
         else {
           this.iframes.set(iframeId, iframe)
-          this.sendEventToIframe({
-            event: 'prepare',
-            iframeId,
-            startTime,
-            otelCarrier: this.traces.getContextCarrier(otelContext),
-          }).then(resolve, error => reject(this.dispatchIframeError(error)))
+          this.waitForReady(iframeId)
+            .then(() => this.sendEventToIframe({
+              event: 'prepare',
+              iframeId,
+              startTime,
+              otelCarrier: this.traces.getContextCarrier(otelContext),
+            }))
+            .then(resolve, error => reject(this.dispatchIframeError(error)))
         }
       }
       iframe.onerror = (e) => {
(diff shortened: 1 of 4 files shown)


packages/browser/src/client/orchestrator.ts
```diff
@@ -250,13 +260,15 @@ export class IframeOrchestrator {
...
         else {
-          this.iframes.set(iframeId, iframe)
-          this.sendEventToIframe({
-            event: 'prepare',
-            iframeId,
-            startTime,
-            otelCarrier: this.traces.getContextCarrier(otelContext),
-          }).then(resolve, error => reject(this.dispatchIframeError(error)))
+          didHandleInitialLoad = true
+          this.iframeRegistry.waitForReady(iframeId, iframeInstanceId)
```
soconnor-seeq: This is where the actual bug-fix is. Instead of just assuming we can `this.sendEventToIframe({ event: 'prepare', ... })` we now wait until the iframe has told us it is ready.

packages/browser/src/client/orchestrator.ts
```diff
@@ -225,7 +223,10 @@ export class IframeOrchestrator {
     startTime: number,
     otelContext: OTELContext,
   ) {
-    const iframe = this.createTestIframe(iframeId)
+    const iframe = document.createElement('iframe')
+    const iframeInstanceId = this.iframeRegistry.register(iframeId, iframe)
```
sheremet-va: Why do we need multiple iframe ids? (`iframeId` + `iframeInstanceId`)

sheremet-va (changes requested):

sheremet-va:
> This PR adds an explicit tester-ready handshake between the browser tester and orchestrator, and tracks `<iframe>` instance generations so stale `ready` events from an older `<iframe>` cannot satisfy readiness for a newly recreated `<iframe>`.

I am confused why we need multiple id trackers. There can only be _one_ iframe on the page at a time, the `iframeId` is the name of the test file. Iframes are always removed when the test is finished even in watch mode, we need to recreate them.

There should never be any "stale" or "reloaded" iframes

packages/browser/src/client/orchestrator.ts
```diff
@@ -240,7 +241,16 @@ export class IframeOrchestrator {
             + `Received URL: ${href || 'unknown due to CORS'}\nExpected: ${iframe.src}`,
           )))
         }
-        else if (this.iframes.has(iframeId)) {
+        else if (this.iframeRegistry.get(iframeId) !== iframe) {
```
sheremet-va: This should physically not be possible, we remove the iframe with the same id before this is ever called

And even then I can't imagine a situation when there _would_ be the same iframe already

test/unit/test/iframe-registry.test.ts
```diff
@@ -0,0 +1,150 @@
+// @vitest-environment happy-dom
```
sheremet-va: I would prefer tests that actually fail without you change (even if not always). Having unit tests for the registry doesn't give me any confidence that API works in a real orchestrator. Especially because there is supposed to only be a single iframe anyway. The idea of having a registry with this premise sounds a bit ridiculous, not gonna lie

sheremet-va:
I've also noticed that after your previous PR we are getting `Test timed out in 360000ms` errors more often in brower tests

soconnor-seeq:
> I've also noticed that after your previous PR we are getting `Test timed out in 360000ms` errors more often in brower tests

Oh dang! Probably best to revert that PR then. :disappointed:

packages/browser/src/client/orchestrator.ts
soconnor-seeq: I (and with Codex's non-help) was being overly cautious here. I saw that the `iframeId` was the test's name/path and/or the `__vitest_all__`. I thought that the hang I was seeing could have been caused by a race between `<iframe>` elements being added and removed by the orchestrater vs the event's being broadcast from the `<iframe>`.

But as it turns out all this extra `iframeInstanceId` stuff wasn't needed. I simplified the code — returning it back to the state that was on main — and just focused on sending the `ready` event from `tester.ts`. Building `vitest` with that change and testing it 20+ times in CI showed no hangs caused by the race-condition where the "orchestrator could send the initial prepare event before the tester <iframe> was actually ready to receive channel messages".

test/unit/test/iframe-registry.test.ts
soconnor-seeq: > The idea of having a registry with this premise sounds a bit ridiculous, not gonna lie

Yea it was. I was trusting Codex's answers about how it was possible that there would be stale events coming from `<iframe>`s that needed to be disambiguated.

I've removed this unit test and with Codex was able to make a `test/browser/specs/readiness.test.ts` integration test file which fails on the current `main` branch but passes w/ this code change.

packages/browser/src/client/orchestrator.ts
```diff
@@ -251,12 +253,14 @@ export class IframeOrchestrator {
         }
         else {
           this.iframes.set(iframeId, iframe)
-          this.sendEventToIframe({
-            event: 'prepare',
-            iframeId,
-            startTime,
-            otelCarrier: this.traces.getContextCarrier(otelContext),
-          }).then(resolve, error => reject(this.dispatchIframeError(error)))
+          this.waitForReady(iframeId)
```
soconnor-seeq, line 258: This is where the actual bug-fix is. Instead of just assuming we can `this.sendEventToIframe({ event: 'prepare', ... })` we now wait until the `<iframe>` has told us it is ready.

packages/browser/src/client/orchestrator.ts
soconnor-seeq: Outdated comment as there is no longer an `this.iframeRegistry`.

packages/browser/src/client/orchestrator.ts
soconnor-seeq: The two iframe ids have been removed — I've removed the `iframeRegistry` code.

packages/browser/src/client/orchestrator.ts
```diff
@@ -16,6 +16,8 @@ export class IframeOrchestrator {
   private cancelled = false
   private recreateNonIsolatedIframe = false
   private iframes = new Map<string, HTMLIFrameElement>()
+  private readyIframes = new Set<string>()
+  private readyWaiters = new Map<string, Array<() => void>>()
```
sheremet-va: Why do we need an array? How could there be multiple waiters?

sheremet-va (changes requested):

packages/browser/src/client/orchestrator.ts
soconnor-seeq: Nope this was Codex being overly cautious again. It is seemingly obsessed with the notion that it is theoretically possible for there to be multiple `<iframe>` = needed an array of waiters. I've updated it to just be a `new Map<string, () => void>()`.

sheremet-va (approved):

sheremet-va:
LGTM, please confirm that the fix works for your use case 🙏

soconnor-seeq:
> LGTM, please confirm that the fix works for your use case 🙏

Yes I tested this PR's build version of `vitest` in my CI pipeline and w/ it's changes the CI builds pass 20/20. Before they were hanging ~25% of the time.

sheremet-va (approved):

Issue #4476, opened 2023-11-11
Coverage providers are prone to out-of-memory errors


AriPerkkio:
### Describe the bug

Vitest runner sends coverage reports to main thread which stores them in memory during test run:

https://github.com/vitest-dev/vitest/blob/cc19117d303ad0a36721ddee01c14f04812245ae/packages/coverage-istanbul/src/provider.ts#L46-L52

https://github.com/vitest-dev/vitest/blob/cc19117d303ad0a36721ddee01c14f04812245ae/packages/coverage-v8/src/provider.ts#L59

This is fine for most projects but larger projects can run into Node's out-of-memory issues. This is common problem for coverage tools. As commented in the Istanbul's `provider.ts`, this was only a matter of time before Vitest started to run into these. Similar reports form others tools:

- https://github.com/bcoe/c8/issues/338
- https://github.com/istanbuljs/nyc/issues/1263
- https://github.com/jestjs/jest/issues/14287

Instead of storing reports in memory, we should write them on file system. This should be done in main thread here without blocking the runner.

https://github.com/vitest-dev/vitest/blob/cc19117d303ad0a36721ddee01c14f04812245ae/packages/coverage-istanbul/src/provider.ts#L118-L130

Example below demonstrates the idea. Note that this is just a simplified example. Generating reports is much more complicated:

```js
import * as fs from "node:fs/promises";

export class IstanbulCoverageProvider {
  pendingPromises = []

  onAfterSuiteRun({ coverage, transformMode, projectName }) {
    // Write reports to file system. Do not block the runner by `await`'in this call.
    // Instead store the promise for later awaiting
    const promise = fs.writeFile('some-filename.json', JSON.stringify({ coverage, transformMode, projectName }))

    this.pendingPromises.push(promise)
  }

  async reportCoverage() {
    // All tests have run, time to generate report
    await Promise.all(this.pendingPromises);

    const coverageMap = { ... };

    // Read results from file system one by one to avoid causing OOMs
    for (const filename of fs.readdir(...)) {
      const coverage = JSON.stringify(fs.readFileSync(filename));
      coverageMap.merge(coverage);
    }
  }
}
```

This should be possible to implement without causing breaking changes. Only the coverage providers need changes.

### Reproduction

https://github.com/vitest-tests/coverage-large

[...]

PR #10506, opened 2026-06-02
perf(coverage): speed up v8 report generation (bounded-memory merge + precompiled globs)

toxik:
### Description

Two independent, self-contained performance fixes to coverage report generation (`generateCoverage` / `BaseCoverageProvider`), measured on a real 1124-test-file app (~39 MB of raw V8 coverage across the run). Numbers below are best-of-3 runs, baseline vs this PR on the same machine:

| Phase | Before | After |
| --- | --- | --- |
| read + merge | ~20.8s | ~0.30s |
| convert (incl. `isIncluded` filtering) | ~22.2s | ~2.6s |
| **total `generateCoverage`** | **~43s** | **~2.9s** |

Peak heap during coverage generation drops from **~2.5 GB → ~1.15 GB**, and resident raw coverage is now bounded to ≤16 MB regardless of suite size.

> `isIncluded` runs per covered file *inside* `convertCoverage`, so the glob fix's savings land in the **convert** row above; the pure `astV8ToIstanbul` transform (~2.6s) is itself unchanged.

### Tests

No behavior change. Both fixes are verified against the existing `test/coverage-test` suites (v8 + istanbul), which assert coverage output and the set of included files. `pnpm typecheck` and `pnpm lint` pass. A backport to the `v4` release line will follow.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


toxik:
> Hello @toxik. Your PR has been labeled `maybe automated` because it appears to have been fully generated by AI with no human involvement.
> 
> 
> 
> 

Hello,

I am indeed a human and there was indeed coding agent help with this PR from Claude Code.
I've reviewed and tested the code prior to opening the PR, as well as tested it in a large project to confirm the numbers.

packages/coverage-v8/src/provider.ts
```diff
@@ -59,31 +59,48 @@ export class V8CoverageProvider extends BaseCoverageProvider implements Coverage
...
 
     const autoAttachSubprocess = this.options.autoAttachSubprocess
 
     await this.readCoverageFiles<RawCoverage>({
       onFileRead(coverage) {
-        merged = mergeProcessCovs([merged, coverage])
-
-        // mergeProcessCovs sometimes loses autoAttachSubprocess
-        const fromExtendedContext = autoAttachSubprocess ? coverage.result.filter(r => r.isExtendedContext) : []
+        coverages.push(coverage)
```
AriPerkkio: We cannot hold these in-memory as it would cause out-of-memory issues in large projects.

AriPerkkio (changes requested):
The early coverage map merging is there intentionally. We cannot hold all coverage results in memory at once, as it would lead to out-of-memory crashes. That's the reason why we do file-system round-trip in the first place, instead of just holding everything in memory.

Related issue:
- https://github.com/vitest-dev/vitest/issues/4476

packages/coverage-v8/src/provider.ts
toxik: Thanks for the review!
I'll look into what can be done to reduce the memory footprint.
In the meantime, would it make sense to open a separate PR for the regex optimization?

packages/vitest/src/node/coverage.ts
```diff
@@ -189,6 +186,25 @@ export class BaseCoverageProvider {
...
+   * every call, which dominates the filtering step on large test suites.
+   */
+  private getGlobMatchers(): { matchExclude: (file: string) => boolean; matchInclude: (file: string) => boolean } {
+    if (!this._globMatchers) {
+      const exclude = this.options.exclude
+      const include = this.options.include || '**'
+
+      this._globMatchers = {
+        matchExclude: exclude.length ? pm(exclude, { dot: true }) : () => false,
+        matchInclude: pm(include, { dot: true, ignore: exclude }),
```
AriPerkkio: This could probably also check `this.options.include`, and default to `() => true` when it's falsy. 🤔

packages/coverage-v8/src/provider.ts
AriPerkkio: Yup sounds good! Also if it makes sense to optimize the `startOffset` and `autoAttachSubprocess` handling separately, feel free to include them there.

packages/vitest/src/node/coverage.ts
toxik: addressed in [43c3903f76e14c51b578dab4a9090e05c21be04a](https://github.com/vitest-dev/vitest/pull/10506/changes/43c3903f76e14c51b578dab4a9090e05c21be04a)

packages/coverage-v8/src/provider.ts
toxik: left things in the same PR though 😅 
batched the files into ~16MB chunks to still keep O(n) instead of O(n^2)
https://github.com/vitest-dev/vitest/pull/10506/changes/01c2c277730c23c5470f31027356df5842c35ca0

toxik: hey @AriPerkkio, have you had a chance to look over the changes, what do you think?
from my measurements the memory usage is lower with this PR than with main

AriPerkkio: Thanks for the changes. Haven't had time to go through them yet.

toxik: My pleasure. Let me know when I can rebase to cleanup the commits.

packages/coverage-v8/src/provider.ts
```diff
@@ -59,31 +65,71 @@ export class V8CoverageProvider extends BaseCoverageProvider implements Coverage
     const start = debug.enabled ? performance.now() : 0
 
     const coverageMap = this.createCoverageMap()
+
+    // `mergeProcessCovs` is the expensive part of coverage generation: every call
+    // rebuilds the full range-tree forest, so its cost (in both time and peak memory)
+    // scales with the number of calls. Merging the whole run in a single pass is
+    // therefore the fastest and lightest option - but it would hold every raw coverage
+    // in memory at once, which can OOM very large suites (#4476).
```
AriPerkkio: Is there a way to fix this directly in `mergeProcessCovs` instead of adding hacks like this downstream? What makes `mergeProcessCovs` expensive?

toxik: I didn't see a straightforward way to do that.

AriPerkkio: Ok, I'll need to check `mergeProcessCovs` first before I understand why these changes are needed.

toxik: Hey @AriPerkkio, let me know how can I help move the PR forward, thanks!

AriPerkkio: Maybe reproduction where we can see the difference this PR introduces would be good. That could also be used to debug why `mergeProcessCovs` is so slow.

AriPerkkio: Ok, so the slowness likely comes from the fact that `mergeProcessCovs` processes the non-changing scripts too. This should be avoidable by using `mergeScriptCovs` instead, and only processing the changed scripts' URLs. No artificial 16MB limit checking needed.  

But we'll still need reproduction to see that any of this has effect on performance.

toxik: ok, check the repro in 109823addda22641119ed0d48f429b9697cd5704
and using mergeScriptCovs instead of mergeProcessCovs in 6c16369021356fc021366c34101d330338b58869

packages/coverage-v8/src/provider.ts
```diff
@@ -59,42 +59,47 @@ export class V8CoverageProvider extends BaseCoverageProvider implements Coverage
...
 
     await this.readCoverageFiles<RawCoverage>({
       onFileRead(coverage) {
-        merged = mergeProcessCovs([merged, coverage])
+        for (const script of coverage.result) {
+          const previous = mergedScripts.get(script.url)
+          const startOffset = previous?.startOffset || script.startOffset || 0
+          const isExtendedContext = autoAttachSubprocess
+            && (previous?.isExtendedContext || script.isExtendedContext)
+          const merged = mergeScriptCovs(previous ? [previous, script] : [script]) as ScriptCoverageWithOffset
```
AriPerkkio:
```suggestion
          const merged: typeof script = mergeScriptCovs(previous ? [previous, script] : [script])
```

packages/coverage-v8/src/provider.ts
```diff
@@ -59,42 +59,47 @@ export class V8CoverageProvider extends BaseCoverageProvider implements Coverage
...
+          mergedScripts.set(merged.url, merged)
+        }
       },
       onFinished: async (project, environment) => {
+        const result = Array.from(mergedScripts.values())
+          .sort((a, b) => a.url.localeCompare(b.url))
+
+        for (const [scriptId, script] of result.entries()) {
+          script.scriptId = String(scriptId)
+        }
```
AriPerkkio: I don't think we use `scriptId` at all. We should be able to remove this completely:


```suggestion
```

packages/coverage-v8/src/provider.ts
```diff
@@ -59,42 +59,47 @@ export class V8CoverageProvider extends BaseCoverageProvider implements Coverage
...
-            const actual = fromExtendedContext.find(r => r.url === result.url)
-            result.isExtendedContext = actual?.isExtendedContext
-          }
-        })
+          mergedScripts.set(merged.url, merged)
+        }
       },
       onFinished: async (project, environment) => {
+        const result = Array.from(mergedScripts.values())
+          .sort((a, b) => a.url.localeCompare(b.url))
```
AriPerkkio: Order of scripts should not matter.

packages/vitest/src/node/coverage.ts
```diff
@@ -197,6 +195,25 @@ export class BaseCoverageProvider {
...
 
+  /**
+   * Compile `coverage.include`/`coverage.exclude` into reusable matchers once.
+   * `picomatch.isMatch(file, patterns, options)` recompiles the patterns on
+   * every call, which dominates the filtering step on large test suites.
+   */
+  private getGlobMatchers(): { matchExclude: (file: string) => boolean; matchInclude: (file: string) => boolean } {
+    if (!this._globMatchers) {
+      const exclude = this.options.exclude
+      const include = this.options.include
```
AriPerkkio: I think we should reset `_globMatchers` in `_initialize` where we set `this.options`. Just in case it's called again with new values.

AriPerkkio (approved):
Looks good now, thanks for addressing all the findings @toxik! 💯
Sorry for long review time. 

On my machine (M4 Pro chip) I saw following improvements using artificial benchmark:

```
┌────────────────┬───────────────────────┬───────────────────┬─────────┐
│ coverage files │ old incremental merge │ new per-URL merge │ speedup │
├────────────────┼───────────────────────┼───────────────────┼─────────┤
│ 200            │ 77 ms                 │ 38 ms             │ 2.0x    │
├────────────────┼───────────────────────┼───────────────────┼─────────┤
│ 800            │ 795 ms                │ 125 ms            │ 6.4x    │
├────────────────┼───────────────────────┼───────────────────┼─────────┤
│ 2000           │ 5368 ms               │ 271 ms            │ 19.8x   │
└────────────────┴───────────────────────┴───────────────────┴─────────┘
```

Memory usage is a bit higher than in `main`, but it's not crashing:
- This PR: https://github.com/vitest-dev/vitest-ecosystem-ci/actions/runs/29315467267/job/87028381562
- Main: https://github.com/vitest-dev/vitest-ecosystem-ci/actions/runs/29314233868/job/87024505831

PR #10511, opened 2026-06-02
feat!: inline `@vitest/runner` package, do not publish it anymore

sheremet-va:
This package is very tightly connected to Vitest and is not used outside of `vitest`. We had to develop a lot of hacky workaround to augment types or pass down objects in a strange way to keep it agnostic, but the benefits do not outweght the negatives at this point.

The `runner` config option still works the same way as it did before, the public types are also not touched. This is mostly a refactor change.

TODO:
- [x] Make utils modules not be bundled together with the runner (like `createFileTask`) since these modules are also used on node side


hi-ogawa (commented):
Looks clean to me.

Can we dedupe this copy of `generateFileHash` now?  https://github.com/vitest-dev/vitest/blob/680ac2ec69a663d3372fb319ea13c8b62586393d/packages/browser/src/client/orchestrator.ts#L484-L502

PR #10516, opened 2026-06-03
feat!: expose `concurrencyId`/`workerId` on TestModule's diagnostics, make id 1-based

### LINKED ISSUE #10306: `concurrencyId` not exposed to custom reporters

Vitest runs tests in parallel across workers, but custom reporters have no way to know which worker / concurrency lane a given test ran on.

Knowing the concurrency lane is useful for diagnosing flakes that correlate with specific workers, e.g. resource contention, port collisions, or state leaked between tests sharing a worker.

### Suggested solution

It would be useful for the reporter API to expose a `concurrencyId` (or worker id) on each `TestCase` or test result. That's what Playwright Test do.

### Alternative

_No response_

### Additional context

_No response_

### Validations

sheremet-va:
Browser Mode previously didn't support this, now it should be available in `testModule.diagnostic().concurrencyId` and in `import.meta.env.VITEST_POOL_ID` and `import.meta.env.VITEST_WORKER_ID`

This also changes `workerId` from 0-based to 1-based (marking as a breaking change because of it)

Fixes #10306

### DIFF
--- a/packages/vitest/src/node/pools/browser.ts
+++ b/packages/vitest/src/node/pools/browser.ts
@@ -291,6 +291,33 @@ class BrowserPool {
     })
   }
 
+  // stable slot id (1..maxWorkers) assigned to each session/orchestrator on its
+  // first run, exposed to the test runner as both `concurrencyId` and `workerId`.
+  // the id lives on the session, so it is freed when the session disconnects, and
+  // the used set is derived from the live orchestrators, so it stays within maxWorkers
+  private getConcurrencyId(sessionId: string): number {
+    const sessions = this.project.vitest._browserSessions
+    const session = sessions.getSession(sessionId)
+    if (session?.concurrencyId) {
+      return session.concurrencyId
+    }
+    const used = new Set<number>()
+    for (const id of this.orchestrators.keys()) {
+      const concurrencyId = sessions.getSession(id)?.concurrencyId
+      if (concurrencyId) {
+        used.add(concurrencyId)
+      }
+    }
+    let concurrencyId = 1
+    while (used.has(concurrencyId)) {
+      concurrencyId++
+    }
+    if (session) {
+      session.concurrencyId = concurrencyId
+    }
+    return concurrencyId
+  }
+
   private getOrchestrator(sessionId: string) {
     const orchestrator = this.orchestrators.get(sessionId)
     if (!orchestrator) {
@@ -358,6 +385,7 @@ class BrowserPool {
           },
         },
         async () => {
+          const concurrencyId = this.getConcurrencyId(sessionId)
           return orchestrator.createTesters(
             {
               method,
@@ -366,6 +394,10 @@ class BrowserPool {
               // so we need to stringify it first to avoid double serialization
               providedContext: this._providedContext || '[{}]',
               otelCarrier: this._traces.getContextCarrier(),
+              concurrencyId,
+              // in the browser there is a single tab per orchestrator,
+              // so the worker id matches the concurrency slot
+              workerId: concurrencyId,
             },
           )
         },
--- a/packages/vitest/src/node/pools/browser.ts
+++ b/packages/vitest/src/node/pools/browser.ts
@@ -264,7 +264,7 @@ class BrowserPool {
       this.project.vitest._browserSessions.sessionIds.add(sessionId)
       const project = this.project.name
       debug?.('[%s] creating session for %s', sessionId, project)
-      let page = this._traces.$(
+      const page = this._traces.$(
         `vitest.browser.open`,
         {
           context: this._otel.context,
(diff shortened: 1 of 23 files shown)


AriPerkkio (dismissed):
Looks good. Maybe Node side test too at:

https://github.com/vitest-dev/vitest/blob/206e8cff82f3157d9f2ac1f3b038ba42a9af6dc3/test/e2e/test/reported-tasks.test.ts#L65-L104

AriPerkkio (dismissed):

packages/vitest/src/node/pools/pool.ts
```diff
@@ -262,7 +262,7 @@ export class Pool {
   }
 
   private getWorkerId() {
-    let workerId = 0
+    let workerId = 1
```
AriPerkkio: This could as well be `undefined` and throw if loop below didn't set any value. Why is this change needed? 1-based should already be used in init of `this.workerIds` 👀

packages/vitest/src/node/pools/pool.ts
```diff
@@ -262,7 +262,7 @@ export class Pool {
...
 
     this.workerIds.forEach((state, id) => {
       if (state && !workerId) {
```
AriPerkkio: Doesn't this check fail now?

packages/vitest/src/node/pools/pool.ts
sheremet-va: > should already be used

It's not, the worker id is 0 without the change

sheremet-va: Although, let me check again, maybe it's only about workerId

sheremet-va: Yes, it was only about `workerId`; changed this to `workerId == null` check

packages/vitest/src/node/pools/browser.ts
```diff
@@ -367,6 +394,10 @@ class BrowserPool {
               // so we need to stringify it first to avoid double serialization
               providedContext: this._providedContext || '[{}]',
               otelCarrier: this._traces.getContextCarrier(),
+              concurrencyId,
+              // in the browser there is a single tab per orchestrator,
+              // so the worker id matches the concurrency slot
+              workerId: concurrencyId,
```
hi-ogawa, line 400: Is it intended to have `workerId` being independently starting from zero for node pool and browser pool? 

This may feel non intuitive for node/browser mixed projects run since the same `diagnostic.workerId === 0` in reporter metadata can come from different node and browser project.

sheremet-va: Yes, it's intended. They are separate pools with separate `workers` and run independently from each other. Until we combine them, they _must_ have different poolId/workerId

packages/vitest/src/node/reporters/reported-tasks.ts
```diff
@@ -741,6 +743,16 @@ export interface ModuleDiagnostic {
...
+  /**
+   * The id of the worker that ran this file. This value cannot be higher than `maxWorkers`.
+   * If file did not run yet, this will be 0.
+   */
+  readonly concurrencyId: number
+  /**
+   * Incremental number of the worker that ran this file. This number increases with each worker.
+   * If file did not run yet, this will be 0.
+   */
+  readonly workerId: number
```
hi-ogawa, line 765: Doc update? https://main.vitest.dev/api/advanced/test-module.html#diagnostic

Also we can probably add jsdoc mentioning runtime equivalent is `import.meta.env.VITEST_POOL_ID/WORKER_ID`.

packages/vitest/src/node/pools/browser.ts
hi-ogawa: For the use case of OP https://github.com/vitest-dev/vitest/issues/10306#issuecomment-4412024407, `concurrencyId` isn't meant to be overlapped when running node/browser projects running in parallel. Don't remember mixed pol case, but can overlap this happen depending on project `groupOrder`? Not blocking, but just a question I wanted to confirm.

sheremet-va: groupOrder happens within the pool, so ids stay within maxWorkers

For the purpose of OP’s use case, they need to take into account that browser pool runs independently, they can do it based on project.isBrowserEnabled (and add a browser prefix, for example)

packages/vitest/src/node/reporters/reported-tasks.ts
sheremet-va: Added to docs

packages/vitest/src/node/pools/browser.ts
sheremet-va: Added a warning to docs

hi-ogawa (approved):

Issue #10436, opened 2026-05-22
`Error.stack` (e.g. inline snapshot) are slow in angular browser mode with V8 coverage enabled


uneknown:
### Describe the bug

We noticed a significant slowdown with inline snapshots in browser mode when using V8 coverage.
 
A first `toMatchInlineSnapshot()` call in a test can take around ~1000ms with coverage enabled (in our CI; GitHub Codespaces has similiar performance), while the same test takes only a few ms without coverage.
 
 The slow part seems to be inline snapshot callsite resolution: `SnapshotState._resolveInlineStack` reads `error.stack`, and with the profiler enabled (which V8 coverage does), the first stack string formatting is very slow.
 
 This makes test suites with many inline snapshots noticeably slower when v8 coverage is enabled in browser mode.

### Reproduction

https://github.com/uneknown/vitest-browser-v8-coverage-inline-snapshot-repro

### System Info

```shell
System:
    OS: macOS 26.3.1
    CPU: (14) arm64 Apple M4 Pro
  Binaries:
    Node: 25.2.1
    npm: 11.6.2
    pnpm: 10.31.0
    Watchman: 2026.01.12.00
  Browsers:
    Chrome: 145.0.7632.160
    Safari: 26.3.1
  npmPackages:
    @vitest/browser: 5.0.0-beta.3 => 5.0.0-beta.3 
    @vitest/browser-playwright: 5.0.0-beta.3 => 5.0.0-beta.3 
    @vitest/coverage-v8: 5.0.0-beta.3 => 5.0.0-beta.3 
    playwright: ^1.60.0 => 1.60.0 
    vitest: 5.0.0-beta.3 => 5.0.0-beta.3
```

### Used Package Manager

npm

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vitest.dev/guide/).
- [x] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vitest/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
[...]

PR #10519, opened 2026-06-04
perf: skip stack trace probing on inline snapshot happy path

hi-ogawa:
### Description

- related https://github.com/vitest-dev/vitest/issues/10436

Did attempt something but it looks like such optimization isn't really sound to validate `different snapshots cannot be called at the same location` properly. 

I think this change can be still justified if there's clear perf benefit. The optimization applies on good CI run for example.

TODO
- [ ] minimal repro
- [ ] benchmark to justify
- [ ] more test cases

### Tests

### Documentation

### Changesets


hi-ogawa:
In the team discussion, we agreed that this behavior change is acceptable for the benefit of happy path performance.

---

The example of behavior change:

#### Example 1:

```js
test.for(["hey", "world"])("test %s", (arg) => {
  expect(arg.length).toMatchInlineSnapshot(`5`);
});
```

Previously
- test hey -> fails because 3 != 5
- test world -> fails because multiple different snapshot at same location

This PR
- test hey -> fails because 3 != 5
- test world -> pass (since we don't resolve stack and cannot know they are same location)

#### Example 2:

```js
test.for(["hey", "world"])("test %s", (arg) => {
  expect(arg.length).toMatchInlineSnapshot(`3`);
});
```

Previously
- test hey -> pass
- test world -> fails because multiple different snapshot at same location

This PR
- test hey -> pass
- test world -> fails because 3 != 5

hi-ogawa:
From the benchmarking I tried, each `toMatchInlineSnapshot` only costs less than 1ms in any scenario (browser mode, coverage), so I don't think this optimization and associated behavioral change is worth it, so closing. Here is my benchmark setup for reference. https://github.com/hi-ogawa/reproductions/tree/main/vitest-10519-inline-snapshot-stack

PR #10531, opened 2026-06-07
fix(cache): persist source map in fs module cache for coverage v8

jaxalo:
### Description

When `experimental.fsModuleCache` is enabled, the v8 coverage provider silently drops files from istanbul output on cache hits.

**Root cause.** `inlineSourceMap` (in `packages/vitest/src/node/environments/fetchModule.ts`) only inlines source maps that have a `version` field. Any transform whose map lacks `version` — the rollup `{ mappings: '' }` sentinel is the common case — is written to cache without an inline `//# sourceMappingURL`. On read, `getCachedModule` recovered the map *only* by re-parsing that inline comment via `extractSourceMap`, so for those files the cached `transformResult.map` was `null`.

Coverage v8 then calls `project.vite.environments[env].transformRequest(filepath)` during `convertCoverage`, gets `result.map === undefined`, hands `sourceMap: undefined` to `ast-v8-to-istanbul`, and the byte-offset → source-position translation silently returns an empty istanbul object for the file. Net effect: cached files show 0% (or strictly less than non-cache runs) coverage even though they executed.

**Fix.** Persist `fetchResult.map` explicitly in the cache meta and prefer it on read. `extractSourceMap(cachedModule.code)` is kept as a fallback so the contract for already-cached files is backwards-compatible.

### Reproduction context

Discovered on a real-world ~3000-file codebase: enabling `fsModuleCache` dropped ~135 files to 0% and reduced another ~100 to roughly half their non-cache coverage. Disabling the cache restored full coverage. The per-file regression is consistent across runs, matching the cache-key behavior (same source = same cache hit = same lost map).

### Tests

Added `test/unit/test/fs-module-cache.test.ts` covering the `FileSystemModuleCache.saveCachedModule` → `getCachedModule` round-trip for three cases:
- Standard rollup source map with `version`
- The `{ mappings: '' }` sentinel (the original failing case)
- Missing map (`null`) — to confirm we don't accidentally fabricate one

All three pass with the fix; the sentinel case fails on `main` because the cache round-trip returns `undefined` for `map`.

### Tests
  Verified `pnpm vitest run test/unit/test/fs-module-cache.test.ts` (3 passed, vmThreads/threads/forks pools).

### Documentation
  No public API change; behavior fix only.

### Changesets
- [x] Changes in changelog are generated from PR name. Please, make sure that it explains your changes in an understandable manner.

### AI disclosure

This PR was prepared with Claude (Anthropic) as a coding assistant. A human reviewed the diagnosis, the fix, and the test against a real codebase that surfaced the bug.

### DIFF
--- a/packages/vitest/src/node/cache/fsModuleCache.ts
+++ b/packages/vitest/src/node/cache/fsModuleCache.ts
@@ -1,4 +1,4 @@
-import type { DevEnvironment, FetchResult } from 'vite'
+import type { DevEnvironment, FetchResult, Rollup } from 'vite'
 import type { Vitest } from '../core'
 import type { ResolvedConfig } from '../types/config'
 import fs, { existsSync, mkdirSync, readFileSync } from 'node:fs'
@@ -110,6 +110,7 @@ export class FileSystemModuleCache {
       code,
       importedUrls: meta.importedUrls,
       mappings: meta.mappings,
+      map: meta.map,
     }
   }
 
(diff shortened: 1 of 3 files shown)


AriPerkkio (changes requested):
As this affects coverage results, we'll need proper e2e test under `test/coverage-test`. That e2e test is essentially a minimal reproduction of the bug.

sheremet-va:
Doubling the size of the cache file for an edge case seems unreasonable. If the issue is with mappings: “”, then we can just inline them too or store only that kind if a source map.

jaxalo:
Closing this. After more investigation, the real bug at our codebase was not the rollup-sentinel-map cache loss this PR addresses — it was a worker OOM that the pool runner was masking. Adding `experimental.fsModuleCache: true` happened to surface it because the cache made init faster, leaving more memory for tests, which let them run longer before OOMing.

The actual mechanism: when a fork worker dies unexpectedly (OOM in our case), `PoolRunner.emitUnexpectedExit` emits an `'error'` event on the internal event emitter, but the per-task resolver in `Pool.schedule()` only attaches an `'error'` listener when `!runner.isStarted`. For shared runners (`isolate: false` with subsequent atoms), no listener exists. The throw escapes into a libuv-callback context, the resolver stays pending forever, `pool.runTests` hangs, `runFiles.finally` never fires, coverage is silently skipped, and main exits cleanly with code 0 once the event loop drains.

I'll open a follow-up PR that fixes that hang directly (permanent per-task error listener so worker death rejects the resolver instead of leaving it pending). The source-map cache change in this PR is still defensible as a defensive improvement, but it doesn't reproduce in a clean test the way I'd want for a coverage-affecting PR, so closing rather than asking you to merge on a thin case.

Thanks for the review.

jaxalo:
Closing — see latest comment. Wrong root cause, opening a follow-up.

PR #10539, opened 2026-06-08
fix: vi.stubEnv(key, undefined) removes the variable after process.env is reassigned

### LINKED ISSUE #10525: vitest 4 broke stubEnv(..., undefined) to unset an environment variable
### Describe the bug

According to:
- https://github.com/vitest-dev/vitest/pull/6359
- https://github.com/vitest-dev/vitest/blob/main/docs/api/vi.md#vistubenv-vi-stubenv

I would expect 

```js
vi.stubEnv("foo", undefined);
```

to delete `process.env.foo`.

This was working fine until vitest 4: now, passing `undefined` as value to `stubEnv()` doesn't seem to do anything anymore.

### Reproduction

In [this PR](https://github.com/ValentinaVasile/vitest-stubenv-repro/pull/4) of a dedicated repo, you will see that the same test:

- passes in vitest 3
- fails in vitest 4
- fails in vitest 5-beta

The test fails in vitest 4 and 5-beta because the initial value of the environment variable remains after the call to `stubEnv(..., undefined)`

### System Info

```shell
System:
    CPU: (4) x64 AMD EPYC 7763 64-Core Processor
    Memory: 14.62 GB / 15.61 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  Binaries:
    Node: 24.16.0 - /opt/hostedtoolcache/node/24.16.0/x64/bin/node
    Yarn: 1.22.22 - /usr/local/bin/yarn
    npm: 11.13.0 - /opt/hostedtoolcache/node/24.16.0/x64/bin/npm
  Browsers:
    Chrome: 148.0.7778.178
    Chromium: 148.0.7778.0
    Firefox: 151.0.1
    Firefox Developer Edition: 151.0.1
  npmPackages:
    vitest: ^4.1.8 => 4.1.8
```

### Used Package Manager

npm

### Validations

VihAMBR:
### Description

`vi.stubEnv(key, undefined)` is documented to delete the variable, and it does in the common case. But when test code reassigns `process.env` to a new object (for example `process.env = {}` in a `beforeEach`), the variable is not removed and `process.env[key]` keeps its value.

The `import.meta.env` proxy created in the worker reads and writes the live `process.env` in its `get` and `set` traps, but it has no `deleteProperty` trap. So `delete` falls back to the default behaviour, which removes the key from the object the proxy was constructed with at worker init. Once `process.env` is reassigned, that captured target is stale, so the delete misses the current `process.env` and the variable survives.

This adds a `deleteProperty` trap that deletes from the live `process.env`, matching `get` and `set`.

Resolves #10525

### Tests

Added two cases to `test/unit/test/stubs.test.ts`: `stubEnv(key, undefined)` removes the var, and removes it even after `process.env` is reassigned. The second fails on `main` and passes with this change. Verified red to green across the threads, forks, and vmThreads pools.

### DIFF
--- a/packages/vitest/src/runtime/worker.ts
+++ b/packages/vitest/src/runtime/worker.ts
@@ -110,5 +110,14 @@ function createImportMetaEnvProxy(): WorkerGlobalState['metaEnv'] {
 
       return true
     },
+    deleteProperty(_, key) {
+      if (typeof key !== 'string') {
+        return true
+      }
+
+      delete process.env[key]
+
+      return true
+    },
   }) as WorkerGlobalState['metaEnv']
 }
--- a/test/unit/test/stubs.test.ts
+++ b/test/unit/test/stubs.test.ts
@@ -82,6 +82,30 @@ describe('stubbing envs', () => {
     expect(process.env.VITE_TEST_UPDATE_ENV).toBe('development')
   })
 
+  it('stubs to undefined to remove the env var', () => {
+    vi.stubEnv('VITE_TEST_UPDATE_ENV', undefined)
+    expect('VITE_TEST_UPDATE_ENV' in process.env).toBe(false)
+    expect(process.env.VITE_TEST_UPDATE_ENV).toBeUndefined()
+    expect(import.meta.env.VITE_TEST_UPDATE_ENV).toBeUndefined()
+  })
+
+  it('stubs to undefined to remove the env var after process.env is reassigned', () => {
+    const original = process.env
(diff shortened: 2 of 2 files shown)


VihAMBR:
I am a real human, and have read reviewed and stand behind this content. 

I understand everything with context to this problem in the codebase,. I have also ensured it follows best practices and is in line with guildlines and templates. 

i have used claude code to write a bulk of this code but i have reviewed it as a person.

DamienCassou:
@sheremet-va why was this closed?

PR #10541, opened 2026-06-08
fix: stale mock metadata breaks automocking with isolate:false (fix #10145)

### LINKED ISSUE #10145: Automocking silently breaks with isolate: false + mixed mock types and resolve aliases
### Describe the bug

When `isolate: false` is set and two test files mock the same module differently, one with a factory (`vi.mock(import('...'), () => (...))`) and one with automocking (`vi.mock(import('...'))`), the automock is silently skipped if the factory-mock file runs first. The automocked module's exports are not replaced with `vi.fn()`, causing `vi.mocked(fn).mockReturnValue(...)` to throw:

```
TypeError: vi.mocked(...).mockReturnValue is not a function
```

The failure is intermittent because it depends on which test file runs first in the worker.

I believe a `resolve.alias` (e.g. `~ --> ./src`) is required to trigger the bug because it causes the URL-based and ID-based registry keys to diverge (this particular diagnosis comes from Opus reviewing the commit that introduced the issue, I'm not necessarily sure enough myself)

This appears to be a regression between 4.1.0-beta.5 and 4.1.0-beta.6, and is still present in the latest versions.

### Reproduction

https://github.com/kade-robertson/vitest-mock-isolate-repro

There are Github Action results on that repo you can inspect which runs the same test suite multiple times in vitest 4.1.0-beta.5 and 4.1.4. If you want to clone and run locally, `pnpm install`, and then run `pnpm test:bad` to run against 4.1.4 and `pnpm test:good` to run against 4.1.0-beta.5.

*Note*: This is a distilled reproduction as I'm not able to share the original test suite that caused this. This was done primarily with AI assistance, so some of the writing or contents of the README may be misleading.

### System Info

```shell
System:
    OS: macOS 26.4
    CPU: (14) arm64 Apple M4 Pro
    Memory: 611.28 MB / 24.00 GB
    Shell: 4.6.0 - /opt/homebrew/bin/fish
  Binaries:
    Node: 24.13.0 - /Users/kade-mi/.local/share/mise/installs/node/24.13.0/bin/node
    npm: 11.6.2 - /Users/kade-mi/.local/share/mise/installs/node/24.13.0/bin/npm
    pnpm: 10.28.0 - /Users/kade-mi/.local/share/mise/installs/node/24.13.0/bin/pnpm
  Browsers:
    Chrome: 147.0.7727.56
    Chrome Canary: 148.0.7757.0
    Firefox: 147.0.3
    Firefox Developer Edition: 141.0
    Safari: 26.4
    Safari Technology Preview: 26.0
  npmPackages:
    @vitejs/plugin-react: 6.0.1 => 6.0.1
    @vitest/browser: 4.1.4 => 4.1.4
    @vitest/browser-playwright: 4.1.4 => 4.1.4
    @vitest/coverage-v8: 4.1.4 => 4.1.4
    @vitest/ui: 4.1.4 => 4.1.4
    playwright: 1.59.1 => 1.59.1
    vite: 8.0.8 => 8.0.8
    vitest: 4.1.4 => 4.1.4
    vitest-browser-react: 2.2.0 => 2.2.0
```

### Used Package Manager

pnpm

### Validations

kade-robertson:
### Description

When `isolate: false` is set and two test files mock the same module differently (one with a factory, one with automocking), the automock is silently skipped if the factory-mock file runs first. This happens because `fetchModule` bakes `mockedModule` (with `code: ''`) into the module node's `meta`, and with `isolate: false` this metadata persists across test files. When a later file registers an automock for the same module, `cachedRequest` was using the stale manual mock from meta instead of the current automock from the registry.

The fix adds an additional check after the early stale / self-import returns to see if `currentMock` and `mockedModule` differ, we re-fetch.

A `resolve.alias` is required to trigger the bug.

Resolves #10145 

### Tests

### Documentation

### Changesets


packages/vitest/src/runtime/moduleRunner/moduleRunner.ts
```diff
@@ -168,23 +168,28 @@ export class VitestModuleRunner
...
         return this._cachedRequest(node.url, node, callstack, metadata)
       }
-      mocked = await this.mocker.requestWithMockedModule(
-        url,
-        mod,
-        callstack,
-        mockedModule,
-      )
+      if (currentMock !== mockedModule) {
+        const freshNode = await this.fetchModule(injectQuery(url, '_vitest_original'))
```
sheremet-va, line 187: If it's a `manual` or `redirect` mock, then this is not needed because it won't be used to autospy/automock

sheremet-va (changes requested):

packages/vitest/src/runtime/moduleRunner/moduleRunner.ts
kade-robertson: If I understand correctly I believe f394cb4 addresses this

kade-robertson:
@sheremet-va Just wanted to check in on this, anything else you'd like me to address?

sheremet-va (approved):

sheremet-va:
LGTM

PR #10542, opened 2026-06-08
fix(pool): reject pending task resolver when worker exits unexpectedly

jaxalo:
### Description

`Pool.schedule()` only registers a per-task `runner.on('error', ...)` listener inside the `if (!runner.isStarted)` branch. That listener closes over the first task's resolver and is added once. For subsequent tasks reusing the same runner (the common case under `isolate: false`), no error listener is registered for the active resolver.

When a fork worker exits unexpectedly mid-test (the original report's case was an OOM under `--max-old-space-size`), `PoolRunner.emitUnexpectedExit` emits an `'error'` event on the runner's internal `EventEmitter`. With no active-task listener, the throw escapes into the libuv child-process exit callback. The current task's resolver is left pending forever — `pool.runTests` hangs on `Promise.allSettled`, `runFiles.finally` never runs, and coverage/teardown is silently skipped. Depending on the surrounding code path the main process either hits a teardown-timeout fallback or exits naturally with code 0 (the silent-failure case I hit on CI).

The fix is a permanent per-task `onTaskError` listener that rejects the resolver on error and pairs with `onFinished` for cleanup so the listener count stays bounded.

### Reproduction (real-world)

On a ~3000-test repo with `pool: 'forks'`, `isolate: false`, and a 1.1 GB heap, one fork worker per CI run would V8-OOM on a long-running test file. Pre-fix that surfaced as silent coverage loss — the failing batch's `coverage-final.json` was simply never written because `runFiles.finally` didn't fire — and the Test job was reported green by CircleCI because main exited with code 0. The downstream Coverage step then failed with per-file 0% thresholds on a seemingly random subset of files, decoupled from the actual cause.

### Test

`test/e2e/test/pool-worker-exit.test.ts` runs vitest as a subprocess against a fixture where the second of two test files on a shared runner SIGKILLs its own worker before sending `testfileFinished`. The test asserts vitest exits and surfaces a \"Worker exited unexpectedly\" error rather than hanging or exiting silently.

- [x] Please, don't make changes to \`pnpm-lock.yaml\` unless you introduce a new test example.

### Tests
- [x] Verified locally — pre-fix the bug manifests as a silent hang in a programmatic runner setup; post-fix vitest exits with the error reported. The subprocess test currently passes both ways via the pool's teardown-timeout fallback, but the fix is what guarantees runFiles.finally fires and coverage/teardown completes deterministically.

### Documentation
- [ ] No public API change.

### Changesets
- [x] PR title prefixed with \`fix:\`.

### AI disclosure

This PR was prepared with Claude (Anthropic) as a coding assistant after I hit the root cause on a real codebase and instrumented vitest's pool to confirm the mechanism.

### DIFF
--- a/packages/vitest/src/node/pools/pool.ts
+++ b/packages/vitest/src/node/pools/pool.ts
@@ -101,11 +101,19 @@ export class Pool {
           }
 
           runner.off('message', onFinished)
+          runner.off('error', onTaskError)
           resolver.resolve()
         }
       }
 
+      const onTaskError = (error: unknown) => {
+        runner.off('message', onFinished)
+        runner.off('error', onTaskError)
+        resolver.reject(error instanceof Error ? error : new Error(String(error)))
+      }
+
       runner.on('message', onFinished)
+      runner.on('error', onTaskError)
 
       if (!runner.isStarted) {
         runner.on('error', (error) => {
--- a/test/e2e/fixtures/pool-worker-exit/crash.test.js
+++ b/test/e2e/fixtures/pool-worker-exit/crash.test.js
@@ -1,0 +1,8 @@
+import { test } from 'vitest'
+
+test('the worker dies before sending testfileFinished', async () => {
+  // SIGKILL the worker process so it can't send testfileFinished back to main.
+  // Pre-fix this caused pool.run() to hang forever instead of rejecting.
+  queueMicrotask(() => process.kill(process.pid, 'SIGKILL'))
+  await new Promise(() => {})
+})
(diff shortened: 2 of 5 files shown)


jaxalo:
Closing — could not reproduce on `main`. Upstream's current pool teardown machinery already handles the worker-death case (`Pool.cancel` chain via teardown timeout writes coverage cleanly even with the listener gap). The bug as observed in vitest 4.0.15 is masked in main. I'll bisect to find which change masked it; if it turns out the protection is fragile, will re-open with a proper reproducer.

PR #10543, opened 2026-06-08
fix(pool): prevent test run hang on worker crash

jaxalo:
### Description

`Pool.schedule()` registers the per-task `runner.on('error', ...)` listener only inside the `if (!runner.isStarted)` branch. The listener closes over the *current* `resolver`, but it's added once when the runner first transitions out of IDLE. For any subsequent task that picks up the same runner from `sharedRunners` (the common case under `isolate: false`), the listener already exists, closed over the **first** task's resolver and the active task's resolver has no error path.

When a fork worker exits unexpectedly mid-test (the original report's case was V8 OOM in a 50-file batch), `PoolRunner.emitUnexpectedExit` emits an `'error'` event. The stale listener fires and calls `reject` on a long-settled resolver (no-op). The active task's resolver stays pending forever. Downstream:

- `pool.run` hangs on `await testFinish.promise`
- `pool.runTests`'s `Promise.allSettled` hangs
- `runFiles.finally` never runs, so `generateCoverage` / `reportCoverage` are skipped
- coverage report for the affected batch is **silently never written**, even though the worker's earlier-completed test files had their per-suite coverage chunks on disk
- main process eventually exits via libuv drain + `process.on('exit')` cleanup, often with code 0 making the failure invisible in CI

The fix is a permanent per-task `onTaskError` listener that rejects the resolver on error and pairs with `onFinished` for cleanup so the listener count stays bounded.

### Reproduction

The new e2e test in `test/e2e/test/pool-worker-exit.test.ts` builds the exact path:

- `pool: 'forks'`, `isolate: false`, `maxWorkers: 2` (so files become separate tasks instead of batched)
- 3 test files — the 3rd (`crash.test.js`) runs on a runner already in STARTED state because the 1st's runner was released into `sharedRunners`
- `crash.test.js` SIGKILLs its own worker mid-run via `queueMicrotask(() => process.kill(process.pid, 'SIGKILL'))`
- assert: `coverage-final.json` exists and contains src.js coverage

**5/5 runs WITHOUT the fix → test FAILS** (no `coverage-final.json` produced; vitest exits in ~1s).
**5/5 runs WITH the fix → test PASSES** (`coverage-final.json` written; vitest exits ~10s after the worker dies, triggered by the now-properly-rejected resolver flowing through to `runFiles.finally`).

### Real-world context

I hit this on a ~3000-test backend repo running vitest 4.0.15 with `pool: 'forks'`, `isolate: false`, and a 1.1 GB heap limit. One fork worker per CI run would V8-OOM on a long-running test file. Pre-fix that manifested as silent coverage loss. The failing batch's `coverage-final.json` was simply never written because `runFiles.finally` didn't fire  and the Test job was reported green by CircleCI because main exited with code 0. The downstream Coverage step then failed with per-file 0% thresholds on a seemingly random subset of files, decoupled from the actual cause.

- [x] Please, don't make changes to \`pnpm-lock.yaml\` unless you introduce a new test example.

### Tests
- [x] e2e test passes 5/5 with the fix and fails 5/5 without (no flakiness).

### Documentation
- [ ] No public API change.

### Changesets
- [x] PR title prefixed with \`fix:\`.

### AI disclosure

This PR was prepared with Claude (Anthropic) as a coding assistant. I hit the root cause on a real codebase, instrumented vitest's pool to confirm the mechanism, and iterated on the reproducer until it was deterministic.


packages/vitest/src/node/pools/pool.ts
```diff
@@ -101,19 +101,21 @@ export class Pool {
...
+
       runner.on('message', onFinished)
+      runner.on('error', onTaskError)
 
       if (!runner.isStarted) {
-        runner.on('error', (error) => {
-          resolver.reject(
-            new Error(`[vitest-pool]: Worker ${task.worker} emitted error.`, { cause: error }),
-          )
-        })
```
AriPerkkio, line 115: This was the cause of task hanging. `resolver` here references the previous test file, while `resolver` outside this scope is the current test file. Emitted error "cancelled" previous test file and left current one hanging.

packages/vitest/src/node/pools/pool.ts
```diff
@@ -148,6 +150,7 @@ export class Pool {
 
       if (
         !task.isolate
+        && !runner.isTerminated
```
AriPerkkio, line 153: Now recycling crashed workers instead of re-using, so that next test files can start properly instead of failing.

packages/vitest/src/node/pools/poolRunner.ts
```diff
@@ -368,6 +368,7 @@ export class PoolRunner {
 
   private emitUnexpectedExit = (): void => {
     const error = new Error(`Worker exited unexpectedly during ${this._state} state`)
+    this._state = RunnerState.STOPPED
```
AriPerkkio, line 371: This will prevent `Timeout terminating worker crashed-worker.test.ts` cases. No need to attempt terminating crashed workers. 

Also makes pool close quickly without teardown timeout activating.

AriPerkkio (approved):
Looks good, thanks @jaxalo 💯 

Noticed that the test case was still hanging in `pool.close()` after worker crash was properly handled, so added couple of more fixes to address that. Now test case runs quickly without any promise hangs. All these issues are strictly limited to `isolate: false` runs.

jaxalo:
Thanks a lot for the thorough review and for taking the time to push the follow-up fixes directly on the branch, @AriPerkkio learned a ton reading through your commits. Really appreciate it 🙏

PR #10550, opened 2026-06-09
chore: setup PR-driven release pipeline

### LINKED ISSUE #10551: Rework release pipeline to be PR-driven
Somewhat verbose AI write up but it takes what it takes. Accompanying PR is in
- https://github.com/vitest-dev/vitest/pull/10550

### Problem Statement

The current release flow depends on a maintainer running the release command locally. That local command directly pushes the version bump commit and release tag to a release branch, which makes the release trigger difficult to review, trace, and audit as a team workflow.

Today, [scripts/release.ts](https://github.com/vitest-dev/vitest/blob/361cf5ddebf625b355b88757a84731b1c0f1b755/scripts/release.ts#L11) runs `bumpp` with `commit: true`, `push: true`, and `tag: true`. This means a human maintainer runs `pnpm release` locally and directly pushes both the release commit and the `v*` tag to the release branch. Then [.github/workflows/publish.yml](https://github.com/vitest-dev/vitest/blob/361cf5ddebf625b355b88757a84731b1c0f1b755/.github/workflows/publish.yml#L3) watches `v*` tag pushes and publishes the package.

This creates a few problems:

- The important release decision is an invisible local action rather than a reviewable PR or a clearly attributable workflow run.
- Release commits can be pushed directly to branches such as `main`, `v4`, or backport/release branches without the same review shape as normal changes.
- The git tag is created before the npm publish has succeeded. Since the tag is the publish trigger, a failed publish can leave behind a release tag that looks authoritative even though npm did not receive the release successfully.

The primary problem is traceability: release preparation should be visible, reviewable, and attributable before it mutates release branches. The tag-ordering issue is a related consequence of the current tag-triggered publish design.

### Goal

Make release preparation PR-driven and keep the final publish/tag steps inside GitHub Actions so:

- A human starts release preparation with a manual workflow dispatch.
- Version bumps are proposed through a reviewable pull request.
- Publishing is triggered by merging the reviewed version bump into a maintained release branch.
- The npm publish happens before the `vX.Y.Z` tag is created.
- The tag and GitHub release notes are created only after npm publish succeeds.
- The flow works for `main` and maintained version branches such as `v4` / `v4.1`.
- The maintainer-facing release branch policy is codified together with the workflow, so branch targets, backports, and publish triggers stay aligned.
- Existing npm trusted publishing/provenance behavior is preserved. The current `.npmrc` enables provenance in [.npmrc](https://github.com/vitest-dev/vitest/blob/361cf5ddebf625b355b88757a84731b1c0f1b755/.npmrc#L1), and the publish workflow already uses `id-token: write` in [.github/workflows/publish.yml](https://github.com/vitest-dev/vitest/blob/361cf5ddebf625b355b88757a84731b1c0f1b755/.github/workflows/publish.yml#L19).

### Alternatives Considered

1. Keep local `pnpm release`, but stop creating the tag locally.

   This would improve tag ordering if the publish workflow created the tag after npm publish, but it would not solve the main traceability problem. The release branch would still be mutated by a local maintainer command rather than a reviewable PR.

2. Use a fully manual `workflow_dispatch` publish that bumps, publishes, commits, and tags in one workflow run.

   This would make the release trigger visible in Actions and avoid local pushes, but it would skip PR review for the version bump commit. It is simpler, but weaker if the goal is to make release preparation look like normal branch mutation.

3. Automatically open the release PR with a GitHub App token.

   This gives the smoothest PR-driven flow and can allow generated PRs to trigger normal workflows, but it requires maintaining GitHub App credentials. This remains an open implementation choice.

### Suggested Change

1. Add a `prepare-publish.yml` workflow.

   This workflow should be manually dispatched with inputs such as:

   - `target_branch`: release target, for example `main`, `v4`, or `v4.1`.
   - `release`: bumpp release type, defaulting to `next`. Supported values should be `next`, `patch`, `minor`, `major`, `prepatch`, `preminor`, and `premajor`.
   - `version`: optional exact semver version, for example `5.0.0-beta.5` or `4.1.9`. When set, it overrides `release`.

   `next` should be the normal release path. With bumpp, `next` advances the current prerelease when the current version is already prerelease, for example `5.0.0-beta.4` -> `5.0.0-beta.5`; otherwise it behaves like a patch bump, for example `4.1.8` -> `4.1.9`.

   The workflow should:

   - Check out the selected target branch.
   - Run the same monorepo-wide version bump currently handled by [scripts/release.ts](https://github.com/vitest-dev/vitest/blob/361cf5ddebf625b355b88757a84731b1c0f1b755/scripts/release.ts#L7), but without creating a tag. If `version` is set, pass that exact version to bumpp. Otherwise pass `release`, defaulting to `next`.
   - Validate the computed version after bumping: it must be valid semver, all published package manifests must match the root version, it must fit the target branch's release line, and it must not conflict with an existing tag or already-published npm version except in an intentional retry/finalization path.
   - Use the computed version to derive the prepare branch name, for example `prepare-vX.Y.Z`.
   - Before pushing the branch, check whether the same prepare branch already exists. If it exists and matches the same target/version, print the compare or existing PR URL and exit without mutating it. If it exists but does not match the expected release metadata, fail loudly.
   - Commit the version bump as `chore: release vX.Y.Z`.
   - Push the prepare branch.
   - Either print the GitHub compare URL for a human to open the pull request back to the selected target branch, or automatically open the PR with a GitHub App token.

   The release pipeline does not depend on which PR-opening mechanism is chosen, but the implementation should choose one explicitly.

   Add workflow concurrency keyed by `target_branch`, `release`, and `version` input, with `cancel-in-progress: false`, so duplicate dispatches for the same release intent do not race each other before the computed version is known.

2. Update or replace `scripts/release.ts`.

   The current script should no longer push release tags as part of release preparation. Options:

   - Convert it into a CI helper for prepare workflow use, with `tag: false` and controlled commit/push behavior.
   - Or split release helpers into explicit scripts, for example `prepare-release.ts` for version bumping and `publish-ci.ts` for npm publishing.

   The important behavior change is that release preparation may create a version bump commit, but it must not create the final `vX.Y.Z` tag.

3. Change `publish.yml` to run from release branch pushes instead of tag pushes.

   The workflow should watch only maintained release branches:

   - `main`
   - supported version branches such as `v4` or `v4.1`

   It should not publish from ad hoc `backport-*` branches. Backport branches should be used to prepare changes that merge into a maintained release branch, and the merge into that maintained branch should trigger publishing. This should follow the branch model already documented in [CONTRIBUTING.md](https://github.com/vitest-dev/vitest/blob/361cf5ddebf625b355b88757a84731b1c0f1b755/CONTRIBUTING.md#L121-L154): `main` is the active line, `vN` is the latest maintained minor for a non-main major, and `vN.M` is an older maintained minor. The workflow should not use `on.push.tags: ['v*']` as the publish trigger.

4. Detect release commits robustly.

   On a push to a maintained release branch, the workflow should determine whether the pushed commit represents a release version bump. Avoid relying only on `HEAD^..HEAD`, because a release PR could be merged with a merge commit or multiple commits.

   Use this detection rule:

   - Scan the pushed commit range, `github.event.before..github.sha`, for exactly one commit subject matching `chore: release vX.Y.Z`.
   - Validate that `HEAD` has root `package.json` and all published package manifests at version `X.Y.Z`.
   - Validate that no `vX.Y.Z` tag exists at a different commit.
   - Do not use npm/unpkg `latest` as the primary trigger, because `main` and maintained version branches may publish different version streams.

   This keeps the workflow independent of the merge method: a squash merge, regular merge commit, or fast-forward can all work as long as the pushed range contains one release commit and `HEAD` contains the release version.

5. Publish to npm before tagging.

   Update [scripts/publish-ci.ts](https://github.com/vitest-dev/vitest/blob/361cf5ddebf625b355b88757a84731b1c0f1b755/scripts/publish-ci.ts#L11) so it can accept a version derived from the release commit or package metadata, not only from `GITHUB_REF_NAME`.

   The script should keep the existing safety check that `VITE_TEST_WATCHER_DEBUG` is `false`, validate root/package versions, derive the npm dist-tag (`beta`, `alpha`, or `latest`), and publish with the existing pnpm publish behavior.

   Idempotency should live in `publish-ci.ts` or a helper it calls, not only in workflow YAML. Before publishing each package, check whether that exact package version already exists on npm. If it exists, skip that package. This lets reruns recover from "npm publish succeeded, tag/changelog failed" without trying to republish immutable npm versions.

6. Add a publish dry run before the real publish.

   Before publishing, run a dry-run publish check with the same package set and publish flags. This should catch missing files, bad package manifests, and packaging errors before the real publish step reaches npm.

   This is a useful practice from related release pipelines and fits Vitest's current goal of making the publish step safer without changing the release artifact semantics.

7. Create the git tag after npm publish succeeds.

   After `publish-ci.ts` succeeds, the workflow should:

   - Create `vX.Y.Z` at the exact release commit that was published.
   - Push the tag.
   - Run `changelogithub --to vX.Y.Z --name vX.Y.Z` so release notes are generated against the newly-created tag.

8. Make retries safe.

   The workflow should handle partial success states:

   - If npm publish succeeds but tag creation or changelog generation fails, rerunning should not attempt to republish packages that already exist.
   - If the tag already exists and points at the current release commit, rerun should proceed to changelog/finalization.
   - If the tag already exists and points elsewhere, fail loudly.
   - If npm already has the version but the tag is missing, skip publish and create/finalize the tag.

9. Preserve release permissions and trusted publishing setup.

   Keep the publish job on GitHub-hosted runners with `id-token: write`, `contents: write`, `environment: Release`, and package-manager cache disabled for the actual publish job. Vitest already uses the `Release` environment in the existing publish workflow, so the new flow should preserve that approval boundary rather than introduce a separate one. If npm trusted publisher settings are scoped to the existing workflow filename and environment, keep those stable where possible.

### Follow-up Actions

1. Backport the workflow change to maintained release branches.

   The new release workflow must exist on every branch that is expected to prepare or publish future releases. After the new workflows and scripts land on `main`, backport them to each actively maintained version branch, such as `v4` / `v4.1` at rollout time. Frozen old branches should not receive the new workflow unless the team expects to publish from them again.

   Backporting should include all files required by the new flow, for example:

   - `prepare-publish.yml`
   - updated `publish.yml`
   - any changed or newly split release scripts
   - any release documentation added as part of the change

2. Verify branch-specific compatibility.

   Supported branches may have different package manager versions, Node ranges, package lists, or workflow pins. Before enabling the new flow on each branch, verify that the workflow still installs, builds, detects the release version, and publishes using that branch's own lockfile and scripts.

3. Configure npm trusted publishing for every publishing workflow/branch combination that needs it.

   If npm trusted publisher settings are scoped by workflow filename, environment, or branch/ref, update npm settings so the new branch-triggered `publish.yml` remains authorized for `main` and supported release branches.

4. Document and codify the release workflow against the release branch policy.

   [CONTRIBUTING.md](https://github.com/vitest-dev/vitest/blob/361cf5ddebf625b355b88757a84731b1c0f1b755/CONTRIBUTING.md#L121-L162) already documents release branches, backport targets, and documentation branches. Extend that maintainer section as part of this change so it explicitly says:

   - `prepare-publish.yml` targets only maintained release branches, for example `main`, `vN`, or `vN.M`.
   - `backport-*` branches are PR/source branches only and must not be publish triggers.
   - workflow changes required for publishing must be backported to every branch that is still expected to release.
   - duplicate prepare dispatches for the same target/version reuse the existing prepare branch instead of force-pushing it.
   - tags and GitHub releases are finalized by Actions after npm publish, not by local maintainer commands.

   Keep public support policy in [docs/releases.md](https://github.com/vitest-dev/vitest/blob/361cf5ddebf625b355b88757a84731b1c0f1b755/docs/releases.md#L19-L31). Use `CONTRIBUTING.md` for the maintainer process that maps that policy to branches and workflow behavior.

5. Do a non-publishing rehearsal.

   Run the prepare workflow against a test branch or draft PR to validate branch creation, version bumping, commit message shape, and compare URL output. Separately validate release detection without actually publishing.

6. Retire the old local release path.

   Once the new flow is active on all supported release branches, update or remove local `pnpm release` usage so maintainers do not accidentally create tags before publish. Any remaining local helper should prepare a branch/commit only, not push final tags.

### Related Pipeline Findings

Rolldown has a close version of the proposed split: a manually dispatched prepare workflow opens a release PR, and a branch-push publish workflow publishes before creating the GitHub release/tag. Its publish trigger is hardcoded to `main`, so it does not directly answer Vitest's branch/backport release problem.

Oxc uses the same general prepare-PR and branch-push publish pattern for some package families. It also uses GitHub App tokens for generated release PRs and has useful npm dry-run / already-published checks. However, Oxc is split across many artifact-specific workflows and does not provide a direct model for Vitest's multi-release-branch npm publishing.

The main takeaway is to borrow the shape, not copy the exact version detection:

- Borrow: prepare PR, publish from branch version bump, dry-run publish checks, publish before tag/release, idempotent retry behavior.
- Vitest-specific call: detect release by scanning the pushed commit range for one `chore: release vX.Y.Z` commit and validating package manifests at `HEAD`.
- Vitest-specific call: publish only from maintained release branches, not ad hoc backport branches.
- Open implementation choice: human-opened compare URLs vs automatically opening PRs with a GitHub App token.
- Already covered: release environment approval exists today and should be preserved.

### Open Question

- Should `prepare-publish.yml` only push the release branch and print a compare URL for a human to open the PR, or should it automatically open the PR with a GitHub App token?

### GitHub App Setup If Auto-PR Is Chosen

If the workflow should automatically open release PRs, create a dedicated GitHub App under the `vitest-dev` organization rather than under an individual maintainer account. This keeps ownership, private-key rotation, and audit trails tied to the organization.

Recommended setup:

- Create an org-owned app, for example `vitest-release-bot`, from `vitest-dev` organization settings: https://github.com/organizations/vitest-dev/settings/apps
- Install it only on `vitest-dev/vitest`.
- Grant repository permissions:
  - `Contents: read/write`, so the workflow can push `prepare-vX.Y.Z` branches.
  - `Pull requests: read/write`, so the workflow can open release PRs.
  - `Metadata: read`, which is implicit.
- Do not grant `Actions: write` unless this app also needs to dispatch workflows.
- Store the app credentials as repo or org secrets scoped to `vitest-dev/vitest`, for example:
  - `RELEASE_GITHUB_APP_ID`
  - `RELEASE_GITHUB_APP_PRIVATE_KEY`

Workflow token generation would look like:

```yaml
- id: app-token
  uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1
  with:
    app-id: ${{ secrets.RELEASE_GITHUB_APP_ID }}
    private-key: ${{ secrets.RELEASE_GITHUB_APP_PRIVATE_KEY }}
    repositories: vitest
    permission-contents: write
    permission-pull-requests: write
```

The existing Ecosystem CI app in `ecosystem-ci-trigger.yml` already uses `actions/create-github-app-token`, but it is currently scoped for cross-repo workflow dispatch with `permission-actions: write`. It could only be reused if org admins add `contents: write` and `pull_requests: write` on `vitest`, but a separate release app is cleaner because release branch mutation and ecosystem-CI dispatch are different trust domains.

### Residual Risks

- The branch list still has to be maintained as support policy changes.
- Human-opened PRs add one manual step. A GitHub App token removes that step but adds credential management to the release path.
- Duplicate prepare dispatches for the same target/version should be treated as reuse of the existing prepare branch, not as a force-push opportunity.
- The commit-range release detection should fail closed if multiple release commits are pushed together; that is intentional, but it means unusual release recovery may need a manual rerun or small workflow override.

hi-ogawa:
### Description

- Closes https://github.com/vitest-dev/vitest/issues/10551

This PR implements the basic idea of https://github.com/vitest-dev/vitest/issues/10551. The maintainer guide of the release process is added to `CONTRIBUTING.md`. 

The test flight is done in my fork repo. See
- https://github.com/hi-ogawa/vitest/pull/6
- https://github.com/hi-ogawa/vitest/pull/7
- publish with target main and release next
  - Prepare Publish https://github.com/hi-ogawa/vitest/actions/runs/27249579624
  - PR https://github.com/hi-ogawa/vitest/pull/10
  - Publish https://github.com/hi-ogawa/vitest/actions/runs/27249819121
  - https://github.com/hi-ogawa/vitest/pull/12
  - https://github.com/hi-ogawa/vitest/actions/runs/27250746105
  - https://github.com/hi-ogawa/vitest/pull/14
  - https://github.com/hi-ogawa/vitest/actions/runs/27252213688
- prepare publish with target v4 and release next
  - Prepare publish https://github.com/hi-ogawa/vitest/actions/runs/27252466560
  - PR https://github.com/hi-ogawa/vitest/pull/16
  - Publish https://github.com/hi-ogawa/vitest/actions/runs/27252543133

There are some aspects that could be improved later but not included here:
- [x] Setup Github App to auto create PR during "prepare publish" workflow dispatch.
  - Currently "prepare publish" job only outputs the compare URL for human to create PR manually. This is because even though auto creating PR from in-repo `GITHUB_TOKEN` won't trigger the actions on the created PR due to Github's chaining action prevention. 
- [x] Automatically derive npm dist-tags for backport release case in `publish-ci.ts` script.
  - `publish-ci.ts` currently publishes backport release of npm package as `release` dist tags. To prevent this, we would need manual fix up for each backport branch like [v3](https://github.com/vitest-dev/vitest/blob/b6d56f8171ae814ee7571df63a35a0da5203dbaa/scripts/publish-ci.ts#L42) or we should implement automatic way to derive dist-tags.
- [ ] Rework changelog and github release generation. For example, changelog generation could be moved ahead to "prepare publish" time.

They are put into follow up issue https://github.com/vitest-dev/vitest/issues/10560

### Tests

### Documentation

### Changesets


.github/workflows/publish.yml
```diff
@@ -45,10 +84,40 @@ jobs:
...
+
+      - name: Push release tag
+        env:
+          TAG: ${{ needs.detect.outputs.tag }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          git config user.name "vitest-release-bot"
+          git config user.email "actions@github.com"
+
+          if git rev-parse --verify --quiet "refs/tags/$TAG"; then
```
sheremet-va: I think it should be a hard error if tag exists (probably somewhere early in the pipeline)

After this is merged, I will make tags immutable

.github/workflows/publish.yml
```diff
@@ -2,18 +2,57 @@ name: Publish Package
...
+        run: pnpm install --frozen-lockfile --prefer-offline
+        env:
+          PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD: '1'
+
+      - name: Detect release
+        id: detect
+        env:
+          RANGE: ${{ github.event.before }}..${{ github.sha }}
+          BRANCH: ${{ github.ref_name }}
+        run: pnpm exec tsx scripts/detect-release.ts --range "$RANGE" --branch "$BRANCH"
```
sheremet-va: Can't we just get the commit name from some of the `github.event` props? Or does it only have SHAs?

.github/workflows/publish.yml
```diff
@@ -45,10 +78,38 @@ jobs:
...
+          VERSION: ${{ needs.detect.outputs.version }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          TAG="v$VERSION"
+          git config user.name "vitest-release-bot"
+          git config user.email "actions@github.com"
+
+          if git rev-parse --verify --quiet "refs/tags/$TAG"; then
+            echo "Tag $TAG already exists"
+            exit 1
```
sheremet-va: Failing CI after _publishing_ is what we are trying to avoid, can we check the existence of a tag before publishing?

hi-ogawa: Yup, moving it earlier. The first draft was a mess, so just cleaning up gradually.

sheremet-va:
As a note, I really like the idea of using `workflow_dispatch` 👍

scripts/publish-ci.ts
```diff
@@ -32,12 +32,16 @@ const releaseTag = version.includes('beta')
...
+console.log(dryRun ? 'Dry-running version' : 'Publishing version', version, 'with tag', releaseTag || 'latest')
 
 if (releaseTag) {
-  await $`pnpm -r publish --access public --no-git-checks --tag ${releaseTag}`
+  await $`pnpm -r publish --access public --no-git-checks --tag ${releaseTag} ${dryRunArgs}`
 }
 else {
-  await $`pnpm -r publish --access public --no-git-checks`
+  // TODO: make stable backport releases use branch-aware npm dist-tags instead of
+  // falling through to `latest`, for example vN -> VN and vN.M -> VN_M.
```
sheremet-va: (Also important to use capital letters, lowercase is a valid semver and it's not allowed by npm)

```diff
- v3
+ V3
```

.github/workflows/prepare-publish.yml
```diff
@@ -0,0 +1,84 @@
...
+          node-version: 24
+          package-manager-cache: false
+
+      - name: Install
+        run: pnpm install --frozen-lockfile --prefer-offline
+        env:
+          PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD: '1'
+
+      # TODO: for now, this step only outputs a compare URL for maintainers to manually create a PR from.
+      # In the future, PR creation can be automated by using a Github App token like .github/workflows/ecosystem-ci-trigger.yml.
```
sheremet-va: For sure we should do that instead of manually copy paste

scripts/release.ts
```diff
@@ -1,22 +1,25 @@
...
   await versionBump({
     files: packages,
-    commit: true,
-    push: true,
-    tag: true,
+    release,
+    tag: false,
+    push: false,
+    confirm: !release,
   })
```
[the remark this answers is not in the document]
hi-ogawa: Default is `commit: true`

.github/workflows/prepare-publish.yml
hi-ogawa: We need to follow `GitHub App Setup If Auto-PR Is Chosen` in https://github.com/vitest-dev/vitest/issues/10551. I cannot access https://github.com/organizations/vitest-dev/settings/apps so you create one.

.github/workflows/publish.yml
```diff
@@ -45,10 +92,32 @@ jobs:
       - name: Build
         run: pnpm build
 
+      - name: Dry run publish
+        env:
+          VERSION: ${{ needs.detect.outputs.version }}
+          PUBLISH_DRY_RUN: 'true'
+        run: pnpm run publish-ci "$VERSION"
```
hi-ogawa: It maybe worth checking dry run in prepare-publish stage, but that involves some moving parts so later.

scripts/release.ts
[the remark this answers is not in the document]
hi-ogawa: TODO: needs `commit: true` indeed.

.github/workflows/prepare-publish.yml
```diff
@@ -0,0 +1,84 @@
+name: Prepare Publish
+
+on:
+  workflow_dispatch:
+    inputs:
+      target_branch:
+        description: Release target branch.
+        required: true
+        type: string
```
hi-ogawa, line 10: TODO: can default `main`.

scripts/release.ts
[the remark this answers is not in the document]
hi-ogawa: TODO: also `printCommits: false`.

scripts/publish-ci.ts
hi-ogawa, line 1: It looks like `publish-ci` is swallowing logs from underlying publish cli on main
- v5 beta https://github.com/vitest-dev/vitest/actions/runs/26744290771/job/78815733650
- v4 https://github.com/vitest-dev/vitest/actions/runs/26743139751/job/78811844196
- v3 https://github.com/vitest-dev/vitest/actions/runs/26747038692/job/78825060103

It's better to have full logs for the publish. Seems like that's possible by ``await $`...`.stdio('inherit')``

hi-ogawa: Also remove stale `#!/usr/bin/env zx`. This is irrelevant.

CONTRIBUTING.md
```diff
@@ -161,6 +161,16 @@ The release branches are also linked with the documentation site releases:
...
+
+1. **Prepare the release PR.** Run the `Prepare Publish` workflow, setting the `target_branch` input to the branch that matches the [release branch](#release-branches) convention for the release line, then choose the `release` or `version` input to control the version bump. The default input is `release: next`, which bumps the version to the next patch version for stable releases (e.g. `4.1.2 -> 4.1.3`), or the next prerelease version when the current version is already a prerelease (e.g. `4.2.0-beta.2 -> 4.2.0-beta.3`). Use `release: conventional` to derive the bump from the commit history, set `release` to a specific bump type, or pass an exact `version` for pre-releases. To preview the version a given `release` input resolves to, you can run `pnpm release` locally first — it lists the commits since the last release and lets you browse available release types and versions interactively (cancel before confirming so nothing is committed). The workflow pushes the version bump to a branch and prints a compare URL; open that URL to create the release PR.
```
hi-ogawa: TODO: Turns out there's another confusing input. Here "Use workflow from" should be almost always aligned with "target branch" for simple mental model. Worth clarifying here. 

Image

hi-ogawa:
@codex review

hi-ogawa:
Included npm dist tag fix here and now test publishing on fork repo:

- main branch -> beta tag https://github.com/hi-ogawa/vitest/actions/runs/27258334578
- v4 branch -> latest tag https://github.com/hi-ogawa/vitest/actions/runs/27258820340
- v3  branch -> V3 tag https://github.com/hi-ogawa/vitest/actions/runs/27259376798

.github/workflows/prepare-publish.yml
hi-ogawa: This doesn't look too bad to be honest. This is single click and not copy paste :smile: 

image

sheremet-va: Still, I don't want to sit and wait for it :D I've been known to forget to approve the release for days 😄

sheremet-va: I created it and added both `RELEASE_GITHUB_APP_ID` and `RELEASE_GITHUB_APP_PRIVATE_KEY`, but they are only allowed in the `Release` environment. Not sure if it needs a separate environment

.github/workflows/publish.yml
```diff
@@ -45,10 +88,37 @@ jobs:
       - name: Build
         run: pnpm build
 
+      - name: Install good npm
```
sheremet-va: Should we also migrate to https://github.blog/changelog/2026-05-22-staged-publishing-and-new-install-time-controls-for-npm/ ?

.github/workflows/publish.yml
```diff
@@ -45,10 +88,37 @@ jobs:
...
+        run: npm i -g npm@^11.5.2
+
+      - name: Publish to npm (dry run)
```
sheremet-va: Why do we need the dry run?

scripts/publish-ci.ts
```diff
@@ -1,43 +1,70 @@
...
+    return 'beta'
+  }
+  if (version.includes('alpha')) {
+    return 'alpha'
+  }
 
-console.log('Publishing version', version, 'with tag', releaseTag || 'latest')
+  // Need to specify tag explicitly for backport version releases
+  // since otherwise `latest` tag would be overwritten.
+  // Note that `main` branch doesn't always mean `latest` tag because of pre-release phase.
```
sheremet-va, line 46: I'm confused here - wouldn't pre-release phase already trigger one of the `ifs` before?

hi-ogawa: This is calling out the following case (and it's actually our current situation):
- main branch -> beta dist-tag
- v4 branch -> latest dist-tag

And so `main` doesn't always mean `latest`. It may be better to flip the implication as in: `latest` dist-tag doesn't always mean `main` branch. It's confusing either way :upside_down_face: Any better wording is welcome.

sheremet-va: Ah, I see, you are right

.github/workflows/publish.yml
hi-ogawa: I saw the dry run in rolldown so I thought why not. I don't think it has downside and hopefully just help catching a unseen surprise.
https://github.com/rolldown/rolldown/blob/2e0340444ae7dd0f97a9b96d1b923ce332b6650f/.github/workflows/publish-to-npm.yml#L120-L122

scripts/publish-ci.ts
sheremet-va: Should we just specify it in the workflow manually? And pass it down in PR body _somehow_. Maybe `chore: release v1.1.0 [latest]`

.github/workflows/publish.yml
sheremet-va: It does double the time it takes to run, but I guess it's fine. I had issues before where one package in the middle failed to release and had to release the rest manually..

scripts/publish-ci.ts
hi-ogawa: I'm satisfied with current `npm view` + `semver.gt` logic though maybe just because I wrote it up.

It's not impossible but currently `chore: release vx.x.x` is generated by bumpp, so would need to figure out how to insert dist-tags or other ways to pass it through.

.github/workflows/publish.yml
hi-ogawa: You mean staged publishing? I don't know how that works. We have manual approval on release environment, so seems redundant.

scripts/publish-ci.ts
sheremet-va: I guess `npm view` works, can't see a version that won't work. Unless we introduce new tags like `rc` 🤔

.github/workflows/publish.yml
```diff
@@ -2,18 +2,50 @@ name: Publish Package
...
 
 jobs:
-  publish:
-    # only run on main, don't trigger in forks
+  # Keep detection outside the Release environment. Environment approval is job-level,
+  # so the publish job is only created after a release commit is detected.
+  detect:
     if: github.repository == 'vitest-dev/vitest'
+    name: Detect release commit
+    runs-on: ubuntu-latest
```
sheremet-va: Should this be `ubuntu-slim`? It just checks commit name

.github/workflows/publish.yml
```diff
@@ -45,10 +88,37 @@ jobs:
       - name: Build
```
sheremet-va, line 88: Do we need a separate build now? dry run also builds every package

This was previously used just to check that build works before publishing

hi-ogawa: No, I don't think either `publish-ci.ts` nor `publish` triggers build. People do that with `prepack` script or alike but we don't do that. For example, this is local simulation

```sh
$ RELEASE_VERSION=5.0.0-beta.5 pnpm release 
Scope: all 42 workspace projects
[WARN] There are cyclic workspace dependencies: /home/hiroshi/code/others/vitest-release-pipeline-poc/packages/browser, /home/hiroshi/code/others/vitest-release-pipeline-poc/packages/vitest, /home/hiroshi/code/others/vitest-release-pipeline-poc/packages/browser-playwright
Lockfile is up to date, resolution step is skipped
Packages: +20
++++++++++++++++++++
Progress: resolved 20, reused 20, downloaded 0, added 20, done
Done in 991ms using pnpm v11.1.2
$ tsx scripts/release.ts
Bumping versions in packages: package.json, packages/browser/package.json, packages/browser-playwright/package.json, packages/browser-preview/package.json, packages/coverage-istanbul/package.json, packages/coverage-v8/package.json, packages/expect/package.json, packages/pretty-format/package.json, packages/mocker/package.json, packages/snapshot/package.json, packages/spy/package.json, packages/utils/package.json, packages/web-worker/package.json, packages/vitest/package.json, packages/ui/package.json, packages/browser-webdriverio/package.json 

$ VITE_TEST_WATCHER_DEBUG=false PUBLISH_DRY_RUN=true PUBLISH_BRANCH=main pnpm publish-ci 5.0.0-beta.5
Scope: all 42 workspace projects
[WARN] There are cyclic workspace dependencies: /home/hiroshi/code/others/vitest-release-pipeline-poc/packages/browser, /home/hiroshi/code/others/vitest-release-pipeline-poc/packages/vitest, /home/hiroshi/code/others/vitest-release-pipeline-poc/packages/browser-playwright
Lockfile is up to date, resolution step is skipped
Packages: +20
++++++++++++++++++++
Progress: resolved 20, reused 20, downloaded 0, added 20, done
Done in 965ms using pnpm v11.1.2
$ tsx scripts/publish-ci.ts 5.0.0-beta.5
== DRY RUN ==
Publishing version '5.0.0-beta.5' with tag 'beta'
📦 @vitest/pretty-format@5.0.0-beta.5 → https://registry.npmjs.org/
[WARN] Skip publishing @vitest/pretty-format@5.0.0-beta.5 (dry run)
📦 @vitest/spy@5.0.0-beta.5 → https://registry.npmjs.org/
[WARN] Skip publishing @vitest/spy@5.0.0-beta.5 (dry run)
📦 @vitest/utils@5.0.0-beta.5 → https://registry.npmjs.org/
[WARN] Skip publishing @vitest/utils@5.0.0-beta.5 (dry run)
📦 @vitest/expect@5.0.0-beta.5 → https://registry.npmjs.org/
[WARN] Skip publishing @vitest/expect@5.0.0-beta.5 (dry run)
📦 @vitest/mocker@5.0.0-beta.5 → https://registry.npmjs.org/
[WARN] Skip publishing @vitest/mocker@5.0.0-beta.5 (dry run)
📦 @vitest/snapshot@5.0.0-beta.5 → https://registry.npmjs.org/
[WARN] 10 other warnings
📦 @vitest/browser@5.0.0-beta.5 → https://registry.npmjs.org/
📦 vitest@5.0.0-beta.5 → https://registry.npmjs.org/
📦 @vitest/browser-playwright@5.0.0-beta.5 → https://registry.npmjs.org/
📦 @vitest/browser-preview@5.0.0-beta.5 → https://registry.npmjs.org/
📦 @vitest/browser-webdriverio@5.0.0-beta.5 → https://registry.npmjs.org/
📦 @vitest/coverage-istanbul@5.0.0-beta.5 → https://registry.npmjs.org/
📦 @vitest/coverage-v8@5.0.0-beta.5 → https://registry.npmjs.org/
📦 @vitest/web-worker@5.0.0-beta.5 → https://registry.npmjs.org/
📦 @vitest/ui@5.0.0-beta.5 → https://registry.npmjs.org/
```

scripts/publish-ci.ts
hi-ogawa: Realized I can get any prerelease through `semver.parse(..)`. Adjusted to use that instead of checking just alpha/beta.

.github/workflows/publish.yml
sheremet-va: It defends from a different type of attack though (like bypassing 2fa). The linked blog even recommends using both

hi-ogawa: Probably it can be follow up. I've added to https://github.com/vitest-dev/vitest/issues/10560

.github/workflows/publish.yml
```diff
@@ -45,10 +88,37 @@ jobs:
...
         run: pnpm build
 
+      - name: Install good npm
+        run: npm i -g npm@^11.5.2
```
sheremet-va:
```suggestion
        run: npm i -g npm@^11.17.0
```

scripts/publish-ci.ts
```diff
@@ -1,43 +1,73 @@
...
+    throw new Error('Missing PUBLISH_BRANCH environment variable')
+  }
+  const releaseTag = await getReleaseTag(version, publishBranch)
+
+  const dryRun = process.env.PUBLISH_DRY_RUN === 'true'
+  if (dryRun) {
+    console.log('== DRY RUN ==')
+  }
+  console.log(`Publishing version '${version}' with tag '${releaseTag}'`)
+  await $$`pnpm -r publish --access public --no-git-checks --tag ${releaseTag} ${dryRun ? ['--dry-run'] : []}`
```
sheremet-va:
```suggestion
  await $$`pnpm -r stage publish --access public --no-git-checks --tag ${releaseTag} ${dryRun ? ['--dry-run'] : []}`
```

.github/workflows/publish.yml
sheremet-va: this could be removed because pnpm doesn't delegate to npm cli anymore: https://pnpm.io/cli/publish

CONTRIBUTING.md
```diff
@@ -161,6 +161,21 @@ The release branches are also linked with the documentation site releases:
 - `release` points to the latest stable release line used for <https://vitest.dev/>. Release managers update it manually for non-beta releases from `main`; it is not moved for older-line backports.
 - `vN` branches are used for old major documentation sites. For example, <https://v3.vitest.dev/> uses `v3`.
 
+### Release process
```
hi-ogawa, line 164: TODO: we'll have one more step for npm staged publishing.

package.json
hi-ogawa, line 1: TODO: update pnpm for `pnpm stage` command

sheremet-va:
We would also need to backport this to v4 and v3, right? Because the release workflow needs to be identical

hi-ogawa:
Did local dry run. So many warnings but looks good:

```sh
$ RELEASE_VERSION=5.0.0-beta.5 pnpm release
[WARN] The "$" version reference syntax in overrides is deprecated (used by: @types/node, rollup, vite). Define the version in a catalog and reference it with the "catalog:" protocol instead. See https://pnpm.io/catalogs
[WARN] The "$" version reference syntax in overrides is deprecated (used by: @types/node, rollup, vite). Define the version in a catalog and reference it with the "catalog:" protocol instead. See https://pnpm.io/catalogs
Scope: all 42 workspace projects
[WARN] There are cyclic workspace dependencies: /home/hiroshi/code/others/vitest-release-pipeline-poc/packages/browser, /home/hiroshi/code/others/vitest-release-pipeline-poc/packages/vitest, /home/hiroshi/code/others/vitest-release-pipeline-poc/packages/browser-playwright
✓ Lockfile passes supply-chain policies (verified 7m ago)
Lockfile is up to date, resolution step is skipped
Packages: +20
++++++++++++++++++++
Progress: resolved 20, reused 20, downloaded 0, added 20, done
Done in 1.3s using pnpm v11.6.0
$ tsx scripts/release.ts
Bumping versions in packages: package.json, packages/browser/package.json, packages/browser-playwright/package.json, packages/browser-webdriverio/package.json, packages/coverage-istanbul/package.json, packages/coverage-v8/package.json, packages/expect/package.json, packages/pretty-format/package.json, packages/snapshot/package.json, packages/browser-preview/package.json, packages/spy/package.json, packages/ui/package.json, packages/mocker/package.json, packages/utils/package.json, packages/web-worker/package.json, packages/vitest/package.json 
```

```sh
$  VITE_TEST_WATCHER_DEBUG=false PUBLISH_DRY_RUN=true PUBLISH_BRANCH=main pnpm publish-ci 5.0.0-beta.5
[WARN] The "$" version reference syntax in overrides is deprecated (used by: @types/node, rollup, vite). Define the version in a catalog and reference it with the "catalog:" protocol instead. See https://pnpm.io/catalogs
[WARN] The "$" version reference syntax in overrides is deprecated (used by: @types/node, rollup, vite). Define the version in a catalog and reference it with the "catalog:" protocol instead. See https://pnpm.io/catalogs
Scope: all 42 workspace projects
[WARN] There are cyclic workspace dependencies: /home/hiroshi/code/others/vitest-release-pipeline-poc/packages/browser, /home/hiroshi/code/others/vitest-release-pipeline-poc/packages/vitest, /home/hiroshi/code/others/vitest-release-pipeline-poc/packages/browser-playwright
✓ Lockfile passes supply-chain policies (verified 7m ago)
Lockfile is up to date, resolution step is skipped
Packages: +20
++++++++++++++++++++
Progress: resolved 20, reused 20, downloaded 0, added 20, done
Done in 1.2s using pnpm v11.6.0
$ tsx scripts/publish-ci.ts 5.0.0-beta.5
== DRY RUN ==
Staging version '5.0.0-beta.5' with tag 'beta'
[WARN] The "$" version reference syntax in overrides is deprecated (used by: @types/node, rollup, vite). Define the version in a catalog and reference it with the "catalog:" protocol instead. See https://pnpm.io/catalogs
📦 @vitest/pretty-format@5.0.0-beta.5 → https://registry.npmjs.org/
[WARN] Skip staging @vitest/pretty-format@5.0.0-beta.5 (dry run)
📦 @vitest/spy@5.0.0-beta.5 → https://registry.npmjs.org/
[WARN] Skip staging @vitest/spy@5.0.0-beta.5 (dry run)
📦 @vitest/utils@5.0.0-beta.5 → https://registry.npmjs.org/
[WARN] Skip staging @vitest/utils@5.0.0-beta.5 (dry run)
📦 @vitest/expect@5.0.0-beta.5 → https://registry.npmjs.org/
[WARN] Skip staging @vitest/expect@5.0.0-beta.5 (dry run)
📦 @vitest/mocker@5.0.0-beta.5 → https://registry.npmjs.org/
[WARN] 10 other warnings
📦 @vitest/snapshot@5.0.0-beta.5 → https://registry.npmjs.org/
📦 @vitest/browser@5.0.0-beta.5 → https://registry.npmjs.org/
📦 vitest@5.0.0-beta.5 → https://registry.npmjs.org/
📦 @vitest/browser-playwright@5.0.0-beta.5 → https://registry.npmjs.org/
📦 @vitest/browser-preview@5.0.0-beta.5 → https://registry.npmjs.org/
📦 @vitest/browser-webdriverio@5.0.0-beta.5 → https://registry.npmjs.org/
📦 @vitest/coverage-istanbul@5.0.0-beta.5 → https://registry.npmjs.org/
📦 @vitest/coverage-v8@5.0.0-beta.5 → https://registry.npmjs.org/
📦 @vitest/web-worker@5.0.0-beta.5 → https://registry.npmjs.org/
📦 @vitest/ui@5.0.0-beta.5 → https://registry.npmjs.org/
+ @vitest/pretty-format@5.0.0-beta.5 (would stage)
+ @vitest/spy@5.0.0-beta.5 (would stage)
+ @vitest/utils@5.0.0-beta.5 (would stage)
+ @vitest/expect@5.0.0-beta.5 (would stage)
[WARN] 11 other warningseta.5 (would stage)
+ @vitest/snapshot@5.0.0-beta.5 (would stage)
+ @vitest/browser@5.0.0-beta.5 (would stage)
+ vitest@5.0.0-beta.5 (would stage)
+ @vitest/browser-playwright@5.0.0-beta.5 (would stage)
+ @vitest/browser-preview@5.0.0-beta.5 (would stage)
+ @vitest/browser-webdriverio@5.0.0-beta.5 (would stage)
+ @vitest/coverage-istanbul@5.0.0-beta.5 (would stage)
+ @vitest/coverage-v8@5.0.0-beta.5 (would stage)
+ @vitest/web-worker@5.0.0-beta.5 (would stage)
+ @vitest/ui@5.0.0-beta.5 (would stage)
```

hi-ogawa:
> We would also need to backport this to v4 and v3, right? Because the release workflow needs to be identical

Yes, I'll backport it. Slightly scary updating pnpm there but we'll see.

sheremet-va (approved):

PR #10554, opened 2026-06-09
refactor!: separate config resolution from the server creation

### LINKED ISSUE #6912: Create only a single Vite server even if browser mode is enabled

### LINKED ISSUE #8572: Possible problem when using vitest node api in conjunction with alias and browser
### Describe the bug

I'm experimenting a migration to Vitest and encountered a problem when using the Vitest Node API in conjunction with alias and browser.
The example repository shows the problem when running the `npm run test-api-browser` script and Vitest reports the following error when trying to resolve an import:
```bash
(!) Failed to run dependency scan. Skipping dependency pre-bundling. Error: The following dependencies are imported but could not be resolved:

  @alias_root/src/sum.js (imported by /Users/doberkofler/MyDev/ljs_app/trunk_test/periscope/test/sum-alias.test.ts)

Are they installed?
    at file:///Users/doberkofler/MyDev/ljs_app/trunk_test/periscope/node_modules/vite/dist/node/chunks/dep-M_KD0XSK.js:11130:33
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async file:///Users/doberkofler/MyDev/ljs_app/trunk_test/periscope/node_modules/vite/dist/node/chunks/dep-M_KD0XSK.js:34823:15
10:04:36 PM [vite] Internal server error: Failed to resolve import "@alias_root/src/sum.js" from "test/sum-alias.test.ts". Does the file exist?
  Plugin: vite:import-analysis
  File: /Users/doberkofler/MyDev/ljs_app/trunk_test/periscope/test/sum-alias.test.ts:2:18
  1  |  import { test, expect } from "vitest";
  2  |  import { sum } from "@alias_root/src/sum.js";
     |                       ^
  3  |  test("sum", () => expect(sum(1, 2)).toBe(3));
  4  |  
      at TransformPluginContext._formatLog (file:///Users/doberkofler/MyDev/ljs_app/trunk_test/periscope/node_modules/vite/dist/node/chunks/dep-M_KD0XSK.js:31527:43)
      at TransformPluginContext.error (file:///Users/doberkofler/MyDev/ljs_app/trunk_test/periscope/node_modules/vite/dist/node/chunks/dep-M_KD0XSK.js:31524:14)
      at normalizeUrl (file:///Users/doberkofler/MyDev/ljs_app/trunk_test/periscope/node_modules/vite/dist/node/chunks/dep-M_KD0XSK.js:29996:18)
      at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
      at async file:///Users/doberkofler/MyDev/ljs_app/trunk_test/periscope/node_modules/vite/dist/node/chunks/dep-M_KD0XSK.js:30054:32
      at async Promise.all (index 1)
      at async TransformPluginContext.transform (file:///Users/doberkofler/MyDev/ljs_app/trunk_test/periscope/node_modules/vite/dist/node/chunks/dep-M_KD0XSK.js:30022:4)
      at async EnvironmentPluginContainer.transform (file:///Users/doberkofler/MyDev/ljs_app/trunk_test/periscope/node_modules/vite/dist/node/chunks/dep-M_KD0XSK.js:31325:14)
      at async loadAndTransform (file:///Users/doberkofler/MyDev/ljs_app/trunk_test/periscope/node_modules/vite/dist/node/chunks/dep-M_KD0XSK.js:26407:26)
 ✓  chromium  test/sum-relative.test.ts (1 test) 2ms
```

Interestingly only this specific use case not work and others do work as expected:
- The test file `sum-relative.test.ts` does not use an alias and always works
- The script `npm run test-api-jsdom` uses `jsom` instead of the `browser` environment and works
- The script `npm run test-config` uses a config file and the same tests run as expected

### Reproduction

Use the repository: https://github.com/doberkofler/vitest-issue

### System Info

```shell
System:
    OS: macOS 15.6.1
    CPU: (8) x64 Intel(R) Core(TM) i7-8559U CPU @ 2.70GHz
    Memory: 315.08 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.8.0 - ~/.nvm/versions/node/v24.8.0/bin/node
    npm: 11.6.0 - ~/.nvm/versions/node/v24.8.0/bin/npm
    bun: 1.2.19 - ~/.bun/bin/bun
    Watchman: 2025.09.08.00 - /usr/local/bin/watchman
  Browsers:
    Chrome: 140.0.7339.133
    Safari: 18.6
  npmPackages:
    @vitest/browser: 3.2.4 => 3.2.4 
    playwright: 1.55.0 => 1.55.0 
    vitest: 3.2.4 => 3.2.4
```

### Used Package Manager

npm

### Validations

sheremet-va:
This PR:

- Reworks `resolveConfig` API to fully resolve the config first
- Passed down the config to `createServer` from vite, and creates a Vitest instance from it
- Adds a "PluginHarness" API to use internally for plugins that need `vitest` access after the config is resolved

Fixes https://github.com/vitest-dev/vitest/issues/6912
Fixes https://github.com/vitest-dev/vitest/issues/8572


packages/browser/src/client/tester/runner.ts
```diff
@@ -240,7 +240,7 @@ export function createBrowserRunner(
         await rpc().onAfterSuiteRun({
           coverage,
           testFiles: files.map(file => file.name),
-          environment: '__browser__',
+          environment: 'client',
```
hi-ogawa, line 243: human nit

```suggestion
          environment: this.viteEnvironment,
```

packages/vitest/src/node/plugins/cliOverride.ts
```diff
@@ -0,0 +1,33 @@
+import type { Plugin } from 'vite'
+import type { ResolvedBrowserOptions } from '../types/browser'
+import type { UserConfig } from '../types/config'
+import { deepMerge } from '@vitest/utils/helpers'
+import { mergeConfig } from 'vite'
+
+export function CliOverride(cliOptions: UserConfig): Plugin {
+  return {
+    // The CLI plugin overwrites config values with CLI options, making them
+    // avalable in the next plugin. We have to do this via plugins because of watch mode.
```
hi-ogawa: typo

```suggestion
    // available in the next plugin. We have to do this via plugins because of watch mode.
```

packages/browser/src/node/commands/fs.ts
```diff
@@ -18,8 +18,8 @@ function assertFileAccess(path: string, project: TestProject) {
 }
 
 function assertWrite(path: string, project: TestProject) {
-  if (!project.config.browser.api.allowWrite || !project.vitest.config.api.allowWrite) {
-    throw new Error(`Cannot modify file "${path}". File writing is disabled because server is exposed to the internet, see https://vitest.dev/config/browser/api.`)
+  if (!project.vitest.config.api.allowWrite) {
```
hi-ogawa: Similar to https://github.com/vitest-community/vitest-webdriverio/pull/1, we cannot drop per-project config part, but replace with unified one:

- `project.config.browser.api` -> `project.config.api`

```suggestion
  if (!project.config.api.allowWrite || !project.vitest.config.api.allowWrite) {
```

packages/vitest/src/node/plugins/api.ts
```diff
@@ -0,0 +1,72 @@
...
+      handler(viteConfig) {
+        // The watcher is set on the resolved config directly because returning
+        // `watch: null` from the `config` hook is a no-op: Vite's config merge
+        // drops `null` overrides, so a user-provided `server.watch` object wins.
+        const server = viteConfig.server
+        const watch = globalConfig?.watch ?? viteConfig.test?.watch
+        if (!watch) {
+          // disable the builtin watcher when Vitest is not in --watch mode
+          server.watch = null
+        }
```
hi-ogawa: For root project, `globalConfig` is undefined so watcher is disabled on plain `vitest` cli by default. I reproduced locally with

- run `pnpm -C examples/basic test`
- edit test file
- test doesn't re-run

packages/vitest/src/node/plugins/server.ts
hi-ogawa, line 1: nit. was confused about the naming `VitestConfigApi` and seeing watcher related logic. Can we split or rename perhaps?

packages/vitest/src/node/plugins/api.ts
hi-ogawa: Also on the other hand for workspace project, the server may get non null watcher since `WorkspaceVitestPlugin` includes `VitestConfigApi` with root level `globalConfig`.

hi-ogawa (commented):
Surfacing first pass review with local agent that spotted some real issues. Still digging in myself.

packages/browser/src/client/tester/runner.ts
sheremet-va: I would prefer to hardcode it. `viteEnvironment` is useless in browser mode, it's always a `client`

hi-ogawa: Okay, no problem at all 😃

packages/vitest/src/node/core.ts
```diff
@@ -194,48 +206,19 @@ export class Vitest {
...
+   * @internal
    */
-  get snapshot(): SnapshotManager {
-    assert(this._snapshot, 'snapshot', 'snapshot manager')
-    return this._snapshot
+  async _start(config: ResolvedViteConfig): Promise<void> {
+    this._setRootConfig(config)
+    await this._attachRootServer()
+    await this._attachProjectServers()
+    this._harness.setVitest(this)
```
hi-ogawa: Some `harness.setVitest` looks redundant since constructor already does.

packages/vitest/src/node/plugins/server.ts
sheremet-va: It's more of a `VitestConfigServer`, I guess. It changes the `config.server` option

packages/vitest/src/node/plugins/api.ts
sheremet-va: "You are absolutely right!"

test/coverage-test/test/cjs-dependency.test.ts
```diff
@@ -10,7 +10,7 @@ test('excludes Vite transforms done for CJS dependency', async () => {
     },
   }, undefined, {
     cacheDir: fileURLToPath(new URL('./node_modules/.vite', import.meta.url)),
-    optimizeDeps: { include: ['@vitest/cjs-lib', '/Users/ari/Git/vitest/test/browser/cjs-lib'] },
+    optimizeDeps: { include: ['@vitest/cjs-lib'] },
```
hi-ogawa, line 13: nice cleanup 😄

test/unit/test/exports.test.ts
hi-ogawa, line 1: Anything to document for new `vitest/node` API?

test/e2e/test/reporters/merge-reports.test.ts
```diff
@@ -351,24 +351,28 @@ test.for([
...
+        ],
+        headless: true,
+      }
     }
+    return baseConfig
   }
 
   const result = await runVitest({
-    ...baseConfig,
+    ...baseConfig(),
```
hi-ogawa, line 375: Good change, but from the curiosity, did it have specific issue with reusing provider instance?

test/workspaces/space_1/test/env-injected.spec.ts
```diff
@@ -10,10 +10,10 @@ test('dev is injected', () => {
...
 
 test('env variable is assigned', () => {
-  // we override it with "local" in .env.local, but dotenv prefers the root .env
-  // this is consistent with how Vite works
-  expect(import.meta.env.VITE_MY_TEST_VARIABLE).toBe('core')
-  expect(process.env.VITE_MY_TEST_VARIABLE).toBe('core')
+  // the project's own .env.local wins over the inherited root value, like every
+  // other option a project can override
+  expect(import.meta.env.VITE_MY_TEST_VARIABLE).toBe('local')
+  expect(process.env.VITE_MY_TEST_VARIABLE).toBe('local')
```
hi-ogawa, line 16: Looks like a breaking change but sounds like a fix and real use case won't probably hit such usage.

packages/vitest/src/node/core.ts
```diff
@@ -115,34 +116,64 @@ export class Vitest {
...
+   */
+  public config!: ResolvedConfig
+  /**
+   * Resolved global vite config.
+   */
+  public viteConfig!: ResolvedViteConfig
+  /**
+   * Global Vite's dev server instance.
+   */
+  public vite!: ViteDevServer
```
hi-ogawa, line 132: `viteConfig` looks like a new thing. Looks reasonable but feels redundant with what people have conventionally did the same with `ctx.vite.config`. Just curious if this addition has some bigger intent.

packages/vitest/src/node/cli/cli-config.ts
```diff
@@ -109,7 +109,7 @@ export const cliOptionsConfig: VitestCLIOptions = {
   },
   api: {
     argument: '[port]',
-    description: `Specify server port. Note if the port is already being used, Vite will automatically try the next available port so this may not be the actual port the server ends up listening on. If true will be set to ${defaultPort}`,
+    description: `Specify server port. Note if the port is already being used, Vite will automatically try the next available port so this may not be the actual port the server ends up listening on. If true will be set to ${defaultPort} or ${defaultBrowserPort} in browser mode`,
```
hi-ogawa, line 112: Just realized that `api` and `browser.api` unification means that now `--ui` and browser mode get served by the single server. We do have coverage in `test/ui` and tested locally and looks fine, but just highlighting.

For example, the default port of `pnpm -C packages/ui test:ui --ui` does change from `localhost:51204` (ui default) to `localhost:63315` (browser deafult)

hi-ogawa (commented):
Some progress. Mostly just some minor observations.

test/e2e/test/reporters/merge-reports.test.ts
sheremet-va: Yes, it broke it. This is expected and already happens if you provide the same project to instances

test/workspaces/space_1/test/env-injected.spec.ts
sheremet-va: Yep, I could keep the previous behaviour, but decided that it was very dumb 🕊️

packages/vitest/src/node/core.ts
sheremet-va: `viteConfig` is available before `vite` is defined. I am also thinking about doing something like `config.$viteConfig` instead (maybe it's better to have a single config for the engine API in the future)

packages/vitest/src/node/cli/cli-config.ts
sheremet-va: Yeah, changing the port is an interesting side effect 🤔 But I think it only affects vitest repo?

packages/vitest/src/node/core.ts
sheremet-va: Yeah, this one is redundant. The others are required to keep the logic the same for `init` and `restart` phases

packages/vitest/src/node/plugins/api.ts
sheremet-va: Looks like a bug, yeah

packages/browser/src/node/plugin.ts
hi-ogawa: `environment.name !== 'client'` filter is now redundant since `contribution.plugins` is guarded by client `applyEnvironment`.

packages/vitest/src/node/plugins/browserLoader.ts
```diff
@@ -0,0 +1,130 @@
...
+        holder.contribution = contribution
+        const browserConfig = await contribution.config(viteConfig, harness)
+        return browserConfig
+      },
+      applyToEnvironment(environment) {
+        const contribution = holder.contribution
+        if (contribution && environment.name === 'client') {
+          // `post` browser plugins are injected by `vitest:browser:loader:post`
+          // instead, so they run after the `post` plugins of the main pipeline
+          // (e.g. `vitest:mocks` hoisting) rather than at this `pre` position.
```
hi-ogawa: As far as I can tell, currently `post` from `contribution.plugins` is only mocker `dynamicImportPlugin` (`vitest:browser:esm-injector`) and that's only affecting critically.

Probably won't hurt calling out explicitly to help navigating the complexity in the future:

```suggestion
          // `post` browser plugins are injected by `vitest:browser:loader:post`
          // instead, so they run after the `post` plugins of the main pipeline
          // rather than at this `pre` position. For example, the mocker's
          // `vitest:browser:esm-injector` must run after `vitest:mocks`, which is
          // added by the main pipeline and is not part of `contribution.plugins`.
```

packages/vitest/src/node/plugins/index.ts
```diff
@@ -1,293 +1,70 @@
...
+    SsrRunnerFixerPlugin(harness),
+    {
+      name: 'vitest:ui-injector',
+      enforce: 'post',
+      async configResolved(config) {
+        if (config.test.ui) {
+          await harness.packageInstaller.ensureInstalled('@vitest/ui', resolve(config.root), harness.version)
+          const uiPlugin = (await import('@vitest/ui')).default(harness)
+          // @ts-expect-error mutate readonly
+          config.plugins.push(uiPlugin)
```
hi-ogawa: This is good. I've been also thinking that we don't really need plugin mechanism for UI since it's just `configureServer` function. We could just export one function and use it like contribution pattern in the future.

packages/browser/src/node/index.ts
```diff
@@ -25,91 +34,483 @@ export function defineBrowserCommand<T extends unknown[]>(
...
+      return new ParentBrowserProject({ config, vitest }, '/')
+    },
+    setupRpc(parent) {
+      setupBrowserRpc(parent as ParentBrowserProject, mockerRegistry)
+    },
+  }
+
+  contribution.plugins = [
+    ...BrowserPlugin(contribution),
+    interceptorPlugin({ registry: mockerRegistry }),
```
hi-ogawa, line 349: `interceptorPlugin` has `configureServer` but is basically no-op because the plugin is added through contribution `applyToEnvironment` (and Vite doesn't support `configureServer` there). It looks like previously the same `configureServer` did run and registered unused ws handler.

Another minor comment to clarify:

```suggestion
    // this plugin's `configureServer` is ignored since it's added through `applyToEnvironment`
    interceptorPlugin({ registry: mockerRegistry }),
```

hi-ogawa (commented):
Omg, I just realized that there are literally zero change in provider packages. Absolute beauty! 😃 

I think I've digested browser related change and new structure looks great (and probably I've understand better than the previous shape, which has been kinda fuzzy). Another review stopping point and will continue review later for core side diff.

test/e2e/test/reporters/merge-reports.test.ts
```diff
@@ -502,6 +506,7 @@ async function getSerializedModuleGraph(ctx: Vitest) {
           ctx,
           projectName,
           file.filepath,
+          file.viteEnvironment,
```
hi-ogawa, line 509: This is the only place using newly added `getModuleGraph` optional `viteEnvironment` argument, but this test passes without it. Is it fine to remove?

packages/vitest/src/integrations/vi.ts
```diff
@@ -779,12 +779,12 @@ function createVitest(): VitestUtils {
...
-      if (_envBooleans.includes(name)) {
-        env[name] = value ? '1' : ''
-      }
-      else if (value === undefined) {
+      if (value === undefined) {
         delete env[name]
       }
+      else if (_envBooleans.includes(name)) {
+        env[name] = value ? '1' : ''
+      }
```
hi-ogawa, line 787: This looks like a small independent fix and without specific test. Worth adding the coverage here?

https://github.com/vitest-dev/vitest/blob/334edef920b35cbb222019090b8abde788a68bff/test/unit/test/stubs.test.ts#L109

packages/vitest/src/node/projects/resolveProjects.ts
```diff
@@ -25,145 +38,307 @@ import { initializeProject, TestProject } from '../project'
...
+}
+
+/**
+ * Aggregate `optimizeDeps` for each browser Vite server across every project
+ * that shares it, then merge the result into the resolved Vite config's `client`
+ * environment. Runs after resolution (so project `include`/`setupFiles` are
+ * known) and before server creation (so `createViteServer` reuses the mutated
+ * resolved config).
+ */
+async function applyBrowserOptimizeDeps(
```
hi-ogawa, line 201: Probably minor nit. The comment around the whole chain sound a bit noisy:
- `applyBrowserOptimizeDeps`
- `contribution.resolveOptimizeDeps`
- `mergeBrowserOptimizeDeps`

The higher level idea looks simpler than this and basically to combine:
- collect all browser test files and use it as `optimizeDeps.entries` (only dynamic part)
- and statically hard-coded `optimizeDeps.include/exclude` by `@vitest/browser` provided through `contribution.resolveOptimizeDeps` mechanism.

Probably I'd tell AI to inline `mergeBrowserOptimizeDeps` to eliminate the trivial indirection. This may be just my taste so non blocking though 😃

hi-ogawa (commented):
I think I've read through all. The end of my review!

packages/vitest/src/integrations/vi.ts
sheremet-va: I think this breaks without the change to how we setup `env`. After this PR we no longer populate `proces.env` in the main process based on the value of `resolvedConfig.env`

packages/vitest/src/node/projects/resolveProjects.ts
sheremet-va: I agree 👍 It's done now

packages/vitest/src/integrations/vi.ts
sheremet-va: Added the test

test/e2e/test/reporters/merge-reports.test.ts
sheremet-va: I think the opposite is better - I added it to more places

packages/vitest/src/node/plugins/index.ts
sheremet-va: Do you think is should be part of this PR? I agree that it could be simplified like that

packages/vitest/src/node/plugins/api.ts
sheremet-va: Should be fixed now

sheremet-va:
Ready for review again

packages/vitest/src/node/core.ts
hi-ogawa: I'd prefer `ctx.viteConfig` over `ctx.config.$viteConfig` if comparing two.
Also if these aren't absolute need yet, then we don't need to make public API yet then?

packages/vitest/src/node/plugins/index.ts
hi-ogawa: No, this looks fine separately in the future.

packages/vitest/src/integrations/vi.ts
hi-ogawa: Side note I forgot to mention. I thought this may be fixing https://github.com/vitest-dev/vitest/issues/10525 but it seems unrelated.

hi-ogawa (approved):
lgtm!

PR #10561, opened 2026-06-10
fix(spy): preserve class mock prototype methods

### LINKED ISSUE #10553: vi.spyOn(...).mockImplementation(class {...}) produces instances with no prototype methods
### Describe the bug

When mocking a **class constructor** with `vi.spyOn(obj, 'key').mockImplementation(SomeClass)`, instances created from the mocked constructor lose all their methods — `instance.method` is `undefined`.

I am using `vi.spyOn()` to replace a class held on an object property, passing a replacement class (which `extends` the original and overrides a method) to `.mockImplementation()`. I expect `new obj.Key()` to return an instance of the replacement class, so that `instance.method()` runs the overridden implementation and returns `"mocked"`. What actually happens is that the created instance has no methods at all: `typeof instance.method` is `"undefined"` instead of `"function"`.

This is explicitly documented as a supported use case. The [`vi.spyOn` docs](https://vitest.dev/api/vi.html#vi-spyon) state:

> If the spying method is a class definition, the mock implementations have to use the `function` or the `class` keyword:

The reproduction below uses the `class` keyword exactly as the documentation prescribes, yet the resulting instance has no methods.

```ts
import { expect, test, vi } from "vitest";

test("vi.spyOn with class mockImplementation bug", () => {
  class OriginalClass {
    method() {
      return "original";
    }
  }

  const myObj = { TestClass: OriginalClass };

  vi.spyOn(myObj, "TestClass").mockImplementation(
    class MockClass extends OriginalClass {
      method() {
        return "mocked";
      }
    },
  );

  const instance = new myObj.TestClass();

  expect(typeof instance.method).toBe("function"); // ❌ received "undefined"
  expect(instance.method()).toBe("mocked");
});
```

**Expected:** the test passes — `instance.method` is a function returning `"mocked"`.

**Actual:** the test fails:

```
AssertionError: expected 'undefined' to be 'function' // Object.is equality

Expected: "function"
Received: "undefined"

 ❯ test.spec.ts:24:34
     24|   expect(typeof instance.method).toBe("function");
       |                                  ^
```

### Reproduction

cf https://github.com/divdavem/vitest-class-mock-bug

```bash
git clone https://github.com/divdavem/vitest-class-mock-bug
cd vitest-class-mock-bug
npm install
npm test
```

### System Info

```shell
System:
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  Binaries:
    Node: 24.16.0 - $HOME/n/bin/node
    npm: 11.13.0 - $HOME/n/bin/npm
  npmPackages:
    vitest: ^4.1.5 => 4.1.8
```

### Used Package Manager

npm

### Validations

hi-ogawa:
### Description

 - closes https://github.com/vitest-dev/vitest/issues/10553

#### Root cause

Vitest constructs class mocks by running the original/mock implementation constructor, but it uses the mock wrapper as `newTarget`. That means the constructor body and class fields run, but the allocated object inherits from `MockClass.prototype` instead of `ActualClass.prototype`. The gap is easiest to see by splitting construction into two conceptual phases: object allocation with a prototype, then constructor/class-field initialization.

```js
class ActualClass {
  constructor(...ctorArgs) {
    // In normal construction, `this` already inherits from ActualClass.prototype.
    // So `this.method1` is visible here.
  }

  method1() {} // lives on ActualClass.prototype
  method2 = () => {} // assigned as an own property during construction
}

const actualInstance = new ActualClass(...ctorArgs)
//
// This is equivalent to:
//
//   Reflect.construct(ActualClass, ctorArgs, ActualClass)
//
// which then semantically almost same as:
//
//   const actualInstance = Object.create(ActualClass.prototype)
//   // Creates this prototype chain:
//   //   actualInstance -> ActualClass.prototype -> Object.prototype
//   // This is why:
//   //   actualInstance.method1                 // found via ActualClass.prototype
//   //   actualInstance instanceof ActualClass  // true
//
//   ActualClass[[constructorBody]].apply(actualInstance, ctorArgs)
//   // Runs constructor/class-field initialization on `actualInstance`.
//   // This is why:
//   //   actualInstance.method2                 // own property assigned during construction

const MockClass = vi.fn(ActualClass)
const mockInstance = new MockClass(...ctorArgs)
//
// Vitest currently implements this as:
//
//   Reflect.construct(ActualClass, ctorArgs, MockClass)
//
// which mostly means:
//
//   const mockInstance = Object.create(MockClass.prototype)
//   // Creates this prototype chain:
//   //   mockInstance -> MockClass.prototype -> Object.prototype
//   //
//   // This is why:
//   //   mockInstance.method1                 // missing: ActualClass.prototype is not in the chain
//   //   mockInstance instanceof MockClass    // true
//   //   mockInstance instanceof ActualClass  // false
//
//   ActualClass[[constructorBody]].apply(mockInstance, ctorArgs)
//   // Runs constructor/class-field initialization on `mockInstance`.
//   // This is why:
//   //   mockInstance.method2                 // own property assigned during construction
```

---

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/spy/src/index.ts
+++ b/packages/spy/src/index.ts
@@ -500,6 +500,21 @@ function createMock(
       try {
         if (new.target) {
           returnValue = Reflect.construct(implementation, args, new.target)
+          const implementationPrototype = (implementation as any).prototype
+          if (
+            prototypeMembers.length === 0
+            && new.target?.prototype
+            && implementationPrototype
+            && Object.getPrototypeOf(returnValue) === new.target.prototype
+          ) {
+            const { properties, descriptors } = getAllProperties(implementationPrototype)
+            for (const property of properties) {
+              if (property === 'constructor' || property in returnValue) {
+                continue
+              }
+              Object.defineProperty(returnValue, property, descriptors[property]!)
+            }
+          }
 
           // jest calls this before the implementation, but we have to resolve this _after_
           // because we cannot do it before the `Reflect.construct` called the custom implementation.
--- a/packages/spy/src/index.ts
+++ b/packages/spy/src/index.ts
@@ -504,9 +504,9 @@ function createMock(
           if (
             // prototypeMembers is used by automocking case and we want to skip that case.
             prototypeMembers.length === 0
-            // TODO: no idea what
-            && new.target?.prototype
+            && new.target.prototype
             && implementationPrototype
+            // skip when constructor returns value
             && Object.getPrototypeOf(returnValue) === new.target.prototype
           ) {
             const { properties, descriptors } = getAllProperties(implementationPrototype)
--- a/packages/spy/src/index.ts
+++ b/packages/spy/src/index.ts
@@ -502,7 +502,9 @@ function createMock(
           returnValue = Reflect.construct(implementation, args, new.target)
           const implementationPrototype = (implementation as any).prototype
           if (
+            // prototypeMembers is used by automocking case and we want to skip that case.
             prototypeMembers.length === 0
+            // TODO: no idea what
             && new.target?.prototype
             && implementationPrototype
             && Object.getPrototypeOf(returnValue) === new.target.prototype
--- a/test/unit/test/mocking/vi-fn.test.ts
+++ b/test/unit/test/mocking/vi-fn.test.ts
@@ -809,6 +809,24 @@ describe('vi.fn() implementations', () => {
     expect(new Mock().method()).toBe('mocked')
   })
 
+  test('vi.fn(class) does not copy prototype methods when constructor returns an object', () => {
+    const returned = { custom: true }
+    const Mock = vi.fn(class {
+      constructor() {
+        return returned as any
+      }
+
+      method() {
+        return 'mocked'
+      }
+    })
+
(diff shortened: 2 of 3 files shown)


packages/spy/src/index.ts
```diff
@@ -500,6 +500,23 @@ function createMock(
       try {
         if (new.target) {
           returnValue = Reflect.construct(implementation, args, new.target)
+          const implementationPrototype = (implementation as any).prototype
```
sheremet-va, line 503: Is there any explanation why `Reflect.construct` just doesn't work? Does it behave the same outside of Vitest?

hi-ogawa: It's is quite mind-bending, but I got the gist. Added root cause in the PR description. I'm not settled with the fix. There is still some divergence and probably need to pick some trade off.

sheremet-va: Ah, I remember this now. I don't think there is a good solution for this, all of them break something, it just needs to be documented

hi-ogawa:
Closing in favor of 
- https://github.com/vitest-dev/vitest/pull/10902

PR #10567, opened 2026-06-10
fix(vitest): render non-Error causes from env setup so the actual diagnostic is not dropped

### LINKED ISSUE #10557: A non-`Error` value thrown from a test environment's `setup()` is dropped from the "Failed to start worker" report (regression from 3.x)
### Describe the bug

When a custom test environment's `setup()` rejects, the pool wraps the failure as
`[vitest-pool]: Failed to start <pool> worker for test files …` and attaches the
original value via `{ cause }`. However, the reporter only prints the cause when it
is an **`Error` instance**:

- `throw new Error("…")`  → report includes `Caused by: Error: …` ✅
- `throw "…"` / `throw { … }` / a wasm-bindgen `JsValue` → the cause is **silently dropped**; the user sees only the generic, location-free pool message ❌

In Vitest 3.x the same non-`Error` throw was surfaced directly (`Unknown Error: <value>`),
so this is a regression. It appears to have come in with the worker-start error handling
added in #9337 / v4.0.17 (which correctly fixed the indefinite hang from #9271, but the
new wrapper only renders `Error`-typed causes).

This bites real-world environments that initialize native/wasm resources in `setup()`
and reject with a non-`Error` value. For example `@stacks/clarinet-sdk`'s
`vitest-environment-clarinet` calls a wasm-bindgen `initSimnet()` in `setup()`; when a
contract fails to compile, the wasm rejects with a `JsValue`, and on v4 the actual
diagnostic  is completely lost — you only see `Failed to start forks worker`.

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-ldzxobzz?file=throwing-env.js

**`package.json`**
```json
{
  "name": "vitest-env-setup-error-swallowed",
  "version": "1.0.0",
  "type": "module",
  "private": true,
  "scripts": {
    "test": "vitest run"
  },
  "devDependencies": {
    "vitest": "^4.1.8"
  }
}
```

**`throwing-env.js`**
```js
// Minimal custom test environment whose async setup() rejects with a NON-Error value.
// Real-world analogue: wasm-bindgen / native addons reject with a plain JsValue
// (e.g. @stacks/clarinet-sdk's simnet init), not an `Error` instance.
export default {
  name: "throwing-env",
  transformMode: "ssr",
  async setup() {
    // A proper `throw new Error("...")` IS surfaced by vitest (as `Caused by:`).
    // A non-Error value is silently dropped — only the generic pool error remains.
    throw "the real reason setup failed (a non-Error value)";
  },
};
```

**`vitest.config.js`**
```js

export default defineConfig({
  test: {
    environment: "./throwing-env.js",
    // pool: "forks" is the default; the message mentions "forks worker".
  },
});
```

**`test/example.test.js`**
```js
import { test, expect } from "vitest"; test("noop", () => { expect(1).toBe(1); });
```

Run `npm install` and `npm test`.

(Swap the `throw` for `throw new Error("…")` to see the cause correctly printed — that
contrast shows the wrapping/`cause` machinery is present; only non-`Error` rendering is
missing.)

### System Info

```shell
System:
    OS: macOS 26.5.1
    CPU: (14) arm64 Apple M4 Pro
    Memory: 120.22 MB / 48.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 25.1.0 - /Users/brice/.nvm/versions/node/v25.1.0/bin/node
    npm: 11.6.2 - /Users/brice/.nvm/versions/node/v25.1.0/bin/npm
  Browsers:
    Chrome: 149.0.7827.55
    Safari: 26.5
  npmPackages:
    vitest: ^4.1.8 => 4.1.8
```

### Used Package Manager

npm

### Validations

spokodev:
## What

When a custom test environment's `setup()` rejects with a non-`Error` value (a string, a plain object, or a wasm-bindgen `JsValue`), the actual diagnostic is silently dropped and the user only sees the generic, location-free `[vitest-pool]: Failed to start <pool> worker for test files …` from the pool wrapper.

Vitest 3.x rendered the same non-`Error` throw as `Unknown Error: <value>`, so the silent drop is a regression. After this PR the cause surfaces correctly for Error, plain-object, and primitive throws.

Closes #10557.

## Why / root cause

The pool wrapper in `packages/vitest/src/node/pools/pool.ts:127` already attaches the original failure as `{ cause: error }` on the synthesised "Failed to start … worker" Error. The reporter in `packages/vitest/src/node/printError.ts` then descended into `cause` only when it satisfied a narrow shape check:

```ts
if (typeof e.cause === 'object' && e.cause && 'name' in e.cause) {
  (e.cause as any).name = `Caused by: ${(e.cause as any).name}`
  printErrorInner(e.cause, project, {...})
}
```

That filter passes for `new Error(...)` (object + truthy + has `name`) but excludes:

- `throw "string"` — `typeof "string" === "object"` is `false`, so the branch is skipped entirely.
- `throw { reason: '...' }` — object + truthy, but no `name` property, so `'name' in cause` is `false`.
- wasm-bindgen `JsValue` — object + truthy, but no `name` (and in some bindings, `'name' in cause` throws on the proxy).

In all three cases the value reaches `printError` via the cause chain but never reaches `printErrorInner`, so the actual diagnostic is lost.

The regression most likely landed alongside the worker-start error wrapping in #9337 (which correctly fixed the indefinite-hang behaviour from #9271). The wrapper is well-formed; only the reporter's filter is too narrow.

The fix normalises the cause into a printable shape *before* calling `printErrorInner` — which already knows how to handle primitives via its `isPrimitive` branch:

- Primitive cause → wrap as `{ name: 'Caused by', message: String(cause), stack: String(cause) }`.
- Object without `name` → spread + `name: 'Caused by'`.
- Object with `name` → preserve the existing behaviour of prefixing `name` with `Caused by:`.

`printErrorInner` then renders the message exactly as it would for a top-level non-`Error` throw, mirroring Vitest 3.x behaviour. Real-world impact: native/wasm environments (e.g. `@stacks/clarinet-sdk`'s `vitest-environment-clarinet`) reject from their wasm `setup()` with a JsValue when contract compilation fails — the failure reason now surfaces instead of being lost behind "Failed to start forks worker".

## Tests added

New e2e test at `test/e2e/test/env-setup-non-error-cause.test.ts` using `runInlineTests` from `test/test-utils`. Three cases, each booting Vitest against a fixture environment whose `setup()` throws a different shape:

1. **`non-Error string thrown from env setup surfaces in stderr`** — the bug case from #10557. Without this PR the string never appears in stderr; with it, the `the real reason setup failed (a non-Error value)` substring shows up.
2. **`non-Error plain object thrown from env setup surfaces in stderr`** — boundary H-case (wasm JsValue analogue). Without the PR the object is dropped entirely; with it, the object's keys / serialised value appear in stderr.
3. **`Error thrown from env setup still renders with Caused by prefix`** — happy-path regression. Confirms the existing `Caused by: Error: …` rendering is unchanged for proper Error instances.

Verified locally:

```bash
pnpm -F vitest build
CI=true pnpm test env-setup-non-error-cause   # 3/3 passing after fix
CI=true pnpm test print-error                 # 2/2 passing (no regression on existing reporter tests)
CI=true pnpm test pool.test                   # 16/16 passing (no regression on pool tests)
```

Before the fix: tests 1 and 2 fail with `Expected to contain '…the real reason…'` / `Expected to match /wasm-init-failed|reason/`. Test 3 already passed on master.

## Repro

Minimal repro from the issue (https://stackblitz.com/edit/vitest-dev-vitest-ldzxobzz):

```js
// throwing-env.js
export default {
  name: 'throwing-env',
  viteEnvironment: 'ssr',
  async setup() {
    throw 'the real reason setup failed (a non-Error value)'
  },
}
```

```js
// vitest.config.js
export default defineConfig({
  test: { environment: './throwing-env.js' },
})
```

`pnpm test` shows only the generic `Failed to start forks worker` message on master; with this PR the cause string appears under `Caused by:`.

## Validation

- `pnpm -F vitest build` — green
- `CI=true pnpm test env-setup-non-error-cause` — 3/3
- `CI=true pnpm test print-error` — 2/2 (no regression)
- `CI=true pnpm test pool.test` — 16/16 (no regression)
- `pnpm lint:fix` — clean

## AI usage disclosure

Drafted with Claude assistance. The root-cause analysis (the `'name' in cause` narrow filter and its three failure modes), the fix design, the test surface choice (`runInlineTests` instead of unit-testing `printError`), and this PR text are mine. I verified the red baseline → fix → green baseline loop locally and have read every line of the diff. I will engage genuinely with any review feedback within the 3-day window described in `AGENTS.md`.

### DIFF
--- a/packages/vitest/src/node/printError.ts
+++ b/packages/vitest/src/node/printError.ts
@@ -246,9 +246,24 @@ function printErrorInner(
     )
   }
 
-  if (typeof e.cause === 'object' && e.cause && 'name' in e.cause) {
-    (e.cause as any).name = `Caused by: ${(e.cause as any).name}`
-    printErrorInner(e.cause, project, {
+  if (e.cause != null) {
+    // Render the cause regardless of its shape. Previously this branch only
+    // fired for object-causes with a `name` property, so non-`Error` throws
+    // (strings, plain objects, wasm-bindgen JsValues) propagated through the
+    // pool wrapper's `{ cause }` but were silently dropped by the reporter —
+    // see https://github.com/vitest-dev/vitest/issues/10557.
+    let cause: any = e.cause
+    if (typeof cause !== 'object' || cause === null) {
+      const causeStr = String(cause)
+      cause = { name: 'Caused by', message: causeStr, stack: causeStr }
+    }
+    else if (!('name' in cause)) {
+      cause = { ...cause, name: 'Caused by' }
+    }
+    else {
+      cause.name = `Caused by: ${cause.name}`
+    }
+    printErrorInner(cause, project, {
       showCodeFrame: false,
       logger: options.logger,
       parseErrorStacktrace: options.parseErrorStacktrace,
--- a/packages/vitest/src/node/printError.ts
+++ b/packages/vitest/src/node/printError.ts
@@ -247,11 +247,6 @@ function printErrorInner(
   }
 
   if (e.cause != null) {
-    // Render the cause regardless of its shape. Previously this branch only
-    // fired for object-causes with a `name` property, so non-`Error` throws
-    // (strings, plain objects, wasm-bindgen JsValues) propagated through the
-    // pool wrapper's `{ cause }` but were silently dropped by the reporter —
-    // see https://github.com/vitest-dev/vitest/issues/10557.
     let cause: any = e.cause
     if (typeof cause !== 'object' || cause === null) {
       const causeStr = String(cause)
--- a/test/e2e/test/env-setup-non-error-cause.test.ts
+++ b/test/e2e/test/env-setup-non-error-cause.test.ts
@@ -1,0 +1,99 @@
+import { expect, test } from 'vitest'
+import { runInlineTests } from '../../test-utils'
+
+// https://github.com/vitest-dev/vitest/issues/10557 — when a test environment's
+// `setup()` rejects with a non-`Error` value (string, plain object, wasm-bindgen
+// JsValue, etc.), the pool wrapper still attaches the value via `{ cause }`, but
+// the reporter only renders causes that are `Error` instances. Without this
+// fix, the actual diagnostic is silently dropped and the user only sees the
+// generic `[vitest-pool]: Failed to start … worker for test files …` message.
+
+test('non-Error string thrown from env setup surfaces in stderr', async () => {
+  const { stderr } = await runInlineTests({
+    'throwing-env.js': `
+export default {
+  name: 'throwing-env',
+  transformMode: 'ssr',
+  async setup() {
+    throw 'the real reason setup failed (a non-Error value)'
+  },
+}
+    `,
+    'vitest.config.js': `
+import { defineConfig } from 'vitest/config'
+export default defineConfig({
+  test: {
+    environment: './throwing-env.js',
+  },
+})
+    `,
+    'example.test.js': `
+import { test, expect } from 'vitest'
+test('noop', () => { expect(1).toBe(1) })
+    `,
+  })
+
+  expect(stderr).toContain('Failed to start')
+  expect(stderr).toContain('the real reason setup failed (a non-Error value)')
+})
+
+test('non-Error plain object thrown from env setup surfaces in stderr', async () => {
+  const { stderr } = await runInlineTests({
+    'throwing-env.js': `
+export default {
+  name: 'throwing-env',
+  transformMode: 'ssr',
+  async setup() {
+    throw { reason: 'wasm-init-failed', code: 42 }
+  },
+}
+    `,
+    'vitest.config.js': `
+import { defineConfig } from 'vitest/config'
+export default defineConfig({
+  test: {
+    environment: './throwing-env.js',
+  },
+})
+    `,
+    'example.test.js': `
+import { test, expect } from 'vitest'
+test('noop', () => { expect(1).toBe(1) })
+    `,
+  })
+
+  expect(stderr).toContain('Failed to start')
+  // The object's contents (or at least the keys) must surface — without the
+  // fix the entire object is dropped.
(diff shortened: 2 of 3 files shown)


spokodev:
Hi, this is Yarik. Spoko-dev.
Of course, I'm human.
I've read everything carefully and agree with everything. The fix is at
`printError.ts:249` where the `'name' in cause` filter let Error instances
through and silently dropped strings / plain objects / wasm JsValues.
Of course, I also use Claude for writing drafts. But rest assured, I'm
human:) I'm writing with my hands on the keyboard, everything is fine,
don't worry:)

packages/vitest/src/node/printError.ts
```diff
@@ -246,9 +246,24 @@ function printErrorInner(
     )
   }
 
-  if (typeof e.cause === 'object' && e.cause && 'name' in e.cause) {
-    (e.cause as any).name = `Caused by: ${(e.cause as any).name}`
-    printErrorInner(e.cause, project, {
+  if (e.cause != null) {
```
sheremet-va, line 249: I like the solution, but the comment is over explaining itself. No reason to have anything here

test/e2e/test/env-setup-non-error-cause.test.ts
```diff
@@ -0,0 +1,99 @@
+import { expect, test } from 'vitest'
+import { runInlineTests } from '../../test-utils'
+
+// https://github.com/vitest-dev/vitest/issues/10557 — when a test environment's
```
sheremet-va: More over explanaition

sheremet-va (changes requested):

spokodev:
Removed both, plus a third in the same style further down the test.

sheremet-va (approved):

PR #10583, opened 2026-06-12
fix(ui)!: harden UI API access

### LINKED ISSUE #10499: Harden browser mode and UI API access

hi-ogawa:
### Description

- Closes https://github.com/vitest-dev/vitest/issues/10499

As described in v5 breaking change doc, this changes so that Vitest UI now requires authentication for accessing `/__vitest__/`.  First time user authenticates through accessing `/__vitest__/?token=...`, then middleware check token and persists it as cookie, then later `/__vitest__/` access will check that cookie. To make required `token` stable between vitest invocation, it's now generated once in `.local/share/vitest/.vitest-secret-token` (and alike on macos and windows).

The rational of the change is that, even though the Vitest API usage is purely for devtools, what's available on the network through `localhost:51204/__vitest__/` is immediately RCE (remote code execution) and it's reasonable to protect from arbitrary network access as a basic security posture. In retrospect, if we were to have this (together with https://github.com/vitest-dev/vitest/pull/10522), the past vulnerabilities related to API didn't exist. 

This is also partly because, intended or not, our security policy frames this way https://github.com/vitest-dev/vitest/blob/main/SECURITY.md#what-vitest-does-not-trust

> What Vitest Does Not Trust:
> Network data and untrusted clients The integration built on top of Vite's dev server must treat all inbound network requests as potentially hostile. This includes malformed requests. Clients may be outside the developer's intended environment because of port-forwarding, shared networks, or accidental exposure to the internet.

(Not sure yet though if I want to go through all vulnerability patch process. We can say this is a matter of interpretation since we do call out exposing allowExec/allowWrite in non localhost shouldn't be done in documentation.)

TODO
- [x] test
- [x] docs
- [x] cleanup

Example:

https://github.com/user-attachments/assets/5934bda1-4aa1-4d8f-8800-cf60212b8bc1

### Tests

### Documentation

### Changesets


hi-ogawa:
Still brainstorming and cookie might indeed work alright. Here is possible flow:

1. Vitest prints an auth URL once:
   `UI auth URL: http://localhost:51204/__vitest__/?token=...`
2. Browser/open behavior still targets plain:
   `http://localhost:51204/__vitest__/`
3. Plain UI without cookie shows a small blocked page:
   “Vitest UI is protected. Open the auth URL printed in the terminal to continue.”
4. Visiting `?token=...` validates token, sets an HttpOnly-ish session cookie, then redirects to plain `/__vitest__/`.
5. Subsequent access uses cookie, so bookmarks/history/devtools mostly show the clean URL.

test/ui/test/ui.spec.ts
```diff
@@ -33,6 +41,36 @@ test.describe('ui', () => {
     await testCrossOriginAccess(page, pageUrl)
   })
 
+  test('blocks unauthenticated ui html requests', async ({ request }) => {
```
hi-ogawa, line 47: TODO: verify token file isn't served by vite via fs.deny.

packages/vitest/src/node/create.ts
```diff
@@ -75,6 +75,17 @@ export async function createVitest(
...
+        // Open the clean UI base URL (without `?token=`) rather than the
+        // authenticated URL printed by the logger. On macOS, `openBrowser`
+        // reuses an existing tab whose URL matches via substring and reloads
+        // it (Vite's `bin/openChrome.js`). Since the 302 redirect strips the
+        // token, an already-authenticated tab lives at the clean URL, so
+        // opening the clean URL matches and reloads it; opening the token URL
+        // would never match and would spawn a new tab on every restart.
+        server.config.server.open = ctx.config.uiBase
+        server.openBrowser()
+      }
```
hi-ogawa, line 103: TODO: we can actually know the first time token is generated and auto-open `/__vitest__/?token=...` just for that case.

packages/browser/src/node/plugin.ts
```diff
@@ -402,6 +404,8 @@ export default (parentServer: ParentBrowserProject, base = '/'): Plugin[] => {
         }
         viteConfig.server.fs ??= {}
         viteConfig.server.fs.allow = viteConfig.server.fs.allow || []
+        viteConfig.server.fs.deny ??= []
+        viteConfig.server.fs.deny.push(API_TOKEN_FS_DENY)
```
hi-ogawa: not just this, but technically browser mode should try to inherit all deny rule from parent?

packages/ui/node/index.ts
```diff
@@ -94,11 +106,25 @@ export default (ctx: Vitest): Vite.Plugin => {
...
+                res.statusCode = 302
+                res.setHeader('Set-Cookie', `${UI_TOKEN_COOKIE}=${encodeURIComponent(ctx.config.api.token)}; Path=${base}; HttpOnly; SameSite=Strict`)
+                res.setHeader('Location', base)
+                res.end()
+                return
+              }
+              const cookieToken = getCookie(req, UI_TOKEN_COOKIE)
+              if (cookieToken !== ctx.config.api.token) {
+                res.statusCode = 403
+                res.end('Use the Vitest UI URL printed by the server.')
```
hi-ogawa: TODO: sounds vague.

packages/ui/node/index.ts
```diff
@@ -9,6 +10,17 @@ import { distClientRoot } from './paths'
 
 export { distClientRoot }
 
+// TODO: verify with proper lib
+const UI_TOKEN_COOKIE = 'vitest-ui-token'
```
hi-ogawa: TODO: use some cookie lib?

sheremet-va (approved):

PR #10589, opened 2026-06-12
fix: collect in-source tests when the module is cached (fix #10577)

### LINKED ISSUE #10577: Some in-source tests are not detected when `isolate: false` and `maxWorkers: 1` are set
### Describe the bug

In a project using in-source testing, running tests with both `isolate: false` and `maxWorkers: 1` causes some of the source files containing in-source tests to fail with `Error: No test suite found in file ...`. With only one of `isolate: false` or `maxWorkers: 1` set, all tests are detected correctly.

Which files fail varies from run to run. From what I have observed, once a test file has run, the in-source tests in its dependencies (direct or transitive) are no longer detected.

### Reproduction

https://github.com/koutaro-masaki/vitest-insource-repro

### System Info

```shell
OS: macOS 26.5.1
CPU: (14) arm64 Apple M4 Pro
Memory: 143.63 MB / 48.00 GB
Shell: 5.9 - /bin/zsh

Vitest: 4.1.8 (also reproduced on 5.0.0-beta.4)
```

### Used Package Manager

pnpm

### Validations

koutaro-masaki:
### Description

When in-source testing is used with both `isolate: false` and `maxWorkers: 1`, files that were already imported by another test file fail with `"No test suite found"`.
  
This happens because tests are registered as a side effect of importing the file. With this configuration, all test files share the same worker's module cache, so files that are already imported are not re-evaluated during collection and no tests are registered.

This was a known problem in the mode that runs tests with Node.js's native loader (`viteModuleRunner: false`), where a workaround is implemented in native.ts and nodejsWorkerLoader.ts in #9210, but it was not covered when running with the default Vite module runner.

This PR applies the same workaround to the default Vite module runner.

Resolves #10577

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/vitest/src/runtime/runners/test.ts
+++ b/packages/vitest/src/runtime/runners/test.ts
@@ -72,7 +72,11 @@ export class TestRunner implements VitestTestRunner {
         },
       },
       () => {
-        if (!this.viteModuleRunner) {
+        const shouldInvalidate = !this.viteModuleRunner
+        // this is require for in-source tests to be invalidated if
+        // one of the files already imported it in --maxWorkers=1 --no-isolate
+          || (source === 'collect' && this.workerState.evaluatedModules.getModuleById(filepath)?.evaluated)
+        if (shouldInvalidate) {
           filepath = `${filepath}?vitest=${Date.now()}`
         }
         return this.moduleRunner.import(filepath)
--- a/packages/vitest/src/runtime/runners/test.ts
+++ b/packages/vitest/src/runtime/runners/test.ts
@@ -58,11 +58,9 @@ export class TestRunner implements VitestTestRunner {
   }
 
   importFile(filepath: string, source: VitestRunnerImportSource): unknown {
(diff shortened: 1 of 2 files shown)


sheremet-va (approved):

sheremet-va (changes requested):
Hm, actually I am not sure how this is treated by plugins for non-ts/js files. If it is a `.vue` file, for example, does it recognize it? (It usually checks for `.vue` extension, but with a query it's harder) Or should we strip the query for transform, but keep it for cache? `viteModuleRunner: false` has its own handling for this (search for `vitest=` in the repo)

koutaro-masaki:
I added a repro test and confirmed that the previous fix breaks plugins that check the file extension (like the vue plugin).
I should have invalidated the module cache from the start, like `source === 'setup'` already does. The fix now does that instead of appending the `?vitest` query.

sheremet-va (approved):

PR #10621, opened 2026-06-19
feat(reporters)!: write json and junit reporter output files to `.vitest` by default

### LINKED ISSUE #10619: Make `json` and `junit` reporters default output use `.vitest` as the report artifact root
## Problem

Vitest now documents `.vitest/` as the shared location for generated reports and artifacts. The `blob` reporter already follows this (`.vitest/blob/`), but `json` and `junit` do not.

Currently, when no `outputFile` is configured:

- `json` prints the report to stdout (`json.ts:231-248`)
- `junit` prints the report to stdout (`junit.ts:249-273`)

This is inconsistent with the `.vitest/` convention and with the `blob` reporter, and means these reporters produce no on-disk artifact by default.

In practice, "print to stdout" pushes users into awkward workarounds to get the file they actually wanted. Common patterns in the wild:

- **Redirect to a file, then upload as a CI artifact:** `vitest run --reporter=json > vitest-report.json` then upload it (e.g. [mizchi/flaker `.github/workflows/ci.yml`](https://github.com/mizchi/flaker/blob/274105c68e099adc7637c85f8ef9d7133362224f/.github/workflows/ci.yml#L95), which redirects → uploads → re-imports the report in a later job).
- **Pipe into a post-processor:** `vitest --reporter=json | jq '.numFailedTests'`.
- **Redirect a bench/report run:** `vitest bench --run --reporter=json > benchmark-results.json` (e.g. [thibmeu/act-ts](https://github.com/thibmeu/act-ts/blob/ec3195d6113d35e630565ec29f0bdf316907dcf1/.github/workflows/ci.yml#L113)).

A scoped `.vitest/` default serves all of these *better*, with zero config:

- Artifact upload: just run `vitest --reporter=json` and upload `.vitest/json/output.json` — no redirect.
- Post-analysis: `jq '.numFailedTests' .vitest/json/output.json` — and it's re-runnable without re-running the tests, since the artifact persists.

So the stdout default isn't a feature these workflows rely on; it's the limitation they work around. Moving to `.vitest/` removes the workaround.

## Goal

When the `json` and `junit` reporters use their default output (no `outputFile`), write to a scoped directory under `.vitest/`:

```txt
.vitest/
  json/
    output.json
  junit/
    output.xml
```

This mirrors the existing `blob` behavior (`blob.ts:88-116`): explicit `outputFile` always wins; otherwise fall back to `createReport(scope)` + a default filename.

## Desired behavior

- No `outputFile` configured:
  - `json` writes `.vitest/json/output.json`
  - `junit` writes `.vitest/junit/output.xml`
- Explicit `outputFile` (string or `outputFile.json` / `outputFile.junit`) is respected and still resolved against `root`, unchanged.

## Breaking change

The default changes from "print to stdout" to "write to file". This affects:

- The documented piping workflow `vitest --reporter=json > out.json` (`docs/guide/reporters.md:63-67`, `319`, `423`).
- Any CI that consumes json/junit from stdout.

**Decision: drop stdout entirely; no sentinel opt-in.** Every workflow stdout enabled (artifact upload, `| jq` post-processing) is served better by reading `.vitest/json/output.json` after a plain `vitest --reporter=json`. Adding an `outputFile: 'stdout'` sentinel would just preserve the worse path. Explicit `outputFile` (file path) is unchanged.

## Scope / notes

- Target the next major (aligns with other v5 reporter changes).
- Sibling effort to #10240 (html). Per #10240, top-level `.vitest/index.html` + `.vitest/ui/` are reserved for the html reporter; other reporters keep scoped subdirectories such as `.vitest/json/` and `.vitest/junit/`.
- Coverage migration is intentionally out of scope.
- Docs (`docs/guide/reporters.md`) and the e2e suites (`test/e2e/test/reporters/json.test.ts`, `junit.test.ts`, which mostly assert on `stdout`) will need updating.

hi-ogawa:
### Description

- closes https://github.com/vitest-dev/vitest/issues/10619

### Tests

### Documentation

### Changesets


docs/guide/migration.md
```diff
@@ -165,6 +165,15 @@ Vitest no longer serves the browser orchestrator UI from a bare `/__vitest_test_
 
 If you manually opened the browser preview by copying the Vite server URL or visiting `/__vitest_test__/` directly, use the URL opened or printed by Vitest instead.
 
+### Generated Reports and Artifacts Use the `.vitest` Directory
+
+Vitest now uses a single `.vitest` directory at the project root as the shared artifact root. The `json` and `junit` reporters now write to this directory by default instead of printing to stdout:
```
sheremet-va: What is wrong with printing to stdout? Why do we remove this completely?

hi-ogawa: For the use case I understand, they look effectively useless and misguided. See https://github.com/vitest-dev/vitest/issues/10619

> In practice, "print to stdout" pushes users into awkward workarounds to get the file they actually wanted. Common patterns in the wild:
> 
> - **Redirect to a file, then upload as a CI artifact:** `vitest run --reporter=json > vitest-report.json` then upload it (e.g. [mizchi/flaker `.github/workflows/ci.yml`](https://github.com/mizchi/flaker/blob/274105c68e099adc7637c85f8ef9d7133362224f/.github/workflows/ci.yml#L95), which redirects → uploads → re-imports the report in a later job).
> - **Pipe into a post-processor:** `vitest --reporter=json | jq '.numFailedTests'`.
> - **Redirect a bench/report run:** `vitest bench --run --reporter=json > benchmark-results.json` (e.g. [thibmeu/act-ts](https://github.com/thibmeu/act-ts/blob/ec3195d6113d35e630565ec29f0bdf316907dcf1/.github/workflows/ci.yml#L113)).
> 
> A scoped `.vitest/` default serves all of these *better*, with zero config:
> 
> - Artifact upload: just run `vitest --reporter=json` and upload `.vitest/json/output.json` — no redirect.
> - Post-analysis: `jq '.numFailedTests' .vitest/json/output.json` — and it's re-runnable without re-running the tests, since the artifact persists.
> 
> So the stdout default isn't a feature these workflows rely on; it's the limitation they work around. Moving to `.vitest/` removes the workaround.

hi-ogawa: Did we discuss maybe supporting `outputFile: "-"` to mean stdout?

test/e2e/test/rollup-error.test.ts
hi-ogawa, line 1: TODO: Don't remember why this test needs junit. Could be replaced with just stderr?

docs/guide/migration.md
sheremet-va: > So the stdout default isn't a feature these workflows rely on; it's the limitation they work around. Moving to .vitest/ removes the workaround.

This seems weird - if it's so much better to keep a file, why don't they use `--outputFile` and read from it? Just because it's not a default?

hi-ogawa: > > So the stdout default isn't a feature these workflows rely on; it's the limitation they work around. Moving to .vitest/ removes the workaround.
> 
> This seems weird - if it's so much better to keep a file, why don't they use `--outputFile` and read from it? Just because it's not a default?

This is just my educated guess and basically my biased opinion.

hi-ogawa: Also I prefer the file because direct stdout writing such as `process.stdout.write` can break stdout redirection approach (or console.log in main process like global setup).

docs/guide/reporters.md
```diff
@@ -60,7 +60,13 @@ export default defineConfig({
...
 ## Reporter Output
 
-By default, Vitest's reporters will print their output to the terminal. When using the `json` or `junit` reporters, you can instead write your tests' output to a file by including an `outputFile` [configuration option](/config/outputfile) either in your Vite configuration file or via CLI. The `html` reporter writes a report directory instead; see its [`outputDir`](#html-reporter) option.
+By default, Vitest's reporters print their output to the terminal. The `json`, `junit` and `html` reporters instead write to a scoped location under `.vitest/`:
+
+- `json` writes `.vitest/json/output.json`
+- `junit` writes `.vitest/junit/output.xml`
+- `html` writes `.vitest/index.html`
+
+The `json` and `junit` locations can be overridden with the `outputFile` [configuration option](/config/outputfile) either in your Vite configuration file or via CLI; the `html` reporter uses its [`outputDir`](#html-reporter) option instead.
```
AriPerkkio: Do we mention `outputFile` reporter option anywhere?

https://github.com/vitest-dev/vitest/blob/c4090147918c53835860ece57a04c8c0c1456458/packages/vitest/src/node/reporters/junit.ts#L50-L51

I would rather start pointing users to configure this via reporter options and remove whole top level `outputFile` in future.

packages/vitest/src/node/reporters/junit.ts
```diff
@@ -256,20 +256,21 @@ export class JUnitReporter implements Reporter {
...
-        await fs.writeFile(this.fileFd, `${text}\n`)
+    this.baseLog = async (text: string) => {
+      if (!this.fileFd) {
+        this.fileFd = await fs.open(this.reportFile!, 'w+')
       }
-    }
-    else {
-      this.baseLog = async (text: string) => this.ctx.logger.log(text)
+
+      await fs.writeFile(this.fileFd, `${text}\n`)
```
AriPerkkio: Out-of-scope of this PR, but probably good refactoring for future:
I'm not exactly sure why this reporter was built to stream its results into the output file, while everything is still handled in `onTestFinished`. Maybe to avoid OOMs? Anyway, we could move this whole `fileId` initialization in `onTestFinished` and keep it in function scope there.

AriPerkkio (dismissed):
Looks good to me. I don't think stdout as JSON or Junit output ever made sense. The output could have contained Node's and Vitest's deprecation warnings, so parsing the XML/JSON from there would be difficult.

sheremet-va:
> I don't think stdout as JSON or Junit output ever made sense

warnings are in stderr though. jest actually forwards every log to stderr so that json reporter is parsable

hi-ogawa:
> > I don't think stdout as JSON or Junit output ever made sense
> 
> warnings are in stderr though. jest actually forwards every log to stderr so that json reporter is parsable

How comprehensive is this though? Does any user code on main process stdout gets collected and redirected to stderr like global setup?

AriPerkkio:
> Does any user code on main process stdout gets collected and redirected to stderr like global setup?

No, not even from test runner workers. If you have `process.stdout.write('\n\n\n\nHello world\n\n\n\n')` in a test file, the final JSON will be broken:

image

docs/guide/reporters.md
hi-ogawa: It looks like there was never a mention of `['json', { outputFile: ... }]` nor junit. Will add some in the doc.

I agree that curent `ouptutFile` option system are rather scattered, but not sure we can drop whole top level one. Before dropping it, probably we need reporter option from cli like `--reporter.json.outputFile=...` somehow.

hi-ogawa: tweaked docs and created cli reporter option request https://github.com/vitest-dev/vitest/issues/10650

AriPerkkio (approved):

sheremet-va:
This argument feels weird to me thought. I am not against changing the default, but I am unsure why we need to take away the stdout output. It feels like a natural programmatic way to use it in a terminal - just because user can manually break it doesn’t mean it’s vitest fault. We allow custom reporters even though they can throw an error.

sheremet-va:
I would keep it as —stdout flag or something. Maybe —reporter.json.stdout like Ari wanted for other reporter options

AriPerkkio (changes requested):
If reading JSON and XML from terminal output is expected use case, let's keep it. I cannot see when it would be useful though.

hi-ogawa:
I also suggested this thing `outputFile: "-"` to mean stdout https://github.com/vitest-dev/vitest/pull/10621#discussion_r3441333449. This feels too stretched overload, so will go with new stdout option then.

test/e2e/test/rollup-error.test.ts
```diff
@@ -3,35 +3,35 @@ import { expect, test } from 'vitest'
 import { rolldownVersion } from 'vitest/node'
 
 test('rollup error node', async () => {
-  const { stdout } = await runVitest({
+  const { stderr } = await runVitest({
     root: './fixtures/rollup-error',
     environment: 'node',
-    reporters: ['junit'],
+    reporters: ['default'],
```
hi-ogawa: can remove explicit reporters?

test/e2e/test/reporters/utils.ts
```diff
@@ -49,6 +55,16 @@ export function getContext(): Context {
     snapshot: {
       summary: { added: 100, _test: true },
     } as any,
+    createReport: (scope: string) => {
```
hi-ogawa: can revert now in favor of `stdout: true`.

test/e2e/test/reporters/json.test.ts
hi-ogawa, line 1: add one trivial test for stdout case here and junit.test.ts

sheremet-va (dismissed):

docs/guide/reporters.md
```diff
@@ -60,7 +60,13 @@ export default defineConfig({
...
 ## Reporter Output
 
-By default, Vitest's reporters will print their output to the terminal. When using the `json` or `junit` reporters, you can instead write your tests' output to a file by including an `outputFile` [configuration option](/config/outputfile) either in your Vite configuration file or via CLI. The `html` reporter writes a report directory instead; see its [`outputDir`](#html-reporter) option.
+By default, Vitest's reporters print their output to the terminal. The `json`, `junit` and `html` reporters instead write to a scoped location under `.vitest/`:
+
+- `json` writes `.vitest/json/output.json`
+- `junit` writes `.vitest/junit/output.xml`
+- `html` writes `.vitest/index.html`
+
+The `json` and `junit` locations can be overridden with the `outputFile` [configuration option](/config/outputfile) in your Vite configuration file or via CLI. The `html` reporter uses its [`outputDir`](#html-reporter) option instead.
```
AriPerkkio:
```suggestion
The `json` and `junit` locations can be overridden with the `outputFile` [configuration option](/config/outputfile) in your Vitest configuration file or via CLI. The `html` reporter uses its [`outputDir`](#html-reporter) option instead.
```

AriPerkkio (dismissed):

AriPerkkio (approved):

PR #10631, opened 2026-06-21
fix: avoid double-prefixing node-only builtins (fix #10630)

### LINKED ISSUE #10630: Vitest double-prefixes `node:`-only built-ins in `resolve.external`
### Describe the bug

Vitest 4 sets `config.resolve.external` by spreading `node:module.builtinModules` and then prefixing every entry with `node:`:

https://github.com/vitest-dev/vitest/blob/29c364d50e45eb6fdec0c4c4eea8bf3949e56b43/packages/vitest/src/node/plugins/runnerTransform.ts#L100-L106

On newer Node versions, `builtinModules` includes some built-ins only in already-prefixed form, for example:

```txt
node:sea
node:sqlite
node:test
node:test/reporters
```
Related documentation: https://nodejs.org/api/modules.html#built-in-modules

This makes Vitest generate invalid double-prefixed entries:

```txt
node:node:sea
node:node:sqlite
node:node:test
node:node:test/reporters
```

This surfaced downstream with `@cloudflare/vite-plugin@1.42.1`, which still rejects Vitest startup because `resolve.external` contains these non-built-in `node:node:*` entries.

### Reproduction

On Node `v24.14.1`:

```js
import { builtinModules } from "node:module";

console.log(
  builtinModules
    .flatMap((m) => [m, `node:${m}`])
    .filter((m) => m.includes("sea") || m.includes("sqlite") || m.includes("test")),
);
```

Output includes:

```txt
node:sea
node:sqlite
node:test
node:test/reporters
node:node:sea
node:node:sqlite
node:node:test
node:node:test/reporters
```

### Expected behavior

Vitest should not add a second `node:` prefix when a built-in already starts with `node:`.

For example:

```ts
config.resolve.external = [
  ...builtinModules,
  ...builtinModules.map((m) => (m.startsWith("node:") ? m : `node:${m}`)),
];
```

### System Info

- Vitest: `4.1.9`
- Node: `v24.14.1`
- OS: macOS arm64
- Package manager: Bun

### Additional context

This appears related in spirit to #8925, but this report is specifically about Node’s own `node:`-only built-ins being double-prefixed.

Filed by GPT-5.5.

koriyoshi2041:
## Problem

Node can expose some builtins only in `node:` form, for example `node:sea`, `node:sqlite`, `node:test`, and `node:test/reporters`. Vitest currently prefixes every entry from `builtinModules`, so those entries become invalid `node:node:*` values in `resolve.external`.

## Fix

Make the builtin external list generation idempotent for entries that already start with `node:` while preserving the existing unprefixed + prefixed coverage for normal builtins.

Closes #10630.

## Test

- `./node_modules/.bin/vitest run test/builtin-external.test.ts --project threads`
- `./node_modules/.bin/eslint packages/vitest/src/node/plugins/runnerTransform.ts test/unit/test/builtin-external.test.ts`
- `git diff --check`

## Risk

Low. The change only avoids adding a second `node:` prefix; it keeps existing builtin entries and their `node:` aliases in the external list.

Disclosure: assisted by Codex.

### DIFF
--- a/packages/vitest/src/node/plugins/runnerTransform.ts
+++ b/packages/vitest/src/node/plugins/runnerTransform.ts
@@ -4,6 +4,13 @@ import { normalize } from 'pathe'
 import { escapeRegExp } from '../../utils/base'
 import { resolveOptimizerConfig } from './utils'
 
+export function resolveBuiltinExternalModules(modules: readonly string[]): string[] {
+  return [
+    ...modules,
+    ...modules.map(m => m.startsWith('node:') ? m : `node:${m}`),
+  ]
+}
+
 export function ModuleRunnerTransform(): VitePlugin {
   let testConfig: NonNullable<UserConfig['test']>
   const noExternal: (string | RegExp)[] = []
@@ -97,10 +104,7 @@ export function ModuleRunnerTransform(): VitePlugin {
         }
 
         // remove Vite's externalization logic because we have our own (unfortunately)
-        config.resolve.external = [
-          ...builtinModules,
-          ...builtinModules.map(m => `node:${m}`),
-        ]
+        config.resolve.external = resolveBuiltinExternalModules(builtinModules)
 
         // by setting `noExternal` to `true`, we make sure that
         // Vite will never use its own externalization mechanism
--- a/test/unit/test/builtin-external.test.ts
+++ b/test/unit/test/builtin-external.test.ts
@@ -1,0 +1,23 @@
+import { expect, test } from 'vitest'
+import { resolveBuiltinExternalModules } from '../../../packages/vitest/src/node/plugins/runnerTransform'
+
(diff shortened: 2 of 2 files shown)


joaoclaudio82:
Independent red/green verification completed against the current `main` branch using Node.js 24 in a network-isolated Docker test run.

A regression test invoking the actual `ModuleRunnerTransform` hooks failed before this change because `resolve.external` contained `node:node:sea`, `node:node:sqlite`, `node:node:test`, and `node:node:test/reporters`.

After applying this PR's prefix guard and rebuilding `packages/vitest`, the same test passed:

```text
Test Files  1 passed (1)
Tests       1 passed (1)
Duration    167ms
```

The extracted `resolveBuiltinExternalModules` unit test in this PR is also a cleaner, lower-coupling version of the integration-style check I used. The fix matches the observed failure.

PR #10645, opened 2026-06-22
fix(spy): update static properties when mockImplementation is called with a class

### LINKED ISSUE #10574: class mock statics do not update after `mockImplementation`
### Describe the bug

Found the edge case while thinking about the fix https://github.com/vitest-dev/vitest/pull/10561 since copying extra thing logic is analogical.

Reproduction https://stackblitz.com/edit/vitest-dev-vitest-b4ogrg4r?file=repro.test.ts

```js
import { expect, test, vi } from 'vitest'

test('vi.fn(class) static properties follow mockImplementation', () => {
  class Original {
    static value = 'original'
  }

  class Replacement {
    static value = 'replacement'
    static replacementOnly = 'replacement-only'
  }

  const Mock = vi.fn(Original) as any

  expect(Mock.value).toBe('original')
  console.log(Mock.value) // -> 'original'

  Mock.mockImplementation(Replacement)
  console.log(Mock.value) // -> still 'original' because static properties are not updated
  console.log(Mock.replacementOnly); // undefined
})
```

Static properties are copied from the initial class/original class when the mock wrapper is created, but later `mockImplementation(class Replacement { static ... })` calls do not update the mock constructor's static surface.

Related history:

- https://github.com/vitest-dev/vitest/pull/7780 introduced static-property copying for spied functions/classes.
- https://github.com/vitest-dev/vitest/pull/8956 extended/fixed the behavior for `vi.fn(class ...)` initialization.

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-b4ogrg4r?file=repro.test.ts

### System Info

```shell
stackblitz
v5 beta
```

### Used Package Manager

npm

### Validations

lazerg:
Fixes #10574

`mockImplementation` with a class doesn't update static props — they stay frozen from when the mock was first created. Added a WeakMap to track which keys were copied, so each `mockImplementation` call removes the old static keys and copies the new ones.

### DIFF
--- a/packages/spy/src/index.ts
+++ b/packages/spy/src/index.ts
@@ -26,6 +26,9 @@ const MOCK_RESTORE = new Set<() => void>()
 // if you stored it before calling `mockClear` where it will be recreated
 const REGISTERED_MOCKS = new Set<Mock<Procedure | Constructable>>()
 const MOCK_CONFIGS = new WeakMap<Mock<Procedure | Constructable>, MockConfig>()
+// Tracks the set of static property keys that were copied from a class/function
+// implementation onto the mock, so they can be removed when the implementation changes.
+const MOCK_COPIED_STATIC_KEYS = new WeakMap<Mock<Procedure | Constructable>, Set<string | symbol>>()
 
 export function createMockInstance(options: MockInstanceOption = {}): Mock<Procedure | Constructable> {
   const {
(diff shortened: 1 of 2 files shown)


lazerg:
yes, it is me). i used claude for assisting, but i it is not "ai fully written pr" , i review and check

lazerg:
btw the failing checks (`Browsers: node-24, windows-latest` and `Build&Test: node-24, macos-latest`) also fail on `main` (run [27953929034](https://github.com/vitest-dev/vitest/actions/runs/27953929034))... 

pre-existing flakiness, unrelated to this change.

sheremet-va:
please read the comments in the issue first, we did not decide if this is a bug or an intentional behavior

PR #10646, opened 2026-06-22
fix(snapshot): support no-unsafe-eval CSP by evaluating snapshot files server-side

### LINKED ISSUE #10615: `getSnapshotData` swallows the parse error, making snapshot failures impossible to debug
### Describe the bug

In browser mode, when the page is served with a CSP that blocks `eval` (no `'unsafe-eval'`), external `.snap` files silently fail to load. The snapshot data ends up empty, which produces 1 failed snapshot and 1 obsolete snapshot. 

The `.snap` file parses fine in Node. The actual cause is never reported.

**NOTE** This happens only when `CI=true` is set, which is the default in GitHub Actions. There is no error in non-CI mode, which makes this even more confusing to debug.

## Why

`getSnapshotData` parses `.snap` files with `new Function("exports", contents)`, which CSP blocks. The resulting error is thrown away:

```js
// @vitest/snapshot/dist/index.js
try {
  snapshotContents = content;
  const populate = new Function("exports", snapshotContents);
  populate(data);
} catch {} // <- error discarded, data stays empty
```

With the `catch {}` in place there's nothing in the logs pointing at CSP/eval. Adding a `console.log(err)` there immediately surfaces the real error.

## Proposed solution
The `catch {}` should be removed, or at least the error should be logged. This would make snapshot failures in browser mode much easier to debug.

### Reproduction

https://github.com/vojty/vitest-browser-snapshot-error

`npm run test:ci`

```
 FAIL   browser (chromium)  src/repro.test.ts > repro > should match snapshot
Error: Snapshot `repro > should match snapshot 1` mismatched

Failure screenshot:
  - src/__screenshots__/repro.test.ts/repro-should-match-snapshot-1.png

 ❯ src/repro.test.ts:5:21
      3| describe("repro", () => {
      4|   it("should match snapshot", () => {
      5|     expect({ a: 1 }).toMatchSnapshot(); // external .snap → new Function → CSP blocks it
       |                     ^
      6|   });
      7| });

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯


  Snapshots  1 failed
             1 files obsolete

 Test Files  1 failed (1)
      Tests  1 failed (1)
   Start at  15:23:36
   Duration  1.64s (transform 0ms, setup 0ms, import 6ms, tests 105ms, environment 0ms)
```

`npm run test` (without `CI=true`) works fine

### System Info

```shell
System:
    OS: macOS 26.5.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 99.92 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.14.1 - /Users/tomas.vojtasek/.nvm/versions/node/v24.14.1/bin/node
    npm: 11.12.1 - /Users/tomas.vojtasek/.nvm/versions/node/v24.14.1/bin/npm
    pnpm: 10.10.0 - /Users/tomas.vojtasek/Library/pnpm/pnpm
    bun: 1.3.14 - /Users/tomas.vojtasek/.bun/bin/bun
    Watchman: 2026.03.30.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 149.0.7827.155
    Safari: 26.5
  npmPackages:
    @vitest/browser-playwright: ^4.1.9 => 4.1.9 
    vite: ^8.0.16 => 8.0.16 
    vitest: ^4.1.9 => 4.1.9
```

### Used Package Manager

npm

### Validations

lazerg:
### Description

In browser mode, when the page is served with a `no-unsafe-eval` CSP policy, `getSnapshotData` silently fails to parse `.snap` files because `new Function()` is blocked. The empty `catch {}` discards the error, leaving snapshot data empty and producing confusing mismatch failures with no hint about the real cause.

This PR introduces an optional `readSnapshotFileData` method on `SnapshotEnvironment` that returns already-evaluated snapshot data. When the environment provides it, `SnapshotState.create()` uses it instead of `readSnapshotFile + new Function`, so the browser tester runtime never has to eval. The browser's `VitestBrowserSnapshotEnvironment` implements the method by delegating to the server via RPC, where the file is read and evaluated safely in Node.js context.

Resolves #10615

### DIFF
--- a/packages/snapshot/src/types/environment.ts
+++ b/packages/snapshot/src/types/environment.ts
@@ -7,6 +7,7 @@ export interface SnapshotEnvironment {
   resolveRawPath: (testPath: string, rawPath: string) => Promise<string>
   saveSnapshotFile: (filepath: string, snapshot: string) => Promise<void>
   readSnapshotFile: (filepath: string) => Promise<string | null>
+  readSnapshotFileData?: (filepath: string) => Promise<Record<string, string> | null>
   removeSnapshotFile: (filepath: string) => Promise<void>
   processStackTrace?: (stack: ParsedStack) => ParsedStack
 }
--- a/packages/browser/src/client/tester/snapshot.ts
+++ b/packages/browser/src/client/tester/snapshot.ts
@@ -23,6 +23,10 @@ export class VitestBrowserSnapshotEnvironment implements SnapshotEnvironment {
     return rpc().readSnapshotFile(filepath)
   }
 
+  readSnapshotFileData(filepath: string): Promise<Record<string, string> | null> {
+    return rpc().readSnapshotFileData(filepath)
+  }
+
   saveSnapshotFile(filepath: string, snapshot: string): Promise<void> {
     return rpc().saveSnapshotFile(filepath, snapshot)
   }
--- a/packages/browser/src/node/rpc.ts
+++ b/packages/browser/src/node/rpc.ts
@@ -268,6 +268,28 @@ export function setupBrowserRpc(globalServer: ParentBrowserProject, defaultMocke
           }
           return fs.readFile(snapshotPath, 'utf-8')
         },
+        async readSnapshotFileData(snapshotPath) {
+          checkFileAccess(snapshotPath)
+          if (!existsSync(snapshotPath)) {
+            return null
+          }
+          const content = await fs.readFile(snapshotPath, 'utf-8')
+          const data = Object.create(null)
+          try {
+            // eslint-disable-next-line no-new-func
+            const populate = new Function('exports', content)
+            populate(data)
+          }
+          catch (cause) {
+            // surface corrupted snapshot files as a hard error instead of
+            // silently ignoring them
+            throw new Error(
+              `Invalid snapshot file, please manually fix or delete it: ${snapshotPath}`,
+              { cause },
+            )
+          }
+          return data
+        },
         async saveSnapshotFile(id, content) {
           checkFileAccess(id)
           if (!canWrite(project)) {
--- a/packages/browser/src/types.ts
+++ b/packages/browser/src/types.ts
@@ -33,6 +33,7 @@ export interface WebSocketBrowserHandlers {
   cancelCurrentRun: (reason: CancelReason) => void
   getCountOfFailedTests: () => number
   readSnapshotFile: (id: string) => Promise<string | null>
+  readSnapshotFileData: (id: string) => Promise<Record<string, string> | null>
   saveSnapshotFile: (id: string, content: string) => Promise<void>
   removeSnapshotFile: (id: string) => Promise<void>
   sendLog: (method: TestExecutionMethod, log: UserConsoleLog) => void
--- a/packages/snapshot/src/env/node.ts
+++ b/packages/snapshot/src/env/node.ts
@@ -1,6 +1,7 @@
 import type { SnapshotEnvironment, SnapshotEnvironmentOptions } from '../types'
 import { existsSync, promises as fs } from 'node:fs'
 import { basename, dirname, isAbsolute, join, resolve } from 'pathe'
+import { evaluateSnapshotFile } from '../port/utils'
 
 export class NodeSnapshotEnvironment implements SnapshotEnvironment {
   constructor(private options: SnapshotEnvironmentOptions = {}) {}
@@ -38,6 +39,14 @@ export class NodeSnapshotEnvironment implements SnapshotEnvironment {
       return null
     }
     return fs.readFile(filepath, 'utf-8')
+  }
+
+  async readSnapshotFileData(filepath: string): Promise<Record<string, string> | null> {
+    const content = await this.readSnapshotFile(filepath)
+    if (content == null) {
+      return null
+    }
+    return evaluateSnapshotFile(filepath, content)
   }
 
   async removeSnapshotFile(filepath: string): Promise<void> {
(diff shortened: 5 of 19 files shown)


lazerg:
the two windows failures aren't from this PR, both `Build&Test` and `Browsers` windows jobs fail the same way on https://github.com/vitest-dev/vitest/actions/runs/27978136425 which is a completely different branch. looks like pre-existing flakiness on windows

hi-ogawa (changes requested):
I think we can go all-in as breaking change like:
- require `readSnapshotFileData`
- surface `populate` error as hard error and say something like "invalid snapshot file, please manually fix or delete (path to snapshot file)"

---

new e2e should cover
- csp wihtout unsave eval
- new hard error populate behavior

packages/snapshot/src/types/environment.ts
```diff
@@ -7,6 +7,7 @@ export interface SnapshotEnvironment {
   resolveRawPath: (testPath: string, rawPath: string) => Promise<string>
   saveSnapshotFile: (filepath: string, snapshot: string) => Promise<void>
   readSnapshotFile: (filepath: string) => Promise<string | null>
+  readSnapshotFileData?: (filepath: string) => Promise<Record<string, string> | null>
```
hi-ogawa, line 10: On 2nd thought, it may be fine this is optional for integrator but fill up the default implementation on environment side somehow.

lazerg: Went with this. Kept `readSnapshotFileData` optional on the interface and added a default implementation to `NodeSnapshotEnvironment` that reads the file and evaluates it in Node. Custom environments that don't implement it still fall back to `readSnapshotFile` + the in-runtime eval, so nothing breaks for them, while the bundled Node and browser environments never eval in the tester.

lazerg:
Updated for the review:

- kept `readSnapshotFileData` optional, with a default impl on `NodeSnapshotEnvironment` so it's not breaking
- bad snapshot files now throw `Invalid snapshot file, please manually fix or delete it: <path>` instead of being swallowed
- added e2e for the CSP case (browser test, `script-src` without `unsafe-eval`) and for the corrupted-file error

Went non-breaking per your second comment, but can make it required instead if you'd prefer.

lazerg:
@hi-ogawa think this is ready for another look whenever you get a chance 🙏

packages/snapshot/src/types/environment.ts
hi-ogawa: What I had in mind was a bit different: instead of a default impl on `NodeSnapshotEnvironment`, just handle the default inline in `SnapshotState.create`. When `readSnapshotFileData` isn't provided, do `readSnapshotFile` + evaluate right there. That way every environment gets the default for free, the `NodeSnapshotEnvironment` impl isn't needed (it evals in the worker anyway), and the old in-runtime eval branch goes away. Browser stays the only environment that overrides `readSnapshotFileData` (to eval server-side).

On "non-breaking": I only mean the `SnapshotEnvironment` interface — `readSnapshotFileData` stays optional, so existing integrator environments keep working untouched. I'm not worried about the `SnapshotState` constructor signature; that's internal, so changing it (e.g. dropping the `snapshotContent` param) is fine.

hi-ogawa (changes requested):

lazerg:
Closing — the approach needs more rework than I can commit to right now. Thanks for the detailed review @hi-ogawa.

PR #10658, opened 2026-06-24
fix(vm): fix external module resolve error with deps optimizer query for encoded URI

### LINKED ISSUE #10657: mThreads/vmForks pool crashes with ERR_MODULE_NOT_FOUND when deps.optimizer.client.enabled is true
### Describe the bug

This is a variation of #9899 
I'm doing exactly the same as is described there, I also get the same error message. I'm using vitest 4.1.9

### Reproduction

```ts
// vitest.config.ts

export default defineConfig({
  test: {
    environment: 'happy-dom',
    pool: 'vmThreads', // also fails with 'vmForks'
    deps: {
      optimizer: {
        client: {
          enabled: true,
          include: ['react', 'react-dom'],
        },
      },
    },
  },
});
```

```ts
// example.test.ts
test('basic', () => { expect(1 + 1).toBe(2); });
```

Run:
```bash
vitest run
```

The difference to #9899 is that the ? is escape in the URL, and therefore the fix isn't working:
```
node_modules/vitest/dist/spy.js%3Fv=1c38a1aa
```

### System Info

| | |
|---|---|
| vitest | 4.1.9 |
| vite | 8.0.10 |
| node | 24.16.0 |
| npm | 11.13.0 |
| OS | Linux 6.8 Ubuntu 24.04.4 LTS 24.04.4 LTS (Noble Numbat) |
| pool | vmThreads / vmForks (both affected) |
| environment | jsdom (also reproducible with happy-dom) |

```Shell
  System:
    OS: Linux 6.8 Ubuntu 24.04.4 LTS 24.04.4 LTS (Noble Numbat)
    CPU: (22) x64 Intel(R) Core(TM) Ultra 7 155H
    Memory: 8.25 GB / 30.85 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  Binaries:
    Node: 24.16.0 - ~/.nvm/versions/node/v24.16.0/bin/node
    npm: 11.13.0 - ~/.nvm/versions/node/v24.16.0/bin/npm
    pnpm: 10.11.0 - /run/user/731821679/fnm_multishells/2261290_1782132283535/bin/pnpm
  Browsers:
    Chrome: 149.0.7827.114
    Chromium: 149.0.7827.53
  npmPackages:
    @vitejs/plugin-react: ^6.0.1 => 6.0.1 
    @vitest/coverage-v8: ^4.1.9 => 4.1.9 
    @vitest/ui: ^4.1.9 => 4.1.9 
    vitest: ^4.1.9 => 4.1.9 
    vitest-sonar-reporter: 3.0.0 => 3.0.0 
```

### Used Package Manager

npm

### Validations

SveLil:
### Description

Resolves #10657 

### Tests
- [x] Run the tests with `pnpm test:ci`

### Changesets

### DIFF
--- a/packages/vitest/src/runtime/moduleRunner/cachedResolver.ts
+++ b/packages/vitest/src/runtime/moduleRunner/cachedResolver.ts
@@ -1,5 +1,6 @@
 import type { WorkerGlobalState } from '../../types/worker'
 import { pathToFileURL } from 'node:url'
+import { splitFileAndPostfix } from '@vitest/utils/helpers'
 import { join, normalize } from 'pathe'
 import { distDir } from '../../paths'
 
@@ -25,9 +26,10 @@ export function getCachedVitestImport(
   const root = state().config.root
   const relativeRoot = relativeIds[root] ?? (relativeIds[root] = normalizedDistDir.slice(root.length))
   if (id.includes(distDir) || id.includes(normalizedDistDir)) {
+    const { file, postfix } = splitFileAndPostfix(id)
     const externalize = id.startsWith('file://')
       ? id
-      : pathToFileURL(id).toString()
+      : `${pathToFileURL(file)}${postfix}`
     externalizeMap.set(id, externalize)
     return { externalize, type: 'module' }
   }
@@ -36,8 +38,9 @@ export function getCachedVitestImport(
     // /node_modules/.pnpm/vitest/dist
     (relativeRoot && relativeRoot !== '/' && id.startsWith(relativeRoot))
   ) {
-    const path = join(root, id)
-    const externalize = pathToFileURL(path).toString()
+    const { file, postfix } = splitFileAndPostfix(id)
+    const path = join(root, file)
+    const externalize = `${pathToFileURL(path)}${postfix}`
     externalizeMap.set(id, externalize)
     return { externalize, type: 'module' }
   }


SveLil:
Note: This is only a fix for ONE cleanUrl, there are quite a few, maybe time for a cleanup?

sheremet-va:
Why does vm receive an encoded string when cleaning the URL? Is it passed down incorrectly?

hi-ogawa (changes requested):
The only source of glitch seems like the single point here in `getCachedVitestImport` https://github.com/vitest-dev/vitest/blob/a60ded0fb1a15a5bc2eb6c26681d9e934d3e17eb/packages/vitest/src/runtime/moduleRunner/cachedResolver.ts#L30 so fix belongs there and not the general `cleanUrl` side.

---

Unfortunately, we get `?v=xxx` only when `vitest` is inside `node_modules` so that seems like the reason we don't repro in our e2e.

SveLil:
I just debugged, for me it's a few lines later:

https://github.com/vitest-dev/vitest/blob/a60ded0fb1a15a5bc2eb6c26681d9e934d3e17eb/packages/vitest/src/runtime/moduleRunner/cachedResolver.ts#L40 
---
I'll push the fix soon.

packages/vitest/src/runtime/moduleRunner/cachedResolver.ts
```diff
@@ -27,7 +27,7 @@ export function getCachedVitestImport(
   if (id.includes(distDir) || id.includes(normalizedDistDir)) {
     const externalize = id.startsWith('file://')
       ? id
-      : pathToFileURL(id).toString()
+      : decodeURIComponent(pathToFileURL(id).toString())
```
hi-ogawa: From what I can guess, the idea isn't decoding. The idea is to have `pathToFileURL` to not include `?param` by splitting id so it aligns with the fix on consumer side https://github.com/vitest-dev/vitest/pull/10024.

hi-ogawa (changes requested):

packages/vitest/src/runtime/moduleRunner/cachedResolver.ts
hi-ogawa: Perhaps something like this here and below.

```js
const { file, postfix } = splitFileAndPostfix(id)
const externalize = id.startsWith('file://') ? id : `${pathToFileURL(file)}${postfix}`
```

SveLil: No problem at all, that works, too

SveLil: Pushed a version that will include the param but returns a non-encoded URI

hi-ogawa (approved):
Thank you!
I verified the repro with pkg.pr.new. Not being able to e2e is unfortunate but can be revisited in the future.

hi-ogawa:
This means vm + optimizer has been broken entire v4 phase. We may want to backport this to v4.

SveLil:
> This means vm + optimizer has been broken entire v4 phase. We may want to backport this to v4.

Yes, I'm on v4 and I will backport it

sheremet-va (approved):

sheremet-va:
Please, backport to v4 branch

SveLil:
See #10661 10661

PR #10659, opened 2026-06-25
fix: fix per-project `sequence` config

### LINKED ISSUE #9645: Project sequence.concurrent not being applied
### Describe the bug

I am currently moving some of our test files to leverage sequence.concurrent and was hoping to leverage a `project` to do so. It seems like the file only respects the root level configuration (or if the describe block manually sets the concurrent = true). 

According to https://vitest.dev/config/sequence.html#sequence-concurrent , the setting isn't a root-level only like the sequencer. 

### Reproduction

https://github.com/joeyslater/vitest-project-concurrent-repro

### System Info

```shell
OS: macOS 15.7.3
    CPU: (14) arm64 Apple M4 Pro
    Memory: 95.13 MB / 48.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.13.0 - /Users/jslater/.nvm/versions/node/v24.13.0/bin/node
    Yarn: 4.12.0 - /Users/jslater/.nvm/versions/node/v24.13.0/bin/yarn
    npm: 11.6.2 - /Users/jslater/.nvm/versions/node/v24.13.0/bin/npm
    Watchman: 2025.11.10.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 141.0.7390.123
    Safari: 26.2
  npmPackages:
    vitest: 4.0.18 => 4.0.18
```

### Used Package Manager

yarn

### Validations

hi-ogawa:
### Description

- Closes https://github.com/vitest-dev/vitest/issues/9645

The current intent of per-project `sequence` config appears to be `Omit<SequenceOptions, 'sequencer' | 'seed'>` i.e. `sequence.sequencer` and `sequence.seed` are root only. This intent is broken both in type `ProjectConfig` and runtime. This PR fixes both.

As one exception not covered by doc, `sequence.shuffle.files` is a root config concept, so it's clarified in doc and excluded from `ProjectConfig` type.

### Tests

### Documentation

### Changesets


packages/vitest/src/node/config/resolveConfig.ts
```diff
@@ -790,10 +790,7 @@ export function resolveConfig(
   }
   resolved.sequence.groupOrder ??= 0
   resolved.sequence.hooks ??= 'stack'
-  // Set seed if either files or tests are shuffled
-  if (resolved.sequence.sequencer === RandomSequencer || resolved.sequence.shuffle) {
-    resolved.sequence.seed ??= Date.now()
-  }
+  resolved.sequence.seed ??= Date.now()
```
hi-ogawa, line 793: I simplified this and `getSeed` since I wasn't sure if leaving seed undefined is intended from the context of:
- https://github.com/vitest-dev/vitest/pull/8592

If we want to keep `getSeed` returning null, we can keep revert this and seed fixup seed can happen later during `resolveProjects`, so let me know.

hi-ogawa: Actually we can keep the semantics of `getSeed` while ensuring seed number at resolved config level. See changes in `getSeed`.

sheremet-va (approved):

PR #10662, opened 2026-06-25
fix(browser): resize the browser ui only if it's not headless

sheremet-va:
I found that this test always fails locally on Mac with Webkit reporting a different resolution:

https://github.com/vitest-dev/vitest/blob/a60ded0fb1a15a5bc2eb6c26681d9e934d3e17eb/test/browser/specs/to-match-screenshot.test.ts#L284

### DIFF
--- a/packages/browser-playwright/src/playwright.ts
+++ b/packages/browser-playwright/src/playwright.ts
@@ -434,7 +434,11 @@ export class PlaywrightBrowserProvider implements BrowserProvider {
       ...contextOptions,
       ignoreHTTPSErrors: true,
     } satisfies BrowserContextOptions
-    if (this.project.config.browser.ui) {
+    // A `null` viewport lets the page adopt the real window size, which is only
+    // meaningful for a headed UI. In headless mode there is no real window, so it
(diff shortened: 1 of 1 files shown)


macarie (approved):
I guess it makes sense, even tho I don't think anyone would run with UI enabled but headless, or at least I hope 😅

_It works on my machine™_ (and CI?) so I wonder why the test fails tho. What are the sizes of the two screenshots?

Also, do you have this option (Appearance → Windows, at the bottom) set to "Always"?

Screenshot of macOS' Show Scroll Bars setting options

sheremet-va:
> I guess it makes sense, even tho I don't think anyone would run with UI enabled but headless, or at least I hope 😅
> 
> _It works on my machine™_ (and CI?) so I wonder why the test fails tho. What are the sizes of the two screenshots?
> 
> Also, do you have this option (Appearance → Windows, at the bottom) set to "Always"?
> 
> Screenshot of macOS&apos; Show Scroll Bars setting options

The size is just doubled, the device pixel ratio on retina kicks in

PR #10665, opened 2026-06-26
fix(snapshot): support no-unsafe-eval CSP by evaluating snapshot files on server

### LINKED ISSUE #10615: `getSnapshotData` swallows the parse error, making snapshot failures impossible to debug

hi-ogawa:
### Description

- closes https://github.com/vitest-dev/vitest/issues/10615
- supersedes https://github.com/vitest-dev/vitest/pull/10646

Currently the snapshot utils always evaluates snapshot file key values with `new Function` in the test runtime, but this would fails on browser mode with stricter CSP.

This PR adds `SnapshotEnvironment.readSnapshotFileData`, so browser environment can evaluate the file on the server so that browser runtime can directly receive napshot key value data. 

Also as a drive-by change, this PR now surfaces snapshot files evaluation errors as an actionable errors instead of silently ignoring it.

TODO
- [x] test

### Tests

### Documentation

### Changesets


packages/browser/package.json
```diff
@@ -78,6 +78,7 @@
   "dependencies": {
     "@blazediff/core": "1.9.3",
     "@vitest/mocker": "workspace:*",
+    "@vitest/snapshot": "workspace:*",
```
hi-ogawa: todo: this is just for `evaluateSnapshotFile`, so should be bundled.

packages/snapshot/src/environment.ts
```diff
@@ -1,2 +1,3 @@
 export { NodeSnapshotEnvironment } from './env/node'
+export { evaluateSnapshotFile } from './port/utils'
 export type { SnapshotEnvironment } from './types/environment'
```
hi-ogawa, line 3: todo: probably shouldn't export from next to node thing since this is imported in browser.

EDIT: Actually it's fine because `evaluateSnapshotFile` is meant to run on server.

packages/browser/package.json
```diff
@@ -92,6 +92,7 @@
     "@types/node": "catalog:",
     "@types/pngjs": "^6.0.5",
     "@types/ws": "catalog:",
+    "@vitest/snapshot": "workspace:*",
```
hi-ogawa, line 95: This adds a dev dep to bundle a small piece of code `evaluateSnapshotFile` about 20 loc. Technically this duplicates main `vitest` package's snapshot bundled dep, but should be fine since small enough.

test/test-utils/index.ts
```diff
@@ -319,8 +319,9 @@ export async function runVitest(
         ? buildErrorProjectTree(modules, options)
         : buildErrorTree(modules, options)
       const errors = ctx?.state.getUnhandledErrors()
-      if (errors && errors.length > 0) {
-        tree.__unhandled_errors__ = errors.map((e: any) => e.message)
+      const root = ctx?.config.root
+      if (errors && errors.length > 0 && root) {
+        tree.__unhandled_errors__ = errors.map((e: any) => replaceRoot(e.message, root))
```
hi-ogawa, line 324: Tweaked error tree to normalize root:

```js
"__unhandled_errors__": [
  "Invalid snapshot file, please manually fix or delete it: <root>/__snapshots__/basic.test.ts.snap",
]
```

packages/browser/src/node/rpc.ts
```diff
@@ -276,6 +277,14 @@ export function setupBrowserRpc(globalServer: ParentBrowserProject, defaultMocke
           }
           return fs.readFile(snapshotPath, 'utf-8')
         },
+        async readSnapshotFileData(snapshotPath) {
+          checkFileAccess(snapshotPath)
```
sheremet-va, line 283: I think this also needs `api.allowExec` check

sheremet-va (changes requested):

packages/browser/src/node/rpc.ts
hi-ogawa: Good point. Updated 👌

sheremet-va (approved):

PR #10675, opened 2026-06-29
refactor!: remove webdriverio package

sheremet-va:
webdriverio provider now lives in https://github.com/vitest-community/vitest-webdriverio


AriPerkkio (approved):
🙏

packages/vitest/package.json
```diff
@@ -120,7 +120,7 @@
     "@types/node": "^22.0.0 || >=24.0.0",
     "@vitest/browser-playwright": "workspace:*",
     "@vitest/browser-preview": "workspace:*",
-    "@vitest/browser-webdriverio": "workspace:*",
+    "@vitest/browser-webdriverio": "^5.0.0-beta.5 || >=5.0.0",
```
hi-ogawa, line 123: I forgot why browser is peer dep. Don't we want to eventually remove webdriverio one at least otherwise version bump seems messy.

pnpm-workspace.yaml
```diff
@@ -8,14 +8,14 @@ trustPolicy: no-downgrade
...
+  - '@wdio/types@9.29.1'
   - '@wdio/repl@9.16.2'
-  - '@wdio/utils@9.20.0'
-  - '@wdio/config@9.20.0'
-  - '@wdio/logger@9.18.0'
-  - '@wdio/protocols@9.16.2'
+  - '@wdio/utils@9.29.1'
+  - '@wdio/config@9.29.1'
+  - '@wdio/logger@9.29.1'
+  - '@wdio/protocols@9.29.1'
```
hi-ogawa, line 18: Are these still needed?

hi-ogawa (commented):
Also as we discussed, we still have `webdriverio` branch in core but removing/reworking these are out of scope. Just surfacing these for posterity:

- `packages/browser/src/node/rpc.ts:323` — `switchContext` RPC gated to `provider.name !== 'webdriverio'`.
- `packages/vitest/src/node/pools/browser.ts:459` — `shouldIgnoreDebugger` special-cases `webdriverio` (chrome/edge only).
- `packages/browser/src/client/tester/tester.ts:185` — a non-trivial iframe context-switch block that runs only when `server.provider === 'webdriverio'`.
- `packages/browser/src/client/tester/tester-utils.ts:90` — shadow-DOM `>>>` selector prefix only for webdriverio.
- `packages/browser/src/client/tester/context.ts:50` — keyboard modifier maps webdriverio → `Ctrl`.
- `packages/browser/src/client/tester/logger.ts:100` — filters `[WDIO]` shadow-root logs.

packages/vitest/package.json
sheremet-va: vitest/browser exports from the webdriverio. without it, the types are not populated. I am not sure if there is a better way

hi-ogawa: Oh that one. Then it's fine for another day in the future.

hi-ogawa: We don't actually add or extend any type through this though

https://github.com/vitest-dev/vitest/blob/d20aa66b6949b826928cbe7708a1a38f0fbb4d1c/packages/browser-playwright/context.d.ts#L1

And actual semi-automatic type extension is through main provider import such as 

https://github.com/vitest-dev/vitest/blob/d20aa66b6949b826928cbe7708a1a38f0fbb4d1c/packages/browser-playwright/src/playwright.ts#L650-L651

It looks like `vitest` should have just `@vitest/browser` as optional peer dep instead of all three. Not sure what I may be missing.

pnpm-workspace.yaml
hi-ogawa: Still needed for https://github.com/vitest-dev/vitest/pull/10675#discussion_r3495372940

hi-ogawa (approved):
lgtm

hi-ogawa:
Btw, in what sense is this breaking? (meaning anything to mention as v5 breaking?)

sheremet-va:
> Btw, in what sense is this breaking? (meaning anything to mention as v5 breaking?)

Hm, it is not breaking, right 🤔

packages/vitest/package.json
sheremet-va: If we do that, then extra types from providers won't be picked up (like options)

PR #10676, opened 2026-06-29
test: remove axios

sheremet-va:
We don't actually use this as a package anywhere, it's only used by mocks, and we can just make our own fixture instead


test/unit/vite.config.ts
```diff
@@ -77,6 +77,7 @@ export default defineConfig({
     alias: [
       { find: /^#/, replacement: resolve(import.meta.dirname, 'src') },
       { find: /^custom-lib$/, replacement: resolve(import.meta.dirname, 'projects', 'custom-lib') },
+      { find: /^http-client$/, replacement: resolve(import.meta.dirname, 'projects', 'http-client') },
```
hi-ogawa: Can we use `"http-client": "file:..."` in package.json? The resolver does different with and without `node_moduels` and using `file:...` would preserve axios mocking closer.

sheremet-va: We could, but it’s not important to the mocker, it looks at moduleDirectories for externalisation

sheremet-va: Done

test/unit/package.json
```diff
@@ -32,7 +32,7 @@
     "@vitest/utils": "workspace:*",
     "@vitest/web-worker": "workspace:^",
     "@vueuse/integrations": "^14.2.1",
-    "axios": "^1.13.4",
+    "http-client": "link:./projects/http-client",
```
hi-ogawa: `link:...` doesn't make it to `node_modules` with symlink resolved.

sheremet-va: Sure

hi-ogawa (approved):

PR #10677, opened 2026-06-29
fix(ui): remove stale task nodes when a file is re-collected

### LINKED ISSUE #10670: Bug: Removing/commenting-out a test does not update the UI sidebar tree
### Describe the bug

In the latest Vitest (v4.1.9) when I delete or comment-out a failing test, it sometimes remains listed in red in the UI sidebar, despite all tests passing in the console. I would expect the entry to disappear.

### Reproduction

Here is a Stackblitz: https://stackblitz.com/edit/vitest-dev-vitest-ytmcjdwe?file=test%2Fbasic.test.ts

1. On initial load all tests pass
2. Uncomment the "FAILING TEST" line at the bottom on `basic.test.ts`, the test will fail
3. Comment out the line again, tests pass in the console, but the "FAILING TEST" entry remains in the UI sidebar

Clicking "Re-run all tests" does *not* fix the issue, it still persists.

Hitting "Refresh" in the browser UI seems to update the sidebar correctly. This is a temporary workaround.

<details><summary><h3>System Info</h3></summary>

```shell
System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 22.22.3 - /usr/local/bin/node
  npmPackages:
    @vitest/ui: latest => 4.1.9 
    vite: latest => 8.1.0 
    vitest: latest => 4.1.9
```

</details>

### Used Package Manager

npm

### Validations

eldardada:
### Description

When a test is removed or commented out, the file is re-collected with fewer tasks. The sidebar tree only ever added or updated nodes while collecting; it never dropped the ones that were gone, so the orphaned entry stayed in the explorer (often still showing red) until a full page reload. Re-running all tests does not help, because the rerun goes through the same collect path. This matches the report in #10670.

`createOrUpdateFileNode` (when `collect` is set) and `createOrUpdateNode` (when `createAll` is set) rebuild a file or suite subtree from the collected tasks, but they only walked the incoming list, so any tracked child missing from it was left behind. This change reconciles the tracked children against the collected tasks after each subtree rebuild and removes any node, along with its descendants, that no longer exists.

The summary counts were already correct because they come from the live task tree rather than the UI nodes, which is why the console showed the right totals while the sidebar did not.

Resolves #10670

### Tests

Added `packages/ui/client/composables/explorer/tree.spec.ts`. It collects a file containing a suite, re-collects it after a test is deleted, and checks that the orphaned node is dropped, that a removed suite takes its children with it, and that untouched and newly added tasks stay. The two removal cases fail on `main` and pass with this change.

I went with a unit test rather than a browser e2e because the report is intermittent ("sometimes remains"), so an e2e around the watch rerun would be timing-dependent, while the reconciliation itself is deterministic.

### Documentation
- [ ] No user-facing API changes, so no documentation update is needed.

### Changesets

### DIFF
--- a/packages/ui/client/composables/explorer/tree.spec.ts
+++ b/packages/ui/client/composables/explorer/tree.spec.ts
@@ -1,0 +1,162 @@
+import type { RunnerTask, RunnerTestFile } from 'vitest'
+import type { FileTreeNode, SuiteTreeNode } from '~/composables/explorer/types'
+import { beforeEach, describe, expect, it } from 'vitest'
+import { explorerTree } from '~/composables/explorer'
+import { createOrUpdateFileNode } from '~/composables/explorer/utils'
+
+interface TestSpec {
+  id: string
+  name: string
+  state?: 'pass' | 'fail'
+}
+
+interface SuiteSpec {
+  id: string
+  name: string
+  children: TestSpec[]
+}
+
+function buildTest(spec: TestSpec, file: RunnerTestFile): RunnerTask {
+  // a test task must not own a `tasks` property, otherwise it is treated as a suite
+  return {
+    id: spec.id,
+    name: spec.name,
+    type: 'test',
+    mode: 'run',
+    file,
+    result: spec.state ? { state: spec.state } : undefined,
+  } as unknown as RunnerTask
+}
+
+function buildSuite(spec: SuiteSpec, file: RunnerTestFile): RunnerTask {
+  return {
+    id: spec.id,
+    name: spec.name,
+    type: 'suite',
+    mode: 'run',
+    file,
+    tasks: spec.children.map(child => buildTest(child, file)),
+  } as unknown as RunnerTask
+}
+
+function buildFile(tasks: Array<TestSpec | SuiteSpec>): RunnerTestFile {
+  const file = {
+    id: 'file-1',
+    name: 'sample.test.ts',
+    type: 'suite',
+    mode: 'run',
+    meta: {},
+    filepath: '/sample.test.ts',
+    projectName: '',
+    result: { state: 'fail' },
+    tasks: [],
+  } as unknown as RunnerTestFile
+
+  file.tasks = tasks.map(task =>
+    'children' in task ? buildSuite(task, file) : buildTest(task, file),
+  )
+
+  return file
+}
+
+function fileNode() {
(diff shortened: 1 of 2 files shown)


eldardada:
The two red jobs look unrelated to this change.

`Browsers: windows-latest` failed in `hooks-timeout.test.ts` with sub-400ms `locator.click` timeouts across chromium, firefox and webkit, while the same browser suite passed on `macos-latest`. `Test: vite@7, macos` failed in `test/watch/reporter-failed.test.ts` (timeout waiting for the `RERUN` line) and `test/oxc.test.ts` (an "Unknown session id" error).

This PR only touches the UI explorer tree under `packages/ui/client`, which is loaded only with `--ui`, so none of those tests run the changed code. Looks like timing/infra flakiness. Happy to rebase if a re-run would help.

SimonEast:
Thanks for working on a fix @eldardada. I don't think I was able to tag you in the original issue (https://github.com/vitest-dev/vitest/issues/10670#issuecomment-4864520743), so I'll repost my comment again here:

I also noticed today that *deleting* a test file triggers the same issue, except it's even worse – it doesn't get removed after a browser refresh. You have to fully restart Vitest in order to get it to pickup the deleted test file. 

Same issue occurs on test file rename, except that you end up with two copies showing in the left pane (one is stale).

Does your fix handle those two scenarios too?

eldardada:
Good catch. Those are a different path from what this PR touches, so I opened a separate one for them: #10702. This PR only reconciles tasks inside a file that gets re-collected, like a test you comment out. Deleting or renaming the whole file never reached the UI at all, because the file node lives at the tree root and the watcher's removal event wasn't forwarded over the websocket. #10702 forwards that event and drops the stale file node, and since a rename is really a delete of the old path plus a collect of the new one, it clears the duplicate too

SimonEast:
Thank you! Really appreciate your work! 😊👏

packages/ui/client/composables/explorer/tree.spec.ts
hi-ogawa, line 1: Can you remove unit test and add e2e in test/ui instead?

hi-ogawa (changes requested):

PR #10681, opened 2026-06-30
fix: don't print column in test names when `includeTaskLocation` is enabled

sheremet-va:
Vitest doesn't support `file:line:column` format when running tests, so one always has to manually remove the column when copying the path. Instead, let's just only print the line. Column is usually irrelevant anyway

### DIFF
--- a/docs/guide/reporters.md
+++ b/docs/guide/reporters.md
@@ -248,10 +248,10 @@ Example output:
 An example with `--includeTaskLocation`:
 
 ```bash
-✓ __tests__/file1.test.ts:2:1 > first test file > 2 + 2 should equal 4 1ms
-✓ __tests__/file1.test.ts:3:1 > first test file > 4 - 2 should equal 2 1ms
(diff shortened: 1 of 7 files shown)


hi-ogawa (commented):
Cannot tell where this change shows up exactly. This isn't the error stack line:column right? Is this only for display side of `includeTaskLocation`?

sheremet-va:
> Cannot tell where this change shows up exactly. This isn't the error stack line:column right? Is this only for display side of `includeTaskLocation`?

Yes. We print the location in verbose reporter if `includeTaskLocation` is enabled, and also in the "Errors" section when we print the name of the test

hi-ogawa (approved):
This looks fine and doesn't change much anyway since the column is just `test/it` indentation.

hi-ogawa:
Can you update PR title to mention `includeTaskLocation` so it won't mislead as general errors display?

PR #10683, opened 2026-06-30
ci: speed up the ci by parallelising and deduplicating tests

sheremet-va:
- Removing `--with-deps` should speed up our Windows tests since the deps are not used by us
- Merging cached with normal tests should remove duplicate jobs
- enabling parallelism should speed up e2e suite (spawned tests still use `maxWorkers: 1`)
- splitting vite7 jobs into two should make the time it takes to runs tests more equal across jobs
- only running pnpm trust check once during lint
- skipped installing docs dependencies that pulled in `sharp` (~12s)


hi-ogawa (approved):
Nice!

AriPerkkio (approved):
> Removing `--with-deps` should speed up our Windows tests since the deps are not used by us

If I remember correctly, on Linux these are also only needed when running `firefox` or `webkit`. So if we don't run any of those browsers, or move them to separate CI job, there's some more optimization opportunities left. 

Also Github Actions might have [some Chrome installed already](https://github.com/web-infra-dev/rstest/commit/16b2e8ef26353460a7681b4e151c0b232c47c430#diff-82a43d8a19c7cf8dbbfa74d160ef5dfc17a923d46b728f77bc2cb0a014815d4c), so whole `playwright install` could be skipped.

sheremet-va:
> > Removing `--with-deps` should speed up our Windows tests since the deps are not used by us
> 
> If I remember correctly, on Linux these are also only needed when running `firefox` or `webkit`. So if we don't run any of those browsers, or move them to separate CI job, there's some more optimization opportunities left.
> 
> Also Github Actions might have [some Chrome installed already](https://github.com/web-infra-dev/rstest/commit/16b2e8ef26353460a7681b4e151c0b232c47c430#diff-82a43d8a19c7cf8dbbfa74d160ef5dfc17a923d46b728f77bc2cb0a014815d4c), so whole `playwright install` could be skipped.

We are running all the browsers, so there is no opportunity to improve here left anymore. And linux is fast anyway 😄

sheremet-va:
This PR improves the overall times by just a few minutes because the windows job is pretty slow. We need structural changes to lower the times.

PR #10684, opened 2026-06-30
feat: add --smart-rerun to prioritize or skip previously failed tests

### LINKED ISSUE #7834: Run failed tests in a clean environment on retry

Thompsonay:
## Summary
- Adds `--smart-rerun`: caches failed test file paths to `.vitest-failed-cache.json` and runs them first on the next invocation, ahead of the rest of the suite. Cache clears automatically once everything passes.
- Adds `--smart-rerun-only-failed` (requires `--smart-rerun`): instead of reordering, skips files that passed last run entirely and only executes previously-failed files. Falls back to a full run if the cache is empty or stale.
- Both flags are off by default and do not alter existing behavior.

## Test plan
- [x] Unit tests for cache save/load/clear, corrupted-cache handling, reorder, and filter logic (`test/unit/test/smart-rerun.test.ts`)
- [x] CLI parsing tests for both flags (`test/unit/test/cli-test.test.ts`)
- [x] `pnpm run build` passes with no type errors
- [x] Full `test/unit` suite passes
- [x] Verified pre-existing/environment-specific failures elsewhere in the monorepo are unaffected by this change (reproduced identically on `main` via `git stash`)

Closes #7834
Related to #10247

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ test/e2e/fixtures/browser-multiple/basic-*
 .vitest
 explainFiles.txt
 .vitest-dump
+.vitest-failed-cache.json
--- a/docs/config/smartrerun.md
+++ b/docs/config/smartrerun.md
@@ -1,0 +1,14 @@
+---
+title: smartRerun | Config
+outline: deep
+---
+
+# smartRerun
+
+- **Type:** `boolean`
+- **CLI:** `--smartRerun`, `--smart-rerun`
+- **Default:** `false`
+
+Cache the file paths of failed tests in `.vitest-failed-cache.json` in the project root. On the next run, previously failed tests are moved to the front of the test queue so they run first, ahead of the rest of the suite. The cache file is automatically removed once all tests pass.
+
+See [`smartRerunOnlyFailed`](/config/smartrerunonlyfailed) to skip files that passed in the previous run instead of just reordering them.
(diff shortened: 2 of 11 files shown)


Thompsonay:
I contributed to the open source

PR #10685, opened 2026-06-30
perf: bundle vitest's own dependencies

sheremet-va:
Bundles vitest's own dependencies into its dist to reduce the boot time by cutting the external module resolutions Node performs when loading the worker bundle:

metric | main | branch | Δ
-- | -- | -- | --
worker boot import (cold, min) | 28.8 ms | 21.3 ms | −26%
wall — forks isolate:true (min) | 1029 ms | 937 ms | −9%
wall — forks isolate:false | 382 ms | 384 ms | ~0 (amortized)
worker boot modules loaded | 61 | 45 | −26%
— node: builtins | 14 | 10 | −4
— vitest dist chunks | 23 | 28 | +5
— external npm resolves | 7 | 2 | −71%
boot graph (static closure) | 605 KB / 23 chunks | 759 KB / 28 chunks | +154 KB / +5
dist size | 2215 KB | 2557 KB | +342 KB

Details:

- `tinyrainbow` is pinned to a single shared chunk: its default export is a mutable color object whose formatters are toggled at runtime, so inlining it into several chunks would give each chunk its own copy and break runtime color toggling. It is re-exported as `disableDefaultColors` from `vitest/node` so programmatic consumers (e.g. our test suite) can disable colors on the instance Vitest actually renders with.
- `NativeModuleRunner` is imported lazily so `local-pkg` stays out of the default worker startup graph; only the `viteModuleRunner: false` path needs it.

> [!NOTE]
> This PR was previously stacked and also carried an isolate-hint diagnostic and the browser optimizeDeps pre-bundling. The stack was split into independent feature PRs: the isolate hint now ships as part of `experimental.diagnostics` in #10710, and the browser pre-bundling moved to #10713.

TODO
- [ ] Move deps to devDeps


packages/vitest/src/node/reporters/base.ts
```diff
@@ -660,7 +661,77 @@ export abstract class BaseReporter implements Reporter {
...
+      return
+    }
+
+    const state = this.ctx.state
+    const numWorkers = state.workersSpawned
+    // only meaningful when at least one non-browser project isolates workers
+    const isolates = this.ctx.projects.some(
+      project => project.config.isolate && !project.config.browser.enabled,
+    )
+    if (!numWorkers || !isolates) {
```
AriPerkkio: Nit: `const isolates` naming doesn't feel right. It's ok to be verbose:

```suggestion
    const hasAnyIsolatedNodeProjects = this.ctx.projects.some(
      project => project.config.isolate && !project.config.browser.enabled,
    )
    if (!numWorkers || !hasAnyIsolatedNodeProjects) {
```

packages/vitest/src/node/core.ts
```diff
@@ -1531,21 +1532,29 @@ export class Vitest {
...
+          try {
+            await this.pool.close?.()
+          }
+          catch (error) {
+            teardownErrors.push(error)
+          }
+
+          this.pool = undefined
+        }
+
```
AriPerkkio, line 1543: This sounds like unrelated bug fix. I'd like to have this in separate PR shipped as bug fix if it makes sense 👀

packages/vitest/src/node/core.ts
```diff
@@ -1566,7 +1575,8 @@ export class Vitest {
    * @param force If true, the process will exit immediately after closing the projects.
    */
   public async exit(force = false): Promise<void> {
-    setTimeout(() => {
+    clearTimeout(this._exitTimeout)
+    this._exitTimeout = setTimeout(() => {
```
AriPerkkio, line 1579: What's this about? Can `exit` be called multiple times? Should we `return` instead when it's already exitting?

packages/vitest/src/node/reporters/base.ts
sheremet-va: this part of the PR is rewritten here - https://github.com/vitest-dev/vitest/pull/10710 (all perf PRs are stacked on top of each other)

packages/vitest/src/node/core.ts
sheremet-va: It's called by tests where we do `Ctrl+C`. There was no way to clean that timeout and it killed subsequent tests after 10s, `this._exitTimeout` exists just so we can clear it

packages/vitest/rollup.config.js
```diff
@@ -52,9 +52,29 @@ const dtsEntries = {
...
 
+// Dependencies inlined into the chunks that use them (instead of staying external
+// npm modules) so a spawned worker loads fewer modules at boot — per-module ESM
+// overhead (resolve + read + compile) dominates over bytes. Kept external:
+//   - chai: users `import 'chai'` to register plugins → must be one shared instance
+//   - @vitest/mocker: mock registry identity
+//   - vite/module-runner, #module-evaluator, @opentelemetry: runtime-provided
+// These are stateless or only consumed through `vitest`, so inlining them is
+// identity-safe (rollup still emits ONE shared chunk for each — no duplication).
+// TODO: remove them from dependencies, this is a prototype
```
sheremet-va: Note for reviewers: I will do this after review

AriPerkkio (dismissed):

packages/utils/display.d.ts
```diff
@@ -0,0 +1 @@
+export * from './dist/display.js'
```
sheremet-va, line 1: These are required for the dts-bundle plugin, it doesn't seem to support export conditions 👀

PR #10686, opened 2026-06-30
fix!: use `>` as separator in `-t`, calculate `only` once

sheremet-va:
- A small performance improvement to avoid walking over tasks every time we need to check for `only` (it's not calculated during collection)
- `-t` now expects the test name to use vitest's full test name (with `>` as a separator instead of an empty space)


hi-ogawa (commented):
This change may not be as breaking as I initially thought, but worth thinking about any downstream / end users have relied on this.

May need to test out but iirc agents filters test title only `-t ...long test title...` without concatenating with suite title, so probably it's fine there.

sheremet-va:
> This change may not be as breaking as I initially thought, but worth thinking about any downstream / end users have relied on this.
> 
> May need to test out but iirc agents filters test title only `-t ...long test title...` without concatenating with suite title, so probably it's fine there.

Yeah, I was planning to notify the ecosystem channel. But it's a good point about agents, it might be too annoying unless we provide a skill, perhaps? Not sure

AriPerkkio (approved):

PR #10690, opened 2026-07-01
ci: separate long running CI tests

sheremet-va:
Testing how long it takes to run when CI is separated into more jobs

- Split the monolithic test job into parallel rest / e2e / coverage suite jobs (test:ci:rest|e2e|coverage)
- Parallelized the rest suite (dropped --sequential; kept the 2 browser packages serial)
- Enabled --shard for the browser suite (extended BaseSequencer); isolated the heavy runner.test.ts into its own job + sharded the rest
- macOS reduced from 5+ jobs to a single e2e smock test
- Browser suite moved to Ubuntu + Windows
- vite7 moved to Ubuntu, slimmed to an e2e + runner.test.ts smoke; fixed its git-commit identity
- Fixed webkit-on-Linux system libs (playwright install-deps), scoped to browser jobs only
- Node jobs install chromium-only (per-browser-set playwright cache key)
- Reduced node 22 from the matrix to get total jobs under the ~20 concurrent cap (still run a single e2e test as a smock test)
- Defaulted the meta-test harness spawn pool to threads; pinned pool: 'forks' on the few fork-dependent tests
zizmor-clean (justified inline ignores for fixed matrix expansions)


.github/actions/setup-playwright/action.yml
```diff
@@ -1,6 +1,12 @@
 name: Setup Playwright
 description: Setup and cache Playwright browser binaries
 
+inputs:
+  browsers:
+    required: false
+    default: chromium
+    description: 'Playwright browsers to install: "chromium" (headless shell only) or "all"'
+
```
AriPerkkio, line 9:
```suggestion
inputs:
  browsers:
    description: 'Playwright browsers to install: "chromium" (headless shell only) or "all"'
    required: true
    type: choice
    default: chromium
    options:
      - chromium
      - all

```

test/test-utils/index.ts
```diff
@@ -184,6 +184,23 @@ export async function runVitest(
...
+        // don't execute in a node pool.
+        {
+          name: 'vitest:test-utils:default-pool',
+          config(config) {
+            if (config.test?.pool == null && !config.test?.browser?.enabled) {
+              return { test: { pool: 'threads' } }
+            }
+          },
+        },
+      ],
```
AriPerkkio, line 203: Last time I wanted to make this change, you had some strong opinions against it. Did something change? Can't find the discussion anymore. 👀

AriPerkkio (dismissed):

test/test-utils/index.ts
sheremet-va: I just saw the coverage already had it, so why not

hi-ogawa (commented):
Looks good. Hard to follow the bullet points and I just needed a quick table 😄 

| platform · node | unit | e2e | coverage | browser |
|---|:---:|:---:|:---:|---|
| **ubuntu 22** | | ✅ | | |
| **ubuntu 24** | ✅ | ✅ | ✅ | ✅ 2 shards (whole suite) |
| **ubuntu 26** | ✅ | ✅ | ✅ | |
| **windows 24** | ✅ | ✅ | ✅ | ✅ runner isolated + 2 shards (excl. runner) |
| **macos 24** | | ✅ | | |
| **ubuntu 24 (vite7)** | | ✅ | | 🟡 runner only |

I think dropping macos suite is fine as I don't think we saw the platform specific failure mode in the exact spot. 

For Browser suite, I'm thinking we should try to split the big `runner.test.ts` so they should get better split, but that can be a follow up.

One thing we need to check is "merge reports" is failing for some https://github.com/vitest-dev/vitest/actions/runs/28518329229/job/84536811214?pr=10690#step:7:95

.github/workflows/ci.yml
```diff
@@ -125,50 +143,75 @@ jobs:
...
           retention-days: 1
           include-hidden-files: true
 
+      - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        if: ${{ matrix.suite == 'e2e' && !cancelled() }}
+        with:
+          name: vitest-results-e2e-${{ matrix.os }}-node-${{ matrix.node_version }}
+          path: test/e2e/.vitest
+          retention-days: 1
+          include-hidden-files: true
```
hi-ogawa: Is this split needed? probably this should work.

```js
name: vitest-results-${{ matrix.suite }}-${{ matrix.os }}-node-${{ matrix.node_version }}
path: |
  README.md
  test/unit/.vitest
  test/e2e/.vitest
  packages/ui/.vitest
```

`path: test/e2e/.vitest` is the reason why missing e2e blob https://github.com/vitest-dev/vitest/actions/runs/28518329229/job/84536811214?pr=10690#step:7:95

hi-ogawa (changes requested):

package.json
```diff
@@ -26,6 +26,9 @@
     "test": "pnpm --filter test-unit test:threads",
     "test:ci": "CI=true pnpm -r --reporter-hide-prefix --stream --sequential --filter '@vitest/test-*' --filter !test-browser run test",
     "test:ci:no-bail": "CI=true pnpm -r --no-bail --reporter-hide-prefix --stream --sequential --filter '@vitest/test-*' --filter !test-browser run test",
+    "test:ci:e2e": "CI=true pnpm --filter @vitest/test-e2e run test",
+    "test:ci:coverage": "CI=true pnpm --filter @vitest/test-coverage run test",
+    "test:ci:rest": "CI=true pnpm -r --reporter-hide-prefix --stream --filter '@vitest/test-*' --filter !test-browser --filter !test-e2e --filter !test-coverage --filter !test-ui --filter !test-workspaces-browser run test && CI=true pnpm -r --reporter-hide-prefix --stream --sequential --filter test-ui --filter test-workspaces-browser run test",
```
hi-ogawa: What's in the rest doesn't look clear. Is it too much to use positive filter instead of negating many?

Well, of course this is the "rest", but since we want add new suite anytime soon, just listing remaining 4 seems easier.

sheremet-va: i feel like it’s dangerous because when we add a new one, we might forget to add it here. the rest is just not coverage and e2e, all dts tests, unit and such. we can just name it unit if it’s easier. ideally we should move most of the rest to e2e anyway

.github/actions/setup-playwright/action.yml
sheremet-va: this syntax is only valid for `workflow_dispatch` / `workflow_call`

sheremet-va:
Updated `rest` to `unit`, and merge `merge reports` into a single one instead of doing `if:`

hi-ogawa (approved):

Issue #8403, opened 2025-08-07
Include an ability to write the actual snapshot to disk (with a debug flag) when there is a snapshot mismatch


michaelwiles:
### Clear and concise description of the problem

So when a snapshot is mismatched the only place where the comparison is presented is in the test output.

And then a rerun is required in order to update the snapshot.

And when there are  lot of snapshot tests that have mismatches this can be quite tedious.

### Suggested solution

Please include an ability that when there is a mismatch to write the actual snapshot to disk as a file with some suitable debug flag included. 

That way devs can compare the actual with expected and conclude whether the difference is significant and then easily update the expected with changes from the new actual.

This [tool](https://github.com/origin-energy/java-snapshot-testing) from java context has this facility and it is very convenient and I miss it with vitest.

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vitest.dev/guide/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

hi-ogawa:
Can you elaborate what gets written in what file more concretely? Would it be a single file listing all mismatches (just like dumping errors showing up in terminal into a dedicated file) or something else like individual actual/expected pairs get written in a full form each file? Couldn't quickly tell what https://github.com/origin-energy/java-snapshot-testing does exaclty.

hi-ogawa:
Linked this as sub issue of
- https://github.com/vitest-dev/vitest/issues/2229

PR #10691, opened 2026-07-01
feat: custom IO operations for `toMatchScreenshot` screenshots

macarie:
### Description

I'd been planning to add this capability for a while, so I'm taking the opportunity now that it's been requested in https://github.com/vitest-dev/vitest/discussions/9664#discussioncomment-15828832.

This feature would allow storing images either on the local filesystem or in remote stores, which is useful for people who don't store their references in the repository.

When and how to write, as well as permissioning, are entirely left up to the user to decide and implement, since it's a custom requirement we know nothing about. Users will have access to the entire `project` data to help them make the correct assumptions.

There are a couple of open questions:

- How do we handle this for artifacts if there are no files on the local fs? Currently the artifacts produced by the assertion expect a path. We could use the raw data instead, but it would require sending more data to the client. Otherwise we could check on the server if the path exists and ignore it.
- How would this work with #8403 (once this assertion's custom snapshot mechanism is merged with the _actual_ snapshot manager)?

### Tests

### Documentation

### Changesets

### DIFF
--- a/packages/browser/src/client/tester/expect/toMatchScreenshot.ts
+++ b/packages/browser/src/client/tester/expect/toMatchScreenshot.ts
@@ -91,12 +91,13 @@ export default async function toMatchScreenshot(
     }
 
     if (attachments.length > 0) {
-      await recordArtifact(this.task, {
-        type: 'internal:toMatchScreenshot',
-        kind: 'visual-regression',
-        message: result.message,
-        attachments,
-      })
+      // @todo understand how to handled this
+      // await recordArtifact(this.task, {
+      //   type: 'internal:toMatchScreenshot',
+      //   kind: 'visual-regression',
+      //   message: result.message,
+      //   attachments,
+      // })
     }
   }
 
--- a/packages/vitest/src/node/types/browser.ts
+++ b/packages/vitest/src/node/types/browser.ts
@@ -575,6 +575,44 @@ export interface ToMatchScreenshotOptions {
    * @default path.resolve(root, attachmentsDir, testFileDirectory, testFileName, `${arg}-${browserName}-${platform}${ext}`)
    */
   resolveDiffPath?: ToMatchScreenshotResolvePath
+  /**
+   * Overrides the default filesystem access used to read and write screenshots.
+   *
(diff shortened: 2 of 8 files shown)


packages/vitest/src/node/types/browser.ts
```diff
@@ -575,6 +575,44 @@ export interface ToMatchScreenshotOptions {
...
+    }) => Promise<void>
+  }
 }
 
 export interface ToMatchScreenshotComparators {}
+
+/**
+ * Binary image data accepted and returned by {@linkcode ToMatchScreenshotOptions.io|io}'s `read` and `write` functions.
+ */
+export type TypedArray = Buffer<ArrayBufferLike> | Uint8Array<ArrayBufferLike>
```
macarie, line 618: Technically a breaking change since the original included `Uint8ClampedArray`, but reading-wise I think the API is the same? 😬

packages/browser/src/client/tester/expect/toMatchScreenshot.ts
```diff
@@ -91,12 +91,13 @@ export default async function toMatchScreenshot(
     }
 
     if (attachments.length > 0) {
-      await recordArtifact(this.task, {
-        type: 'internal:toMatchScreenshot',
-        kind: 'visual-regression',
-        message: result.message,
-        attachments,
-      })
+      // @todo understand how to handled this
```
macarie, line 94: tbd

packages/vitest/src/node/types/browser.ts
```diff
@@ -575,6 +575,44 @@ export interface ToMatchScreenshotOptions {
...
+    write: (data: {
+      /** The path resolved by {@linkcode resolveScreenshotPath|browser.expect.toMatchScreenshot.resolveScreenshotPath} or {@linkcode resolveDiffPath|browser.expect.toMatchScreenshot.resolveDiffPath}. */
+      path: string
+      /** The image data to write, as a `Buffer` or `Uint8Array`. */
+      data: TypedArray
+      /** Indicates which type of image is being written, so implementations can apply different handling (e.g. retention policies) for reference, actual, and diff images. */
+      kind: 'reference' | 'actual' | 'diff'
+      /** The {@linkcode https://vitest.dev/api/advanced/test-project|TestProject} the test belongs to. */
+      project: TestProject
+    }) => Promise<void>
```
macarie, line 609: Would it make sense to provide the default implementation too? So users could defer diff and actual writing to Vitest and only handle references.

macarie (commented):
Not sure if it makes sense to merge this before the saved snapshots one, as we might have to change this implementation or have some blockers on that one...

I'm not sure how to best handle artifacts either.

PR #10702, opened 2026-07-02
fix(ui): remove file nodes when a test file is deleted or renamed

eldardada:
### Description

Deleting a test file leaves its entry in the UI sidebar until Vitest is fully restarted, and renaming a file leaves a stale duplicate next to the new one. This was reported by @SimonEast in #10670 (and on #10677).

The watcher already handles the deletion server-side: `onFileDelete` drops the file from `state.filesMap` and reports `onTestRemoved`. But `onTestRemoved` was only a terminal-reporter hook and was never part of `WebSocketEvents`, so the browser never learned the file was gone. The explorer tree keeps file nodes directly under its root and only ever adds/updates them while collecting, so a removed file lingered (and a rename left a duplicate) because nothing removed the root node.

This forwards `onTestRemoved` over the websocket and handles it on the client:

- `WebSocketReporter.onTestRemoved` broadcasts the removed file path to every connected client (and `onTestRemoved` is added to `WebSocketEvents` / the birpc `eventNames`).
- On the client, `StateManager.removeFiles` drops the file and its task ids from `filesMap`/`idMap`, so it is not resurrected on the next collect.
- `ExplorerTree.removeFiles` drops the matching file node and its descendants from the tree, refreshes `uiFiles`, and re-runs the filter.

Rename is covered by the same change: the old path is removed via `onTestRemoved` while the new path is added through the normal collect, so no duplicate remains.

This is a follow-up to #10677, which only reconciled tasks *within* a file that is re-collected (a test removed or commented out), not the file node itself.

### Tests

Added `packages/ui/client/composables/explorer/tree-remove.spec.ts`: it collects two files, removes one, and checks that the removed file node and every descendant are dropped while the untouched file stays; a no-op case where no tracked file matches; and that `StateManager.removeFiles` clears the file and its task ids from the run state.

I went with a unit test rather than a browser e2e because reproducing the watcher delete/rename through the UI is timing-dependent, while the tree/state reconciliation is deterministic.

### Documentation
- [ ] No user-facing API changes, so no documentation update is needed.

### Changesets

### DIFF
--- a/packages/ui/client/composables/explorer/tree-remove.spec.ts
+++ b/packages/ui/client/composables/explorer/tree-remove.spec.ts
@@ -1,0 +1,147 @@
+import type { RunnerTask, RunnerTestFile } from 'vitest'
+import type { FileTreeNode } from '~/composables/explorer/types'
+import { beforeEach, describe, expect, it } from 'vitest'
+import { explorerTree } from '~/composables/explorer'
+import { createOrUpdateFileNode } from '~/composables/explorer/utils'
+// relative on purpose: importing the client state before `~/composables/explorer`
+// trips the client module's circular init (config/testRunState TDZ)
+import { StateManager } from '../client/state'
+
+interface TestSpec {
+  id: string
+  name: string
+  state?: 'pass' | 'fail'
+}
+
+interface SuiteSpec {
+  id: string
+  name: string
+  children: TestSpec[]
+}
+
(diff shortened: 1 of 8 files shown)


packages/ui/client/composables/explorer/tree-remove.spec.ts
hi-ogawa, line 1: e2e in test/ui please.

hi-ogawa (changes requested):

PR #10707, opened 2026-07-03
fix(spy): point ESM namespace spy error to module mocking docs

### LINKED ISSUE #9467: Misleading error messages relating to spyOn modules. - "Cannot spy on export  "[X]". Module namespace is not configurable in ESM"
### Describe the bug

If I write some code like this: 

```
// main.test.ts
import { vi } from "vitest";
import * as lodash from "lodash";
vi.spyOn(lodash, "after");


```

Then this code will error with 

```typescript
TypeError: Cannot spy on export "after". Module namespace is not configurable in ESM. See: https://vitest.dev/guide/browser/#limitations
 ❯ src/main.test.ts:3:4
      1| import { vi } from "vitest";
      2| import * as lodash from "lodash";
      3| vi.spyOn(lodash, "after");
       |    ^
      4| 

Caused by: TypeError: Cannot redefine property: after
```

This links us to this documentation: 

https://vitest.dev/guide/browser/#limitations

This documentation appears to relate to browser mode specifically. 

However this error occurs whether we are using browser mode or not.* 
This causes some confusion when encountering this error. 

Furthermore - this kind of code _is_ valid in vitest v3. This breaking change is not reflected in in the migration guide. 

https://vitest.dev/guide/migration.html#vitest-4

If I'm understanding it right vitest v4 changes the way spys are applied, and directly use of spyon against external modules is no longer supported at all. 

*Note that we can reproduce a similar error that _does_ only occur in browser mode. 

```typescript
import { vi, describe, it, expect } from "vitest";
import * as foo from "./main";
vi.spyOn(foo, "foo");
```

### Reproduction

https://github.com/dwjohnston/vitest-spyon-module-issue

### System Info

```shell
System:
    OS: macOS 15.6.1
    CPU: (12) arm64 Apple M4 Pro
    Memory: 328.81 MB / 24.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.21.1 - /Users/davidjohnston/.nvm/versions/node/v22.21.1/bin/node
    npm: 11.6.2 - /Users/davidjohnston/.nvm/versions/node/v22.21.1/bin/npm
    pnpm: 10.16.1 - /Users/davidjohnston/Library/pnpm/pnpm
    bun: 1.2.21 - /Users/davidjohnston/.bun/bin/bun
  Browsers:
    Brave Browser: 143.1.85.120
    Chrome: 144.0.7559.59
    Edge: 143.0.3650.139
    Safari: 18.6
  npmPackages:
    @vitest/browser-playwright: ^4.0.17 => 4.0.17 
    vitest: 4.0.17 => 4.0.17
```

### Used Package Manager

npm

### Validations

ryandiginomad:
### Description

The `Cannot spy on export "…". Module namespace is not configurable in ESM` error links users to a **browser-mode-only** docs anchor (`/guide/browser/#limitations`). But this error fires in Node too — the reproduction in the issue (`vi.spyOn(lodash, 'after')`) has nothing to do with browser mode, so the link sends Node users somewhere irrelevant.

This points the message at the environment-agnostic module mocking guide (`/guide/mocking/modules#mocking-a-module`) instead — the standard `vi.mock(path, { spy: true })` solution — following @sheremet-va's steer in the issue to "lean into the standard" rather than suggest a mode-specific config. The message body and the guard are unchanged; only the link moves. No environment branching is added.

I used the `#mocking-a-module` section anchor (not the bare page) to match the deep-linking convention already used for the sibling module-mocking error in `packages/mocker/src/node/hoistMocks.ts`.

Resolves #9467

### Tests

Verified locally (`fix` reverted → test fails, `fix` applied → passes):

- **Node** (`test/unit`): added a regression test in `vi.spyOn() edge cases` that fakes a non-configurable ESM namespace and asserts the message points at the mocking guide and **not** at `/guide/browser`. This is the path the browser snapshot test didn't cover.
- **Browser** (`test/browser`): updated the existing inline snapshot to the new link.

### Open questions for maintainers

- The `{ spy: true }` remedy on the target page currently sits inside a "Browser Mode Support" danger callout — a Node reader landing there might still hesitate. Happy to open a small follow-up to make that section's framing environment-agnostic if you'd like, but kept it out of this PR to keep the change minimal.
- I left the error message body as-is (link only). If you'd prefer it to inline the `vi.mock(path, { spy: true })` hint directly, I can do that too — just wanted to avoid over-reaching (the earlier #9505 went that route).
- The issue also notes the v3→v4 behavior change isn't in the migration guide. That felt separable from this message fix, so I left it out — happy to follow up if useful.

### 🤖 AI assistance

Developed with Claude Code. I diagnosed the throw site, verified the docs anchor lands on the right content, and confirmed the fix with a local red→green test run. I've reviewed every line of this diff.

### DIFF
--- a/packages/spy/src/index.ts
+++ b/packages/spy/src/index.ts
@@ -410,7 +410,7 @@ export function spyOn<T extends object, K extends keyof any>(
         || error.message.includes('can\'t redefine non-configurable property'))
     ) {
       throw new TypeError(
-        `Cannot spy on export "${String(key)}". Module namespace is not configurable in ESM. See: https://vitest.dev/guide/browser/#limitations`,
+        `Cannot spy on export "${String(key)}". Module namespace is not configurable in ESM. See: https://vitest.dev/guide/mocking/modules#mocking-a-module`,
         { cause: error },
       )
     }
--- a/test/browser/test/mocking.test.ts
+++ b/test/browser/test/mocking.test.ts
@@ -12,7 +12,7 @@ it('spying on an esm module prints an error', () => {
     }
   })()
   expect(error.name).toBe('TypeError')
-  expect(error.message).toMatchInlineSnapshot(`"Cannot spy on export "calculator". Module namespace is not configurable in ESM. See: https://vitest.dev/guide/browser/#limitations"`)
+  expect(error.message).toMatchInlineSnapshot(`"Cannot spy on export "calculator". Module namespace is not configurable in ESM. See: https://vitest.dev/guide/mocking/modules#mocking-a-module"`)
 
   expect(error.cause).toBeInstanceOf(TypeError)
 })
--- a/test/unit/test/mocking/vi-spyOn.test.ts
+++ b/test/unit/test/mocking/vi-spyOn.test.ts
@@ -19,6 +19,39 @@ describe('vi.spyOn() edge cases', () => {
     expect(fn3.length).toBe(3)
   })
 
+  test('spying on a non-configurable ESM namespace export points to the module mocking docs, not browser mode', () => {
+    // Fake an ES module namespace: `Symbol.toStringTag` is 'Module' and the
+    // export is non-configurable, so `Object.defineProperty` throws
+    // "Cannot redefine property" — the same guard the real ESM path hits.
+    const namespace: Record<string, unknown> = {}
+    Object.defineProperty(namespace, Symbol.toStringTag, { value: 'Module' })
+    Object.defineProperty(namespace, 'answer', {
+      value: () => 42,
+      enumerable: true,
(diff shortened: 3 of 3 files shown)


ryandiginomad:
Quick check-in on this one — it's been in the queue a couple of weeks. I've rebased it on main and CI is green.

It's a small fix following the direction @sheremet-va outlined in #9467: the `vi.spyOn` ESM namespace error currently points Node users at browser-mode docs, so this points it at the environment-agnostic mocking guide instead. Happy to adjust the wording or the target anchor if you'd prefer something different.

sheremet-va (approved):

PR #10708, opened 2026-07-04
perf: serve warm modules to workers in one round-trip, enable the node compile cache

sheremet-va:
Two runtime performance changes.

## What changed

**1. Warm-module snapshot (`fetchWarmModules`).** With `isolate: true`, every fresh worker fetched each module of its test file's import graph through one strictly sequential RPC round-trip — on an app-shaped fixture that is ~122 round-trips per file and ~97 ms of pure serial waiting, *even when the server had already transformed every module*. Most of the per-call latency is queueing on the main process event loop, which serves all workers at once, so the cost compounds with worker count.

Workers now ask the server **once per run request** for every module of their files' import graphs that is already stored on disk (the forks pool tmp files, or `experimental.fsModuleCache`) and read the code directly; only misses go through the per-module `fetch` path. Externalize verdicts for resolved urls ride the same snapshot.

Correctness notes:
- The snapshot is requested at the first *dependency* fetch — by then the entry file's transform has connected its import graph on the server, so the snapshot actually covers the file's transitive dependencies.
- It is re-requested per run request, and the server builds it from the live module graph: a file edit invalidates the module server-side and it simply drops out of the next snapshot, keeping reused (`isolate: false`) workers in sync in watch mode.
- The mock/builtin resolution ladder runs *before* the snapshot lookup, so `vi.mock` semantics are unchanged.
- Only externalize verdicts for already-resolved urls are shared (an unresolved bare specifier resolves through the requesting environment's plugin container, so its verdict is environment/importer-specific), keyed by the Vite server instance so a restart drops them.

**2. Node compile cache.** The `vitest` bin now calls `module.enableCompileCache()` before importing the CLI and propagates `NODE_COMPILE_CACHE` to workers (child processes do not inherit it otherwise). The CLI graph and every worker's bundle skip V8 recompilation across processes and runs. Respects a pre-set `NODE_COMPILE_CACHE` and `NODE_DISABLE_COMPILE_CACHE`; no-op before Node 22.8. Workers get the cache disabled when coverage is enabled, because V8 serializes compile-cached scripts without the source positions precise coverage relies on.

**3.** `experimental.fsModuleCache` now remembers the on-disk location when *saving* a fresh transform (previously only when reading it back), so the snapshot is effective in the first session already, not only in the next one.

## Benchmarks

App fixture = 176 TS source modules + 40 test files with heavily overlapping import graphs (layered utils/services/features, a 60-export barrel, a depth-15 chain, one externalized dep); `tiny`/`huge` bracket the startup-dominated and execution-dominated extremes. min of N interleaved runs, Apple Silicon 10 cores, Node 24. Measured while this PR was stacked on #10685 (the base column is that branch); the PR is now based on `main` directly and the numbers were not re-run — the changes themselves are unchanged.

| benchmark | base | this PR | this PR + `fsModuleCache` |
|---|---:|---:|---:|
| tiny (1 test, 1 file, forks) | 267 ms | 205 ms (−23%) | 152 ms (−43%) |
| huge (10k tests, 1 file, forks) | 921 ms | 864 ms (−6%) | 574 ms (−38%) |
| app node, forks isolate:true (default) | 1266 ms | 1050 ms (−17%) | 751 ms (−41%) |
| app node, forks isolate:false | 629 ms | 613 ms (−3%) | 346 ms (−45%) |
| app node, threads isolate:true | 967 ms | 773 ms (−20%) | 652 ms (−33%) |
| app node, threads isolate:false | 481 ms | 411 ms (−15%) | 283 ms (−41%) |
| app jsdom, forks isolate:true | 3809 ms | 3393 ms (−11%) | 3209 ms (−16%) |
| app jsdom, forks isolate:false | 1184 ms | 1095 ms (−8%) | 880 ms (−26%) |
| app happy-dom, forks isolate:true | 2134 ms | 1969 ms (−8%) | 1787 ms (−16%) |

Why the shape of the numbers: the snapshot removes a cost that scales with modules × test files (biggest on isolated many-file runs and on warm `fsModuleCache` sessions where transform work is already gone), while the compile cache removes a fixed per-process boot cost (biggest on single-file runs and visible in every worker). The jsdom rows dilute in relative terms because jsdom setup (~550 ms per worker) dominates isolated DOM runs — that cost is untouched here and remains the main argument for the environment diagnostic in #10710.

## Decomposition and cold/warm follow-up

The table above is base → PR → PR + `fsModuleCache`. These two blocks break that down: which of the three changes each delta comes from, and the cold-vs-warm behaviour of `fsModuleCache`. Reproduced on the original stack (base = #10685, this PR cherry-picked on top), same fixtures and harness, `min` of N interleaved runs, Apple Silicon 10-core / Node 24. Each block is its own session — compare per-row deltas, not absolute ms across blocks (the two blocks are self-consistent where they overlap: `+ compile cache` below == `fsModuleCache off` further down, and `+ fsModuleCache (warm)` == `warm`).

### Per-lever contribution

Each row turns on one more lever on top of the previous. warm-modules is measured by running `dist/cli.js` directly (the bin is bypassed, so the compile cache is off for that row).

| stacked on #10685 | app forks isolate:true | app threads isolate:true | app forks isolate:false |
|---|---:|---:|---:|
| base | 1167 ms | 880 ms | 673 ms |
| + warm-modules | 1061 ms (−9%) | 850 ms (−3%) | 607 ms (−10%) |
| + compile cache | 1018 ms (−13%) | 799 ms (−9%) | 577 ms (−14%) |
| + `fsModuleCache` (warm) | 789 ms (−32%) | 663 ms (−25%) | 352 ms (−48%) |

- warm-modules carries the isolated-**forks** win on its own (−9…−10%): the forks pool writes tmp copies (`cacheFs: true`), so the snapshot already has on-disk modules to serve without `fsModuleCache`.
- On **threads** warm-modules alone is ~0 (−3%, noise): threads don't write tmp copies, so without `fsModuleCache` the snapshot finds nothing on disk and falls back to per-module fetches — the threads delta in that row is the compile cache. `fsModuleCache` writes on-disk copies for every pool, which is what makes the snapshot pay off on threads as well.
- compile cache is a flat ~−4…−6 pts everywhere, and the dominant lever on single-file runs (`tiny` above).
- `fsModuleCache` is the largest lever (−15…−33 additional pts). It is default-on (separate PR), so the `fsModuleCache` column is the default steady state, not an opt-in.

### `fsModuleCache` cold vs warm

The `fsModuleCache` gain is a warm-cache steady state; the run that *populates* the cache pays for it. App fixture, forks isolate:true, on this PR (compile cache + Vite optimizeDeps kept warm, so cache presence is the only variable; % relative to `fsModuleCache` off):

| `fsModuleCache` | wall (min) | transform phase |
|---|---:|---:|
| off | 1010 ms | 2.06 s |
| cold — first run, writes 217 files / 0.9 MB | 1080 ms (+7%) | 2.20 s |
| warm — cache present | 780 ms (−23%) | 0.52 s |

The cold run is slightly *slower* (full transforms plus the disk writes); the warm run collapses the transform phase (2.06 s → 0.52 s) because a fresh worker reads the transformed modules off disk instead of re-transforming them. Keys are content-hashed, so an edit invalidates only the changed module and leaves the rest warm. With `fsModuleCache` default-on, the warm row is what a repeat run — or a CI shard with a restored cache — actually gets.

### DIFF
--- a/packages/vitest/src/node/cache/fsModuleCache.ts
+++ b/packages/vitest/src/node/cache/fsModuleCache.ts
@@ -34,7 +34,7 @@ export class FileSystemModuleCache {
   private rootCache: string
   private metadataFilePath: string
 
-  private version = '1.0.0-beta.4'
+  private version = '1.0.0-beta.5'
   private fsCacheRoots = new WeakMap<ResolvedConfig, string>()
   private fsEnvironmentHashMap = new WeakMap<DevEnvironment, string>()
   private fsCacheKeyGenerators = new Set<CacheKeyIdGenerator>()
@@ -214,6 +214,7 @@ export class FileSystemModuleCache {
           mode: config.mode,
           consumer: config.consumer,
           resolve: config.resolve,
+          injectCjsGlobal: vitestConfig.injectCjsGlobals,
           // plugins can have different options, so this is not the best key,
           // but we cannot access the options because there is no standard API for it
           plugins: config.plugins
--- a/packages/vitest/src/node/cache/fsModuleCache.ts
+++ b/packages/vitest/src/node/cache/fsModuleCache.ts
@@ -1,4 +1,5 @@
-import type { DevEnvironment, FetchResult } from 'vite'
+import type { DevEnvironment } from 'vite'
+import type { ModuleType, VitestFetchResult } from '../../types/general'
 import type { Vitest } from '../core'
 import type { ResolvedConfig } from '../types/config'
 import fs, { existsSync, mkdirSync, readFileSync } from 'node:fs'
@@ -110,12 +111,13 @@ export class FileSystemModuleCache {
       code,
       importedUrls: meta.importedUrls,
       mappings: meta.mappings,
+      moduleType: meta.moduleType,
     }
   }
 
-  async saveCachedModule<T extends FetchResult>(
+  async saveCachedModule(
     cachedFilePath: string,
-    fetchResult: T,
+    fetchResult: VitestFetchResult,
     importedUrls: string[] = [],
     mappings: boolean = false,
   ): Promise<void> {
@@ -126,6 +128,7 @@ export class FileSystemModuleCache {
         url: fetchResult.url,
         importedUrls,
         mappings,
+        moduleType: fetchResult.moduleType,
       } satisfies Omit<CachedInlineModuleMeta, 'code'>
       debugFs?.(`${c.yellow('[write]')} ${fetchResult.id} is cached in ${cachedFilePath}`)
       await atomicWriteFile(cachedFilePath, `${fetchResult.code}${cacheComment}${this.toBase64(result)}`)
@@ -366,6 +369,7 @@ export interface CachedInlineModuleMeta {
   code: string
   mappings: boolean
   importedUrls: string[]
+  moduleType?: ModuleType
 }
 
 /**
--- a/packages/vitest/src/node/pool.ts
+++ b/packages/vitest/src/node/pool.ts
@@ -127,6 +127,14 @@ export function createPool(ctx: Vitest): ProcessPool {
             ...project.config.env,
           }
 
+          // V8 serializes compile-cached scripts without the source positions
+          // that precise coverage relies on — never let workers (or the
+          // processes they spawn) use the compile cache when coverage is on
+          if (ctx.config.coverage.enabled) {
+            delete env.NODE_COMPILE_CACHE
+            env.NODE_DISABLE_COMPILE_CACHE = '1'
+          }
+
           // env are case-insensitive on Windows, but spawned processes don't support it
           if (isWindows) {
             for (const name in env) {
--- a/packages/vitest/src/node/pool.ts
+++ b/packages/vitest/src/node/pool.ts
@@ -128,9 +128,12 @@ export function createPool(ctx: Vitest): ProcessPool {
           }
 
           // V8 serializes compile-cached scripts without the source positions
-          // that precise coverage relies on — never let workers (or the
-          // processes they spawn) use the compile cache when coverage is on
-          if (ctx.config.coverage.enabled) {
+          // that precise coverage relies on, so the compile cache must stay off
+          // for the v8 provider (and custom providers, whose mechanism we can't
+          // assume) in workers and any process they spawn. istanbul instruments
+          // the source at transform time, so the cache is harmless there and the
+          // boot speedup is kept.
+          if (ctx.config.coverage.enabled && ctx.config.coverage.provider !== 'istanbul') {
             delete env.NODE_COMPILE_CACHE
             env.NODE_DISABLE_COMPILE_CACHE = '1'
           }
--- a/packages/vitest/src/node/pools/rpc.ts
+++ b/packages/vitest/src/node/pools/rpc.ts
@@ -4,11 +4,13 @@ import type { RuntimeRPC } from '../../types/rpc'
 import type { TestProject } from '../project'
 import type { ResolveSnapshotPathHandlerContext } from '../types/config'
 import { existsSync, mkdirSync } from 'node:fs'
+import { readFile } from 'node:fs/promises'
 import { fileURLToPath } from 'node:url'
 import { cleanUrl } from '@vitest/utils/helpers'
 import { isBuiltin, toBuiltin } from '../../utils/modules'
 import { handleRollupError } from '../environments/fetchModule'
 import { normalizeResolvedIdToUrl } from '../environments/normalizeUrl'
+import { detectModuleType } from '../resolver'
 
 interface MethodsOptions {
   cacheFs?: boolean
@@ -25,6 +27,16 @@ interface MethodsOptions {
 // Keyed by the Vite server so a server restart drops the accumulated verdicts
 // together with the resolver that produced them.
 const warmExternals = new WeakMap<ViteDevServer, Record<string, FetchResult>>()
+
+// `detectModuleType` only reads the original source in the ambiguous case where
+// the transformed code both looks like ESM and mentions a CommonJS variable;
+// mirror the guard `ModuleFetcher.sourceLoader` uses for virtual modules
+function warmSourceLoader(file: string | null): (() => Promise<string | null>) | undefined {
+  if (!file || file.startsWith('\x00') || file.startsWith('virtual:')) {
+    return undefined
+  }
+  return () => readFile(file, 'utf-8').then(source => source, () => null)
+}
 
 export function createMethodsRPC(project: TestProject, methodsOptions: MethodsOptions = {}): RuntimeRPC {
   const vitest = project.vitest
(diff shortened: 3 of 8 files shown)


packages/vitest/src/node/pool.ts
```diff
@@ -127,6 +127,14 @@ export function createPool(ctx: Vitest): ProcessPool {
             ...project.config.env,
           }
 
+          // V8 serializes compile-cached scripts without the source positions
```
AriPerkkio, line 130: What does this mean in practice?

packages/vitest/src/node/pools/rpc.ts
```diff
@@ -56,6 +78,70 @@ export function createMethodsRPC(project: TestProject, methodsOptions: MethodsOp
...
+
+        const transformResult = node.transformResult
+        if (!transformResult || node.id == null) {
+          continue
+        }
+        // the transformed code is already stored on disk either by the forks
+        // pool (`cacheFs`) or by `experimental.fsModuleCache` — the worker can
+        // read the file itself instead of fetching each module separately.
+        // invalidated modules lose `transformResult` and drop out automatically
+        const tmp = transformResult.__vitestTmp ?? (transformResult as { _vitest_tmp?: string })._vitest_tmp
```
AriPerkkio, line 121: Now that `_vitest_tmp` is used/leaking outside the `fetchModule.ts`, maybe it's time to add it to typings next to `__vitestTmp`.

AriPerkkio (dismissed):
So when worker requests test file (or anything), main thread instructs where the dependencies of the requested file are on the cache, so that no new RPC calls need to be made? Sounds like good optimization.

packages/vitest/vitest.mjs
```diff
@@ -1,2 +1,19 @@
...
+
+// Enable Node's on-disk compile cache before importing the CLI so both the CLI
+// graph and (via the inherited env variable) every spawned worker skip V8
+// recompilation of unchanged modules. `enableCompileCache()` only affects the
+// current process — child processes pick the cache up from NODE_COMPILE_CACHE.
+// Respects an explicit NODE_COMPILE_CACHE and NODE_DISABLE_COMPILE_CACHE; the
+// API is not available before Node 22.8.
+try {
+  const result = module.enableCompileCache?.()
+  if (result?.directory && !process.env.NODE_COMPILE_CACHE) {
```
AriPerkkio, line 12: Curious how much this speeds up test runs on its own 👀

sheremet-va: around 8%

packages/vitest/src/node/pools/rpc.ts
sheremet-va: Opened a separate PR to merge them, actually - https://github.com/vitest-dev/vitest/pull/10733

packages/vitest/src/node/pool.ts
hi-ogawa: Looks like this is basically called out in https://nodejs.org/api/module.html#limitations-of-the-compile-cache

> Currently when using the compile cache with [V8 JavaScript code coverage](https://v8project.blogspot.com/2017/12/javascript-code-coverage.html), the coverage being collected by V8 may be less precise in functions that are deserialized from the code cache. It's recommended to turn this off when running tests to generate precise coverage.

packages/vitest/src/node/pool.ts
```diff
@@ -127,6 +127,14 @@ export function createPool(ctx: Vitest): ProcessPool {
...
+          // that precise coverage relies on — never let workers (or the
+          // processes they spawn) use the compile cache when coverage is on
+          if (ctx.config.coverage.enabled) {
```
hi-ogawa: Can we keep compile cache for istanbul provider?

packages/vitest/src/node/pools/rpc.ts
```diff
@@ -14,6 +16,16 @@ interface MethodsOptions {
...
 
+// externalize verdicts served during this session, shared with fresh workers
+// via `fetchWarmModules`. Only verdicts for already-resolved urls are stored:
+// an unresolved specifier (a runtime-variable dynamic import of a bare name)
+// resolves through the requesting environment's plugin container, so its
+// verdict is environment- and importer-specific and cannot be shared. Resolved
+// ids always externalize the same way (the resolver memoizes by resolved id).
+// Keyed by the Vite server so a server restart drops the accumulated verdicts
+// together with the resolver that produced them.
+const warmExternals = new WeakMap<ViteDevServer, Record<string, FetchResult>>()
```
hi-ogawa: My agent called out externals should be keyed by `environment` too. I didn't repro but it sounds right. Edge case but wouldn't switching  `WeakMap<DevEnvironment, Record<..>>` to more natural?

<details><summary>AI review</summary>

The comment is incorrect because it conflates the requested URL with the resolved ID.

Call chain:

1. `rpc.fetch()` receives a URL and selects an environment at `packages/vitest/src/node/pools/rpc.ts:49`.
2. `ModuleFetcher.fetch()` passes even leading-slash URLs to that environment’s module graph at `fetchModule.ts:66`.
3. Vite’s `ensureEntryFromUrl()` resolves the URL through the environment-specific plugin container:
   - `node_modules/vite/dist/node/chunks/node.js:23694`
   - `node_modules/vite/dist/node/chunks/node.js:24717`
4. Externalization is decided using the resulting `moduleGraphModule.id` at `fetchModule.ts:77`.
5. The PR nevertheless caches the verdict under the original `url` at `rpc.ts:70`.

Vite explicitly permits environment-specific resolution. This repository already has an example where the same source resolves differently according to `this.environment.config.consumer`:

`test/coverage-test/fixtures/configs/vitest.config.conditional.ts:8`

A concrete failure can occur like this:

```text
ssr environment:
  /conditional -> /node_modules/pkg/server.js -> externalized

client environment:
  /conditional -> /project/src/browser.ts -> should be inlined
```

After the SSR fetch, the current cache contains:

```text
server -> "/conditional" -> external result
```

`fetchWarmModules()` then adds that result to the client snapshot without consulting the client environment. The worker hits `/conditional` and bypasses client resolution entirely.

So:

- Server scoping is reasonable.
- Resolved-ID externalization caching is reasonable.
- The implementation caches by unresolved environment URL, not resolved ID.
- `url[0] === '/'` does not prove that no further resolution occurs.

The minimal safe fix is to scope by environment object:

```ts
WeakMap<DevEnvironment, Record<string, FetchResult>>
```

or by server, then environment name, then URL. This is a concrete correctness issue, not merely speculative.

</details>

hi-ogawa (commented):
Very cool!

hi-ogawa:
Also another AI artifact to help me picture the idea https://artifacts.hiro18181.workers.dev/vitest-pr-10708-warm-modules#example

packages/vitest/src/node/pools/rpc.ts
sheremet-va: Sure, makes sense. The vitest externalisation happens on the server level though, so the only way this can break is if the plugin does conditional resolve (which is what is pointed out by the summary). It's a bit confusing

sheremet-va: implemented

sheremet-va:
I also added the cache for the moduleType, otherwise we need to resolve it all the time

packages/vitest/src/node/cache/fsModuleCache.ts
hi-ogawa, line 1: Was thinking to point out in https://github.com/vitest-dev/vitest/pull/10709 that if we fs cache `moduleType` then we need to bust config hash with `injectCjsGlobal` flag.

sheremet-va: good point

hi-ogawa: Also technically I don't think fs caching `moduleType` is too crucial for warm snapshot design because the single warm snapshot can still batch compute `moduleType` so rpc saving is effective.

sheremet-va: fixed

sheremet-va: > Also technically I don't think fs caching moduleType is too crucial for warm snapshot design because the single warm snapshot can still batch compute moduleType so rpc saving is effective.

It will recompute it for every worker every time; it doesn't affect performance much, but why do that when we can just cache it, like we do other things

hi-ogawa (approved):
Looks good 👍

PR #10709, opened 2026-07-04
feat: add toggable `injectCjsGlobals` option

### LINKED ISSUE #10292: Have a `strictModuleType` option to disable CJS-ESM interop

Currently Vitest injects `import.meta`, static `import` support, `__dirname`, `__filename`, `require`, `exports`, `module` into every inlined module. This is not how modules work in the wild. Browsers do not support CJS globals and CJS doesn't support static imports and `import.meta` property.

The main reason why Vitest keeps supporting this is that it is faster to just inject these variables rather than try to find what kind of module is running (ESM or CJS). It also removes the burden from the user to setup their testing pipeline.

However, this also means that Vitest environment is different from the target runtime. Tests shouldn't have access to CJS variables unless they are written in a CJS module (an external module for example or CJS files that use ESM TypeScript syntax).

### Suggested solution

Introduce a `strictModuleType` flag to disable CJS<->ESM interop:

```ts
export default defineConfig({
  test: {
    strictModuleType: true, // CJS globals are not injected in ESM
  },
})
```

### Alternative

`experimental.viteModuleRunner: false` solves this by running code in the target environment instead of using a module runner. This is a viable solution for many teams, but it is also limited in its feature set.

### Additional context

Vitest runs code using Vite's module runner. Module runner works by replacing all static and dynamic imports with a separate `__vite_ssr_import__` function that intercepts requests. This means that Vitest features only support files that use ESM syntax (real ESM files or TypeScript files that use ESM syntax, but are bundled into CJS files separately from tests).

Vitest does not support CJS syntax directly (we don't even expose a CJS version of `vitest` entry), but we do support importing CJS dependencies. Required files leave the module runner runtime and are executed directly by Node (this means that mocking doesn't work, for example). This means that we can inject CJS globals only in inlined _dependencies_.

### Validations

sheremet-va:
This PR adds `injectCjsGlobals` for projects that keep using the module runner to run tests, but need a more fair environment (ESM files do not have access to CJS globals).

- CJS globals are still injected in inlined modules if they have the type `cjs` (we use Node.js algorithm to check it)
- Fixes https://github.com/vitest-dev/vitest/issues/10292

Thanks Fable 

I am hoping to disable this by default in Vitest 6


packages/vitest/src/node/resolver.ts
```diff
@@ -166,6 +173,114 @@ async function isValidNodeImport(id: string) {
...
+  return 'esm'
+}
+
+const packageScopeTypeCache = new Map<string, ModuleType | 'none'>()
+
+// mirrors LOOKUP_PACKAGE_SCOPE from the ESM resolution algorithm:
+// the lookup stops at the first package.json and never crosses
+// the "node_modules" boundary, so typeless dependencies don't
+// inherit the `type` field of the user's project
+function lookupPackageScopeType(directory: string): ModuleType | 'none' {
```
hi-ogawa: The logic here is similar to `findNearestPackageData` (it's used above `isValidNodeImport` for externalization heuristics) where the diff is apparently what AI got right over the original `findNearestPackageData` around `node_modules` case.

I don't doubt AI implemented wrong, but would like some higher level clarification regarding why not `findNearestPackageData` and if `findNearestPackageData` should be also unified.

packages/vitest/src/runtime/moduleRunner/moduleEvaluator.ts
```diff
@@ -298,22 +313,50 @@ export class VitestModuleEvaluator implements ModuleEvaluator {
...
+      __vite_track_exportName__ || __vite_ssr_exportName__,
+    ]
+
+    // TODO@discuss switch the default in Vitest 6(?)
+    // backwards compat for vite-node
+    const injectCjsGlobals = this.options.injectCjsGlobals !== false
+      // the module type is provided by the server only when `injectCjsGlobals` is disabled.
+      // CommonJS modules always receive these variables because they are part
+      // of the module scope, without them the module cannot be evaluated at all
+      || (module.meta as { moduleType?: ModuleType } | undefined)?.moduleType === 'cjs'
```
hi-ogawa: This `as` typing looks too surgical. Can we get type through `type VitestFetchResult` or something so it connects producer and consumer?

packages/vitest/src/node/environments/fetchModule.ts
```diff
@@ -281,6 +296,9 @@ class ModuleFetcher {
       tmp: cachePath,
       url: cachedModule.url,
       invalidate: false,
+      moduleType: this.detectModuleType
+        ? await detectModuleType(cachedModule.file, cachedModule.code, this.sourceLoader(cachedModule.file))
+        : undefined,
```
hi-ogawa, line 299: Was wondering if `moduleType` also get cached, but my AI debunked me saying it wouldn't give you much.

hi-ogawa (commented):
lgtm. Made a few comments.

> I am hoping to disable this by default in Vitest 6

Also agree.

packages/vitest/src/node/resolver.ts
sheremet-va: It did use that one, but I asked it to check if node actually does the same, turns out there are some subtle differences. For example, node doesn’t leave node_modules folder when looking for pkg.json

hi-ogawa: To clarify, I want that kind of clarification as comments more than AI comment just explaining what it implements locally 😃

sheremet-va: I don’t understand why it needs to be in the code? It’s a user decision. If it wasn’t AI, the code would’ve been perfectly fine to merge, it explains the difference  already

sheremet-va: But they could probably just be merged

packages/vitest/src/runtime/moduleRunner/moduleEvaluator.ts
sheremet-va: moved

packages/vitest/src/node/resolver.ts
sheremet-va: replaced the older function with a new one

sheremet-va:
Important to sync with https://github.com/vitest-dev/vitest/pull/10708 (or the other way around) because they touch the same mechanism

hi-ogawa (approved):

packages/vitest/src/node/resolver.ts
hi-ogawa: Great! Thanks for unifying them.

PR #10710, opened 2026-07-04
feat: add `vitest doctor`

sheremet-va:
Part 3 of 3 of the perf discoverability work. Stacked on #10821 (performance hints), which is stacked on #10820 (duration breakdown percentages) — only the last commits are new here.

`vitest doctor` turns the hints' estimate into a measurement: resolves the config, prunes candidates from it (`threads`, `vmThreads` for DOM suites, `environment: 'happy-dom'` for jsdom projects, `isolate: false`, `threads + isolate: false` as the honest competitor for vm-pool configs, and `fsModuleCache: true` when the cache is off), runs the suite under each, and recommends with a ready-to-paste snippet:

```
Results (min of 3 runs each)

  baseline (pool: forks · isolate: true)  4.08s
  pool: 'threads'                         3.64s (-11%)
  pool: 'vmThreads'                       1.33s (-67%)
  isolate: false                          1.28s (-69%)

Recommendation: pool: 'vmThreads' (-67%)

  // vitest.config.ts

  export default defineConfig({
    test: {
      pool: 'vmThreads', // measured -67% on this suite
    },
  })
```

- `isolate: false` is validated with two shuffled-order runs (explicit distinct seeds, so a user-pinned `sequence.seed` can't silently produce the same order twice); a failing validation excludes the candidate and reports which tests depend on isolation.
- Failing candidates print an excerpt of their errors — "this suite can't move to vmThreads" comes with the reason.
- "Keep the current configuration" is always a measured verdict: even with no candidates the baseline is measured and reported.
- Every measured run executes through a generated runner script (`createVitest` + JSON overrides) instead of the CLI binary, so the baseline and every candidate pay the identical entry cost, and candidates can express what CLI flags cannot: `environment: 'happy-dom'` swaps the environment of jsdom projects only — other projects keep theirs (covered by a fixture whose node project fails if the swap leaks). Offered whenever any project runs jsdom and the package actually resolves — checked with `import.meta.resolve` from the vitest package, which is exactly how the environment imports it. A recommendation comes with a "happy-dom implements the DOM differently" caveat.
- `fsModuleCache: true` gets one untimed priming run before its measured reps (candidates support `primeRuns`), so the reported number is what repeated runs pay with a populated cache — consistent with the warm steady state everything else is measured in.
- The baseline always gets ≥2 runs so a cold first run (which warms caches for everyone) is never compared against warm candidates; candidates near-tied with the fastest are resolved in favor of the one that keeps isolation.
- Every measurement runs the full suite, browser projects included — `isolate: false` affects browser mode too, and a recommendation is only honest when measured on the run the user actually pays for. Candidates that cannot affect browser projects (`pool`, `environment`, the fs cache) are still picked from the node-side projects only, and a browser-only workspace still gets the `isolate: false` candidate and the `maxWorkers` probes. Candidates are killed after 4× the baseline duration; SIGINT/SIGTERM kills the active child.
- Doctor also probes lower `maxWorkers` values on top of the winning configuration - every worker funnels transforms through the single main-thread Vite server, so past a certain count more workers are slower. Greedy descent from half the current count, keeping a step only when it's ≥5% faster:

```
  measuring maxWorkers: 4 (with pool: 'vmThreads')... 0.94s
  measuring maxWorkers: 2 (with pool: 'vmThreads')... 1.12s

Recommendation: pool: 'vmThreads' + maxWorkers: 4 (-74%)
```

Docs: `vitest doctor` in `cli.md`, plus the doctor pointers in `improving-performance.md` and `config/experimental.md`.

### DIFF
--- a/docs/config/experimental.md
+++ b/docs/config/experimental.md
@@ -461,6 +461,8 @@ Environment  jsdom was created 40 times · 23.80s total, 79% of tracked time
 
 Hints never suggest changing an option that was set explicitly: if the config defines `pool`, other pools are not suggested, and an explicitly configured `isolate` is never suggested to be disabled. Hints are also printed in CI. Set the option to `false` to disable all hints, or disable them individually.
 
+To measure the impact of a configuration change instead of estimating it, run [`vitest doctor`](/guide/cli#vitest-doctor).
+
 ### experimental.diagnostics.isolate {#experimental-diagnostics-isolate}
 
 - **Type:** `boolean`
--- a/docs/config/experimental.md
+++ b/docs/config/experimental.md
@@ -533,6 +533,19 @@ interface DiagnosticsOptions {
    * @default true
    */
   environment?: boolean
+  /**
+   * Hint when test files repeatedly evaluate the same module graph
+   * (typical for barrel-file imports) and `isolate: false` would
+   * evaluate it once per worker.
+   * @default true
+   */
+  import?: boolean
+  /**
+   * Hint when transforming modules dominates the run and
+   * `experimental.fsModuleCache` would persist the results across runs.
+   * @default true
+   */
+  transform?: boolean
 }
 ```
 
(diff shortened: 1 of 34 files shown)


packages/vitest/src/node/plugins/index.ts
```diff
@@ -221,8 +221,11 @@ export async function VitestPlugin(
           delete viteConfigTest.alias
         }
 
+        // provenance must be captured before the defaults are merged in
+        const providedOptions = captureProvidedOptions(viteConfigTest, options)
```
sheremet-va: All the config shenanigans would look better after we merge the config refactor

sheremet-va:
@hi-ogawa forgot to ask for review 😢

sheremet-va:
> Isolate hint — the pool now tracks how long each worker start takes (spawn + bundle load + environment setup, summed as state.startupTime / state.workersSpawned); when isolate: true spends a significant share of the run spawning a fresh worker per test file, the summary estimates what isolate: false would save.

Technically, it should also remove the time it takes to load the module graph

> Transform hint — when transforms dominate instead (≥2s and ≥25% of tracked time) and nothing persists them, the hint points at the fs module cache; it only speaks up when the import hint didn't (e.g. isolate was set explicitly), so the two disclose progressively:

Should also note that on CI this doesn't help. Plus it's not longer in experimental

> environment: 'happy-dom' is only offered when every project runs jsdom (the CLI override hits all projects at once) and the package actually resolves — checked with import.meta.resolve from the vitest package, which is exactly how the environment imports it. A recommendation comes with a "happy-dom implements the DOM differently" caveat.

We should manually update the config for every project (do we have APIs that could do that?), "all projects use jsdom" is pretty useless as one will probably use a combination of environments

> Browser projects are excluded from the measurements via --project (no candidate affects them; they'd only dilute deltas); candidates are killed after 4× the baseline duration; SIGINT/SIGTERM kills the active child.

We should always run all tests. isolation also affects browser mode, for example.

PR #10713, opened 2026-07-05
perf(browser): pre-bundle the vitest runtime in optimizeDeps

sheremet-va:
Pre-bundles the vitest runtime for browser mode so the browser fetches a few optimized chunks instead of ~20 separately-served dist chunks (faster startup, fewer module requests).

`vitest`, `vitest/internal/browser` and `@vitest/browser/client` are moved from `optimizeDeps.exclude` to `include`. They are optimized together in a single pass, so esbuild dedupes their shared stateful chunks (the test collector, the runner, the RPC client) to a single instance — preserving module identity between the test files' `import 'vitest'` and the tester. Their transitive deps (`@vitest/utils`, `@vitest/spy`, `pathe`, `tinyrainbow`, …) are inlined into these bundles. The virtual per-server modules (`@vitest/browser/context`, …), `vite/module-runner` and `msw` must stay excluded (see the comment in `index.ts`).

**Shared-server resolver fix.** The `vitest:resolve-core` plugin (`vitestResolver.ts`) short-circuits bare `vitest` imports straight to the running copy's dist file. It now lives on the single Vite server shared by every environment, so it also governs the browser `client` environment where the runtime is pre-bundled — without care the tester would use the optimized bundle while test files receive the raw dist graph, producing two `@vitest/runner` instances and failing every `test()` with "Vitest failed to find the current suite". When the environment includes `vitest` in `optimizeDeps`, the plugin now routes the import through the optimizer (`this.resolve`) so both share one instance; every other environment keeps the pinned dist resolution.

Stack frames pointing into the optimizer cache (`.vite/vitest`) are ignored in stack traces like the un-bundled `/vitest/dist/` ones. The console-logging spec now runs with an explicit `default` reporter and asserts on Vitest's own console forwarding: once the runtime is pre-bundled the tester loads a stubbed `@vite/client`, so Vite's client console relay no longer runs.


hi-ogawa (commented):
Didn't look through yet, but it looks like it involves many independent changes. Is there a reasonable way to split PRs to ease reviews? (perhaps sourcemap related stuff?)

sheremet-va:
There are already 4 PRs stacked on top of each other 😩 It’s gonna be a merge headache. I’ll ask claude to split it 😔

packages/utils/src/source-map.ts
```diff
@@ -31,6 +31,10 @@ const stackIgnorePatterns: (string | RegExp)[] = [
   '/deps/@vitest',
   '/deps/loupe',
   '/deps/chai',
+  // the Vitest runtime is pre-bundled into a dedicated optimizer cache
+  // (`node_modules/.vite/vitest/<hash>/deps/`) in browser mode, so its internal
+  // frames must be ignored the same way as the un-bundled `/vitest/dist/` ones
+  '/.vite/vitest/',
```
hi-ogawa: Doesn't this change by vite `cacheDir` config?

packages/browser/src/node/index.ts
```diff
@@ -431,23 +431,18 @@ function resolveBrowserOptimizeDeps(
     ...(testConfig.snapshotSerializers || []),
   ]
 
+  // Keep these external (never pre-bundle by esbuild):
```
hi-ogawa: now rolldown too

```suggestion
  // Keep these external (never pre-bundle by esbuild and rolldown):
```

hi-ogawa (changes requested):
lgtm

packages/utils/src/source-map.ts
sheremet-va: Yes, this is not even needed because we already do deps/vitest

hi-ogawa: I tested one and I see `test/browser/fixtures/assertion-helper/node_modules/.vite/vitest/da39a3ee5e6b4b0d3255bfef95601890afd80709/deps/vitest.js`, which doesn't match `deps/vitest_` in this list. And also it wouldn't match if user have `cacheDir: '.../.not-vite/`. 

---

Also I think what I realized that 'm confused about is that the comment claims `the Vitest runtime`, but this `/.vite/vitest/` pattern is about ignoring all pre-bundled frames, which sound related to the topic of https://github.com/vitest-dev/vitest/pull/10728 but I lost the track...

hi-ogawa (changes requested):

packages/utils/src/source-map.ts
sheremet-va: I removed this, not sure why it wasn't pushed 🤔

sheremet-va:
Screenshot 2026-07-23 at 10 45 00

First time seeing this 🤷

hi-ogawa:
It's spinning for me too.

packages/utils/src/source-map.ts
hi-ogawa: Ah alright, you did fix it but I just couldn't see it yet :octocat:

sheremet-va:
Should be done spinning @hi-ogawa 

Removing the line is enough, I also added `deps/vitest.js` just in case, although most of the code will probably end up in chunks 👀  (But CI doesn't confirm this, everything passes 🤷 )

hi-ogawa (approved):
Now I see the commit!

PR #10721, opened 2026-07-06
fix(pool): vm pools to respect cgroupsv2 memory limit

milas:
### Description
When determining memory based on a percentage of the total available, use the constrained memory value if present to avoid OOM in scenarios where the process has been restricted to less memory than the host physically has.

For example, if running tests in a Kubernetes pod, the container resource requets/limits are used to configure the cgroupsv2 namespace. Exceeding this will result in being OOM killed.

Vitest might see that it's scheduled on a node with 128GiB of RAM even though the container itself has been limited to 4GiB.

It will then allow each worker to have 1/N * 128Gi instead of 1/N * 4Gi of memory.
This can quickly result in an OOM situation, requiring manually setting the `vmMemoryLimit` and `maxWorkers` in the conf.

   - ⬆️  Adding tests is unfortunately not straightforward here, requires mocking `process` stuff. I'm happy to give it a shot if given guidance on preferred approach from a maintainer.

### Tests

### Documentation
  - ⬆️ I don't think this warrants a change in the current docs, but I'm happy to clarify that it _does_ respect cgroupsv2 now if desired.
 
### Changesets
  - ⬆️ I marked this as `feat` but feel free to retitle as `fix` if desired!

### DIFF
--- a/packages/vitest/src/node/pool.ts
+++ b/packages/vitest/src/node/pool.ts
@@ -6,6 +6,7 @@ import type { TestProject } from './project'
 import type { TestSpecification } from './test-specification'
 import type { BuiltinPool, ResolvedConfig } from './types/config'
 import * as nodeos from 'node:os'
+import process from 'node:process'
 import { isatty } from 'node:tty'
 import { resolve } from 'pathe'
 import { version as viteVersion } from 'vite'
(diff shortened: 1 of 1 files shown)


AriPerkkio (approved):
Looks good to me. @hi-ogawa are you familiar with `process.constrainedMemory`?

hi-ogawa (approved):
Didn't know the API but it makes sense to me. Just skimmed doc https://nodejs.org/api/process.html#processconstrainedmemory

Issue #9800, opened 2026-03-05
"Both esbuild and oxc options were set" on Vite 8.0.0-beta.16 + Vitest 4.1.0-beta.6


wojtekmaj:
### Describe the bug

When running unit tests, I'm seeing a warning:

```
Both esbuild and oxc options were set. oxc options will be used and esbuild options will be ignored. The following esbuild options were set: `{ legalComments: 'inline' }`
```

The thing is, neither in my `vitest.config.ts` nor `vite.config.ts` I have `esbuild.legalComments` option set. I only have `build.rolldownOptions.output.legalComments` set.

I believe this comes from the following default:

https://github.com/vitest-dev/vitest/blob/0c2c01361a95dd26d0d7fd7bc38bcca8dbc6e5d2/packages/browser/src/node/plugin.ts#L358-L361

but following intuition and setting `esbuild: false` in my Vitest config, I just get a different warning:

```
`esbuild` option is set to false, but `oxc` option was not set to false. `esbuild: false` does not have effect any more. If you want to disable the default transformation, which is now handled by Oxc, please set `oxc: false` instead.
```

### Reproduction

I believe I have provided sufficient information for this bug to be resolved. If I'm mistaken, please let me know!

### System Info

```shell
System:
    OS: macOS 26.3
    CPU: (10) arm64 Apple M1 Max
    Memory: 834.45 MB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.11.1 - /usr/local/bin/node
    Yarn: 4.10.3 - /Users/wojciech.maj/.yarn/switch/bin/yarn
    npm: 11.6.2 - /usr/local/bin/npm
  Browsers:
    Chrome: 144.0.7559.134
    Firefox: 141.0.3
    Safari: 26.3
```

### Used Package Manager

npm

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vitest-dev/vitest/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vitest.dev/guide/).
- [x] Check that there isn't [already an issue](https://github.com/vitest-dev/vitest/issues) that reports the same bug to avoid creating a duplicate.
[...]

PR #10725, opened 2026-07-07
fix: close the pool before the Vite servers, stabilize flaky tests

sheremet-va:
Standalone stability fixes that previously rode along in #10685, extracted so they can land independently.

## `Vitest.close()` closes the pool before the Vite servers

Closing a Vite server releases its port while automated browser pages may still be alive: a page's websocket client would auto-reconnect onto the next server that binds the same port and fail with `Unknown session id` (the reconnect race behind #9800). The pool (and the browser pages with it) is now closed first, and only then the per-project Vite servers.

## The force-exit watchdog is kept disarmable

`Vitest.exit()` arms an unref'd watchdog that force-exits the process after `teardownTimeout` if teardown hangs. In-process tests that stub `process.exit` to survive `exit()` (the `q` / double-CTRL+C shortcuts) left that watchdog armed, and it killed the shared e2e test worker ~10s later, in the middle of an unrelated test file. The timeout handle is now stored on the instance (`_exitTimeout`) so tests can disarm it; the watchdog itself still fires in real runs.

## Browser test stability

- `processTimeoutOptions` keeps a 250ms buffer (was 100ms) so the provider's descriptive, source-mapped locator error wins the race against the generic task timeout on loaded CI machines running several browsers.
- `commands.test.ts` uses a per-browser file name — all browser instances run the file against the same cwd in parallel.
- `wheel.test.ts` polls for the wheel event, which the browser dispatches asynchronously.
- `cancel-run.test.ts` polls for the `process.exit` call instead of racing the teardown.

### DIFF
--- a/packages/browser/src/client/tester/tester-utils.ts
+++ b/packages/browser/src/client/tester/tester-utils.ts
@@ -247,8 +247,10 @@ export function processTimeoutOptions<T extends { timeout?: number }>(options_: 
   const remainingTime = Math.floor(endTime - currentTime)
   // keep some buffer to process the timeout, but always hand the provider a
   // positive value so it surfaces a descriptive, source-mapped locator error
-  // instead of letting the task timer win the race with a generic timeout
-  options_.timeout = Math.max(remainingTime - 100, 1)
+  // instead of letting the task timer win the race with a generic timeout;
+  // the buffer covers the provider->server->client round-trip of the rejection,
+  // which can exceed 100ms on loaded CI machines running several browsers
+  options_.timeout = Math.max(remainingTime - 250, 1)
   return options_
 }
 
--- a/packages/vitest/src/node/core.ts
+++ b/packages/vitest/src/node/core.ts
@@ -1519,8 +1519,7 @@ export class Vitest {
         // close the pool (and the browser pages with it) BEFORE the Vite
         // servers: closing a server releases its port while automated pages may
         // still be alive — a page's websocket client would auto-reconnect onto
-        // the next server that binds the same port and fail with
-        // "Unknown session id" (#9800)
+        // the next server that binds the same port and fail with "Unknown session id"
         if (this.pool) {
           try {
             await this.pool.close?.()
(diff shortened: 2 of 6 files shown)


packages/browser/src/client/tester/tester-utils.ts
```diff
@@ -247,8 +247,10 @@ export function processTimeoutOptions<T extends { timeout?: number }>(options_:
   const remainingTime = Math.floor(endTime - currentTime)
   // keep some buffer to process the timeout, but always hand the provider a
   // positive value so it surfaces a descriptive, source-mapped locator error
-  // instead of letting the task timer win the race with a generic timeout
-  options_.timeout = Math.max(remainingTime - 100, 1)
+  // instead of letting the task timer win the race with a generic timeout;
+  // the buffer covers the provider->server->client round-trip of the rejection,
+  // which can exceed 100ms on loaded CI machines running several browsers
+  options_.timeout = Math.max(remainingTime - 250, 1)
```
hi-ogawa, line 253: I was also thinking to suggest bumping this for timeout rework https://github.com/vitest-dev/vitest/pull/10667#discussion_r3503052370. Apparently playwright does use 250ms for capping, so should be good.

packages/vitest/src/node/core.ts
```diff
@@ -1515,21 +1516,29 @@ export class Vitest {
           })
         }
 
+        // close the pool (and the browser pages with it) BEFORE the Vite
+        // servers: closing a server releases its port while automated pages may
+        // still be alive — a page's websocket client would auto-reconnect onto
+        // the next server that binds the same port and fail with
+        // "Unknown session id" (#9800)
```
hi-ogawa: The reference doesn't seem right.
- https://github.com/vitest-dev/vitest/issues/9800

hi-ogawa (commented):
lgtm!

packages/vitest/src/node/core.ts
sheremet-va: not sure where it got it from, removed the reference

PR #10726, opened 2026-07-07
perf(browser): open browser sessions adaptively instead of maxWorkers upfront

sheremet-va:
Opens browser sessions adaptively instead of `min(maxWorkers, files.length)` tabs upfront.

Sessions open one at a time while the queue justifies another tab: the projected queue drain time (queue length × measured per-file cost, split across the open sessions) has to exceed twice the measured session-open cost. Every tab pays a context + page + full module-graph bring-up that competes with the already-running sessions for the same Vite server, so fast suites finish sooner with fewer tabs, while suites with slower files still scale up to `maxWorkers` (without a per-file signal the pool keeps opening sessions like before). The per-file cost is an EMA that ignores each session's first file, which pays the tester bootstrap on top of the test.

The dispatch also warms the test file's transform (`transformRequest`) while the iframe is booting, so the file import doesn't wait for the transform on the critical path.

### DIFF
--- a/packages/vitest/src/node/pools/browser.ts
+++ b/packages/vitest/src/node/pools/browser.ts
@@ -187,6 +187,16 @@ class BrowserPool {
   private _promise: DeferPromise<void> | undefined
   private _providedContext: string | undefined
 
+  private _scaling: Promise<void> | undefined
+  // EMA of how long a single file takes in this pool; undefined until the
+  // first file finishes, which means "no signal yet — keep opening sessions"
+  private _fileCostEma: number | undefined
+  // refined with the real duration after every session open
+  private _sessionOpenCost = 250
+  // a session's first file pays the tester bootstrap on top of the test
+  // itself, so it would wildly overestimate the steady per-file cost
+  private _warmedUpSessions = new Set<string>()
+
   private readySessions: Set<string>
 
   private _traces: Traces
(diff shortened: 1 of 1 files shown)


packages/vitest/src/node/pools/browser.ts
```diff
@@ -373,7 +441,14 @@ class BrowserPool {
     const orchestrator = this.getOrchestrator(sessionId)
     debug?.('[%s] run test %s', sessionId, file)
 
+    // warm the transform cache while the iframe is booting so the test
+    // file import doesn't wait for the transform; mirrors the URL the
+    // tester will request (see `importFile` in the browser runner)
+    const fileUrl = `/${/^\w:/.test(file.filepath) ? '@fs/' : ''}${file.filepath}`.replace(/\/+/g, '/')
+    void this.project.browser?.vite.transformRequest(fileUrl).catch(() => {})
```
hi-ogawa: Should this be defined or optional? If we can semi-prove, I prefer dropping the optional defense.

hi-ogawa (dismissed):
Nice! It all makes sense once I see it. I had to ask AI again for visualization to capture the intuition https://artifacts.hiro18181.workers.dev/vitest-pr-10726-adaptive-sessions

One thing I noticed is that this scaling idea looks like universal work scheduling optimization and this can technically apply to the cases like reusing pools for slow windows forks for example.

packages/vitest/src/node/pools/browser.ts
sheremet-va: After config refactor we can actually just do `this.project.vite.transformRequest`

sheremet-va: Pushed the change

sheremet-va:
> One thing I noticed is that this scaling idea looks like universal work scheduling optimization and this can technically apply to the cases like reusing pools for slow windows forks for example.

I'll explore it, but beware that it will be one more PR to review 😄

hi-ogawa (approved):

PR #10727, opened 2026-07-07
perf(browser): prewarm the browser while the Vite server starts

sheremet-va:
Starts preparing the browser while the Vite server is still being created, so the launch latency overlaps server startup.

Browser providers gain an optional `prewarm({ config, vitest })` hook, invoked in `createClusterServer` right before the browser Vite server is created (the `TestProject` doesn't exist yet at that point, so the hook receives the resolved config). The playwright provider uses it to start `import('playwright')` and the browser launch concurrently with server startup.

The speculative launch is always safe: the launch options are resolved by the same code as the real launch and compared at adoption time — on any mismatch the prewarmed browser is discarded and the normal launch path runs. A prewarmed browser that is never adopted is closed with the provider. Prewarming is skipped for `connectOptions`, `persistentContext`, and inspector flows. Every browser name resolved from `browser.name` and `browser.instances` is warmed.


packages/browser-playwright/src/playwright.ts
```diff
@@ -96,12 +96,133 @@ export function playwright(options: PlaywrightProviderOptions = {}): BrowserProv
...
+      const browser = ctx.config.browser
+      const names = new Set<string>(browser.name ? [browser.name] : [])
+      for (const instance of browser.instances ?? []) {
+        if (instance.browser) {
+          names.add(instance.browser)
+        }
+      }
+      for (const name of names) {
+        prewarmBrowser(ctx, options, name)
+      }
```
hi-ogawa: Adversarial reviews caught that this is warming filtered out browser instances too. For example, running the fixture with `--project=chromium` spawns all three browsers

```sh
$ pnpm -C test/browser test-fixtures basic.test.ts  --project=chromium --browser.headless=false
```

```sh
$ ps -axo pid=,comm= | perl -ne 'print "$1 $2\n" if /^\s*(\d+).*\/(Google Chrome for Testing|firefox|Playwright)$/'
60868 Google Chrome for Testing
60869 firefox
60875 Playwright
```

This isn't the case on main. Maybe something mixed up after config rework.

hi-ogawa (changes requested):

packages/vitest/src/node/plugins/browserLoader.ts
```diff
@@ -125,6 +126,20 @@ export async function createClusterServer(
...
 
+  // let the provider start preparing the browser (e.g. launching it)
+  // while the vite server is being created; like the loader plugin above,
+  // the provider can also be configured per instance
+  const provider = config.browser.provider
+    ?? config.browser.instances?.find(instance => instance.provider)?.provider
+  if (provider?.prewarm) {
+    // instances filtered out by `--project` never become projects, so no
+    // provider would ever adopt (or close) a browser prepared for them
+    const instances = filterProjectBrowserInstances(vitest.config.project, config)
```
hi-ogawa: Probably nit. Instead of extracting and reusing `filterProjectBrowserInstances`, would it also possible to stash already available filtered result somewhere and grab it back here? I think config resolution (and project filtering) is already done till reaching here.

packages/browser-playwright/src/playwright.ts
```diff
@@ -146,18 +139,27 @@ function prewarmBrowser(project: LaunchContext, options: PlaywrightProviderOptio
...
   if (byName.has(browserName)) {
     return
   }
-  const launchOptions = resolveLaunchOptions(project, options, browserName)
+  // `headless` is the only launch-relevant option an instance can override;
+  // this mirrors how the instance's project config is cloned from the parent
+  const browserConfig = instance.headless == null
+    ? project.config.browser
+    : { ...project.config.browser, headless: instance.headless }
+  const launchOptions = resolveLaunchOptions(browserConfig, project.vitest.config.inspector, options, browserName)
```
hi-ogawa: This looks similar point and I assume ground truth config resolution is available at the point of `createClusterServer` doing prewarm instead of duplicating instance inheritance behavior.

Slop alert 🤖 but if I nudge agent, it said something like this:

<details><summary>gpt says</summary>

A lower-repetition design would carry the already cloned instance configs from expansion into server creation.

**1. Add Handoff Metadata**

At `packages/vitest/src/node/types/config.ts:1270`:

```ts
export interface ResolvedProjectEntry {
  viteConfig: ResolvedViteConfig
  projectConfig: ResolvedConfig
  hidden?: boolean

  /** Resolved browser-instance configs owned by this hidden parent. */
  browserProjects?: ResolvedConfig[]
}
```

**2. Capture Resolved Configs**

At `packages/vitest/src/node/projects/resolveProjects.ts:458`, filtering already determines the selected instances.

Before adding the hidden parent at line 468:

```ts
const browserProjects: ResolvedConfig[] = []

result.push({
  ...entry,
  hidden: true,
  browserProjects,
})
```

Inside the instance loop, the authoritative config is created at lines 553-554:

```ts
const clonedConfig
  = cloneProjectConfigForBrowserInstance(projectConfig, instance)
clonedConfig.name = name

browserProjects.push(clonedConfig)

result.push({
  viteConfig,
  projectConfig: clonedConfig,
})
```

The same resolved config object is therefore used both for project creation and prewarming.

**3. Pass Them Into Server Creation**

At `resolveProjects.ts:930`:

```ts
const {
  viteConfig,
  projectConfig,
  hidden,
  browserProjects,
} = entry
```

At line 958:

```ts
const { server, parent } = await createClusterServer(
  vitest,
  viteConfig,
  projectConfig,
  browserProjects,
)
```

**4. Prewarm Effective Providers**

At `packages/vitest/src/node/plugins/browserLoader.ts:111`:

```ts
export async function createClusterServer(
  vitest: Vitest,
  viteConfig: ResolvedViteConfig,
  config: ResolvedConfig,
  browserProjects: ResolvedConfig[] = [],
) {
```

Each resolved project contains its effective provider at `project.browser.provider`. Grouping by provider also fixes the per-instance-provider issue:

```ts
const projectsByProvider = Map.groupBy(
  browserProjects,
  project => project.browser.provider,
)

for (const [provider, projects] of projectsByProvider) {
  provider?.prewarm?.({ vitest, projects })
}
```

**5. Consume Resolved Browser Config**

Change the prewarm context at `packages/vitest/src/node/types/browser.ts:30`:

```ts
prewarm?: (ctx: {
  vitest: Vitest
  projects: ResolvedConfig[]
}) => void
```

Then Playwright at `packages/browser-playwright/src/playwright.ts:100` becomes:

```ts
prewarm(ctx) {
  for (const project of ctx.projects) {
    prewarmBrowser(ctx.vitest, options, project.browser)
  }
}
```

`project.browser` already contains:

- Effective browser name from `resolveProjects.ts:707`
- Effective `headless` from line 705
- Effective provider from line 706
- Inherited trace, UI, and parent browser settings

This removes the second filtering call and the manual reconstruction of `browserConfig` inside Playwright.

</detasil>

hi-ogawa (changes requested):

packages/vitest/src/node/plugins/browserLoader.ts
sheremet-va: Agree, this should be simplified

hi-ogawa (approved):
lgtm!

PR #10728, opened 2026-07-07
perf(browser): stop serving framework sourcemaps in headless runs

sheremet-va:
Stops serving sourcemaps that nothing consumes during headless runs, while keeping reported error stacks readable.

**Framework sourcemaps.** In headless runs (with the inspector off), Vitest no longer generates and inlines sourcemaps for its own pre-built modules — their stack frames are hidden from stack traces anyway. Serving them multiplied the bytes each browser context downloads by ~5 and burned server CPU regenerating maps for minified code on every run.

**`browser.dependencySourcemaps: false`** (also `--browser.dependencySourcemaps=false`) extends this to all of `node_modules` for projects that don't step into dependency code in devtools. It respects `server.sourcemapIgnoreList`, so workspace code that resolves to a `node_modules` path (e.g. with `resolve.preserveSymlinks`) can be opted back in. Sourcemaps of user source files are always served.

**Readable error stacks either way.** Error stack frames pointing into a pre-bundled dependency (or a module whose map was stripped above) are mapped using the sourcemaps stored on disk in the optimizer cache. Previously the transform-pipeline map resolved back into the optimizer cache, so these frames were reported at positions in the optimized chunk (e.g. `deps/react.js?v=…:468:41`) instead of the dependency's real sources. This works regardless of the `dependencySourcemaps` setting.

### DIFF
--- a/packages/browser/src/node/index.ts
+++ b/packages/browser/src/node/index.ts
@@ -9,7 +9,7 @@ import { toArray } from '@vitest/utils/helpers'
 import { join, resolve } from 'pathe'
 import sirv from 'sirv'
 import c from 'tinyrainbow'
-import { isFileServingAllowed, isValidApiRequest, rolldownVersion, distDir as vitestDist } from 'vitest/node'
+import { isCSSRequest, isFileServingAllowed, isValidApiRequest, rolldownVersion, distDir as vitestDist } from 'vitest/node'
 import { version } from '../../package.json'
 import { distRoot } from './constants'
 import { createOrchestratorMiddleware } from './middlewares/orchestratorMiddleware'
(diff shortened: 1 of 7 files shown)


packages/browser/src/node/index.ts
```diff
@@ -347,11 +347,68 @@ body {
...
+  }
+  // sibling instances share this server (and its module graph cache, so a
+  // late-spawned instance would receive already-cached transforms) and can
+  // override `headless` — check the static instance options instead of the
+  // lazily populated `children` to stay deterministic across runs
+  const instances = parentServer.config.browser.instances ?? []
+  return instances.every(instance => instance.headless !== false)
+}
+
+function cleanQueryUrl(url: string): string {
```
hi-ogawa: nit, there's same `cleanUrl` somewhere.

packages/browser/src/node/index.ts
```diff
@@ -347,11 +347,68 @@ body {
...
+        }
+        // users that never debug into node_modules can drop dependency
+        // sourcemaps entirely; `server.sourcemapIgnoreList` (default:
+        // node_modules) can opt paths back in even then, e.g. with
+        // `preserveSymlinks` where workspace code keeps its node_modules
+        // path and would be wrongly treated as a dependency
+        if (
+          parentServer.config.browser.dependencySourcemaps === false
+          && path.includes('/node_modules/')
+          && (parentServer.vite?.config.server.sourcemapIgnoreList(path, path) ?? true)
```
hi-ogawa: `parentServer.vite` isn't optional in type anymore.

hi-ogawa (commented):
lgtm.

hi-ogawa (changes requested):

hi-ogawa (approved):

PR #10729, opened 2026-07-07
perf(browser): serve framework assets as immutable

sheremet-va:
Serves Vitest's own pre-built modules with `Cache-Control: immutable` instead of having every tester iframe revalidate them.

Vitest's framework assets only change with the Vitest version and browser contexts don't outlive the run, so revalidating them per iframe is wasted work. They are now pinned as `public,max-age=31536000,immutable`. Skipped when `persistentContext` is enabled, since its disk cache would survive a Vitest upgrade. The new immutable policy and the existing optimized-deps `no-cache` policy (which must stay `no-cache` so per-test module mocking works) now live in a single middleware.


hi-ogawa (commented):
This looks good, but the browsers on preview provider would keep stale immutable cache forever.

packages/browser/src/node/index.ts
```diff
@@ -219,20 +232,28 @@ body {
         )
       }
 
+      // vitest's own dist files only change with the vitest version and
+      // browser contexts don't outlive the process, so there is no reason
+      // to revalidate them in every tester iframe. Skipped for persistent
+      // contexts — their disk cache would survive a vitest upgrade
+      const persistentContext = (parentServer.config.browser.provider?.options as { persistentContext?: unknown } | undefined)?.persistentContext
```
hi-ogawa, line 239: Perhaps another provider level API contract (could be internal) so provider side can say if it's a fresh context?

sheremet-va:
> This looks good, but the browsers on preview provider would keep stale immutable cache forever.

I don’t think that’s true? Vite serves optimized dependencies with the same mechanism in dev server. We manually disabled it because of mocking; this pr just removes the check for vitest since it can’t be mocked anyway.

hi-ogawa:
> > This looks good, but the browsers on preview provider would keep stale immutable cache forever.
> 
> I don’t think that’s true? Vite serves optimized dependencies with the same mechanism in dev server. We manually disabled it because of mocking; this pr just removes the check for vitest since it can’t be mocked anyway.

Oh wait, that's the story. That makes sense, but also confused as to why we are opting out persistentContext?

sheremet-va:
> Oh wait, that's the story. That makes sense, but also confused as to why we are opting out persistentContext?

Good point, I think AI hallucinated a little bit here. Optimized deps are served with a `?v=` query. Although I am not sure if we merged the part where vitest is now served optimized 🤔

hi-ogawa (changes requested):

sheremet-va:
Let's come back here after https://github.com/vitest-dev/vitest/pull/10713 since without it vitest would never be revalidated (it has no `v=` query)

sheremet-va:
@hi-ogawa actually this PR is correct. The immutability is applied only to static imports like `__vitest_browser__`, not to `vitest` imports, so there is no `?v=` attached.

Regarding the new contract - I am not sure. The `.options` API is already kind of internal and I don't see a reason to expose it differently? This is the only way to reuse context at the moment anyway.

packages/browser/src/node/index.ts
```diff
@@ -219,20 +232,28 @@ body {
...
+      const immutablePrefixes = persistentContext
```
macarie: Shouldn't this check if `fileParallelism` is true and running a headless browser since it's ignored in that case?

Also not sure I understand why this gets disabled when a persistent context is set 🤔

sheremet-va: contexts are normally destroyed after the browser is closed, so these urls are cached only during the test run (so different tests don’t request the same files). persistent context will keep the cache even after the browser is closed. this means that even after the user upgrades vitest to the new version, they will receive the old content

I don’t understand why file parallelism needs to be checked?

macarie: I see, thanks!

Regarding the parallelism, I see the Playwright provider checks for that (https://github.com/vitest-dev/vitest/blob/1ffafc0878db61c92ace581b11d926e4699ca079/packages/browser-playwright/src/playwright.ts#L336-L338), is the option already _normalized_ by the time it reaches this point?

sheremet-va: I don't think it can be normalized because it's dynamically decided when tests are actually running, not during the config phase.

I think it's better to just check for `persistentContext` to avoid false positives. Even if `persistentContext` is enabled and it's not in parallel, we can eat the performance penalty - in this case user also sees a warning anyway.

macarie (dismissed):

hi-ogawa (changes requested):
> actually this PR is correct. The immutability is applied only to static imports like `__vitest_browser__`, not to `vitest` imports, so there is no `?v=` attached.

I don't follow the logic. How does this explain we can immutable cache preview provider but opt-out for playwright persistent context?

packages/browser/src/node/index.ts
```diff
@@ -219,20 +232,28 @@ body {
...
+        ? []
+        : ['/__vitest_browser__/', `/@fs${vitestDist}`, `/@fs${distRoot}`]
```
hi-ogawa, line 242: Since we pre-bundle vitest, shouldn't target `/deps/vitest...` for immutable? From quick testing on this pr branch, pre-bundled vitest assets get `no-cache`. 

On the other hand, browser `dist/locators.js` gets immutable, so that's working good (but for the point https://github.com/vitest-dev/vitest/pull/10729/#pullrequestreview-4783791577, it would mean user browser with preview provider can get stale assets).

Image

Image

---

There's also an issue with `/@fs/` assumption. Vite resolves to `/@fs/<abs-path>` when the source is out of vite root (e.g. monorepo package testing other package inside monorepo), but if `node_modules` is available at root, then Vite resolves to `/node_modules/...` without `@fs` full path.

sheremet-va: Because this is about our assets, not vitest package. There was actually a separate PR for optimized dependencies, this PR has nothing to do with them. The vitestDist was probably just hallucinated.

The fs issue doesn’t apply here because it’s not vite that uses it, we manually build urls that start with fs/ when importing these assets

sheremet-va:
`preview` is a good point, I will add a guard for it too.

PR #10734, opened 2026-07-07
feat: promote fsModuleCache to a top-level option

### LINKED ISSUE #10701: Stabilize `fsModuleCache`
I propose to stabilize and enable by default in Vitest 5:
- [`experimental.fsModuleCache`](https://vitest.dev/config/experimental.html#experimental-fsmodulecache)
- [`experimental.fsModuleCachePath`](https://vitest.dev/config/experimental.html#experimental-fsmodulecachepath)

We haven't received any issues regarding this feature, and we have also been running tests with this flag enabled for a while now.

TBD:
- How does `fsModuleCachePath` interact with `cacheDir` and `cache.dir`

Related:
- https://github.com/vitest-dev/vitest/discussions/9221

sheremet-va:
Move `fsModuleCache` and `fsModuleCachePath` out of `experimental` (`fsModuleCache` defaults to off). The cache is a single workspace-root `node_modules/.vitest-cache` directory shared by every project. The old `experimental.*` options are migrated with a deprecation warning.

Fixes #10701


sheremet-va:
Should we enable it by default only in non-CI environments? There is usually a performance hit for the first (cold) run and in CI there is usually no caching anyway

docs/config/fsmodulecache.md
```diff
@@ -0,0 +1,88 @@
...
+```js [vitest.config.js]
+import { defineConfig } from 'vitest/config'
+
+export default defineConfig({
+  plugins: [
+    {
+      name: 'vitest-cache',
+      api: {
+        vitest: {
+          ignoreFsModuleCache: true,
```
hi-ogawa, line 77: I don't remember the use case of per plugin `ignoreFsModuleCache`. From what I can imagine, this only basically plugin's existence to not affect cache key, so it's like adding/removing the plugin conditionally for different test run:

```js
plugins: [
  process.env.SOME_CONDITION && {
    ...some logic...
    api: { vitest: { ignoreFsModuleCache: true }}
  }
]
```

Is this the right use case? The option name `ignoreFsModuleCache` and documentation here doesn't seem to give much intuition. Not mean to bike-shed now and no blocking though.

hi-ogawa (dismissed):
lgtm

docs/config/fsmodulecache.md
sheremet-va: Yes, kind of. It will be ignored when generating a cache key. This is mostly useful for library authors or when calling vitest programmatically (we even pass down this flag ourselves in a fee places). For config it doesn’t make much sense because config content (its file content - the string) affects the cache key.

This is for plugins that don’t affect transformation. If you move the order, it also won’t affect it for example.

hi-ogawa (approved):

PR #10742, opened 2026-07-08
perf: make the Node compile cache opt-in, persist worker caches on teardown

sheremet-va:
#10708 enabled Node's compile cache unconditionally. That turned out to be a bad default in two ways:

- On cold environments — CI that doesn't persist the cache directory — it makes runs slower: workers pay the code-cache production cost, and thread workers additionally serialize their whole compiled graph when they are terminated, with zero payoff since the directory is thrown away.
- For the default `forks` pool the cache was write-only anyway: SIGTERM'd workers never reach Node's exit-time flush, so worker entries (runtime, test environment, externalized deps) never reached the disk even when the directory did persist.

This PR makes the cache opt-in and makes opting in actually work:

- The executable no longer calls `module.enableCompileCache()`. Setting `NODE_COMPILE_CACHE=<dir>` enables the cache for the CLI (Node reads it at startup) and Vitest propagates it to every worker.
- Workers persist their compiled modules during graceful teardown (`module.flushCompileCache()` in the `stop` handler, both teardown paths; default import because a named import fails to link before Node 22.10). This is what makes the cache useful for `forks`.
- The `v8` coverage provider still strips the cache from workers (V8 serializes cached scripts without the source positions precise coverage relies on) — now covered by a test.
- Documented in `docs/guide/improving-performance.md` as a tuning knob for environments where the directory survives between runs.

40-file happy-dom `isolate: true` suite, median of 5 interleaved runs (Apple M4 ×10, Node 24):

| pool | scenario | main (always on) | this PR |
|---|---|---:|---:|
| forks | cold, cache not persisted (CI) | 2123 ms | **1924 ms (−9%)** |
| forks | `NODE_COMPILE_CACHE`, first run | — | 2366 ms |
| forks | `NODE_COMPILE_CACHE`, subsequent runs | — | 1812 ms |
| threads | cold, cache not persisted (CI) | 2115 ms | **1709 ms (−19%)** |
| threads | `NODE_COMPILE_CACHE`, first run | — | 2123 ms |
| threads | `NODE_COMPILE_CACHE`, subsequent runs | — | 1618 ms |

Cold runs get faster by default. Opting in pays the first-run serialization once and runs ~23% faster than that first run afterwards; the benefit over not enabling the cache at all grows with the size of the natively-imported graph (heavier environments and dependencies).

The e2e test covers env propagation into workers, `NODE_DISABLE_COMPILE_CACHE` precedence, and the coverage exclusion.

### DIFF
--- a/packages/vitest/src/runtime/workers/init.ts
+++ b/packages/vitest/src/runtime/workers/init.ts
@@ -2,6 +2,9 @@ import type { WorkerRequest, WorkerResponse } from '../../node/pools/types'
 import type { MetaEnv, WorkerSetupContext } from '../../types/worker'
 import type { FileSpecification } from '../runner/types'
 import type { VitestWorker } from './types'
+// default import: `flushCompileCache` only exists since Node 22.10, a named
+// import would fail to link on older versions
+import Module from 'node:module'
 import { serializeError } from '@vitest/utils/error'
 import { disableDefaultColors } from 'tinyrainbow'
 import { Traces } from '../../utils/traces'
@@ -241,6 +244,20 @@ export function init(worker: Options): void {
       case 'stop': {
         await runPromise
 
+        // Persist this worker's compile cache before the parent tears the
+        // worker down — forks are SIGTERM'd and never reach Node's exit-time
+        // flush, so without this the cache stays write-only for them. Runs
+        // even when teardown throws (the compiled modules are still worth
+        // persisting). A no-op when the cache is disabled or was fully loaded
+        // from disk, and cheap (~tens of ms) otherwise, so every worker can
+        // afford it.
+        const persistCompileCache = () => {
+          try {
+            Module.flushCompileCache?.()
+          }
+          catch {}
+        }
+
         try {
           const context = traces.getContextFromCarrier(message.otelCarrier)
 
@@ -256,9 +273,13 @@ export function init(worker: Options): void {
 
           await traces.finish()
 
+          persistCompileCache()
+
           send({ type: 'stopped', error, __vitest_worker_response__ })
         }
         catch (error) {
+          persistCompileCache()
+
           send({ type: 'stopped', error: serializeError(error), __vitest_worker_response__ })
         }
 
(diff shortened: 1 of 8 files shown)


sheremet-va:
@AriPerkkio I should've tested more without trusting AI, you were right to question what are the actual numbers 😄 I don't think the cold start cost justifies the warm improvement here since CI will always run cold

packages/vitest/src/runtime/workers/init.ts
```diff
@@ -241,6 +244,20 @@ export function init(worker: Options): void {
...
         await runPromise
 
+        // Persist this worker's compile cache before the parent tears the
+        // worker down — forks are SIGTERM'd and never reach Node's exit-time
+        // flush, so without this the cache stays write-only for them. Runs
+        // even when teardown throws (the compiled modules are still worth
+        // persisting). A no-op when the cache is disabled or was fully loaded
+        // from disk, and cheap (~tens of ms) otherwise, so every worker can
+        // afford it.
+        const persistCompileCache = () => {
```
sheremet-va, line 254: since we kill the child, forks never actually flush the cache otherwise

hi-ogawa (approved):
makes sense!

packages/vitest/src/runtime/workers/init.ts
AriPerkkio: There's actually small time window where the worker can gracefully exit, before main thread kills it. But anyways, flushing the cache manually like this is good approach.

sheremet-va: threads workers do flush gracefully, but we for forks we kill them and they don't have enough time (or even knowledge?) to flush. in my tests they never flushed the cache

AriPerkkio: Removing these two lines shows that forks worker does exit gracefully without the need of main thread killing it:

https://github.com/vitest-dev/vitest/blob/83ab9a44f23d98ce4e5b4dbb0214ee47cab0d433/packages/vitest/src/node/pools/workers/forksWorker.ts#L85

https://github.com/vitest-dev/vitest/blob/83ab9a44f23d98ce4e5b4dbb0214ee47cab0d433/packages/vitest/src/node/pools/workers/forksWorker.ts#L80-L83

```sh
cd test/unit
pnpm run test test/math.test.ts --project forks --run
```

Maybe enabling compile cache prevents worker from exiting. 🤔

PR #10743, opened 2026-07-08
fix: keep per-file isolation in vm pools when maxWorkers is 1

sheremet-va:
### Description

vm pools resolve `isolate` to `false` during config resolution — their isolation comes from a fresh VM context per run request, not from worker lifecycle. The single-worker batching in `groupSpecs` treated that resolved flag as a user opt-in: with `maxWorkers: 1`, all same-environment files were packed into one run request, and `runVmTests` executes a whole run request in a single VM context. Every test file then shared one context — `globalThis`, DOM and module state leaked across files:

```ts
// a.test.ts
test('pollutes', () => {
  globalThis.__leaked = 'from a'
})

// b.test.ts — fails with vmThreads + maxWorkers: 1, passes with maxWorkers >= 2
test('isolated', () => {
  expect(globalThis.__leaked).toBeUndefined()
})
```

`vmThreads`/`vmForks` are now excluded from that batching and fall back to one run request per file on the shared worker, restoring the same per-file context isolation as any other `maxWorkers` value. No opt-in is lost: explicit `isolate` settings are already overwritten for vm pools, so `isolate: false` there never expressed user intent to share state.

### DIFF
--- a/packages/vitest/src/node/pool.ts
+++ b/packages/vitest/src/node/pool.ts
@@ -431,8 +431,12 @@ function groupSpecs(specs: TestSpecification[], environments: WeakMap<TestSpecif
       throw new Error(`Projects "${last}" and "${spec.project.name}" have different 'maxWorkers' but same 'sequence.groupOrder'.\nProvide unique 'sequence.groupOrder' for them.`)
     }
 
-    // Non-isolated single worker can receive all files at once
-    if (isolate === false && maxWorkers === 1) {
+    // Non-isolated single worker can receive all files at once.
+    // vm pools are excluded: their `isolate: false` comes from config
+    // resolution rather than the user, because their isolation is a fresh VM
+    // context per run request — batching files into a single run request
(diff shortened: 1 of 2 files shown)


AriPerkkio (approved):
Proper fix for this should be that vm pools themselves handle vm context recycling. They should be able to receive all specs from main thread in a single batch.

This PR seems to add work-around.

hi-ogawa:
Oh well, I was just about to report the bug and realized it's fixed already 😄

AriPerkkio:
> Proper fix for this should be that vm pools themselves handle vm context recycling. They should be able to receive all specs from main thread in a single batch.

Actually this might not be possible due to how [`vmMemoryLimit`](https://vitest.dev/config/vmmemorylimit.html#vmmemorylimit) works. So the approach this PR introduced is probably the only way to do it simply. 

I wonder if `vmMemoryLimit` was actually broken due to this bug. 🙃

PR #10744, opened 2026-07-08
perf: reuse compiled code across vm pool contexts and prewarm the module graph

sheremet-va:
### Description

vm pools isolate test files with a fresh VM context per run request, which means every file re-compiles and re-evaluates its entire module graph — including externalized `node_modules` and the Vitest runtime itself. Profiling jsdom + `vmThreads` runs of fixtures with real dependencies (lodash, lodash-es, date-fns, moment, rxjs, zod, react, react-dom/server) showed three costs this PR removes:

1. **Re-compilation per context.** Compiled code has no per-context state — only its evaluation does. This PR shares it per worker:
   - inlined modules: compiled `vm.Script` instances are shared across contexts (safe: Vite rewrites dynamic imports to `__vite_ssr_dynamic_import__`, so no per-context import callback is baked into the script);
   - externalized CJS: per-worker V8 `cachedData`, produced after the first execution so it carries the compiled module body (the `vm.Script` itself stays per-context because it embeds the per-context `importModuleDynamically` closure);
   - externalized ESM: `SourceTextModule` `cachedData`, produced before evaluation as V8 requires.

   Entries are keyed by module id, and `mock:` ids never collide with their originals.

2. **An idle server during worker startup.** On a cold start every vm worker spends the first ~0.5s of its life importing the environment package (jsdom), while the server has nothing to do; the file's transforms only start after that, serially. vm workers now fire a `prewarmModuleGraph` RPC before `environment.setupVM`: the server BFS-transforms the file's import graph through the same fetcher as live fetches during that window. Since the prewarmed transforms set `__vitestTmp`, the worker's `fetchWarmModules` snapshot then covers the graph even on the first cold run, and its fetches short-circuit to local disk reads. Prewarm failures are ignored — the worker's own fetch reports errors with the proper import context.

3. **Re-resolution of every import edge, per context.** `ExternalModulesExecutor.resolve` calls `import.meta.resolve` for every ESM import edge of every externalized module, and Node re-derives the resolved module's package scope on each uncached call — re-parsing large `exports` maps along the way (date-fns's package.json is 199KB with 741 export entries). Node's own resolution cache is keyed by `(specifier, parentURL)`, so each worker still pays the full external graph once (~3300 calls ≈ 700ms of a 9-worker profile on the deps fixture, ×9 workers). Two changes:
   - relative specifiers (restricted to a conservative charset — anything URL-special falls back to Node) resolve by plain URL join. This is exactly what `import.meta.resolve` returns for them: relative ESM specifiers never consult package.json, and Node's resolver doesn't check existence either — the executor already recreates `ERR_MODULE_NOT_FOUND` itself;
   - bare specifiers get a worker-wide resolution cache in front of `import.meta.resolve` (behind the custom vite/mock resolvers, which are unaffected), and module type/existence detection is memoized per worker. Both follow the same worker-lifetime staleness invariants as the existing `fileMap`/`packageCache`.

Isolation semantics are unchanged: evaluation still happens once per context, verified by an e2e test that fails if module state leaks through the shared scripts.

Benchmarks (jsdom, `vmThreads`, `isolate: true`, fs module cache on; cold = all caches wiped before every rep, warm = populated caches; median of 3 interleaved reps in a single window, M4/Node 24.13, whole-process wall clock; trivial assertions):

| fixture | cold | warm |
|---|---|---|
| 342 modules + real deps (lodash, lodash-es, date-fns, react, react-dom/server), 40 test files | 2605ms → 1805ms (**−31%**) | 2556ms → 1710ms (**−33%**) |
| 1022 modules + heavier deps (adds moment, rxjs, zod), 160 test files | 7078ms → 5127ms (**−28%**) | 7010ms → 5149ms (**−27%**) |
| 176 modules, no npm deps, 40 test files | 1210ms → 1135ms (**−6%**) | 1200ms → 1136ms (**−5%**) |

`vmForks` tracks `vmThreads` (1022-module fixture: 6528ms → 5296ms cold, **−19%** against the resolution change alone). The split matches the mechanisms: fixtures with externalized dependencies win from the resolution fast path and compile reuse (the costs scale with the external graph), while the no-deps fixture only collects the transform/jsdom-import overlap.

### Transform time reporting

The prewarm broke the reported transform time, and investigating that exposed that the stat was already unreliable: it summed each fetch's wall time, but concurrent fetches (parallel workers, and now the prewarm) all wait on the same deduplicated in-flight transforms, so the sum counts the same work many times. Measured on the deps fixture (cold `vmThreads` run, ~2.5s wall; serial `--maxWorkers=1` reference: 356ms of real transform work):

| strategy | reported "transform" |
|---|---|
| summed fetch walls, with prewarm | 348–495s |
| summed walls deduplicated to the first in-flight caller | 61–82s (a wide BFS still measures queue position, not work) |
| summed fetch walls on `main` today, no prewarm, default workers | 2.16s on a 2.52s run; warm runs report fs-cache *reads* as 2.4s of "transform" |
| **this PR: pipeline busy time** | **1.11s cold, 0ms warm** |

`transformTime` is now the union of in-flight transform intervals, measured inside the fetcher around the actual transform (`fetchAndProcess`) — cache hits and fs-cache reads don't tick the clock. It has a crisp meaning ("wall time during which the server was transforming"), is bounded by the run's duration, and behaves identically with and without the prewarm. Warm runs with a populated `fsModuleCache` now truthfully report ~0 transform time. Per-module durations (`metadata.duration`, used by the UI module graph) are still recorded from direct worker fetches only — the prewarm's per-module walls would measure BFS queue position rather than the module's own cost.


sheremet-va:
The same prewarm trick doesn't work with treads/forks because we need to reevaluate the environment for every worker. Here we spawn the workers -> import jsdom in parallel in X workers once -> prewarm every test file

sheremet-va:
The transform changes affects how it's shown in https://github.com/vitest-dev/vitest/pull/10710

packages/vitest/src/node/environments/fetchModule.ts
```diff
@@ -18,6 +18,18 @@ import { normalizeResolvedIdToUrl } from './normalizeUrl'
...
 const readFilePromises = new Map<string, Promise<string | null>>()
 
+/**
+ * Tracks the wall time during which at least one transform is running.
+ * Durations of individual fetches cannot be summed instead: concurrent
+ * fetches (parallel workers, the vm pool graph prewarm) all wait on the same
+ * deduplicated in-flight transforms, so per-caller wall times overcount the
+ * actual work by orders of magnitude.
+ */
+export interface TransformClock {
```
sheremet-va, line 28: in short, we calculate the REAL time it takes to transform. instead of summing individual times

packages/vitest/src/runtime/moduleRunner/moduleEvaluator.ts
```diff
@@ -25,6 +25,28 @@ import { ModuleDebug } from './moduleDebug'
...
 const isWindows = process.platform === 'win32'
 
+// Compiled scripts of inlined modules, shared across vm contexts: vm pools
+// evaluate every module again in each fresh context, but the compiled script
+// holds no per-context state (Vite rewrites dynamic imports to
+// `__vite_ssr_dynamic_import__`, so no per-context import callback is baked
+// in) — only its evaluation has to happen per context. Keyed by module id
+// (`mock:` ids stay distinct from their originals) and guarded by the exact
+// code, so invalidated modules replace their entry.
+const vmInlineScriptCache = new Map<string, { code: string; script: vm.Script }>()
```
AriPerkkio: Does this increase memory usage a lot? It's kind of introducing memory leak for every script that was evaluated.

AriPerkkio (commented):
Does the new transform time measuring also remove time that workers idle when waiting for server to handle their request? Transform time is now measured on main thread side, right?

> the server BFS-transforms the file's import graph

What's BFS?

sheremet-va:
> BFS

Breadth-first search, as apposed to DFS (depth-first search). Do they not ask you this in the interviews :O

packages/vitest/src/runtime/moduleRunner/moduleEvaluator.ts
sheremet-va: Good question, I will validate before merging. _I think_ that it's fine since tests usually import the same files anyway. At worst, this holds every source file (which is already just one test file that imports a barrel file)

These are also just strings, not actual export values

AriPerkkio:
> Breadth-first search, as apposed to DFS (depth-first search). Do they not ask you this in the interviews :O

Nope, never heard of these terms. I don't think I've ever run into this term in my ~10 year career, or in the engineering studies before that. Happy to ask all the stupid questions out loud. 😇

packages/vitest/src/runtime/moduleRunner/moduleEvaluator.ts
AriPerkkio: Yep, caching just the source strings shouldn't be as bad as the og `vm.Script` memory leaks of executed context. But good to check it before merge.

sheremet-va:
> > Breadth-first search, as apposed to DFS (depth-first search). Do they not ask you this in the interviews :O
> 
> Nope, never heard of these terms. I don't think I've ever run into this term in my ~10 year career, or in the engineering studies before that. Happy to ask all the stupid questions out loud. 😇

These algorithms have been following me for years now 😢

hi-ogawa:
Haven't read but my review agent said more PRs to split.

> The PR could reasonably be split into:
> 
> 1. Reuse compiled inlined and externalized VM code. These could be one PR or two because they use different cache mechanisms.
> 2. Cache external-module resolution and module information.
> 3. Prewarm VM module graphs, including warm snapshots, the RPC contract, error-path coverage, and transform-time accounting.

sheremet-va:
> Haven't read but my review agent said more PRs to split.
> 
> > The PR could reasonably be split into:
> > 
> > 1. Reuse compiled inlined and externalized VM code. These could be one PR or two because they use different cache mechanisms.
> > 2. Cache external-module resolution and module information.
> > 3. Prewarm VM module graphs, including warm snapshots, the RPC contract, error-path coverage, and transform-time accounting.

I'd argue the PR is small enough to review and merge.

packages/vitest/src/runtime/vm/commonjs-executor.ts
```diff
@@ -109,10 +113,16 @@ export class CommonjsExecutor {
...
       _compile(code: string, filename: string) {
         const cjsModule = Module.wrap(code)
+        const codeCache = executor.codeCache
+        const cachedData = codeCache?.get(filename, cjsModule)
         const script = new vm.Script(cjsModule, {
           filename,
+          cachedData,
           importModuleDynamically: options.importModuleDynamically,
         } as any)
+        if (cachedData && (script as any).cachedDataRejected) {
```
hi-ogawa: The type seems available

```suggestion
        if (cachedData && script.cachedDataRejected) {
```

hi-ogawa (commented):
Vm resolve/script/source caching looks good!
Now following on prewarm story.

packages/vitest/src/runtime/vm/commonjs-executor.ts
sheremet-va: oh yeah, AI loves this

packages/vitest/src/runtime/moduleRunner/moduleEvaluator.ts
hi-ogawa: One thing I'd add is that, these script/code per file are retained for each worker (thread/fork), so if the same module is evaluated in different worker, then these in-memory cache is also duplicated. So multiplicative factor also depends on `maxWorkers`. 

The analogy I'm imagining is that, basically our main process already have code-level module graph of entire test run. The cache retained in worker runtime side can potentially multiply that by `maxWorkers` depending on tests and imported modules distribution.

packages/vitest/src/node/state.ts
hi-ogawa: Was lost how the clock API is connected. Does doing this hurt? (this allows IDE to link `transformStarted/finished`)

```ts
import type { TransformClock } from './environments/fetchModule'

export class StateManager implements TransformClock {
```

hi-ogawa (changes requested):
The vm prewarm idea sounds good, but the reason on why vm only seems to assume heavy per vm environment setup. Should we also measure bench for non jsdom/happy-dom case i.e. default node environment? 

Labeling "request changes" on that ground and also vm worker side memory retain justification.

packages/vitest/src/runtime/moduleRunner/moduleEvaluator.ts
hi-ogawa: Counter point on my own comment though is that, retaining cache/modules in this way on worker runtime side is how forks/threads + no-isolate boosts perf, so the design feels sound.

sheremet-va:
> The vm prewarm idea sounds good, but the reason on why vm only seems to assume heavy per vm environment setup. Should we also measure bench for non jsdom/happy-dom case i.e. default node environment?
> 
> Labeling "request changes" on that ground and also vm worker side memory retain justification.

I did measure it for every combination. It makes sense to do only if we spawn non isolated workers that import jsdom once (which is the bottleneck - it takes 500ms per worker TO IMPORT. Setting up is cheap). Node doesn’t import anything and happy-dom could benefit from it, but it’s just 200ms.

In short, wamp up only makes sense in jsdom+vm combination and potentially happy-dom+vm.

The warmup only works because the main thread doesn’t do anything otherwise (it just waits for the jsdom to be imported in every worker in parallel).

packages/vitest/src/node/state.ts
sheremet-va: Should be fine

packages/vitest/src/runtime/moduleRunner/moduleEvaluator.ts
sheremet-va: I think we should be able to remove the `code` check since we destroy workers on change and it should never be hit, then we only keep scripts (which also holds the string, but it's half as much)

sheremet-va: Results from Sol (I also removed `{ code }` and just store by `vm.Script`)

`vmInlineScriptCache` does not substantially increase peak memory usage in the tested workload.

Compared directly with `main`, the complete PR used 7-12% more peak aggregate RSS with 4 or 8 workers, but 30-41% less with one worker. A controlled build with only `vmInlineScriptCache` removed showed that the cache itself added about 3.4 MiB in one configuration and reduced peak RSS in every other configuration. The multi-worker increase against `main` therefore comes from other differences in the PR, not `vmInlineScriptCache`.

## Complete PR Versus `main`

| Pool | Workers | `main` | PR | Difference |
| --- | ---: | ---: | ---: | ---: |
| `vmThreads` | 1 | 497 MiB | 293 MiB | -204 MiB (-41.0%) |
| `vmThreads` | 4 | 722 MiB | 773 MiB | +51 MiB (+7.1%) |
| `vmThreads` | 8 | 894 MiB | 988 MiB | +93 MiB (+10.4%) |
| `vmForks` | 1 | 575 MiB | 400 MiB | -175 MiB (-30.4%) |
| `vmForks` | 4 | 969 MiB | 1,081 MiB | +112 MiB (+11.5%) |
| `vmForks` | 8 | 1,446 MiB | 1,560 MiB | +114 MiB (+7.9%) |

The PR avoids repeated compilation churn, which explains the substantially lower single-worker peak. At higher worker counts, the complete PR retains more memory than `main`, but this comparison includes all changes on both diverged branches.

hi-ogawa:
> In short, wamp up only makes sense in jsdom+vm combination and potentially happy-dom+vm.

Do we settle on doing this only for jsdom/happy-dom or for everything on vm?

sheremet-va:
> > In short, wamp up only makes sense in jsdom+vm combination and potentially happy-dom+vm.
> 
> Do we settle on doing this only for jsdom/happy-dom or for everything on vm?

I don't see the harm of doing it everywhere even if it doesn't improve performance there.

hi-ogawa (approved):
If it's not negatively affecting perf, then that should be alright.

sheremet-va:
> If it's not negatively affecting perf, then that should be alright.

Confirmed with Fable:

No regression anywhere, and the answer to the node question is: keep the prewarm, unconditionally. The PR description's rationale (overlapping the jsdom import) indeed doesn't apply to node — but the overlap mechanism generalizes, and the measurement shows prewarm is net-positive on node whenever the suite has a real module graph.

Node with dependencies is actually the branch's biggest relative win.

Screenshot 2026-07-23 at 15 08 49

Why prewarm still pays on node: the overlap window isn't only the environment import — it's everything the worker does before and between inline fetches: resolving, compiling, and evaluating the externalized dependency graph per context, booting the runtime into the context, and (vmForks) spawning the process. The server BFS runs ahead of the worker's serial evaluation-driven fetch waterfall during that time. That's why vmForks shows the largest node-side prewarm win (−7.5%): slower worker boot, bigger window. The only cell where prewarm costs anything is node + tiny no-deps suite (+44ms cold), where workers start fetching immediately and 40 prewarm RPCs just contend with live fetches — and even that cell never drops below main, because the code cache absorbs it. Gating node off would trade 50–310ms of real-suite wins (appbig: 309ms cold) for a ≤44ms saving on suites that are already faster than main. This is the opposite of the base-pool prewarm null result from earlier — there the parent was the bottleneck; in vm pools the parent is idle during worker-side external work even on node.

PR #10746, opened 2026-07-09
fix(expect): handle nullish subject in toHaveProperty (#10735)

### LINKED ISSUE #10735: expect(null | undefined).toHaveProperty() throws an internal TypeError instead of a clean matcher error
### Description
When the received value is `null` or `undefined`, `toHaveProperty` crashes with a raw internal error rather than failing the assertion cleanly (or, for the negated form, passing). `.not.toHaveProperty` also crashes, when under vitest's chai-based semantics it should pass on a nullish subject.

### Steps to reproduce
```ts
expect(null).toHaveProperty('x')        // throws
expect(null).not.toHaveProperty('x')    // throws (should pass)
expect(undefined).toHaveProperty('x')   // throws
```
Observed: every call throws `TypeError: Cannot convert undefined or null to object`. Expected: a proper matcher error / assertion failure (jest emits a clean matcher error here).

### Root cause
`packages/expect/src/jest-expect.ts` -- the `getValue()` helper calls `Object.hasOwn(actual, propertyName)` unconditionally, before the null-safe `utils.getPathInfo(actual, propertyName)` fallback. `Object.hasOwn(null, ...)` throws; `actual` is never guarded for nullish.

### Suggested fix
Guard the subject before the own-property check: `const hasOwn = actual != null && Object.hasOwn(actual, propertyName)`, which lets the null-safe `getPathInfo` fallback run (fails cleanly, `.not` passes) — or throw a jest-style matcher error when `actual == null`.

xianjianlf2:
### Description

`expect(...).toHaveProperty()` crashes with a raw internal `TypeError: Cannot convert undefined or null to object` when the received value is `null` or `undefined`, instead of failing the assertion cleanly. The negated form `.not.toHaveProperty()` crashes too, when it should pass for a nullish subject.

```ts
expect(null).toHaveProperty('x')        // throws TypeError (should fail cleanly)
expect(null).not.toHaveProperty('x')    // throws TypeError (should pass)
expect(undefined).toHaveProperty('x')   // throws TypeError (should fail cleanly)
```

**Root cause:** in `packages/expect/src/jest-expect.ts`, `toHaveProperty`'s `getValue()` helper calls `Object.hasOwn(actual, propertyName)` before the null-safe `getPathInfo` fallback. `Object.hasOwn(null, …)` throws, and `actual` is never guarded for nullish values.

**Fix:** guard the own-property fast path with `actual != null` so the null-safe `getPathInfo` fallback runs. A nullish subject now fails with a proper matcher error (`expected null to have property "x"`), and the negated form passes — matching Jest's behavior.

Resolves #10735

### Tests
- [x] Added a regression test in `test/unit/test/jest-expect.test.ts` covering nullish subjects for both `toHaveProperty` and `.not.toHaveProperty`. Verified it fails on `main` and passes with this change. Ran the `jest-expect` suite, `eslint`, and `tsc` typecheck locally — all green.

### Documentation
- [x] No new functionality; bug fix only, no docs changes needed.

### Changesets
- [x] PR title is prefixed with `fix:` and describes the change.

### DIFF
--- a/packages/expect/src/jest-expect.ts
+++ b/packages/expect/src/jest-expect.ts
@@ -469,10 +469,8 @@ export const JestChaiExpect: ChaiPlugin = (chai, utils) => {
       const actual = this._obj as any
       const [propertyName, expected] = args
       const getValue = () => {
-        const hasOwn = Object.hasOwn(
-          actual,
-          propertyName,
-        )
+        const hasOwn
+          = actual != null && Object.hasOwn(actual, propertyName)
         if (hasOwn) {
           return { value: actual[propertyName], exists: true }
         }
--- a/test/unit/test/jest-expect.test.ts
+++ b/test/unit/test/jest-expect.test.ts
@@ -384,6 +384,23 @@ describe('jest-expect', () => {
     }).toThrowErrorMatchingInlineSnapshot(`[AssertionError: expected { a: { b: { c: 1 } } } to deeply equal { a: { b: { c: 2 } } }]`)
   })
 
+  it('toHaveProperty on null/undefined subjects (#10735)', () => {
+    // A nullish subject must fail cleanly instead of throwing a raw
+    // `TypeError: Cannot convert undefined or null to object`.
+    expect(null).not.toHaveProperty('x')
+    expect(undefined).not.toHaveProperty('x')
+    expect(null).not.toHaveProperty('a.b')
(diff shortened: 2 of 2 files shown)


xianjianlf2:
代码使用cc 来生成的，修复了  expect(null).toHaveProperty('x'), Vitest 没有给出正常的断言失败信息，直接抛出原始的内部错误

PR #10749, opened 2026-07-09
perf: scale reusable pool workers adaptively

sheremet-va:
### Description

Starts pool workers adaptively instead of one per queued task up front. For reusable workers (`isolate: false`, and vm pools' shared workers) the upfront strategy pays `maxWorkers × (spawn + runtime bring-up + environment setup)` even when a couple of reused workers would drain the queue sooner. Follows the same idea as the browser pool's adaptive sessions (#10726), which @hi-ogawa pointed out is a universal scheduling optimization that should transfer to the process pools.

- The first `min(maxWorkers, 4)` workers start in parallel, exactly like before — any queue that reaches this point justifies them, and it keeps small machines (CI runners have 4 vcpus, so `maxWorkers` is 3) on the previous behavior byte for byte.
- Beyond that, workers ramp up by doubling (in-flight starts never exceed already-running workers), so a long queue still reaches `maxWorkers` in logarithmic time.
- Once a per-task cost signal exists, another worker only starts when the projected queue drain time across the running workers exceeds twice the measured worker start cost. The signal is an EMA over reused workers' tasks — a worker's first task pays the environment bring-up on top of the test, so it doesn't count — and resets per task group, since e.g. a typecheck group after a unit group has a completely different per-task cost.

Only tasks with `isolate: false` are gated. Isolated tasks pay a worker per task no matter what — there is no avoidable bring-up cost for the scaling to save — so they keep the exact previous behavior.

### Benchmarks

macOS (M4, 10 cores → `maxWorkers` 9), 40/160-file fixtures, jsdom environment unless noted, interleaved A/B (median of 3, same build):

| cell | before | after | Δ |
|---|---:|---:|---:|
| forks `isolate:false`, 40 files | 1085ms | 936ms | **−14%** |
| threads `isolate:false`, 40 files | 1005ms | 883ms | **−12%** |
| forks `isolate:false`, node env, 40 files | 568ms | 532ms | −6% |
| forks `isolate:false`, 160 files, 1022 modules | 3007ms | 3074ms | ~flat |
| forks/threads `isolate:true`, vmThreads | — | — | ~flat (not gated / cost justifies max) |

CI-sized runners (4 vcpus → `maxWorkers` 3, same-runner interleaved A/B, node env, median of 5):

| cell | windows before | windows after | Δ | ubuntu Δ |
|---|---:|---:|---:|---:|
| forks `isolate:false`, 40 files | 827ms | 846ms | +2.3% (noise) | −0.6% |
| forks `isolate:true`, 40 files | 3946ms | 3981ms | +0.9% | −0.5% |
| forks `isolate:false`, 160 files | 1378ms | 1381ms | +0.2% | −0.1% |
| threads `isolate:false`, 40 files | 850ms | 855ms | +0.6% | −0.5% |

The win scales with core count and per-worker bring-up cost (environment setup dominates it for jsdom), so the beneficiaries are many-core machines; small runners stay on the previous behavior by construction.


packages/vitest/src/node/pools/pool.ts
```diff
@@ -121,16 +148,30 @@ export class Pool {
...
+            clearTimeout(id)
+            this._startingCount--
+          })
+
+        if (!resolver.isRejected) {
+          this._spawnCost = performance.now() - startedAt
+        }
+
+        // the next worker can start while this one runs its task
+        void this.schedule()
```
AriPerkkio, line 170: Looks a lot like https://github.com/vitest-dev/vitest/pull/8755

packages/vitest/src/node/pools/pool.ts
```diff
@@ -38,6 +38,21 @@ export class Pool {
...
+  // Workers start one at a time while the queue justifies another one
+  // (mirrors the browser pool's adaptive session scaling): every worker pays
+  // a spawn + runtime bring-up that competes with the already-running workers
+  // for the same CPUs and Vite server, so fast suites finish sooner with
+  // fewer workers, while suites with slower files still scale up to
+  // `maxWorkers`.
+  private _startingCount = 0
+  // refined with the measured duration after every worker start
+  private _spawnCost = 150
+  // EMA of how long a single task takes; undefined until the first signal,
```
AriPerkkio, line 50: What's EMA?

AriPerkkio (commented):
All this calculation seems complex. What does it do in practice? I didn't understand most of this PR based on the generated description.

hi-ogawa (changes requested):
The browser side PR was partly reverted https://github.com/vitest-dev/vitest/pull/10753. How heuristics can fail is pictured in https://artifacts.hiro18181.workers.dev/vitest-pr-10726-adaptive-sessions#failure-mode

This PR changes on node pool has slightly different code flow, but it should have effectively same characteristics where once it caps spawning new worker, it cannot grow back. This is because, while this PR changed so that one `schedule` can call two `schedule`, this "schedule ramp up chance" is given only to new worker (first we have 4 chances), so once each "chance" misses scaling opportunity due to small `fileCost` for earlier tests, it doesn't grow back even when later `fileCost` turn out to be large.

We should craft a similar adversarial examples like https://github.com/vitest-dev/vitest/pull/10753 to verify.

PR #10750, opened 2026-07-09
feat!: inline projects extend the root config by default

### LINKED ISSUE #7225: [v3] Automatically inherit common test config in inline projects

Project configurations can be repetitive, it would be nice for us and for Vitest to have a clear path towards defining what is common and what is specific to that project in one place. Inline projects are a great tool for that, I'm so thankful they made it into Vitest.

However, extending doesn't quite work as I expected, here I expected the project to inherit `globals: true`:

```ts
{
  test: {
    globals: true,
    workspace: [
      {
        test: {
          name: "node",
          environment: "node",
        },
      },
    ],
  },
}
```

because otherwise any config defined at the root `test` is a noop when `workspace` exists, right?

Next I tried setting `extends: true` on the project to get it to work, but it didn't. I don't really know what it does by default, I suppose it's looking up the directory tree.

What did work was configuring `extends: "./vitest.config.ts"`.

### Suggested solution

I would like not having to configure `extends` on projects, I would like inheritance to just happen. If that makes sense for the direction of v3.

### Alternative

Type definitions could be altered in a way that prohibits noop root configs in case `workspace` is defined, using a discriminated union.

### Additional context

_No response_

### Validations

sheremet-va:
### Description

`extends` now defaults to `true`: inline projects inherit all options from the root config, including plugins. Set `extends: false` to keep the project isolated.

- `name` and `projects` are never inherited. `globalSetup` is not inherited from the root config because it already runs once per test run (extending a non-root config still inherits it). The project's own `tags` replace the inherited array.
- Extending projects also inherit the programmatic config passed to `createVitest`/`startVitest`, so inheritance works even when there is no root config file. Root `plugins` from the programmatic config are not inherited because their instances are owned by the root server. The programmatic `browser` configuration is also not inherited: its instances describe a single project, and inheriting them would create duplicate instance names (the `--browser` flags have the same guard).
- The hardcoded `experimental` options propagation is removed; the values are inherited like everything else, and `extends: false` discards them. `--experimental.*` CLI flags are applied to every project like the other project-level CLI flags.

Projects referenced as config files or directories are not affected. Documented in the migration guide and the projects guide.

Resolves #7225

---

Stacked on #10823, which deduplicates the run-level option copies and adds the "Project Configuration Resolution" section to the advanced guide; this PR updates that section for the new semantics.


packages/vitest/src/node/projects/resolveProjects.ts
```diff
@@ -340,20 +344,99 @@ async function resolveDeclaredProjectEntries(
...
+  // must never be un-merged
+  const isInlineEntry = typeof workspacePath === 'number'
+  const inheritsRootConfig = isInlineEntry
+    && options.extends !== false
+    && typeof options.extends !== 'string'
+  // `extends: './path'` can still point back to the root config file
+  const extendsRootConfig = inheritsRootConfig
+    || (!!configFile && configFile === globalViteConfig.configFile)
+
+  // programmatic overrides are part of the effective root config, so an
```
sheremet-va: Something weird is going on here...

packages/vitest/src/node/projects/resolveProjects.ts
```diff
@@ -345,20 +351,100 @@ async function resolveDeclaredProjectEntries(
...
+    },
+    api: {
+      vitest: {
+        ignoreFsModuleCache: true,
+      },
+    },
+  }
+}
+
+/**
```
sheremet-va, line 396: It is a bit of a mind fck, but I don't see a better way. `viteOverrides` were a mistake!

docs/api/advanced/plugin.md
```diff
@@ -117,11 +117,11 @@ Note that this will only affect projects injected with [`injectTestProjects`](#i
 :::
 
 ::: tip Referencing the Current Config
-If you want to keep the user configuration, you can specify the `extends` property. All other properties will be merged with the user defined config.
+Inline configurations inherit the root config by default. If you want to inherit the configuration of a specific project instead, set the `extends` property to its config file. All other properties will be merged with the user defined config.
```
macarie: I'm not sure the relationship between "a specific project" and "its config file" (like why is it a different config file?) is clear, but I have no idea how to word it differently...

macarie (dismissed):
Looks good from what I can tell _(given my limited knowledge of this part)_ 👍🏼

docs/api/advanced/plugin.md
sheremet-va: Changed to

> If you want to inherit a specific configuration file instead, set the `extends` property to its path

sheremet-va:
The plan after this PR:

- support projects inside non-root configs ("nested projects")
- share the vite server between projects that don't define settings that affect the server

macarie (approved):

PR #10753, opened 2026-07-10
revert: adaptive browser session scaling (#10726)

sheremet-va:
Reverts the adaptive session opening from #10726 back to upfront `min(maxWorkers, files.length)` opens. The test-file transform pre-warm from that PR is independent of scheduling and is kept.

The heuristic can under-provision sessions when the files it measures early are much faster than the files still queued (raised by @hi-ogawa in the #10726 review): the projected queue drain `queue × fileCostEMA / sessions` underestimates the remaining work, the pool stops opening tabs, and the slow tail runs in extra serial waves. Once the scaling loop stops, nothing restarts it for the rest of the run, so the penalty scales with the hidden tail's duration.

Measured impact (10-core mac, chromium headless, cold runs, interleaved reps):

- Adversarial ordering (6 trivial files followed by 9×3s files): adaptive 7.09s vs upfront 4.26s (**+66%**).
- Real suites with the #10726 scheduler backported into their own vitest 4.1.x, so only scheduling differs: vueuse (85 browser test files) 7004ms upfront vs 7125ms adaptive (neutral); bootstrap.native (12 files) 7860ms vs 7466ms (adaptive −5%).
- Replayed per-file duration profiles of 6 OSS suites through the real pool (cold size-desc ordering reproduced): adaptive −0.1% to −12%.

The measurable upside on real suites is small, while the worst case grows with the duration of the mispredicted tail. Correcting the false cap needs in-flight cost lower bounds plus re-scaling machinery, which measured ~0 improvement on real suites — the simple upfront behavior wins on risk vs complexity.

### DIFF
--- a/packages/vitest/src/node/pools/browser.ts
+++ b/packages/vitest/src/node/pools/browser.ts
@@ -187,16 +187,6 @@ class BrowserPool {
   private _promise: DeferPromise<void> | undefined
   private _providedContext: string | undefined
 
-  private _scaling: Promise<void> | undefined
-  // EMA of how long a single file takes in this pool; undefined until the
-  // first file finishes, which means "no signal yet — keep opening sessions"
-  private _fileCostEma: number | undefined
-  // refined with the real duration after every session open
-  private _sessionOpenCost = 250
-  // a session's first file pays the tester bootstrap on top of the test
-  // itself, so it would wildly overestimate the steady per-file cost
-  private _warmedUpSessions = new Set<string>()
-
   private readySessions: Set<string>
 
   private _traces: Traces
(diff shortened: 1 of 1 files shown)


hi-ogawa (approved):
This is the effective diff after revert with https://github.com/vitest-dev/vitest/pull/10726 (6db8de2e2 is merge commit of the PR).

```diff
$ git diff 6db8de2e2^ revert-browser-adaptive-sessions -- packages/vitest/src/node/pools/browser.ts
diff --git a/packages/vitest/src/node/pools/browser.ts b/packages/vitest/src/node/pools/browser.ts
index 8c92d037c..9a846d835 100644
--- a/packages/vitest/src/node/pools/browser.ts
+++ b/packages/vitest/src/node/pools/browser.ts
@@ -373,6 +373,12 @@ class BrowserPool {
     const orchestrator = this.getOrchestrator(sessionId)
     debug?.('[%s] run test %s', sessionId, file)
 
+    // warm the transform cache while the iframe is booting so the test
+    // file import doesn't wait for the transform; mirrors the URL the
+    // tester will request (see `importFile` in the browser runner)
+    const fileUrl = `/${/^\w:/.test(file.filepath) ? '@fs/' : ''}${file.filepath}`.replace(/\/+/g, '/')
+    void this.project.vite.transformRequest(fileUrl).catch(() => {})
+
     this.setBreakpoint(sessionId, file.filepath).then(() => {
       // this starts running tests inside the orchestrator
       const testersPromise = this._traces.$(
```

PR #10757, opened 2026-07-11
feat!: enable mocking Temporal without fake timers

### LINKED ISSUE #10345: `Temporal.Now` is not mocked when timer mocks are enabled
### Describe the bug

`Temporal.Now` is not mocked when `vi.setSystemTime` is called (both with and without `vi.useFakeTimers()`)

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-agrzhfx6?file=test%2Ftime.test.ts

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    @vitest/ui: latest => 4.1.6 
    vite: latest => 8.0.12 
    vitest: latest => 4.1.6
```

### Used Package Manager

npm

### Validations

fabon-f:
### Description

Resolves #10345 

#10654 added support for mocking Temporal with `useFakeTimers`, but `setSystemDate` without fake timers still mocked only `Date` (it's a nice-to-have feature, cf. https://github.com/vitest-dev/vitest/pull/10654#issuecomment-4787538247).

This pull request makes `setSystemDate` mock `Temporal.Now` if it is available on the global namespace, in addition to `Date`.

### Tests

### Documentation

### Changesets


hi-ogawa (commented):
I'm not sure we like to bring another own custom "mock date" machinery like `packages/vitest/src/integrations/mock/temporal.ts`. 

I'm wondering if we can drop our `mock/date.ts` too now and go all in with fake timer. Basically making `setSystemDate` to be a sugar of something like:

```js
// setSystemDate(date)
// =>
vi.useFakeTimers({
  toFake: ['Date', 'Temporal'],
  now: date
})
```

Probably we still keep a internal state matrix of `_fakingTime x _fakingDate` but switch to the above sugar for `_fakingTime = false, _fakingDate = truthy` scenario. 

https://github.com/vitest-dev/vitest/blob/ad065a7d62af6ff5818197f3424f17c54d238ba3/packages/vitest/src/integrations/mock/timers.ts#L21-L28

Such change doesn't look much breaking.

fabon-f:
I updated the code in line with the suggestion.

hi-ogawa (approved):
Thanks for follow-up!

Regarding the breaking change aspect of the sugar idea, on the contrary to what I thought, it looks like this is effectively zero breaking change and basically pure refactoring to borrow fake timer `Date` implementation.

What I was wrongly thinking is that, the change may surface bare `vi.setSystemDate` would make `vi.isFakeTimers = true` and thus it may kicks in `expect.poll/vi.waitFor/until`'s fake timer auto advances, but actually we still keep internal states `_fakingTime/_fakingDate` so `isFakeTimers` is still false. So, this still hold true:

```ts
vi.setsystemtime(0)
await vi.waitFor(() => sleep(100))
console.log(Date.now()) // => 0 (doesn't auto advance still)
```

fabon-f:
Note that we no longer need to keep `_fakingDate` by ourselves because we can get it from `this._clock.now`. Vitest have to remember just whether fake timers are enabled, only `Date` and `Temporal` are mocked, or nothing is mocked.

sheremet-va (approved):

